PR #652, opened 2023-11-08
Fix code formatting in f.d.q.subclass-test ns

cap10morgan:

PR #662, opened 2023-11-10
Fix subclassOf inferencing

cap10morgan:
Fixes https://github.com/fluree/core/issues/48

### DIFF
--- a/src/fluree/db/json_ld/vocab.cljc
+++ b/src/fluree/db/json_ld/vocab.cljc
@@ -1,11 +1,11 @@
 (ns fluree.db.json-ld.vocab
-  (:require [fluree.db.flake :as flake]
-            [fluree.db.constants :as const]
-            [fluree.db.util.async :refer [<? go-try]]
+  (:require [fluree.db.constants :as const]
+            [fluree.db.flake :as flake]
+            [fluree.db.json-ld.ledger :as jld-ledger]
             [fluree.db.query.range :as query-range]
-            [fluree.db.util.schema :as schema-util]
-            [clojure.set :as set]
-            [fluree.db.json-ld.ledger :as jld-ledger]))
+            [fluree.db.util.async :refer [<? go-try]]
+            [fluree.db.util.log :as log]
+            [fluree.db.util.schema :as schema-util]))
 
 #?(:clj (set! *warn-on-reflection* true))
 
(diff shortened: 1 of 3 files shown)


cap10morgan:
Current dependencies on/for this PR:
* main
  * **PR #652** <a href="https://app.graphite.dev/github/pr/fluree/db/652?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #653** <a href="https://app.graphite.dev/github/pr/fluree/db/653?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #661** <a href="https://app.graphite.dev/github/pr/fluree/db/661?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #662** <a href="https://app.graphite.dev/github/pr/fluree/db/662?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/fluree/db/662?utm_source=stack-comment).

src/fluree/db/json_ld/vocab.cljc
```diff
@@ -243,29 +243,62 @@
...
+                    (log/trace "filter-vocab-flakes adding" f "b/c its sid matches a ref flake's o")
+                    (set-conj vfs f))
+
+                  (class-or-prop-flake-sids fs)
+                  (do
+                    (log/trace "filter-vocab-flakes adding" f "b/c its sid matches a type flake's sid")
+                    (set-conj vfs f))
+
+                  :else vfs)))
+            ref-flakes all-flakes)))
```
dpetran, line 301: If we do a `(or ref-flakes #{})` here we can save the overhead of a `nil` check on every `conj`.

src/fluree/db/json_ld/vocab.cljc
```diff
@@ -243,29 +243,62 @@
...
+  [predicate-flakes all-flakes]
+  (let [ref-flakes               (::ref predicate-flakes)
+        class-or-prop-flakes     (::class-or-prop predicate-flakes)
+        ref-flake-os             (->> ref-flakes (map flake/o) set)
+        class-or-prop-flake-sids (->> class-or-prop-flakes (map flake/s) set)
+        predicate-sids           (->> all-flakes (map flake/p) set)
+        set-conj                 (fnil conj #{})]
+    (reduce (fn [vfs f]
+              (let [fs (flake/s f)]
+                (cond
```
dpetran, line 284: Do all of the cases in this `cond` do the same thing?

dpetran (approved):
I've got some questions about code organization but this looks good.

I think I would prefer a lot less trace logging - when I'm debugging I tend to prefer very granular/bespoke logging for my specific problem , rather than the firehose of every step. But it's not a blocker by any means.

src/fluree/db/json_ld/vocab.cljc
cap10morgan: Yeah except for the logging, which I guess we can drop now. :)

dpetran: I think it's useful context, so maybe they can live on as comments?

cap10morgan:
> I've got some questions about code organization but this looks good.
> 
> I think I would prefer a lot less trace logging - when I'm debugging I tend to prefer very granular/bespoke logging for my specific problem , rather than the firehose of every step. But it's not a blocker by any means.

This is the same logic that motivated changing logs like these from debug to trace not too long ago. We need to figure out our logging philosophy and overall approach. To me, `trace` implies a firehose. But besides other log levels, there are also ns filters. But regardless of my opinions / understanding, consensus on an overall team-wide approach would be good here. In the meantime I'll probably just start removing these from most of my PRs because I have no idea what we want to end up in the shipped codebase.

zonotope:
> > I've got some questions about code organization but this looks good.
> > I think I would prefer a lot less trace logging - when I'm debugging I tend to prefer very granular/bespoke logging for my specific problem , rather than the firehose of every step. But it's not a blocker by any means.
> 
> This is the same logic that motivated changing logs like these from debug to trace not too long ago. We need to figure out our logging philosophy and overall approach. To me, `trace` implies a firehose. But besides other log levels, there are also ns filters. Regardless of my opinions / understanding, consensus on an overall team-wide approach would be good here. In the meantime I'll probably just start removing these from most of my PRs because I have no idea what we want to end up in the shipped codebase.

I agree with @cap10morgan that TRACE implies a firehose, but I think I also agree with @dpetran than having all of these logs from all of these namespaces makes the logging system much less useful in general. I think moving these logs from DEBUG to TRACE is much better than what we had before in development, but I still find myself editing logback-test.xml locally to set the log level to DEBUG when I run my tests because the default log level when we run our tests is still TRACE. 

When I have a test failure and get a wall of irrelevant logs in my terminal, it's impossible for me to find what the actual cause of the failure was. A lot of the time there would be so many logs printed out that the scrollback buffer for my terminal fills up and I can't even see the actual test failure. I usually have to edit logback-test.xml at that point to set the log level higher and then add my own logs to pinpoint on the cause of the problem. I had been meaning to bring this up, but just haven't gotten around to it yet.

I'm fine with keeping all of these "stepping through the code" logs at trace level, but I think we should change the default log level for tests to DEBUG like we have in development. I don't think TRACE should be enabled by default anywhere, but it might be useful in some situation to be able to trace the code execution step by step, so we should still have the option.

So, in short, my preference would be that the default log level is DEBUG for test and development profiles, and INFO for production, and that TRACE isn't enabled by default anywhere, but we still have the option to enable it through the logging config.

PR #669, opened 2023-11-13
Add extra condition and logging for fluree/notify

bplatz:
This adds extra logging condition for calls to fluree/notify, and also will log out if supplied map doesn't look like a commit map to alert the user.


dpetran (approved):
:deciduous_tree:

PR #670, opened 2023-11-13
Adds connect-file CLJ API to default file connection

bplatz:
This creates a `connect-file` API similar to the existing `connect-ipfs` and `connect-memory` that allows a file connection to work by default without any options passed (but options passed will still be used if provided).

I think it could be (and welcome) debate over if we should have this or not, but I personally find it nice and convenient - and we have other helpers but this one was missing. If we don't want it, then I think we should remove all the helpers.

So this will work:
`(def conn @(fluree/connect-file nil))`

### DIFF
--- a/src/fluree/db/conn/file.cljc
+++ b/src/fluree/db/conn/file.cljc
@@ -219,7 +219,8 @@
 (defn connect
   "Create a new file system connection."
   [{:keys [defaults parallelism storage-path lru-cache-atom memory serializer nameservices]
-    :or   {serializer (json-serde)} :as _opts}]
+    :or {serializer (json-serde)
+         storage-path "data/ledger"} :as _opts}]
   (go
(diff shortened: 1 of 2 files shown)


dpetran (approved):
This change looks fine.

I think I'm on the side of remove-all-the-helpers, for a few reasons:

I don't think they provide a ton of convenience, and it seems the main usage is to pass `nil` as an arg, which always makes me want to peek at the implementation anyways to see what in the world is going on that the function needs a `nil` argument.

Plus they add maintenance overhead in terms of documentation, and in propagating changes out to all the other sdks.

And I have a general preference for smaller apis so less scrolling is necessary to find the functions I care about.

In all, I don't think the convenience they add is really worth it. But I don't feel so strongly about it that if someone else thinks these costs are worth the benefit that I'd insist on their removal.

PR #674, opened 2023-11-14
relax context requirement

dpetran:
Fixes https://github.com/fluree/core/issues/54

Users no longer need to supply the fluree context in their transactions, unless they are wrapping them in a verifiable credential.

### DIFF
--- a/src/fluree/db/util/context.cljc
+++ b/src/fluree/db/util/context.cljc
@@ -198,13 +198,24 @@
 (defn txn-context
   "Remove the fluree context from the supplied context."
   [txn]
-  (validate-txn-context txn)
-  (let [supplied-context (->> (get txn "@context" (:context txn))
-                              (util/sequential)
-                              (remove #{"https://ns.flur.ee"}))]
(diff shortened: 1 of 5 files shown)


src/fluree/db/util/context.cljc
```diff
@@ -198,14 +198,25 @@
...
   [txn]
-  (validate-txn-context txn)
-  (let [supplied-context (->> (get txn "@context" (:context txn))
-                              (util/sequential)
-                              (remove #{"https://ns.flur.ee"}))]
+  (let [supplied-context (when (or (contains? txn :context)
+                                   (contains? txn "@context"))
+                              (->> (get txn "@context" (:context txn))
+                                   (util/sequential)
+                                   (remove #{"https://ns.flur.ee"})))]
```
zonotope, line 205: what happens if the user also supplies the fluree context?

dpetran: We remove it. This is the status quo, though, and this commit doesn't change it. Once we properly remove default context this hack will no longer be necessary.

zonotope (approved):
🌵

PR #675, opened 2023-11-14
Add a test for queries w/ faux compact IRI

cap10morgan:
Bug report but it was already fixed. I figured the test was a good addition against any regressions.


mpoffald (approved):
🔖

PR #684, opened 2023-11-15
Task/internal renaming

dpetran:
This is just a followup to the dead code removal - we can remove the suffixes from all the functions that no longer need them.

I also moved a function to the only namespace where it used.


cap10morgan (approved):
📛

PR #685, opened 2023-11-17
adjust db boostrap

dpetran:
This fixes an error where loading a ledger with a new FileConnection conn was throwing an error. The error was being thrown because a newly initialized db had a `:t` of -1, and when it tried to apply the first commit with a `:t` of -1it threw an error about the "gap" between the commit and the db.

The error was confusing because it mentioned a gap and then showed two identical t values, so I updated that to be more useful.

And I fixed it by changing an initialized db to have a `:t` of 0. This was the old behavior, before I changed it last week because transactions were unable to look up any of the base flakes while the db `:t` was greater than the flake `t`, which is only the case on the initial tx. But then I was able to narrow our base flakes down to just the identity flake, which does not need to be looked up during transactions, only during queries. So the problem of a 0 `:t` doesn't exist anymore.

I was also able to clean up some obsolete bootstrap code and all that was left was a call to `commit-data/update-novelty`, so I just moved it into the db creation function.


mpoffald (approved):
🥄

PR #693, opened 2023-11-18
Expand policy opts role using supplied context

mpoffald:
Part of https://github.com/fluree/core/issues/61

Up to now, the `wrap-policy` process has relied entirely on default context to expand the iri of the provided `:role` and look up its subject id. The call it has been making to `subid` has a  default behavior that expands using the db's default context.

This PR updates our policy opts helper (renamed to `policy-identity`) to expand the iri itself when given a context. Also updates our api fns to pass the supplied context from the request to that helper fn.

I removed the use of default context from a large number of the policy tests, but not all of them.

I did not want to mess with removing any existing default context logic yet (and have to change _many_ unrelated tests), so there's some redundancy here. I have introduced a context helper that just pulls out the supplied context from the request map, with no interaction with the default context whatsoever. I just use this when processing opts, leaving existing context retrieval/calculation processes (which do use default context) intact.

### DIFF
--- a/src/fluree/db/json_ld/policy.cljc
+++ b/src/fluree/db/json_ld/policy.cljc
@@ -493,8 +493,9 @@
   [{:keys [context] :as identity-map}]
   (when-let [{:keys [role] :as identity} (-> (select-keys identity-map [:did :role :credential])
                                              not-empty)]
-    (cond-> identity
-      (and role context) (update :role json-ld/expand-iri (json-ld/parse-context context)))))
+    (if (and role context)
+      (update identity :role json-ld/expand-iri (json-ld/parse-context context))
+      identity)))
 
 (defn role-sids-for-sid
   [db sid]
--- a/src/fluree/db/json_ld/policy.cljc
+++ b/src/fluree/db/json_ld/policy.cljc
@@ -1,6 +1,7 @@
 (ns fluree.db.json-ld.policy
   (:require [clojure.core.async :as async]
             [fluree.db.constants :as const]
+            [fluree.json-ld :as json-ld]
             [fluree.db.util.async :refer [<? go-try]]
             [fluree.db.dbproto :as dbproto]
             [fluree.db.json-ld.policy-validate :as validate]
@@ -488,10 +489,12 @@
          (throw e))))))
 
 
-(defn policy-opts
-  [opts]
-  (-> (select-keys opts [:did :role :credential])
-      not-empty))
+(defn policy-identity
+  [{:keys [context] :as identity-map}]
+  (when-let [{:keys [role] :as identity} (-> (select-keys identity-map [:did :role :credential])
+                                             not-empty)]
+    (cond-> identity
+      (and role context) (update :role json-ld/expand-iri context))))
 
 (defn role-sids-for-sid
   [db sid]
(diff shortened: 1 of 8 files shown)


src/fluree/db/json_ld/policy.cljc
```diff
@@ -488,10 +489,12 @@
          (throw e))))))
 
 
-(defn policy-opts
-  [opts]
-  (-> (select-keys opts [:did :role :credential])
-      not-empty))
+(defn policy-identity
+  [{:keys [context] :as identity-map}]
+  (when-let [{:keys [role] :as identity} (-> (select-keys identity-map [:did :role :credential])
```
cap10morgan, line 494: Just a very nitpick-y style thing you can take or leave, but the threading macro might be doing its job* even better if `identity-map` alone were the first arg.

* Making nested forms more readable for humans

src/fluree/db/json_ld/policy.cljc
```diff
@@ -488,10 +489,12 @@
...
+                                             not-empty)]
+    (cond-> identity
+      (and role context) (update :role json-ld/expand-iri (json-ld/parse-context context)))))
```
cap10morgan, line 497: Another take it or leave it thing (and this is even more personal preference than style): I'm not a huge fan of using `cond->` w/ only one condition. I think plain ol' `if` `else` is much more readable for these scenarios, especially for outside contributors.

test/fluree/db/policy/basic_test.clj
```diff
@@ -184,30 +194,47 @@
...
-                                          :commit-details true
-                                          :opts           {:did  alice-did
-                                                           :role :ex/userRole}}))
-              "Alice should not be able to see any history for John's ssn")
+          (testing "query-connection"
+            (is (= [["Alice" "111-11-1111"] ["John" nil]]
+                   @(fluree/query-connection conn
+                                             {:context context
+                                              :from "policy/a"
+                                              :select '[?name ?ssn]
```
cap10morgan, line 212: `"policy/a"` is misaligned w/r/t the other map vals

test/fluree/db/policy/basic_test.clj
```diff
@@ -184,30 +194,47 @@
...
+                                              :where  '[{:id          ?p
+                                                         :schema/name ?name}
+                                                        [:optional {:id         ?p
+                                                                    :schema/ssn ?ssn}]]
+                                              :opts   {:did  alice-did
+                                                       :role :ex/userRole}}))
+                "Both user names should show, but only SSN for Alice"))
+          (testing "history query"
+            (is (= []
+                   @(fluree/history ledger {:context context
```
cap10morgan, line 222: map val alignment

cap10morgan (approved):
Looks good! Since we'd been discussing it last week, I posted some style / code readability nitpick-y stuff (read: cranky opinions). Feel free to ignore! :)

src/fluree/db/json_ld/policy.cljc
zonotope: agreed

zonotope (approved):
📑

src/fluree/db/json_ld/policy.cljc
mpoffald: For some reason I'm not following what the suggestion is, what should the change be?

mpoffald: Oh wait, I think I get it, you mean on its own as the first arg to `->`

PR #694, opened 2023-11-18
Add pending test for missing predicate in policy-enforced query result

mpoffald:
Related bug report: https://github.com/fluree/core/issues/61

### DIFF
--- a/test/fluree/db/policy/basic_test.clj
+++ b/test/fluree/db/policy/basic_test.clj
@@ -392,7 +392,7 @@
                                 :opts {:role "ex:userRole"
                                        :did alice-did}}))))))
 
-(deftest ^:pending ^:integration identity-equals-test
+(deftest ^:integration identity-equals-test
   (let [conn         @(fluree/connect {:method :memory})
         context      {"ex"     "http://example.org/"
                       "schema" "http://schema.org/"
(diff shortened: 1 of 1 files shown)


mpoffald:
@aaj3f has reported that the bug in question seems to be resolved (https://github.com/fluree/core/issues/61#issuecomment-1828332917). So I figured I'd remove the `^:pending` tag and PR this as a passing test, but it still fails for me in the same way as the original bug report: https://github.com/fluree/db/actions/runs/7011610886/job/19074589954#step:4:168

I'm going to convert this PR to a draft for now while I figure out what's going on. Eventually I will either:
- update the test (if maybe I'm testing the wrong thing?)
- if there is a bug I'll reopen with a fix
- close this PR altogether

mpoffald:
I figured out the discrepancy between this failing test and @aaj3f report that it's working: this test transacts the policy first, before everything else. That causes the policy-enforced query to misbehave.

He sent over the steps he used to check on this bug, and he has been transacting the policy _after_ the data itself. That works fine, and is in fact how all of our policy tests work already.

It also works if you transact the identity first:
```clj
{"@context" {"ex" "http://example.org/"
             "schema" "http://schema.org/"
             "f"      "https://ns.flur.ee/ledger#"}
 "insert"   {"@id"     did
             "ex:user" {"@id" "ex:freddy"}
             "f:role"  {"@id" "ex:yetiRole"}}}
```
Followed by the policy, and then the data.

You just can't transact policy first.

Rearranging this test to make it pass basically replicates existing tests, so I'm going to close this PR. If we decide we need to support transacting the policy first in cases like this, then that can be separate work related to its own issue.

PR #695, opened 2023-11-26
Remove legacy v2 tag proto

bplatz:
This removes the db protocol for v2's -tag and -tag-id which is no longer relevant in v3.

In v2, 'tags', replicated namespaced IRIs (for tag uniqueness) and some sugar that shortened the longer namespaced names. v3 now fully embraces namespaced IRIs, and natively handles the v2 tag concept.

If a user wants to create "tags" in v3, each tag should just be an IRI, e.g. colors might be `[{"@id": "tag:Red"}, {"@id": "tag:Blue"}]`. @context can be used to shorten the tags in query results if desired (in this case, i.e.: `{"tag": "http://example.org/tags/"}`). If a specific property should be restricted to only allow a specific set of enumerated tags (enum), SHACL can be used to create this restriction.


zonotope (approved):
✂️

PR #696, opened 2023-11-28
Fix read-only f:allNodes policy

mpoffald:
Closes https://github.com/fluree/core/issues/65

We weren't building the `:policy` map properly in cases where permissions on `f:allNodes` only had one `f:action`. We were expecting `f:action` to always have a collection of actions in it, without ensuring that this was the case. This resulted in `nil` keys in the `:policy` map.


dpetran (approved):
:bullettrain_front:

PR #699, opened 2023-11-29
Add @type IRI flake for select clause projection

bplatz:
When select clause returns predicate matches, and the predicate is @type, the string value for @type was not being returned as there was no flake to resolve it. Instead, it was returning nil.


zonotope (approved):
🍃

PR #704, opened 2023-12-13
JWS support

dpetran:
https://github.com/fluree/core/issues/56

This adds basic support for JWS's as an auth method. This deliberately does not support EDN transactions or queries at this time, it assumes that the JWS payload is a json string. We can add EDN support in the future if necessary.

### DIFF
--- a/src/fluree/db/json_ld/credential.cljc
+++ b/src/fluree/db/json_ld/credential.cljc
@@ -1,13 +1,14 @@
 (ns fluree.db.json-ld.credential
   (:require [alphabase.core :as alphabase]
-            [fluree.db.util.async :refer [go-try]]
             [clojure.string :as str]
-            [fluree.db.util.core :as util #?(:clj :refer :cljs :refer-macros) [try* catch*]]
-            [fluree.crypto :as crypto]
-            [fluree.json-ld :as json-ld]
             #?(:cljs [cljs.core.async.interop :refer-macros [<p!]])
-            [fluree.json-ld.processor.api :as jld-processor]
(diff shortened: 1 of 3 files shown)


src/fluree/db/json_ld/credential.cljc
```diff
@@ -62,51 +59,64 @@
...
         ;; everything is good
-        {:subject subject :did auth-did}))))
+       {:subject subject :did auth-did}))))
+
+(defn verify-jws
+  [jws]
+  (let [{:keys [payload pubkey]} (crypto/verify-jws jws)
+        id                       (crypto/account-id-from-public pubkey)
+        auth-did                 (did/auth-id->did id)]
+    {:subject (json/parse payload false) :did auth-did}))
```
dpetran, line 116: Hardcoding `json/parse` of the payload precludes the use of edn strings in the payload. This can be fixed in the future by allowing users to specify the header for their JWS and the content-type for the payload. Since that adds a lot of complexity to the implementation that I'm not sure that we'll actually need I've not supported that here.

zonotope: We've done a lot of work to make fluree feel natural in Clojure, and Clojure developers will be one of the first populations we focus on early in the new year. We should at the very least have a ticket for enabling edn strings.

zonotope (approved):
🏛️

src/fluree/db/json_ld/credential.cljc
dpetran: I agree, I've created https://github.com/fluree/core/issues/66 to track it.

PR #705, opened 2023-12-20
Feature/store protocol

dpetran:
This introduces the Store protocol which is a very simple KV store api that should allow us to abstract over different storage methods.

This abstraction is useful because it allows us to avoid adding more content-specific methods to the conn like "-ctx-write" and "-ctx-read", or in this case, "-txn-write" and "-txn-read".

The Store protocol is an abstraction-in-progress. This PR just lays the basic foundation for different Stores, it will need more refinements as we move to support other storage methods and multiple storage methods at the same time.

Closes:
https://github.com/fluree/core/issues/57
https://github.com/fluree/core/issues/58


src/fluree/store/core.cljc
```diff
@@ -0,0 +1,69 @@
+(ns fluree.store.core
```
zonotope, line 1: I think we should move away from `*core` namespace names, and this should be under `fluree.db`. 

There's already a storage namespace. The protocol should probably live there.

src/fluree/db/conn/memory.cljc
```diff
@@ -35,63 +36,60 @@
...
+  [store address]
+  (read-data store address))
 
 (defn write-context!
-  [data-atom context-data]
-  (write-data! data-atom context-data))
+  [store context-data]
+  (write-data! store context-data))
 
 (defn read-context
```
zonotope, line 76: I didn't realize I left these functions in here. Not for this pr, but we should probably clean these up at some point.

src/fluree/db/nameservice/memory.cljc
```diff
@@ -21,82 +22,66 @@
...
+        (-> head-commit (get "address"))
+        (throw (ex-info (str "Unable to lookup ledger address from conn: "
+                             ledger-alias)
+                        {:status 500 :error :db/missing-head})))))
 
 (defn ledger-list
-  [state-atom opts]
-  (go (-> @state-atom keys)))
+  [store opts]
+  (go (->> (keys @(:storage-atom store))
```
zonotope, line 55: this breaks the abstraction barrier. there should be a list method on storage that takes a prefix that this function can use. nothing should be peeking inside of storage implementations and looking at the storage atom.

src/fluree/db/util/filesystem.cljc
```diff
@@ -78,6 +78,22 @@
...
                             "path"    (.-path e)})))))))
 
+(defn delete-file
+  "Delete the file at `path`."
+  [path]
+  #?(:clj
+     (try
+       (io/delete-file (io/file path))
+       :deleted
+       (catch Exception e
```
zonotope, line 88: i don't think we should be catching exceptions here so we can allow downstream consumers of this api to handle errors as they arise.

src/fluree/store/core.cljc
```diff
@@ -0,0 +1,69 @@
...
+   LocalStorageConfig
+   MemoryConfig])
+
+(defn start
+  [{:keys [:store/method] :as config}]
+  (if-let [config-error (m/explain StoreConfig config)]
+    (throw (ex-info "Invalid Store config."
+                    {:error  config-error
+                     :config config}))
+    (case method
```
zonotope, line 45: maybe not for this pr, we should use an open multi method instead of a closed case statement to decide which storage implementation to load. this allows for more flexibility in which implementations we support, even user supplied ones.

src/fluree/store/file.cljc
```diff
@@ -0,0 +1,51 @@
...
+(defn file-write
+  [storage-path k v {:keys [content-address?] :as opts}]
+  (when (not (store-util/hashable? v))
+    (throw (ex-info "Must serialize v before writing to FileStore."
+                    {:storage-path storage-path
+                     :k            k
+                     :v            v
+                     :opts         opts})))
+  (let [hash (crypto/sha2-256 v)
+        k*   (if content-address?
```
zonotope, line 17: we shouldn't have a content-address? option here, and we should also move away of embedding control flow in predicate? options in general. this is the point of protocols and records. the storage api needs to be content addressed, and if there's a use case that cannot be content addressed, then that use case should be supported by a different api because they are performing fundamentally different actions.

dpetran: Same as with the proto point, I'd like to defer this to the IPFS conn work in the interest of shipping https://github.com/fluree/core/issues/57 and https://github.com/fluree/core/issues/58

If you don't agree, I can absolutely do it as part of this, though.

src/fluree/store/proto.cljc
```diff
@@ -0,0 +1,8 @@
+(ns fluree.store.proto
```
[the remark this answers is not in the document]
dpetran: In the interest of getting https://github.com/fluree/core/issues/57 and https://github.com/fluree/core/issues/58 shipped, I'd like to defer this work to when we fix the IPFS conn, since that will involve additional refinements to the Store protocol.

bplatz:
I am trying to understand without spending too much time in the code if (a) this approach will be compatible with consensus (fluree/server), and (b) if this will be compatible with IPFS. I have some concerns after glancing it will be, but hard to tell. Do we know for sure this will all work with fluree/server which has a different approach towards storage? For IPFS, it is a key part of our next sprint so shipping something that breaks it means we can't do that work in the next sprint.

Our methods of `-c-write`, etc actually end up having a lot of logic they must implement and do differently depending on the circumstance. Downstream of those, ultimately a key-val storage works but it is downstream - not at this level. Again, might be how this works, but seems like it might not be from glancing at it.

When running consensus, you never "write" a commit and related files - consensus actually writes them, the transaction server itself does not. If the transaction server wrote the files and then sent it through consensus, you end up with inconsistent states which then require additional functionality try to clean it all up. This is the main issue fluree/server deals with, and I want to make sure this work doesn't end up breaking fluree/server and its associated guarantees.

There are downstream operations (like indexing) that are not part of consensus approval which are saving to have a simple storage writing protocol for however, or things like storing transactions for queued processing.

Again, hard to tell and it might all be fine, but I want to confirm we don't expect any issues for fluree/server (where consensus writes) and the complexities of IPFS, else this will not end up being code we can use.

dpetran:
> is this approach compatible with consensus?

Yes, I believe it's completely orthogonal to our approach to consensus.

> is this approach compatible with IPFS?

Yes, though it doesn't support IPFS right now. I've discussed with Ben approaches to properly extend the abstraction to work with IPFS/IPNS, and nothing in this PR will preclude IPFS support.

I'm not sure I understand what you mean by "consensus" writing the commits. My understanding is that our consensus mechanism ensures a total ordering of transactions, which are then processed in order by whatever cluster node is the leader for that ledger. The store protocol works completely downstream of consensus and is not meant to provide more than a storage method abstraction.

Per your point about `-c-write` doing work, it doesn't look like any of the current `-c-write` implementations are doing more than directly writing the commit to whatever storage mechanism is in play, so delegating this job to a proper protocol for it may allow us to consolidate Connection implementations in the future and hopefully reduce the risk of storage-method-specific bugs from divergent Conn implementations.

bplatz:
@dpetran thanks for the reply and good to hear. I guess I'd want to make sure nothing here was incompatible with IPFS or fluree/server.

As for consensus, the "leader" for a given ledger can think it is the leader when it is not due to several circumstances - typically due to some network latency. This ends up having two (or more) servers processing the same transaction, one that thinks it is the leader and one that actually is the leader. Because commits are not deterministic (although the underlying data changes should mostly be) we now have two commits for the same transaction.

If both servers write the commit, it means one server will end up having to "undo" its commit. Therefore, commit writing should happen via consensus where only one commit could ever be written.

Another scenario is in e.g. a two-server setup the leader creates a commit but the other server dies and never receives it. It once again means consensus cannot be reached however one server already wrote a commit, and would be inconsistent.

src/fluree/db/store.cljc
```diff
@@ -0,0 +1,74 @@
...
+                     :config config}))
+    (case method
+      :file         (file-store/create-file-store config)
+      :localstorage (localstorage-store/create-localstorage-store config)
+      :memory       (mem-store/create-memory-store config)
+
+      (throw (ex-info (str "No Store implementation for :store/method: " (pr-str method))
+                      config)))))
+
+(defn write
```
zonotope, line 54: As it stands now, these functions will be redundant when we move the protocol into this namespace. There might be more functionality we build into them later, but this redundancy is part of the reason the protocol should be here to begin with.

src/fluree/db/api/transact.cljc
```diff
@@ -19,20 +19,21 @@
           opts))
 
 (defn stage
-  [db txn opts]
+  [db txn {:keys [raw-txn] :as opts}]
```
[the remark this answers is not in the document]
dpetran: The raw-txn allows for us to store the initial incoming transaction without touching it at all. It's a bit awkward because we need to expand it in order to figure out which ledger to load, so we need to pass it down.

src/fluree/db/conn/file.cljc
```diff
@@ -68,14 +44,13 @@
                         (when branch (str "/" branch))
                         (str "/" type-dir "/")
                         hash ".json")
-        write-path (str (fs/local-path storage-path) "/" path)]
-    (log/debug (str "Writing " (name data-type) " at " write-path))
-    (fs/write-file write-path bytes)
+        ;; TODO: build migration to remove .json suffix so we can use :content-address? true
```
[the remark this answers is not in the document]
dpetran: We need non-content addressed writes for the ledger head currently. 

In order to handle this going forward, we'd need dual implementations of most of the store methods:
- FileStore
- ContentAddressedFileStore
- MemoryStore
- ContentAddressedMemoryStore
- etc.

The content-addressed versions of the store can be implemented in terms of the non-content addressed stores. The one exception is IPFS, which would have the IPFSStore and the IPNSSTore, where the IPFSStore would be used to write the commits and the IPNSStore would be used to write the head.

I'm not as opposed to having a boolean option meant for control flow, as it will save writing a ton of boilerplate, but I don't feel very strongly about it.

src/fluree/db/store.cljc
```diff
@@ -0,0 +1,74 @@
+(ns fluree.db.store
+  (:refer-clojure :exclude [read list exists?])
+  (:require [fluree.db.store.proto :as store-proto]
+            [fluree.db.store.file :as file-store]
+            [fluree.db.store.localstorage :as localstorage-store]
+            [fluree.db.store.memory :as mem-store]
+            [malli.core :as m]))
+
+
+(def BaseConfig
```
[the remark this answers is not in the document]
dpetran: Because it's data the schema can be composed into a larger config schema upstream to allow validation to happen at config ingestion:

```clojure
(def ConnectSchema
   {:storage/config store/StoreConfig
    :other/component component/ThatComponentConfig
    ;; etc.
  }
```
That would allow us to define the config relevant to storage right next to the storage, which I think is nice, but still use it wherever we want. That's what I like about malli!

src/fluree/db/conn/file.cljc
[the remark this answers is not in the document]
bplatz: > We need non-content addressed writes for the ledger head currently.
> 
> In order to handle this going forward, we'd need dual implementations of most of the store methods:

Potentially, but I think this part is the concern of the NameService. It is possible the NameService does need its own way of persisting - but it is also possible some nameservices are at a location different than data storage (e.g. Etherium as a  NameService, but IPFS as storage)

It is true that someone using FileStore will likely want the FileStore Nameservice and then both do write... and likewise IPFS will want to use IPNS for Nameservice (which relies on IPFS files) so two purposes.

But I'd rather see the NameServices implement their own methods for how they store... and that be abstracted from the protocol which looks more like `-lookup`, `-update`, etc.

src/fluree/db/store.cljc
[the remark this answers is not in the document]
zonotope: That needlessly couples code with configuration, which means any change to the format of the configuration file requires changes to the constructors that create the storage implementation and vice versa. it also spreads validation logic around the code that couples us to a particular validation library. 

Config validation should happen at the application boundary during initialization, and the same code that validates the config should then decompose the config and call the constructors as necessary. 

For example: there's no need for the `create-file-store` function to take a map. we should be able to initialize it with `(create-file-store "/some/path/")`, and the initialization code should know how to pull the right thing out of the config after it's validated and call the right function. this de-couples our configuration format from our internal application code. giving us more flexibility on both sides.

zonotope (approved):
🖼️

PR #716, opened 2024-01-19
allow users to specify annotations on commits

dpetran:
Similar to author and txn, this only applies to 1-to-1 stage/commit operations. The extra flake in the flake count is for the new commit schema flake.

Closes https://github.com/fluree/core/issues/78

### DIFF
--- a/src/fluree/db/json_ld/transact.cljc
+++ b/src/fluree/db/json_ld/transact.cljc
@@ -210,9 +210,10 @@
    (stage db nil txn parsed-opts))
   ([db fuel-tracker txn parsed-opts]
    (go-try
-     (let [ctx           (:context parsed-opts)
-           parsed-txn    (q-parse/parse-txn txn ctx)
-           db*           (if-let [policy-identity (perm/parse-policy-identity parsed-opts ctx)]
+     (let [{:keys [context raw-txn did]} parsed-opts
+
+           parsed-txn    (q-parse/parse-txn txn context)
+           db*           (if-let [policy-identity (perm/parse-policy-identity parsed-opts context)]
                            (<? (perm/wrap-policy db policy-identity))
                            db)
            tx-state      (->tx-state db*)
--- a/src/fluree/db/json_ld/transact.cljc
+++ b/src/fluree/db/json_ld/transact.cljc
@@ -1,23 +1,25 @@
 (ns fluree.db.json-ld.transact
-  (:require [clojure.core.async :as async :refer [go alts!]]
+  (:require [clojure.core.async :as async :refer [alts! go]]
             [fluree.db.constants :as const]
-            [fluree.db.json-ld.policy :as perm]
             [fluree.db.dbproto :as dbproto]
             [fluree.db.flake :as flake]
             [fluree.db.json-ld.branch :as branch]
             [fluree.db.json-ld.commit-data :as commit-data]
             [fluree.db.json-ld.ledger :as jld-ledger]
+            [fluree.db.json-ld.policy :as perm]
             [fluree.db.json-ld.shacl :as shacl]
             [fluree.db.json-ld.vocab :as vocab]
             [fluree.db.ledger.proto :as ledger-proto]
             [fluree.db.policy.enforce-tx :as policy]
-            [fluree.db.query.fql.parse :as q-parse]
             [fluree.db.query.exec :as exec]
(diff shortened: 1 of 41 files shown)


test/fluree/db/query/history_test.clj
```diff
@@ -959,3 +959,50 @@
                           (-> (get c "f:commit")
                               (update "f:data" select-keys ["f:t"])
                               (select-keys ["f:author" "f:txn" "f:data"]))))))))))
+
+(deftest ^:integration txn-annotation
```
zonotope, line 963: Could you add some English documentation to this test using both the `testing` macro to explain what's being tested, and the second argument to the `is` macro to describe the expected outcome?

test/fluree/db/query/history_test.clj
```diff
@@ -959,3 +959,50 @@
...
+                                       "opts" {"annotation" {"originator" "txn" "bar" "foo"}}})
+                   (fluree/commit! ledger)
+                   (deref))]
+      (is (= [{"f:txn" "fluree:memory://annotationtest/txn/5c54eee1fd6be197402fcd76b035cd6f42b55723f35c294992d76e6eea2cf874"}
+              {"f:txn" "fluree:memory://annotationtest/txn/6ab2e58c44af9db30feb461c212b3b566a35fdfe92d53b61018eab425ca0c342"
+               "f:annotation" {"originator" "opts" "foo" "bar"}}
+
+              {"f:txn" "fluree:memory://annotationtest/txn/e342e6a53d26b78b6222446eac29060c5237f54175e1d6639cb900103b729ccc"
+               "f:annotation" {"originator" "txn" "bar" "foo"}}]
+             (->> @(fluree/history ledger {:context        context
```
zonotope, line 1002: could you break this query and formatting apart and bind them to descriptive names in the `let` binding vector?

src/fluree/db/json_ld/transact.cljc
```diff
@@ -183,7 +184,7 @@
           {:keys [last-pid last-sid]} @iri->sid
 
           db-after (-> db-before
-                       (update :txns (fnil conj []) [txn-id author-did])
+                       (update :txns (fnil conj []) [txn-id author-did annotation])
```
zonotope, line 187: How is this `:txns` entry used?

bplatz:
I think we need a different approach to this similar to v1/v2, but perhaps this is needed short-term and we should do that separately.

Transactional metadata should be "just more triples", where the subject is the transaction's @id value. That approach also allows it to all be queried using the normal flureeQL sysntax.

It appears here we are storing maps. They are not indexable, or queryable - and also lack the other benefits of RDF data.

JSON-LD already has a strong notion of this, but given that our "transactions" can be combinations of where/insert/delete statements we'd need to adopt... but this example using a named graph right out the spec illustrates this:

```
{
  "@context": {
    "generatedAt": {
      "@id": "http://www.w3.org/ns/prov#generatedAtTime",
      "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
    },
    "Person": "http://xmlns.com/foaf/0.1/Person",
    "name": "http://xmlns.com/foaf/0.1/name",
    "knows": {"@id": "http://xmlns.com/foaf/0.1/knows", "@type": "@id"}
  },
  "@id": "http://example.org/foaf-graph",
  "generatedAt": "2012-04-09T00:00:00",
  "@graph": [
    {
      "@id": "http://manu.sporny.org/about#manu",
      "@type": "Person",
      "name": "Manu Sporny",
      "knows": "https://greggkellogg.net/foaf#me"
    }, {
      "@id": "https://greggkellogg.net/foaf#me",
      "@type": "Person",
      "name": "Gregg Kellogg",
      "knows": "http://manu.sporny.org/about#manu"
    }
  ]
}
```
Here, the `generatedAt` property is transactional metadata.

I have some ideas of how we should handle this syntactically... but don't want to be perspecriptive but happy to share.

Ultimately this will provide for the functionality that we want to tell customers they can. us... and I believe the approach done here would not meet those requirements (correct me if I'm wrong).

src/fluree/db/json_ld/transact.cljc
dpetran: This is a vector of per-stage txn metadata that gets accumulated at every stage. During the commit it gets processed into commit data and removed from the db.

PR #720, opened 2024-02-05
fix flake meta json serialization

dpetran:
Fixes https://github.com/fluree/core/issues/91

Serializing a flake with meta would incorrectly override the `op` value of the flake. This fixes that.

old, incorrect:
(serialize-flake #Flake [211106232533006 1008 "two" 1 -4 true {:i 2}]) [211106232533006 1008 "two" 1 -4 {"i" 2} {:i 2}]

new, correct:
(serialize-flake #Flake [211106232533006 1008 "two" 1 -4 true {:i 2}]) [211106232533006 1008 "two" 1 -4 true {"i" 2}]

This was causing list values to be materialized on load in an arbitrary order, which was sometimes correct, sometimes not.

### DIFF
--- a/src/fluree/db/serde/json.cljc
+++ b/src/fluree/db/serde/json.cljc
@@ -100,7 +100,7 @@
   [flake]
   (-> (vec flake)
       (update 2 serialize-flake-value (flake/dt flake))
-      (cond-> (flake/m flake) (assoc 5 (util/stringify-keys (flake/m flake))))))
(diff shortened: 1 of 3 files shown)


dpetran:
This may fix https://github.com/fluree/core/issues/75 as well.

dpetran:
This may fix https://github.com/fluree/core/issues/74 as well

zonotope (approved):
🎹

PR #721, opened 2024-02-09
Feature/ipfs store

dpetran:
This creates an ipfs store to underlie the ipfs connection. It is not used by the ipns nameservice - that has its own config and is managed separately.

### DIFF
--- a/src/fluree/db/storage/proto.cljc
+++ b/src/fluree/db/storage/proto.cljc
@@ -2,6 +2,7 @@
   (:refer-clojure :exclude [read list exists?]))
 
 (defprotocol Store
+  (address [store k] "Returns the address that would be constructed by writing to `k`.")
   (write [store k v opts] "Writes `v` to Store associated with `k`. Returns value's address.")
   (exists? [store address] "Returns true when address exists in Store.")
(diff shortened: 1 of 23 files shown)


src/fluree/db/storage/proto.cljc
```diff
@@ -2,6 +2,7 @@
   (:refer-clojure :exclude [read list exists?]))
 
 (defprotocol Store
+  (address [store k] "Returns the address that would be constructed by writing to `k`.")
```
zonotope, line 5: What is this method used for?

dpetran:
I've backed out the usage of the store protocol in the nameservice implementations. It collides with what Brian has been working on. All that remains are the bare minimum changes necessary for it to work.

src/fluree/db/storage/proto.cljc
zonotope: Ok. I haven't looked in to the nameservice implementation recently, but It doesn't seem right to me at first blush that something would need to know from the storage implementation what the address will be as a result of a given write before the fact, especially since it's very hard (last i checked impossible) to know this for IPFS without actually writing the file.

zonotope (approved):
🥮

PR #722, opened 2024-02-09
convert s3 conn to use s3 store

dpetran:
This converts the s3 connection to use the s3 store abstraction.


src/fluree/db/conn/s3.clj
```diff
@@ -17,14 +17,14 @@
             [fluree.db.util.core :as util]
```
cap10morgan, line 17: Not for this PR, but this ns seems barely S3-specific now. Is there an opportunity for more code consolidation emerging here?

cap10morgan (approved):
🍶

PR #723, opened 2024-02-13
Fix querying nonexistent ledgers

JaceRockman:


JaceRockman:
Not sure how to document this, but this should solve Issue 83 https://github.com/fluree/core/issues/83

PR #726, opened 2024-02-20
Update file nameservice format, use with IPFS conn

bplatz:
IPFS connection types now by default use two nameservices - IPNS and filesystem. The filesystem nameservice is synchronous, and requires a successful write before a commit is confirmed. IPNS is asynchronous and will update as IPNS allows it to.

Note that any connection can have any number of nameservices configured and fed into the conn as an option - so the old behavior is still available if needed, just not the default.

At the same time, the file nameservice needed to be updated as it previously only wrote file system path names as the commit-address - but now it needs to be able to write IPFS commit addresses (and potentially others).

This PR not only makes this address update to the file nameservice, but also
- stores the nameservice file as JSON-LD which can contain additional metadata about the ledger and branches
- Starts to use the branch specification format, as previously discussed, as `(<branch-name>)` appended to the nameservice address. Because branching isn't fully built out, this is all hidden under the hood for now

Future steps:
- Cache the nameservice file so a read isn't required each time (likely with a TTL, and coupled with a notification system that invalidates cache on known updates)
- Implement a Fluree database nameservice - using the JSON-ld format, the nameservice files could easily be stored in a special nameservice fluree database
- Allow `create` API which doesn't create a DB (it wouldn't require a transaction), but reserves the namespace so a db can be placed there
- Have a `status` property on the JSON-LD, where actions such as reindexing could be noted, which would make a ledger temporarily unavailable
-


src/fluree/db/nameservice/proto.cljc
```diff
@@ -8,13 +8,11 @@
...
   (-sync? [nameservice] "Indicates if nameservice updates should be performed synchronously, before commit is finalized. Failure will cause commit to fail")
-
-  (-exists? [nameservice ledger-address] "Returns true if ledger exists (must have had at least one commit), false otherwise")
   (-ledgers [nameservice opts] "Returns a list of ledger aliases registered with this nameservice")
 
   (-close [nameservice] "Closes all resources for this nameservice")
 
   (-alias [nameservice ledger-address] "Given a ledger address, returns ledger's default alias name else nil, if not avail")
-  (-address [nameservice ledger-alias key] "Returns address/iri for provided ledger alias specific to the connection type"))
+  (-address [nameservice ledger-alias key] "Returns full nameservice address/iri which will get published in commit. If 'private', return nil."))
```
dpetran, line 16: What is `private`? I'm not familiar with that semantic.

src/fluree/db/nameservice/filesystem.cljc
```diff
@@ -1,85 +1,225 @@
...
-
-        work        (fn [complete]
-                      (log/debug (str "Updating head at " write-path " to " commit-path "."))
-                      (fs/write-file write-path (bytes/string->UTF8 commit-path))
-                      (complete (file-address head-path)))]
-    #?(:clj  (let [p (promise)]
-               (future (work (partial deliver p)))
-               p)
-       :cljs (js/Promise. (fn [resolve reject] (work resolve))))))
+  "The file nameservice will eventually hold metadata including various
```
dpetran, line 95: I generally try to keep notes for future work as comments instead of docstrings, it makes it easier to tell what the function _actually_ does as opposed to what it will do.

src/fluree/db/nameservice/filesystem.cljc
```diff
@@ -1,85 +1,225 @@
...
   (let [[_ _ path] (str/split address #":")]
     path))
 
-(defn address-full-path
-  [local-path address]
-  (str local-path "/" (address-path address)))
-
-(defn file-address
-  "Turn a path or a protocol-relative URL into a fluree file address."
+(defn list-files-in-path
```
dpetran, line 39: This seems like it could live alongside the other file system utils in `fluree.db.util.filesystem`.

src/fluree/db/nameservice/filesystem.cljc
```diff
@@ -1,85 +1,225 @@
...
             [fluree.db.util.async :refer [<? go-try]]
             [clojure.string :as str]
-            [fluree.db.util.log :as log]))
+            [fluree.db.util.json :as json]
+            [fluree.db.util.log :as log]
+            #?(:clj [clojure.java.io :as io])))
 
 #?(:clj (set! *warn-on-reflection* true))
 
+(defn ns-record
```
dpetran, line 15: If I understand correctly, this is the new HEAD document.

zonotope (approved):
🥬

src/fluree/db/nameservice/proto.cljc
bplatz: private means the nameservice address is not published in the commit file - but the local server can still know how to find it.

src/fluree/db/nameservice/filesystem.cljc
bplatz: Fixed in forthcoming PR to apply this changes to latest main

bplatz:
Plan to close this PR without merging, instead favoring PR #758 - leaving open for now as we have at least one user currently relying on this branch.

cap10morgan:
@bplatz Good to close this now?

PR #732, opened 2024-02-23
Feature/blank node ids

zonotope:
I woke up in a cold sweat haunted by the possibility of collisions in our blank node ids so I decided to make those far less likely. 

This patch integrates [nano-id](https://github.com/zelark/nano-id) to generate blank node ids instead of using `rand-int` like we did previously. It retains the timestamp prefix so we still won't have too much index thrashing. Using a nano-id length of 8, we'd need to [generate 2 million ids in one millisecond to get a collision probability of 1%](https://zelark.github.io/nano-id-cc/). 

```clojure
user> (new-blank-node-id)
"_:fdb-1708699899809-sfqiEdMv"
```


dpetran (approved):
:safety_vest:

PR #733, opened 2024-02-26
Cleanup/storage

zonotope:
This patch implements the following code organization changes with the storage protocol and implementations that I think will make them more flexible and generalizable going forward. I would like to eventually make similar changes to the connection and database protocols, but those will come later. 

These changes:
- Consolidated the `fluree.db.storage.proto` and `fluree.db.storage.util` namespaces into `fluree.db.storage`.
- Simplified the storage record constructor functions to take normal arguments instead of configuration maps to make those functions more generally useful and less tied to user input/configuration formats. 
- Remove the configuration validation for storage records. These were no longer necessary because of the simplified constructor functions. We will need to eventually need to add some sort of configuration validation for user input while constructing a *connection*, but external users will not supply any input to construct storage directly because it's an internal api. We will soon overhaul connections and we'll add back validation schemas for connection constructors as a whole during that work. 
- The current connections are tied closely to the storage back ends, so we rely on the connection constructors to initialize the storage they need. When we overhaul connections to make them more general, we'll add a mechanism to allow a user to specify their storage back end through the configuration the pass to the `fluree/connect` function. 
- We had previously defined `-txn-read` and `-txn-write` methods on the `fluree.db.conn.proto/iStorage` protocol, but it just wasn't implemented for any of the connection types. So instead of breaking the connection abstraction and reaching in to the store directly to write out transactions, I implemented those methods for all the existing connection types and used that method during the transact process. We will soon overhaul the connection protocols (possibly even removing the protocol all together), but there will still be a "connection" abstraction that we should not break.
- Defined the localstorage store in a cljs file because it will only be used from clojurescript
- Besides that, there are some small error handling fixes, bug fixes, and cosmetic changes


dpetran (approved):
:+1:

PR #738, opened 2024-02-29
Correctly return nil when read-address fails

JaceRockman:


bplatz:
Hi @JaceRockman I suggest we close this as per above... let me know if you disagree, else please feel free to close.

PR #739, opened 2024-02-29
remove constant metadata from min/max sid and flake components

zonotope:
While this metadata didn't cause any issue running the app in the tests or at the repl, it did cause issues running dependent apps through an uberjar. This patch removes this metadata for now until we can figure out how, or if, to add it back.

### DIFF
--- a/src/clj/fluree/db/flake.cljc
+++ b/src/clj/fluree/db/flake.cljc
@@ -23,12 +23,12 @@
 (def max-p iri/max-sid)
 (def min-dt iri/min-sid)
 (def max-dt iri/max-sid)
-(def min-t 0)
-(def max-t util/max-long)
-(def min-op false)
-(def max-op true)
-(def min-meta util/min-integer)
-(def max-meta util/max-integer)
+(def ^:const min-t 0)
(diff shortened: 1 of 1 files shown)


src/clj/fluree/db/flake.cljc
```diff
@@ -17,18 +17,18 @@
...
+(def min-p iri/min-sid)
+(def max-p iri/max-sid)
+(def min-dt iri/min-sid)
+(def max-dt iri/max-sid)
+(def min-t 0)
+(def max-t util/max-long)
+(def min-op false)
+(def max-op true)
+(def min-meta util/min-integer)
+(def max-meta util/max-integer)
```
cap10morgan, line 31: Do we want to remove all of them, or just the ones that interact with the `SID` class?

cap10morgan (approved):
One question but LGTM either way

src/clj/fluree/db/flake.cljc
zonotope: good point. all add the non-sid ones back

cap10morgan:
Looks even better now :)

PR #743, opened 2024-03-13
Index after every commit in migration

zonotope:
This patch changes the migration script to index after every commit instead of all at once at the end. That way, the indexer will decide if the index should actually need to be run, and it's less likely that we'll run into memory issues migrating larger ledgers. 

It also creates a unique file changes channel for each commit, and merges those channels into the changes channel that was passed in to the migrator. This is so we can run the indexer after each commit because it closes the changes channel after it runs, but we'd like to keep the original changes channel open in this case.

### DIFF
--- a/src/clj/fluree/db/json_ld/migrate/sid.cljc
+++ b/src/clj/fluree/db/json_ld/migrate/sid.cljc
@@ -119,9 +119,9 @@
                                 :main)
           ledger            (<? (jld-ledger/->ledger conn ledger-alias {:branch branch}))
           commit-opts*      (assoc commit-opts :branch branch)
-          tuples-chans      (->> all-commit-tuples
-                                 (map (fn [commit-tuple]
-                                        [commit-tuple (async/chan)])))
+          tuples-chans      (map (fn [commit-tuple]
+                                   [commit-tuple (async/chan)])
+                                 all-commit-tuples)
           changes-chs       (map second tuples-chans)
           _                 (-> (async/merge changes-chs)
                                 (async/pipe changes-ch))
--- a/src/clj/fluree/db/json_ld/migrate/sid.cljc
+++ b/src/clj/fluree/db/json_ld/migrate/sid.cljc
@@ -42,7 +42,7 @@
     (let [db-address (-> commit
                          (get-first const/iri-data)
                          (get-first-value const/iri-address))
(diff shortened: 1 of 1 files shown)


src/clj/fluree/db/json_ld/migrate/sid.cljc
```diff
@@ -115,11 +118,13 @@
...
-          update-commit     (commit/update-commit-fn db commit-opts*)
-          indexed-db        (<? (indexer/do-index indexer db {:changes-ch    changes-ch
-                                                              :update-commit update-commit}))]
-      (ledger/-db-update ledger indexed-db)
+          tuples-chans      (->> all-commit-tuples
+                                 (map (fn [commit-tuple]
+                                        [commit-tuple (async/chan)])))
+          changes-chs       (map second tuples-chans)
+          _                 (-> (async/merge changes-chs)
+                                (async/pipe changes-ch))
```
cap10morgan, line 127: Might be better as:

```clojure
(-> changes-chs
    async/merge
    (async/pipe changes-ch))
```

Or:

```clojure
(let [merged (async/merge changes-chs)
      _      (async/pipe merged changes-ch)
```

Or even just:

```clojure
(async/pipe (async/merge changes-chs) changes-ch)
```

cap10morgan (approved):
Few small formatting things and a question, but looks good once addressed

src/clj/fluree/db/json_ld/migrate/sid.cljc
```diff
@@ -95,17 +94,21 @@
...
+              update-commit (commit/update-commit-fn merged-db commit-opts)
+              indexed-db    (<? (indexer/do-index indexer merged-db
+                                                  {:changes-ch    ch
+                                                   :update-commit update-commit}))]
+          (recur r indexed-db))
         db))))
 
 (defn migrate
-  [conn address changes-ch commit-opts]
+  [conn address commit-opts changes-ch]
```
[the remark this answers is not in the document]
zonotope: Yeah, but the only caller is in the server repo in a branch that also hasn't been merged yet, but I've updated it there.

src/clj/fluree/db/json_ld/migrate/sid.cljc
zonotope: Went with your first option

Issue #351, opened 2023-01-24
loaded db and live db stats do not match


dpetran:

cap10morgan:
branch with failing tests for this (they're in `f.d.json-ld.api-test`): https://github.com/fluree/db/tree/fix/inconsistent-db-stats

dpetran:
`asserts` - `retracts` measurement tells us the size of the flakes an `[?s ?p ?o]` query would return. 

`asserts` + `retracts` tells us the rough size the whole db would take on disk - not exactly, since there's the index node overhead, but gives you a rough idea of how much space you would need.

I'm not sure which measure is more useful.

cap10morgan:
Or maybe they're both useful and we just need to come up with good names for the two different keys in `:stats`?

dpetran:
@bplatz do you have a preference here? I think all we need is a decision of one or the other or both, and I think you have more customer context to be able to make that call.

bplatz:
@dpetran I've always considered the total size of the DB to be important - both 't' values and size, as that would give someone a sense of how many commits and how large they are. Also if the size was a proxy for someone using it for billing, total size would be more important here as well.

dpetran:
Alright, then we will just do the `asserts` + `retracts` implementation to fix this bug.

PR #745, opened 2024-03-25
Do not subtract retracts from db :size

JaceRockman:
This is a fix for https://github.com/fluree/db/issues/351

### DIFF
--- a/src/clj/fluree/db/json_ld/commit_data.cljc
+++ b/src/clj/fluree/db/json_ld/commit_data.cljc
@@ -375,7 +375,7 @@
          (update-in [:novelty :opst] flake/revise ref-add ref-rem)
          (update-in [:novelty :tspo] flake/revise add rem)
          (update-in [:novelty :size] + flake-size)
-         (update-in [:stats :size] + flake-size)
+         (update-in [:stats :size] + (flake/size-bytes add))
          (update-in [:stats :flakes] + flake-count)))))
 
 (defn add-tt-id
--- a/src/clj/fluree/db/json_ld/transact.cljc
+++ b/src/clj/fluree/db/json_ld/transact.cljc
@@ -185,7 +185,7 @@
                       (update :txns (fnil conj []) [txn-id author-did])
                       (assoc :policy policy) ;; re-apply policy to db-after
                       (assoc :t t)
-                      (commit-data/update-novelty add remove)
+                      (commit-data/update-novelty add)
                       (commit-data/add-tt-id)
                       (vocab/hydrate-schema add))]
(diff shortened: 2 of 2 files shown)


zonotope:
@JaceRockman the tests fail on this branch. Do you still want to keep this pull request open?

JaceRockman:
> Hi @JaceRockman I think this might not be `false` `.-op` (retraction) flakes, but flakes that are completely removed during staging.
> 
> Therefore I think the counts are accurate.
> 
> For context, note this is the main place this is called when transacting:
> 
> https://github.com/fluree/db/blob/2a9db1b669784b3897c99b8ace5980c7acc6b869/src/clj/fluree/db/json_ld/transact.cljc#L188
> 
> And for the function as a whole, on line 181:
> 
> https://github.com/fluree/db/blob/2a9db1b669784b3897c99b8ace5980c7acc6b869/src/clj/fluree/db/json_ld/transact.cljc#L177-L191
> 
> That `remove` is only ever used for stage operations (in-between commits). This means that `add` actually has both addition and retraction flakes in it (probably could have used a better name).
> 
> The main thing is `remove` is different than `retract`, and therefore the counts looks like they were accurate.

@bplatz So is issue https://github.com/fluree/db/issues/351 not actually a problem then? Or do you think it was fixed by some other change?

bplatz:
@JaceRockman I'm not sure if that is still a problem... but if so I don't think this is the source.

The ticket says:

> fluree.db.json-ld.transact/final-db calculates the size of the asserts and subtracts the size of the retracts

But I *think* this is incorrect... it subtracts the size of `remove`, not `retract` (what I think was meant by "subtracts" above)... and `remove` is only ever used with multiple `fluree/stage` calls in-between commits. I think our naming here is confusing, because in `final-db` what it calls `add` is actually both assertions + retractions.

If this is still a problem, my best guess is that when we reify, we include commit metadata flakes when calculating size, but when transacting we don't include the commit metadata flakes. But that is just a guess.

JaceRockman:
@cap10morgan @bplatz I ran the tests in this branch https://github.com/fluree/db/tree/fix/inconsistent-db-stats on the most recent version of main and they passed with flying colors, so I think the original issue got fixed over the course of some other changes. I wanted to check with you before closing just to be sure though.

cap10morgan:
> @cap10morgan @bplatz I ran the tests in this branch https://github.com/fluree/db/tree/fix/inconsistent-db-stats on the most recent version of main and they passed with flying colors, so I think the original issue got fixed over the course of some other changes. I wanted to check with you before closing just to be sure though.

I would say it's safe to close then. If we discover it should be re-opened later we can do that.

PR #746, opened 2024-03-28
Generalize Filter Clauses

zonotope:
This patch generalizes filter clauses by removing the constraint that they only reference a single variable defined in the where clause and removes the optimization of applying prospective filters defined in filter clause only in index range scans directly. Removing that optimizations allows us to apply filters to variables not bound by range scans, like from a bind clause. It also allows us to filter for the absences of a variable, like from an optional clause.

This patch also adds a new "variable map" syntax where single-variable filters can be defined where the index range scan optimization will be applied. This allows users to still have an option of directing the query engine to apply the index range scan optimization where possible, while still being able to define more general filters referencing multiple variables or variables not bound in an index range scan.

### DIFF
--- a/src/clj/fluree/db/constants.cljc
+++ b/src/clj/fluree/db/constants.cljc
@@ -61,6 +61,7 @@
 (def ^:const iri-value "@value")
 (def ^:const iri-language "@language")
 (def ^:const iri-type "@type")
+(def ^:const iri-filter "@filter")
 (def ^:const iri-json "http://www.w3.org/2001/XMLSchema#json")
 (def ^:const iri-anyURI "http://www.w3.org/2001/XMLSchema#anyURI")
 (def ^:const iri-rdf-type "http://www.w3.org/1999/02/22-rdf-syntax-ns#type")
--- a/src/clj/fluree/db/query/exec/where.cljc
+++ b/src/clj/fluree/db/query/exec/where.cljc
@@ -1,6 +1,6 @@
 (ns fluree.db.query.exec.where
   (:require [fluree.db.query.range :as query-range]
-            [clojure.core.async :as async :refer [go take! put!]]
+            [clojure.core.async :as async :refer [go]]
             [fluree.db.flake :as flake]
             [fluree.db.fuel :as fuel]
             [fluree.db.index :as index]
@@ -68,13 +68,6 @@
           (matched-sid? mch))
     const/iri-anyURI
     (::datatype-iri mch)))
-
-(defn get-datatype-sid
-  [mch db-alias]
-  (if (or (matched-iri? mch)
-          (matched-sid? mch))
-    const/$xsd:anyURI
-    (get-in mch [::datatype-sids db-alias])))
 
 (defn matched?
   [match]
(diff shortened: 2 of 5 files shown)


dpetran (approved):
:+1: 

I think this is a good change, here's my understanding of the changes:

1) the `[:filter ...]` pattern 
- can use any number of variables, bound or unbound (this lack of restriction is what's new)
- this does not get applied during the range scan (less efficient but more flexible)
2) a `:filter` value in an object position in a `where` pattern 
- applied during a range scan
- can only bind the one object
- has to be specified in a value map, with the `:value` keyword being the binding

This seems like a pretty good way to hold on to the efficiency of scan filters while still providing the flexibility of a `:filter` pattern.

bplatz:
@zonotope I'm still having issues making this work for my primary use case of this feature. Here is a reproducible case that should be able to be pasted directly into your REPL:

```
(let [conn   @(fluree/connect-memory nil)
      ledger @(fluree/create conn "test/db1")
      db     @(fluree/stage
                (fluree/db ledger)
                {"@context" {"ex" "http://example.org/"}
                 "insert"   {"@id"       "ex:bob"
                             "ex:father" [{"@id" "ex:alex-jr"}, {"@id" "ex:aj"}]}})]
  @(fluree/query db {:context {"ex" "http://example.org/"}
                     :select  '[?s ?f1 ?f2]
                     :where   '[{"@id"       ?s
                                 "ex:father" ?f1}
                                {"@id"       ?s
                                 "ex:father" ?f2}
                                ;; remove below filter to see all results (should be 4 total)
                                ["filter" "(not= ?f1 ?f2)"]]})
  ;; => []   ;; but should be: [["ex:bob" "ex:alex-jr" "ex:aj"], ["ex:bob" "ex:aj" "ex:alex-jr"]]
  )
```

zonotope:
> @zonotope I'm still having issues making this work for my primary use case of this feature. Here is a reproducible case that should be able to be pasted directly into your REPL:
> 
> ```
> (let [conn   @(fluree/connect-memory nil)
>       ledger @(fluree/create conn "test/db1")
>       db     @(fluree/stage
>                 (fluree/db ledger)
>                 {"@context" {"ex" "http://example.org/"}
>                  "insert"   {"@id"       "ex:bob"
>                              "ex:father" [{"@id" "ex:alex-jr"}, {"@id" "ex:aj"}]}})]
>   @(fluree/query db {:context {"ex" "http://example.org/"}
>                      :select  '[?s ?f1 ?f2]
>                      :where   '[{"@id"       ?s
>                                  "ex:father" ?f1}
>                                 {"@id"       ?s
>                                  "ex:father" ?f2}
>                                 ;; remove below filter to see all results (should be 4 total)
>                                 ["filter" "(not= ?f1 ?f2)"]]})
>   ;; => []   ;; but should be: [["ex:bob" "ex:alex-jr" "ex:aj"], ["ex:bob" "ex:aj" "ex:alex-jr"]]
>   )
> ```

@bplatz I just pushed a fix for this along with a test based on the code you provided. that code made tracking this down pretty easy, so thanks for that. 

The problem was from the way we represent iris internally, which is different from scalar values to allow for federated queries to work across different ledgers where the sids might not be the same. 

The filtering code was only checking for the case of scalar values and not iri references, so it always got `nil` when trying to extract an iri value, so `not=` always returned false.

PR #747, opened 2024-04-01
Fix :schema map generation

bplatz:
The :schema map generation changed, and no longer worked correctly for rdfs:subClassOf processing.

This adds a test that would fail prior to this patch, but succeeds with it.

The issue originated from any db change generating a new :schema map, then later merging it with the prior one. The issue is with at least rdfs:subClassOf, there can be dependencies on the prior schema map.

This now carries the current schema map through instead of re-generating it with every update and later merging it. The merge is no longer necessary as the new map is based on the old.


zonotope (approved):
🎒

PR #748, opened 2024-04-01
Remove :ref?, :refs from :schema map

bplatz:
In the :schema map of the db, :ref? and :refs keys are no longer used - this removes them


zonotope (approved):
✂️

PR #749, opened 2024-04-03
Add decode-iri API, rename internal-id->encode-iri

bplatz:
Added a `decode-iri` API to correspond with the old `internal-id` API, which got renamed to `encode-iri`.

Had a use case that needed IRI decoding, which prompted this PR.

`internal-id` was the old API when we had internal integer subject IDs. Now that we always have IRIs with a special IRI encoding, `encode-iri` I think is a better description.

We didn't have the opposite API, so this is now in place and called `decode-iri`.

### DIFF
--- a/src/clj/fluree/db/json_ld/api.cljc
+++ b/src/clj/fluree/db/json_ld/api.cljc
@@ -328,9 +328,25 @@
    (json-ld/expand-iri compact-iri
                        (json-ld/parse-context context))))
 
-(defn internal-id
-  "Returns the internal Fluree integer id for a given IRI.
+(defn encode-iri
+  "Returns the internal Fluree IRI identifier (a compact form).
   This can be used for doing range scans, slices and for other
   more advanced needs."
   [db iri]
   (iri/encode-iri db iri))
+
+(defn internal-id
+  "Deprecated, use encode-iri instead."
+  {:deprecated true}
+  [db iri]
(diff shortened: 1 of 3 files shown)


zonotope (approved):
I didn't add this function before because I thought the iri would be fine in all use cases outside of the db library since they now have a direct mapping that's cheap to compute, but, since there's still some nondeterminism in the conversion because of the namespace mapping, if there's a use case for sids outside of the library, then it makes sense to have this function.

bplatz:
@zonotope Thinking about this... that you are correct but also want to consider end-users (or tools) doing multiple index-range scans.. which means we'll be decoding/encoding at every step. But if the work is trivial it would be great to have this 100% hidden.

Anyhow, mulling this over before merging.

zonotope:
> @zonotope Thinking about this... that you are correct but also want to consider end-users (or tools) doing multiple index-range scans.. which means we'll be decoding/encoding at every step. But if the work is trivial it would be great to have this 100% hidden.
> 
> Anyhow, mulling this over before merging.

These conversions should be _very_ fast. It first does a hash map lookup in a relatively small map, so that part should be very fast. Then there are a few bit-shift and bitwise logical operations, and those tend to be the fastest thing a digital computer can do. There are also a few string allocations in there too, but I don't think that should matter in the context of an index range scan.

PR #752, opened 2024-04-08
Persist index updates

zonotope:
This patch started as a fix for overly ridged commit validation that rejected any index updates against commits from the past. In high volume ledgers, the indexer might not always keep up with commits so indexes might come back for commits > 2 t-values in the past. Rejecting these index updates had the effect of stopping the indexer until there's a pause in transactions, which had an impact on both query and transaction performance during periods of high load. This patch fixes that issue by adding a [function that considers both the index and t values separately and takes the latest of both](https://github.com/fluree/db/blob/936701f4282f824c9b0d93cf7022de8f06a88790/src/clj/fluree/db/json_ld/branch.cljc#L59-L80) during each commit/index update.

In fixing this bug, I discovered another issue where the newly completed indexes were always [thrown away and replaced with `nil`](https://github.com/fluree/db/blob/main/src/clj/fluree/db/json_ld/branch.cljc#L64) during each index index update because we stored commit and index state both in the branch and the branches latest-db record without ensuring that those two items stayed in sync. I have also fixed this issue, and I began to change our branch state to only store the current db without repeating any other state in branch metadata that needed to be kept in sync, but I ran into some circular dependency issues between connections, ledgers, branches, dbs, and indexers that started looking like a larger change, so I'm saving the rest of that for a later pr.


src/clj/fluree/db/json_ld/branch.cljc
```diff
@@ -33,110 +33,92 @@
...
+    (if (= db-t new-t)
+      (if (updatable-commit? current-commit new-commit)
+        (update-db branch-map db)
+        (throw (ex-info (str "Commit failed, latest committed db is " current-t
+                             " and you are trying to commit at db at t value of: "
+                             new-t ". These should be one apart. Likely db was "
+                             "updated by another user or process.")
+                        {:status 400 :error :db/invalid-commit})))
+      (throw (ex-info (str "Unexpected Error updating commit database. "
+                           "New database has an inconsistent t from it's commit:"
```
cap10morgan, line 117: `s/it's/its/`

src/clj/fluree/db/json_ld/commit.cljc
```diff
@@ -31,31 +28,30 @@
...
-                        (datatype/time-type? pdt)
-                        ;;need to retain the `@type` for times
-                        ;;so they will be coerced correctly when loading
-                        (assoc "@type"
-                               (get-s-iri pdt db compact-fn)))
-            next-acc' (conj acc' obj*)]
-        (if (seq r')
-          (recur r' all-refs? next-acc')
-          [next-acc' all-refs?])))))
+  (loop [[p-flake & r'] p-flakes
```
cap10morgan, line 31: Should `r'` just be `r` here?

src/clj/fluree/db/json_ld/commit.cljc
```diff
@@ -31,31 +28,30 @@
...
+         all-refs? nil
+         acc'      nil]
```
cap10morgan, line 33: Also no need for a prime here, I think?

src/clj/fluree/db/json_ld/commit.cljc
```diff
@@ -31,31 +28,30 @@
...
+                            [{"@value" (-> p-flake
+                                           flake/o
+                                           (serde-json/serialize-object pdt))}
+                             false])
+          obj*      (cond-> obj
+                      list? (assoc :i (-> p-flake flake/m :i))
+                      (datatype/time-type? pdt)
+                      ;;need to retain the `@type` for times
+                      ;;so they will be coerced correctly when loading
+                      (assoc "@type"
```
cap10morgan, line 49: This is super nitpick-y and totally up to you, but it took me a sec to understand what's going on here w/ the comment splitting the pred and the threaded form. Might be ever-so-slightly more readable to put them comment above the pred? Could even put a blank line above them for even more obvious grouping.

src/clj/fluree/db/json_ld/commit.cljc
```diff
@@ -31,31 +28,30 @@
...
+                             (get-s-iri pdt db compact-fn)))
+          next-acc' (conj acc' obj*)]
```
cap10morgan, line 51: Maybe either rename `next-acc'` to just `acc'` or `next-acc`?

src/clj/fluree/db/json_ld/commit.cljc
```diff
@@ -217,61 +166,91 @@
...
+          (let [sid   (flake/s (first s-flakes))
+                s-iri (get-s-iri sid db compact-fn)
+                [assert* retract*]
+                (if (and (= 1 (count s-flakes))
+                         (= const/$rdfs:Class (->> s-flakes first flake/o))
+                         (= const/$rdf:type (->> s-flakes first flake/p)))
+                  ;; we don't output auto-generated rdfs:Class definitions for classes
+                  ;; (they are implied when used in rdf:type statements)
+                  [assert retract]
+                  (let [{assert-flakes  true,
```
cap10morgan, line 196: No need for comma w/ line break

src/clj/fluree/db/json_ld/commit.cljc
```diff
@@ -217,61 +166,91 @@
...
     (or (nil? ledger-t)
-        (> (commit-data/t db-commit)
-           ledger-t))))
+        (flake/t-after? (commit-data/t db-commit)
+                        ledger-t))))
 
 (defn do-commit+push
   "Writes commit and pushes, kicks off indexing if necessary."
-  [{:keys [ledger commit] :as db} {:keys [branch did private] :as _opts}]
+  [ledger {:keys [commit] :as db} {:keys [branch did private] :as _opts}]
```
cap10morgan, line 253: 👍🏻

src/clj/fluree/db/json_ld/migrate/sid.cljc
```diff
@@ -127,5 +126,5 @@
                                 async/merge
                                 (async/pipe changes-ch))
           db                (<? (merge-commits ledger commit-opts* tuples-chans))]
-      (ledger/-db-update ledger db)
+      (jld-ledger/db-update ledger db)
```
cap10morgan, line 129: Was the idea here to support other implementations of the ledger protocol? Any reason to try to preserve that abstraction?

src/clj/fluree/db/json_ld/transact.cljc
```diff
@@ -80,25 +78,27 @@
...
-      :t             t
-      :reasoner-max  10 ;; maximum number of reasoner iterations before exception
-      :reasoned      reasoned-from-IRI})))
+  "Generates a state map for transaction processing. When optional
+  reasoned-from-IRI is provided, will mark any new flakes as reasoned from the
+  provided value in the flake's metadata (.-m) as :reasoned key."
+  ([db txn author-did]
+   (->tx-state db txn author-did nil))
+
+  ([db txn author-did reasoned-from-IRI]
```
cap10morgan, line 87: `reasoned-from-iri`?

src/clj/fluree/db/json_ld/transact.cljc
```diff
@@ -184,12 +184,12 @@
...
   "Returns map of all elements for a stage transaction required to create an
   updated db."
-  [db new-flakes {:keys [stage-update? policy t txn-id author-did] :as _tx-state}]
+  [db new-flakes {:keys [stage-update? policy t txn author-did] :as _tx-state}]
   (let [[add remove] (if stage-update?
                        (stage-update-novelty (get-in db [:novelty :spot]) new-flakes)
                        [new-flakes nil])
         db-after  (-> db
-                      (update :txns (fnil conj []) [txn-id author-did])
+                      (update :staged conj [txn author-did])
```
cap10morgan, line 192: Are we good with generating lists (instead of vectors) on nil here?

src/clj/fluree/db/ledger/json_ld.cljc
```diff
@@ -37,15 +38,16 @@
...
 (defn db-update
   "Updates db, will throw if not next 't' from current db.
   Returns original db, or if index has since been updated then
   updated db with new index point."
   [{:keys [state] :as _ledger} {:keys [branch] :as db}]
-  (-> (swap! state update-in [:branches branch] branch/update-db db)
-      (get-in [:branches branch :latest-db])))
+  (-> state
+      (swap! update-in [:branches branch] branch/update-db db)
+      (get-in [:branches branch :current-db])))
```
cap10morgan, line 50: 👍🏻

src/clj/fluree/db/ledger/json_ld.cljc
```diff
@@ -54,13 +56,14 @@
   If index in provided db is newer, updates latest index held in ledger state."
   [{:keys [state] :as ledger} branch-name db]
   (log/debug "Attempting to update ledger's db with new commit:"
-             (:alias ledger) "branch:" branch-name)
+             (:alias ledger) "branch:" branch-name "to t" (:t db))
```
cap10morgan, line 59: Is it confusing that the alias comes after `commit: ` in the log message?

src/clj/fluree/db/ledger/json_ld.cljc
```diff
@@ -247,22 +235,15 @@
...
+  (let [db (jld-db/create ledger)]
+    (db-update ledger db)))
 
 (defn create*
-  [conn ledger-alias {:keys [include] :as opts}]
+  [conn ledger-alias opts]
   (go-try
     (let [ledger (<? (->ledger conn ledger-alias opts))]
-      (<? (initialize-db! ledger include))
+      (initialize-db! ledger)
```
cap10morgan, line 246: Love how much async removal this branch has!

cap10morgan (approved):
The usual smattering of mostly minor comments, questions, and suggestions. Looks great overall!

src/clj/fluree/db/indexer/default.cljc
```diff
@@ -510,15 +502,15 @@
...
         {:keys [tempid port]} index-state]
     (if lock?
       ;; when we have a lock, reindex and put updated db onto pc.
       (go
         (try*
           (push-index-event indexer :index-start index-state)
-          (let [indexed-db   (<? (refresh indexer db opts))
-                index-state* (unlock-indexer (:state-atom indexer) branch tempid indexed-db)
+          (let [indexed-db                      (<? (refresh indexer db opts))
+                index-state*                    (unlock-indexer (:state-atom indexer) branch tempid indexed-db)
```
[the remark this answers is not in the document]
zonotope: yeah, i agree. I find it annoying too, but emacs just does it. i know i can configure this because you can configure everything in emacs. the question is whether i can figure out how to configure this in less than an hour...

in any case, i'll take this one out by hand.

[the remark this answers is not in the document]
dpetran: I usually just enter a line of whitespace between the destructuring form and the rest, do a quick `C-c SPC`, and then remove the whitespace. Or not remove the whitespace, depends.

src/clj/fluree/db/json_ld/transact.cljc
dpetran: I don't think so, lists conj to the front and vectors conj to the back, so the order would be reversed.

zonotope: Yeah, I think this is fine. This should never be `nil` because it's [set in the db `create` function](https://github.com/fluree/db/blob/b6d41fb528c93573e9ce08d2f98549ef7b75ba90/src/clj/fluree/db/db/json_ld.cljc#L186), and it's also an explicit field on the db record.

src/clj/fluree/db/ledger/json_ld.cljc
zonotope: I'll reword this.

src/clj/fluree/db/indexer/default.cljc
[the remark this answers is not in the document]
zonotope: yeah, i do that too sometimes, but i also don't really like needing to enter that empty line just because there's a huge destructuring form. it isn't the worst though.

src/clj/fluree/db/json_ld/migrate/sid.cljc
zonotope: I'm not concerned with that in this migration ns because this should be one-time use code to bring existing data to the current (as of April 2024) data format, which is a state where multiple versions of the ledger protocol are still not supported.

Long term though, I think it could very well be the case that we have multiple implementations of the ledger protocol (it's also possible that it won't be necessary, i'm not sure). Even if we do have multiple implementations though, I doubt that we'd want a `db-update` method on that protocol as a ledger's latest db is an internal implementation detail about it's own state that shouldn't be manipulated from the outside. The ledger protocol would have a method like `commit!` or `transact!` instead, and a `db-update` function would be internal to a particular ledger and would be called in the implementation of `commit!` or `transact!` instead. That's why I removed that method from the protocol in this branch. 

Of course, there's still work left to do in streamlining the protocols; this is just a start.

PR #753, opened 2024-04-08
Add `with-meta` support to #Flake

bplatz:
This adds support for Clojure's `with-meta` and `meta` to Fluree's Flake custom type. This was done as there was a need to store internal metadata about a Flake (specifically, originating reasoner rules that generated the Flake).

`(with-meta flake {:some "metadata"})`

Along with the complementary function:
`(meta flake)`

While we already have the `.-m` Flake field, which we call meta, it is used in the equality comparisons (as intended). Therefore an otherwise identical flake but with different metadata will be deemed as different.

But this means we don't have a place to store metadata that is not part of equality. That would involve one of 4 solutions:
1) store this 'non-equality' metadata on a special key on the existing `.-m`. This would have conflated .-m, and also comparators would have to perform a `dissoc` frequently which I was concerned could hurt comparator performance
2) Add a new field on the Flake and implement custom functions
3) Add a new field on the Flake and implement the existing Clojure protocols that add more basic support for with-meta and meta
4) Move to defrecord instead of deftype for Flakes, which would automatically support `with-meta`, `meta`

I opted for #3. This will at a small amount of memory overhead per Flake, but will not require extra comparator logic.

Now that we have this, our current `.-m` field ideally would have been called something other than "meta", but not sure that is worth a more invasive change at the moment.

### DIFF
--- a/src/clj/fluree/db/flake.cljc
+++ b/src/clj/fluree/db/flake.cljc
@@ -31,7 +31,7 @@
 (def ^:const max-meta util/max-integer)
 
 
-(deftype Flake [s p o dt t op m _meta]
+(deftype Flake [s p o dt t op m -clj-meta]
   #?@(:clj  [clojure.lang.Seqable
              (seq [f] (list (.-s f) (.-p f) (.-o f) (.-dt f) (.-t f) (.-op f) (.-m f)))
 
@@ -59,10 +59,10 @@
              (assoc [f k v] (assoc-flake f k v))
 
              clojure.lang.IMeta
-             (meta [_] _meta)
+             (meta [_] -clj-meta)
 
              clojure.lang.IObj
-             (withMeta [_ meta] (Flake. s p o dt t op m meta))
+             (withMeta [_ clj-meta] (Flake. s p o dt t op m clj-meta))
 
              Object
              (hashCode [f] (hash (seq f)))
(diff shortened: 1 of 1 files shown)


zonotope (approved):
🎩

src/clj/fluree/db/flake.cljc
```diff
@@ -31,7 +31,7 @@
 (def ^:const max-meta util/max-integer)
 
 
-(deftype Flake [s p o dt t op m]
+(deftype Flake [s p o dt t op m _meta]
```
zonotope, line 34: Tiny nitpick, but "_" is usually indicates arguments/fields that are never used. We don't use `_meta` in equality calculations, but we do/will use it in other ways. "-" usually indicates arguments/fields/functions that are "private" and shouldn't be used directly, so perhaps `-meta` better conveys the intention. 

I've been trying to think of a better name than "meta" to avoid confusion with the `m` field, but I haven't come up with anything yet. The best I can do is "clj-meta", which I'm not in love with. I don't think it's worth it to change the name of the `m` field at this point because that would be a HUGE change.

bplatz: I just changed it to `-clj-meta` as I like that suggestion the best. Take a look at let me know if it feels wrong.

Maybe we just need to find another word that starts with `m` and act like `.-m` was always for that. :)

PR #757, opened 2024-04-12
dataset API, query any federated dataset

bplatz:
The allows federated query of any number of graph dbs across any number of connections with any unique configurations per db desired (permissions, time-travel).

In order to query against any arbitrary set of graph dbs, you compose those dbs together using the new API `dataset`.

e.g. to compose these 3 dbs into a single dataset:
```
(let [authors (fluree/db @(fluree/ledger conn1 "my/authors"))
      books   (fluree/db @(fluree/ledger conn1 "my/authors"))
      movies  (fluree/db @(fluree/ledger conn1 "my/authors"))
      ds      (fluree/dataset {"test/authors" authors
                               "test/books"   books
                               "test/movies"  movies})]
  @(fluree/query ds <query-here>))
```
In this case, all the dbs are composed as a default graph, but they can still be targeted invidually for specific where statements using the `graph` function:
```
  {...
   'where': [...
             ['graph' <graph-alias> <query-pattern>]]
   ...}
```

If you'd like to have only specific dbs part of the "default graph", you can supply the alias names as a list with a second argument to `fluree/dataset`:
```
(def ds (fluree/dataset {"test/authors" authors
                         "test/books"   books
                         "test/movies"  movies}
                         ["test/books", "test/movies"])
```
In this case, only `test/books` and `test/movies` would be part of the default graph.

This closes issue: https://github.com/fluree/core/issues/96


test/fluree/db/query/federated_test.clj
```diff
@@ -0,0 +1,285 @@
...
+                                                         "@type" "Person"
+                                                         "name"  "Margaret Mitchell"}]})
+          books   @(fluree/create-with-txn conn
+                                           {"@context" ["https://ns.flur.ee" context "https://schema.org"]
+                                            "ledger"   "test/books"
+                                            "insert"   [{"id"     "https://www.wikidata.org/wiki/Q3107329",
+                                                         "type"   ["Book"],
+                                                         "name"   "The Hitchhiker's Guide to the Galaxy",
+                                                         "isbn"   "0-330-25864-8",
+                                                         "author" {"@id" "https://www.wikidata.org/wiki/Q42"}}
```
zonotope, line 34: Ok, I'm really late to this, but I just noticed that the iri here ends with "Q42" :)

zonotope (approved):
📚

dpetran:
Will we also support datasets in transactions? Say I want to update data in one graph based on data in another graph.

PR #758, opened 2024-04-16
ipfs name service changes from PR#726

bplatz:
This hand-applies all of the changes from PR #726 against the latest main branch, as the conflicts were numerous and it was easier this way.

Pasting the description from that PR here:

> IPFS connection types now by default use two nameservices - IPNS and filesystem. The filesystem nameservice is synchronous, and requires a successful write before a commit is confirmed. IPNS is asynchronous and will update as IPNS allows it to.
> 
> Note that any connection can have any number of nameservices configured and fed into the conn as an option - so the old behavior is still available if needed, just not the default.
> 
> At the same time, the file nameservice needed to be updated as it previously only wrote file system path names as the commit-address - but now it needs to be able to write IPFS commit addresses (and potentially others).
> 
> This PR not only makes this address update to the file nameservice, but also
> 
> stores the nameservice file as JSON-LD which can contain additional metadata about the ledger and branches
> Starts to use the branch specification format, as previously discussed, as (<branch-name>) appended to the nameservice address. Because branching isn't fully built out, this is all hidden under the hood for now
> Future steps:
> 
> - Cache the nameservice file so a read isn't required each time (likely with a TTL, and coupled with a notification system that invalidates cache on known updates)
> - Implement a Fluree database nameservice - using the JSON-ld format, the nameservice files could easily be stored in a special nameservice fluree database
> - Allow create API which doesn't create a DB (it wouldn't require a transaction), but reserves the namespace so a db can be placed there
> - Have a status property on the JSON-LD, where actions such as reindexing could be noted, which would make a ledger temporarily unavailable


zonotope (approved):
🦘

PR #759, opened 2024-04-18
return a SHACL Validation Report for validation exceptions

dpetran:
https://www.w3.org/TR/shacl/#validation-report

This utilizes the official SHACL vocabulary for validation report results. I had to augment the official vocabulary with `f:expectation`, as I think that's a very useful part of the validation result.

I also added support for user-specified severity and message keys on the shape, according to the spec.

While updating the test expectations I restructured multiple tests to allow the data and the expectation to live adjacent to each other.


dpetran:
@aaj3f If you've got any thoughts about how errors should look, this is a great PR to discuss them on.

zonotope (approved):
This looks good besides a few small suggestions.

PR #760, opened 2024-04-22
Feature/shacl lang+logical constraints

dpetran:
We were very close and the work was not hard, so I decided to make our SHACL support more complete. 

This PR adds support (and tests) for all remaining SHACL constraints, namely:
- `sh:languageIn`
- `sh:uniqueLang`
- `sh:and`
- `sh:or`
- `sh:xone`

It also adds support for SHACL `sh:alternativePath` paths.

It also adds tests for existing `sh:targetSubjectsOf` and `sh:targetNode` target declarations.

The only unsupported part of SHACL that remains are the recursive `sh:path` variants:
- `sh:zeroOrMorePath`
- `sh:oneOrMorePath`
- `sh:zeroOrOnePath`

### DIFF
--- a/test/fluree/db/shacl/shacl_basic_test.clj
+++ b/test/fluree/db/shacl/shacl_basic_test.clj
@@ -2920,3 +2920,43 @@ Subject ex:InvalidHand path [\"ex:digit\"] violates constraint sh:qualifiedValue
                (ex-data db2)))
         (is (= "Subject ex:invalid path [\"ex:myProperty\"] violates constraint sh:maxCount of shape _:fdb-3 - count 2 is greater than maximum count of 1."
                (ex-message db2)))))))
+
+(deftest target-node
+  (let [conn    @(fluree/connect {:method :memory})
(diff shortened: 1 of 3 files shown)


dpetran:
> I think we are also still missing class inheritance (parent class shapes should apply to child classes).
Thanks, that's absolutely right, I had forgotten about hierarchies.

test/fluree/db/shacl/shacl_basic_test.clj
```diff
@@ -1425,6 +1425,90 @@ WORLD!")
           (is (= "Subject :ex/john path [:ex/birthYear] violates constraint :sh/pattern of shape _:fdb-3 - value \":ex/ref\" does not match pattern \"(19|20)[0-9][0-9]\"."
                  (ex-message db-ref-value))))))))
 
+(deftest language-constraints
```
zonotope, line 1428: These test needs more commentary. What are you testing? What do you expect to happen in these tests?

PR #761, opened 2024-04-24
ledger query-string-t-val

bplatz:
This allows a ledger alias to contain http query strings to specify the moment in time a db should be loaded.

This is used in `fluree/query-connection` whose primary consumer is fluree/server - allowing more granular time travel capabilities in federated (multi-db) queries.

This expanded syntax is supported for both `from` and `from-named` keys in the query syntax, and can now specify a t value like:
   - my/db?t=42
   - my/db?t=2020-01-01T00:00:00Z
 
global-t is still supported as an opt in `query-connection` and will be used if a more specific t value is not supplied for a db as per above.

API test for this syntax in fluree/server

### DIFF
--- a/src/clj/fluree/db/api/query.cljc
+++ b/src/clj/fluree/db/api/query.cljc
@@ -2,6 +2,7 @@
   "Primary API ns for any user-invoked actions. Wrapped by language & use specific APIS
   that are directly exposed"
   (:require [clojure.core.async :as async]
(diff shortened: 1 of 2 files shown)


zonotope (approved):
This looks great

I think we should eventually restrict global-t to *only* allow ISO-8601 datetime strings because integer t-values will not translate across ledgers, but that restriction can of course come later.

PR #762, opened 2024-04-29
Pretty Print DB Records

zonotope:
This patch adds a method to pretty-print db records by selecting a subset of the keys to print as the regular print-methods do. There's also some miscellaneous cleanup.


dpetran (approved):
:+1:

PR #764, opened 2024-04-29
Allow bare id maps in where clause

zonotope:
We used to support matching all subjects in a ledger by including a bare id map like `{"@id" ?s}` in a where clause. We implemented this by matching the "@id" predicate directly. We however removed id predicates in the switch to deterministic subject ids, so this patch adds back the bare id map search functionality by adding a new match pattern time that filters subject ids from any flakes.


dpetran (approved):
:star:

PR #766, opened 2024-05-06
Content addressed storage

zonotope:
This patch moves our storage implementations to be more content addressed. There were a few places where code outside the storage implementation produced content hashes that were then passed down and used by the storage implementation, and these hashes often didn't match the data that was actually being stored. This patch relies on the storage implementations themselves to produce content hashes and faithfully stores the data under those hashes that was passed to it. All of the paths and addresses are also generated by the storage implementations and not by any code outside of it. This should further separate the storage concern to be behind the storage abstraction and should help these storage records to be truly interchangeable. 

This patch also removes the ".json" suffix from stored files because we can't assume all the files we store will always be json, and it removes `address` from the storage abstraction because the address because the address will only be reliably known after the data is stored, and nothing outside of the storage abstraction should need to generate an address outside of the context of storing data, so they should rely on the `:address` field of the map returned by a successful storage operation. 

I also found one error in legacy nameservice lookups where an async operation was treated as synchronous. I fixed that issue in https://github.com/fluree/db/commit/95ba0919119dfc538ec7bfec31a794e87ecb8aa4 and https://github.com/fluree/db/commit/ff60dd2e091feeb55753b8b90fd0bb3d9d076134

### DIFF
--- a/src/clj/fluree/db/conn/file.cljc
+++ b/src/clj/fluree/db/conn/file.cljc
@@ -33,10 +33,9 @@
           bytes    (bytes/string->UTF8 json)
           hash     (crypto/sha2-256 bytes :hex)
           type-dir (name data-type)
-          path     (str alias
-                        (when branch (str "/" branch))
-                        (str "/" type-dir "/")
-                        hash ".json")
+          path     (->> [alias branch type-dir hash]
+                        (remove nil?)
+                        (str/join "/"))
 
           {:keys [hash address]} (<? (storage/write store path bytes))]
       {:name    path
--- a/src/clj/fluree/db/conn/file.cljc
+++ b/src/clj/fluree/db/conn/file.cljc
@@ -2,7 +2,6 @@
   (:require [clojure.core.async :as async :refer [go]]
             [fluree.db.util.async :refer [<? go-try]]
             [clojure.string :as str]
-            [fluree.crypto :as crypto]
             [fluree.db.util.core :as util]
             [fluree.json-ld :as json-ld]
             [fluree.db.index :as index]
@@ -31,13 +30,12 @@
                      data
                      (json-ld/normalize-data data))
           bytes    (bytes/string->UTF8 json)
-          hash     (crypto/sha2-256 bytes :hex)
           type-dir (name data-type)
-          path     (->> [alias branch type-dir hash]
+          path     (->> [alias branch type-dir]
                         (remove nil?)
                         (str/join "/"))
 
-          {:keys [hash address]} (<? (storage/write store path bytes))]
+          {:keys [path hash address]} (<? (storage/write store path bytes))]
       {:name    path
        :hash    hash
        :json    json
--- a/src/clj/fluree/db/conn/memory.cljc
+++ b/src/clj/fluree/db/conn/memory.cljc
@@ -10,7 +10,6 @@
             [fluree.db.conn.cache :as conn-cache]
             [fluree.db.indexer.default :as idx-default]
             [fluree.json-ld :as json-ld]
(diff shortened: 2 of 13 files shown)


src/clj/fluree/db/conn/file.cljc
```diff
@@ -31,14 +30,12 @@
                      data
                      (json-ld/normalize-data data))
           bytes    (bytes/string->UTF8 json)
-          hash     (crypto/sha2-256 bytes :hex)
           type-dir (name data-type)
-          path     (str alias
-                        (when branch (str "/" branch))
-                        (str "/" type-dir "/")
-                        hash ".json")
```
dpetran, line 39: Are we going to write a migration to handle this breaking change? I really regretted putting this suffix here but didn't remove it when I did the storage refactor in order to maintain compatibility. I think we do need to remove it in order to have a proper storage protocol, though.

zonotope: I'm going to roll it in to the sid migration if it's necessary since that migration hasn't been released yet and no one that I know of has run it.

bplatz: I think the .json was useful because we knew the encoding of the file (as a side benefit, it was also useful to double-click on commit or data files and have the editor pop open the file without extra steps). The file store is the most obvious place where this surfaces... because IPFS actually embeds the encoding into its format, and memory store doesn't really matter. I don't think anyone touched S3... so not sure if it ever did the right thing.

The nice thing about JSON-LD is the standard defines all three of (a) serialization, (b) encoding, (c) hashing algo... but if we want to support other formats we need to consider how to store potentially all three of those pieces of metadata somehow.

I don't see us favoring anything but JSON-LD for commit and data files, but I could see a strong argument for storing index files in a more compact/speedy format. We could again rely on file extension here (e.g. `.avro`) or we'll need to move to some existing or custom format where this is encoded into either the file name or some binary metadata.

zonotope: I could see us eventually storing a more compact representation even for commit and especially data files at some point because there's a lot of opportunity for compression. There are definitely advantages to using the suffix when we know the format, but I don't think we can rely on the format being the same, especially since we could use the same storage implementation to store commit files, data files, and indexes, which you could argue could each be optimally stored in a different format. I think baking in the json suffix ties our hands

I think the storage implementation is most flexible if it's only concerned with putting and pulling bytes too/from disk. We can rely on the things built on top of it to understand what format those bytes are and how to interpret them, and those higher level components can worry about performing whatever (de)serialization required before things are written and after they're read from storage.

bplatz: My comment is more along the lines that if storage is just pushing/pulling bytes, then that's probably all it should do. Right now it is it is determining hash algorithm, file name and encoding.

Unfortunately IPFS combines all of these things into one protocol... which sort of means we have no choice but to also embed these things into every storage protocol... so it can't be a simple as pushing/pulling bytes.

But now our FS storage no longer does this. It used to (using the .json suffix) but now doesn't. It seems we would need a mechanism to do this if we want to support multiple file formats.

I'm not saying `json` can't be passed as an argument to storage instead of hard coded like it was, and storage (a) might append it as suffix `.json`, while storage (b) might encode it in a binary header (like IPFS does), or storage (c) might use multi-format and encode it into the file name characters.

zonotope: Ok, I think I understand what you're getting at, and I think that makes sense. I think we can add back the suffix in the file store for now and figure out what to do if/when we start supporting multiple formats.

PR #767, opened 2024-05-09
Include reasoning logic when restricting db

JaceRockman:


JaceRockman:
This is meant to solve Story #1 from https://github.com/fluree/core/issues/102

JaceRockman:
If anyone has ideas on how to get a db of rules from just an alias when using `query` instead of `query-connection`, let me know, otherwise, I think this is good to merge.

src/clj/fluree/db/api/query.cljc
```diff
@@ -160,11 +171,16 @@
...
-        (<? (restrict-db db t* context opts)))
+            address            (<? (nameservice/primary-address conn alias nil))
+            ledger             (<? (jld-ledger/load conn address))
+            db                 (ledger/-db ledger)
+            t*                 (or explicit-t t)
+            rules-db           (let [dbs-or-aliases (:reasoner-rules-db opts)]
+                                 (if (string? (first dbs-or-aliases))
+                                   [(ledger/-db (<? (jld-ledger/load conn (first dbs-or-aliases))))]
+                                   dbs-or-aliases))
+            opts*              (assoc opts :reasoner-rules-db rules-db)]
```
cap10morgan, line 182: Very minor, but if possible can you configure your editor to ignore destructuring forms for alignment purposes? They just tend to get pretty long and make the other bindings in the let form align too far to the right. So many exceptions to every rule... ;)

src/clj/fluree/db/api/query.cljc
```diff
@@ -43,13 +47,20 @@
...
+                           db)
+          time-travel-db (-> (if t
+                               (<? (time-travel/as-of policy-db t))
+                               policy-db))
+          reasoned-db    (let [{:keys [reasoners reasoner-rules reasoner-rules-db]} opts]
+                           (if reasoners
+                             ;; Currently we only support one rule source, so we take the first db or first
+                             ;; reason graph that we find.
+                             (<? (reasoner/reason
+                                  time-travel-db reasoners (or (first reasoner-rules-db) (first reasoner-rules)) opts))
```
cap10morgan, line 61: Can you split this form up into multiple lines so it doesn't extend so far to the right?

src/clj/fluree/db/api/query.cljc
```diff
@@ -43,13 +47,20 @@
...
-          db** (-> (if t
-                     (<? (time-travel/as-of db* t))
-                     db*))]
-      (assoc-in db** [:policy :cache] (atom {})))))
+    (let [policy-db      (if-let [policy-identity (perm/parse-policy-identity opts context)]
+                           (<? (perm/wrap-policy db policy-identity))
+                           db)
+          time-travel-db (-> (if t
+                               (<? (time-travel/as-of policy-db t))
+                               policy-db))
```
cap10morgan, line 55: Love these more descriptive var names!

cap10morgan (approved):
Looks good overall! I noted a couple of very minor formatting things to address just in the interest of minimizing whitespace churn in the future. Thanks!

src/clj/fluree/db/api/query.cljc
bplatz: With the new policy PR, this has changed a little bit so the above wouldn't be compatible. Now you compose dbs, instead of a bunch of if conditions in the API. In this case if you wanted a db to be policy-wrapped, you'd call that API first explicitly. This no longer looks for this option in a regular query.

Also if a query is wrapped in a verifiable credential, there is a new explity query API for that... so this check won't be done always.

JaceRockman: It looks like that PR has failing checks, should I wait for it to be merged and then adjust this PR to work with it?

PR #768, opened 2024-05-10
Index wasn't updating latest commit t file

bplatz:
When an index was in-process but a new commit happens before it finishes, the index lock state updates the latest commit such that when the indexing finishes, it can update that commit.

It was updating the latest commit in the wrong key, therefore once the indexing completed it was always returning the same commit when the indexing started, not the more updated commit if it existed.

This fixes the key location where it updates the latest commit.


zonotope (approved):
I'm working on some changes that will leave both the indexing lock state and update commit function unnecessary, but I think this is a good fix for the underlying problem in the interim.

PR #769, opened 2024-05-10
SPARQL VALUES patterns

dpetran:
In SPARQL, a VALUES pattern can appear inside a WHERE clause alongside other patterns. In FQL :values is its own clause. This mismatch makes translating directly awkward, since we need to insert our translated pattern at a different point in the parse tree than where we are at translation time.

We got around it by translating it into a :bind pattern. However, a :bind pattern can only bind a single value to a variable, so we limited the scope of a VALUES pattern to only a single value.

This gets around it by "tagging" the translated value pattern and then after the whole parsing is done, moving it to the correct place in the tree: see `format-values`. It's awkward, but it gives much more flexibility to the user.


bplatz:
Should we move :values into the where clause? I'd imagine it is there because there could be multiple where clauses, or sub-queries which might have its own set of values.

dpetran:
Rebased on https://github.com/fluree/db/pull/777

dpetran:
Rebased on https://github.com/fluree/db/pull/778

dpetran:
> > We got around it by translating it into a :bind pattern. However, a :bind pattern can only bind a single value to a variable, so we limited the scope of a VALUES pattern to only a single value.
> 
> `:bind` patterns can take any number of mappings of values to variables.

Right, maybe it is better stated that a :bind pattern can only bind a single value _per_ variable, whereas :values can bind multiple values per variable.

src/clj/fluree/db/query/sparql/translator.cljc
```diff
@@ -484,6 +542,5 @@
 
 (defn translate
   [parsed]
-  (reduce (fn [fql rule] (into fql (parse-rule rule)))
-          {}
-          parsed))
+  (->> parsed
```
zonotope, line 545: I don't think you need the threading macro here because there's just one step.

src/clj/fluree/db/query/sparql/translator.cljc
```diff
@@ -484,6 +542,5 @@
...
+       (reduce (fn [fql rule] (into fql (parse-rule rule))) {})))
```
zonotope, line 546: this would be much more readable if you moved the function body after the argument vector to the next line

zonotope (dismissed):

zonotope (commented):
I didn't mean to approve because of the data type issue.

dpetran:
Rebased on https://github.com/fluree/db/pull/777

src/clj/fluree/db/query/fql/parse.cljc
```diff
@@ -79,8 +79,10 @@
         (let [expanded (json-ld/expand-iri val context)]
           (where/match-iri var-match expanded))
         (where/match-value var-match val dt-iri))
-      (let [dt (datatype/infer-iri val)]
-        (where/match-value var-match val dt)))))
+      (if-let [lang (get attrs const/iri-language)]
+        (where/match-value var-match val const/iri-string {:lang lang})
```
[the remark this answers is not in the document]
dpetran: I had originally use `const/iri-lang-string` here but that broke other tests, so I changed it what you see here. After investigating after your comment it looks like we weren't actually assigning the correct datatype, so I've fixed that and this instance.

PR #771, opened 2024-05-13
Indexing process state management

bplatz:
This is a temporary fix to solve the problem of bulk loads under high server demand. It should be replaced by work being done by @zonotope which eliminates ever attempting to update a commit file.

If an index finishes, it attempts to update the latest commit. In a high load scenario, by the time it can update the latest commit there is a new commit. This has some bad effects.

We now have a temporary atom state that maintains the latest index, the two main process changes are:
1) Upon index completion, that index atom is updated
2) When writing new commits, that index atom is always consulted for the latest index files
3) When updating an existing commit with a new index, before writing a check is performed if there have been any newer commits and if so, it no longer will attempt an update at all

While this gets things working fairly reliably, there are still some race conditions that exist and this is not by iteself a permanent fix. At the point we no longer try to update commits it is irrelevant however.

### DIFF
--- a/src/clj/fluree/db/indexer/default.cljc
+++ b/src/clj/fluree/db/indexer/default.cljc
@@ -518,11 +518,11 @@
                 {:keys [update-commit-fn port]} index-state*]
             ;; in case event listener wanted final indexed db, put on established port
             (when (fn? update-commit-fn)
-              (let [result (<! (update-commit-fn indexed-db))]
+              (let [result (<! (update-commit-fn indexed-db))] ;; will return 'nil if commit not updated
                 (when (util/exception? result)
                   (log/error result "Exception updating commit with new index: " (ex-message result))
                   (throw result))
-                (when changes-ch
+                (when (and changes-ch result)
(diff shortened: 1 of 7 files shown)


src/clj/fluree/db/json_ld/reify.cljc
```diff
@@ -302,13 +302,13 @@
...
 
-          {:keys [previous issuer message] :as commit-metadata}
+          {:keys [previous issuer message data] :as commit-metadata}
           (commit-data/json-ld->map commit db)
 
           commit-id          (:id commit-metadata)
           commit-sid         (iri/encode-iri db commit-id)
           [prev-commit _]    (some->> previous :address (read-commit conn) <?)
-          db-sid             (iri/encode-iri db alias)
+          db-sid             (iri/encode-iri db (:id data))
```
zonotope, line 311: Wow. good catch!

zonotope (approved):

bplatz:
Closing because main now has features that address this isssue.

PR #772, opened 2024-05-14
Fix/db sid load

bplatz:
When generating commit metadata flakes, the `db-id` was trying to use the db's alias instead of the `@id` of the data file in the commit in two of the three places it is used.

This resulted in `nil` sids, and while it could load in simple dbs, once indexed we'd start getting errors.

This uses @id instead of alias and all three locations are consistent.

Also made change to log out merge flake exceptions, even if verbose. This is a place many `nil` flake issues are caught as it throws when trying to create a db 'size' calc on `nil`

### DIFF
--- a/dev/commit_explore.clj
+++ b/dev/commit_explore.clj
@@ -1,0 +1,67 @@
+(ns commit-explore
+  (:require [clojure.java.io :as io]
+            [fluree.db.util.json :as json]))
+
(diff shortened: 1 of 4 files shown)


bplatz:
This was approved by @zonotope  in PR # 771 (where this branch was merged into that branch)

PR #773, opened 2024-05-14
vocab's :class not longer used, updated asserts

bplatz:
`:class` in vocab/schema was not being used - removed.

Updated some asserts/doc strings to be current.


dpetran (approved):
:+1:

PR #774, opened 2024-05-15
Normalized branch state

zonotope:
This patch removes the cyclic dependency between ledgers -> branches -> and dbs by removing references to the ledger from the db record and loading a db directly from a commit (and only that way). 

It also adds an explicit "genesis commit" at t = 0 that helps us load new dbs and will (in the future) help us create ledgers without an explicit transaction. The genesis commit is currently being persisted with this patch, but I decided to leave both pushing the genesis commit out to the nameservices, adding genesis commit metadata flakes to the ledger, and using the genesis commit to detect if a ledger exists, which will allow us to create ledgers without transactions, for later. Those tasks require some reorganization of how the name services work, and there are higher priority changes to index processing that I'd like to tackle before that.


bplatz:
Any chance you can merge #766 and re-target this to main?

dpetran (approved):
Looks good :+1:

src/clj/fluree/db/json_ld/migrate/sid.cljc
```diff
@@ -33,98 +33,98 @@
...
+;;           assert           (reify/db-assert db-data)
+;;           asserted-flakes  (reify/assert-flakes ns-mapping t-new assert)
+;;           retract          (reify/db-retract db-data)
+;;           retracted-flakes (reify/retract-flakes ns-mapping t-new retract)
+;;           db*              (set-namespaces db ns-mapping)
 
-          {:keys [previous issuer message data] :as commit-metadata}
-          (commit-data/json-ld->map commit db*)
+;;           {:keys [previous issuer message data] :as commit-metadata}
+;;           (commit-data/json-ld->map commit db*)
```
cap10morgan, line 56: Should we delete all this commented-out code?

zonotope: I decided to comment it out for now because we will very soon have to re-write this code, so it won't stay this way for long, and because it won't be a wholesale rewrite since the underlying logic of the migration will be the same. There's just a few indexing and db loading apis that have/will change.

PR #775, opened 2024-05-17
Feature/schema in index

bplatz:
This PR serializes the full `:schema` in the index-root.

Previously, the `:schema` was partially serialized, but required a series of queries upon db load to get all the data that was required for a full `:schema`. This results in loading possibly many index nodes and was a substantial delay for large databases.

In one large database used for testing, this change reduces the db load time from ~45 seconds to ~15 seconds (still a long time, but substantially better).

Now, a db can 'load' without requiring any queries.

### DIFF
--- a/src/clj/fluree/db/json_ld/iri.cljc
+++ b/src/clj/fluree/db/json_ld/iri.cljc
@@ -191,8 +191,12 @@
    (iri->sid iri default-namespaces))
   ([iri namespaces]
    (let [[ns nme] (decompose iri)]
-     (when-let [ns-code (get namespaces ns)]
-       (->sid ns-code nme)))))
+     (if-let [ns-code (get namespaces ns)]
+       (->sid ns-code nme)
+       (throw (ex-info (str "Unexpected error: Namespace not registered in the database "
+                            "for iri: " iri ".")
+                       {:status 500
+                        :error :db/unexpected-error}))))))
 
 (defn next-namespace-code
   [namespaces]
--- a/src/clj/fluree/db/json_ld/vocab.cljc
+++ b/src/clj/fluree/db/json_ld/vocab.cljc
@@ -376,3 +376,106 @@
             (assoc :pred pred-map)
             (update-with db t vocab-flakes)
             (add-pred-datatypes (filterv #(> (count %) 1) preds)))))))
+
+;; schema serialization
+(def ^:const serialized-pred-keys [:id :datatype :subclassOf :parentProps :childProps])
+(def ^:const serialized-pred-keys-reverse (reverse serialized-pred-keys))
+
+(defn schema-tuple
+  [pred-map]
+  (reduce
+   (fn [acc next-key]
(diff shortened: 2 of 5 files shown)


src/clj/fluree/db/json_ld/vocab.cljc
```diff
@@ -372,3 +376,106 @@
...
+                       (mapv schema-tuple))]
+    {"t"    t
+     "pred" {"keys" pred-keys
+             "vals" pred-vals}}))
+
+(defn deserialize-pred-tuple
+  "Takes list of keys and tuples containing values
+  and turns them into a map with the respective keys and tuples"
+  [namespace-codes positions pred-vals]
+  (let [base-map base-property-map
```
zonotope, line 432: I don't think we need this extra binding here. We can just use the `base-property-map` directly below while initializing the `loop`.

src/clj/fluree/db/json_ld/iri.cljc
```diff
@@ -191,8 +191,12 @@
    (iri->sid iri default-namespaces))
   ([iri namespaces]
    (let [[ns nme] (decompose iri)]
-     (when-let [ns-code (get namespaces ns)]
-       (->sid ns-code nme)))))
+     (if-let [ns-code (get namespaces ns)]
+       (->sid ns-code nme)
+       (throw (ex-info (str "Unexpected error: Namespace not registered in the database "
```
zonotope, line 196: 👍🏾

zonotope (approved):

src/clj/fluree/db/util/core.cljc
```diff
@@ -206,7 +206,7 @@
   (reduce-kv
     (fn [acc k v]
       (if (keyword? k)
-        (assoc acc (name k) v)
+        (assoc acc (subs (str k) 1) v)
```
dpetran, line 209: If we're keeping the namespace of namespaced keywords I think the docstring for this function should be updated as well.

src/clj/fluree/db/json_ld/vocab.cljc
bplatz: I do this intentionally when I'm using a constant and the function is not pure. Because it is so infrequent to do that, this helps call out this is happening.

But, it comes at a tiny cost, and especially if it isn't helping achieve that goal it serves no purpose - so I'll go ahead and get it change.

src/clj/fluree/db/util/core.cljc
bplatz: I'll revert this back to its original state - I needed this behavior for a feature and realized it would still work with the other uses.

But then I removed the feature, so the change is no longer needed.

src/clj/fluree/db/db/json_ld.cljc
bplatz, line 1: @zonotope - this addresses having to query the db with only deserialized `:preds` to create the entire schema. It requires an intermediary db to use for querying, which then the final `:schema` can be attached to.

PR #765, opened 2024-05-01
SPARQL -> FQL translator refactor

dpetran:
This is a rewrite of the SPARQL translator that solves several issues, and should make adding support for more SPARQL syntax easier.

Specifically fixes :union patterns in where clauses, subject iri parsing, and REGEX functions.

fixes https://github.com/fluree/core/issues/90

PR #777, opened 2024-05-17
Feature/values pattern

dpetran:
This adds support for `:values` as a where-clause pattern. This allows users to scope their `:values` to just a `:union` or `:optional` pattern, instead of constraining the whole `:where` clause.

This will enable us to translate the SPARQL VALUES pattern directly to FQL.


src/clj/fluree/db/query/exec/where.cljc
```diff
@@ -566,6 +566,23 @@
...
+  (let [[_ inline-solutions] pattern
+        ;; need to compute sids on inline data so they can match equal solutions
+        inline-solutions* (mapv (fn [solution]
+                                  (->> solution
+                                       (mapv (fn [[var match]]
+                                               [var (if (matched-iri? match) (compute-sid match db) match)]))
+                                       (into {} )))
+                                inline-solutions)]
+    ;; filter out any solutions that don't match inline solution data
+    (if (every? (fn [inline-solution] (= (select-keys solution (keys inline-solution))
```
zonotope, line 580: I don't think this will work in all cases. There could be other metadata that we pull from the dbs that won't be in the inline solution. There might also be problems in the context of federated queries where a subject was pulled from a different database than the one being evaluated now, so the sid that you compute above won't match. I think you have to pull the value or iri out specifically instead of relying on the entire match map to be equal with the inline one for each inline variable.

src/clj/fluree/db/query/fql/parse.cljc
```diff
@@ -570,7 +575,7 @@
 (defn parse-analytical-query
   [q]
   (let [context       (context/extract q)
-        [vars values] (parse-values q context)
+        [vars values] (parse-values (:values q) context)
```
zonotope, line 578: Was there a problem with passing the entire query to the `parse-values` function? I would prefer to follow that pattern because there are some steps which might need to get information from multiple parts of the query, and I think the parsing code is clearer when each parsing function acts on the query consistently.

dpetran: Not so much a problem as it being more inconvenient. `parse-values` is called in three places:
- `parse-analytical-query`, where passing the entire query is easy
- `parse-txn`, where the values have to be put into an artificially constructed query map
- `parse-pattern :values`, where we would have to artificially construct a query map

On the balance of uses, I thought it was cleaner to refactor it to just accept the values instead of assuming that it comes attached to a query, since most of the time it's actually not.

src/clj/fluree/db/query/exec/where.cljc
dpetran: Hmm, maybe `:lang` from :meta as well? I'm not sure about `:i`, though.

src/clj/fluree/db/query/exec/where.cljc
```diff
@@ -566,6 +566,22 @@
                           clause-ch)
     out-ch))
 
+(defmethod match-pattern :values
+  [db fuel-tracker solution pattern error-ch]
+  (let [inline-solutions (pattern-data pattern)
+        ;; need to remove ::sids from the solution as those may differ based on which graph they are from
+        solution* (reduce (fn [soln [var match]]
+                            (assoc soln var (cond-> (dissoc match ::sids)
```
zonotope, line 574: I think this is still brittle and could unintentionally break if we add more information to the matches. Instead of removing known divergent information from the solution and doing a top level equality check, I think we should reduce over the variables in the inline solution and explicitly check if the value and datatype of the match bound to that variable in the solution is the same as the value and datatype of the inline solution and, if there is no value/datatype, we should check if it's a matched iri and the iris are the same. The `get-value` and `get-datatype` functions will help you in the first case, and the `get-iri` will help you in the second. 

We should explicitly check the things we know must be the same, instead of explicitly eliminating the things we think might be different, because we could add more things that might be different later.

dpetran: Good point, I've updated this section (and discovered the very handy `update-vals` core fn, thanks Rich) to select the important bits.

dpetran:
Rebased on https://github.com/fluree/db/pull/765

zonotope (approved):
🥼

dpetran:
The CI was stuck waiting for a container so I merged and ran CI locally.

PR #782, opened 2024-05-28
Manage indexing jobs with a queue

zonotope:
This patch schedules indexing jobs for new commits with a dropping queue to ensure we are always working on the latest commit. It also separates commit updates to ledger/branch state from index updates to simplify and streamline those updates, and all of these updates and decisions about when/if to make them are executed entirely within an atomic `swap!` to branch state to eliminate race conditions. 

In order to enable atomic updates to index state, new dbs are loaded in the background within an `AsyncDB`. To enable interoperability between `AsyncDB` and `JsonLdDb`, I had to add more protocols for history queries, policy restriction, and time travel.

I also added a new indexing protocol for db record types that replaces the standalone indexer records and `iIndex` protocols. With this patch, all index and commit updates to ledger state are atomic and without any locks.

There is still _a lot_ of cleanup and reorganization to do, but this patch is the minimal changeset necessary to manage indexing jobs with a queue.


src/clj/fluree/db/ledger/json_ld.cljc
```diff
@@ -90,22 +96,155 @@
...
-      (jld-commit/commit ledger db* opts*)
-      (throw (ex-info (str "Cannot commit db, as committed 't' value of: " committed-t
-                           " is no longer consistent with staged db 't' value of: " t ".")
-                      {:status 400 :error :db/invalid-commit})))))
+  "Finds all uncommitted transactions and wraps them in a Commit document as the subject
+  of a VerifiableCredential. Persists according to the :ledger :conn :method and
+  returns a db with an updated :commit."
+  [{:keys [alias conn] :as ledger} {:keys [t stats commit] :as db} opts]
+  (go-try
+    (let [{:keys [did message tag file-data? index-files-ch] :as opts*}
```
dpetran, line 200: Does anything actually supply an `index-files-ch` opt yet? Is that something that `server` will supply?

src/clj/fluree/db/db/json_ld.cljc
```diff
@@ -373,7 +397,62 @@
...
+      (validate/allow-iri? db sid)))
+
+  indexer/Indexed
+  (collect [db changes-ch]
+    (if (idx-default/novelty-min? db reindex-min-bytes)
+      (idx-default/refresh db changes-ch)
+      (go)))
+
+
+  TimeTravel
```
dpetran, line 413: I'm liking these new protocols a lot. This looks like a nice step towards decomplecting : )

src/clj/fluree/db/ledger/json_ld.cljc
dpetran: Also, a nitpick: `index-files-ch` may be a slight misnomer if the FileStore isn't being used. Maybe `index-node-ch`?

bplatz: fluree/server supplies it, and it is used to distributed the index file notifications across the network.

dpetran (approved):
:+1:

PR #783, opened 2024-05-28
Feature/sparql graph

JaceRockman:


src/clj/fluree/db/query/sparql/translator.cljc
```diff
@@ -366,6 +366,16 @@
   [[_ source]]
   (parse-term source))
 
+(defmethod parse-term :GraphGraphPattern
+  ;; GraphGraphPattern ::= <'GRAPH'> VarOrIri GroupGraphPattern
+  [[_ & [var-or-iri group-graph-pattern]]]
+  [:graph (parse-term var-or-iri) (into [] (parse-term group-graph-pattern))])
```
dpetran, line 372: I suspect you don't need the `into` here, since the `:GroupGraphPattern` should return a vector.

JaceRockman: Without it, I get `({@id ?x, foaf:mbox ?mbox})` instead of `[{@id ?x, foaf:mbox ?mbox}]`

dpetran: Ah, never mind then :+1:

dpetran (approved):
Nice work!

PR #780, opened 2024-05-23
Feature/sparql from named

JaceRockman:
Support for translating SPARQL `FROM` and `FROM NAMED` clauses to FQL.

PR #784, opened 2024-05-28
Feature/fql exists not exists

dpetran:
This adds support for filtering based on the existence or non-existence of specified patterns in the data, analogous to SPARQL's `FILTER EXISTS` and `FILTER NOT EXISTS` functionality.


test/fluree/db/query/negation_test.clj
```diff
@@ -0,0 +1,46 @@
+(ns fluree.db.query.negation-test
+  (:require  [clojure.test :as t :refer [deftest testing is]]
+             [fluree.db.json-ld.api :as fluree]
+             [fluree.db.test-utils :as test-utils]))
+
+(deftest minus
```
zonotope, line 6: This tests needs more commentary. It's easier on people coming to the test later if you describe, in English, what the setup code is doing and *why* the queries should return what they do. `clojure.test` includes facilities to give that commentary with the `testing` macro and the second argument to the `is` macro. We should use them more.

dpetran:
Rebased on https://github.com/fluree/db/pull/780

dpetran:
Rebased on https://github.com/fluree/db/pull/780

PR #787, opened 2024-05-29
Fix formatting in a few namespaces

cap10morgan:


cap10morgan:
* **#788** <a href="https://app.graphite.dev/github/pr/fluree/db/788?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#787** <a href="https://app.graphite.dev/github/pr/fluree/db/787?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @cap10morgan and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

src/clj/fluree/db/database/async.cljc
```diff
@@ -84,30 +83,27 @@
...
+            (log/error e "Error loading database")
+            (>! error-ch e))))
       prop-ch))
 
   (-iri-visible? [_ iri]
     (go-try
       (let [db (<? db-chan)]
         (<? (jld-response/-iri-visible? db iri)))))
 
-
```
zonotope, line 96: i think these extra spaces make things more readable because it better separates the different protocol implementations

cap10morgan: Yeah, I don't necessarily disagree, but the tradeoff is that I don't think we can easily encode that as a rule exception in a tool like cljfmt. And I would rather move towards automatic formatting w/ a tool like that and give up some of these nice formatting exceptions.

cap10morgan: The other option is that we could disable the default-true `:remove-consecutive-blank-lines?` setting in cljfmt. But we have quite a bit of arbitrary inconsistency here that I would prefer to eliminate as a source of incidental git churn over preserving the places where it can aid in readability.

zonotope: yeah, that makes sense. we can't automate these checks and have them be perfect all the time. at least not yet.

cap10morgan:
### Merge activity

* **May 30, 12:20 PM EDT**: @cap10morgan started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/fluree/db/787).

PR #789, opened 2024-05-30
Refactor/flake db

zonotope:
This patch is part of a series that reorganizes the code in light of the new db types. This patch renames the `JsonLdDb` record to `FlakeDB`, and puts the namespace that defines that record as well as all of it's supporting namespace under the `fluree.db.flake....` namespace hierarchy. I think `JsonLdDb` is ambiguous because all of our DB types represent json-ld data, but I think `FlakeDB` is more precise to how this specific DB type represents that data.


dpetran (approved):
:+1:

dpetran (approved):
Looks good :+1:

zonotope:
closing in favor of #842 and #844

PR #792, opened 2024-05-31
Shacl optimization

zonotope:
This patch ensures that we run a query for shacl shapes present in a db only once while validating the set of modified subjects in a transaction, and that we bypass shacl all together if there are no shapes to validate. 

There's also an optimization for new index jobs processed in the index queue to always use the latest known index before processing the new indexing job. This should prevent the indexer from repeating work it has already done. 

this patch depends on #785, so please review that one first.


dpetran (approved):
If I'm understanding correctly, these are the optimizations:

For SHACL, we only look up the shape sids once, instead for every single subject in the transaction.

Also, when processing a db from index queue we'll make sure to assoc in the most recent index, instead of whatever the index was when the db was put on the indexing queue.

zonotope:
> If I'm understanding correctly, these are the optimizations:
> 
> For SHACL, we only look up the shape sids once, instead for every single subject in the transaction.
> 
> Also, when processing a db from index queue we'll make sure to assoc in the most recent index, instead of whatever the index was when the db was put on the indexing queue.

Not quite for the second one. We don't assoc in the recent index because that wouldn't actually build the index in the db. We reload the db with the recent index so both the index roots and novelty are correct.

PR #794, opened 2024-06-03
Feature/subquery

bplatz:
Adds new subquery capability. Supports multiple parallel subqueries, or nested subqueries.

A new `solution-value` interface is for select statements to pass formatted solutions for :select clauses instead of final projections as a normal select clause does.

e.g. in EDN - this has a subquery that calculates the average of all favNums, then in the parent query filters to only include users who have a favNum > than the average:
```
{:context  {:ex "http://example.org/ns/"}
 :select   '[?iri ?favNums]
 :where    ['{:id         ?iri
              :ex/favNums ?favNums}
            [:filter "(> ?favNums ?avgFavNum)"]
            [:query {:where  '{:ex/favNums ?favN}
                     :select '[(as (avg ?favN) ?avgFavNum)]}]]
 :order-by '[?iri ?favNums]}
```

Tests/examples in: `fluree.db.query.subquery-test`


src/clj/fluree/db/query/exec.cljc
```diff
@@ -42,22 +43,85 @@
...
 
+(defn extract-subquery
+  [query-smt]
+  (when (and (sequential? query-smt)
+             (= :query (first query-smt)))
+    (second query-smt)))
+
+(defn execute
+  [ds fuel-tracker q error-ch initial-soln]
+  (->> (where/search ds q fuel-tracker error-ch initial-soln)
```
zonotope, line 54: I think there's some room to consolidate everything before the format step into it's own function, and to consolidate the offset and limit steps into another function too since so much logic is shared between the `execute` and `execute-subquery` functions.

src/clj/fluree/db/query/exec/select.cljc
```diff
@@ -112,8 +122,15 @@
...
               (catch* e
-                (log/error e "Error applying aggregate selector")
-                (>! error-ch e))))))
+                      (log/error e "Error applying aggregate selector")
+                      (>! error-ch e)))))
+  (solution-value
+    [_ error-ch _]
+    (async/put!
+     error-ch
+     (ex-info "Aggregate selectors cannot be used in subqueries without 'as' syntax."
```
zonotope, line 131: Ideally we'd catch this in the parsing phase. We have enough information to be able to return an error before we even start reading the indexes. Malli should be able to detect this and throw before we execute any part of the query.

src/clj/fluree/db/query/exec/select.cljc
```diff
@@ -157,7 +177,14 @@
                          where/get-iri)
                      subj)]
       (json-ld-resp/format-node ds iri context compact spec iri-cache
-                                fuel-tracker error-ch))))
+                                fuel-tracker error-ch)))
+  (solution-value
+    [_ error-ch _]
+    (async/put!
+     error-ch
+     (ex-info "Graph crawl select statements cannot be used in subqueries."
```
zonotope, line 185: Another thing we should detect during the parsing phase.

src/clj/fluree/db/query/exec/select.cljc
```diff
@@ -52,16 +53,19 @@
 
 (defprotocol ValueSelector
   (implicit-grouping? [this]
-   "Returns true if this selector should have its values grouped together.")
+    "Returns true if this selector should have its values grouped together.")
   (format-value [fmt db iri-cache context compact fuel-tracker error-ch solution]
-   "Formats a where search solution (map of pattern matches) by extracting and displaying relevant pattern matches."))
+    "Async format a search solution (map of pattern matches) by extracting relevant match.")
+  (solution-value [fmt error-ch solution]
```
zonotope, line 59: I feel like this should be in a new subquery specific protocol because the solution-value method doesn't apply in all the cases where ValueSelector should be defined.

src/clj/fluree/db/query/exec/select.cljc
```diff
@@ -52,16 +53,19 @@
...
+    "Formats value for subquery select statement as k-v tuple - synchronous."))
 
 ;; This exists because many different types of data structures in :select
 ;; clauses get implicit-grouping? called on them. So this defaults them to false.
 (extend-type #?(:clj Object :cljs object) ; https://cljs.github.io/api/cljs.core/extend-type
   ValueSelector
   (implicit-grouping? [_] false)
-  (format-value [_ _ _ _ _ _ _ _] nil))
+  (format-value [_ _ _ _ _ _ _ _] nil)
+  (solution-value [_ _ _] nil))
```
zonotope, line 68: If `solution-value` is in a new protocol, I don't think we should have a default implementation because there are some cases where it explicitly should not be allowed.

src/clj/fluree/db/query/fql/parse.cljc
```diff
@@ -567,21 +567,30 @@
...
+         parse-having
+         (parse-select context)
+         parse-fuel))))
+
+(defmethod parse-pattern :query
+  [[_ sub-query] _vars context]
+  (let [sub-query* (-> sub-query
+                       syntax/coerce-query
+                       (parse-analytical-query context))]
+    [(where/->pattern :query sub-query*)]))
```
zonotope, line 593: This probably doesn't matter, but I don't think this needs to be a `where` pattern because subqueries are executed by different machinery. It is not explicitly executed by the where-clause processing and doesn't have a corresponding `match-pattern` method in the `where` namespace like all other items created by `where/->pattern`.

src/clj/fluree/db/query/exec.cljc
```diff
@@ -42,22 +43,85 @@
...
+  [subquery-chans]
+  (when (seq subquery-chans)
+    (if (= 1 (count subquery-chans))
+      (first subquery-chans)
+      (let [out-ch (async/chan)]
+        (go
+          (let [all-solns (loop [chans subquery-chans
+                                 acc   []]
+                            (if-let [next-chan (first chans)]
+                              (let [solns (async/<! (async/into [] next-chan))]
```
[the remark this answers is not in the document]
bplatz: I pondered this and weighed the complexity against the following assumptions:

a) subqueries will more often than not be doing aggregates/calculations, and have limited results (often just 1 result). If you are doing large joined solutions, I think you'd typically use a union or optional.

b) In looking for example subqueries online, having a multiple subqueries is uncommon. For the common single-subquery case, it is wired to use the lazy chan in a very efficient way.

I think this adds support and will mostly be fine. A bullet-proof solution is better of course, but I feel this is much lower probability here than other areas we have memory concerns around. We can always do more work on this in the future.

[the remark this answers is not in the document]
zonotope: Yeah, that makes sense. This code is modular enough that we can always revisit if the need arises.

src/clj/fluree/db/query/exec.cljc
bplatz: I wanted to consolidate these more, but the logic *is* different, even though some functions are shared. Wrapping these in multiple steps IMO makes it harder to follow the logic and right now I think they both read very clearly.

I can change if you feel differently, but I did go down that path and reverted back here.

src/clj/fluree/db/query/exec/select.cljc
bplatz: I started this change, but both `format-value` and `solution-value` are tied to `implicit-grouping?`. So either `solution-value` needs to rely on the other protocol. At that point it seems they should be part of the same protocol.

Do you still want them separated out?

src/clj/fluree/db/query/exec/select.cljc
bplatz: Based on what I can figure out from Malli, the only way to do this is to create a new subquery definition that is mirrors query which is defined here: https://github.com/fluree/db/blob/f3dafc7d81c0741cb096304d32312747635e4a34/src/clj/fluree/db/query/fql/syntax.cljc#L35-L76  with a nearly identical function that duplicates everything except the two small tweaks to select.

This duplicationwoudl be a lot of duplication, but my larger concern is it is error-prone as any future features/syntax updates to query will have to be made to *both* definitions to keep them consistent.

Admittedly I'm no expert here, but I've spent a good deal of time here not only in the malli the docs, but also been in source code and don't see another path. If there is one I'd love to learn about it.

src/clj/fluree/db/query/exec/select.cljc
bplatz: Note previous related comment.

src/clj/fluree/db/query/exec/select.cljc
bplatz: Note previous related comment

src/clj/fluree/db/query/exec/select.cljc
zonotope: Yeah, I have had reservations about the addition of `implicit-grouping?` as a protocol method, and especially as a method on _this_ protocol since it was introduced because whether or not something is grouped implicitly is outside the scope of selecting values. 

I think the best thing would be to [go back to the way it was before](https://github.com/fluree/db/blob/e21b2039bb3a086510c378634115b1f0dac43ebd/src/fluree/db/query/exec/select.cljc#L137-L140) and just add the other aggregate selector types to that function as necessary. That would eliminate problems like this and result in more flexibility. 

I think that's probably beyond the scope of this particular pr so I might just make that change along with the changes I suggested here in a separate cleanup branch at some point.

src/clj/fluree/db/query/exec/select.cljc
zonotope: I think we could refactor that function to make it have a minimal amount of duplication and allow future changes that apply to both query and subquery definitions only need to happen in one place because Malli schemas are just normal Clojure data. We could break out the common parts of the schema into it's own definitions, and then add to/transform them as necessary to create the specific schemas we need:

```suggestion
(def common-query-schema
  [:map {:closed true}
   [:from {:optional true} ::from]
   [:from-named {:optional true} ::from-named]
   [:where {:optional true} ::where]
   [:t {:optional true} ::t]
   [:context {:optional true} ::context]
   [:orderBy {:optional true} ::order-by]
   [:order-by {:optional true} ::order-by]
   [:groupBy {:optional true} ::group-by]
   [:group-by {:optional true} ::group-by]
   [:having {:optional true} ::function]
   [:values {:optional true} ::values]
   [:limit {:optional true} ::limit]
   [:offset {:optional true} ::offset]
   [:maxFuel {:optional true} ::max-fuel]
   [:max-fuel {:optional true} ::max-fuel]
   [:depth {:optional true} ::depth]
   [:opts {:optional true} ::opts]
   [:prettyPrint {:optional true} ::pretty-print]
   [:pretty-print {:optional true} ::pretty-print]])

(defn wrap-query-map-schema
  [schema]
  [:and
   [:map-of ::json-ld-keyword :any]
   [:fn {:error/fn
         (fn [_ _]
           (str "Query does not have exactly one select clause. "
                "One of 'select', 'selectOne', 'select-one', 'selectDistinct', or 'select-distinct' is required in queries. "
                "See documentation here for more details: "
                docs/error-codes-page "#query-missing-select"))}
    one-select-key-present?]
   schema])

(defn query-schema
  "Returns schema for queries, with any extra key/value pairs `extra-kvs`
  added to the query map.
  This allows eg http-api-gateway to amend the schema with required key/value pairs
  it wants to require, which are not required/supported here in the db library."
  [extra-kvs]
  (-> common-query-schema
      (into [[:select {:optional true} ::select]
             [:selectOne {:optional true} ::select]
             [:select-one {:optional true} ::select]
             [:selectDistinct {:optional true} ::select]
             [:select-distinct {:optional true} ::select]])
      (into extra-kvs)
      wrap-query-map-schema))

(defn subquery-schema
  [extra-kvs]
  (-> common-query-schema
      (into [[:select {:optional true} ::subquery-select]
             [:selectOne {:optional true} ::subquery-select]
             [:select-one {:optional true} ::subquery-select]
             [:selectDistinct {:optional true} ::subquery-select]
             [:select-distinct {:optional true} ::subquery-select]])
      (into extra-kvs)
      wrap-query-map-schema))
```
that 👆🏾 should work if `::subquery-select` is defined, and changes to any non-select part of the schema would automatically apply to both queries and subqueries. We could also change the `wrap-query-map-schema` function definition to take the error message and validator function as arguments, but it seems like the same function should work in both places so I opted not to do that.

src/clj/fluree/db/query/exec/select.cljc
bplatz: Addressed in e041dc93c2e681f21fbbc5814cd98ae80aa956c9

src/clj/fluree/db/query/exec/select.cljc
bplatz: Resolved in d99b491bb69a12e2c0fd175d737005e4ca2419b9

src/clj/fluree/db/query/exec/select.cljc
bplatz: I don't plan to resolve this here - I don't think it is specific to the subquery functionality and can sit as a separate change. We are also still doing some query work, so opinions on ideal solution might change.

src/clj/fluree/db/query/exec/select.cljc
bplatz: Addressed in d99b491bb69a12e2c0fd175d737005e4ca2419b9

bplatz:
I think all the issues are addressed here @zonotope. @dpetran would be great to move ahead with SPARQL subquery support to go with this.

zonotope (approved):
🌞

PR #796, opened 2024-06-04
translate SPARQL EXISTS/NOT EXISTS filters into FQL

dpetran:
Adds support for translating SPARQL EXISTS and NOT EXISTS filters into FQL.

### DIFF
--- a/src/clj/fluree/db/query/sparql/translator.cljc
+++ b/src/clj/fluree/db/query/sparql/translator.cljc
@@ -64,14 +64,12 @@
                     {:status 400 :error :db/invalid-query}))))
 
 (defmethod parse-term :ExistsFunc
-  [r]
-  (throw (ex-info "EXISTS is not a supported SPARQL clause"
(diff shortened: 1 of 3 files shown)


src/clj/fluree/db/query/sparql/translator.cljc
```diff
@@ -446,7 +444,10 @@
 (defmethod parse-term :Filter
   ;; Filter ::= <'FILTER'> WS Constraint
   [[_ constraint]]
-  [:filter [(parse-term constraint)]])
+  (let [parsed-constraint (parse-term constraint)]
+    (if (contains? #{"exists" "not-exists"} (first parsed-constraint))
```
dpetran, line 448: Because FQL implements `exists` and `not-exists` as top-level where patterns instead of as subordinate to a `filter` pattern like SPARQL I had to do a little manual unpacking of the parse tree in the translator here.

zonotope (approved):

PR #798, opened 2024-06-05
Fix irregular deletions after indexing

zonotope:
There were two issues that prevented delete (and other) transactions from correctly modifying data and/or queries from returning accurate results after data was modified. 

The first issue stemmed from the `max-namespace-code` field on `JsonLdDb` records not being incremented after a transaction. This caused new namespace codes to be generated for previously encountered namespaces, and left flakes that should have had the same subject id (or other iri flake components) with different subject ids because there were two different namespace codes for the same namespace. This was fixed in 311f375.

The second issue stemmed from our flake deserialization process converting `nil` into `{}` for the flake meta component. That meant that flakes in novelty could have `nil` as the meta value, but `{}` when they move from novelty to the index nodes. This caused queries to return inconsistent results because retracting flakes in novelty could not be linked to their assertions in the index nodes. This was fixed in 32c799c.

The rest of these changes are a bit of cleanup and debugging aides.

### DIFF
--- a/src/clj/fluree/db/storage.cljc
+++ b/src/clj/fluree/db/storage.cljc
@@ -34,6 +34,10 @@
      :method method
      :local  local}))
 
+(defn parse-local-path
+  [address]
+  (-> address parse-address :local))
+
 (defprotocol Store
   (write [store k v] "Writes `v` to Store associated with `k`. Returns value's address.")
   (exists? [store address] "Returns true when address exists in Store.")
--- a/src/clj/fluree/db/db/json_ld.cljc
+++ b/src/clj/fluree/db/db/json_ld.cljc
@@ -363,6 +363,10 @@
           db        (assoc file-data "f:address" db-address)]
       (json-ld/expand db))))
 
+(defn get-max-ns-code
+  [ns-codes]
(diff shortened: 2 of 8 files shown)


src/clj/fluree/db/storage.cljc
```diff
@@ -34,6 +34,10 @@
      :method method
      :local  local}))
 
+(defn parse-local-path
```
dpetran, line 37: Does this get used anywhere?

dpetran (approved):
:deciduous_tree:

src/clj/fluree/db/storage.cljc
zonotope: It does not get used (yet). I have another branch that implements this function and uses it, and I have some as yet un checked in index exploration code (similar to Brian's excellent [commit exploration code](https://github.com/fluree/db/blob/0f557334b9dd77179c1d91bd36745584ab3e5edb/dev/commit_explore.clj#L1)) that also uses it. I don't think the index exploration code is quite ready for committing yet, and I was going to remove this function, but I decided to leave it in since I know it will soon be useful, but I'm happy to take it out if you or anyone else disagrees.

PR #800, opened 2024-06-05
Include opts in SPARQL queries

JaceRockman:
Allow opts to be added to SPARQL queries for credential purposes.


src/clj/fluree/db/api/query.cljc
```diff
@@ -215,36 +215,36 @@
...
 
 (defn query-connection-fql
-  [conn query]
+  [conn query {:keys [role did]}]
   (go-try
-    (let [{query :subject, did :did} (or (<? (cred/verify query))
-                                         {:subject query})
-          {:keys [t opts] :as query*}  (update query :opts sanitize-query-options did)
+    (let [credentialed-opts-query     (update query :opts #(merge {:role role :did did} %))
+          {query :subject, did :did}  (or (<? (cred/verify credentialed-opts-query))
```
dpetran, line 221: I don't think we can modify the query and still expect to verify the results here. The query may not even be a map, it could be just a JWS string.

JaceRockman: If the query is just a JWS string, would there be some way to add the `role` and `did` to it?

dpetran: `cred/verify` returns a map of `:subject` and `:did`, if whatever it is verifying is valid. I don't think we should override the `did`, though - the supplied did should only be used if a query (or txn) is not signed.

bplatz:
Is this still desired, or should this be closed?

src/clj/fluree/db/api/query.cljc
JaceRockman: Ok, it sounds like I should just be inserting the :role and :did when we convert the sparql to fql, that way it won't mess with any JWS style queries that are submitted.

JaceRockman:
> Is this still desired, or should this be closed?

I think we want to be able to restrict SPARQL queries by role still. Why would we no longer want that?

src/clj/fluree/db/api/query.cljc
JaceRockman: Although, this is part of the `query-connection-fql` function, so can that be a JWS string? It sounds like it would have to be fql.

bplatz:
> I think we want to be able to restrict SPARQL queries by role still. Why would we no longer want that?

That sounds good. The new policy-queries no longer has roles, there is the new `wrap-identity-policy` API which would be the closest approximation to this.

src/clj/fluree/db/query/api.cljc
```diff
@@ -207,36 +207,36 @@
...
                         {:status 400, :error :db/invalid-query}))))))
 
-
 (defn query-connection-sparql
-  [conn query]
+  [conn query {:keys [role did] :as request-opts}]
   (go-try
-    (let [fql (sparql/->fql query)]
-      (<? (query-connection-fql conn fql)))))
+    (let [fql              (sparql/->fql query)
```
dpetran, line 234: You'll need to do a `verify` before this call, as you may not have an actual sparql string for a query right here.

Something like this:
```
(let [{sparql :subject, did :did}  (or (<? (cred/verify credentialed-opts-query))
                                                          {:subject query})
         fql  (sparql/->fql query)]
    ....)
```

JaceRockman: That feels confusing, because this function will only get called if the request's `Content-Type` is set to `application/sparql-query`. So even for sparql queries, the query itself can be submitted as a JWS string?

dpetran: All the way down here we don't know that this function call is the result of an http request - that's all handled by `server`. So while we know that `server` only calls this function after verifying the query, there will be other callers that are not `server`. So we need to call `cred/verify` in this path as well.

JaceRockman: Ok, I think this must have been broken previously then, because we weren't verifying any credentials here previously. This will also make it so that we get the query in `query-connection`, and if it's sparql, then we verify the credentials in the sparql->fql step and then verify the credentials again in `query-connection-fql`. But maybe I'm just not understanding what exactly our verification process for credentials is doing or when it's necessary.

dpetran: Yeah, we actually verify multiple times. If the thing we're verifying isn't a credential or a jws, then `cred/verify` just returns `nil`, so that's why it's always wrapped in an `or` expression. And I think you're right that it was broken before, I don't think we were really considering authentication for sparql queries.

src/clj/fluree/db/json_ld/credential.cljc
```diff
@@ -132,13 +135,17 @@
...
-                        :error  :db/invalid-credential}))))))
+  ([signed-command]
+   (verify signed-command nil))
+  ([signed-command {format :format}]
+   (go-try
+     (let [result (if (jws? signed-command)
+                    (verify-jws signed-command format)
+                    (<? (verify-credential signed-command)))]
+       (println signed-command)
+       (println result)
```
dpetran, line 146: Let's remove the `println`s.

src/clj/fluree/db/api.cljc
```diff
@@ -295,7 +295,7 @@
   ([ds cred-query {:keys [default-allow? values-map] :as opts}]
    (promise-wrap
     (go-try
-     (let [{query :subject, identity :did} (<? (cred/verify cred-query))]
+     (let [{query :subject, identity :did} (<? (cred/verify cred-query opts))]
```
dpetran, line 298: I think we need to do this in `credential-query-connection` as well.

test/fluree/db/json_ld/credential_test.cljc
```diff
@@ -224,7 +229,21 @@
                 (async/<!! (cred/generate {:history (:id auth)
                                            :t       {:from 1}}
                                           (:private pleb-auth)))))
-           "history query credential - forbidding access"))))
+           "history query credential - forbidding access")
+
+       (is (= [["D"]]
```
dpetran, line 234: Let's add another test for `credential-query-connection` too.

src/clj/fluree/db/query/api.cljc
```diff
@@ -209,36 +208,36 @@
...
+          credentialed-fql (update fql :opts #(merge {:role role :did did} %))]
+      (<? (query-connection-fql conn credentialed-fql request-opts)))))
 
 (defn query-connection
-  [conn query {:keys [format did] :as opts :or {format :fql}}]
-  (case format
-    :fql (query-connection-fql conn query did)
-    :sparql (query-connection-sparql conn query did)))
-
+  [conn query {:keys [format] :as request-opts}]
```
dpetran, line 240: Instead of `request-opts`, can we just call this `opts`? It's not guaranteed that this will be called during the processing of a request.

JaceRockman: I was hoping to differentiate it somehow between the opts that we put directly into queries and transactions. Would something like `external-opts` be better?

dpetran: I'm not sure there is a meaningful distinction there, we use opts to refer to both everywhere in the api, so I think it's best to be consistent with that.

dpetran: That's a good point, and I'm a believer that different things should have different names. Can you open a ticket for renaming the opts that aren't part of the document? Then we can decide the best name for it there and get it fixed across the board, rather than just fixing it locally here.

src/clj/fluree/db/query/api.cljc
```diff
@@ -209,36 +208,36 @@
...
+                        {:status 400, :error :db/invalid-query}))))))
 
 (defn query-connection-sparql
-  [conn query did]
+  [conn query {:keys [role did] :as request-opts}]
   (go-try
-   (let [fql (sparql/->fql query)]
-     (<? (query-connection-fql conn fql did)))))
+    (let [fql              (sparql/->fql query)
+          credentialed-fql (update fql :opts #(merge {:role role :did did} %))]
```
dpetran, line 236: Instead of merging the opts into the fql, I think we can just pass them as `opts` to `query-connection-fql`.

JaceRockman: I don't think the :role makes its way into the query at any point if I just include it as `opts` in `query-connection-fql` unless that isn't how roles are supposed to work anymore?

dpetran: I don't think we use `role` anymore. @bplatz did we deprecate the `role` opt?

bplatz: Yes, `role` is no longer part of policy enforcement nor used anywhere else.

src/clj/fluree/db/api.cljc
```diff
@@ -293,7 +293,7 @@
...
```
bplatz, line 296: Per the many comments in https://github.com/fluree/db/pull/843, nobody wanted to create additional options to `cred/verify` - so most of the commits after the first one were to take an alternative approach.

This will need to be refactored to take the same (or different) approach.

src/clj/fluree/db/json_ld/credential.cljc
```diff
@@ -113,14 +113,17 @@
...
                                      (throw (ex-info (str "Invalid JWS: " jws)
                                                      {:status 400
                                                       :error  :db/invalid-credential})))
         id                       (crypto/account-id-from-public pubkey)
-        auth-did                 (did/auth-id->did id)]
-    {:subject (json/parse payload false) :did auth-did}))
+        auth-did                 (did/auth-id->did id)
+        processed-payload        (if (= :sparql format)
+                                   payload
+                                   (json/parse payload false))]
```
bplatz, line 125: Per the alternative approach taken in https://github.com/fluree/db/pull/843 we should not be parsing JSON in the function at all.

src/clj/fluree/db/query/api.cljc
```diff
@@ -204,33 +204,33 @@
...
+    (let [fql (sparql/->fql query)]
+      (<? (query-connection-fql conn fql opts)))))
 
 (defn query-connection
-  [conn query {:keys [format did] :as _opts :or {format :fql}}]
-  (case format
-    :fql (query-connection-fql conn query did)
-    :sparql (query-connection-sparql conn query did)))
+  [conn query {:keys [format] :as opts}]
+  (case (or format :fql)
```
bplatz, line 234: Usually we'd handle this in as part of the destructuring syntax (if we were destructuring, which we are here).

src/clj/fluree/db/query/api.cljc
```diff
@@ -204,33 +204,33 @@
...
-       (let [ds       (<? (load-dataset conn default-aliases named-aliases t opts))
-             query**  (update query* :opts dissoc :meta :max-fuel ::util/track-fuel?)
-             max-fuel (:max-fuel opts)]
-         (if (::util/track-fuel? opts)
-           (<? (track-query ds max-fuel query**))
-           (<? (fql/query ds query**))))
-       (throw (ex-info "Missing ledger specification in connection query"
-                       {:status 400, :error :db/invalid-query}))))))
-
+    (let [{:keys [t opts] :as sanitized-query} (-> query
```
bplatz, line 209: This appears to overwrite the `opts` function argument... so it is unused.

bplatz: Actually I see (:did opts) is pulled from it before overwriting.

If only `:did` is being used, then it is far more clear to have `did` as a function argument than an opaque `opts` map what we don't know what is supposed to be inside it.

src/clj/fluree/db/query/api.cljc
```diff
@@ -34,10 +34,10 @@
     did (assoc :did did :issuer did)))
 
 (defn restrict-db
-  [db t opts]
+  [db t {:keys [did default-allow?]}]
   (go-try
-    (let [db*  (if-let [did (:did opts)]
-                 (<? (perm/wrap-identity-policy db did true nil))
+    (let [db*  (if did
+                 (<? (perm/wrap-identity-policy db did default-allow? nil))
```
bplatz, line 40: I think this change may break some `fluree/server` tests. You should verify you don't need a new PR to `fluree/server` to still work based on the changes here. If so, that PR should also get queued up.

src/clj/fluree/db/api.cljc
```diff
@@ -351,8 +351,7 @@
          (and query identity)
          (let [policy-db (<? (policy/wrap-identity-policy latest-db
                                                           identity
-                                                          (or (true? default-allow?)
-                                                              false)
+                                                          default-allow?
```
bplatz, line 354: This should be accompanied by a malli change that ensures only boolean values are include.

This changes means instead of having to be `true`, it only needs to but truthy... so even an empty string (`""`) or empty vector (`[]`) or anything other than `nil` would dramatically change the security characteristics of the ledger.

For things affecting security, we must ensure the most restrictive options are always taken until explicit intention is shown. This loosens that.

src/clj/fluree/db/query/api.cljc
JaceRockman: I ran the `fluree/server` tests pointed at these latest changes and they all passed.

bplatz (approved):
LGTM!

PR #802, opened 2024-06-06
Stop throwing error when namespace hasn't been registered

zonotope:
Throwing errors in this case causes queries for iris with namespaces that haven't previously been registered to throw an error while queries for iris that have registered namespaces but don't exist in the database return an empty result set. These two types of queries should behave the same because they are both querying for non-existent iris.


bplatz (approved):
Good change! We may end up wanting a separate API to check if a namespace is registered/exists - but we can wait till we have a use case.

PR #804, opened 2024-06-11
Include reasoning logic when restricting db

JaceRockman:
This is a redo of PR #767. I reverted it by mistake so I built a new branch and readded the changes.

### DIFF
--- a/src/clj/fluree/db/api/query.cljc
+++ b/src/clj/fluree/db/api/query.cljc
@@ -40,7 +40,7 @@
   [db t context opts]
   (go-try
     (let [policy-db      (if-let [policy-identity (perm/parse-policy-identity opts context)]
-                           (<? (perm/wrap-policy db policy-identity))
+                           (<? (perm/wrap-identity-policy db policy-identity false nil))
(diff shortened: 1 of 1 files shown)


src/clj/fluree/db/api/query.cljc
```diff
@@ -152,11 +166,16 @@
...
-            t*      (or explicit-t t)]
-        (<? (restrict-db db t* context opts)))
+            address  (<? (nameservice/primary-address conn alias nil))
+            ledger   (<? (jld-ledger/load conn address))
+            db       (ledger/-db ledger)
+            t*       (or explicit-t t)
+            rules-db (let [dbs-or-aliases (:reasoner-rules-db opts)]
+                       (if (string? (first dbs-or-aliases))
+                         [(ledger/-db (<? (jld-ledger/load conn (first dbs-or-aliases))))]
+                         dbs-or-aliases))
```
bplatz, line 176: Here it is assumed there is a valid db object/var. I think this is the right point to ensure that is the case or throw an exception

PR #805, opened 2024-06-12
Set max namespace code at commit instead of stage

zonotope:
This patch sets the match namespace code at commit time instead of at the time the database is staged. This allows us to accurately compute the new namespaces generated in a commit and save them to the commit file, which, in turn, allows us to load the correct namespaces afterwards after a new index update.


dpetran (approved):
:lock:

PR #806, opened 2024-06-16
Revert "Merge pull request #790 from fluree/feature/policy-query2"

zonotope:
The policy-query2 branch causes test to fail when integrated with fluree/server. This patch reverts this branch temporarily to allow us to continue to create working server images while identifying and fixing the test failures.

This reverts commit f9423304a85176e81ffdc1ecc3f2f396a2bdd724, reversing changes made to 8f1f988831a4d14a7531b6d27cc39f9bc3650030.


bplatz:
There are some tests on fluree/server that transact f:equals data and tries policy around them, and that older policy mechanism is deprecated. Those tests need to be modified so they use the new policy query... but in the interim they can also be removed as they are not intended to work.

I'll get some new tests in fluree/server to replace them that use the new policy framework.

zonotope:
> There are some tests on fluree/server that transact f:equals data and tries policy around them, and that older policy mechanism is deprecated. Those tests need to be modified so they use the new policy query... but in the interim they can also be removed as they are not intended to work.
> 
> I'll get some new tests in fluree/server to replace them that use the new policy framework.

That works. We don't have to merge this branch if we know what's going on. I mainly needed to get a new image with some recent bug fixes out so I needed something to build off of that both passed ci and had the bug fixes.

bplatz:
FYI, working on updating fluree/server to support the policy maps but it exposes a couple additional features that were needed as well to control overall access to the server itself. Plan to have some PRs this week.

bplatz:
There is now a PR that upates fluree/server to use the new policy queries, so this can be closed.

https://github.com/fluree/server/pull/73

PR #810, opened 2024-06-19
handle `as` and `AS` in SPARQL queries

dpetran:
The spec only defines an all-caps AS expression in the SPARQL grammar, but it is common for SPARQL processors to be case indifferent


cap10morgan (approved):
*️⃣

PR #812, opened 2024-06-22
Fix/nested query results

zonotope:
@fluree/core this patch fixes a bug where query results didn't include all nested data that was previously committed because that nested data did not make it into the indexes.

The problem arose because the `:leftmost?` key was not being serialized when we wrote out index roots. This in turn meant that any new sid committed after the index root was written and was ordered lower than the rest of the data in the index root would never make it into an index node because we use the `:leftmost?` key to determine which is the lowest index node to incorporate such flakes with low order sids into. 

This has been a long standing bug (maybe back to version 2 even), but we haven't run into it until now because of the confluence of events necessary to trigger it. 

The other changes here are a dev namespace to help load and read index files for debugging, and some general cleanup.


zonotope:
@aaj3f I think you'll be interested in this too.

dpetran (approved):
:+1:

PR #813, opened 2024-06-24
Fluree db api namespace

zonotope:
This patch moves the top level fluree db api namespace from `fluree.db.json-ld.api` to `fluree.db`.


bplatz:
Do you think we should retain fluree.db.json-ld.api (perhaps with deprecation warnings) for the interim?

I'm OK with either, just want to highlight it as a decision to weigh in on.

zonotope:
> Do you think we should retain fluree.db.json-ld.api (perhaps with deprecation warnings) for the interim?
> 
> I'm OK with either, just want to highlight it as a decision to weigh in on.

Good point. I'll restore it and add deprecation warnings.

src/clj/fluree/db.cljc
```diff
@@ -0,0 +1,521 @@
...
+  [conn commit-map]
+  (promise-wrap
+    (if (map? commit-map)
+      (notify-ledger conn commit-map)
+      (go
+        (ex-info (str "Invalid commit map, perhaps it is JSON that needs to be parsed first?: " commit-map)
+                 {:status 400 :error :db/invalid-commit-map})))))
+
+
+(defn index
```
dpetran, line 181: Should we remove all of these stubs? I don't know that we even expect to use all of these anymore.

src/clj/fluree/db/json_ld/api.cljc
```diff
@@ -30,7 +31,9 @@
 
 (declare query)
 
-(defn promise-wrap
+(defn ^{:deprecated    "3.0"
```
dpetran, line 34: Do these deprecations issue any warnings to the user? And if so, is it at build time or run time?

src/clj/fluree/db.cljc
zonotope: yeah, it's probably best to remove the stubs, especially from this namespace. it won't break the API if/when we add the functions back upon implementation.

src/clj/fluree/db/json_ld/api.cljc
zonotope: As far as I know, only dev time linters and IDE tools like clj-kondo and clojure-lsp (which uses clj-kondo) report on these metadata; there's no compiler option to print a warning as far as I can tell. We also should be able to make use of them in our docs generator to report on them as well (but I have to confirm this).

Do you think we need something more explicit?

dpetran: I think a `log/warn` message would be appropriate, something simple like `<fn> DEPRECATED. Superseded by <fn>`.

bplatz:
> This patch moves the top level fluree db api namespace from `fluree.db.json-ld.api` to `fluree.db`.

It appears this is now `fluree.db.api` (discussion above).

Would be great to get this merged (from my perspective) ASAP. I have a couple new APIs and will probably rebase here as to not create conflicts.

dpetran (approved):
:mountain:

zonotope:
> > This patch moves the top level fluree db api namespace from `fluree.db.json-ld.api` to `fluree.db`.
> 
> It appears this is now `fluree.db.api` (discussion above).
> 
> Would be great to get this merged (from my perspective) ASAP. I have a couple new APIs and will probably rebase here as to not create conflicts.

It's `fluree.db.api` instead of `fluree.db` because the rename to `fluree.db` inexplicably broke the browser tests.
```
Chrome Headless 126.0.6478.126 (Linux x86_64): Executed 0 of 16 SUCCESS (0 seLOG: 'Testing fluree.db-test'
Chrome Headless 126.0.6478.126 (Linux x86_64): Executed 0 of 16 SUCCESS (0 seChrome Headless 126.0.6478.126 (Linux x86_64) ERROR
  Uncaught TypeError: Cannot read properties of undefined (reading 'load')
  at browser-tests.js:21077:7
  
  TypeError: Cannot read properties of undefined (reading 'load')
      at Function.$fluree$db$database$async$load_$_state_machine__86847__auto____1$$ [as $cljs$core$IFn$_invoke$arity$1$] (browser-tests.js:81805:86)
      at $cljs$core$async$impl$ioc_helpers$run_state_machine_wrapped$$ (browser-tests.js:21074:140)
      at browser-tests.js:81862:12
      at $cljs$core$async$impl$dispatch$process_messages$$ (browser-tests.js:20545:135)
      at z (browser-tests.js:174:210)
      at m (browser-tests.js:175:160)
```
I had thought this was because of the [known limitations of cljs namespaces that allows namespace segments and functions to clobber each other](https://clojurescript.org/about/differences#_namespaces), but I renamed everything that could possibly clash as a test and got the same error. 

It looks like this error is due to the `AsyncDB` type not being defined for some reason, but I haven't been able to identify why that type is missing from the browser build. `fluree.db.api` works for now, but I would like to revisit this error later if we have time to see if `fluree.db` is possible, but I think we have higher priority items left to be done.

PR #815, opened 2024-06-27
fix SPARQL unicode grammar regex for CLJS

dpetran:
Clojure and Clojurescript default to using their respective platforms' RegExp implementations, and unfortunately the syntax for unicode characters is mutually incompatible.

I've tried several variations to get it to work in the same file, but have unfortunately failed. So the next best thing is the dumb solution, which is two separate grammars, one for cljs and one for clj.

The only difference is in the PN_CHARS_BASE terminal, where we need to specify the unicode flag (the `(?u)` prefix) and use ES2015 Unicode-aware regular expressions syntax: `[\u{10000}-\u{EFFFF}]`.

These are the sources I used to figure it out:
https://mathiasbynens.be/notes/es6-unicode-regex
https://yo-dave.com/2018/04/20/notes-on-clojurescript-regular-expressions/

### DIFF
--- a/resources/sparql.js.bnf
+++ b/resources/sparql.js.bnf
@@ -1,0 +1,251 @@
+<Query> ::=  <COMMENT?> WS Prologue WS ( SelectQuery | ConstructQuery | DescribeQuery | AskQuery ) WS Modifiers
+UpdateUnit ::= Update
+Prologue ::= ( BaseDecl | PrefixDecl )*
+BaseDecl ::= <'BASE'> WS IRIREF
+PrefixDecl ::= <'PREFIX'> WS PNAME_NS WS IRIREF
+SelectQuery ::= WS SelectClause WS DatasetClause WS WhereClause WS SolutionModifier WS
+SubSelect ::= SelectClause WhereClause SolutionModifier ValuesClause
+SelectClause ::= WS <'SELECT'> WS ( 'DISTINCT' | 'REDUCED')? ( ( WS ( Var | ( <'('> Expression As Var <')'> ) ) )+ | ( WS '*' ) )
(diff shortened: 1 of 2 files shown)


resources/sparql.js.bnf
```diff
@@ -0,0 +1,251 @@
...
+EXPONENT	  ::=  	#"[eE][+-]?[0-9]+"
+<STRING_LITERAL1>	  ::=  	<"'"> ( #"[^\x27\x5C\x0A\x0D]" | ECHAR )* <"'">
+<STRING_LITERAL2>	  ::=  	<'"'> ( #"[^\x27\x5C\x0A\x0D]" | ECHAR )* <'"'>
+<STRING_LITERAL_LONG1>	  ::=  	"'''" ( ( "'" | "''" )? ( #"[^'\\]" | ECHAR ) )* "'''"
+<STRING_LITERAL_LONG2>	  ::=  	'"""' ( ( '"' | '""' )? ( #"[^'\\]" | ECHAR ) )* '"""'
+ECHAR	  ::=  	#"\\[tbnrf]"
+NIL	  ::=  	'(' WS* ')'
+<WS>	  ::=  	<#"[\x20\x09\x0D\x0A]*\#[^\n]*\n*[\x20\x09\x0D\x0A]*|[\x20\x09\x0D\x0A]*">
+ANON	  ::=  	'[' WS* ']'
+<PN_CHARS_BASE>	  ::=  	#"(?u)[a-zA-Z]|[\u00C0-\u00D6]|[\u00D8-\u00F6]|[\u00F8-\u02FF]|[\u0370-\u037D]|[\u037F-\u1FFF]|[\u200C-\u200D]|[\u2070-\u218F]|[\u2C00-\u2FEF]|[\u3001-\uD7FF]|[\uF900-\uFDCF]|[\uFDF0-\uFFFD]|[\u{10000}-\u{EFFFF}]"
```
dpetran, line 241: This is the line that required different syntax. The important bits are the flag prefix: `(?u)` and the unicode character syntax for multibyte codepoints: `[\u{10000}-\u{EFFFF}]`

PR #816, opened 2024-06-28
fix pending subjects-as-predicates test

dpetran:
We were not using the right data for the test - db3 was unused and it was the only db that had an edge between ex:subject-as-predicate and ex:nested via ex:unlabeled. As soon as I changed it to the correct data source the results met expectations.

### DIFF
--- a/test/fluree/db/query/property_test.clj
+++ b/test/fluree/db/query/property_test.clj
@@ -142,7 +142,7 @@
                                                    "ex:parent" ?parent}}))))
             "returns all sub properties of ex:parent property")))))
 
-(deftest ^:integration ^:kaocha/pending subjects-as-predicates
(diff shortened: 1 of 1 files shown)


dpetran:
Closes https://github.com/fluree/core/issues/98

zonotope (approved):
Nice work tracking this down. 

This would have been much easier (at least for me) to track down if we had more explicit names for the database bindings, grouped the setup code along with the assertions that directly use that setup code, and used the `testing` and `is` macros to describe in English what was being tested and what the expectations were instead of initializing all the database values in the top and then calling all of the assertions to save a bit of nesting. I think this whole tests should be restructured to make catching problems like this easier, but it's outside the scope of this particular pr right now.

I'm glad you were able to track down the issue, and that it didn't point to a bug in the code 👍🏾

PR #822, opened 2024-07-08
deterministic where patterns

dpetran:
If a node-pattern in a where clause has multiple predicates, the order that they are converted into triple patterns isn't deterministic. This commit sorts the keys before converting them into new patterns.

### DIFF
--- a/src/clj/fluree/db/query/fql/parse.cljc
+++ b/src/clj/fluree/db/query/fql/parse.cljc
@@ -683,7 +683,9 @@
                        :else            (where/match-iri id))]
     (reduce (partial parse-pred-cmp allowed-vars subj-cmp)
             triples
-            (dissoc node :id :idx))))
+            (->> (dissoc node :id :idx)
+                 ;; deterministic patterns for each pred
+                 (sort-by (comp str first))))))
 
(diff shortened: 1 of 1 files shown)


src/clj/fluree/db/query/fql/parse.cljc
```diff
@@ -683,7 +683,9 @@
                        :else            (where/match-iri id))]
     (reduce (partial parse-pred-cmp allowed-vars subj-cmp)
             triples
-            (dissoc node :id :idx))))
+            (->> (dissoc node :id :idx)
+                 ;; deterministic patterns for each pred
+                 (sort-by (comp str first))))))
```
cap10morgan, line 688: What is `(str (first node))` getting exactly? A string-ified key-value pair?

dpetran: A stringified key, actually. `sort-by` coerces the node map into a seq of entries `([k v] ...)`, so `first` gets the key and `str` stringifies it - I wanted to avoid any chance that keywords wouldn't be comparable with strings.

zonotope (approved):
🐇

PR #823, opened 2024-07-08
Feature/multiple reasoner sources2

JaceRockman:


bplatz:
This looks like it supplants https://github.com/fluree/db/pull/804 - is that so? If the case, we should close that PR.

src/clj/fluree/db/reasoner.cljc
```diff
@@ -218,24 +235,42 @@
             (recur r db**))
           db*)))))
 
+(defn deduplicate-raw-rules
```
bplatz, line 238: I think this could use a docstring to explain what it is doing. There is too much going on here and the code itself is not explaining it well.

In lieu of a docstring, perhaps if more of it was broken into multiple functions that were more descriptive.

src/clj/fluree/db/reasoner.cljc
```diff
@@ -218,24 +235,42 @@
...
+  [raw-rules]
+  (let [rule-ids (map first raw-rules)
+        duplicate-ids (filter #(< 1 (last %)) (frequencies rule-ids))]
+    (reduce (fn [rules [duplicate-id occurances]]
```
bplatz, line 242: It looks like this attempts to retain every rule that has identical names by modifying their names with an indexed value.

I guess the options here would be:
1) Throw an exception
2) "merge" behavior, where only one is retained (with the idea they should be identical if they have the identical IRI)
3) renaming duplicated names

You opted for #3 - is there a reason for that over the others?

src/clj/fluree/db/reasoner.cljc
```diff
@@ -218,24 +235,42 @@
...
+                    updated-rules-list
+                    (let [updated-rule [(str duplicate-id suffix) (last (first rules-to-update))]]
+                      (recur (dec suffix) (rest rules-to-update) (conj updated-rules-list updated-rule)))))))
+            raw-rules duplicate-ids)))
+
 (defn reason
-  [db methods graph-or-db {:keys [max-fuel reasoner-max]
-                           :or   {reasoner-max 10} :as _opts}]
+  [db methods
+   {:keys [rule-graphs rule-dbs]}
```
bplatz, line 255: If we do support a list of sources here, I don't think we should require special key syntax to distinguish them. By requiring this special map syntax, you are going to require our users to have to look up this special syntax likely every time they use the API - even if they are only passing in a single rule set (which is the most common case).

 Instead, you could easily call `db?` on each item and do the right thing without needing any special syntax at all.

src/clj/fluree/db/reasoner.cljc
```diff
@@ -218,24 +235,42 @@
...
+                (loop [suffix occurances
+                       rules-to-update (get grouped-rules true)
+                       updated-rules-list (get grouped-rules false)]
+                  (if (empty? rules-to-update)
+                    updated-rules-list
+                    (let [updated-rule [(str duplicate-id suffix) (last (first rules-to-update))]]
+                      (recur (dec suffix) (rest rules-to-update) (conj updated-rules-list updated-rule)))))))
+            raw-rules duplicate-ids)))
+
 (defn reason
```
bplatz, line 253: I think this adds complexity to the API in the wrong spot.

If working directly with the API, I think one could instead do:
```
(-> db 
    (reason :owl [{...some json-ld-rules}])
    (reason :datalog some-rules-db)
    (reason :owl some-other-rules-db))
```

Supporting multiple at once is more an issue for stateless transactions, and therefore any special syntax for it should be moved to `query-connection` (or successors) where we handle stateless interactions.

What still needs to be built in doing the above is to "merge in" each successive rule set - and whatever logic we decide on for that (per my other comments) should get implemented there.

src/clj/fluree/db/reasoner.cljc
JaceRockman: My thinking for option 3 was that if we are getting two rules from different sources with the same IRI, we would still want to preserve the functionality of both of those rules. If we wanted to retain one and remove the other, we would probably want the person submitting the query to have control over that which gets complicated I think. We also discussed this a few weeks ago and that was the solution that you leaned towards at the time.

bplatz: Apologies if I leaned towards that previously, sometimes it takes me getting into things deeper to have a more natural feel of what seems right.

I think I'd lean towards (2) now, and use merge semantics (left to right). This would allow the flexibility of overwriting rules. I could see this if global governance had a rule set that must be followed, regardless of if a local db owner tried to overwrite the rule.

While the rules still remain an 'or' condition, once we had a rule designated as "important" (similar to CSS's important!) it would allow and conditions.

The combination of merge semantics + important would allow global governance to guarantee a certain set of rules are always followed. Without it, I think there could be possibilities that folks could get around rules.

Also, with the suggestion below of:
```
(-> db 
    (reason :owl [{...some json-ld-rules}])
    (reason :datalog some-rules-db)
    (reason :owl some-other-rules-db))
```

It allows the flexibility of holding a db, applying some query/tx on it, then applying another rule set and applying another rule set. Supporting a list here likely precludes changing rules on an existing db, it would have to be a "reset"... whereas adding rules one at a time allows more flexibility.

JaceRockman: In option you say that we should be doing merge behavior under the assumption that identical ids should have identical rules, but if we're making that assumption, then we wouldn't need to make any changes here because one of the rules already gets overridden in that case. This is moreso meant to solve the problem of identical ids with distinct rules. In that case, should we be throwing an exception?

I also think I'm not fully understanding how an `important` tag enables `and` conditions. If there are two rules applied as an `or` condition, then for a given subject, it will be returned if either one of those rules matches it. For an `and` condition, a subject would only be returned if both rules match it, but if we have two rules and one of them is getting overridden because the other rule is marked as `important`, then it's not evaluating the other rule at any point.

And what should we do if neither rule is marked as `important` or both rules are marked as `important`?

If something is marked as `important` I don't think we need merge semantics to ensure that those are always followed. I would just have to add a check in the deduplicate function that if any of the rules with duplicate IRIs are marked as `important` then we ignore the rest of those duplicated rules.

If we are adding rules one at a time, I think that might cause issues where the order that we apply the rules becomes much more important. If we have a rule that adds `ex:uncle` to an entity under some condition and another rule that adds `ex:aunt` to an entity if they are married to an `ex:uncle`, then the order that those rules are applied will determine whether or not `ex:aunt` gets added to that entity.

bplatz: Merge behavior with two identical rule ids could:
a) overwrite without warning
b) overwrite with warning
c) throw

To me (b) feels like the right answer, not that it needs to be this way and do everything in a single PR.

I think we should strive for predictable, simple behavior and merge feels right (perhaps not perfect). I think it can provide a functional feature as well, which is if there were a common set of rules in some ledger, I could "deactivate/replace" some of them by merging another rule set with just those ids.

As for `important` (or a different semantic might be better), the idea is to allow AND conditions. Right now only OR conditions are supported. By marking a rule as `important` it means it *must* pass. If there are 2+ `important` rules for the same Flake, they must all pass (AND). In that case none of the other rules matter which are not `important` and they can be effectively ignored. Even though it might be ignored for a specific Flake, it might be relevant for other flakes as rules can have narrow or broad scopes.

Again, not all for the same PR - just explaining the breadth of what I think the reasoning feature can do, as it helps inform the right behavior of what to do when we have multiple reasoning rule sets which is the focus here.

dpetran: How do you warn? If it's just log messages most users will never see them. I would lean towards (a) with documentation or (c), I think.

bplatz: I think with (c) you'd loose a valuable feature. I think (a) is definitely an option. I still prefer (b) with log messages.

Your assumption that they won't see the log messages assumes this is coming from someone running a production server. I think folks who are developing and merging reasoning rules are likely be running this as a developer locally getting the behavior they are looking for. If for personal analytics, it will never run as a production server - or if a production environment that logic they develop locally will be part of their app. If they want to turn off the logging once in production they can easily do so with logback.xml file.

JaceRockman: Ok, I misunderstood your comment about the `important` tag to be saying that that's how you think conflicting rule ids should be resolved. You're just saying that you think the merge behavior makes more sense with the eventual implementation of the `important` tag. I have more questions about the implementation of that, but it can wait until we are actually thinking of working on it.

So for this PR then, how do we know which source of rules to prioritize for overriding then? You mentioned in one of your other comments that the rule sources would be better as just a list instead of using the `:rule-graphs` and `:rule-dbs` map syntax, are you thinking it just takes that list and merges from left to right regardless of what type of source it is coming from?

bplatz: I think it will require keeping a map of rules on the db with ids as keys (or similar) - for the primary purpose of merging in new rules if added and then applying the result of that merge as the working rule set.

JaceRockman: Maybe I'm not understanding, but are you saying that the ids of the rules in the provided rule sources should be transacted to the db? I'm not sure what that accomplishes over just getting the raw rules in `reasoner/reason` and then merging those maps together?

bplatz: No, they shouldn't be transacted - but they need to be retained somewhere... e.g. `(:reasoning-rules db)`... so we are hanging them on the db object as a new key-val pair. This way if more rules are added later we can perform the merge and apply the result as the new set of rules.

The other thing that needs to be considered is equality checks on the dbs that are reasoned, but that could be addressed separately.

src/clj/fluree/db/flake/reasoner.cljc
```diff
@@ -163,28 +165,52 @@
...
+  [methods inserts dbs]
+  (let [clause-chan (async/to-chan! dbs)
+        out-chan (async/chan)]
+    (doall
+     (for [method methods]
+       (async/pipeline-async
+        1
+        out-chan
+        (fn [db out-ch]
+          (async/pipe (go-try
```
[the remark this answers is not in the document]
JaceRockman: I'm trying to loop over the `clause-chan` with something like this:
```
(defn rules-from-dbs
  [methods inserts dbs]
  (let [clause-chan (async/to-chan! dbs)
        out-chan (async/chan)]
    (doall
     (for [method methods]
       (clojure.core.async/go-loop [from-ch clause-chan]
         (let [db (<? from-ch)]
           (if (nil? db)
             out-chan
             (do (async/pipe (go-try
                               (as-> db $
                                 (<? (resolve/rules-from-db $ method))
                                 (rules-from-graph method inserts $)))
                             out-chan)
                 (recur from-ch)))))))))
```
Maybe I misunderstood what you were saying, but I'm just replacing the `pipeline-async` call with a `go-loop`. With `go-loop` I can process elements from the `from-chan` and `pipe` them onto the `out-chan`, but this ends up requiring me to call `<?` twice on the results since `go-loop` creates a channel that contains the `out-chan`. So I think I'm doing something wrong, but I'm not sure what.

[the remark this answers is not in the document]
zonotope: `<?` throws any exceptions that come off of a channel immediately. When an exception is thrown inside of a go block, it is silently swallowed and the channel returned by the go block closes without returning any more results. Therefore, you need to explicitly handle any errors that could possibly be thrown with a `try`/`catch` block or something similar. 

You should only use the higher order pipeline functions when you are certain that none of the code inside the pipeline can throw for this reason, or you are certain that you have handled any possible errors that could be thrown.

You need to account for the possiblity that `<?` is an exception, and do the right thing in each case it is. If it's impossible for it to be an exception, then you shouldn't use `<?`; you should use `<!` instead.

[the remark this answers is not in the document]
JaceRockman: This sounds like it is all assuming that I would be going the route of using the higher level pipeline functions and explicitly handling any exceptions. You were originally recommending that I use a loop though, which is what I was struggling with figuring out. Is there one of these solutions that is better than the other?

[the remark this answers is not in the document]
JaceRockman: Right, but your first comment on this thread was about `<?` not working with the `pipeline-` functions, but now it sounds like you are just talking about `<?` needing explicit error handling regardless of whether I'm using the `pipeline-` functions. So should I just keep the `pipeline-` functions and do explicit error handling since it sounds like I will need to do that either way.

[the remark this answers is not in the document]
zonotope: Here's what I said:
> go-try/<? does not work well with the higher order core.async pipeline functions like pipeline-async and pipe because any errors that are thrown won't be handled properly and instead be propagated downstream.

`go-try` takes any exception that is thrown inside it's block and puts it on its output channel. `<?` takes any exception from the channel it's pulling from and throws it. That behavior is not compatible with the pipeline functions because the whole point of using them is so you can process collections as a single unit. That no longer works when there could be exception objects mixed in to those collections. 

If you are going to use the pipeline functions, then you need to make sure that none of the code in the pipeline functions can throw exceptions (which `<?` does), and you need to make sure that none of the code upstream of those functions will put exceptions on their input channel, which `go-try` does.

src/clj/fluree/db/flake/reasoner.cljc
```diff
@@ -163,28 +165,45 @@
     (log/debug "Reasoner - source OWL rules: " graph)
     (owl-datalog/owl->datalog inserts graph)))
 
+(defn rules-from-dbs
+  [methods inserts dbs]
+  (let [rules (atom [])]
```
dpetran, line 170: I don't think you need an atom here, you can accumulate the data you want in the loops - here's a rough example:
```clojure
(go-try
  (loop [[method & r-methods] methods
         rules []]
    (if method
      (loop [[db & r-dbs] dbs
             rules* rules]
        (if db
          (let [db-rules (<? (resolve/rules-from-db db method))]
            (recur r-dbs (into rules* (rules-from-graph method inserts db-rules))))
          (into rules rules*)))
      rules)))
```

JaceRockman: This feels more complicated than the code I had before wrapping a `for` loop in `doall` to remove the laziness. I think that also reads better than this. Is there a reason that isn't ideal?

dpetran: Can you elaborate on what feels more complicated? . This imperative-style mutation-with-a-loop is pretty unidiomatic in Clojure. Mutation is inherently complex and if we're not doing it for performance reasons (and I believe in this case it's actually slower, though probably not critically so) it's good practice to avoid it. 

I also think it's simpler  - it's less code, the structure is pretty clear, and it should be plenty fast.

JaceRockman: I'm thinking the for-loop would look something like this:

JaceRockman: Ok, that explains some of the issues I've been running into trying to get it to work. All this async behavior seems really finicky and unintuitive on when it works and when it doesn't.

dpetran: I find it much easier to read recursive functions if the `recur` doesn't contain multi-line expressions. Can you restructure this along the lines of what I sketched out earlier?

src/clj/fluree/db/flake/reasoner.cljc
```diff
@@ -163,28 +165,47 @@
...
-                                      (throw e))))]
-      (loop [[method & r] methods
-             rules []]
-        (if method
-          (let [rules-graph* (or supplied-rules*
-                                 (<? (resolve/rules-from-db rules-db method)))
-                rules*       (rules-from-graph method inserts rules-graph*)]
-            (recur r (into rules rules*)))
-          rules)))))
+    (let [rule-graphs           (filter #(and (map? %) (not (queryable? %))) rule-sources)
```
dpetran, line 191: It looks like you have `fluree.db.query.exec` required twice. I prefer not `:refer`ing functions so it's easy to tell at a glance what dependencies a function has just by looking at the ns, so I think just using `exec/queryable` would be fine here. About the only thing I ever `:refer` is macros like `go-try` and `try*`/`catch*`.

src/clj/fluree/db/flake/reasoner.cljc
```diff
@@ -163,28 +165,47 @@
     (log/debug "Reasoner - source OWL rules: " graph)
     (owl-datalog/owl->datalog inserts graph)))
 
+(defn rules-from-dbs
+  [methods inserts dbs]
+  (go-try
```
zonotope, line 170: This function has way too much nesting. You should break it up into multiple functions that compute the steps you need and put them together here.

src/clj/fluree/db/flake/reasoner.cljc
```diff
@@ -163,28 +165,47 @@
...
+                                          (mapcat (fn [parsed-rules-graph]
+                                                    (rules-from-graph method inserts parsed-rules-graph))
+                                                  parsed-rule-graphs))
+                                        methods)
+          rule-dbs              (filter #(or (string? %) (queryable? %)) rule-sources)
+          all-rule-dbs          (if (or (nil? rule-dbs) (empty? rule-dbs))
+                                  [db]
+                                  (conj rule-dbs db))
+          all-rules-from-dbs    (<? (rules-from-dbs methods inserts all-rule-dbs))
+          all-rules             (concat all-rules-from-graphs all-rules-from-dbs)]
```
zonotope, line 207: you should return the concatenation directly instead of making a new binding that you just return immediately.

src/clj/fluree/db/flake/reasoner.cljc
```diff
@@ -235,21 +256,47 @@
...
+  (let [rule-ids (map first raw-rules)]
+    (->> rule-ids
+         frequencies
+         (filter #(< 1 (val %))))))
+
+(defn deduplicate-raw-rules
+  "Given a list of reasoning rules, identifies rules with duplicate ids and renames them using
+  indexes."
+  [raw-rules]
+  (let [duplicate-ids (find-duplicate-ids raw-rules)]
```
zonotope, line 270: This binding should have a different name; these are not duplicate ids but a map from duplicate ids to their number of occurrences.

test/fluree/db/reasoner/owl2rl_test.clj
```diff
@@ -538,8 +551,13 @@
                                                          "rdfs" "http://www.w3.org/2000/01/rdf-schema#"}
                                         "@id"           "ex:child"
                                         "@type"         ["owl:ObjectProperty"]
-                                        "owl:inverseOf" {"@id" "ex:parent"}}])]
+                                        "owl:inverseOf" {"@id" "ex:parents"}}])]
 
+     (println "JMR" (fluree/query db-reasoned
```
dpetran, line 556: We can remove this `println` I think.

test/fluree/db/reasoner/owl2rl_test.clj
```diff
@@ -288,7 +301,7 @@
                                        "owl" "http://www.w3.org/2002/07/owl#"}
                            "@id"      "ex:mother"
                            "@type"    ["owl:ObjectProperty" "owl:FunctionalProperty"]}])]
-
+
```
dpetran, line 304: I try to keep an eye out for whitespace-only changes like this and avoid staging them - it just adds noise to the git history. Not a big deal, but something to keep in mind in the future.

test/fluree/db/reasoner/owl2rl_test.clj
```diff
@@ -200,40 +213,40 @@
...
                                             "owl"  "http://www.w3.org/2002/07/owl#"
                                             "rdfs" "http://www.w3.org/2000/01/rdf-schema#"}
-                              "@id"        "ex:parent"
+                              "@id"        "ex:parents"
                               "@type"      ["owl:ObjectProperty"]
-                              "rdfs:range" [{"@id" "ex:Person"} {"@id" "ex:Parent"}]}])]
+                              "rdfs:range" [{"@id" "ex:Person"} {"@id" "ex:parents"}]}])]
 
-          (is (= (list "ex:Parent" "ex:Person")
+          (is (= (list "ex:Person" "ex:parents")
```
dpetran, line 234: This `list` call is extraneous here 
```clojure
(= (list 1 2 3)
   [1 2 3]
   '(1 2 3))
;;=> true
```

dpetran: Is the result order not stable? My thought is that it should be, but I'm not that familiar with it.

JaceRockman: I'm not sure, I think this used `list` before my changes. A lot of the existing tests are formatted like this.

dpetran (approved):
Looks good to me. @zonotope any other feedback, or do you reckon this is ready to merge?

src/clj/fluree/db/flake/reasoner.cljc
```diff
@@ -235,21 +258,47 @@
...
           db*)))))
 
+(defn find-duplicate-ids
+  [raw-rules]
+  (let [rule-ids (map first raw-rules)]
+    (->> rule-ids
+         frequencies
+         (filter #(< 1 (val %))))))
+
+(defn deduplicate-raw-rules
```
zonotope, line 268: I think this algorithm is needlessly wasteful. You go over the rules map multiple times to first find the frequencies, and then multiple `group-by` calls for each duplicate rule. I think it would be much more efficient to group the rules by the rule id once, and then filter the map returned by `group-by` that have more than one rule in the value of each rule entry.

src/clj/fluree/db/flake/reasoner.cljc
```diff
@@ -235,21 +258,47 @@
...
-                              resolve/rules->graph
-                              add-rule-dependencies)
-          db**            (if-let [inserts* @inserts]
-                            (<? (process-inserts db* fuel-tracker inserts*))
-                            db*)]
+          raw-rules          (<? (all-rules methods db* inserts rule-sources))
+          _                  (log/debug "Reasoner - extracted rules: " raw-rules)
+          duplicate-id-freqs (find-duplicate-ids raw-rules)
+          deduplicated-rules (when (not (empty? duplicate-id-freqs))
+                               (log/error "Duplicate ids detected. Some rules will be overwritten:" (apply str (map first duplicate-id-freqs))))
```
zonotope, line 295: This is not an error and should not be logged as such. At most this should be logged at the warn level, but I think it could reasonably be logged at info or debug.

src/clj/fluree/db/query/api.cljc
```diff
@@ -146,16 +157,31 @@
...
-            ledger  (<? (jld-ledger/load conn address))
-            db      (ledger/-db ledger)
-            t*      (or explicit-t t)]
-        (<? (restrict-db db t* opts)))
+            address      (<? (nameservice/primary-address conn alias nil))
+            ledger       (<? (jld-ledger/load conn address))
+            db           (ledger/-db ledger)
+            t*           (or explicit-t t)
+            rule-sources (<? (load-aliased-rule-dbs conn (:rule-sources opts)))
+            opts*        (assoc opts :rule-sources rule-sources)]
```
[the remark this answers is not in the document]
JaceRockman: Because `rule-sources` is already part of the opts that are being provided, it feels odd to also be providing that as an additional argument. I think I can just do the `load-aliased-rule-dbs` inside `restrict-db` and that might be cleaner.

src/clj/fluree/db/flake/reasoner.cljc
JaceRockman: This function actually isn't getting used anymore, after talking with Brian we decided to instead go the route of overwriting duplicate ids and throwing a warning: https://github.com/fluree/db/pull/823#discussion_r1671003352. So I'll just get rid of this function.

zonotope (approved):

PR #824, opened 2024-07-10
Fix list retraction

cap10morgan:
The retraction code was conceptually duplicated from the assertion code, but implemented pretty differently (when 99% of the time the only necessary difference was the `false` arg to `flake/create`). This was causing list retractions to fail b/c lists weren't handled in the retraction code.

This DRY's up the assert/retract code and fixes the list retraction issue. Ideally I would add a test in here for the list retraction issue, and I will attempt to do that. But I don't want to spend too much time on it. I do have an integration test in another project that went from failing to passing with these changes.

### DIFF
--- a/src/clj/fluree/db/db/json_ld.cljc
+++ b/src/clj/fluree/db/db/json_ld.cljc
@@ -34,8 +34,11 @@
             [fluree.db.serde.json :as serde-json]
             [fluree.db.time-travel :refer [TimeTravel]]
             [fluree.db.util.async :refer [<? go-try]]
-            [fluree.db.util.core :as util :refer [get-first get-first-value
-                                                  get-first-id vswap!]]
+            [fluree.db.util.core :as util
+             #?@(:clj  [:refer [get-first get-first-value get-first-id vswap!
+                                try* catch*]]
+                 :cljs [:refer [get-first get-first-value get-first-id vswap!]
+                        :refer-macros [try* catch*]])]
             [fluree.db.util.log :as log]
             [fluree.json-ld :as json-ld])
   #?(:clj (:import (java.io Writer))))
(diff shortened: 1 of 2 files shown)


src/clj/fluree/db/db/json_ld.cljc
```diff
@@ -474,60 +477,84 @@
...
-                          (not (-> node-entry key keyword?))))
-                (mapcat (fn [[prop value]]
-                          (let [pid (iri/encode-iri db prop)]
-                            (assert-property db sid pid t value)))))
+          (comp (remove #(-> % key keyword?))
+                (mapcat
+                 (fn [[prop value]]
+                   (let [pid (if assert?
+                               (iri/encode-iri db prop)
+                               (try*
```
dpetran, line 543: Instead of doing this messy exception handling here, could we just let the `nil` from `iri/encode-iri` trickle down to `value-map->flake` and throw if `pid` or `sid` are `nil` when we actually are about to call `flake/create`?

Also, I think we should update the error value on the exception, this has nothing to do with committing and I think that can only cause confusion (though I realize you're not the one who wrote the error initially).

cap10morgan: > Instead of doing this messy exception handling here, could we just let the `nil` from `iri/encode-iri` trickle down to `value-map->flake` and throw if `pid` or `sid` are `nil` when we actually are about to call `flake/create`?

I like keeping the exception handling more explicit and closely and directly connected to the problem it's dealing with. I hate it when I run into code handling "maybe this is `nil`?" and I then have to go figure out why / when it could be nil. This doesn't seem all that messy to me compared to that.

> Also, I think we should update the error value on the exception, this has nothing to do with committing and I think that can only cause confusion (though I realize you're not the one who wrote the error initially).

These fns are all in support of `merge-commit`, so it seems commit-related to me, unless I'm misunderstanding?

dpetran: Fair enough, regarding letting nils by. 

Do we need `encode-existing-iri` though? We could just use an `or` instead of `try*` for the same effect in fewer lines.

As for whether this is commit related, this is all actually downstream of `stage`, and `merge-commit` just happens to reuse that machinery. If there's any problem it'll only surface during a call to `stage`. Commits should only have known-good data.

cap10morgan: > Fair enough, regarding letting nils by.
> 
> Do we need `encode-existing-iri` though? We could just use an `or` instead of `try*` for the same effect in fewer lines.

I like giving things like that intention-revealing names. It wasn't immediately obvious to me at first glance why this was different for retracts until I thought about it for a second. But I'm not super wedded to it in this case, so if you don't think it's worth it I can go back to an `or`.

> As for whether this is commit related, this is all actually downstream of `stage`, and `merge-commit` just happens to reuse that machinery. If there's any problem it'll only surface during a call to `stage`. Commits should only have known-good data.

OK I can tweak the error message.

cap10morgan: Oh, wait. I now realize I don't think I follow your suggested change to the error. Can you elaborate?

dpetran: I just found the pattern of throwing an exception then immediately catching it and throwing a different exception to be a little heavy for this case, the exception should provide enough context about why it's necessary.

cap10morgan: Oh, I see. I did that because in the `encode-existing-iri` context it shouldn't assume things that the original error messages contained (see the deleted retract code). But I didn't want to lose information that those error messages had in case anyone was relying on them. So I throw an error from `encode-existing-iri` only containing what it knows went wrong, but then added in the additional context from the original errors in the calling catches.

dpetran: Hold on, this isn't downstream of `stage` at all - I was wrong. I thought they shared the same mechanism but it looks like it's different. So it looks like the error will only show up while attempting to load a ledger.

Here's kind of what I'm thinking:
```
(if assert?
  (iri/encode-iri db prop)
  (or (iri/encode-iri db prop)
        (throw (ex-info "Cannot retract IRI with unknown namespace." 
                                    {:status 400 :error :db/invalid-retraction :iri prop}))))
```

cap10morgan:
I added a test

dpetran (approved):
:fleur_de_lis:

test/fluree/db/transact/retraction_test.clj
```diff
@@ -41,4 +41,36 @@
...
+          db-after-retract @(fluree/stage
+                             db
+                             {"@context" context
+                              "delete"   {"id"        "ex:list-test"
+                                          "ex:items1" "?items1"
+                                          "ex:items2" "?items2"}
+                              "where"    {"id"        "ex:list-test"
+                                          "ex:items1" "?items1"
+                                          "ex:items2" "?items2"}})
+          after-retract    @(fluree/query db-after-retract q1)]
```
dpetran, line 71: I think your `after-retract` needs to be against a db that's been committed and re-loaded in order to exercise the `merge-commit` call stack.

PR #825, opened 2024-07-11
Feature/credential apis

bplatz:
This PR creates the remaining credential/signature specific APIs. Now queries, transactions, etc. that are wrapped in signatures have an independent API.

The primary motivation for this work right now is getting back a "zero-trust" setting for fluree/server like we had in v2 and this supports that being done. In the zero-trust (formerly called open-api: false), all interactions must happen with signatures. I also think those APIs make more sense separate as they not only have a different input format, but are quite intentional - you should know if you need it or not. I think it also helps ensure you don't accidentally do the wrong thing, as security here is of high concern.

In addition, there was some intersecting work, the biggest change is that `f:txn` (source transaction) *only* gets written to disk and included in the commit if it was originally wrapped in a signature. Storing every transaction is expensive, and it's purpose is for replayable cryptographic proof - when a ledger is being used without signatures it does nothing for that use case.

There were a number of tests that used test-utils/transact, which I assume was created before we had the fluree/transact! API - so those tests now use our external API and the helper util is removed.

### DIFF
--- a/src/clj/fluree/db/api/transact.cljc
+++ b/src/clj/fluree/db/api/transact.cljc
@@ -15,10 +15,11 @@
             [fluree.db.ledger :as ledger]))
 
 (defn parse-opts
-  [parsed-opts opts]
-  (reduce (fn [opts* [k v]] (assoc opts* (keyword k) v))
-          parsed-opts
-          opts))
+  [expanded-txn opts txn-context]
+  (-> (util/get-first-value expanded-txn const/iri-opts)
+      (merge opts) ;; override opts 'did', 'raw-txn' with credential's if present
+      (util/keywordize-keys)
+      (assoc :context txn-context)))
 
 (defn stage
   [db txn opts]
@@ -26,10 +27,7 @@
    (let [txn-context (or (ctx-util/txn-context txn)
                          (:context opts))
          expanded    (json-ld/expand (ctx-util/use-fluree-context txn))
-         parsed-opts (-> (util/get-first-value expanded const/iri-opts) ;; get opts out of transaction
-                         (merge opts) ;; merge with upstream supplied opts (e.g. did, raw-txn
-                         (assoc :context txn-context)
-                         (util/keywordize-keys))
+         parsed-opts (parse-opts expanded opts txn-context)
          track-fuel? (or (:maxFuel parsed-opts)
                          (:meta parsed-opts))
          parsed-txn  (q-parse/parse-txn expanded txn-context)
@@ -65,10 +63,7 @@
                                         {:status 400 :error :db/invalid-transaction})))
           address     (<? (nameservice/primary-address conn ledger-id* nil))
 
-          parsed-opts (-> (util/get-first-value expanded const/iri-opts)
-                          (merge opts)
-                          (util/keywordize-keys)
-                          (assoc :context txn-context))]
+          parsed-opts (parse-opts expanded opts txn-context)]
       (if-not (<? (nameservice/exists? conn address))
         (throw (ex-info "Ledger does not exist" {:ledger address}))
         (let [ledger   (<? (jld-ledger/load conn address))
@@ -104,10 +99,7 @@
                         (throw (ex-info "Invalid transaction, missing required key: ledger."
                                         {:status 400 :error :db/invalid-transaction})))
           address     (<? (nameservice/primary-address conn ledger-id nil))
-          parsed-opts (-> (util/get-first-value expanded const/iri-opts)
-                          (merge opts) ;; override opts 'did', 'raw-txn' with credential's if present
-                          (util/keywordize-keys)
-                          (assoc :context txn-context))]
+          parsed-opts (parse-opts expanded opts txn-context)]
       (if (<? (nameservice/exists? conn address))
         (throw (ex-info (str "Ledger " ledger-id " already exists")
                         {:status 409 :error :db/ledger-exists}))
--- a/src/clj/fluree/db/api/transact.cljc
+++ b/src/clj/fluree/db/api/transact.cljc
@@ -21,89 +21,103 @@
           opts))
 
 (defn stage
-  [db txn {:keys [raw-txn] :as opts}]
+  [db txn opts]
   (go-try
-    (let [{txn* :subject did :did} (or (<? (cred/verify txn))
-                                       {:subject txn})
-          txn-context              (or (:context opts)
-                                       (ctx-util/txn-context txn*))
+   (let [txn-context (or (ctx-util/txn-context txn)
+                         (:context opts))
+         expanded    (json-ld/expand (ctx-util/use-fluree-context txn))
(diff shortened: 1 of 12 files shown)


src/clj/fluree/db/json_ld/commit_data.cljc
```diff
@@ -557,15 +563,18 @@
                              (issuer-flakes db t commit-sid issuer-iri))
         message-flakes     (when message
                              (message-flakes t commit-sid message))
+        txn-flake          (when txn
```
zonotope, line 566: I think shadowing function names with local let bindings can be a source of both bugs and confusion. Since there's already a `txn-flake` function definition in scope here could we either change the name of that function to something like `create-txn-flake`, or change the binding here to `txn-f` or something?

src/clj/fluree/db/json_ld/credential.cljc
```diff
@@ -115,8 +115,22 @@
         auth-did                 (did/auth-id->did id)]
     {:subject (json/parse payload false) :did auth-did}))
 
+(defn is-JWS?
```
zonotope, line 118: I think just `jws?` would be a more idiomatic function name.

src/clj/fluree/db/query/api.cljc
```diff
@@ -64,24 +64,25 @@
...
+  ([ds query {:keys [did issuer] :as _opts}]
+   (go-try
+    ;; TODO - verify if both 'did' and 'issuer' opts are still needed upstream
+    (let [{:keys [t opts] :as query*} (update query :opts sanitize-query-options (or did issuer))
 
           ;; TODO: extracting query context here for policy only to do it later
           ;; while parsing the query. We need to consolidate both policy and
           ;; query parsing while cleaning up the query api call stack.
           q-ctx    (ctx-util/extract query*)
+          ;; TODO - remove restrict-db from here, restriction should happen upstream if needed
```
zonotope, line 77: 👍🏾

zonotope: Also, once we make the change to restricting the db upstream, we should remove the `opts` argument here and accept `did` and `issuer` directly. I think open ended options maps should be reserved for user facing api functions, and the options maps should be parsed, validated, and destructured immediately at the api surface instead of passed directly to internal functions, and the signatures of non-user facing functions should be explicit.

src/clj/fluree/db/query/api.cljc
```diff
@@ -207,11 +208,9 @@
         (dataset db-map defaults)))))
 
 (defn query-connection-fql
-  [conn query]
+  [conn query {:keys [did] :as opts}]
```
zonotope, line 211: I think we should pass `did` directly here instead of introducing another open-ended options argument as this isn't a user-facing function

src/clj/fluree/db/json_ld/commit_data.cljc
```diff
@@ -524,6 +522,14 @@
   [t commit-sid message]
   [(flake/create commit-sid const/$_commit:message message const/$xsd:string t true nil)])
 
+;; TODO - txn should really be an IRI, not a string
```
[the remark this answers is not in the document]
bplatz: We persist it to disk based on its unique hash-based id (content-addressed). So the @id should be the unique identifier, and if we follow the pattern we already have established for commits and data files, we would also include "f:address" as the physical storage location(s).

On disk would be JSON-LD looking something like:
```
{"@id": "fluree:txn:sha256:...." ;; eg IPFS IRI address, local file, s3
 "@type": ["f:Transaction"]
 "f:address" "...some physical storage location addres ..."
 "f:tx-raw" ".... stringified raw data ..."}  ;; this could be JSON, JWS, or even some byte data for different crypto
```
It could also include other metadata with it... e.g. for a different type of crypto it might include a "f:verificationProof", or whatever... it just just JSON-LD and infinitely flexible. Technically it doesn't have to include everything above either, as some could be inferred (if we were super storage-concerned).

So for the purposed of the commit file, it would likely only have the {"@id": "...", "f:address": "..."} embedded to minimize space.... any other metadata (if any besides the raw txn) would have to go to disk to retrieve.

This is a bigger change of course, which is why I left this TODO.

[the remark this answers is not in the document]
zonotope: This makes sense. 👍🏾

src/clj/fluree/db/json_ld/credential.cljc
```diff
@@ -115,8 +115,22 @@
...
+  [signed-transaction]
+  (string? signed-transaction))
```
dpetran, line 120: A problem Jace ran across is SPARQL queries are just strings, so we can't tell the difference between a SPARQL query and a JWS with just the `string?` predicate. 

This regex should work: `#"(^[A-Za-z0-9-_]*\.[A-Za-z0-9-_]*\.[A-Za-z0-9-_]*$)"`.

src/clj/fluree/db/json_ld/commit_data.cljc
bplatz: Thanks for noticing... corrected in 653f8c9c64ed88b0348de53e211d7505a403830a

src/clj/fluree/db/json_ld/credential.cljc
bplatz: OK. Resolved with 78f1732d6950d41a79de9a561adf640f159b21df

src/clj/fluree/db/json_ld/credential.cljc
bplatz: Ahh thank you! I actually ran into this issue with fluree/server and didn't know why and haven't had time to dig in. That saved me a bunch of time I suspect.

bplatz: I've really been thinking SPARQL should have its own API... but I am also cognizant that the API surface is growing.

src/clj/fluree/db/query/api.cljc
bplatz: Ok! Done in eaecd865cbebedee14a1cb314c4a9c3568964e16

src/clj/fluree/db/api/transact.cljc
```diff
@@ -21,89 +21,103 @@
...
+  [db txn opts]
   (go-try
-    (let [{txn* :subject did :did} (or (<? (cred/verify txn))
-                                       {:subject txn})
-          txn-context              (or (:context opts)
-                                       (ctx-util/txn-context txn*))
+   (let [txn-context (or (ctx-util/txn-context txn)
+                         (:context opts))
+         expanded    (json-ld/expand (ctx-util/use-fluree-context txn))
+         parsed-opts (-> (util/get-first-value expanded const/iri-opts) ;; get opts out of transaction
```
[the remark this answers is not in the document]
bplatz: Addressed in 20eaaa098effae890431b716bb83d8af12e9b7cb

zonotope (approved):
🌙

PR #826, opened 2024-07-13
Refactor/policy identity

bplatz:
This removes the old policy 'roles' fns. Roles are no longer part of policy.

`fluree/query-connection` (used by fluree/server) was broken because it was still relying on this. Removed remaining traces.

This also defaults identity-policy execution to default-allow? true.

This is based on PR #825


zonotope (approved):
🪪

PR #827, opened 2024-07-15
Feature/history api updates

dpetran:
Fixes https://github.com/fluree/core/issues/110

This PR is meant to allow users to be receive more granular information from the history api. It teases apart some different categories of data:
- `:commit` - the commit metadata that wraps the actual data: `f:address` `f:alias` `f:author` `f:branch` `f:data` `f:previous` `f:time` `f:v` `id` `f:annotation` and `f:txn`.
- `:data` - the asserts and retracts that represent the facts changed: `f:assert` `f:retract`
- `:txn` - the raw transaction string (or `nil` if there was no signed transaction associated with the commit.
 
Each of these history keys can be composed with the existing keys `:history` and `:t`, but overrides the `:commit-details` and is meant to be a replacement for that option.

### DIFF
--- a/src/clj/fluree/db/async_db.cljc
+++ b/src/clj/fluree/db/async_db.cljc
@@ -146,13 +146,13 @@
       (let [db (<? db-chan)]
         (<? (history/-history db context from-t to-t commit-details? error-ch history-q)))))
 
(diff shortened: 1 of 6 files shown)


zonotope (approved):
👠

PR #828, opened 2024-07-15
Add load-or-notify API fn

cap10morgan:
This is in support of nexus-query-server and cloud-transactor where transactions happen in another process and nexus-query-server might have the ledger already loaded & cached but at an older t value. This fn allows it to load the ledger if it hasn't already or just update its cache to the new t if it has.

This also inverts the boolean logic of `register-ledger` so that it returns a `cached?` value instead of `not-cached?` which was confusing to reason about.

### DIFF
--- a/src/clj/fluree/db/api.cljc
+++ b/src/clj/fluree/db/api.cljc
@@ -18,7 +18,7 @@
             [fluree.db.util.log :as log]
             [fluree.db.query.range :as query-range]
             [fluree.db.nameservice.core :as nameservice]
-            [fluree.db.connection :refer [notify-ledger]]
+            [fluree.db.connection :as conn]
             [fluree.db.json-ld.credential :as cred]
             [fluree.db.reasoner :as reasoner]
             [fluree.db.flake :as flake]
@@ -172,11 +172,28 @@
   [conn commit-map]
   (promise-wrap
     (if (map? commit-map)
-      (notify-ledger conn commit-map)
+      (conn/notify-ledger conn commit-map)
       (go
         (ex-info (str "Invalid commit map, perhaps it is JSON that needs to be parsed first?: " commit-map)
                  {:status 400 :error :db/invalid-commit-map})))))
 
+(defn load-or-notify
+  "Loads the ledger specified by alias if it isn't already, or notifies it of
+  the new commit in commit-map if it is already loaded. Returns a promise with
+  the loaded / updated ledger."
+  [conn alias commit-map]
+  (promise-wrap
+   (go
+     (log/trace "load-or-notify:" alias)
+     (<! (let [[cached? ledger-chan] (conn/register-ledger conn alias)]
+           (if cached?
+             (do
+               (log/trace "load-or-notify notifying cached ledger"
+                          (async/poll! ledger-chan))
+               (conn/notify-ledger conn commit-map))
+             (let [ledger-addr (<! (alias->address conn alias))]
+               (log/trace "load-or-notify loading ledger" alias)
+               (jld-ledger/load* conn ledger-chan ledger-addr))))))))
 
 (defn stage
   "Performs a transaction and queues change if valid (does not commit)"
--- a/src/clj/fluree/db/connection.cljc
+++ b/src/clj/fluree/db/connection.cljc
@@ -52,23 +52,26 @@
   in-memory on the conn.
 
   Returns a two-tuple of
-  [not-cached? promise-chan]
+  [cached? promise-chan]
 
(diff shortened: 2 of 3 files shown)


cap10morgan:
* **#832** <a href="https://app.graphite.dev/github/pr/fluree/db/832?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#831** <a href="https://app.graphite.dev/github/pr/fluree/db/831?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#830** <a href="https://app.graphite.dev/github/pr/fluree/db/830?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#829** <a href="https://app.graphite.dev/github/pr/fluree/db/829?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#828** <a href="https://app.graphite.dev/github/pr/fluree/db/828?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @cap10morgan and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

dpetran (approved):
:notes:

src/clj/fluree/db/api.cljc
```diff
@@ -172,11 +172,28 @@
...
-      (notify-ledger conn commit-map)
+      (conn/notify-ledger conn commit-map)
       (go
         (ex-info (str "Invalid commit map, perhaps it is JSON that needs to be parsed first?: " commit-map)
                  {:status 400 :error :db/invalid-commit-map})))))
 
+(defn load-or-notify
+  "Loads the ledger specified by alias if it isn't already, or notifies it of
+  the new commit in commit-map if it is already loaded. Returns a promise with
+  the loaded / updated ledger."
```
bplatz, line 183: This may have a use case I'm not thinking of, but the reason there is separate `load` and `notify` APIs is that if you receive update notifications but don't have the ledger loaded, it should be ignored. If a ledger is needed for a query, `load` can be used and then `notify` will keep it updated automatically. Loading a ledger is expensive, and keeping it up to date is even more expensive - especially expensive on memory... work we want to avoid if a ledger is not being used on a server.

In the case of Nexus this is exacerbated until they have more specific pub/sub on a per ledger basis. (the 'subscribe' API, which is not yet available in v3, and likely needs to act a little different than v2's version). At scale a server could be getting thousands of notify messages of which almost all of them can be ignored.

Also not yet done is an `unload` API, which after a period of inactivity for a ledger should unload it completely - so that future notify messages can be ignored.

cap10morgan: The use case is so that nexus-query-server doesn't have to reach around the public db API to do this: https://github.com/fluree/nexus-query-server/blob/0c89d31d86c5a0d6331832b4160ff177ad0c1fe4/src/fluree/query_server/handlers/sns.clj#L50-L56

...when it receives a notification of a successful transaction from cloud-transactor. I don't have any kind of spec of what nexus-query-server actually needs to do here to say if this is necessary or not, but I did want to get it to use only public API fns from db.

bplatz: I don't know what calls that code upstream, but the design is intended to be able to avoid this - as if this is performed for every commit it will be unnecessarily taxing on the servers. I don't know what calls this upstream and why - but if there was a reason the design pattern wasn't working then it would be good to know and hopefully address in a better way.

The intended pattern is:
a) When a query is executed, the server should hit `fluree/load` - and if already loaded a cached version will be returned.
b) `fluree/notify` can be called at any scale and volume but will only keep ledgers updated where `fluree/load` had been called (they are cached). If a ledger is not cached, the message should be ignored. If a query for that ledger comes in later, it will call `fluree/load` and then keep it maintained from there.
c) When a ledger has had no query activity for some period of time, it should be dropped. It can always be `fluree/load` in the future if a query does come in. This is not yet implemented, but meant to garbage collect and avoid keeping a ledger updated that nobody is using.

The function you reference basically invalidates all of the above. Again, maybe there was a reason the above wasn't working right.

cap10morgan: @JaceRockman Any reason the above approach can't work here instead?

cap10morgan: I implemented this locally and all the tests still pass and it makes logical sense to me. So I'm going to close this PR.

PR #836, opened 2024-07-16
Add status codes to crypto/credential failures

bplatz:
Add error message status code to crypto verification exceptions

### DIFF
--- a/src/clj/fluree/db/json_ld/credential.cljc
+++ b/src/clj/fluree/db/json_ld/credential.cljc
@@ -98,19 +98,26 @@
            auth-did      (did/auth-id->did id)]
        (when (not= jws-header-json header)
          (throw (ex-info "Unsupported jws header in credential."
-                         {:error :credential/unknown-signing-algorithm
+                         {:status 400
+                          :error :credential/unknown-signing-algorithm
                           :supported-header jws-header-json
                           :header header
                           :credential credential})))
 
        (when (not (crypto/verify-signature pubkey signing-input signature))
-         (throw (ex-info "Verification failed." {:error :credential/invalid-signature :credential credential})))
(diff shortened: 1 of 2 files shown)


src/clj/fluree/db/json_ld/credential.cljc
```diff
@@ -98,19 +98,26 @@
...
+                         {:status 400
+                          :error :credential/unknown-signing-algorithm
                           :supported-header jws-header-json
                           :header header
                           :credential credential})))
 
        (when (not (crypto/verify-signature pubkey signing-input signature))
-         (throw (ex-info "Verification failed." {:error :credential/invalid-signature :credential credential})))
+         (throw (ex-info "Verification failed, invalid credential."
+                         {:status 400
```
dpetran, line 109: Should we consider using a 403 here?

bplatz: That's what I did at first but then reconsidered as the request itself is invalid but definitely a gray area. 

I did a quick google search and found other tools doing 400 so that was enough to nudge in this direction. Certainly open for debate though.

Jackamus29: Yeah, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403):
> The HTTP 403 Forbidden response status code indicates that the server understands the request but refuses to authorize it.

I would use a 400 for malformed requests (which includes an invalid credential) and 403 for a valid credential, but insufficient permissions.

dpetran: That makes perfect sense, I'm happy with the 400.

dpetran (approved):
:stadium:

PR #837, opened 2024-07-16
Fix/wildcard selector

dpetran:
During the SPARQL work I discovered a weird additional nesting that wildcard queries introduced. It was introduced by select notation like this:
```
{:select ["*"] ...}
```
In SPARQL, wildcard selectors cannot be composed with any other selectors, and putting it in a sequence like that introduces the extra layer of nesting in the results (see the changes to sparql_test.cljc). The solution was to never translate "SELECT *" into `:select ["*"]`, but rather to this:
```
{:select "*" ...}
```

In the course of investigating this, I discovered that in FQL we do allow the composition of a wildcard selector alongside other selectors. I don't think this composition was intentional, but I do think if users come across it they will be confused by the results. The second commit, 111d0c449d7c251661f15f19f4d8433836cdd1a0, rearranges the FQL grammar to only allow a single wildcard in select and disallows composition with other selectors. You can still have multiple wildcards if you have multiple subgraph selectors: 
```
{:select [{?s ["*"]} {?foo ["*"]}]}
```

### DIFF
--- a/src/clj/fluree/db/query/exec/select.cljc
+++ b/src/clj/fluree/db/query/exec/select.cljc
@@ -93,18 +93,13 @@
   (implicit-grouping? [_] false)
   (format-value
     [_ db iri-cache _context compact _fuel-tracker error-ch solution]
-    (go-loop [ks (keys solution)
-              formatted {}]
-      (let [k          (first ks)
-            fv         (-> solution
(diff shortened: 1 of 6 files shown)


zonotope (approved):
🍏

PR #838, opened 2024-07-17
Update subquery tests to new API

bplatz:
For some reason the subquery PR https://github.com/fluree/db/pull/794 showed green on checks but never ran CICD on the last merge.

There was the one eastwood exception that popped up on `main` with that CICD run.

Fixed here.


dpetran (approved):
:+1:

PR #842, opened 2024-07-18
Rename main db record to `FlakeDB`

zonotope:
This patch is a smaller subset of the changes in #789. That pull request became very large and is now out of date. Instead of resolving the conflicts there and landing all those changes at once, I figured it would be less disruptive to break that pull request into more manageable pieces.

This patch renames the main db record from `JsonLdDb` to `FlakeDB`, and changes the implementing namespace from `fluree.db.db.json-ld` to `fluree.db.flake.flake-db`. Since we now have multiple db records, I think these names are more descriptive


dpetran (approved):
:snowflake:

PR #843, opened 2024-07-18
Fix sparql query-credential APIs

bplatz:
This fixes SPARQL support for credential-query and credential-query-connection.


src/clj/fluree/db/api.cljc
```diff
@@ -314,11 +315,13 @@
 
 (defn credential-query-connection
   ([conn cred-query] (credential-query-connection conn cred-query {}))
-  ([conn cred-query opts]
+  ([conn cred-query {:keys [format] :as opts}]
    (promise-wrap
     (go-try
-     (let [{query :subject, identity :did} (<? (cred/verify cred-query))]
-       (<? (query-connection conn query (assoc opts :did identity))))))))
+     (let [json? (not= :sparql format)
```
dpetran, line 321: Would it be possible to pass the actual format key down? I tend to prefer arguments that carry a bit of semantics with them, rather than a naked boolean.

bplatz: I think the issue here is that `cred-verify` should be a verifier and not a JSON parser. Those two are conflated here.

I'd rather split up those roles... but my goal here wasn't refactoring code, it was fixing what was not working.

bplatz: And by that I also mean that fluree/server relies on cred/verify and it's JSON parsing... so any change in that behavior is going to affect not just fluree/db/

zonotope: I think it makes sense for us to change that behavior sooner rather than later because quick short term changes to get things working have a tendency to be codified into the api.

bplatz: Made a ticket here: https://github.com/fluree/db/issues/845

bplatz: > I think it makes sense for us to change that behavior sooner rather than later because quick short term changes to get things working have a tendency to be codified into the api.

The function already did JSON parsing, that is not new. What is new is that a new arity option can tell it to not do that.

zonotope: Right. What I meant was that we should separate parsing from verification sooner rather than later if that's what we wanted to do. 

I'm not the biggest fan of the boolean argument either. If we were to keep the parsing here long term, then I would support instead splitting the functions up in a different way. Instead of a `verify` function with a `json?` argument, I think it would be better to have two functions: `verify-json` and `verify-string`, and the function here would call one of them depending on whether the format was `:sparql` or not.

bplatz: Not sure why CI/CD stopped working - bunch of tests are now failing so pulling this down for now. Maybe re-enabling it after I figure out the failures will kick it off again.

bplatz: Hopefully addressed this. Changed the following:
1) removed JSON option that was added
2) remove JSON parsing from `cred/verify-jws`, and moved to `cred/verify` only for JWS
3) Refactored credential APIs for SPARQL to call `cred/verify-jws` directly, and no longer use `cred/verify`

In summary, `cred/verify` functionality is identical to that before and everything relying on it will remain working, but for SPARQL we now no longer use that function (and therefore avoid needing adding the extra JSON parsing flag)

bplatz: Updated https://github.com/fluree/db/issues/845 to reflect this latest set of changes, removing options of how to handle boolean flag because it is no longer being used.

src/clj/fluree/db/query/api.cljc
```diff
@@ -34,15 +34,19 @@
...
-                 (<? (perm/wrap-identity-policy db did true nil))
-                 db)
-          db** (-> (if t
-                     (<? (time-travel/as-of db* t))
-                     db*))]
-      (assoc-in db** [:policy :cache] (atom {})))))
+   (let [db*  (if did
+                (<? (perm/wrap-identity-policy db
+                                               did
+                                               (or (true? default-allow?)
```
dpetran, line 42: Is this `or` expression even necessary? I think it just boils down to `default-allow?`

JaceRockman:
Most of these changes are already covered in https://github.com/fluree/db/pull/800/

src/clj/fluree/db/query/api.cljc
bplatz: That's a good point... let me look.

bplatz: Added malli option here 7b3d4d51bc8c9268d584e9b1af7902aee90d4c73 , and updated params in 1eea5cb038b5ac7b38a538e7641a46aa224c5170

dpetran (approved):
:crescent_moon:

bplatz:
> Most of these changes are already covered in #800

@JaceRockman Do you want me to hold off merging this and work on the corresponding changes in your branch?

I originally tackled this based on your ask in Slack... and albeit it turned from a single-commit fix that took about an hour to to something that ended up being much larger to address the good feedback and some other related changes that were not fix-specific.

JaceRockman:
> > Most of these changes are already covered in #800
> 
> @JaceRockman Do you want me to hold off merging this and work on the corresponding changes in your branch?
> 
> I originally tackled this based on your ask in Slack... and albeit it turned from a single-commit fix that took about an hour to to something that ended up being much larger to address the good feedback and some other related changes that were not fix-specific.

Yeah, I see some of your comments on my PR, I can fix those there if we just want to make all of these changes as one PR.

dpetran:
I'm inclined to merge this one, and then rebase #800 on this work. This one is ready to go and is a subset of that one, so that seems like the smoothest path.

bplatz:
Closed in favor of #800

PR #844, opened 2024-07-18
Refactor/db protocol implementations

zonotope:
This patch is the second in a series of replacements for #789 that cleans up some of the `FlakeDB` protocol implementations (more work to be done on that front later) and adds a new protocol for reasoning. The reasoning api functions currently in main operate on a `FlakeDB` directly, so they would break if they were ever called on an `AsyncDB`.


src/clj/fluree/db/query/exec.cljc
```diff
@@ -7,10 +7,16 @@
             [fluree.db.query.exec.order :as order]
             [fluree.db.query.exec.having :as having]
             [fluree.db.util.core :as util]
+            [fluree.db.query.exec.select.subject :as subject]
             [fluree.db.util.log :as log :include-macros true]))
 
 #?(:clj (set! *warn-on-reflection* true))
 
+(defn queryable?
```
dpetran, line 15: Is this ever used in a hot path? `satisfies?` is [very slow](https://clojure.atlassian.net/browse/CLJ-1814) and should be avoided if performance is going to be an issue.

zonotope: I hadn't realized that about `satisfies?`. This function is only ever used [here, when the reasoner looks up rules](https://github.com/fluree/db/pull/844/files#diff-840707c6ff2530f8a56111942cc8facba0efd3e648581a48d2d9020347e95b3cR171), which only happens once per reasoning run. 

This code used to check directly if the supplied instance was a `FlakeDB`, which won't work anymore because we now support multiple db types (and we'll most likely add more in the future). The important bit that we need to know is whether the argument can work with the `where/Searchable` and `subject/SubjectFormatter` protocols because they are used to extract the rules.

I'll try to think of a way to avoid `satisfies?`, but since it's only called once per reasoning run, it probably isn't the end of the world if we don't have any alternatives.

dpetran: That's sounds good, I think it's fine as long as its on our radar.

dpetran (approved):
:+1:

PR #846, opened 2024-07-20
Allow indexing options & persist

bplatz:
We lost the ability to set indexing options `reindex-min-bytes` and `reindex-max-bytes`.

This restores that and gives the ability to do so when creating a ledger.

In addition, it stores these settings in the index-root, so when loading a DB from disk it will retain the original settings, but there is the ability to override it (and then preserve the new settings upon the next index).

I discovered this issue when working on indexing garbage collection which is broken, but this needed to get addressed first.


zonotope (approved):
🎭

PR #847, opened 2024-07-20
Remove some old dev namespaces

bplatz:
Came across some old dev namespaces that I don't think are needed/relevant any longer.


zonotope (approved):
🧹

PR #848, opened 2024-07-20
Write garbage addr, previous idx info into index-root

bplatz:
Builds on https://github.com/fluree/db/pull/846

This captures the garbage file address and puts it in the index-root, so the file can be found and utilized for garbage collection in the future.

In addition, the previous index's t value written to the index-root was wrong, and the previous index's address wasn't being captured at all. This would make it impossible to find previous index root files from the current one, and therefore impossible to do garbage collection (without having to read every historical commit file ever produced).


zonotope (approved):
🛷

PR #849, opened 2024-07-22
Automatic garbage collection

bplatz:
Builds on PR https://github.com/fluree/db/pull/848.

This adds automatic old index garbage collection.

By default we will retain 3 old indexes, but any more than that will be automatically deleted to minimize disk space. If the default setting is not desired, it can be set when creating a new ledger (see below) and the setting is retained in the index-root(s), therefore upon a db load the setting will be preserved.

The easiest way to try this would be to turn `reindex-min-bytes` setting per below way down, so a new index is created with every commit. With `max-old-indexes` set to 1, you'll see at most one old index retained on disk. The rest will be auto garbage collected.

```
 (def ledger 
  @(fluree/create conn "mydb" {:indexing {:reindex-min-bytes 1000 ; 1 kb
                                          :max-old-indexes   1}}))
```

### DIFF
--- a/src/clj/fluree/db/indexer/garbage.cljc
+++ b/src/clj/fluree/db/indexer/garbage.cljc
@@ -1,0 +1,91 @@
+(ns fluree.db.indexer.garbage
+  (:require [clojure.core.async :as async :refer [<! go]]
+            [fluree.db.indexer.storage :as storage]
+            [fluree.db.util.async #?(:clj :refer :cljs :refer-macros) [<? go-try]]
+            [fluree.db.util.core :as util]
+            [fluree.db.util.log :as log]))
+
+#?(:clj (set! *warn-on-reflection* true))
+
+(defn garbage-meta-map
+  "Garbage file metadata that will be passed to the
+  garbage cleaning function that allows it to report
+  on status and have filename addresses."
+  [{:keys [prev-index garbage] :as idx-root}]
+  (when garbage ;; first index will not have garbage
+    (assoc garbage
+      ;; retain not the current index file, but the previous one
+      ;; where the garbage in this file originated
+      :t (:t prev-index)
+      :index (:address prev-index))))
+
+(defn trace-idx-roots
+  [conn commit]
+  (go-try
+   (loop [next-idx-root (<! (storage/read-db-root conn (-> commit :index :address)))
+          garbage       []]
+     (if (or (nil? next-idx-root) ;; no more idx-roots
+             (util/exception? next-idx-root)) ;; if idx-root already deleted, will be exception
+       garbage
+       (let [garbage-meta  (garbage-meta-map next-idx-root)
+             prev-idx-addr (get-in next-idx-root [:prev-index :address])
+             garbage*      (if garbage-meta
+                             (conj garbage garbage-meta)
+                             garbage)]
+         (recur (<! (storage/read-db-root conn prev-idx-addr))
+                garbage*))))))
+
+(defn clean-garbage-record
+  "Cleans up a complete garbage file, which will contain
+  many index segment garbage items within it. Garbage file
+  looks like:
+  {:alias 'ledger-alias',
+   :branch 'main',
+   :garbage ['fluree:file://.../index/segment/1.json', ...]
+   :t 1}"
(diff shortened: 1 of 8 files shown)


src/clj/fluree/db/indexer/garbage.cljc
```diff
@@ -0,0 +1,91 @@
...
+
+(defn clean-garbage
+  "Cleans up garbage data for old indexes, but retains
+  the last `max-indexes` indexes. Note that if not retaining
+  at least one prior index, current queries or handles on the latest
+  db might fail, therefore `max-indexes` is recommended to typically
+  be 1 or more."
+  [{:keys [conn commit] :as _db} max-indexes]
+  (go-try
+   (assert (and (int? max-indexes)
```
dpetran, line 74: `assert` is something that makes me nervous because `AssertionError`s aren't part of the `Exception` tree, so it will bubble through all our exception handling and take down the process.

src/clj/fluree/db/indexer/garbage.cljc
```diff
@@ -0,0 +1,91 @@
...
+                (>= max-indexes 0)))
```
dpetran, line 75: If a `max-indexes` of zero will break queries some percentage of the time, should we just disallow it entirely?

dpetran: If we could validate this wherever it comes from instead of all the way down here where it's used I would feel better.

src/clj/fluree/db/indexer/garbage.cljc
bplatz: At this point this is a background process to a background process (indexing) - there is nothing that is looking at the response at all, and nothing to capture any exceptions. If exceptions occur, it won't stop the operation of the ledger, but they will appear in the logs.

The main goal of the assert is to stop processing to avoid unexpected results, but not to produce anything that might use it upstream.

I'm happy to change this to an ex-info however - in case something does use the results of this some day it will be in a better place.

src/clj/fluree/db/indexer/garbage.cljc
bplatz: > If a `max-indexes` of zero will break queries some percentage of the time, should we just disallow it entirely?

It will only break queries if you are holding onto an old `db` object which points to the old indexes - but that risk exists even if you have 100 old indexes. It is wholly dependent on how you use the API and how often index updates are happening - but in most circumstances it wouldn't present an issue at all.

Sometimes garbage can consume a lot of disk space, and if you are running this in a constrained environment and know the `0` setting won't impact how you use fluree, then this will be useful to have.

bplatz: > If we could validate this wherever it comes from instead of all the way down here where it's used I would feel better.

It mostly comes from the index-root that we write to disk, which is where we read the value from. We don't currently validate any index-related data on disk.

bplatz: Adjusted docstring to more accurately reflect risks of index garbage collection here: b8059a06afb624e55fd8282c4c8436a4df111f11

src/clj/fluree/db/indexer/garbage.cljc
bplatz: Adjusted here: 20b919c2991b2ced63309270dab9f355186eec21

src/clj/fluree/db/indexer/garbage.cljc
```diff
@@ -0,0 +1,98 @@
...
+  Note that any db's held as a variable will rely on the index-root
+  from when they were pulled from the ledger via (fluree/db <ledger>).
+  If these db vars are held over time, you might want to adjust this
+  setting such that old index-roots are not garbage collected during that
+  expected timeframe. The frequency of new indexes being created is
+  dependent on the frequency and size of updates that is ledger-specific
+  against the ledger's 'reindex-min-bytes' setting."
+  [{:keys [conn commit] :as _db} max-indexes]
+  (go
+    (if (and (int? max-indexes) (>= max-indexes 0))
```
dpetran, line 79: `nat-int?` should cover both properties of this `and`

PR #850, opened 2024-07-22
don't include author in commit for unsigned transactions

dpetran:
Instead of saving an empty string, we now omit the `author` flake for unsigned transactions.


zonotope (approved):
♐

PR #851, opened 2024-07-24
Nameservice push! refactor to rely on only JSON-LD commit

bplatz:
While working on fluree/server, the raft implementation was broken because nameservice record support wasn't working properly.

While all the files were getting replicated properly and consensus was working fine, the nameservice record itself kept locally on every server wasn't getting written out properly. Previous refactors around nameservices appear to have broken this and it wasn't realized. This prevented all the servers from being able to properly 'load' ledgers they were maintaining together.

This PR refactors the nameservice push! operations to rely *only* on the JSON-LD commit record, which is known in raft consensus, and means fluree/server can easily maintain proper NS records across a distributed network without a lot of extra work.

Before this PR, several pieces of metadata were sent to `push!` - including the JSON-LD commit we now only rely on, but also the keyword-map commit, the write file results, and even the `ledger`'s state atom. The work in this PR removed the dependency on all of those other pieces of metadata such that just the single JSON-LD commit could be used in all nameservice implementations and work.

### DIFF
--- a/src/clj/fluree/db/ledger/json_ld.cljc
+++ b/src/clj/fluree/db/ledger/json_ld.cljc
@@ -311,11 +311,7 @@
   {:closed?  false
    :branches branches
    :branch   current-branch
-   :graphs   {}
-   :push     {:complete {:t   0
-                         :dag nil}
-              :pending  {:t   0
-                         :dag nil}}})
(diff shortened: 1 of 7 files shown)


zonotope (approved):
🏋🏾‍♂️

PR #853, opened 2024-07-31
change cache size opts to cache-max-mb

bplatz:
The default LRU cache for storing index segments defaulted to 10, and recent changes to fluree/server was also not picking up the correct options key name and was being ignored - so regardless of the config settings LRU cache was always 10.

This significantly slowed down queries/transactions for any but the smallest dbs.

This changes the default if no settings exist to 100Mb of memory for the cache (1,000 LRU cache items)

This also renames the setting into connection creation from `memory` to the better named `cache-max-mb`, which is what the new fluree/server settings is used - so it will now pick up the fluree/server config settings.


dpetran (approved):
:mega:

PR #854, opened 2024-07-31
Refactor/lru cache

bplatz:
This makes the following changes:
1) consolidates the index-resolver code which was identical across all connections
2) change the LRU cache to store promise-chans and not values. This has two effects:
  a) Concurrent requests for the same uncached resource are highly likely to share the same request now, whereas they were highly likely previously to do the same work multiple times.
  b) index-resolver is called a lot, and it creates a new chan every time to put the ultimate value on (even if cached). Now we don't need to create new chans for every call that don't serve much of a purpose.
3) Moves exception-handling when caching to the cache. It was not handled consistently before.

### DIFF
--- a/dev/remote_conn.clj
+++ b/dev/remote_conn.clj
@@ -13,8 +13,12 @@
 
   (def db (fluree/db ledger))
 
-  @(fluree/query db {:select {"?s" ["*"]}
-                     :where  [["?s" "ex:name" nil]]})
+
+
+  @(fluree/query db {"@context" {"ex" "http://example.org/"}
+                     "select"   {"?s" ["*"]}
+                     "where"    {"@id"     "?s"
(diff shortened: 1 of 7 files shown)


zonotope (approved):
This looks like a big improvement. 

I think we're really close to being able to remove the `index/Resolver` protocol entirely. I added that a while back before we had a general storage protocol, and we needed a protocol for retrieving index nodes from storage. Now that we have a general purpose storage protocol, we should eventually use that instead for retrieving index nodes as we use it for everything else.

The only thing the `index/Resolver` protocol is doing for us now is providing an abstraction over caching. I think we eventually want to add caches to the storage protocol implementations directly, and then the storage implementations will do everything that the `index/Resolver` protocol defines and we can switch to plain functions operating on storage records for storing and retrieving index nodes.

PR #855, opened 2024-07-31
Fix/query optimization

bplatz:
This does query offset/limit before query selection formatting in addition to increasing the parallelism to 3 for selection formatting.

These changes have a strong performance impact for queries that have large graph crawls in the 'select'.


zonotope (approved):
👏🏾

PR #856, opened 2024-08-01
Remove `implicit-grouping?` protocol method

zonotope:
This small patch removes the `implicit-grouping?` method from the `ValueSelector` protocol and reinstates the `implicit-grouping?` function because how solutions should be grouped should not be a concern of any of the individual selector protocols. It also separates the `format-value` and `solution-value` methods into different protocols because these two methods won't always be definable in the same contexts.


dpetran (approved):
:parasol_on_ground:

PR #857, opened 2024-08-03
Datatype filtering

zonotope:
This patch adds features for querying with datatypes. There are examples of the added features in the [datatype test](https://github.com/fluree/db/blob/933ced3e47f0df0c8934e586f3a77462e0d82b1b/test/fluree/db/query/fql_test.clj#L322-L383)

First, it adds a `datatype` function that users can use to bind the datatype of a matched object to a variable or for use in more complex functions. It also adds the ability to filter and restrict the datatype of a match in an `@value` map within a triple (similar to how users can set language tags with `@language`).

This patch also adds the `iri` function so that users can explicitly set a string to be an iri within functions because this was essential for allowing users to compare a datatype extracted from the database with one they included in code. More work needs to be done to extend the `iri` function to make it work everywhere a user would expect, but this is just a start.

As an implementation detail, this patch adds a new `IRI` record type that both the `iri` and `datatype` function returns. This enables the type inference code to detect datatypes automatically. I would like to eventually parse all of the iris into `IRI` records because that would unlock more functionality, but that's a monumental task that I think we should take on gradually. 

Beyond that, this patch includes some cleanup of select clause processing because we no longer need to rely on cached sids/iris while displaying values since we changed the way we manage and generate sids a while ago, so there was a lot of dead code that would never be reached. This cleanup also has the advantage of making the display multimethod no longer need to be async. 

I also would like to clean up query processing to unify both value matches and iri matches. This separation was a vestige of previous versions of the database when we couldn't easily compute an iri from an sid or vice versa. Now that we can, this separation is no longer necessary, and I think the code would be much simpler and less error prone without it. I started some of that unification here to deliver these features, but I think finishing it is beyond the scope of this pr.

### DIFF
--- a/src/clj/fluree/db/json_ld/iri.cljc
+++ b/src/clj/fluree/db/json_ld/iri.cljc
@@ -23,29 +23,30 @@
   [s]
   (->IRI s))
 
-(defn iri->string
-  [iri]
-  (:s iri))
-
 (defn unwrap
   [x]
   (if (iri? x)
-    (iri->string x)
+    (str x)
     x))
 
-(defn display-iri ^String
+(defn serialize-iri ^String
   [iri]
-  (str "<\"" iri "\">"))
+  (str "\"" iri "\""))
 
 #?(:clj (defmethod print-method IRI [^IRI iri ^java.io.Writer w]
-          (let [iri-str (display-iri iri)]
+          (let [iri-str (serialize-iri iri)]
             (doto w
               (.write "#fluree/IRI ")
               (.write iri-str))))
    :cljs (extend-protocol IPrintWithWriter
            IRI
            (-pr-writer [iri writer _]
-             (write-all writer "#fluree/IRI " (display-iri iri)))))
+             (write-all writer "#fluree/IRI " (serialize-iri iri)))))
+
+#?(:clj (defmethod print-dup IRI
+          [^IRI iri ^java.io.Writer w]
+          (let [iri-string (str iri)]
+            (.write w (str "#=" `(string->iri ~iri-string))))))
 
 (defmethod pprint/simple-dispatch IRI [^IRI iri]
   (pr iri))
--- a/src/clj/fluree/db/json_ld/iri.cljc
+++ b/src/clj/fluree/db/json_ld/iri.cljc
@@ -2,6 +2,7 @@
   (:require [fluree.db.util.core :as util]
             [fluree.db.util.log :as log]
             [fluree.db.util.bytes :as bytes]
+            [clojure.pprint :as pprint]
             [clojure.string :as str]
             [clojure.set :refer [map-invert]]
             [nano-id.core :refer [nano-id]]
@@ -13,6 +14,23 @@
 (defrecord IRI [^String s]
   Object
   (toString [_] s))
+
+(defn display-iri ^String
+  [iri]
+  (str "<\"" iri "\">"))
+
+#?(:clj (defmethod print-method IRI [^IRI iri ^java.io.Writer w]
+          (let [iri-str (display-iri iri)]
+            (doto w
+              (.write "#fluree/IRI ")
+              (.write iri-str))))
+   :cljs (extend-protocol IPrintWithWriter
+           IRI
+           (-pr-writer [iri writer _]
+             (write-all writer "#fluree/IRI " (display-iri iri)))))
+
+(defmethod pprint/simple-dispatch IRI [^IRI iri]
+  (pr iri))
 
 (defn iri?
   [x]
--- a/src/clj/fluree/db/json_ld/iri.cljc
+++ b/src/clj/fluree/db/json_ld/iri.cljc
@@ -15,6 +15,24 @@
   Object
   (toString [_] s))
 
+(defn iri?
+  [x]
+  (instance? IRI x))
+
+(defn string->iri
+  [s]
+  (->IRI s))
+
+(defn iri->string
+  [iri]
+  (:s iri))
+
+(defn unwrap
+  [x]
+  (if (iri? x)
+    (iri->string x)
+    x))
+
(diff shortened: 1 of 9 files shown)


src/clj/fluree/db/query/exec/eval.cljc
```diff
@@ -289,21 +296,35 @@
   [term expressions]
   (contains? (set expressions) term))
 
+(def context-var
+  (symbol "$-CONTEXT"))
+
+(defmacro iri
+  [s]
+  `(-> ~s
+       (json-ld/expand-iri ~context-var)
```
dpetran, line 305: Just trying to understand this - why wouldn't an IRI already be expanded by the time it gets here?

src/clj/fluree/db/json_ld/iri.cljc
```diff
@@ -10,6 +11,46 @@
...
+
+(defn unwrap
+  [x]
+  (if (iri? x)
+    (str x)
+    x))
+
+(defn serialize-iri ^String
+  [iri]
+  (str "\"" iri "\""))
```
dpetran, line 34: Do we actually use this to store IRIs on disk? If it's just for pretty printing maybe `display-iri` is a better function name, so nobody is tempted to use it for storage in the future.

src/clj/fluree/db/query/exec/eval.cljc
zonotope: Because that's what this query "function" is doing. The user puts this inside of a query function definition. The parser isn't going to be able to expand iris during parse time because it might not have access to enough information for the expansion. It has to be expanded when the function is evaluated. For example:

```clojure
[:filter (= ?s (iri (str ?userNs ":" ?firstName)))]
```

src/clj/fluree/db/json_ld/iri.cljc
zonotope: I could go either way. We don't store iris on disk right now, but I do think we might eventually want to store iris on disk since sids can be ambiguous. The same namespace code could point to different namespaces from different ledgers. I had used "display" as this function name before, but when I came back and implemented the data reader for iris, serialize made more sense because the data reader takes the output of this function and turns it back into the iri, so serialize fit better considering all the ways this function is used.

test/fluree/db/query/fql_test.clj
```diff
@@ -319,32 +319,68 @@
...
-                               {:ex "http://example.org/ns/"}]
-                     :select  '[?age ?dt]
-                     :where   '[{:ex/age ?age}
-                                [:bind ?dt (datatype ?age)]]}
-            results @(fluree/query db query)]
-        (is (= [["forever 10" "xsd:string"] [36 "xsd:long"]]
-               results))))
-    (testing "filtering query results with datatype fn")
-    (testing "filtering query results with @type value map")))
+(deftest ^:integration datatype-test
```
dpetran, line 322: I attempted to generate an iri and ran into an issue where we were expecting iris to be strings. Here's the setup:

```
(def conn @(fluree/connect {:method :memory}))
  (def ledger @(fluree/create conn "upserttest"))

  (def db0 (fluree/db ledger))

  (def db1 @(fluree/stage db0 {"insert"
                               [{"@id" "ex:dan"
                                 "@type" "ex:Person"
                                 "ex:email" "definitely@exists.com"}]}))

  (def db2 @(fluree/stage db1 {"where"
                               [["optional" {"@id" "?maybeFriend1" "ex:email" "definitely@exists.com"}]
                                ["optional" {"@id" "?maybeFriend2" "ex:email" "nobody@doesntexist.com"}]
                                ["bind" "?maybeFriend3" "(or ?maybeFriend2 (iri (concat \"ex:\" (struuid))))"]]
                               "insert"
                               [{"@id" "ex:hasFriend"
                                 "@type" "ex:Person"
                                 "ex:friend" "?maybeFriend1"}
                                {"@id" "ex:newFriend"
                                 "@type" "ex:Person"
                                 "ex:newProp" "yup"
                                 "ex:friend" "?maybeFriend3"}
                                {"@id" "?maybeFriend3"
                                 "ex:email" "nobody@doesntexist.com"}]}))
```

I can investigate further, but I thought I'd bookmark it here.

zonotope: There are going to be lots of places where we expect iris to be strings that we'll have to deal with in turn. That's what I was talking about in the description of this pr when I mentioned that we'd need to convert the existing strings into iris, and make existing code work with iris, gradually. This pr is just a first step that has the minimum functionality to make data type filtering work.

dpetran: I'm a bit hesitant to merge this in while things are actively broken, though - we typically keep `main` in a working state. Is it possible to have a list of things that need to be updated? That way we'll at least have scope outlined and we can keep track of the ways in which it's broken.

zonotope: Things are not "actively broken". the iri function is a new feature that explicitly only works right now in specific contexts. using the iri function anywhere outside of a where clause is not currently supported. we can add support later. Actively broken means something was working before and isn't working now. I have added a new feature, but it only works in specific contexts. We can expand that support later.

dpetran: This is in the where clause, though? I'm ok with merging it in as long as we know which contexts it does work in and which it doesn't, and I don't understand which is which.

zonotope: More precisely, you are using the _result_ of the iri function in the insert, and that's what's causing the problem.

dpetran: Ok, thanks for clearing that up. Can we make a ticket for the followup work so we can do it? I think Andrew's going to need it soon one way or the other : )

dpetran (approved):
:record_button:

PR #858, opened 2024-08-05
Refactor/nameservice protocol

zonotope:
This patch adds a "storage-backed" nameservice (better name suggestions welcome!) to replace the file, memory, and s3 nameservice implementations. The storage-backed nameservice uses a generic store that implements the `fluree.db.storage/Storage` protocol, so we can decide where the storage-backed nameservice stores its data based on what kind of storage implementation we pass to it. 

These changes are based on #876 (despite the numbering) so please review that first if you haven't already.


test/fluree/db/query/stable_hashes_test.clj
```diff
@@ -28,10 +28,10 @@
                        :schema/age   30}]})
           db1    @(fluree/commit! ledger db0)]
       (testing "stable commit id"
-        (is (= "fluree:commit:sha256:bbnabt4dx7g2cetpxnpmj7yw3dwwztdo6r4pusndprm2cedpvjixm"
+        (is (= "fluree:commit:sha256:blsm452njcnvya2yr2t54y3stj6rgi3rmwgnkeut3njqkumjcj4a"
                (get-in db1 [:commit :id]))))
       (testing "stable commit address"
-        (is (= "fluree:memory://4be8331e870ec61478ca8c4f0fb834f392bcd8cc341be5c081ba9bf7c70ef11c"
+        (is (= "fluree:memory://098da2e8b5dd5ae8176a403f6739f269b24ad9b7103543a14946827f134cad15"
```
dpetran, line 34: Do you know what changes in what we persist to disk caused these hash changes?

zonotope: I do not, but I'll look into it

zonotope: Ok, I got it. This is from an inconsistency with the old file nameservice vs the memory nameservice. Addresses from the memory nameservice [include "/<branch>/head" at the end](https://github.com/fluree/db/blob/b8ce181559a36d021c0b1d611672ed84d438ad50/src/clj/fluree/db/nameservice/memory.cljc#L69-L71), while the file nameservice addresses [do not include any information about the branch](https://github.com/fluree/db/blob/b8ce181559a36d021c0b1d611672ed84d438ad50/src/clj/fluree/db/nameservice/filesystem.cljc#L57-L60). 

I based the behavior of the storage-backed nameservice on the file nameservice, but it would be easy to change it to match what the memory nameservice did. I chose the file nameservice because that's what has gotten the most real-world usage in a persisted environment so far.

dpetran: I don't think we need to change it back - the memory nameservice doesn't persist so we won't be breaking anything by changing the convention and I'm in favor of being consistent.

dpetran (approved):
:card_file_box:

PR #859, opened 2024-08-07
Feature/sid migration

dpetran:
We've changed several things since v3, and this is a tool to bring existing v3 ledgers up to compatibility with main.

When this migration is run, it will:
1. generate a nameserver record for the ledger head
2. create a new genesis commit for the ledger
3. process each commit in order in order to:
    - generate index nodes that have correctly serialized SIDs
    - update commit data with new namespace codes
    - remove empty string author entries
    - correctly link to the previous commit, including the new genesis commit
 
Afterwards the old files will be left behind in the `<ledger>/<branch>` directory, which can be safely delete manually once everything is confirmed to be working correctly. We may write an additional migration in the future to delete the legacy data directory.

### DIFF
--- a/src/clj/fluree/db/json_ld/migrate/sid.cljc
+++ b/src/clj/fluree/db/json_ld/migrate/sid.cljc
@@ -4,7 +4,7 @@
             [fluree.db.async-db :as async-db]
             [fluree.db.connection :as connection]
             [fluree.db.constants :as const]
-            [fluree.db.flake.flake-db :as db]
+            [fluree.db.flake.flake-db :as flake-db]
             [fluree.db.flake.transact :as flake.transact]
             [fluree.db.json-ld.commit-data :as commit-data]
             [fluree.db.json-ld.iri :as iri]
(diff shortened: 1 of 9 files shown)


dpetran:
This has only been tested using a filesystem nameserver, there may be some kinks to work out with other nameserver methods.

src/clj/fluree/db/json_ld/migrate/sid.cljc
```diff
@@ -21,109 +24,139 @@
...
-;;       (-> db*
-;;           (reify/merge-flakes t-new all-flakes)
-;;           (assoc :previous (:commit db*))
-;;           (assoc :commit commit-metadata)))))
+(defn migrate-commits
+  "Reduce over each commmit and integrate its data into the ledger's db."
+  [ledger branch tuples-chans]
+  (go-try
+    (loop [[[commit-tuple ch] & r] tuples-chans
+           db (<? (async-db/deref-async (jld-ledger/current-db ledger)))]
```
zonotope, line 113: it's possible that current db is a FlakeDB. I think you're going to have to check what kind of db it is before trying to deref it.

src/clj/fluree/db/json_ld/migrate/sid.cljc
```diff
@@ -1,17 +1,20 @@
 (ns fluree.db.json-ld.migrate.sid
-  (:require [fluree.db.constants :as const]
-            [fluree.db.query.exec.update :as update]
+  (:require [clojure.core.async :as async]
+            [clojure.string :as str]
+            [fluree.db.async-db :as async-db]
+            [fluree.db.connection :as connection]
+            [fluree.db.constants :as const]
+            [fluree.db.flake.flake-db :as db]
```
zonotope, line 7: I think this alias should be `flake-db` so it's more clear what the full namespace is when it's used.

zonotope (approved):
✍🏾

src/clj/fluree/db/json_ld/migrate/sid.cljc
dpetran: This ledger is a brand new one returned from a call to `jld-ledger/create*` with no data transacted into it, so I think it will always be an AsyncDB the first time you call `current-db` on it, but I'll add a guard just in case.

PR #860, opened 2024-08-09
Inline Transaction Filtering

zonotope:
This patch adds support for specifying either explicit or variable transactions within a value map during a query, and you can see [examples for using this feature](https://github.com/fluree/db/compare/feature/inline-datatype-filtering...feature/inline-transaction-filtering?expand=1#diff-d9f2d68ae5213e13d8a2b9d402cc073f631b4eb0de343b6a4d85312b9306dd9cR422-R499) in the `fluree.db.query.fql-test/t-test`. 

This patch also fixes a number of bugs, including one with `:bind` patterns where solutions were still considered valid even if a bound variable was already set which didn't agree with the value from the binding, as well as a bug with language and data type matching when bound variables were already set. 

This patch builds on #857.

### DIFF
--- a/test/fluree/db/query/fql_test.clj
+++ b/test/fluree/db/query/fql_test.clj
@@ -384,12 +384,13 @@
         (testing "filtered with the datatype function"
           (let [query   {:context [test-utils/default-context
                                    {:ex "http://example.org/ns/"}]
-                         :select  '[?age ?dt]
-                         :where   '[{:ex/age ?age}
+                         :select  '[?name ?age ?dt]
+                         :where   '[{:ex/name ?name
+                                     :ex/age  ?age}
                                     [:bind ?dt (datatype ?age)]
                                     [:filter (= (iri :xsd/long) ?dt)]]}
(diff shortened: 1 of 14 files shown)


test/fluree/db/query/fql_test.clj
```diff
@@ -418,6 +419,85 @@
...
+                                             :type  :xsd/int}}
+                                  {:id      :ex/bart
+                                   :ex/name "Bart"
+                                   :ex/age  "forever 10"}]})
+          db1*   @(fluree/commit! ledger db1)
+          db2    @(fluree/stage db1* {"@context" ["https://ns.flur.ee"
+                                                  test-utils/default-context
+                                                  {:ex    "http://example.org/ns/"
+                                                   :value "@value"
+                                                   :type  "@type"}]
```
bplatz, line 448: Never thought about aliasing these... I guess we get that mostly for "free" in expansion.

bplatz (approved):
This is great - very excited to have this functionality!

PR #861, opened 2024-08-09
Fix/conn indexing opts

bplatz:
This addresses carrying :ledger-defaults on the conn as a default - specifically for indexing.

Previously, indexing default options were not being passed through.

This also adds some more functionality to the development index-explore namespace for troubleshooting indexing.


zonotope (approved):
🍡

PR #862, opened 2024-08-09
Use "@id" as the data type for IRIs

zonotope:
This patch uses "@id" as the internal datatype for IRI values. It builds off of #860


bplatz (approved):
I was thinking this would affect the `values` for policy enforcement but looks like you addressed that as well.

Nice to have this evolution of how we think of 'node' identifiers without needing an explicit scalar datatypes.

src/clj/fluree/db/json_ld/policy/enforce.cljc
```diff
@@ -55,7 +55,7 @@
                    :TODO
                    [(into ["?$this"] (keys values-map))
                     [(into [{"@value" this-var
-                             "@type"  "http://www.w3.org/2001/XMLSchema#anyURI"}] (vals values-map))]])]
+                             "@type"  "@id"}] (vals values-map))]])]
```
dpetran, line 58: I think we should use `const/iri-id` here.

dpetran (approved):
:+1:

zonotope:
I'm holding off on merging this because it needs a migration for existing databases. I am still working on some indexing stuff, but I'll return to the migration when I'm done with that.

PR #863, opened 2024-08-11
Fix/balance index

zonotope:
The reindexing algorithm created a new level in the index tree each time a branch node needed to be split because it overflowed, but this is only necessary for the root node. This patch changes the algorithm to split each branch node in place, and if the root node was split, it creates a new root node as the parent of the split roots. This should result in a significant performance improvement for large databases.


bplatz (approved):
Looks good!

PR #864, opened 2024-08-15
order-by syntax wasn't parsing correctly

bplatz:
`order-by` parsing had the following problems, which didn't allow for specifying "DESC" ordering:
1) when a direction was specified, e.g. [:desc ?myvar] - it was checking for `list?` which would fail for a vector, and therefore fail for parsed JSON. This is now changed to use `sequential?`.
2) Our docs specify using `"DESC"` and `"ASC"` - however we were only checking for `"desc"` and `"asc"`.

As an additional note for comment, the ordering needs to include the direction first, then the variable. This seems a bit counterintuitive to me.. perhaps that is the way its always been. It seems more natural to specify the variable first, then the direction... eg:
`[?score "DESC"]`
instead of how it currently needs to be:
`["DESC" ?score]`

Thoughts?

### DIFF
--- a/src/clj/fluree/db/query/fql/parse.cljc
+++ b/src/clj/fluree/db/query/fql/parse.cljc
@@ -603,11 +603,18 @@
            ensure-vector
            (mapv parse-var-name)))
 
+(defn wrap-order-statement
+  [order-by]
+  (if (syntax/order-by-tuple? order-by)
+    [order-by]
+    order-by))
+
 (defn parse-ordering
   [q]
   (some->> (or (:order-by q)
                (:orderBy q))
            ensure-vector
+           wrap-order-statement
            (mapv (fn [ord]
                    (if-let [v (parse-var-name ord)]
                      [v :asc]
--- a/src/clj/fluree/db/query/fql/syntax.cljc
+++ b/src/clj/fluree/db/query/fql/syntax.cljc
@@ -14,11 +14,11 @@
 
 (defn asc?
   [x]
-  (boolean (#{'asc "asc" :asc} x)))
+  (boolean (#{'asc "asc" "ASC" :asc} x)))
 
 (defn desc?
   [x]
-  (boolean (#{'desc "desc" :desc} x)))
+  (boolean (#{'desc "desc" "DESC" :desc} x)))
 
 (defn one-select-key-present?
   [q]
@@ -165,7 +165,7 @@
                          [:desc [:fn desc?]]]
     ::ordering          [:orn {:error/message "Ordering must be a var or two-tuple formatted ['ASC' or 'DESC', var]"}
                          [:scalar ::var]
-                         [:vector [:and list?
+                         [:vector [:and sequential?
                                    [:catn
                                     [:direction ::direction]
                                     [:dimension ::var]]]]]
--- a/src/clj/fluree/db/query/fql/syntax.cljc
+++ b/src/clj/fluree/db/query/fql/syntax.cljc
@@ -20,6 +20,12 @@
   [x]
   (boolean (#{'desc "desc" "DESC" :desc} x)))
 
+(defn order-by-tuple?
+  [x]
+  (and (sequential? x)
+       (or (-> x first desc?)
+           (-> x first asc?))))
+
 (defn one-select-key-present?
   [q]
   (log/trace "one-select-key-present? q:" q)


zonotope:
Ordering with directions is explicitly a list and not just sequential on purpose. This is to differentiate from a user supplying multiple orderings (vector) from a single ordering with direction (list). 

The direction has always come first (at least in version 3). I have analogized it with function execution, which also fits with the list requirement.

And, with the function analogy, specifying `"(desc ?myVar)"` in the json order clause does work.

I think all of these should work. If not, then they are bugs:

```json
{
  "@context": {
    "schema": "http://schema.org/",
    "ex": "http://example.org/ns/"
  },
  "select": [ "?name", "?email", "?age", "?favNums" ],
  "where": {
    "schema:name": "?name",
    "schema:email": "?email",
    "schema:age": "?age",
    "ex:favNums": "?favNums"
  },
  "groupBy": "?name",
  "orderBy": "?name"
}

{
  "@context": {
    "schema": "http://schema.org/",
    "ex": "http://example.org/ns/"
  },
  "select": [ "?name", "?email", "?age", "?favNums" ],
  "where": {
    "schema:name": "?name",
    "schema:email": "?email",
    "schema:age": "?age",
    "ex:favNums": "?favNums"
  },
  "groupBy": "(desc ?name)",
  "orderBy": "?name"
}

{
  "@context": {
    "schema": "http://schema.org/",
    "ex": "http://example.org/ns/"
  },
  "select": [ "?name", "?email", "?age", "?favNums" ],
  "where": {
    "schema:name": "?name",
    "schema:email": "?email",
    "schema:age": "?age",
    "ex:favNums": "?favNums"
  },
  "groupBy": [ "?name", "?age" ],
  "orderBy": "?name"
}

{
  "@context": {
    "schema": "http://schema.org/",
    "ex": "http://example.org/ns/"
  },
  "select": [ "?name", "?email", "?age", "?favNums" ],
  "where": {
    "schema:name": "?name",
    "schema:email": "?email",
    "schema:age": "?age",
    "ex:favNums": "?favNums"
  },
  "groupBy": [ "?name", "?age" ],
  "orderBy": ["?age" "(desc ?name)"]
}
```

zonotope:
To be clear, while I prefer the current behavior, I don't feel strongly about this, so if you do feel strongly and would like to change it, then I'm fine with that. My comment 👆🏾 was only meant to express the considerations behind the behavior and my mental model for it.

bplatz:
> I think all of these should work. If not, then they are bugs

They don't, and don't even get past our spec validation which is here:
https://github.com/fluree/db/blob/fbb8c415678de864e7a105ccb099c8051eab38cb/src/clj/fluree/db/query/fql/syntax.cljc#L166-L171

Currently you cannot order by descending unless you are using CLJ and use an quoted fn `'(desc ?x)`

In v2 this is not how it worked, but happy to have an updated syntax... just want to make the feature work.

zonotope:
Ok, then it _is_ a bug. We're missing the malli transformer to turn the ordering strings into function calls pre-validation. I could add that right quick (if we want to keep/enable the behavior I described) or we can go with this. Either way.

bplatz:
> Ok, then it _is_ a bug. We're missing the malli transformer to turn the ordering strings into function calls pre-validation. I could add that right quick (if we want to keep/enable the behavior I described) or we can go with this. Either way.

I'm happy to go with your described format, if you'd like to update Malli that would be great and I'll close this.

I do struggle a bit with having "code in strings" which we have also with bind/filter - so definitely not new ground here and this isn't an ideal time to revisit it. Not sure any alternative is better.

dpetran:
I've been playing around with an s-expression based json syntax for expressions, something like this:
```
;; current syntax:
{"where" [["bind" "?newvar" "(concat ?oldvar \"mystring\")"]]}
;; proposed syntax
{"where" [["bind" "?newvar" {"expr" ["concat" "?oldvar" "mystring"]}]]}
```
I started a little experiment last winter because quoting string args was driving me batty. I never took the time to finish it but I think it would be pretty easy to get that working. It's a little more verbose because you need to label it an "expr", but that's worth it in my book. And there's no reason we couldn't support both syntaxes side-by-side if we want.

zonotope:
I just opened https://github.com/fluree/db/pull/865

bplatz:
Closing in favor of https://github.com/fluree/db/pull/865

PR #865, opened 2024-08-15
Allow ordering directions to be specified as string

zonotope:
This patch fixes a bug where string ordering direction weren't being parsed correctly. It uses a malli decoder to turn strings to edn in the ordering before trying to validate it.

### DIFF
--- a/src/clj/fluree/db/query/fql/syntax.cljc
+++ b/src/clj/fluree/db/query/fql/syntax.cljc
@@ -96,7 +96,6 @@
 
 (defn string->ordering
   [x]
-  (log/info "Decoding ordering:" x)
   (if (string? x)
     (edn/read-string x)
     x))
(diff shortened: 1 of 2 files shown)


dpetran (approved):
:+1:

src/clj/fluree/db/query/fql/syntax.cljc
```diff
@@ -163,7 +170,9 @@
     ::direction         [:orn {:error/message "Direction must be \"asc\" or \"desc\""}
                          [:asc [:fn asc?]]
                          [:desc [:fn desc?]]]
-    ::ordering          [:orn {:error/message "Ordering must be a var or two-tuple formatted ['ASC' or 'DESC', var]"}
+    ::ordering          [:orn {:error/message "Ordering must be a var or two-tuple formatted ['ASC' or 'DESC', var]"
```
bplatz, line 173: We should update this error message.

PR #866, opened 2024-08-15
Ensure `leftmost?` is set for index nodes

zonotope:
There were some situations where the `:leftmost?` key was not set for index nodes after they or their parent nodes overflowed. This caused the system to lose track of some of these nodes during ancestor checks in certain situations because *both* the `first-flake` and `leftmost?` keys are required for accurate ancestor checks. This patch sets the `leftmost?` key in each case a node or it's ancestors could be split.

This patch also contains an optimization to eagerly push index nodes downstream for writing during a reindexing process instead of pushing fully resolved nodes onto the stack to wait for their ancestor nodes to be processed. Their ancestor nodes needs to have access to all of it's child nodes to determine if and how it should update it's `children` key, but we don't need to keep fully resolved nodes in the stack waiting for this. Instead, we can write the fully resolved node immediately and push an unresolved node (the same node, but without the `flakes` or `children` keys) onto that stack that it's ancestor can access. This should result in a significant reduction in memory consumption while reindexing large databases.

### DIFF
--- a/dev-resources/logback.xml
+++ b/dev-resources/logback.xml
@@ -7,7 +7,7 @@
         </encoder>
     </appender>
 
-    <root level="DEBUG">
+    <root level="INFO">
         <appender-ref ref="STDOUT"/>
     </root>
 
--- a/dev/index_explore.clj
+++ b/dev/index_explore.clj
@@ -33,10 +33,16 @@
             r))
         roots))
 
-(defn latest-t
+(defn t-values
   [roots]
   (->> roots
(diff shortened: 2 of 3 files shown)


bplatz (approved):
This is great. I think this might solve the memory pressure we had seen in the past under huge import load.

I'm also thinking this might solve the garbage collection issue, so I'll get that re-tested. The issue might not have been garbage, but the `leftmost?` bug - which appears would show during large data imports.

PR #867, opened 2024-08-15
Feature/vector query

bplatz:
This adds initial vector-based query capabilities allowing you to score dense vectors stored in Fluree against a target vector. This introduces some new query syntax and is hopefully enough to allow folks to play with the feature and give feedback on the syntax.

*note* This does not yet include any vectoring indexing, whose initial work is located here: https://github.com/fluree/vector-db

This allows regular Fluree queries to now include dense vector comparisons using dot product, cosine similarity or euclidean distance scoring. Test are [here](https://github.com/fluree/db/commit/ae460b9476d284a2542b84d4ffd8797be30cc751) containing many examples, but and example is:
```
{"@context": {"ex" :"http://example.org/ns/"},
 "select":   ["?x", "?score", "?vec"],
 "where":    [{"@id":    "?x",
               "ex:age": 36},
              {"@id":     "?x",
               "ex:xVec": {"@value"  "?vec",
                           "@vector": [0.7, 0.6],
                           "@score":  "?score",
                           "@metric": "dotproduct"}}]}
```
Next steps:
1) Collect any feedback on the syntax and feature
2) Incorporate vector index for fast retrieval of score-based sorted vectors when needed, and decide on syntax for calling the index vs just scoring (as this PR does).


zonotope:
These vector comparisons feel more like function execution to me than attributes and metadata on a value map. The `"@vector"` attribute, for example, doesn't have anything to do with the value itself, but it is an argument of an external operation on that value. Similarly, the `"@score"` attribute isn't intrinsic to the value, but the output of a comparison operation with a specific, unrelated other value.

I think using our existing facilities for function execution would be more consistent, so my syntactic suggestion would be:
```json
{"@context": {"ex" :"http://example.org/ns/"},
 "select":   ["?x", "?score", "?vec"],
 "where":    [{"@id":    "?x",
               "ex:age": 36,
               "ex:xVec": "?vec"},
              ["bind", "?score", "(dotProduct ?vec ?targetVec)"]],
 "values":   ["?targetVec", [{"@value": [0.7, 0.6], "@type": "f:vector"}]]}
                "?score", "(dotProduct ?vec ?targetVec)"]]]}
```

I think the main drawback of this syntax is that it's more verbose, but the main benefit of it is that it's more consistent with the way the rest of the system works and how you'd accomplish similar tasks today. At least, what I intuitively think are similar tasks.

This is just my 2 cents, so you can take it or leave it.

src/clj/fluree/db/json_ld/ledger.cljc
```diff
@@ -86,9 +86,7 @@
           "http://www.w3.org/ns/shacl#qualifiedMinCount"        const/sh_qualifiedMinCount
           "http://www.w3.org/ns/shacl#qualifiedMaxCount"        const/sh_qualifiedMaxCount
           "http://www.w3.org/ns/shacl#qualifiedValueShapesDisjoint" const/sh_qualifiedValueShapesDisjoint
-          ;; fluree
-          const/iri-role                                        const/$f:role
-          const/iri-target-class                                const/$fluree:targetClass}))
```
dpetran, line 91: Can you remind me what we used `targetClass` and `role` for? And is removing these from our predefined properties backwards compatible?

dpetran:
I think I'm in agreement with Ben on the syntax, and it should be a bit easier to tackle it as a function instead of having to do all the extra work around the match api.

dpetran:
Here's a quick working example I was able throw together, you've got all the bits already implemented so it went pretty quick. The big difference is that I'm calling `vectorize` inside the `dotproduct` function, instead of doing it at parse time.

https://github.com/fluree/db/commit/136b67ec338a665ba6e03f87cb9ce1e80bce1b1d

bplatz:
The syntax in a bind statement is good however we've yet to implement the functionality of the most common way vector search will be used - which is by hitting a vector index. 

The vector index will return at least 3 vars that we will want and maybe more. The score, the matching sid, the matching property, and sometimes matching vector (.-o) , the 't' and possibly the idx if a list (eg each paragraph of a doc has its own vector and idx would tell you the matching paragraph). None of those vars will be known ahead of time. 

Bind won't support this. We could go down the route of a new thing similar to optional or union where multiple vars can be captured. Or the syntax I have here is the same as how we capture multiple similar vars in any other statement. 

I think we could support new functions in bind however where you already know some vars amf just want to do math.

src/clj/fluree/db/json_ld/ledger.cljc
bplatz: These were related to the old policy functionality which is replaced by policy queries.

dpetran:
The vector index is sort of like a different graph, could we use the `graph` pattern to bind data specifically from that graph, then use bind to bind as many transformations as we need?

dpetran:
```
{"@context" {"ex" "http://example.org/ns/"}
                     "from"     ["graph/triple" "graph/vector"]
                     "select"   ["?x" "?score" "?vec"]
                     "values"   ["?targetVec" [{"@value" [0.7, 0.6] "@type" const/iri-vector}]]
                     "where"    [["graph" "graph/triple" {"@id" "?sid" "ex:xVec" {"@value" "?vec" "@t" "?T"}}]
                                 ["graph" "graph/triple" {"@id" "ex:xVec" "ex:score" "?score"}]
                                 ["bind" "?var1" "(vector-xf ?vec ?score)" "?var2" "(vector-xf2 ?var1 ?t)"]]}
```
Then we could be polymorphic on the type of the graph, we already have `-match-tuple` as part of the graph protocol.

bplatz:
> Vectors seem more to me like metadata describing an object than an object in it's own right

In a simple full-text search context I think that is an OK assumption.

In even slightly more involved full text search scenario it already starts to break that assumption. E.g. a sentence transformer is good at creating a vector per sentence... a single property with many sentences now needs multi-cardinality vector embeddings.

1 multi-cardinality vec -> 1 object (in a single index)

Embeddings/vectors are quite specialized - so it will be common to use multiple strategies for different use cases on the same data. e.g. I might use what I have used in the vector-db tests: all-MiniLM-L6-v2, which is a great general-purpose semantic search. It is not good at keyword search. If I want to find "similar content" I'd use that... if I wanted to provide a user a google-like keyword search box I'd want to use BM 25 or SPLADE instead. If I want both, I need both.

many models (vecs) -> 1 object (multiple indexes)

Then there will be the desire to combine multiple 'objects' into a single index ... e.g. ex:title + ex:summary as a single embedding.

Or I'll want to create an embedding for the entire subject (using a model that would be good at that) - so i can provide a JSON object and say "do I have anything similar in the DB already to this new thing".

1 vec -> 2+ "objects" (same subject)

And data scientists are constantly creating new/refined models, so I expect they will always be playing around with new vectors - and want to have multiple strategies in play at the same time.

These are all just examples to say flexibility is key - and flexibility I think best comes from what we already do well - ability to add properties onto subjects at your heart's content.

bplatz:
```json
{ ...
 "where":    [["find", "?score", "(dotProduct ?bio ?targetBio)"]
...}
```
Each vector index you want to create is tied to the embedding strategy + scoring calc (e.g. dotproduct). If you want to change either of those you need to build a different/new index.

If not hitting an index, you can do the scoring calc on the fly, and I see that as being super useful for heavily filtered candidates - but it will be magnitudes slower as it isn't optimized in an index. I see this being good for results in the 10s of potential candidates, but unusable for large numbers - which of course is where the index comes in.

I imaging you specifying the indexes you want us to create with metadata on the properties.
```json
{"@id": "ex:summarySplade",
 "@type": "rdf:Property",
 "rdf:label": "Splade embedding model on combination of ex:title and ex:summary for user keyword search",
 "f:vectorIndex": {
     "@type": "f:HNSW-Index-Config",
     "f:vectorMetric": {"@id": "f:dotProduct"},
     "f:efConstruction": 100,
     "f:efSearch": 25
  }
}
```
Those are examples of vector index configs... HNSW has several more that a user might want control over (other vector DBs allow the users to do this sort of tuning to optimize results).

I'd like to support more than HNSW indexes too, which is fairly slow but tried and true and consistent... so easiest one to get out the door quickly. There are tradeoffs for different indexing strategies that affect recall... HNSW is only about 90% of "perfect", but perfect isn't performant over large numbers.

bplatz:
Also as a reference point, here is pgvector (Postgres vector extension) - every vector type is created in a column (which is what I propose with our properties). They push the scoring into the order by clause.

It does provide a "sub query" like syntax possibility for handling vector search. It is also an option to consider although I'm not sure it makes it better at the cost of extra syntax complexity... need to think about it.

https://github.com/pgvector/pgvector

bplatz:
Have given this some further thought and I think for 'on the fly' scoring the proposed syntax works best with `bind`. So this is fine, but will never hit a vector index:
```json
{"@context": {"ex" :"http://example.org/ns/"},
 "select":   ["?x", "?score", "?vec"],
 "where":    [{"@id":    "?x",
               "ex:age": 36,
               "ex:xVec": "?vec"},
              ["bind", "?score", "(dotProduct ?vec ?targetVec)"]],
 "values":   ["?targetVec", [{"@value": [0.7, 0.6], "@type": "f:vector"}]]}
```

My hope was to have basically identical syntax whether doing that or hitting the vector index, but I think I've given up on that idea as I don't see a clear path to it.

So if hitting a vector index (which would have been pre-configured in a TBD manner), I propose these two syntaxes. The first avoids the need to bind `@id` outside of the normal pattern. The second feels more 'intentional' and I think I prefer it:

Option 1:
```
{"@context" {"ex" "http://example.org/ns/"},
 "select"   ["?x", "?title", "?score" "?vec"],
 "where"    [{"@id"      "?x",
              "ex:title" "?title"
              "ex:xVec"  {"@value"  "?vec",
                          "@search" {"@compare" [0.7, 0.6]
                                     "@score"   "?score"
                                     "@limit"   10}}}]}
```

Option 2:
```
{"@context" {"ex" "http://example.org/ns/"},
 "select"   ["?x", "?title", "?score" "?vec"],
 "where"    [["search" "ex:vec" {"@id"      "?id"
                                 "@value"   "?vec"
                                 "@compare" [0.7, 0.6]
                                 "@score"   "?score"
                                 "@limit"   10}]
             {"@id"      "?x"
              "ex:title" "?title"}]}
```

I hope both give flexibility to support different types of searching in the future, as they both provide effectively a flexible 'options map' that can be specific to what is needed for different value types (e.g. GeoJSON search might have different opts).

bplatz:
Option 3:

This is another option that was discussed.

While I don't personally think this option makes things more usable, I think it can work.

```json
{"@context": {"ex":            "http://example.org/ns/",
              "vecSearch":     "http://ns.flur.ee/hnsw/",
              "vecSearchInst": "http://ns.flur.ee/hnsw/instance/"},
 "select":   ["?x", "?title", "?score", "?vec"],
 "where":    [{"@type":             "vecSearchInst:myVecIndex",
               "vecSearch:compare": [0.7, 0.6],
               "vecSearch:limit":   10,
               "vecSearch:result":  {"vecSearch:id":    "?x",
                                     "vecSearch:score": "?score",
                                     "vecSearch:value": "?vec"}},
              {"@id":      "?x",
               "ex:title": "?title"}]}
```

In this case however we do loose the direct tie to referencing a specific property (`ex:xVec` in the examples above). That tie would have to be done in the index creation.. which in this case you would name `vecSearchInst:myVecIndex`. This is where you'd specify index options like metric, dimensionality, efConstruction (for hnsw) etc.

One advantage is those options would be different for each index type we create, which would be namespaced as in the example below: `http://ns.flur.ee/hnsw/` where a vocabulary could be defined that explains the options.

So the other side of this, in creating the index itself would be something like:
```json
{"@context": {"ex":            "http://example.org/ns/",
              "vecSearch":     "http://ns.flur.ee/hnsw/",
              "vecSearchInst": "http://ns.flur.ee/hnsw/instance/"},
 "insert":   {"@id": "http://ns.flur.ee/hnsw/instance/myVecIndex",
              "vecSearch:onProperty": {"@id": "ex:Bio"}, // this would be the property we'd pull vectors from
              "vecSearch:metric": {"@id" "vecSearch:dotproduct"},
              "vecSearch:efConstruction": 100,
              "vecSearch:efSearch": 25}}
```

bplatz:
@dpetran @zonotope Based on our discussions, this should be ready to go and for now simply adds a few vector scoring calculations (no vector indexes) for use with Fluree using the `bind` pattern.

I used @dpetran commit and expanded it for the other tests / math types - and removed all the other special syntax originally included here.

Next step will be to implement a vector index using some of the syntax we discussed here.

src/clj/fluree/db/vector/scoring.cljc
```diff
@@ -0,0 +1,37 @@
...
+  "Takes vector collection v and wraps it in a more optimized vector format
+  for faster vector math."
+  [v]
+  #?(:clj  (v/vec v)
+     :cljs v))
+
+(defn dotproduct
+  [v1 v2]
+  (when (and (vector? v1) (vector? v2))
+    #?(:clj  (v/dot (vectorize v1) (vectorize v2))
```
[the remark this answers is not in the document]
bplatz: I knew this wasn't super efficient but figured deferring speed optimization could be left for a later effort if needed.

Went ahead and did this now. It required 3 different changes:
a) coercing input values (in-memory flakes hold optimized vectorized objects instead of plain vectors)
b) coercing query `values` - this was probably a bug as I think it should be have been done all along, but this effort exposed it
c) coercing values returned by 'select'

dpetran (approved):
:vampire:

PR #868, opened 2024-08-16
fix sparql filter translation

dpetran:
We were generating FQL that was incompatible with our syntax schema, this changes the translator to produce compliant FQL.


zonotope (approved):
🪸

PR #870, opened 2024-08-18
Rebalance Indexes

zonotope:
I needed to write a migration for the changes introduced in #862 to be compatible with existing databases. This required me to load all the leaf nodes, extract their flakes, change the data type of affected flakes, and then rebuild leaves and branches from those new flakes. 

We will eventually need a reindex process that is very similar to that migration process. The only difference is the addition of changing the data types, which is just a mapping transducer.

Because these two processes were so similar, I decided to just implement the reindex process and then use the same machinery to implement the migration instead of implementing the migration with throwaway code. 

Besides that, this patch continues the flake db refactor by moving the flake index related namespaces under the `fluree.db.flake.*` namespace hierarchy.


dpetran (approved):
:balance_scale:

PR #872, opened 2024-08-21
support whitespace in additive expressions

dpetran:
The grammar we were using didn't allow for whitespace around numeric operators, this updates the grammar to support that.


zonotope (approved):
🍇

PR #873, opened 2024-08-22
Refactor/tx api

bplatz:
In order to support a new 'import' API in fluree/server which can accept Turtle (TTL) data, the transact API needs to now accept parsed triples directly, which is implemented in the new function `stage-triples`.

There was also some redundancy in work being done across this APIs, so tried to minimize the duplicate work.


zonotope (approved):
🍶

PR #874, opened 2024-08-22
Add consts rdf first, rest, nil

bplatz:
Add rdf:first, rdf:rest, rdf:nil

Needed for parsing TTL in fluree/server.


dpetran (approved):
:rabbit:

PR #876, opened 2024-08-24
Refactor/store protocol

zonotope:
This patch adds a `ByteStore` protocol to store and retrieved non-content-addressed data, and it splits the `iNameService` protocol by adding separate protocols for publishing and managing subscriptions. A number of nameservice implementations threw unsupported exceptions from their method definitions, which should never happen in a protocol implementation and indicates that a protocol should be split.


dpetran (commented):
I really like `ByteStore` vs `ContentAddressedStore` :evergreen_tree:

dpetran (approved):
:evergreen_tree:

PR #877, opened 2024-08-26
Feature/polymorphic compare

dpetran:
This adds the ability to compare non-numeric datatypes. I've added datatype inference support so literal return types of nested expressions and simple literals are supported as well.


src/clj/fluree/db/query/exec/eval.cljc
```diff
@@ -191,15 +191,17 @@
 
 (defn var->lang-var
   [var]
-  (-> var
-      (str "$-LANG")
-      symbol))
+  (when (where/variable? var)
```
zonotope, line 194: why is this check necessary? i don't think this should be called until we've already established that `var` is a variable, unless i'm missing something.

src/clj/fluree/db/query/exec/eval.cljc
```diff
@@ -210,6 +212,99 @@
...
+    #{const/iri-xsd-boolean
+      const/iri-anyURI
+      const/iri-id
+      const/iri-xsd-dateTime
+      const/iri-xsd-date
+      const/iri-xsd-time}))
+
+(defn infer-dt-iri
+  [x]
+  (get {const/$xsd:string  const/iri-string
```
zonotope, line 251: This is probably minor, but it wouldn't hurt to bind this map in a `def` so it doesn't have to be initialized and garbage collected with each invocation of this function.

src/clj/fluree/db/json_ld/iri.cljc
```diff
@@ -14,7 +14,10 @@
 
 (defrecord IRI [^String s]
   Object
-  (toString [_] s))
+  (toString [_] s)
+  Comparable
+  (compareTo [_ o]
```
zonotope, line 19: Should we add an `IRI` type hint to `o` to make sure we only compare iris to each other?

src/clj/fluree/db/query/exec/eval.cljc
dpetran: This is required so that we can compare literals and nested expression results. In those cases we may (and do, there were broken tests before I added this) get numbers passed in, which results in a compile exception as we try to construct symbols like `101$-DATATYPE`.

If we're going to allow composing expressions arbitrarily we need this var-guard somewhere, and this seemed like the best place to put it.

src/clj/fluree/db/json_ld/iri.cljc
dpetran: I thought about it, but after the javadoc for `Comparable` I saw that it defines the method in terms of T:
`Interface Comparable<T>`
[compareTo](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/Comparable.html#compareTo(T))([T](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/Comparable.html) o)

So I figured that I could assume that callers would do it correctly and that if it blows up when called incorrectly it should be fine. Also, when I added the type hint to the signature it throws a syntax error:
```
Syntax error (IllegalArgumentException) compiling deftype* at (src/clj/fluree/db/json_ld/iri.cljc:15:1).
Can't find matching method: compareTo, leave off hints for auto match.
```
So I just thought to leave it off. I can add a hint to the `(str o)` call, though.

zonotope (approved):
🪮

PR #879, opened 2024-08-28
Feature/vector flatrank

bplatz:
This adds flat-rank vector searching, and while this feature is useful, the bigger items here is a syntax/framework to call "virtual graphs" to effectively execute any function with parameters (as JSON-LD) and bind the output of the function to variables (as JSON-LD but with variable bindings).

For example, here is a vector search query against a "virtal graph":
```json
{"@context" {"ex":     "http://example.org/ns/",
             "fidx": "http://ns.flur.ee/index#"},
 "select":   ["?targetSubj", "?x", "?score"],
 "where":    [{"@id":     "?targetSubj",
               "ex:xVec": "?targetVec"},
              ["graph",
               "##Flatrank-Cosine",
               {"fidx:search":  "?targetVec",
                "fidx:property": {"@id": "ex:xVec"},
                "fidx:limit":    10,
                "fidx:result":   {"@id":    "?x",
                                    "fidx:score": "?score"}}],
              ["filter", "(not= ?targetSubj ?x)"]]}
```

Here, we call `virtual graph` where the graph's iri (`"##Flatrank-Cosine"`) is a special one registered as a virtual graph. In this case our virtual graph performs flat-rank vector scoring using cosine similarity.

The graph's where clause acts as the input parameters to the virtual graph "function call", and the `idx-fr:result` represents the result values which can be bound to variables as any normal query could do.

The concept here is the vector index has infinite subjects each representing a different comparison (the parameters) and those subjects have a `result` reference which contain the results based on that subject.

Next step is to implement actual HNSW. Flat-rank is a useful, but slow algorithm for scoring vectors. HNSW will be super fast, but not as accurate.


src/clj/fluree/db/vector/index_graph.cljc
```diff
@@ -0,0 +1,171 @@
...
+           (= iri-limit (prop-iri triple))
+           (assoc acc ::limit (obj-val triple solution))
+
+           (= iri-id (prop-iri triple))
+           (assoc-in acc [::result ::id] (obj-var triple))
+
+           (= iri-score (prop-iri triple))
+           (assoc-in acc [::result ::score] (obj-var triple))
+
+           (= iri-vector (prop-iri triple))
```
dpetran, line 86: Might be worth pulling the result of `(prop-iri triple)` into a `let` to avoid re-evaluating it each time.

src/clj/fluree/db/query/fql/parse.cljc
```diff
@@ -472,10 +472,16 @@
 
 (defmethod parse-pattern :graph
   [[_ graph where] vars context]
-  (let [graph* (or (parse-variable graph)
-                   graph)
-        where* (parse-where-clause where vars context)]
-    [(where/->pattern :graph [graph* where*])]))
+  (let [graph*       (or (parse-variable graph)
+                         graph)
+        graph-iri    (when (string? graph*)
```
dpetran, line 477: Is it possible to have an index graph bound to a variable? Or could it be a namespaced keyword, like `:idx-fr/DotProduct`?

src/clj/fluree/db/vector/index_graph.cljc
```diff
@@ -0,0 +1,171 @@
...
+                            {:status 400
+                             :error  :db/invalid-query} e)))))
+
+(defn extract-vectors
+  [{::keys [property]} db]
+  (let [pid (iri/encode-iri db property)]
+    ;; For now, pulling all matching values from full index
+    ;; once hitting the actual vector index, we'll only need
+    ;; to pull matches out of novelty (if that)
+    (query-range/index-range db :post = [pid])))
```
dpetran, line 104: If this is too large to fit into memory, what will happen?

dpetran (commented):
This looks pretty good to me!

Here's how I understand it working:

1) The user transacts a bunch of vectors into a ledger as values of various predicates. 2) At query time a user can specify the metric they want to score the vectors on by using a special `f-idx-flatrank-ns` IRI in a `:graph` clause.
3) We parse `:graph` patterns that use our special prefix into `:index-graph` patterns
4) When processing the `:index-graph` pattern, we invoke the new `Searcher` protocol
5) `-search` will materialize all of the vectors into memory, score them, and produce new solutions with the score and vector
6) Which can then be composed with other patterns

In the future we may extend it by implementing other scoring metrics.

src/clj/fluree/db/flake/flake_db.cljc
```diff
@@ -393,12 +394,24 @@
...
+      db
+
+      :else
+      (throw (ex-info (str "Unknown graph alias: " alias')
+                      {:status 400 :error :db/invalid-query}))))
 
   (-aliases [_]
     [alias])
 
+  where/Searcher
```
dpetran, line 411: Does `AsyncDb` need to implement this as well?

src/clj/fluree/db/query/fql/parse.cljc
bplatz: It can be a variable or a compact IRI (which would be expanded if so).

src/clj/fluree/db/vector/index_graph.cljc
bplatz: OOM Exception - which is the same thing that any query whose results won't fit into memory today will have happen, even a :limit 10 but with an order-by has the same risk, or aggregates over large results. So it isn't even large results that have this risk.

We've talk about disk spill-overs but right now that isn't something we have.

src/clj/fluree/db/flake/flake_db.cljc
bplatz: I don't think so, but can't fully recall why I thought that any longer. I probably should have made a comment. I can look at it again.

src/clj/fluree/db/vector/index_graph.cljc
bplatz: Yes good point. This is only hit at parse time, so won't have a measurable impact on speed - but should have done this regadless.

bplatz: Addressed in c749929213c37034e4d187e61ed969049197a5d0

test/fluree/db/vector/index_test.clj
```diff
@@ -0,0 +1,192 @@
...
+                                  "@type"  const/iri-vector}
+                                 {"@value" [0.2, 0.9]
+                                  "@type"  const/iri-vector}]}]})]
+
+      (testing "dot product scoring"
+        (let [query   {"@context" {"ex"   "http://example.org/ns/"
+                                   "fidx" iri/f-idx-ns}
+                       "select"   ["?x", "?score", "?vec"]
+                       "where"    [["graph"
+                                    "##Flatrank-DotProduct"
```
zonotope, line 32: What happens when we query datasets? For example, what if we'd like to query multiple dbs at once and we'd like to flat-rank search all of them together inside that query? What about if we'd like to search only the entries in a particular db inside of the broader query?

src/clj/fluree/db/query/exec/where.cljc
```diff
@@ -613,9 +621,18 @@
 
 (defn match-alias
   [ds alias fuel-tracker solution clause error-ch]
-  (-> ds
-      (-activate-alias alias)
-      (match-clause fuel-tracker solution clause error-ch)))
+  (try*
+    (let [alias-ds (-activate-alias ds alias)]
+      (if (virtual-graph? alias)
```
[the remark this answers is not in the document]
zonotope: Ok, I've thought about it. I think we can avoid adding an extra `Searcher` protocol by adding a `finalize` method to the `Matcher` protocol that we would call at the end of the `match-clause` function and pass either the solution or the whole solution channel. 

`finalize` would be a no-op for the current implementations of the `Matcher` protocol, but that method would perform the flat-rank computation for the virtual graph implementation.

The virtual graph implementation would process the search graph on a triple by triple basis using the other methods on the `Matcher` protocol, adding values to the solutions in flight like we process regular graph queries in the system. By the time `finalize` gets called, all the information to perform the computation should be present on the solutions.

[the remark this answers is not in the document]
zonotope: I created a new branch based on this one and opened #926 implementing this idea. I think it's more flexible and avoids creating a new protocol.

dpetran (approved):
:flat_shoe:

PR #881, opened 2024-08-30
Feature/function refactor

dpetran:
In the course of implementing support for polymorphic comparisons and the various rdf functions I had to make some changes to how we handle function evaluation in general. I ended up replacing the IRI records with a more general TypedValue record. The main problem was that there are cases were we need to know the datatype of a value when that datatype cannot be inferred. We can only infer datatypes of:
- boolean
- string
- long
- decimal

If we have a function that returns a datatype not in that set, we couldn't do it. The IRI record was meant to allow inference of IRI types (distinct from strings), but it wasn't a general approach, and we needed a general approach.

This PR converts all of our function implementations to take TypedValues and return TypedValues.

### DIFF
--- a/src/clj/fluree/db/query/exec/eval.cljc
+++ b/src/clj/fluree/db/query/exec/eval.cljc
@@ -633,7 +633,7 @@
 
                     ;; literal
                     (not (sequential? x))
-                    (where/->TypedValue x (datatype/infer x) nil)
+                    (where/->TypedValue x (datatype/infer-iri x) nil)
 
                     :else
                     x))
--- a/src/clj/fluree/db/query/exec/eval.cljc
+++ b/src/clj/fluree/db/query/exec/eval.cljc
@@ -678,4 +678,5 @@
     (eval `(fn [~soln-sym ~var]
              (-> ~soln-sym
                  (assoc (quote ~var) ~var)
-                 ~f)))))
+                 ~f
+                 :value)))))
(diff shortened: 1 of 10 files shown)


dpetran:
Rebased on main.

src/clj/fluree/db/query/exec/eval.cljc
```diff
@@ -25,192 +25,196 @@
 
 (defn sum
   [coll]
-  (reduce + coll))
+  (where/->typed-val (reduce + (mapv :value coll))))
```
bplatz, line 28: For `sum`, `avg`, `median`, etc where we are performing math, I think we should have a try*/catch* in case there is a non-numeric datatype, and output a friendly ex-message of that cause.

It doesn't need to be in here as it wasn't in here before these changes, but just noticing now.

src/clj/fluree/db/query/exec/eval.cljc
```diff
@@ -281,183 +282,277 @@
...
+(defn less-than-or-equal
+  [{a :value a-dt :datatype-iri}
+   {b :value b-dt :datatype-iri}]
+  (where/->typed-val
+    (or (= a b)
+        (neg? (compare* a a-dt b b-dt)))))
+
+(defn greater-than
+  [{a :value a-dt :datatype-iri}
+   {b :value b-dt :datatype-iri}]
```
bplatz, line 299: Destructuring which is quite slow, and some of these functions could be hit thousands of times+ in larger query results. Because in many cases here the destructured value is only used once, I think we should prefer either `get` or could add a  `value` and `datatype` method to `TypedValue`.

Doesn't need to be done here, I think this is good for now to fix the issues... but as a note for future.

bplatz (approved):
Great! Thanks for getting this adjusted so it works as expected!

PR #882, opened 2024-09-01
General Connection Record

zonotope:
This patch remove the specific connection types in favor of a generalized connection record with a configuration system to provide specific features the old connections had. The configuration and system is managed with integrant, and I think we should add validation to connection configs in a later pull request, but I left that out here.

This patch also removes many methods from the connection protocols, but more work needs to be done to disentangle and remove circular dependencies from the code that operates on connections before we can remove the connection protocols all together, and that work will come in a later pull request. 

This patch also adds an index-store to back `FlakeDB` records that combines storage, serialization, and caching so that `FlakeDB` records will eventually no longer need access to connections to function. More work has to be done to disentangle dbs from connections which will come in a later pull request. 

These changes are based on #858, so please review that first if you haven't already.


src/clj/fluree/db/connection/system.cljc
```diff
@@ -0,0 +1,159 @@
...
+            [fluree.db.nameservice.storage-backed :as storage-nameservice]
+            [fluree.db.nameservice.ipns :as ipns-nameservice]
+            [fluree.db.flake.index.storage :as index.storage]
+            #?(:clj  [fluree.db.storage.s3 :as s3-store]
+               :cljs [fluree.db.storage.localstorage :as localstorage-store])
+            [integrant.core :as ig]))
+
+(derive :fluree.storage/file :fluree/content-storage)
+(derive :fluree.storage/file :fluree/byte-storage)
+(derive :fluree.storage/file :fluree/json-archive)
```
dpetran, line 18: What's the semantic distinction between these types of storage? `content-storage` and `byte-storage` I get, but what's a `json-archive`?

src/clj/fluree/db/connection.cljc
```diff
@@ -1,31 +1,28 @@
...
   (-c-read [conn commit-key] "Reads a commit from storage")
   (-c-write [conn ledger-alias commit-data] "Writes a commit to storage")
   (-txn-write [conn ledger-alias txn-data] "Writes a transaction to storage and returns the key. Expects string keys.")
-  (-txn-read [conn txn-key] "Reads a transaction from storage")
-  (-index-file-write [conn ledger-alias idx-type index-data] "Writes an index item to storage")
-  (-index-file-read [conn file-address] "Reads an index item from storage")
-  (-index-file-delete [conn file-address] "Deletes an index item from storage"))
+  (-txn-read [conn txn-key] "Reads a transaction from storage"))
 
 (comment
```
dpetran, line 27: Is this comment still accurate? My understanding is that the Connection ledger state is a map of ledger-alias->channel[Ledger].

dpetran (approved):
Ok, this looks good to me. 

Here's how I understand what's different:

There is now only one Connection (!!!). The connection has these storage keys:
- `:store`, for implementing `iConnection`, reading and writing commits and txns
- `:index-store`, for reading and writing index nodes
- `:primary-ns` and `:aux-nses`, which use a mutable :byte-storage store.

I'm a fan of all the different protocols for storage, but I'm not sure I understand why `:index-store` is separate from `:store`, since I think they are the same in practice (I think, for instance, that both keys point to the same store instance).

A Connection is still a single-store-for-all-ledgers record, you couldn't have one ledger using a MemoryStore and another using a FileStore in the same Connection, I think.

src/clj/fluree/db/connection/system.cljc
```diff
@@ -0,0 +1,159 @@
...
+                                       :primary-ns  (ig/ref :fluree/nameservice)
+                                       :aux-nses    []
+                                       :parallelism parallelism
+                                       :defaults    defaults}})
+
+(defn memory-config
+  [parallelism cache-max-mb defaults]
+  (-> (base-config parallelism cache-max-mb defaults)
+      (assoc :fluree.storage/memory {})
+      (assoc-in [:fluree.nameservice/storage-backed :address-prefix] "fluree:memory://")))
```
dpetran, line 130: This is a personal bugaboo, so feel free to not take it seriously, but I would really love to see the double-slash `//` removed. Tim Berners-Lee calls the decision to add the double slash his [chief regret](https://archive.nytimes.com/bits.blogs.nytimes.com/2009/10/12/the-webs-inventor-regrets-one-small-thing/), and I'd really love to not perpetuate it : )

bplatz: The intention here has always been that this allows us to distinguish a location. it is probably misused a bit here with 'memory' as doesn't really make sense in this situation.

When a // exists, we know a remote location is being specified (e.g. IP/dns address, but it could be a P2P network address or otherwise).

I've looked into multi-addr as an alternative, but it has more complexity than I think we'll ever need and it is something new people have to learn. For that reason I have preferred the // as, like it or not, it is an established convention.

src/clj/fluree/db/connection/system.cljc
zonotope: json-archive corresponds to the `fluree.db.storage/JsonArchive` protocol. It's a read-only protocol with a single method for retrieving json data. It's the other side of the content-addressed store protocol, which only has a method for writing data. most of the storage implementations implement both protocols, but the remote storage implementation only implements JsonArchive because you can't modify data in a remote store; you can only read it.

zonotope:
> I'm a fan of all the different protocols for storage, but I'm not sure I understand why `:index-store` is separate from `:store`, since I think they are the same in practice (I think, for instance, that both keys point to the same store instance).

An index store might be backed by the same underlying storage found under the `:store` key, but it is a different in practice. An index store isn't just storage, it is [a record that encompasses storage, a cache, and a serializer](https://github.com/fluree/db/pull/882/files#diff-232ca7baa6ea70b4fd3cd36191c8fc9bb5edb2c432b9efc1a0acc8098c18f599R19). The system writes index data to/reads index data from an index store, and the store itself handles the serialization and caching. We don't do this for commits, so that's why there are two separate entries.

But, besides all that, I would like to further decouple index storage from commit storage because these are two different types of data that the system uses in different ways, so there shouldn't be a constraint that they use the same storage medium. I could easily imagine a use case where a user would like to store indexes on the local filesystem, but commits on ipfs, for example. 

> A Connection is still a single-store-for-all-ledgers record, you couldn't have one ledger using a MemoryStore and another using a FileStore in the same Connection, I think.

This should also change eventually. I think a user should be able to configure storage on a per-ledger basis. They should even be able to configure multiple storage media on a single ledger, but I'm changing enough right now so I think this specific change should wait until later.

PR #883, opened 2024-09-04
add support for a singular function selector

dpetran:
This allows users to use a singular function selector, just like a singular variable selector or a wildcard selector:
```
{
  "from": "graph",
  "where": {
    "@id": "?s",
    "favNum": "?favNum"
  },
  "select": "(count ?favNum)",
  "groupBy": "?s"
}
```


zonotope (approved):
🦶🏾

PR #886, opened 2024-09-05
Fix/schema datatype

bplatz:
This finishes the move of inferred datatypes for integers, which are now cast to xsd:integer instead of xsd:long per spec. Tests adjusted.

After this change, I noticed `dbproto/-p-prop` shouldn't be used and was actually getting in the way of SHACL validation reporting. I removed this protocol.

Removal of the protocol obliviated the need to track `:datatype` in the `:schema`,s "preds" - so that was removed which also simplified some code and should offer a tiny boost in speed.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -12,7 +12,7 @@
          metosin/malli                   {:mvn/version "0.16.2"}
          nano-id/nano-id                 {:mvn/version "1.1.0"}
          com.fluree/json-ld              {:git/url "https://github.com/fluree/json-ld.git"
-                                          :git/sha "97e5c713e1e9a2c833a509c71c615595ff5fcda0"}
+                                          :git/sha "452efeb51b6b7244de9dbbd461f068d62a33d53c"}
 
          ;; logging
          org.clojure/tools.logging       {:mvn/version "1.3.0"}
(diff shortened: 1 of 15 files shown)


zonotope (commented):
📰

zonotope (approved):

bplatz:
> 🏫
> 
> Do we need a migration for all the new datatype work? I don't think we can migrate commits alone, but if we have original transactions we could reprocess them to get correct datatypes.

There are two migration concerns here - reading existing commits, and then there was a change to the `schema` we now persist in the index root (to eliminate `datatype`). For the latter, no migration needed. When I did that persistent work I anticipated some changes and designed it in a way that would minimize certain types of changes. This falls in that category. Old indexes will continue to load fine even though they have additional data, and the next index will 'drop' that data.

For the former, default data types, I don't expect the change to impact anyone but there is a possibility it could however. If someone had written queries that filtered based on data type, and assumed the default was xsd:float or xsd:long then those queries would have to get updated. Otherwise the only other concern could be if for some reason we output BigDec notation into a commit, however I looked at this and it appears we didn't but it was of slight concern (e.g. instead of `99.99` if it were `99.99M` it would import as BigDec. I think this is covered.

No guarantees, but feel pretty good all should be fine.

PR #887, opened 2024-09-09
Feature/default policy

bplatz:
Removes `default-allow?` policy flag in favor of default policies.

When looking at configuring fluree/server for zero-trust, it introduced the challenge of specifying a config for `default-allow?` on a per ledger basis. This made it clear that this behavior really belonged as part of the policy itself, and such a config shouldn't be needed.

Having the 'default' behavior part of the policy also offered these enhanced capabilities:
1) instead of true/false, any policy query could be defined as the 'default', allowing more granular control.
2) Different users/identities might have different default behavior
3) Default policy behavior is co-resident with other policy behavior
4) Default policies are defined the identical way as property or class policies, so one thing to explain

As we look to have zero-trust mode, this also makes it possible to create a 'root' policy (just a default policy that always returns truthy) and attach it to an identity, presumably the same identity that created the ledger. This is how v2 worked.

Creating a 'root' policy is just like any other policy, except there is *no* `f:onProperty` or `f:onClass` specified (thus, default). The following is the most permissive/simple example of a root policy:
```
{"@id"      "ex:defaultAllowViewModify"
 "@type"    ["f:AccessPolicy" "ex:RootPolicy"]
 "f:action" [{"@id" "f:view"}, {"@id" "f:modify"}]
 "f:query"  {"@type"  "@json"
                    "@value" {}}}
```

To attached an identity to the above policy, just use `f:policyClass` set to the `@type` in the policy define above (`ex:RootPolicy`):
```
{"@id"          <some-root-did-here>
 "f:policyClass" [{"@id" "ex:RootPolicy"}]}
```

While any query `where` condition can be specified for `f:query` - an empty map has no where condition, so it allows `?$this` to just flow through (truthy).

### DIFF
--- a/src/clj/fluree/db/api.cljc
+++ b/src/clj/fluree/db/api.cljc
@@ -231,11 +231,11 @@
    (ledger/-db ledger)))
 
 (defn wrap-policy
-  ([db policy default-allow?]
-   (wrap-policy db policy default-allow? nil))
-  ([db policy default-allow? values-map]
+  ([db policy]
+   (wrap-policy db policy nil))
+  ([db policy values-map]
    (promise-wrap
-    (policy/wrap-policy db policy default-allow? values-map))))
+    (policy/wrap-policy db policy values-map))))
 
 (defn wrap-identity-policy
   "For provided identity, locates specific property f:policyClass on
@@ -244,11 +244,11 @@
 
   With the policy classes, finds all policies containing that class
   declaration."
-  ([db identity default-allow?]
-   (wrap-identity-policy db identity default-allow? nil))
-  ([db identity default-allow? values-map]
+  ([db identity]
+   (wrap-identity-policy db identity nil))
+  ([db identity values-map]
    (promise-wrap
-    (policy/wrap-identity-policy db identity default-allow? values-map))))
+    (policy/wrap-identity-policy db identity values-map))))
 
 (defn dataset
   "Creates a composed dataset from multiple resolved graph databases.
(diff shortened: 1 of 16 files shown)


src/clj/fluree/db/api.cljc
```diff
@@ -244,11 +244,11 @@
 
   With the policy classes, finds all policies containing that class
   declaration."
-  ([db identity default-allow?]
-   (wrap-identity-policy db identity default-allow? nil))
-  ([db identity default-allow? values-map]
+  ([db identity]
+   (wrap-identity-policy db identity nil))
+  ([db identity values-map]
```
dpetran, line 249: I think removing `default-allow?` is a breaking change. What's the best way to manage it? (I'm very much in favor of removing it)

dpetran (commented):
This removes the `default-allow?` flag for creating a policy-wrapped db and requires the user to specify a policy they can use to read or modify the db. It also provides an example policy that provides read-all/modify-all authorization.

One thing I'm not clear on, though, is what "default" means now. By my read we are _removing_ the default and now there is no default, the user must always be explicit. Or is there some fallback mechanism I missed?

src/clj/fluree/db/api.cljc
bplatz: Policy queries were just introduced and we are pre-release so I'm not concerned about it.

bplatz:
> One thing I'm not clear on, though, is what "default" means now. By my read we are _removing_ the default and now there is no default, the user must always be explicit. Or is there some fallback mechanism I missed?

I think of it as default, and the old method was default as well - it was just inflexible. This is now flexible and definable in policies.

Default means it doesn't target a property or a class, so it is for "everything else" where no explicit policy exists.

dpetran (approved):
:closed_lock_with_key:

PR #888, opened 2024-09-10
Remove cyclic namespace dependencies and Connection/Ledger protocols

zonotope:
This patch removes cyclic dependencies in a lot of namespaces involving both the connection and ledger namespaces, which allowed us to also remove the connection protocols, and allows us to use connections as the "entry point" and api for the database so that different components are now agnostic about which components they sit alongside of to manage the system. 

This patch is based on #882, so please review that, and all of the prs that #882 depends on, first.


src/clj/fluree/db/connection.cljc
```diff
@@ -14,16 +16,6 @@
 
 #?(:clj (set! *warn-on-reflection* true))
 
-(defprotocol iConnection
-  (-did [conn] "Returns optional default did map if set at connection level")
-  (-nameservices [conn] "Returns a sequence of all nameservices configured for the connection."))
-
-(defprotocol iStorage
```
dpetran, line 21: :D

dpetran (approved):
This looks great. I love to see the payoff of all the refactoring you've been doing, having everything be dispatched from the Connection is a big step forward.

PR #891, opened 2024-09-13
Feature/policy classes

bplatz:
Adds two features needed for Nexus (and generally useful)
1) Adds `wrap-class-policy` API to the existing `wrap-policy` and `wrap-identity-policy` for enforcing policy on a db. This addition allows you to pass one or more class IRIs to enforce policy, which will query the db for policies of those class(es) and apply them. This feature provides a similar capability to v2's 'roles' option.
2) Allows new policy features to be utilized from the `query-connection` API which Nexus primarily uses. In the 'opts' for a query, you can now include the following keys:
a) `policyClass` - which performs the feature as described in (1) above.
b) `policy` - which allows you to pass in any json-ld policy definitions and will call `wrap-policy` on your behalf
c) `policyValues` - which allows you to pass in a values map that policy enforcement will inject into the policy queries (via the policy query's `values` key).

*note - in the case of `policy` and `policyClass`, json-ld/expand will be called on them utilizing the query's `@context`. In the process added the expansion to the existing `did` option which previously was never attempted to expand.

### DIFF
--- a/src/clj/fluree/db/json_ld/policy.cljc
+++ b/src/clj/fluree/db/json_ld/policy.cljc
@@ -1,7 +1,15 @@
 (ns fluree.db.json-ld.policy
-  (:require [fluree.db.constants :as const]))
+  (:require [clojure.core.async :refer [go <!]]
+            [fluree.db.dbproto :as dbproto]
+            [fluree.db.constants :as const]
+            [fluree.db.util.core :as util]
+            [fluree.db.util.log :as log]))
 
 #?(:clj (set! *warn-on-reflection* true))
+
+(defprotocol Restrictable
+  (wrap-policy [db policy-rules values-map])
+  (root [db]))
 
 (def root-policy-map
   "Base policy (permissions) map that will give access to all flakes."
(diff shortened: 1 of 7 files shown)


src/clj/fluree/db/json_ld/policy.cljc
```diff
@@ -1,8 +1,16 @@
...
+  (:require [clojure.core.async :refer [go <!]]
+            [fluree.db.dbproto :as dbproto]
+            [fluree.db.constants :as const]
+            [fluree.db.util.core :as util]
+            [fluree.db.util.log :as log]))
 
 #?(:clj (set! *warn-on-reflection* true))
 
+(defprotocol Restrictable
+  (wrap-policy [db policy-rules values-map])
```
zonotope, line 11: Nice simplification of this protocol

src/clj/fluree/db/json_ld/policy/rules.cljc
```diff
@@ -140,50 +140,31 @@
...
-      %)
-   query-results))
+(defn validate-values-map
+  [values-map]
+  (or (map? values-map)
+      (throw (ex-info (str "Invalid policy values map. Must be a map. Received: " values-map)
+                      {:status 400
+                       :error  :db/invalid-values-map}))))
+
+(defn expanded?
```
zonotope, line 150: This is fine for now, but I would prefer more robust code paths where we only call this function with data we already know is expanded. That has its own benefits, and this method introduces a dependency on the proprietary `:idx` key and limits our flexibility with respect to expansion going forward. Making the code paths more robust is probably a much larger refactor, so I'm ok with this for now.

zonotope (approved):
🥘

src/clj/fluree/db/json_ld/policy/rules.cljc
bplatz: Good point. I ran into this issue when using with fluree/server so added a quick fix.

I just fixed it up property and removed this.

PR #892, opened 2024-09-13
Fix/policy modify issue

bplatz:
This fixes an issue where if a vew-only policy is applied, modifications could still be allowed.

### DIFF
--- a/src/clj/fluree/db/api/transact.cljc
+++ b/src/clj/fluree/db/api/transact.cljc
@@ -16,10 +16,10 @@
 
 (defn parse-opts
   [expanded-txn opts txn-context]
-  (-> (util/get-first-value expanded-txn const/iri-opts)
-      (merge opts) ;; override opts 'did', 'raw-txn' with credential's if present
-      (util/keywordize-keys)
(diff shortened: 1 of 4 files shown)


aaj3f:
@bplatz -- do we need to update `stage-triples` so that if `transact!` is used w/ `opts.policy` or `opts.policyClass`, we are not only looking for `opts.did` to produce a policy-wrapped db, but we also can use those? 

https://github.com/fluree/db/blob/8e2b07eeb06c3db08fa5f31b4c7facd3697b4dcb/src/clj/fluree/db/api/transact.cljc#L24-L50

I may be wrong, but it appears that `server`, for example, ultimately uses `api/transact!` which seems to use `transact/stage-triples` to produce a `policy-db` for `tx/stage`, but `stage-triples` doesn't seem to account for `opts.policy` or `opts.policyClass`

bplatz:
@aaj3f Per your comment, transact! policy options is now address in this separate PR: https://github.com/fluree/db/pull/895

zonotope (approved):
🐻‍❄️

PR #893, opened 2024-09-16
Enable fuel tracking for SPARQL queries by default

JaceRockman:


dpetran:
Do we always want this? Do we plan to do it for regular FQL queries as well, or history and transact apis?

JaceRockman:
> Do we always want this? Do we plan to do it for regular FQL queries as well, or history and transact apis?

We already do it by default for FQL queries, history, and transact by adding what we need into the opts before sending it to db, but because SPARQL queries don't have query opts, I had to handle that after it reaches db. I did confirm with @jakep36 that we do want this to be the default behavior though.

dpetran:
I don't think we want to hardcode that behavior into `db`, though. I think it would be best if you pass `:meta true` in from `server` (or `cloud-transactor`). If you work off my [opt cleanup pr](https://github.com/fluree/db/pull/900) the db changes necessary to get the `:meta` opt down to `track-query` are already done.

JaceRockman:
> I don't think we want to hardcode that behavior into `db`, though. I think it would be best if you pass `:meta true` in from `server` (or `cloud-transactor`). If you work off my [opt cleanup pr](https://github.com/fluree/db/pull/900) the db changes necessary to get the `:meta` opt down to `track-query` are already done.

This is only for SPARQL queries though, which don't support query opts which is where I would add `:meta true` to. Neither `fluree/query` nor `fluree/query-connection` get the opts past the point where the SPARQL gets converted to fql, so I think it has to be handled on the db side of things as it currently stands.

dpetran:
I've refactored the query api to allow override opts to be passed alongside the query, so `fluree/query` and `fluree/query-connection` both support that now.

JaceRockman:
> I've refactored the query api to allow override opts to be passed alongside the query, so `fluree/query` and `fluree/query-connection` both support that now.

Is that in a PR somewhere? That's not how I'm seeing `query` or `query-connection` behave.

dpetran:
Yeah, you can find it here: https://github.com/fluree/db/pull/900

PR #894, opened 2024-09-19
Remove special assert syntax

bplatz:
This code appears to create new flakes for `@type` values (classes) when being read from commits.

To my knowledge with the latest SID changes, this isn't needed and isn't done with initial transactions. At one point we did need a flake for every IRI, which is what I assume this was put in for.

I believe this would have the effect of a different db state for the same data when originally transacted vs read from commit files - as the latter would include flakes for every class, and the former would not.

After removing this code, all tests pass -  but definitely want more eyeballs in case I'm missing something.

For a large commit, this code does a decent amount of work, so will help performance by removing it as well.

### DIFF
--- a/src/clj/fluree/db/flake/flake_db.cljc
+++ b/src/clj/fluree/db/flake/flake_db.cljc
@@ -123,63 +123,14 @@
            :namespace-codes new-ns-codes
            :max-namespace-code max-namespace-code*)))
 
-(defn enrich-values
-  [id->node values]
-  (mapv (fn [{:keys [id type] :as v-map}]
-          (if id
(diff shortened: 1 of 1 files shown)


zonotope (approved):
This code looks like it's combining all the maps in an assert sequence with the same id in case you have something like `{"@id" "ex:foo", "ex:friend" {"@id" "ex:bar", "ex:age" 27}}`, and `{"@id" "ex:bar", "ex:favColor" "green"}` in the same sequence. i don't think we need this functionality either. since the tests pass, i say we get rid of it and deal with any issues when they come up.

👍🏾

PR #895, opened 2024-09-20
Fix/policy transact

bplatz:
This adds support for the `transact!` API to support the same policy options that were recently added to the `query-connection` API in this PR: https://github.com/fluree/db/pull/891

Now transact! "opts" can include the keys of `policyClass` or `policy` to pass along policy information, in addition to `did` which it already supported.

### DIFF
--- a/src/clj/fluree/db/api/transact.cljc
+++ b/src/clj/fluree/db/api/transact.cljc
@@ -26,28 +26,29 @@
    internal Fluree triples format."
   [db parsed-txn parsed-opts]
   (go-try
-   (let [track-fuel? (or (:maxFuel parsed-opts)
-                         (:meta parsed-opts))
-         identity    (:did parsed-opts)
-         policy-db   (if identity
-                       (<? (policy/wrap-identity-policy db identity nil))
-                       db)]
-     (if track-fuel?
-       (let [start-time #?(:clj (System/nanoTime)
-                           :cljs (util/current-time-millis))
-             fuel-tracker       (fuel/tracker (:maxFuel parsed-opts))]
-         (try*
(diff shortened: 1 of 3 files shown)


bplatz:
@jakep36 this is the transact! support we discussed.

jakep36:
🙏🏼

src/clj/fluree/db/api/transact.cljc
```diff
@@ -70,36 +71,42 @@
...
-                       txn
-                       (q-parse/parse-txn expanded txn-context))
-         parsed-opts (parse-opts expanded opts txn-context)
-         db          (<? (stage-triples (ledger/-db ledger) triples parsed-opts))
-         ;; commit API takes a did-map and parsed context as opts
-         ;; whereas stage API takes a did IRI and unparsed context.
-         ;; Dissoc them until deciding at a later point if they can carry through.
-         cmt-opts    (dissoc parsed-opts :context :did)] ;; possible keys at f.d.ledger.json-ld/enrich-commit-opts
-     (<? (ledger/-commit! ledger db cmt-opts)))))
+    (let [expanded    (if expanded?
```
dpetran, line 74: Isn't `transact-ledger!` an internal api? I suspect we can just always pass an expanded txn in without having to bother with this flag.

dpetran (approved):
I really like consolidating the policy opts handling into one function, that'll simplify things.

src/clj/fluree/db/api/transact.cljc
zonotope: i addressed this in one of my unmerged prs, so i don't think we have to worry about it here

PR #896, opened 2024-09-23
add IFlureeDb/-query impl for AsyncDB

dpetran:
This fixes the error Andrew was seeing when trying to query with an identity policy in place after a server restart.


dpetran:
Fixes https://github.com/fluree/core/issues/126

zonotope (approved):
This protocol will eventually go away entirely. That's why it wasn't implemented when we added AsyncDB records. There were still some circular dependencies lurking when I tried to get rid of this protocol, so I held it off until later. 

I'm ok with this for now to fix the immediate bug.

PR #897, opened 2024-09-25
Fix SPARQL :InlineDataFull Bug

aaj3f:
`parse-term :InlineDataFull` wasn't properly returning its `:values` vector.

A SPARQL statement like
```sparql
    VALUES (?state ?startDate) { 
        ( "New York" "2023-03-01"^^xsd:date ) 
        ( "Florida" "2023-04-01"^^xsd:date ) 
    }
```

Was getting returned as
```
[:values [
     [["?state" "?startDate"]]
     ...
  ]
]
```
instead of
```
[:values [
     ["?state" "?startDate"]
     ...
  ]
]
```

There was a test that was passing (but shouldn't have been). But the only reason it was passing was because it was (a) a VALUES statement within a WHERE and (b) it was only testing for a certain shape and not actually executing the query. As an aside, this false-positive test was also why inline VALUES maps were parsing to FQL correctly but SPARQL queries that used them would return `[]` result sets

### DIFF
--- a/src/clj/fluree/db/query/sparql/translator.cljc
+++ b/src/clj/fluree/db/query/sparql/translator.cljc
@@ -1,8 +1,7 @@
 (ns fluree.db.query.sparql.translator
   (:require #?(:cljs [cljs.tools.reader :refer [read-string]])
             [clojure.string :as str]
-            [fluree.db.constants :as const]
-            [fluree.db.util.log :as log]))
+            [fluree.db.constants :as const]))
 
 (defn rule?
   [x]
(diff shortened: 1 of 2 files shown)


src/clj/fluree/db/query/sparql/translator.cljc
```diff
@@ -1,7 +1,8 @@
 (ns fluree.db.query.sparql.translator
-  (:require [fluree.db.constants :as const]
+  (:require #?(:cljs [cljs.tools.reader :refer [read-string]])
             [clojure.string :as str]
-            #?(:cljs [cljs.tools.reader :refer [read-string]])))
+            [fluree.db.constants :as const]
+            [fluree.db.util.log :as log]))
```
dpetran, line 5: It doesn't look like we're using this `log` namespace.

dpetran (approved):
:sparkle: 

Good catch!

src/clj/fluree/db/query/sparql/translator.cljc
aaj3f: Fixed

PR #898, opened 2024-09-25
Fix/single select aggregates

dpetran:
We were getting some strange behavior when calling `count` in a single select:
```
{:where [{:id '?s :ex/nums '?num}]
 :select '(count ?num)}
```
Turns out we weren't properly checking for non-sequential selectors when deciding when to aggregate and when deciding to modify solutions.

This fixes that.


zonotope (approved):
👣

PR #899, opened 2024-09-26
Allow `compare*` for Non-Numeric `min`/`max`

aaj3f:
@dpetran had updated `compare*` to allow `<`, `>`, etc comparisons between non-numeric-yet-comparator-friendly types like `LocalDate` and `OffsetDateTime`, but `min` and `max` were still using clojure.core defaults, disabling their use as aggregate functions for dates and dateTimes.

This just adds the use of `compare*` to those eval fns.

One small note, it seemed in the following that if the `coll` elements are `{ :value :datatype-iri }` that we should return both, but data that had been added as `xsd:date` or `xsd:dateTime` was returning `LocalDate` and `OffsetDateTime` data types, which I think would be more confusing than useful. As a result, I left the implementation for that commented out, and just returned the `:value` for now...

```clojure
(defn min
  [coll]
  (let [compare-fn (fn [a b]
                     (if (neg? (compare* (:value a) (:datatype-iri a)
                                         (:value b) (:datatype-iri b)))
                       a
                       b))
        ;; I think we should also return the datatype-iri here, 
        ;; but currently for date-like values, that is java.time.LocalDate 
        ;; or java.time.OffsetDateTime, which ought actually to be returned 
        ;; as xsd:date or xsd:dateTime, respectively.
        ;;
        ;;     {:keys [value datatype-iri]} (reduce compare-fn coll)]
        ;; (where/->typed-val value datatype-iri)))
        value      (:value (reduce compare-fn coll))]
    (where/->typed-val value)))
```

### DIFF
--- a/src/clj/fluree/db/query/exec/eval.cljc
+++ b/src/clj/fluree/db/query/exec/eval.cljc
@@ -322,16 +322,8 @@
                      (if (pos? (compare* (:value a) (:datatype-iri a)
                                          (:value b) (:datatype-iri b)))
                        a
-                       b))
-        ;; I think we should also return the datatype-iri here, 
-        ;; but currently for date-like values, that is java.time.LocalDate 
-        ;; or java.time.OffsetDateTime, which ought actually to be returned 
-        ;; as xsd:date or xsd:dateTime, respectively.
-        ;;
-        ;;     {:keys [value datatype-iri]} (reduce compare-fn coll)]
-        ;; (where/->typed-val value datatype-iri)))
-        value      (:value (reduce compare-fn coll))]
-    (where/->typed-val value)))
+                       b))]
+    (reduce compare-fn coll)))
 
 (defn min
   [coll]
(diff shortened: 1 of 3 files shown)


src/clj/fluree/db/query/exec/eval.cljc
```diff
@@ -326,6 +316,40 @@
...
+                       a
+                       b))
+        ;; I think we should also return the datatype-iri here, 
+        ;; but currently for date-like values, that is java.time.LocalDate 
+        ;; or java.time.OffsetDateTime, which ought actually to be returned 
+        ;; as xsd:date or xsd:dateTime, respectively.
+        ;;
+        ;;     {:keys [value datatype-iri]} (reduce compare-fn coll)]
+        ;; (where/->typed-val value datatype-iri)))
+        value      (:value (reduce compare-fn coll))]
```
dpetran, line 350: I think we could just return the value produced by `reduce`, which will be a TypedValue with the correct datatype (and lang, if applicable).

zonotope (approved):
LGTM. @dpetran do you have any thoughts?

dpetran (approved):
:thermometer:

PR #903, opened 2024-10-01
Fix/union parsing

dpetran:
This is a branch that I was working on with Andrew to get subqueries working in SPARQL. Depends on https://github.com/fluree/db/pull/899

### DIFF
--- a/src/clj/fluree/db/query/fql/parse.cljc
+++ b/src/clj/fluree/db/query/fql/parse.cljc
@@ -436,7 +436,10 @@
 (defmethod parse-pattern :union
   [[_ & unions] vars context]
   (let [parsed (mapv (fn [clause]
-                       (parse-where-clause clause vars context))
+                       (if (and (vector? clause)
+                                (= (first clause) :query)) 
+                         (parse-pattern clause vars context)
+                         (parse-where-clause clause vars context)))
                      unions)]
     [(where/->pattern :union parsed)]))
 
--- a/src/clj/fluree/db/query/fql/parse.cljc
+++ b/src/clj/fluree/db/query/fql/parse.cljc
@@ -436,8 +436,7 @@
 (defmethod parse-pattern :union
   [[_ & unions] vars context]
   (let [parsed (mapv (fn [clause]
-                       (if (and (vector? clause)
-                                (= (first clause) :query))
+                       (if (and (vector? clause) (keyword? (first clause)))
                          (parse-pattern clause vars context)
                          (parse-where-clause clause vars context)))
                      unions)]
--- a/src/clj/fluree/db/query/fql/parse.cljc
+++ b/src/clj/fluree/db/query/fql/parse.cljc
@@ -268,14 +268,18 @@
                        [var f]))))
         binds))
 
+(defn higher-order-pattern?
+  "A non-node where pattern."
+  [pattern]
+  (and (sequential? pattern) (keyword? (first pattern))))
+
 (defn parse-where-clause
   [clause vars context]
-  ;; a single non-node where pattern is already sequential, so we need to specially handle it
-  (let [clause* (if (and (sequential? clause) (keyword? (first clause)))
+  ;; a single higher-order where pattern is already sequential, so we need to check if it needs wrapping
+  (let [clause* (if (higher-order-pattern? clause)
                   [clause]
                   (util/sequential clause))]
     (->> clause*
-         util/sequential
          (mapcat (fn [pattern]
                    (parse-pattern pattern vars context)))
          where/->where-clause)))
--- a/src/clj/fluree/db/query/fql/parse.cljc
+++ b/src/clj/fluree/db/query/fql/parse.cljc
@@ -270,11 +270,15 @@
 
 (defn parse-where-clause
   [clause vars context]
-  (->> clause
-       util/sequential
-       (mapcat (fn [pattern]
-                 (parse-pattern pattern vars context)))
-       where/->where-clause))
+  ;; a single non-node where pattern is already sequential, so we need to specially handle it
+  (let [clause* (if (and (sequential? clause) (keyword? (first clause)))
+                  [clause]
+                  (util/sequential clause))]
+    (->> clause*
+         util/sequential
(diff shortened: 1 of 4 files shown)


dpetran:
That branch was merged to main so this PR had its base branch changed to main.

src/clj/fluree/db/query/fql/parse.cljc
```diff
@@ -436,7 +436,10 @@
 (defmethod parse-pattern :union
   [[_ & unions] vars context]
   (let [parsed (mapv (fn [clause]
-                       (parse-where-clause clause vars context))
+                       (if (and (vector? clause)
```
zonotope, line 439: Why couldn't this be built into `parse-where-clause` instead of special casing here? Could we have the same issues with other higher level pattern types like `optional` that operate on nested where clauses?

The only constraint I can think of with this approach is that where clauses consisting of only a single non-map pattern is required to be wrapped in a vector, but I think, if this is the reason, then it should be handled in the sparql translator and not here.

dpetran: The problem is that our `::where` grammar is ambiguous:
```
[:union
  [:query1 ...]
  [:query2 ...]]
```
Could either be two single-pattern clauses or one multi-pattern clause. All of our higher-order patterns indicate that they accept multiple clauses, but looking at the parser we assume for all but `:union` that we are only receiving a single clause.

I think for the short term I'll generalize the condition so that it will work on all higher-order patterns by changing it to this:
```
(and (vector? clause) (keyword? (first clause))) ;; clause is a single pattern
```
But I think we'll need what our actual grammar is - is the parser the source of truth or is the malli schema?

dpetran: Ok, I think I found a better way to handle it. `parse-where-clause` was calling `util/sequential` to ensure that regardless of whether the clause was a single where-pattern or a sequence of where-patterns we would treat it as a sequence. However, all non-node where patterns are already sequential, so this was not wrapping it in a vector. Now we correctly handle all single-pattern types and turn them into a sequence of where-patterns.

zonotope: Union is supposed to handle multiple clauses. A clause can either be a map or a vector. A map is the case of a singleton subject-node map (with possible nested maps in the case where the value of a property is a reference). 

If it is a vector, then it is a vector of where patterns. A single vector is not valid. 

**Single-pattern where clauses need to be wrapped in vectors unless they are node maps**, and this is what [the malli schema](https://github.com/fluree/db/blob/main/src/clj/fluree/db/validation.cljc#L323-L325) should be checking for. 

```
[:union
  [:query1 ...]
  [:query2 ...]]
```
is not valid. It should be:
```
[:union
  [[:query1 ...]]
  [[:query2 ...]]]
```

dpetran: Ok, there's the source of the confusion. I got the test syntax from @bplatz , who also seemed confused that we only supported single node patterns instead of single where patterns of any type.

So, now that we have a way to support single where patterns of any type, do we want that or no?

zonotope: > A single vector is not valid
This is not clear. I meant to say, a single vector pattern is not valid unless it is wrapped in another vector. so you can't have the following inside of a where clause
```
{...
 :where [:optional {:foo/bar ?baz}]
 ...}
```
It must be:
```
{...
 :where [[:optional {:foo/bar ?baz}]]
 ...}
```

The rules for what can come in an optional and union are the same as what can come in a where clause.

zonotope: > Ok, I think I found a better way to handle it. `parse-where-clause` was calling `util/sequential` to ensure that regardless of whether the clause was a single where-pattern or a sequence of where-patterns we would treat it as a sequence. However, all non-node where patterns are already sequential, so this was not wrapping it in a vector. Now we correctly handle all single-pattern types and turn them into a sequence of where-patterns.

I'm not sure why this would have ever happened, unless there was a bug in validation. If it got past the validation step, then any single where patterns that were sequential should have already been wrapped. It must have been that the `:query` pattern type wasn't being validated properly.

zonotope: > Ok, there's the source of the confusion. I got the test syntax from @bplatz , who also seemed confused that we only supported single node patterns instead of single where patterns of any type.
> 
> So, now that we have a way to support single where patterns of any type, do we want that or no?

I had considered checking for a keyword as the first element when I first wrote the query parser but then I decided against it. I just can't remember _why_ I decided against it. 

Since I can't remember, then it probably wasn't that good of a reason not to do it, so let's go with it for now and we can revisit if I remember why I didn't do it in the first place and/or some problems pop up.

src/clj/fluree/db/query/fql/parse.cljc
```diff
@@ -270,11 +270,15 @@
 
 (defn parse-where-clause
   [clause vars context]
-  (->> clause
-       util/sequential
-       (mapcat (fn [pattern]
-                 (parse-pattern pattern vars context)))
-       where/->where-clause))
+  ;; a single non-node where pattern is already sequential, so we need to specially handle it
+  (let [clause* (if (and (sequential? clause) (keyword? (first clause)))
```
zonotope, line 274: This condition should be factored into a predicate function with a name like `higher-order-pattern?`

src/clj/fluree/db/query/fql/parse.cljc
```diff
@@ -270,11 +270,15 @@
...
+                  [clause]
+                  (util/sequential clause))]
+    (->> clause*
+         util/sequential
```
zonotope, line 278: This is no longer necessary since you've ensured that the clause is sequential above.

dpetran: I think you still need it if you've got a single node pattern, since that will fail the `sequential?` check above.

dpetran: Wait, nevermind.

PR #904, opened 2024-10-03
allow value-maps in function expressions

dpetran:
We ran into a problem when translating SPARQL into FQL - RDF tagged literals were being translated into json-ld value maps, but we did not support value maps inside of function expressions.

```
PREFIX ex: <http://example.com/ns#>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
SELECT ?s
FROM <fluree-jld/369435906932736>
WHERE {
    ?s ex:dob ?date .
    FILTER (?date >= "2023-08-01"^^xsd:date && ?date <= "2023-08-31"^^xsd:date)
}
;; => translates to
{:context {"ex" "http://exmpale.com/ns#", "xsd" "http://www.w3.org/2001/XMLSchema#"},
   :select ["?s"],
   :from ["fluree-jld/369435906932736"],
   :where [{"@id" "?s", "ex:dob" "?date"}
           [:filter "(and (>= ?date {\"@value\" \"2023-08-01\", \"@type\" \"xsd:date\"}) (<= ?date {\"@value\" \"2023-08-31\", \"@type\" \"xsd:date\"}))"]]}
```

I started by trying to change the translator to detect datatype-tagged literals and move them to a :values pattern, but it turned out to be much easier, and in retrospect much more flexible, to add support for value maps within expressions.

The only tricky part was handling the `in` function, where we use a vector as a set literal notation.

### DIFF
--- a/src/clj/fluree/db/query/exec/eval.cljc
+++ b/src/clj/fluree/db/query/exec/eval.cljc
@@ -4,12 +4,13 @@
   (:require [fluree.db.query.exec.group :as group]
             [fluree.db.query.exec.where :as where]
             [fluree.db.vector.scoring :as score]
+            [fluree.db.util.core :as util]
             [fluree.db.util.log :as log]
             [fluree.json-ld :as json-ld]
             [fluree.db.json-ld.iri :as iri]
             [clojure.set :as set]
             [clojure.string :as str]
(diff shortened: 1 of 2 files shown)


zonotope (approved):
🪴

PR #905, opened 2024-10-03
add support for functions as data

dpetran:
When a user needs to write a function as a string all literal strings within the expression need to be escaped, which is difficult to read and prone to frustrating errors.

example:
```
  {"where" [{"@id" "?s" "ex:name" "?name"}
            ["filter" "(strStarts ?name \"B\")"]]
   ...}
```

This allows users who need to use json to express their functions as nested arrays:

```
  {"where" [{"@id" "?s" "ex:name" "?name"}
            ["filter" ["strStarts" "?name" "B"]]
   ...}
```
I tossed this off because I've been bothered by this for ages and I recently had to do a lot of work with a function like this:
```
"(and (>= ?date {\"@value\" \"2023-08-01\", \"@type\" \"xsd:date\"}) (<= ?date {\"@value\" \"2023-08-31\", \"@type\" \"xsd:date\"}))"
```
Which looks a lot nicer written like this:
```
["and"
  [">=" "?date" {"@value" "2023-08-01" "@type" "xsd:date"}]
  ["<=" "?date" {"@value" "2023-08-31" "@type" "xsd:date"}]]
```

### DIFF
--- a/src/clj/fluree/db/validation.cljc
+++ b/src/clj/fluree/db/validation.cljc
@@ -249,7 +249,7 @@
     ::function             [:orn
                             [:string-fn [:and :string [:re #"^\(.+\)$"]]]
                             [:list-fn [:and list? [:cat :symbol [:* any?]]]]
-                            [:data-fn [:and vector? [:cat [:enum :expr] [:* any?]]]]]
+                            [:vector-fn [:and vector? [:cat [:enum :expr] [:* any?]]]]]
     ::as-function          [:orn {:error/message "subquery aggregates must be bound to a variable with 'as' e.g. '(as (sum ?x) ?x-sum)"}
                             [:string-fn [:and :string [:re #"^\(as .+\)$"]]]
                             [:list-fn [:and list? [:cat :symbol [:* any?]]]]]
--- a/src/clj/fluree/db/validation.cljc
+++ b/src/clj/fluree/db/validation.cljc
@@ -249,7 +249,7 @@
     ::function             [:orn
                             [:string-fn [:and :string [:re #"^\(.+\)$"]]]
                             [:list-fn [:and list? [:cat :symbol [:* any?]]]]
-                            [:data-fn [:and vector? [:cat :string [:* any?]]]]]
+                            [:data-fn [:and vector? [:cat [:enum :expr] [:* any?]]]]]
     ::as-function          [:orn {:error/message "subquery aggregates must be bound to a variable with 'as' e.g. '(as (sum ?x) ?x-sum)"}
                             [:string-fn [:and :string [:re #"^\(as .+\)$"]]]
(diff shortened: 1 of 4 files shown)


bplatz:
My first reaction is very excited for this. I could see a hiccup-like syntax support for the CLJ community as well - although we do allow them to current pass symbols, but the escaping still needs to be done.

I'll be curious what @zonotope thinks of this.

zonotope:
I don't have a problem with this change under the constraints of JSON. 

The one thing that gives me pause is that `[]` now means two different things in fql (both functions and higher order where patterns). We could run into trouble if we ever have a function with the same name as a where pattern, and deeply nested functions within higher order where patterns like bind or union might start looking like a sea of `[]` that could be hard to decipher.

I've never had a problem with escaping quotes inside strings, but maybe I just haven't needed to do it enough for it to be annoying. Emacs completes the escaping for me automatically when i do the opening `\"`, so I don't really think about it most of the time. It's unreasonable of us to require that our users use Emacs with my specific configuration, (even though they should 😉 )

For clojure users specifically, I think passing symbols in lists would be superior to an edn syntax because it differentiates it from higher order where patterns and is more familiar, but json isn't as expressive as edn, so compromises in json might make sense in some circumstances.

dpetran:
In my first implementation I actually forced the expression to be tagged, like so:
```
{:where [[:filter [:f/expr ["strStarts" "?foo" "B"]]]] ...}
```
It actually makes parsing easier. In the end, though, the parse tree is not ambiguous about what the meaning was so I was able to safely discard it. I could switch it back to that if we're concerned about future conflicts.

zonotope:
> For clojure users specifically, I think passing symbols in lists would be superior to an edn syntax ...

I meant to say that passing symbols in lists would be superior to a **hiccup** syntax ...

dpetran:
I've added back the ["expr"...] tag for data fns. The reason I took it out originally is that there is in fact no ambiguity syntactically speaking, so it's just extra typing.

src/clj/fluree/db/validation.cljc
```diff
@@ -248,7 +248,8 @@
     ::subject              ::iri
     ::function             [:orn
                             [:string-fn [:and :string [:re #"^\(.+\)$"]]]
-                            [:list-fn [:and list? [:cat :symbol [:* any?]]]]]
+                            [:list-fn [:and list? [:cat :symbol [:* any?]]]]
+                            [:data-fn [:and vector? [:cat [:enum :expr] [:* any?]]]]]
```
zonotope, line 252: tiny nitpick, but lists are also data. perhaps this could be called `:vector-fn` or `:expr-pattern`.

zonotope (approved):
This looks good to me besides the ambiguous usage of `:data-*`.

PR #908, opened 2024-10-09
Fix/compare non serializable data literals

dpetran:
There were two issues preventing the comparison of dates:

1) If the user had `f` as a prefix in their context, the expansion wouldn't pick up the expanded value map. I fixed this by using an already expanded iri.

2) value maps in expressions don't have their values coerced into their respective datatypes. Most of the time this doesn't matter but it does matter when comparing values. 

I had hoped to do it at compile time because that would be most efficient, but that doesn't work for some data types. A Date, for example, is represented by an opaque, non-serializable `java.time.Date` object and `eval` does not work on it. So it has to happen while making the comparison.


dpetran:
Fixes https://github.com/fluree/core/issues/124

dpetran:
I added a new, distinct equality semantic that takes datatypes into consideration and will throw an error if you try to compare incomparable types.

zonotope:
Could you introduce new [tagged literals](https://clojure.org/reference/reader#tagged_literals) in `data_readers.clj` to (de)serialize the date objects you need?

dpetran:
I added tagged literals for time datatypes and it should be a big performance win. Now each literal gets coerced once at function compilation time instead of on every invocation. EDN is brilliant : )

zonotope:
> I added tagged literals for time datatypes and it should be a big performance win. Now each literal gets coerced once at function compilation time instead of on every invocation. EDN is brilliant : )

This looks good, but I would like to confirm that the library you pulled in doesn't clobber our own [`data_readers.cljc`](https://github.com/fluree/db/blob/main/src/clj/data_readers.cljc) file because we've defined some fluree specific tagged literals there. 

I think it suffices to start up the repl and see if something like `#fluree/Flake [s p o t dt op m]` still works.

dpetran:
Our tagged flakes still work the same as before, just verified.

zonotope (approved):
🎩

PR #909, opened 2024-10-10
Use override opts in queries

JaceRockman:


src/clj/fluree/db/query/api.cljc
```diff
@@ -35,7 +35,8 @@
   [query {:keys [identity did issuer] :as override-opts}]
   (update query :opts (fn [{:keys [max-fuel meta] :as opts}]
                         ;; ensure :max-fuel key is present
-                        (-> (assoc opts :max-fuel max-fuel)
+                        (-> (merge override-opts opts)
```
dpetran, line 38: Good catch! If we want the override opts to override we'll have to swap the order of the merge:
`(merge opts override-opts)`

src/clj/fluree/db/query/api.cljc
```diff
@@ -35,7 +35,8 @@
   [query {:keys [identity did issuer] :as override-opts}]
   (update query :opts (fn [{:keys [max-fuel meta] :as opts}]
                         ;; ensure :max-fuel key is present
-                        (-> (assoc opts :max-fuel max-fuel)
+                        (-> (merge opts override-opts)
+                            (assoc :max-fuel max-fuel)
```
dpetran, line 39: Oh, I just realized that the `(assoc :max-fuel max-fuel)` is using the `:max-fuel` from `opts`, so you'll be clobbering the `override-opts`. I think we can simplify the whole thread form:

```
(-> (assoc opts :max-fuel max-fuel)
      (merge opts override-opts)
      ;; get rid of :did, :issuer opts
      (update :identity #(or % (:did opts) (:issuer opts)))
      (dissoc :did :issuer))

JaceRockman: Ah, that makes sense, good catch!

dpetran (approved):
:thread:

PR #910, opened 2024-10-11
Other fixes based on the kebab-case API parsing

bplatz:
This addresses remaining API issues that stemmed from the changes in #900 and also relates to #909


dpetran (approved):
:safety_pin:

PR #911, opened 2024-10-11
fix reify-schema version check

dpetran:
I'm a little confused how this scooted by undetected for so long, but I think this condition was reversed. And it was also not forward compatible.

This manifested as the db :schema not having a :shapes cache, which causes validation to blow up.


dpetran:
Fixes https://github.com/fluree/core/issues/133

PR #915, opened 2024-10-21
publish genesis commit on ledger creation

dpetran:
Now a ledger will `exist?` once created.

### DIFF
--- a/src/clj/fluree/db/ledger.cljc
+++ b/src/clj/fluree/db/ledger.cljc
@@ -139,24 +139,6 @@
    :branch   current-branch
    :graphs   {}})
 
-(defn instantiate
-  "Creates a new ledger, optionally bootstraps it as permissioned or with default
-  context."
-  [conn ledger-alias ledger-address branch commit-catalog index-catalog
-   indexing-opts did latest-commit]
-  (let [branches {branch (branch/state-map ledger-alias branch commit-catalog index-catalog
-                                           latest-commit indexing-opts)}]
-    (map->Ledger {:conn                 conn
-                  :id                   (random-uuid)
-                  :did                  did
-                  :state                (atom (initial-state branches branch))
-                  :alias                ledger-alias
-                  :address              ledger-address
-                  :commit-catalog       commit-catalog
-                  :index-catalog        index-catalog
-                  :cache                (atom {})
-                  :reasoner             #{}})))
-
 (defn normalize-alias
   "For a ledger alias, removes any preceding '/' or '#' if exists."
   [ledger-alias]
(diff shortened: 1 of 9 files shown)


src/clj/fluree/db/ledger.cljc
```diff
@@ -163,6 +165,16 @@
     (subs ledger-alias 1)
     ledger-alias))
 
+(defn publish-commit
```
zonotope, line 168: This function operates on a connection, so it should be in the connection namespace. Ledgers shouldn't reach in to connections, and they should not know anything about name services, or anything else attached to a connection besides the ledger itself. Publishing commits is not and should not be the concern of ledger objects.

src/clj/fluree/db/storage.cljc
```diff
@@ -48,7 +48,8 @@
 (defn split-address
   "Splits `address` into the fully qualified storage method and local path."
   [address]
-  (str/split address #":(?!.*:)" 2))
+  (let [components (str/split address #":")]
+    [(str/join ":" (butlast components)) (last components)]))
```
zonotope, line 52: was there a bug how it was previously? this looks like it's doing a bunch of extra work to split the entire string only to immediately join it again.

src/clj/fluree/db/ledger.cljc
dpetran: Good catch, I've refactored to do the correct thing instead of the expedient thing :+1:

src/clj/fluree/db/storage.cljc
dpetran: Yes, I've detailed the problem in the commit message of [ab50edc](https://github.com/fluree/db/pull/915/commits/ab50edcb7715b10dcc5c62e5b733bc3075fbfc9d). I've update this implementation to be more efficient.

bplatz:
I still have a philosophical issue with the genesis block.

We've tried to make dbs 'content addressed', and therefore without content, there is nothing to address (except commit metadata - but that isn't a db).

Aliases should be mappable to any "db" - but that is controlled by the nameservice(s) you are using - which might also be your own local nameservice (just like you can locally modify DNS with `/etc/hosts`)

Granted we are working with what we've got, so my input here might be idealistic and less practical for right now, but my preference would be to:
a) register/unregister/update aliases on the nameservice. The alias can be part of the commit, but is just a 'suggestion'.
b) There is no genesis commit, only a first commit of data which is content-addressed

zonotope:
> I still have a philosophical issue with the genesis block.
> 
> We've tried to make dbs 'content addressed', and therefore without content, there is nothing to address (except commit metadata - but that isn't a db).
> 
> Aliases should be mappable to any "db" - but that is controlled by the nameservice(s) you are using - which might also be your own local nameservice (just like you can locally modify DNS with `/etc/hosts`)
> 
> Granted we are working with what we've got, so my input here might be idealistic and less practical for right now, but my preference would be to: a) register/unregister/update aliases on the nameservice. The alias can be part of the commit, but is just a 'suggestion'. b) There is no genesis commit, only a first commit of data which is content-addressed

I think in order for this to work, we'd have to remove the alias and branch from the commit document and instead associate it with a name service. I think we'd also need to introduce hierarchical name services too as right now name services are flat, and a connection round-robin checks each one. So, instead of having a local file name service sitting along side an ipfs name service that sits lower in the priority (for example) we would have a file name service that referenced the ipfs name service and looked up addresses after doing whatever remapping it would need to do. 

I don't think I disagree with any of this, but I'm just trying to work through the implications and what we'd need to do to make it work. In any case, I don't think we should keep what we have now because the `create-with-transaction` flow is both confusing for users and has lead to bugs in implementation (you can `create` a ledger with the db library but not when using that db library in a stateless context like in fluree/server, so an error during the first create could lead to other errors about a phantom ledger already existing).

bplatz:
> I would like to make a case for it, though, on the philosophical level : )

I agree with the case - I disagree on how we enable it is all. I think how you create a ledger is you reserve the name in the desired authoritative nameservice. Now nobody else can use that name or create it using that name, and presumably nobody else can publish commit updates to that name that don't have authorization.

We talked about the nameservice being managed via a "system" Fluree ledger... this could also open up all the goodness around policy enforcement and time travel that Fluree offers. The Fluree ledger then also has the controls for guaranteed ordering.

Ultimately that's what I'd like to see... and it not only solves this problem it solves other problems.

While unlikely, I also think your proposed approach doesn't have thread safety... whereas a Fluree ledger solves that for you. If two parties request the identical ledger name at the identical moment in time both would incorrectly succeed I believe - unless those are running through consensus (so this wouldn't be a problem if using Raft).

Ben and I see this problem a little differently I think - so if he and you agree then I'm happy to have the proposed changes you suggest. I think this other approach however will create a better system - we just don't have all the pieces in place for it yet.

dpetran:
Ok, that clarifies a lot, thanks. It'll take some thought to figure out the path from here to there, but I don't think I object to your approach.

dpetran:
Closing for now, cljs fixes here: https://github.com/fluree/db/pull/977

PR #921, opened 2024-10-29
JSON-LD connection configs

zonotope:
This patch brings the json-ld configuration system from server into the db library so that users can now initialize connections using json-ld based configurations. This allows the maximum flexibility in which components make up a connection. I also left in place the `connect-file` and `connect-memory` api for the simple cases where there's only one storage type on the connection. 

I also made some changes to the tests to make them less brittle by eliminating direct comparison of string error messages and specific blank node ids in the shacl tests.

In fixing the tests, I identified a bug in the history queries where retracted flakes were never returned. The tests passed because the `pred-match?` test helper did not ensure that collections that should match had the same number of elements, so if one of the collections was empty and the other was not, the test would implicitly pass. I fixed that issue and marked the tests that failed as a result pending. I will write a bug issue to either fix the history queries or update the tests.


test/fluree/db/shacl/shacl_basic_test.clj
```diff
@@ -66,30 +62,28 @@
...
-                 (ex-data db-no-names)))
-          (is (= "Subject :ex/john path [:schema/name] violates constraint :sh/minCount of shape _:fdb-2 - count 0 is less than minimum count of 1."
-                 (ex-message db-no-names)))))
+                             db
+                             {"@context" ["https://ns.flur.ee" context]
+                              "insert"
+                              {:id              :ex/john
+                               :type            :ex/User
+                               :schema/callSign "j-rock"}})]
+          (is (pred-match? {:status 422,
```
dpetran, line 71: I object to removing the error message tests and using `pred-match?` for the error data tests. The error message tests are important, as is confirming that the correct shape is identified in the error data. 

As a developer, the `pred-match?` predicate gives awful test errors that make it difficult to tell what the actual problem is, and it is prone to producing false positive results as you've noted in your PR message.

I would much rather just use defined ids for the shapes if that is the problem than throw away these tests. I'm happy to help convert as I know it's a tedious task, but I think it's a better solution than relaxing our criteria.

zonotope: I think the error message tests were too brittle, and are the wrong specificity for integration tests like these. I think in these high level integration tests, we should be checking that error cases do indeed return errors while cases that should succeed do. 

We have code that generates error messages, and I'm not saying that we shouldn't test that code; I just don't think we should test it here. We should have unit tests that check that the different types of error messages are properly generated given a specific error report. Separating the tests like that means that we're free to change our error messages, or allow for the normal drift of things like ids, without needing to update hundreds of tests each time. 

I would support adding more unit tests to test the error message generation code specifically, and leave integration tests for the more general behavior of the system as a whole.

zonotope: I also agree with you about `pred-match?`, but I think it's a useful tool for high level integration tests checking for things that contain hashes, timestamps, and ids that will change. I think there are things that we can do to improve its error reporting, like logging out which pieces don't match (predicate function or value) instead of printing the whole structure and, of course, fixing any bugs in it.

I think it's a much better option than `with-redefs` which is, in my opinion, a code smell, or hoping that your input data will never change so things like hashes will always be the same. I think we should test the code as close to how it is run in practice, so something like `pred-match?` enables that without making the tests overly brittle.

zonotope: I think we still have the brittleness because we are testing all the errors. If we change the error text at all, or something similar to what just happened occurs, we'd have to turn around and update all the tests again. 

I think integration tests should be for making sure the different components work together, and they should be general enough to test interactions. More specific behavior should be reserved for unit tests. 

I think (but haven't looked into it deeply enough to be certain so I might be wrong) that you could get away with the same coverage in unit tests for the error message generation because you're testing fewer things. I remember there being multiple assertions where the error message had the same structure that could be covered with one assertion in a unit test.

dpetran: I'm fine with that, but my concern is when we'll find the time to re-add this coverage. I'm hesitant to remove the coverage we have because adding it back will always be lower priority than other work, at least until we git bitten by it : /

dpetran: I threw together this PR: https://github.com/fluree/db/pull/928

I don't want to hold up this PR, which besides this testing issue looks good, so I'll approve it and defer to your judgment here. If you care to use it I would be glad, but I'll defer to you - feel free to merge it if you like it.

dpetran (approved):
:chains:

PR #922, opened 2024-10-29
Add `Catalog` implementations of storage protocols

zonotope:
This patch adds catalog implementations for the compatible storage protocols so that `Catalog` records can be used more interchangeably with the more primitive storage records.


dpetran (approved):
:black_cat:

Issue #918, opened 2024-10-25
Unicode Chars in @id field of TX's cause server hang!!


alan-icebreaker:
Full Discord Thread: https://discord.com/channels/896089675511508992/896089675511508995/1299186180210036747

Note - I originally thought this was related to GC, or the `Refreshing Index` stage, but that was a red herring.

TL;DR

A transaction inserting an @id which contains a unicode character (confirmed with ஃ, •) causes fluree to hang, sometimes indefinitely, but usually for many minutes, preventing any further TX's from processing.

I've sometimes had to just restart the server to restore it to normal state. 

Here's the smallest TX I've been able to craft which triggers the issue...

```
{
  "@context": {
    "user": "https://lod.icebreaker.xyz/schemas/user/1.0/"
  },
  "ledger": "icebreaker/core",
  "insert": [
    {
      "@id": "user:ஃ", 
      "@type": "user:Foo"     
    }
  ]
}
```

alan-icebreaker:
A few additional points...

1. While the server is processing this bad TX, all subsequent TX's will timeout (as noted above)
2. I've seen hangs of a few minutes to almost indefinite hanging (IIRC).
3. The logs in the server don't indicate anything wrong with _this_ bad TX, so its difficult to know anything's hung
4. w.r.t. unicode in @id's - I know unicode is legal inside IRI's, but not URL's - So I guess the question is what does fluree support for @id's - URL's or IRI's?

bplatz:
Hi @alan-icebreaker - this issue has been isolated a bit more in #924 - thanks for reporting it.

To recap, unicode chars are allowed. There is an issue when a unicode character is used as the first character after the namespace in an IRI - but other situations with unicode characters seem to be fine.

We'll look to address this issue, but if possible for now just make sure unicode isn't the first character in your IRI "name" portion - appears anywhere else it works fine.

I'll close this issue, and it is linked to the more specific issue created.

PR #923, opened 2024-10-31
Test for unicode chars in @id iri

bplatz:
This adds a unicode IRI test along with a commented out test that bricks the db. I spent some time trying to identify the issue, but it will take more investigation.

Unicode is supported in iris, however this shows it doesn't work when a special character is the first char, only when it is a later character.

This works: 
{"@id": "ex:aஃ", ...}

This bricks the db:
{"@id": "ex:ஃb", ...}

I've written up a ticket here: 

This exploration was an effort to fix this issue here: https://github.com/fluree/db/issues/918

### DIFF
--- a/test/fluree/db/transact/transact_test.clj
+++ b/test/fluree/db/transact/transact_test.clj
@@ -547,3 +547,38 @@
     (is (= "Subject ex:letti path [\"schema:age\"] violates constraint sh:datatype of shape ex:PropertyShape/age - the following values do not have expected datatype xsd:integer: alot."
            (ex-message db3))
         "datatype constraint is restored after a load")))
+
+(deftest ^:integration ^:json transaction-iri-special-char
+  (testing "transaction with special iri characters in @id"
+    (let [
+          conn      @(fluree/connect {:method :memory})
+          ledger-id "transaction-iri-special-char"
+          ledger    @(fluree/create conn ledger-id)
+          db0       (fluree/db ledger)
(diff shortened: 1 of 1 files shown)


dpetran (approved):
:mag:

test/fluree/db/transact/transact_test.clj
```diff
@@ -547,3 +547,38 @@
...
+                     "select"   {"ex:aஃ" ["*"]}}
+          q1b       {"@context" {"ex" "http://example.org/"}
+                     "from"     ledger-id
+                     "select"   {"ex:ஃb" ["*"]}}]
+      (is (= [{"@id"     "ex:aஃ",
+               "@type"   "ex:Foo"
+               "ex:desc" "try special ஃ as second iri char"}]
+             @(fluree/query db1a q1a)))
+      ;; TODO - below test should pass once unicode as first IRI char supported above
+      #_(is (= [{"@id"     "ex:ஃb",
```
zonotope, line 580: I think we should mark this test as pending by adding the `^:kaocha/pending` metadata to the `deftest` form instead of commenting out the assertion and setup code. That way this tests will show up in the stats report after each test run so we don't forget about it, and it also will be included if we try to run just the pending tests using the tooling provided by the test runner.

bplatz: Done. Did it other way originally as the one test was valid and didn't think it should be split into two tests.

PR #925, opened 2024-10-31
fix inline values patterns when there is a blank solution

dpetran:
Fixes https://github.com/fluree/core/issues/137


zonotope (approved):
🦘

PR #926, opened 2024-11-03
Implement vector flat ranking in terms of the `Matcher` protocol

zonotope:
This patch changes #879 by adding a `-finalize` method to the `fluree.db.query.exec.where/Matcher` protocol which is called after all of the patterns in a clause is matched to complete any remaining computation or clean up any left over state. Then vector flat ranking is implemented in terms of the `Matcher` protocol by creating three new index graph records that wrap an underlying flake db and add comparisons based on the 3 metrics.

### DIFF
--- a/src/clj/fluree/db/async_db.cljc
+++ b/src/clj/fluree/db/async_db.cljc
@@ -73,7 +73,7 @@
   (-aliases [_]
     [alias])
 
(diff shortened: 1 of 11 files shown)


bplatz (approved):
This works well and was able to build on it for BM25.

PR #927, opened 2024-11-04
large txn performance improvements

bplatz:
This gives a significant boost in transaction times, although significant increases are most noticable with larger transactions.

This improved transaction time by > 30% in a transaction containing 1 million flakes / 15MB of data.

Txn time went from ~28 seconds to ~19 seconds.

The improvements are achieved by running some operations in parallel on the JVM that were coded as serial.

Notable:
1) Calculating flake-size (for db size) in commit data is much slower than I expected. Not sure what the bottleneck is, didn't investigate - but of original 28 seconds, this operation alone took 6 seconds. This is done in a background thread now.
2) post, opst updates takes about 4x longer than spot, tspo indexes. I think this makes sense because there are large numbers of the same 'p' in most data sets, so the first equivalency tests still will have a large number of secondary items to check. These two indexes only are updated in the background... spot & tspo will both finish before post, so left those as-is for now.

### DIFF
--- a/src/clj/fluree/db/json_ld/commit_data.cljc
+++ b/src/clj/fluree/db/json_ld/commit_data.cljc
@@ -395,34 +395,47 @@
   [flakes]
   (filter ref? flakes))
 
+;; TODO - flakes-size takes considerable time for lg txns, see if can be optimized
+(defn calc-flake-size
+  [add rem]
+  (cond-> 0
+          add (+ (flake/size-bytes add))
+          rem (- (flake/size-bytes rem))))
+
 (defn update-novelty
   ([db add]
    (update-novelty db add []))
 
   ([db add rem]
    (try*
-    (let [ref-add     (ref-flakes add)
-          ref-rem     (ref-flakes rem)
-          flake-count (cond-> 0
-                              add (+ (count add))
-                              rem (- (count rem)))
-          flake-size  (cond-> 0
-                              add (+ (flake/size-bytes add))
-                              rem (- (flake/size-bytes rem)))]
-      (-> db
-          (update-in [:novelty :spot] flake/revise add rem)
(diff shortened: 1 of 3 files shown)


src/clj/fluree/db/json_ld/commit_data.cljc
```diff
@@ -395,34 +395,47 @@
...
+                               rem (- (count rem)))
+           ;; launch futures for parallellism on JVM
+           flake-size  #?(:clj  (future (calc-flake-size add rem))
+                          :cljs (calc-flake-size add rem))
+           post        #?(:clj  (future (flake/revise (get-in db [:novelty :post]) add rem))
+                          :cljs (flake/revise (get-in db [:novelty :post]) add rem))
+           opst        #?(:clj  (future (flake/revise (get-in db [:novelty :opst]) (ref-flakes add) (ref-flakes rem)))
+                          :cljs (flake/revise (get-in db [:novelty :opst]) (ref-flakes add) (ref-flakes rem)))]
+       (-> db
+           (update-in [:novelty :spot] flake/revise add rem)
```
zonotope, line 422: Does it hurt anything to do spot and tspo in the background as well? While unlikely, there could be a situation where there are lots of properties on a single subject updated at once that might make updating novelty for these indexes slow.

bplatz: There is some overhead in spawning a new thread, but small. I think it is probably in its most optimal state for almost all datasets so I decided to stop here.

It wouldn't be a huge hit to put them all in threads however if desired. The dataset I was testing already had some "s-p" pairs that were ~1,000 flakes however, and still saw this performance difference at this level. They were mostly ref flakes however. I think the circumstance that could push it would be very small "s-p" pairs, but a large number of flakes, where the 'o' consisted of mostly strings.

PR #928, opened 2024-11-05
Task/shacl explicit id tests

dpetran:
This PR does the tedious work of changing the SHACL tests to use explicit node identifiers to preserve the coverage and legibility of the tests without the brittleness of depending on either deterministic blank node ids or broad `pred-match?` style tests.


zonotope (approved):
I'll take this pr because it maintains our coverage of the error messages whre we would have lost it.

I think we should convert the tests assertions checking the explicit text in the error message to unit tests gradually and eventually because it will result in fewer tests with the same amount of coverage and will allow us to more easily change how we generate error messages without needing to update hundreds of tests. 

I think that in the future we should use more unit tests to check for specific functionality like whether an error message is generated correctly on a specific class of error conditions, and reserve integration tests to check that the high level components of the system work together correctly.

PR #929, opened 2024-11-05
Feature/query perf

bplatz:
This aims to improve cold query performance by reducing work done needed to generate a cached index node for a given 't'. Once cached, query performance is fast.

There is some more to be had, but wanted to get this smaller piece out since it is ready. This achieves a ~20% improvement in speed for cold queries.

In a test with ~1 million flakes in novelty after a large transaction, a simple cold query takes 6.6 seconds, whereas the same query once cached takes 8ms. This aims to reduce the 6.6 seconds, and with the following changes the query now takes 5.2 seconds.

This should also improve new index creation by a similar amount, however I didn't test it. I also noted this improved transaction times by a small percentage.

Changes:
- novelty-t arg - `novelty-subrange` filters out any flake values later than the requested 't' - however in most cases the novelty is at the same 't' as the requested 't', in which case the filter work is unnecessary. By passing in the db's t (and therefore novelty's t) we can check if these are equal and if so, avoid doing that work. This reduces cold queries by 15%.
- flake/conj-all -> into - conj-all performed the same function as into, which also uses transients. This resulted in negligible, but tiny improvement in speed (~1 ms for larger novelty).

### DIFF
--- a/src/clj/fluree/db/flake.cljc
+++ b/src/clj/fluree/db/flake.cljc
@@ -471,19 +471,6 @@
   [comparator & entries]
   (apply avl/sorted-map-by comparator entries))
 
-(defn conj-all
-  "Adds all flakes in the `to-add` collection from the AVL-backed sorted flake set
-  `sorted-set`. This function uses transients for intermediate set values for
-  better performance because of the slower batched update performance of
-  AVL-backed sorted sets."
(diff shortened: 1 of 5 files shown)


zonotope (approved):
🏉

PR #930, opened 2024-11-05
use subquery :values to form initial solution

dpetran:
We were explicitly supplying an empty initial solution instead of allowing the query executor to correctly derive the initial subquery solution from the subquery values clause.

Fixes https://github.com/fluree/core/issues/139. Because the subquery values clause wasn't being used to populate the initial subquery solution, the variables used in the query were bound to `nil`, which caused a incomparable datatypes error when executed.


zonotope (approved):
🐩

PR #935, opened 2024-11-08
add support for backpressure when novelty-max-bytes is exceeded

dpetran:
Fixes https://github.com/fluree/core/issues/130

### DIFF
--- a/src/clj/fluree/db/flake/index/novelty.cljc
+++ b/src/clj/fluree/db/flake/index/novelty.cljc
@@ -27,11 +27,12 @@
   [children]
   (> (count children) *overflow-children*))
 
-(defn novelty-min?
+(defn min-novelty?
   "Returns true if ledger is beyond novelty-min threshold."
-  [db reindex-min-bytes]
-  (let [novelty-size (get-in db [:novelty :size])]
-    (> novelty-size reindex-min-bytes)))
+  [db]
+  (let [novelty-size (get-in db [:novelty :size])
+        min-novelty  (:reindex-min-bytes db)]
+    (> novelty-size min-novelty)))
 
 (defn max-novelty?
   "Returns true if ledger is beyond novelty-max threshold."
--- a/src/clj/fluree/db/flake/index/novelty.cljc
+++ b/src/clj/fluree/db/flake/index/novelty.cljc
@@ -33,11 +33,12 @@
   (let [novelty-size (get-in db [:novelty :size])]
     (> novelty-size reindex-min-bytes)))
 
-(defn novelty-max?
+(defn max-novelty?
   "Returns true if ledger is beyond novelty-max threshold."
-  [db reindex-max]
-  (let [novelty-size (get-in db [:novelty :size])]
-    (> novelty-size reindex-max)))
+  [db]
+  (let [novelty-size (get-in db [:novelty :size])
+        max-novelty  (:reindex-max-bytes db)]
+    (> novelty-size max-novelty)))
 
 (defn dirty?
   "Returns `true` if the index for `db` of type `idx` is out of date, or if `db`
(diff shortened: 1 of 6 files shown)


src/clj/fluree/db/flake/transact.cljc
```diff
@@ -120,6 +121,9 @@
 (defn stage
   [db fuel-tracker context identity author annotation raw-txn parsed-txn]
   (go-try
+    (when (novelty/max-novelty? db)
+      (throw (ex-info "Maximum novelty exceeded, try again later."
+                      {:status 429 :error :db/max-novelty-exceeded})))
```
zonotope, line 126: I think the status code should be [503](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503)

src/clj/fluree/db/query/fql/syntax.cljc
```diff
@@ -140,7 +140,15 @@
                          [:file-data? {:optional true} :boolean]
                          [:index-files-ch {:optional true} :any]
                          [:time {:optional true} :string]]
-    ::txn-opts          [:and ::stage-opts ::commit-opts]
+    ::ledger-opts       [:map
```
zonotope, line 143: I don't think these options are valid for any transaction, but only when a new ledger is created. I think it would be better for there to be separate create and transact option schemas.

src/clj/fluree/db/flake/transact.cljc
```diff
@@ -120,6 +121,9 @@
 (defn stage
   [db fuel-tracker context identity author annotation raw-txn parsed-txn]
   (go-try
+    (when (novelty/max-novelty? db)
+      (throw (ex-info "Maximum novelty exceeded, try again later."
```
zonotope, line 125: I think this error message should have better directions. The user would have to instantiate a new db after an index process has completed, so we should mention that.

src/clj/fluree/db/flake/index/novelty.cljc
```diff
@@ -33,11 +33,12 @@
   (let [novelty-size (get-in db [:novelty :size])]
     (> novelty-size reindex-min-bytes)))
 
-(defn novelty-max?
+(defn max-novelty?
```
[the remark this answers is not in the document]
dpetran: Would it be fine if I flip the other one? As a predicate I think the `<condition>-novelty?` construction makes a lot more sense when reading.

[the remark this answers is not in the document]
zonotope: sure. i just think they have to match.

src/clj/fluree/db/flake/transact.cljc
dpetran: "Max novelty exceeded, wait until indexing is completed."
^ How's that?

dpetran: I went with this:
> "Maximum novelty exceeded, no transactions will be processed until indexing has completed."

bplatz: Only if using the library, correct? I think such a message would confuse someone if they were using fluree/server - so I'd opt to leave that out.

Should it work like this? I think it is odd that a user has a db-t that won't work, but then reloads the same "immutable" db-t to get it to work. Seems like we should be handling that somehow for the user. While I'm not thrilled with it, the new BM25 holds some index state metadata in an atom, and is aware of background updates - replacing/updating the index once complete. Not for this PR - but something I think we should have some thought into.

zonotope: This is a good point re server vs the library. I now agree that we should leave the reloading the db part out. If someone is using the db library, they probably would have a better understanding about how indexing works anyway.

I think there are definitely warts with this approach forcing the user to update their own db after the fact, but there's a lot of complexity involved with keeping a handle of all dbs that have been instantiated and having them update themselves as well. I agree that we should probably handle this for the user eventually though.

dpetran: I'm glad you mentioned that Brian, because what I came up with is "... Obtain a new Transactable record with a reference to the new index." And that seemed way too inside-baseball for a regular user.

zonotope (approved):
🔔

PR #936, opened 2024-11-12
Initial working BM25

bplatz:
This creates an initial BM25 implementation that can be built from, but is working.

TODO before releasing:
- serialize/deserialize BM25 index into index-root
- Handle retractions

I'll try to work on at least the retractions piece, but putting up here for comments & review, and anyone else to build off of.

There are a number of additional TODOs related to catching exceptions, updating virtual graph settings, etc. These are marked with `;; TODO - VG -` prefix so they can be searched for in the code base.

Passing tests for BM25, where you can also see the syntax is in the `test` directory with namespace: `fluree.db.vector.bm25-test`


src/clj/fluree/db/virtual_graph/proto.cljc
```diff
@@ -0,0 +1,7 @@
+(ns fluree.db.virtual-graph.proto)
```
zonotope, line 1: I'm wary of protocol only namespaces because they allow cyclic dependencies to develop. Why can't this protocol live in the `fluree.db.virtual-graph.index-graph` namespace?

src/clj/fluree/db/virtual_graph/index_graph.cljc
```diff
@@ -0,0 +1,97 @@
...
+    [db* alias initialized-vg]))
+
+(defn load-virtual-graph
+  [db alias]
+  (or (get-in db [:vg alias])
+      (throw (ex-info (str "Virtual graph requested: " alias " does not exist for the db.")
+                      {:status 400
+                       :error  :db/invalid-query}))))
+
+(defn update-vgs
```
zonotope, line 80: This functionality feels like it should eventually be handled by the [`fluree.db.indexer/Indexable` protocol](https://github.com/fluree/db/blob/43013bc5020b3b180330950f02bae404b83f8cdf/src/clj/fluree/db/indexer.cljc#L6-L7). and the [implementation of that protocol for `FlakeDB` instances](https://github.com/fluree/db/blob/43013bc5020b3b180330950f02bae404b83f8cdf/src/clj/fluree/db/flake/flake_db.cljc#L355-L359)

I understand these virtual graphs to be just different kind of indexes sitting alongside the 4 sorted set-based flake indexes, and a `db` is a unified view of all of these indexes. Since the `Indexable` protocol handles updating the sorted set-based indexes after a transaction, I think it could be adapted to also handle updating the bm25 index too, as well as any other indexes that might be defined on a ledger.

src/clj/fluree/db/virtual_graph/proto.cljc
```diff
@@ -0,0 +1,7 @@
...
+
+(defprotocol UpdatableVirtualGraph
+  (upsert [this source-db new-flakes remove-flakes] "Updates the virtual graph with the provided flakes. Returns async chan with new updated VirtualGraph or exception.")
+  (initialize [this source-db] "Initialize a new virtual graph based on the provided db - returns promise chan of eventual result")
+  (serialize [this] "Returns a JSON serializable representation of the virtual graph (does not serialize to JSON)")
+  (deserialize [this source-db data] "Reifies the virtual graph from the provided data structure"))
```
zonotope, line 7: I think we might be able to adapt the existing [`fluree.db.serde.protocol/StorageSerializer` protocol](https://github.com/fluree/db/blob/43013bc5020b3b180330950f02bae404b83f8cdf/src/clj/fluree/db/serde/protocol.cljc#L6-L14) to work with virtual graphs.

src/clj/fluree/db/virtual_graph/index_graph.cljc
bplatz: Agree and I didn't actually try to do that so don't have any reasons as to why that could be difficult or easy.

Couple potential notes that could affect that:
- Unlike our regular indexes which can use novelty to always be current, BM25 is only usable at current 't' once a index is complete, and while fast, it is slower than our other SS-based indexes.
- Mostly because of the above point, you can opt to `sync: false` - which means your query doesn't depend on the latest index, so use the most recently completed one. This means a BM 25 virtual graph needs access to two indexes, and will use one or the other depending on that sync flag (default is `sync: true`).
- Also because of the first point, indexes need to be run with every db update

zonotope: One idea that might accommodate the time difference in updating a vector index vs the instant novelty update is tweaking what we expect to be returned from the `Indexable/-index` method. 

Right now we expect a channel that will eventually contain a single new indexed db instance. Instead, we could expect a channel that could contain multiple indexed dbs. As each indexing process finishes, we return the updated db on the channel, and then we close the channel when all the indexing processes finish.

Downstream code could keep pulling on the channel until it closes and update the latest db by incorporating the new indexes from each value that comes off the channel. 

I think there might be some issues still to hammer out in write heavy scenarios where new transactions come in before the vector index update completes, but I think we can resolve those issues by reusing any previously returned indexing channels that are still open after the `-index` method gets called because a new transaction comes in.

src/clj/fluree/db/virtual_graph/proto.cljc
bplatz: This work was done in #942

src/clj/fluree/db/virtual_graph/proto.cljc
bplatz: This protocol namespace was removed in #942

PR #937, opened 2024-11-15
CHORE: Fix typo in test

alan-icebreaker:
Fix typo in test case. Please see https://en.wikipedia.org/wiki/Star_Wars_opening_crawl for reference :)


zonotope (approved):
How could we have missed this?!

alan-icebreaker:
> How could we have missed this?!

Spending too much time with friends and power converters :)

PR #938, opened 2024-11-17
Feature/bm25 retractions

bplatz:
This add retractions and updates to the initial BM25 search PR here https://github.com/fluree/db/pull/936

Now, when a new "document" needs to get indexed, it first checks if the document is already in in the index and if so, retracts it before adding the updated one.

Also full retractions are working - when you delete all the properties for an indexed document so there is nothing left for the given IRI, a full retraction is now applied.

Retractions don't attempt to reduce the dimensionality of the index, nor do I think it should. E.g. in the tests the words apples and oranges were part of the index, but then the document containing those words. With no other documents containing those words, the dimensionality could be reduced but it retains those terms as sparse vector positions. I think we'll want this for time travel, and the number of orphaned terms should be relatively low with any modest sized index.

### DIFF
--- a/src/clj/fluree/db/virtual_graph/bm25/index.clj
+++ b/src/clj/fluree/db/virtual_graph/bm25/index.clj
@@ -95,41 +95,41 @@
 (defn search
   [{:keys [stemmer stopwords k1 b index-state] :as bm25} solution error-ch out-ch]
   (go
-   (try*
-    (let [{::vg-parse/keys [target limit timeout] :as search-params} (vg-parse/get-search-params solution)
-          {:keys [pending-ch index]} @index-state
+    (try*
+      (let [{::vg-parse/keys [target limit timeout] :as search-params} (vg-parse/get-search-params solution)
+            {:keys [pending-ch index]} @index-state
 
-          ;; TODO - check for "sync" options and don't wait for pending-ch if sync is false
+            ;; TODO - check for "sync" options and don't wait for pending-ch if sync is false
 
-          index*      (if pending-ch
-                        (let [timeout*   (or timeout 10000)
-                              timeout-ch (async/timeout timeout*)
-                              [idx ch] (alts! [timeout-ch pending-ch])]
-                          (if (= timeout-ch ch)
-                            (put! error-ch (ex-info (str "Timeout waiting for BM25 index to sync after "
-                                                         timeout* "ms. Index is " (percent-complete index-state)
-                                                         "% complete. Please try again later. To configure a "
-                                                         "different timeout, set " const/iri-index-timeout " in the virtual "
-                                                         "graph query to an integer of milliseconds.")
-                                                    {:error  :db/timeout
-                                                     :status 408}))
-                            idx))
-                        index)
-          {:keys [vectors item-count avg-length terms]} index*
-          query-terms (bm25.search/parse-query target terms item-count stemmer stopwords) ;; parsed terms from query with idf calculated
-          candidates  (reduce #(into %1 (:items %2)) #{} query-terms)] ;; reverse index allows us to know which docs contain each query term, put into single set
-      (->> candidates
-           (score-candidates query-terms vectors avg-length k1 b)
-           (sort-by :score #(compare %2 %1))
-           (vg-parse/limit-results limit)
-           (vg-parse/process-results bm25 solution search-params true)
-           (async/onto-chan! out-ch)))
-    (catch* e
-            (log/error e "Error ranking vectors")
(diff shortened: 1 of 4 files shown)


src/clj/fluree/db/virtual_graph/bm25/index.clj
```diff
@@ -182,23 +181,48 @@
...
         affected-iris  (map #(iri/decode-sid db %) affected-sids)
+        item-count     (count affected-iris)
         pq             (parsed-query bm25)
-        iri-var        (-> pq :select :subj)
-        iri-values     (map #(hash-map iri-var (where/match-iri %)) affected-iris)
-        pq*            (assoc pq :values iri-values)
-        upsert-docs-ch (exec/query db nil pq*)]
 
-    (bm25-upsert* bm25 db upsert-docs-ch)))
+        ;; TODO - change below single query to upsert-queries to handle multiple queries
```
zonotope, line 213: it looks like this todo no longer applies. there is also commented code on line 214

src/clj/fluree/db/virtual_graph/bm25/index.clj
```diff
@@ -182,23 +181,48 @@
...
+        upsert-docs-ch (upsert-queries db pq affected-iris) #_(exec/query db nil pq*)]
+
+    (bm25-upsert* bm25 db item-count upsert-docs-ch)))
 
 (defn bm25-initialize
   [{:keys [query-parsed] :as bm25} db]
-  (let [index-items-ch (exec/query db nil query-parsed)]
-    (bm25-upsert* bm25 db index-items-ch)))
+  (let [query-result (exec/query db nil query-parsed)
+        ;item-count  (count query-result)
```
zonotope, line 221: Should this line be commented?

src/clj/fluree/db/virtual_graph/bm25/index.clj
```diff
@@ -182,23 +181,48 @@
...
+        items-ch     (async/chan 1 (map #(vector ::bm25.update/upsert %)))]
+    ;; break up query results into individual document items on a new chan
+    (async/pipeline-async 1 items-ch #(async/onto-chan! %2 %1) query-result)
+    (bm25-upsert* bm25 db 999 items-ch)))
```
zonotope, line 225: Where does the 999 come from? should this be the commented out `item-count`?

zonotope (approved):
lgtm. i just have a few minor questions/suggestions.

src/clj/fluree/db/virtual_graph/bm25/index.clj
```diff
@@ -182,23 +181,48 @@
...
+(defn upsert-queries
+  [db parsed-query affected-iris]
+  (let [results-ch (async/chan)
+        iri-var    (-> parsed-query :select :subj)]
+    (async/go
+      (loop [[next-iri & r] affected-iris]
+        (if next-iri
+          (let [values  [{iri-var (where/match-iri next-iri)}]
+                pq      (assoc parsed-query :values values)
+                result  (async/<! (exec/query db nil pq))
```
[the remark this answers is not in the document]
bplatz: Addressed in 53027b2983e0e76a19ed122d693b00769968b623

src/clj/fluree/db/virtual_graph/bm25/index.clj
```diff
@@ -182,23 +181,48 @@
...
+                result* (if (vector? result)
+                          (first result)
```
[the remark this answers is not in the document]
bplatz: Only ever one value, but handle the :select-one case now at parsing time so this logic no longer required. in commit 28f55a1b1ffe491aefed7950b61f3f21319ae04c

src/clj/fluree/db/virtual_graph/bm25/index.clj
bplatz: Since indexing can take a long time, we track status of completion to report for timed out queries waiting on the index.

In the case of initializing the index for the first time, we query for all items to index but handle it lazily, as in theory it could query the entire DB. Because it is lazy, we don't have item-count in this scenario (but we do for updates to an existing index).

I've reworked the logic here to report out the best status we can, while still keeping this query lazy.

Addressed in 3a427cc65a38d44d04d7a37eda27250bc20c5972

src/clj/fluree/db/virtual_graph/bm25/index.clj
bplatz: Per below comment, addressed in 3a427cc65a38d44d04d7a37eda27250bc20c5972

PR #939, opened 2024-11-18
Feature/policy values

dpetran:
Closes https://github.com/fluree/core/issues/128

This change makes the `opts.policyValues` use the same syntax as an fql values clause or a values pattern.
```
["?a" [1 2]]
;; or
[["?a "?b] [["a1" "b1"] ["a2" "b2"]]]
```
It composes with an existing values clause on the policy query by adding the specified values as a values pattern in the where clause.

### DIFF
--- a/src/clj/fluree/db/query/fql/parse.cljc
+++ b/src/clj/fluree/db/query/fql/parse.cljc
@@ -5,7 +5,6 @@
             [fluree.db.constants :as const]
             [fluree.db.datatype :as datatype]
             [fluree.db.json-ld.iri :as iri]
-            [fluree.db.json-ld.policy :as policy]
             [fluree.db.query.exec.eval :as eval]
             [fluree.db.query.exec.select :as select]
             [fluree.db.query.exec.where :as where]
@@ -13,6 +12,7 @@
             [fluree.db.util.context :as context]
             [fluree.db.util.core :as util :refer [try* catch*]]
             [fluree.db.util.log :as log :include-macros true]
+            [fluree.db.util.parse :as util.parse]
             [fluree.db.validation :as v]
             [fluree.json-ld :as json-ld]))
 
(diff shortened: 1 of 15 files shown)


src/clj/fluree/db/query/fql/parse.cljc
```diff
@@ -145,19 +146,15 @@
...
-      (if (every? (fn [binding]
-                    (= (count binding) var-count))
-                  vals*)
-        [vars* (mapv (fn [vals**]
-                       (parse-value-binding vars* vals** context))
-                     vals*)]
-        (throw (ex-info (str "Invalid value binding: "
-                             "number of variables and values don't match: "
-                             values)
+    (let [[vars vals] (policy/normalize-values values)
```
zonotope, line 149: If that function will be used for more than just policy related features then it should live in a different namespace.

test/fluree/db/policy/query_test.clj
```diff
@@ -293,3 +290,95 @@
...
+                                            "ex:capitol" "Charlotte"}
+                                           {"@id" "usa:co"
+                                            "@type" "usa:state"
+                                            "ex:name" "Colorado"
+                                            "ex:capitol" "Denver"}
+                                           {"@id" "usa:pr"
+                                            "@type" "usa:territory"
+                                            "ex:name" "Puerto Rico"
+                                            "ex:capitol" "San Juan"}]})]
+    (testing "no policyValues"
```
zonotope, line 319: What's supposed to happen when there are no policyValues? This test needs more commentary

zonotope (approved):
I have two changes that I pointed out, but this looks good otherwise.

src/clj/fluree/db/query/fql/parse.cljc
dpetran: I had it in `parse` originally, but that unfortunately introduced a massive cyclic dependency. I'll see if I can untangle it another way, I agree that this is not the right home for it.

dpetran: I ended up putting it in a new `util` ns, I couldn't really find another place that seemed correct and this seems like the least bad way.

PR #941, opened 2024-11-22
Fix/long >string

dpetran:
This fixes an OutOfMemoryError that occurred when we exhaust the heap space while trying to turn the long array on a SID back into a string. The IRI in question in this case was is `http://dbpedia.org/resource/Permian%E2%80%93Triassic_extinction_event`.

I've started adding property tests to ensure that all UTF-8 strings can be roundtripped properly, but that has exposed a few more uncovered cases around how we deal with padding. In the interest of making this fix available quickly, this PR doesn't include those (failing) tests yet.

### DIFF
--- a/src/clj/fluree/db/util/bytes.cljc
+++ b/src/clj/fluree/db/util/bytes.cljc
@@ -35,15 +35,9 @@
 
 
 (defn long->UTF8
-  [n]
-  (loop [result '()
-         n'     n]
-    (if (= 0 n')
-      #?(:clj (byte-array result)
-         :cljs (.from js/Int8Array result))
-      (let [b   (bit-and n' 0xFF)
-            n'' (bit-shift-right n' 8)]
-        (if (= b 0)
-          (recur result n'')
-          (recur (conj result b)
-                 n''))))))
+  [l]
+  (->> [56 48 40 32 24 16 8 0] ;; byte offsets
+       (map (fn [i] (bit-and (bit-shift-right l i) 0xFF)))
+       (remove zero?) ;; get rid of trailing padding
+       #?(:clj (byte-array)
+          :cljs (.from js/Int8Array))))
--- a/src/clj/fluree/db/util/bytes.cljc
+++ b/src/clj/fluree/db/util/bytes.cljc
@@ -33,11 +33,14 @@
               (bit-shift-left result (* diff 8)))
             result))))))
 
-
 (defn long->UTF8
   [l]
-  (->> [56 48 40 32 24 16 8 0] ;; byte offsets
-       (map (fn [i] (bit-and (bit-shift-right l i) 0xFF)))
-       (remove zero?) ;; get rid of trailing padding
-       #?(:clj (byte-array)
-          :cljs (.from js/Int8Array))))
+  (loop [offset 56
+         bs     []]
+    (if (neg? offset)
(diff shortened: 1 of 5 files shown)


src/java/fluree/db/SID.java
```diff
@@ -3,33 +3,25 @@
...
+    public final int namespaceCode;
+    public final long[] nameCodes;
 
-    public SID(int nsCode, long... nameCodes) {
+    public SID(int nsCode, long[] nameCodes) {
         this.namespaceCode = nsCode;
         this.nameCodes = nameCodes;
     }
 
-    public int getNamespaceCode() {
```
zonotope, line 14: Why did the getters need to be removed?

zonotope: We need to add them back. This is a Java class, not Clojure. Those fields should not be public, and they should not be settable. This change makes them settable. 

I don't think anything about the Java class needed to change to fix this bug.

dpetran: My understanding is that `final` indicates that the field cannot be mutated and is more than sufficient for safety, but I've gone ahead and reverted the changes.

src/clj/fluree/db/util/bytes.cljc
```diff
@@ -35,15 +35,9 @@
...
-         :cljs (.from js/Int8Array result))
-      (let [b   (bit-and n' 0xFF)
-            n'' (bit-shift-right n' 8)]
-        (if (= b 0)
-          (recur result n'')
-          (recur (conj result b)
-                 n''))))))
+  [l]
+  (->> [56 48 40 32 24 16 8 0] ;; byte offsets
+       (map (fn [i] (bit-and (bit-shift-right l i) 0xFF)))
```
zonotope, line 40: This probably doesn't matter in practice, but I think we should keep the loop structure here instead of using `map` + `remove`. This code ideally should be as fast as possible because it's often called repeatedly when processing a query.

There's a lot of other things to optimize, and the functional implementation is still way faster than having to do an async query the way it was before, so I don't feel that strongly about changing it back.

zonotope (approved):

src/java/fluree/db/SID.java
zonotope: I think we should strive to write code as idiomatically as possible in whatever language we're writing in. I would not have written a clojure namespace hiding data behind a function, but that is the Java way, and this is a Java class.

src/clj/fluree/db/util/bytes.cljc
dpetran: You raise a good point, I've rewritten it to be a loop. `criterium/quick-bench` pegs it as a little more than twice as fast now - ~300ns per long vs ~850ns with the seq version. I've also added a property test to make sure that it's working for inputs we didn't consider.

src/java/fluree/db/SID.java
dpetran: My bold take is that idiomatic java is in this case just wrong - getters are a bad design and the convention for using them comes from a time when `final` did not exist. 

Especially since this is in performance sensitive code my thought is why waste a stack frame on something we don't need. Same thought with changing the signature from `long...` to `long[]`, we only ever supply an array so no need to generate vararg handling. But in the end it's only a stack frame and I'm sure the JIT will handle both cases correctly anyways.

PR #942, opened 2024-11-26
Serialize and persist bm25 indexes

zonotope:
This patch adds functions to serialize and persist any virtual graph indexes as part of the normal indexing process and to reload any persisted indexes when a db is loaded from storage. 

This pr currently has draft status because the index isn't persisted if the minimum novelty threshold isn't reached and the graph indexing process doesn't kick off. I'm still working on the best way to detect when the bm25 index has changed and needs persisting at commit time, and how to wait until the bm25 indexing process completes to persist the index without holding up the transactor, but I wanted to publish the persistence code now for feedback.


bplatz (commented):
Awesome!

Are you thinking we should write out new VG indexes with every commit, or try to rebuild the index from the last FlakeIndex 't' and just hook into that process?

PR #945, opened 2024-12-06
Feature - Required Policy

dpetran:
## What's changed in this PR:

The execution semantics of when a policy is evaluated changed. We segregate policies into three groups:
1. "property" policies - have an `f:onProperty` constraint
2. "class" policies - have an `f:onClass` constraint
3. "default" policies - have neither `f:onClass` nor `f:onProperty` constraints

Before, we would only evaluate the policies in a single group, checked in the order specified above. This had the effect that if a "property" policy existed we would never check the "class" policies.

Now, we will check every policy regardless of its grouping, as outlined below.

This PR also introduces the `f:required` property, the semantics of which are outlined below.

I spent a lot of time making sure I understood how policies worked before I made changes, so below is a detailed description of how they work now.

## Authorization Policies

A Policy is a document that is used to decide which flakes can be viewed or modified. We use this vocabulary:

- `f:query`
  This is a Fluree query is used to determine if a policy allows access to a flake. It is the only required property on a policy. It is stored as an `@type` `@json` blob with its own `@context` that is separate from the query or transaction context.
- `@id` (optional)
- `@type` (optional)
- `f:onProperty` (optional)
  This is used to constrain when a policy is evaluated - only when a flake has a predicate that matches the properties specified here.
- `f:onClass` (optional)
  This is used to constrain when a policy is evaluated - only when a flake's subject has an @type that matches the classes specified here. This is more expensive because we need to look up the flake's subject's @type refs in order to find the policies that are relevant before we can evaluate the policy.
- `f:action` (optional)
  This is used to constrain when a policy is evaluated:
  `f:view` - during a query
  `f:modify` - during a transaction
  If no `f:action` is specified then a policy is applied for both.
- `f:required` (optional)
  If true, all non-required policies will be ignored. All required policies must evaluate to true to permit access.
- `f:exMessage` (optional)
  A custom error message to return if a policy prevents modification during a transaction.

At a conceptual level, policies are processed in this way for each flake that will be returned:
1. All `f:required` policies must evaluate to `true` to permit access the flake. Non required policies are ignored.
2. If no policies are "required", then all the policies are checked in sequence. The first one to evaluate to `true` permits access. If no policy evaluates to true, access to the flake is denied.

During policy evaluation, the `f:query` is executed with certain variables bound:
`?$this` - the subject of the flake being tested
`?$identity` - the identity of the principal performing the action

The query returns an authorization result:
- 1+ results: flake is allowed
- 0 results: flake is not allowed
- error: action is unwound

In order to have policies enforced at all, the correct opts need to be supplied: `identity`, `policyClass`, or `policy`. When present, we "wrap" the db in the applicable policies:
- `identity`: all policies with an `@type` found in the `f:policyClasses` specified on the `<identity>` subject.
- `policyClass`: all the policies with an `@type` that matches any of the supplied values
- `policy`: just the policies provided by this property.

The `credential-*` api functions derive the identity from the signed action and supply the `identity` opt.

### DIFF
--- a/src/clj/fluree/db/constants.cljc
+++ b/src/clj/fluree/db/constants.cljc
@@ -59,6 +59,7 @@
 (def ^:const iri-rule (fluree-iri "rule"))
 (def ^:const iri-query (fluree-iri "query"))
 (def ^:const iri-onClass (fluree-iri "onClass"))
+(def ^:const iri-required (fluree-iri "required"))
 (def ^:const iri-onProperty (fluree-iri "onProperty"))
 (def ^:const iri-exMessage (fluree-iri "exMessage"))
 (def ^:const iri-policyClass (fluree-iri "policyClass"))
--- a/src/clj/fluree/db/json_ld/policy/enforce.cljc
+++ b/src/clj/fluree/db/json_ld/policy/enforce.cljc
@@ -69,7 +69,6 @@
                "Policy enforcement prevents modification.")
            {:status 403 :error :db/policy-exception}))
 
-
 (defn policies-allow?
   "Once narrowed to a specific set of policies, execute and return
   appropriate policy response."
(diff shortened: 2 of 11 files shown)


dpetran:
Closes https://github.com/fluree/core/issues/158

zonotope (approved):
🔑

PR #949, opened 2025-01-13
Feature/integrated server

zonotope:
This small patch separates the `commit!` api function by adding a new `apply-stage!` function that returns a map of commit metadata including the commit address, hash, size, new t value, and the new db record. The pre-existing `commit!` function now only returns the updated db. These two pieces of functionality used to be complected within the same `commit!` function distinguished with the `file-data?` option. Having two separate functions leads to less complexity than handling control flow with opaque options.

This patch also includes a `prepare` function for parsing and expanding json-ld configuration documents into internal integrand config maps that's useful for repl sessions in dependent applications.

### DIFF
--- a/src/clj/fluree/db/api.cljc
+++ b/src/clj/fluree/db/api.cljc
@@ -222,8 +222,7 @@
 
 (defn apply-stage!
   ([ledger staged-db]
-   (promise-wrap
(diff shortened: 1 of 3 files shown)


dpetran (approved):
:apple:

PR #950, opened 2025-01-15
Optimize Stale Flake Removal

zonotope:
This patch brings ins @aaj3f 's fix for inconsistent indexing configuration options and includes an optimization for removing repeated equivalent facts from query results by maintaining a set of previously checked facts to avoid traversing the list of flakes in a node multiple times.


dpetran (approved):
:rabbit2:

PR #951, opened 2025-01-16
Optimize time travel and novelty resolution

zonotope:
This patch has a few optimizations of how flake sets are manipulated when resolving index nodes to specific `t` values. The optimizations here avoid multiple traversals of flake sets where possible and makes better use of transient sets.


dpetran (approved):
:snowflake:

PR #952, opened 2025-01-17
Feature/property path

dpetran:
Adds support for a single zero-or-more and one-or-more transitive property.

The syntax we use is to wrap the property path in angle brackets and then use plain SPARQL property path expressions.
```
;; one-or-more `ex:y`
{"where" [{"@id" "?x" "<ex:y+>" "?y"}]
 "select" ["?x" "?y"]}

;; zero-or-more `ex:y`
{"where" [{"@id" "?x" "<ex:y*>" "?y"}]
 "select" ["?x" "?y"]}
```
More complex property paths are not supported: sequential, alternate, and negated are not part of the scope of this PR.


dpetran:
Rebased on main.

src/fluree/db/query/fql/parse.cljc
```diff
@@ -366,14 +368,37 @@
     (parse-variable id)
     (parse-subject-iri id context)))
 
+(declare parse-predicate)
+(defn parse-property-path
```
zonotope, line 372: There currently isn't a clojure friendly property path syntax. What's wrong with something like `[:ex/foo :+]`?

dpetran: I'm not sure that there's anything wrong with that, I was just trying to keep parsing as simple as I could - once you start composing path operators it can get pretty gnarly and I wanted to avoid introducing divergent behavior at this stage.

zonotope (approved):
🦯

PR #954, opened 2025-01-20
Fix/base context in queries

dpetran:
Closes https://github.com/fluree/core/issues/88

This is a fix to support using `@base` in queries. Subject iris were not getting expanded with the `@base` context and this PR adds support.


zonotope (approved):
🛷

PR #955, opened 2025-01-22
Fix existing ledger instantiation

zonotope:
This patch fixes an error instantiating existing ledgers that caused indexing options to be treated as identity keypairs. The fix is in commit 23ab9284f55a66626c97833de6a12536d8fffe1b, and the rest of the changes here are just cleanup for things I came across when looking for the source of this bug.

I guess this is another point in favor of statically typed languages 🤷🏾‍♂️

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -59,7 +59,7 @@
    :ns-default build}
 
   :dev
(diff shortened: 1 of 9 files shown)


dpetran (approved):
:tv: 

Static types are _occasionally_ useful

PR #956, opened 2025-01-24
External API - throw if invalid 'conn'

bplatz:
On many occasions when working with the API, I forget to deref @(fluree/connect ...).

It ends up throwing a Nil exception within an async call stack, and is worse than useless (to me) as it makes me start to think there is a weird bug.

This adds a check, and hopefully useful error to anyone doing this on what the issue likely is. This only affects the public API calls that take 'conn', which this extra check should have no impact on performance.

### DIFF
--- a/src/clj/fluree/db/api.cljc
+++ b/src/clj/fluree/db/api.cljc
@@ -43,6 +43,13 @@
                (reject res)
                (resolve res))))))))
 
+(defn- validate-connection
+  "Throws exception if x is not a valid connection"
+  [x]
+  (when-not (connection/connection? x)
+    (throw (ex-info "Unable to create new ledger, connection is not valid. fluree/connect returns a promise, did you deref it?"
+                    {:status 400 :error :db/invalid-connection}))))
+
 (defn connect
   "Forms connection to ledger, enabling automatic pulls of new updates, event
   services, index service.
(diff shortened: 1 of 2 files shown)


src/clj/fluree/db/api.cljc
```diff
@@ -43,6 +43,13 @@
                (reject res)
                (resolve res))))))))
 
+(defn- validate-connection
+  "Throws exception if x is not a valid connection"
+  [x]
+  (when-not (connection/connection? x)
+    (throw (ex-info "Unable to create new ledger, connection is not valid. fluree/connect returns a promise, did you deref it?"
```
zonotope, line 50: Minor suggestion that might not be worth it, but you could use `(instance? clojure.lang.IDeref x)` to tell if it's most likely an unrealized promise or something else, and you could then give a better error message. 

I say it might not be worth it because that will also return true for atoms and delays, but those most likely wouldn't be passed to these api functions. Either way, you can take or leave this suggestion.

zonotope (approved):
🏛️

src/clj/fluree/db/api.cljc
bplatz: Would you expect any CLJS compatibility issues with that approach? I could require that record type with :import and #?(:cljs ... but since it is a clojure name, I was concerned about the name working x-platform.

zonotope: I did not try it on cljs, so I'm not sure if it would work. I know there is a corresponding `IDeref` protocol on that platform, so the right reader conditional would do the trick if it doesn't work cross platform out of the box. In any case, it's no big deal

PR #957, opened 2025-01-25
Flexible policy targeting

dpetran:
This PR adds a more expressive and consistent way to control when a policy is evaluated. 

It introduces three policy keys:
```
| policy key     | value    |
|----------------+----------|
| targetSubject  | WhenExpr |
| targetProperty | WhenExpr |

// allow either static values or a dynamic query
WhenExpr = Value[] | WhereClause
// <where-clause> needs to bind results to the variable `?$target`.
WhereClause = {"@type" "@json" "@value" <where-clause>}
```
These new policy keys are a strict superset of the functionality of the existing targeting keys: `onProperty` and `onClass`. `onClass` could be considered syntax sugar for the following:
```
{...
  "onSubject" {"where" [{"@id" "?$target" "@type" <my-class>}]}
  ...
}
```


src/clj/fluree/db/json_ld/policy/enforce.cljc
```diff
@@ -52,6 +52,19 @@
     (get-in policy-map [const/iri-modify :default])
     (get-in policy-map [const/iri-view :default])))
 
+(defn policies-for-flake
+  [{:keys [policy namespace-codes] :as db} [s p o :as flake] modify?]
+  (let [s-iri (iri/sid->iri s namespace-codes)
+        p-iri (iri/sid->iri p namespace-codes)
+        o-val (if (iri/sid? o) (iri/sid->iri o namespace-codes) o) ]
```
bplatz, line 59: This, and the code below is extremely performance sensitive... it could be done millions of times in a single query. This work is irrelevant if these new types of policies are not present.

We definitely need a way to skip all of this extra work preemptively. At minimum a flag on the policy could be checked, but checking the flag is still potentially extra work... but seems like a reasonable short-term solution. I think a composed policy function that only composes the relevant checks per flake that a given policy enforces feels like a more efficient way, but could be something we do later.

src/clj/fluree/db/json_ld/policy/rules.cljc
```diff
@@ -84,68 +72,106 @@
...
+(defn parse-targets
+  [db target-exprs]
+  (let [out-ch (async/chan 2)
+        in-ch  (async/to-chan! (mapv #(or (util/get-id %) (util/get-value %)) target-exprs))]
+    (async/pipeline-async 2
+                          out-ch
+                          (fn [target-expr ch]
+                            (if (map? target-expr)
+                              ;; maps are query where clauses
+                              (-> (dbproto/-query db (assoc target-expr "select" "?$target"))
```
bplatz, line 84: Is there a reason to require special variable syntax? I'd favor people writing a "normal" query with their own select syntax... but should do some validation that "select" is bound to just a single var.

Certainly a judgement call... so not opposed to this approach, just on surface I don't prefer it.

src/clj/fluree/db/json_ld/policy/enforce.cljc
```diff
@@ -52,6 +52,19 @@
     (get-in policy-map [const/iri-modify :default])
     (get-in policy-map [const/iri-view :default])))
 
+(defn policies-for-flake
+  [{:keys [policy namespace-codes] :as db} [s p o :as flake] modify?]
+  (let [s-iri (iri/sid->iri s namespace-codes)
```
bplatz, line 57: Why do this for every flake? If the `s-targets` is already encoded, this conversion isn't required at all.

src/clj/fluree/db/json_ld/policy/rules.cljc
```diff
@@ -84,68 +72,106 @@
...
+                          in-ch)
+    (async/into #{} out-ch)))
+
+(defn parse-policy
+  [db policy-doc]
+  (go-try
+    (let [id (util/get-id policy-doc) ;; @id name of policy-doc
+
+          subject-targets  (when-let [target-exprs (get policy-doc const/iri-targetSubject)]
+                             (<? (parse-targets db target-exprs)))
```
bplatz, line 98: Per prior comments, I think we should encode the return IRIs when parsing... perhaps here or down below in the fn.

What if @context is used in the query? If so, you won't be able to encode the IRIs for comparison as they will already be "shortened".

For this reason, I added a `:selection-context` internal field in a different PR which, if present, gets used in the select statement. It can be used to remove @context from applying to `select` statements.

src/clj/fluree/db/json_ld/policy/rules.cljc
```diff
@@ -84,68 +72,106 @@
...
+          property-targets (when-let [target-exprs (get policy-doc const/iri-targetProperty)]
+                             (<? (parse-targets db target-exprs)))
+          object-targets   (when-let [target-exprs (get policy-doc const/iri-targetObject)]
+                             (<? (parse-targets db target-exprs)))
```
bplatz, line 102: Why perform all 3 of these queries serially? In theory it would speed up this process 3x if parallel.

src/clj/fluree/db/json_ld/policy/rules.cljc
```diff
@@ -84,68 +72,106 @@
...
+                         :error  :db/invalid-policy}))))))
+
+(defn build-wrapper
+  [db]
+  (fn [wrapper policy]
+    (cond
+      (or (:s-targets policy)
+          (:p-targets policy)
+          (:o-targets policy))
+      (add-default-restriction policy wrapper)
```
bplatz, line 147: Am I reading this correct in that these will only work if `:on-property` or `:on-class` have no policy matches? As it stands, `:on-class` would only get checked if there was no `:on-property`. Wondering if there is an execution skipping condition here, or if always run.

bplatz (commented):
Would be nice to see some tests to get a feel for how this will actually look for an end user.

src/clj/fluree/db/json_ld/policy/enforce.cljc
dpetran: Good point, I can easily move SID encoding to the parsing stage.

src/clj/fluree/db/json_ld/policy/rules.cljc
dpetran: I was trying to make it work first, doing these in parallel is good low-hanging fruit for perf gains.

src/clj/fluree/db/json_ld/policy/rules.cljc
dpetran: If any of the `target-*` properties are present, we ignore `onClass` and `onProperty` along the same lines of how we ignore `onClass` if there's an `onProperty`.

src/clj/fluree/db/json_ld/policy/enforce.cljc
```diff
@@ -52,6 +52,19 @@
...
+        p-iri (iri/sid->iri p namespace-codes)
+        o-val (if (iri/sid? o) (iri/sid->iri o namespace-codes) o) ]
+    (->> (default-policies policy modify?)
+         (keep (fn [{:keys [s-targets p-targets o-targets default?] :as policy}]
+                 (when (or (contains? s-targets s-iri)
```
dpetran, line 62: Currently you can compose `target-*` properties with `OR`, but after talking with Andrew maybe `AND` semantics would be more useful.

src/clj/fluree/db/json_ld/policy/rules.cljc
dpetran: I did the hard-coded `?$target` variable in the spirit of our current harcoded vars: `?$identity` and `?$this`. It simplifies parsing, since we don't have to validate select and we can just document what we expect, but I don't think it's a big deal to just require them to return a single var anyways.

dpetran:
> Would be nice to see some tests to get a feel for how this will actually look for an end user.

Tests are still incoming, I was working by rewriting existing tests but started to put together a more comprehensive and complex scenario to ensure that policies can express everything we need them to.

src/clj/fluree/db/json_ld/policy/rules.cljc
bplatz: I think onClass goes away and this PR gives an alternate tool to accomplish this and it was never a great thing to have anyhow. 

I think however these need to be secondary to onProperty. We want policies to compose and if a global governance person wrote a policy for e.g. ex:ssn, to apply to all dbs in the company, it would be easy to subvert it with one of these policies locally.

src/clj/fluree/db/json_ld/policy/rules.cljc
bplatz: In the "virtual-graphs" (full text search, vector search) we allow people to format the queries using their own vars... but just when parsing the query, verify the `:select` is in a suitable format else provide an error with useful feedback.

The reason ?$this, ?$identity exists is because we inject those with `:values` into the query so we need a fixed variable name. I don't think that is happening here since the query is run once initially.

src/clj/fluree/db/json_ld/policy/enforce.cljc
bplatz: We've always discussed a `required` flag on a policy that needs `AND` semantics... `OR` I think is the preferred default (and how it currently works). We have yet to implement the `required` flag. Once a `required` flag exists for a given flake condition, you can ignore every other rule as they become irrelevant.

But we've never had the situation yet that different rule "types" could apply to the same Flake. It seems this introduces this new type of condition... and that then will potentially mean it will be more likely to have to run many computationally expensive rules per flake. But more pertinent than that, is it makes `AND` semantics more complex... is that an AND for `s`, or AND for `p`, or AND for `o`. Or is that AND for `s` and AND for `p` and AND for `0`.

dpetran: `required` actually landed back in early December: https://github.com/fluree/db/pull/945

The idea of this whole targeting functionality is to drastically limit the number of times we have to run the `f:query` query, since that's by far the most expensive part of the authorization process. 

I think composing separate _policies_ with OR is the right semantics, but @aaj3f pointed out that composing the different targeting methods with AND would be a lot more useful when it comes to limiting when the `f:query` is actually evaluated.

bplatz: > The idea of this whole targeting functionality is to drastically limit the number of times we have to run the f:query query, since that's by far the most expensive part of the authorization process

I think here you are referring to `onClass` policies, which require a class lookup... however `onProperty` require no lookup so that would still be the most performant way to make a policy. For this reason I think `onClass` should go away in favor of this PR. Do I have that correct?

My question is more around how there could be many `targetSubject` policies that apply to the same flake (I think those should OR, unless `required`), but then there could also one or more `targetProperty` that applies to the that flake as well, further one or more `targetObject` that applies to the same flake, and further one or more `onProperty` that apply to the same flake.

I think those should all `OR`... but it brings up more pressing a question of - should we inject into `f:query` not only `?$this`, but also `?$this-property`, and also `?$this-object`... and once you go that far you sort of need `?$this-datatype`. That would allow you to compose "and" semantics into your query - but it makes policy more expensive.

bplatz: Or, are you implying that multiple `target-*` on the same policy can exist, and all conditions are required to exist for the policy to run? I don't think the code currently reflects this but perhaps you are still working on it... if that's the case they would each have their own query, in addition to having a single policy `f:query` if I understand that. In that case you are producing up to 3 sets (one from each `target-*` query) and then a flake needs to belong to all sets in order to then run `f:query`? Maybe I'm way off here, but perhaps that was where you were implying.

dpetran: > multiple target-* on the same policy can exist, and all conditions are required to exist for the policy to run
Yes, this is what Andrew was asking for. It's not implemented here (it is locally, though).

aaj3f:
@dpetran I know there's some ongoing discussion above, but I wanted to ask (just for my own ability to keep track) if this PR will also include:

1. the change to have `targetSubject`, `targetProperty`, and `targetObject`, if used in combination on a single policy, for that to be combinatory (AND) rather than disjunctive (OR) when producing the flake filters for which flakes would trigger the policy (see discussion [here](https://fluree-internal.slack.com/archives/C04G5QYU4QL/p1738084177011589?thread_ts=1738079120.758839&cid=C04G5QYU4QL))
2. the inclusion of some pattern to abstract a desire for mere `ALLOW`/`true` & `DENY`/`false` (see discussion [here](https://fluree-internal.slack.com/archives/C04G5QYU4QL/p1738093317589509?thread_ts=1738093196.941589&cid=C04G5QYU4QL)). For example, `f:query: { "@id": "f:true" }` or `f:query: true` (or separated from `f:query` entirely) something like, but not necessarily named, `f:eval: { "@id": "f:deny" }`, e.g.
```json
{
  "f:targetSubject": "...",
  "f:action": "...",
  "f:query": { "@type": "@json": { "where": { "@id": "?$identity", "user": "?$this" } } },
  "f:eval": { "@id": "f:deny" }
  }
```

It's okay if the answer is no (although preferably I'd love to have characteristic 1 even if we punt on characteristic 2)

dpetran:
@aaj3f Yes for #1, I've got it working and am adding tests now, pushing after lunch :+1: 

No for #2, that will be a second PR, I think we should consider the possible approaches there, though I don't think any individual approach will be difficult to implement.

aaj3f:
> * `targetProperty` - there isn't a good use case where someone needs a query for properties... and we already of `onProperty` which you can list properties. Perhaps we could rename `onProperty` to `targetProperty` if that name was preferred.

@bplatz  Not a good use case, but the only one I can imagine is something like "I know that `secret-project-1` is not only classified, but any property that it uses to describe its triples should also be a classified property", so that you would need to do something like this
```json
{
  "f:targetProperty": {
    "@type": "@json",
    "@value": {
      "where": { "@id": "secret-project-1", "?$target": "?o" }
    }
  }
}
```

bplatz:
> "I know that `secret-project-1` is not only classified, but any property that it uses to describe its triples should also be a classified property"

We could consider making `onProperty` optional provided `allowSubjects` and/or `denySubjects` was present. That I believe would solve that use case.

Also worth nothing and I should have said it above, `f:query` would become optional (it is currently required). So these can create generic allow/deny lists without needed to further qualify if something is allowed via a query.

aaj3f:
>We could consider making onProperty optional provided allowSubjects and/or denySubjects was present. That I believe would solve that use case.

@bplatz I do think `onProperty` should be optional so that you can allow/deny on all triples for a subject (i.e. triples matching `[subj-match, any, any]`). I still am not sure this would solve my (entirely arbitrary and made-up scenario). 

Let's say we had
```json
[
  { "@id": "secret-project-1", "classifiedPropertyOne": "foo", "classifiedPropertyTwo": "bar" },
  { "@id": "andrew", "name": "Andrew", "classifiedPropertyOne": "foo" }
]
```

In this scenario, I don't know what properties are on `secret-project-1` and also they may change, so I don't want to hardcode them into a list for `onProperty`. So I want `targetProperty` as a query condition that finds those properties. Now, I want any subject that has triples on those properties to be restricted on those properties, but not any others. So if I had

```json
{
  "f:targetProperty": {
    "@type": "@json",
    "@value": {
      "where": { "@id": "secret-project-1", "?$target": "?o" }
    }
  },
  // and then some way to express "deny", TBD
}
```

Then queries could possibly still see `{ "@id": "andrew", "name": "Andrew" }`

bplatz:
> In this scenario, I don't know what properties are on `secret-project-1` and also they may change

I'm not sure I'm bought into this use case, but I can think of a way to solve it (albeit not extremely efficiently):

a) allowSubject - where condition queries for `secret-project-1` properties bound in a var, then pull every subject that has those properties
b) `f:query` - In addition to "injecting" `?$this`, we could also inject `?$this-property`, and the query can see if `{'@id": "secret-project-1", "?$this-property": "?x"}` returns truthy, and if so deny (not)

dpetran:
I think these are good ideas, I'm working through adding them now.

I am partial to allowing a query for `onProperty` (exactly how `targetProperty` works) and not making it `required` - my intuition is that running `f:query` is a lot slower than checking set membership for `onProperty`.

To make sure I'm understanding optional `f:query`:

If no `f:query` is present on a policy that will be enforced for a flake, the we allow the flake if it matches the "allow" subjects and forbid it if it matches the "forbid" subjects. If there are no forbidden or allowed subjects, we allow the flake.

dpetran:
I've talked with Andrew and Derek, and for at least this week we are going to stick with `targetSubject` and `targetProperty` and hopefully figure out a way to note to users that this is an exploratory api. That will allow nexus to build a ui to test it out. For now all the existing policy types still work, including `onClass`. I think we'll have to deprecate them in the docs before we remove support (if we even need to do that, they can coexist).

I've added the ability to reference `?$identity` in the `targetSubject` and `targetProperty` queries. I started to remove the implicit `?$target` variable that we `select`. I'm hesitating, though - deciding where to validate that is tricky. At the time the policy is applied, the user may have no ability to change policy being applied - it'll just be opaquely broken. Far better would be to validate it on transaction, but we don't validate specific vocabulary for well-formedness by default (SHACL or Fluree policies), and I think that's the right choice. If a user wants to validate that policies are well formed, I think we should provide a SHACL shape for that.

That said, I can still switch it to use a user-supplied select and just assume that it is a single variable and handle that gracefully.

I spent a good deal of time yesterday writing policies for a simple wishlist application (`target_test.clj`) to ensure that we have support for comprehensive security around a whole application. I found the developer UX very difficult - I struggled with knowing which policies were actually being run on which flakes, what the results of each policy application were, and reasoning about the performance implications of the policies. I was able to do it, but I had extensive internal logging and a repl that allowed me to inspect the interior state. I think writing comprehensive security policies for applications without that level of introspection will be very challenging for users.

I also found it much harder to reason about modification policy than view policy. Targeting subjects is pretty much useless for transactions, as the subjects we want to validate may not exist before the policy is parsed. Transactions have to be validated by `f:query` almost exclusively, which is slow.

I think it would be prudent to design tools to characterize the performance of different policies - counting how many times a `f:query` is executed for a given action at the very least. I think that will help guide us towards an objectively better policy framework, rather than the sort of ad-hoc design we're doing right now. As it stands, I find reasoning about it in the abstract much more difficult than just measuring would be.

src/clj/fluree/db/json_ld/policy/rules.cljc
```diff
@@ -71,81 +61,136 @@
...
+  [db policy-values target-exprs]
+  (let [in-ch  (async/to-chan! target-exprs)
+        out-ch (async/chan 2 (map (fn [iri] (iri/iri->sid iri (:namespaces db)))))]
+    (async/pipeline-async 2
+                          out-ch
+                          (fn [target-expr ch]
+                            (if (map? target-expr)
+                              ;; maps are query where clauses
+                              (let [context (get target-expr "@context")
+                                    sid-xf  (map #(json-ld/expand-iri % (json-ld/parse-context context)))
```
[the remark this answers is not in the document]
dpetran: Each query may have a different context, there's no guarantee they will all be the same.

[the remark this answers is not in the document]
zonotope: Isn't the same context bound in the transducer? If there is a different query with a different context, how is this transducer being redefined to pick up the new context?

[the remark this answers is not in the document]
dpetran: Realistically there will only be one query, though, so it will only happen once.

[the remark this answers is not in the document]
dpetran: Sorry, I should have refreshed before replying. There's a new transducer `sid-xf` per query that is used to turn iris into sids, and it closes over the context from the `target-expr`, which is a query map.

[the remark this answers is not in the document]
zonotope: Yes. The context it closes over is parsed repeatedly inside that transducer, once for each item that passes through the transducer. If you parse the context first and close over that, you would only parse the context (possibly an expensive operation) once instead of however many items are returned by the query.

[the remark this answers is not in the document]
dpetran: Oh, I understand what you mean now. Very good point, I'll pull that out.

PR #958, opened 2025-01-27
Improve accuracy of garbage reporting/exceptions

bplatz:
When cleaning garbage, the logging shows inaccurate garbage cleaning reports and with 'nil' values reported and inaccurate garbage collection counts.

This is because the oldest index root file will point to an even older garbage file which was already cleaned by the prior garbage collection. This is not the case in the first run of garbage collection, but only subsequent runs.

Not only is the count wrong, but the logging reports the data being removed for each garbage file - and when the file is gone, it reports a number of 'nil' values - which appears there is some sort of system error.

This now checks the oldest garbage file(s) for existence before reporting the garbage collection count so reporting is accurate.

### DIFF
--- a/src/clj/fluree/db/indexer/garbage.cljc
+++ b/src/clj/fluree/db/indexer/garbage.cljc
@@ -33,7 +33,8 @@
              garbage*      (if garbage-meta
                              (conj garbage garbage-meta)
                              garbage)]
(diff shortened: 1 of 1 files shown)


zonotope (approved):
🧹

PR #960, opened 2025-02-04
don't tally fuel unless a limit is supplied

dpetran:
Just a quick PR for a minor perf improvement I found.

If we don't have a fuel `:limit`, there's no reason to tally the fuel cost of every concurrent range scan. This should save a little bit of work on every flake when we have a fuel tracker.


zonotope (approved):
🔒

PR #962, opened 2025-02-05
fix match-id implementation

dpetran:
We were passing the args to `where/commit-sid` in the wrong order, causing the range scan to return all subjects, not just the one supplied.


zonotope (approved):
🏟️

PR #963, opened 2025-02-07
handle nil value during normalization

dpetran:
We need to return a `nil` normalized values, otherwise we get an empty values pattern `[[] []]`, which won't match anything. Tests are taken from Andrew's `server` `bug/targetSubject-with-opts-query` branch.


aaj3f:
@dpetran -- sorry to bypass our normal process here, but because this is a very small code update and because it passes all our external integration tests for policy, I'm going to merge so we can get an image of `server` deployed

PR #965, opened 2025-02-08
Load latest index from nameservice, not commit

bplatz:
When loading an existing db, the index was always pulled from the latest commit file which did not reflect the latest index (if one was written since the latest commit).

The latest index is stored in the nameservice, and this now relies on the nameservice latest information rather than the latest commit file.

There were also some bugs I discovered in reading the serialized index file - not sure how things were working when relying on indexes on disk (my guess is they were not, if there were subclasses, etc.)


zonotope (approved):
🐗

PR #968, opened 2025-02-13
Return full commits from remote nameservices

zonotope:
#965 changed the `lookup-commit` nameservice api method to return the full latest commit document instead of just the address of the latest commit.  However, the commit document returned was expanded in our proprietary format instead of expanded or compact jsonld, and this change wasn't implemented for remote nameservices. 

This patch extends on #965 to return compact jsonld from the `lookup-commit` api method, and it changes the behavior of remote nameservices to match storage-based and ipns based nameservices.

I thought about fixing these issues while keeping the original goal of #965 by making changes to stop storing the index in the commit document, and instead store indexes and commits separately while adding another nameservice api method to retrieve the latest index. I think this is what we should do eventually as it will simplify the system and prevent other issues, but I chose to do the most expedient thing right now to fix the immediate problem.

### DIFF
--- a/src/clj/fluree/db/connection.cljc
+++ b/src/clj/fluree/db/connection.cljc
@@ -362,23 +362,22 @@
   [{:keys [commit-catalog index-catalog] :as conn}
    ledger-chan address]
   (go-try
-    (let [commit       (<? (lookup-commit conn address))
-          _            (if-not commit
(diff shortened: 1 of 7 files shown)


dpetran (approved):
:pie:

PR #970, opened 2025-02-14
Old predefined shape cache prevented shacl speedup

bplatz:
PR #966 unlocked great speed gains when SHACL wasn't defined, but an old pre-defining of the shape cache made it look like shapes were present where there were none.

This fixes that, and unlocks the speed.


dpetran (approved):
:racing_car:

PR #971, opened 2025-02-16
Allow 'bytes' configs to accept long values

bplatz:
We have a few settings that take 'bytes', but with large values (over 2.1GB) it won't work as it coerces to an Integer.

This coerces instead to a Long.


zonotope (approved):
🥫

PR #973, opened 2025-02-19
SPARQL Query Results JSON format

dpetran:
This adds the ability to produce query results in [SPARQL 1.1 Query Results JSON Format](https://www.w3.org/TR/sparql11-results-json/). 

https://github.com/fluree/core/issues/170

This introduces the `:output` query option. If it is not supplied it defaults to `:fql`, which is what we currently return. Supplying `:sparql` provides the output like this:
```
{
  "head" : {
    "vars" : [ "favNum", "handle" ]
  },
  "results" : {
    "bindings" : [ {
      "handle" : {
        "value" : "bbob",
        "type" : "literal"
      },
      "favNum" : {
        "value" : "23",
        "type" : "literal",
        "datatype" : "http://www.w3.org/2001/XMLSchema#integer"
      }
    }, {
      "handle" : {
        "value" : "jdoe",
        "type" : "literal"
      },
      "favNum" : {
        "value" : "42",
        "type" : "literal",
        "datatype" : "http://www.w3.org/2001/XMLSchema#integer"
      }
    }, {
      "handle" : {
        "value" : "jdoe",
        "type" : "literal"
      },
      "favNum" : {
        "value" : "99",
        "type" : "literal",
        "datatype" : "http://www.w3.org/2001/XMLSchema#integer"
      }
    } ]
  }
}
```
Note that it retains the datatype and language metadata about values, as well as distinguishes IRIs from strings. It's also a tabular format, so aggregates are "unnested" and flattened.


dpetran:
There's still room for further cleanup, especially around the `select/display` abstraction, but I want to leave it as-is while I tackle `CONSTRUCT` support, which will introduce a new case to display results as json-ld. I consider this a working but still immature abstraction.

src/clj/fluree/db/query/exec/select.cljc
```diff
@@ -15,30 +15,65 @@
...
+                               []
+                               results)]
+          (recur r results*))
+        results))))
+
 (defmulti display
   "Format a where-pattern match for presentation based on the match's datatype.
   Return an async channel that will eventually contain the formatted match."
-  (fn [match _compact]
+  (fn [match output-format _compact]
```
zonotope, line 42: This new `output-format` binding exists only to provide runtime polymorphism. This implementation also litters this `output-format` argument throughout the call stack forcing already working existing code to be modified to support this new feature.

 We should use the built in language facilities to support run time polymorphism instead of rolling our own. Either new implementations of existing protocols/multi-methods, or a new protocol or multimethod would be a better fit.

dpetran: My current plan is to change the `where/match` into a record and implement a display method for it. I've got it work-in-progress for `construct` support, but didn't want to block Nexus and CAM work while I was still figuring out the proper abstraction.

dpetran:
Ok, I've overhauled the approach here. Instead of checking the output type each time we format a result, we now just check during query parsing and add the correct implementation of `format-value` for the output type.

dpetran:
Rebased on `main`

dpetran:
Rebased on main.

src/fluree/db/query/exec/select.cljc
```diff
@@ -15,29 +16,8 @@
...
-(defmethod display const/iri-id
-  [match compact]
-  (some-> match where/get-iri compact))
-
-(defmethod display const/iri-vector
-  [match _compact]
-  (some-> match where/get-value vec))
-
 (defprotocol ValueSelector
+  :extend-via-metadata true
```
zonotope, line 20: I think this namespace is convoluted by relying on this, and I don't see any benefit.

I think instead, there should be two additional namespaces: fluree.db.query.exec.select.fql, and fluree.db.query.exec.select.sparql for the different output formats, each with their own implementations of the `ValueSelector` protocol that don't reference each other.

Then, during parse time, you can read the output format once and decide which `ValueSelector` records to instantiate based on the select clause.

src/fluree/db/query/exec/select/display.cljc
```diff
@@ -0,0 +1,65 @@
+(ns fluree.db.query.exec.select.display
+  (:require [fluree.db.constants :as const]
+            [fluree.db.query.exec.where :as where]
+            [fluree.db.util.json :as json]))
+
+(defmulti fql (fn [match compact] (where/get-datatype-iri match)))
```
zonotope, line 6: I think this should also be split. Instead of this new namespace with these two multimethods, we should instead put the `display/fql` multimethod in `fluree.db.query.exec.select.fql` and the sparql multimethod in the sparql ns.

Essentially, we should take everything except for the protocol definitions from `fluree.db.exec.select.fql` and move them to a new namespace `fluree.db.query.exec.select.fql`, and then define a parallel `fluree.db.query.exec.select.sparql` namespace with the sparql functionality.

src/fluree/db/query/api.cljc
```diff
@@ -122,17 +123,28 @@
         (<? (track-query ds* max-fuel query**))
         (<? (fql/query ds* query**)))))))
 
+(defn wrap-sparql-results
+  [results]
+  (let [format (fn [results] {"head" {"vars" (vec (sort (keys (first results))))}
```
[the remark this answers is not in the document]
dpetran: It was to keep the function definition local. This entire function is obsolete, in the CONSTRUCT PR (https://github.com/fluree/db/pull/985) I move all the wrapping into the collect-results phase of query execution.

src/fluree/db/query/api.cljc
```diff
@@ -266,9 +278,12 @@
...
   [conn query override-opts]
   (go-try
-    (let [fql (sparql/->fql query)]
-      (log/debug "query-connection SPARQL fql: " fql "override-opts:" override-opts)
-      (<? (query-connection-fql conn fql override-opts)))))
+    (let [fql     (sparql/->fql query)
+          _       (log/debug "query-connection SPARQL fql: " fql "override-opts:" override-opts)
+          results (<? (query-connection-fql conn fql override-opts))]
+      (if (= (:output override-opts) :sparql)
+        (wrap-sparql-results results)
```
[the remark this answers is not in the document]
dpetran: No, and this is made possible in the CONSTRUCT PR by moving the wrapping into the collect-results phase of query execution.

src/fluree/db/query/exec/select.cljc
dpetran: I chose `:extend-via-metadata` because the other protocols the records implement (`ValueAdapter`, `SolutionModifier`) aren't affected by the display and if I create separate `Selector` records for FQL vs SPARQL I'd have to move the common implementations to a third, common namespace or duplicate them, and on the balance it seemed that `:extend-via-metadata` was the simpler choice.

src/fluree/db/query/api.cljc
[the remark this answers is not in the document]
dpetran: I'll merge that PR into this one so the whole chain is visible.

src/fluree/db/query/exec/select.cljc
```diff
@@ -68,38 +39,29 @@
...
-              (recur vars (assoc result var display-var)))
-            result))
-        (catch* e
-                (log/error e "Error formatting wildcard")
-                (>! error-ch e)))))
   ValueAdapter
-  (solution-value
-    [_ _ solution]
-    solution))
+  (solution-value [_ _ solution] solution))
```
zonotope, line 50: I think the argument list and function definitions should both be on new lines.

src/fluree/db/query/exec/select.cljc
```diff
@@ -68,38 +39,29 @@
...
+  (let [selector (->WildcardSelector)]
+    (case output
+      :sparql (with-meta selector {`format-value select.sparql/format-wildcard-selector-value})
+      (with-meta selector {`format-value select.fql/format-wildcard-selector-value}))))
 
 (defrecord AggregateSelector [agg-fn]
   ValueSelector
   (format-value
-    [_ _ _ _ _ _ error-ch solution]
+    [_ _ _ _ compact _ error-ch solution]
```
zonotope, line 64: It doesn't look like `compact` is used.

src/fluree/db/query/exec/select.cljc
zonotope: Ok. I'm worried about the performance implications of extending via metadata as well as the semantic implications of having opaque functions hanging off of records. 

This pr has been held open for long enough so I'll let this through and we can address these concerns later if they prove to be a problem.

zonotope (approved):

src/fluree/db/query/fql/parse.cljc
```diff
@@ -22,7 +22,8 @@
   "Returns x as a symbol if x is a valid variable, or nil otherwise. A valid
   variable is a string, symbol, or keyword whose name starts with '?'."
   [x]
-  (when (v/variable? x)
+  (when (or (v/variable? x)
```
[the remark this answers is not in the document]
dpetran: We actually need to distinguish between the two types - in a transaction you can have bnodes and they are _not_ variables, they are "fresh" identifiers that will be mapped to new, internal identifiers. So it's only inside of a query where pattern that we need these `or` statements.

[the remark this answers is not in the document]
zonotope: Ok, so I think it makes sense to have a general function for detecting variables in queries to remove the `or` repetition here.

I see this branch is already merged, so we can address this later.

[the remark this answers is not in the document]
dpetran: Here's a follow-up PR to address it:
https://github.com/fluree/db/pull/989

PR #974, opened 2025-02-20
Multiple storage config

zonotope:
This patch allows connections to use commit and index storage catalogs backed by multiple different storage back ends.


dpetran (approved):
:leaves:

PR #975, opened 2025-02-21
Refactor/sid

bplatz:
This removes the Java SID class in favor of deftype.

It improves transaction speeds by about 20%, but also has a substantial impact on indexing times - for large imports where indexing can be the limiting time factor, it can improve import times by around 50%. The larger the import the more improvement.

There are other performance improvement tweaks that are possible here and I tried many of them, but our use of SIDs as constants creates many hurdles (that print-dup doesn't address on its own) and I decided to take what's here as a solid win for now.

This change also means fluree/db no longer requires a compile / prepare step.

### DIFF
--- a/src/clj/fluree/db/json_ld/iri.cljc
+++ b/src/clj/fluree/db/json_ld/iri.cljc
@@ -77,16 +77,19 @@
   #?@(:clj  [Object
              (equals [this sid]
                (sid-equiv? this sid))
-             (hashCode [this] (hash (seq this)))
+             (hashCode [_]
+               (clojure.lang.Util/hashCombine namespace-code (hash name)))
 
              clojure.lang.IHashEq
-             (hasheq [this] (hash (seq this)))
+             (hasheq [_]
+               (clojure.lang.Util/hashCombine namespace-code (hash name)))
 
              java.lang.Comparable
              (compareTo [this other] (compare-SIDs this other))]
 
-             (-hash [this] (hash (seq this)))
       :cljs [IHash
+             (-hash [_]
+                    (hash-combine namespace-code (hash name)))
 
              IEquiv
              (-equiv [this sid] (sid-equiv? this sid))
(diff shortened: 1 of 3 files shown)


src/clj/fluree/db/query/exec/eval.cljc
```diff
@@ -382,6 +382,7 @@
        :cljs (js/Date.))
     const/iri-xsd-dateTime))
 
+;; TODO - date functions below all look incorrect for CLJS - should (string? datetime) be (map? datetime)?
```
dpetran, line 385: These have all been fixed here: https://github.com/fluree/db/pull/915

I need to cherry pick those out into a separate PR.

src/clj/fluree/db/json_ld/iri.cljc
```diff
@@ -70,6 +68,101 @@
...
+;;
+(deftype SID #?(:clj [^int namespace-code name]
+                :cljs [^number namespace-code name])
+  #?@(:clj  [clojure.lang.Seqable
+             (seq [this] (list (.-namespace-code this) (.-name this)))
+
+             Object
+             (equals [this sid]
+               (sid-equiv? this sid))
+             (hashCode [this] (hash (seq this)))
```
zonotope, line 83: I think we could implement this function with `(hash (list namespace-code name))`, then we wouldn't have to implement `Seqable`, unless there's some other reason.

bplatz: I think you are right... I just copied over the same pattern we did for Flake.

bplatz: @zonotope I replaced it with a more optimized hashing and looks like another ~5% txn performance improvement - at least for large transactions.

zonotope (approved):
This is great. I was trying to think of a way to hash a sid without a new `list` allocation that would just be immediately garbage collected, but I didn't know about `hashCombine`. Nice find!

PR #976, opened 2025-02-21
Speed up flake size calc

bplatz:
This assumes a consistent SID size for both subject and property so they don't need to be constantly calculated when figuring out DB size.

Small change, but this gets called a lot, especially for large transactions.

This change speeds up the size calc a good amount, and as it relates to an overall transaction time, it speeds those up by ~3%.

I'd like to see if we can just take a commit/data file size as a db-size... and if pulling those values is faster than this calculation. That might be easier said than done, but I'll plan to try it.


src/clj/fluree/db/flake.cljc
```diff
@@ -494,6 +494,12 @@
   [ss]
   (->> ss rseq first))
 
+(def ^:const flake-size-base (+ 30 ;;sid
```
dpetran, line 497: sids are variable in size, are we just ignoring that? We use size to determine when to kick off indexing, so adding fudge to this may have knock-on effects.

dpetran: Wait, I see there's another PR that changes SIDs, disregard this until I review that :o

bplatz: That was my thought, but certainly open to debate.

I am also exploring the impact of interning SIDs, and then storing their size with them. This approach would eliminate the constant re-calculations of the same thing(s) but also solve your concern.

But my view is that once we extract the namespace, the "name" portion of the SID is going to be in a fairly narrow size range a very high percentage of the time.

If that's true, then getting a speed bump is worth it. If that is not true, then perhaps it is not worth it.

dpetran: I don't know that we can actually know what the distribution of "name" sizes is going to be, and I think it could create a very confusing user experience if the measurement is inaccurate, with users with not-average-sized flakes having indexing triggered at unpredictable times.

I really like the idea of of interning the size right on the SID, though, that seems like the best of both worlds. Or maybe just changing the entire idea of novelty size to be number of flakes? But I think novelty is meant to be tuned to hardware, so having some analog to actual physical resource usage seems important.

zonotope (approved):
I think the performance benefit is worth the decreased precision of the new size calculation, but it might lead to memory issues for large databases where the max novelty is set too close to the maximum available memory on the machine. We'll have to monitor how this change impacts users and possibly revisit if necessary.

PR #977, opened 2025-02-24
Fix/cljs update

dpetran:
A collection of cljs fixes from https://github.com/fluree/db/pull/915

### DIFF
--- a/src/clj/fluree/db/query/exec/eval.cljc
+++ b/src/clj/fluree/db/query/exec/eval.cljc
@@ -419,7 +419,7 @@
        :cljs
        (.getHours (if (string? x)
                     (datatype/coerce (:value x) (:datatype-iri x))
-                    datetime)))))
+                    x)))))
 
 (defn minutes
   [x]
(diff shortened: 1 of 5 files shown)


zonotope (approved):
🫃

src/clj/fluree/db/query/exec/eval.cljc
```diff
@@ -454,9 +454,9 @@
               (.toString (.getOffset ^OffsetDateTime (->offset-date-time (:value x))))
               #{const/iri-xsd-time}
               (.toString (.getOffset ^OffsetTime (->offset-time (:value x)))))
-       :cljs (.getTimeZoneOffset ^js/Date (if (string? datetime)
-                                            (datatype/coerce (:value datetime) (:datatype-iri datetime))
-                                            datetime)))))
+       :cljs (.getTimeZoneOffset ^js/Date (if (string? x)
+                                            (datatype/coerce (:value x) (:datatype-iri x))
+                                            x)))))
```
bplatz, line 459: I think above still has an error in it...

If `x` is a `string?`, then I don't think `(:value x)`, etc. would work.

Should it be `(map? x)`, or maybe `(string? (:value x))`?

dpetran: Good catch, it should be `(string? (:value x))`

PR #978, opened 2025-02-24
handle TypedValue nesting in tz fn

dpetran:
Fix `tz` handling of TypedValue in cljs.


bplatz:
This same issue affects all the date related functions... not just the one you fixed. There are several fns right above it with the same issue.

dpetran:
Good catch, Brian, I was rushing and not paying attention. I've updated the other datetime functions now.

PR #981, opened 2025-02-27
update load process for parallel data loads, commit meta updates

bplatz:
When doing a db `load` from disk, this now traces just the commits to find the first commit after index, but doesn't yet load the db 'data' in that first pass.

The db 'data' is then loaded for each new commit processed.

This accomplishes two things:
1) we don't need to hold as much data in memory at once, which has a significant memory impact for large commits
2) We can load the data and json-ld/expand in parallel using an async chan... so this parallelizes some of the work and reduces db load times in my tests by about 15-20%.

In addition, I consolidated much of the 'commit metadata flake' generation. This was called from 3 places, each doing much of the same work and was not all done consistently.

Lastly, I removed the "prev-data" flake. We were storing `:f/previous` for both the commit ID and the db-data ID. The latter is not necessary as you can get to it easy enough from  the previous commit (and commits always points to its data).

In our load processing, creating this flake was doing another redundant commit load from disk, further adding time to load process for no obvious value.

### DIFF
--- a/src/clj/fluree/db/commit/storage.cljc
+++ b/src/clj/fluree/db/commit/storage.cljc
@@ -1,10 +1,12 @@
 (ns fluree.db.commit.storage
   (:require [clojure.string :as str]
+            [clojure.core.async :as async :refer [go <! chan]]
             [fluree.db.constants :as const]
             [fluree.db.storage :as storage]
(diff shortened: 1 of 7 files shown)


zonotope (approved):
🍸

PR #982, opened 2025-02-27
Only a single 'staged' db per commit

bplatz:
There was a partial assumption multiple staged dbs could be in a single commit - but this is not something we have supported (although v1/v2 supported multiple txns in a block).

It is the case that you can 'stage' multiple times, but those stages get merged into a single commit.

There were some code assumptions that multiple stages could exist..

This removes the code assumptions and cleans things up a little bit.


zonotope (approved):
🥌

PR #983, opened 2025-02-28
Refactor/fluree ns tx

bplatz:
This removes ns.flur.ee context from transaction processing.

It required an extra json-ld/expand, and complected the processing.

Also removed from all tests.

### DIFF
--- a/src/clj/fluree/db/query/fql/parse.cljc
+++ b/src/clj/fluree/db/query/fql/parse.cljc
@@ -672,25 +672,10 @@
     (assoc q :fuel max-fuel)
     q))
 
-(defn extract-values
-  [q]
-  (or (get q "values")
-      (get q :values)))
-
-(defn extract-where
-  [q]
-  (or (get q "where")
-      (get q :where)))
-
-(defn extract-delete
-  [q]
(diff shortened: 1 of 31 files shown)


src/clj/fluree/db/query/fql/parse.cljc
```diff
@@ -672,13 +672,40 @@
     (assoc q :fuel max-fuel)
     q))
 
+(defn extract-values
```
zonotope, line 675: This pattern is repeated a lot in this pr. Perhaps we could generalize it with something like this:

```clojure
(defn get-named
  [jsonld nme]
  (or (get jsonld nme)
      (get jsonld (keyword nme))))

(let [values (get-named jsonld "values")]
  ...)
```

zonotope (approved):
I'm very happy for this change

src/clj/fluree/db/query/fql/parse.cljc
bplatz: Resolved in [5528eef](https://github.com/fluree/db/pull/983/commits/5528eef213b5e65bb88e7c9cbdd1351ad7c63dc5)

PR #984, opened 2025-02-28
Take db-id, commit-id from storage write hash

bplatz:
In the commit process, we were serializing the JSON just to get a hash, then updating the data, then on writing to storage we get another hash.

This change skips the first step and only uses the latter has for the db-id, commit-id.

The JSON serialization is the the most time consuming part of the 'commit' process - and it was being done twice. Now, just once.

The storage serialization hash is in may ways the better hash, as one can verify just from bytes on disk. Before, you'd have to read the bytes, deserialize them, dissoc the @id, re-serialize, and re-hash for verification.

The downside is the @id value is no longer present in the written files - but they are present by anything wrapping them (e.g. db-id is present in the commit).

There are some proofs (some VC proofs) that require specific serialization rules where neither method would have worked anyhow - so if/when those are supported we can encode the logic and pay the penalty for those at that point. Regardless they'd always be optional, and people who preferred a faster DB vs a special proof wouldn't have to pay a penalty.

### DIFF
--- a/src/clj/fluree/db/connection.cljc
+++ b/src/clj/fluree/db/connection.cljc
@@ -448,19 +448,27 @@
   [(assoc commit-map :address commit-address)
    (assoc commit-jsonld "address" commit-address)])
 
+(defn update-commit-id
+  "Once a commit address is known, which might be after the commit is written
+  if IPFS, add the final address into the commit map."
+  [[commit-map commit-jsonld] commit-hash]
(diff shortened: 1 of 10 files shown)


bplatz:
Also moved everything to base32 hashing. We were using base16 (hex) for memory storage and as a default file storage hash.

zonotope (approved):
🪮

PR #985, opened 2025-02-28
Support for SPARQL CONSTRUCT clauses

dpetran:
Return an RDF graph result using `CONSTRUCT`, according to the [SPARQL spec](https://www.w3.org/TR/sparql11-query/#construct)

https://github.com/fluree/core/issues/170

The specification does not specify how compact the results should be, so I've elected to "flatten" the data so that all of a subject's predicates are grouped into one json object. This does come at a performance penalty, so we could elide that step and still be compliant, but I figured this way is perhaps more usable than the alternative, which is each triple produces a unique object.

I've also elected to treat all objects as multicardinal values returned in arrays, even if there's only a single item. We could do the extra step of unpacking single values, but that complicates result processing and I elected not to do it unless someone asks for it, in which case I would be happy to add it.


test/fluree/db/query/construct_test.clj
```diff
@@ -0,0 +1,107 @@
...
+             @(fluree/query db1 {"@context" context
+                                 "where" [{"@id" "?s" "person:fullName" "?fullName"}
+                                          {"@id" "?s" "person:friend" "?friend"}
+                                          {"@id" "?friend" "person:fullName" "?friendName"}
+                                          {"@id" "?friend" "person:favNums" "?friendNum"}]
+                                 "construct" [{"@id" "?s" "myname" "?fullName"}
+                                              {"@id" "?s" "friendname" "?friendName"}
+                                              {"@id" "?friend" "name" "?friendName"}
+                                              {"@id" "?friend" "num" "?friendNum"}]}))))
+    #_(testing "bnode template"
```
dpetran, line 86: Using bnodes in construct templates produces nondeterministic responses since we process results in parallel. I don't have a good solution for this yet, so the flaky tests are commented out.

src/clj/fluree/db/query/exec/select.cljc
```diff
@@ -268,3 +283,17 @@
...
       (and (instance? AsSelector selector)
            (:aggregate? selector))))
+
+(defn wrap-construct
+  [{:keys [orig-context context]} results]
+  (let [id-key (json-ld/compact const/iri-id context)]
+    (cond-> {"@graph" (->> results
+                           (sort-by #(get % id-key))
+                           (partition-by #(get % id-key))
+                           (mapv display/nest-multicardinal-values))}
```
dpetran, line 293: This sorting, partitioning, and nesting is not strictly required by the spec, but may greatly improve usability. If users are mainly using `construct` to directly inject the results into another graph processor, then this is probably unnecessary. If they are trying to use the json-ld document directly, this is probably very useful. Until we discover which use case is more common I've tried to strike a middle path.

dpetran:
Rebased on main.

zonotope (approved):
🎩

dpetran:
Updated to account for the removal of the `display` ns in https://github.com/fluree/db/pull/973

PR #986, opened 2025-03-11
Refactor/config parsing

zonotope:
The changes in this patch were necessary to better integrate with fluree/server as well as other hosted system. They include:

- Refactors config parsing to make the config system more extensible and apply to other contexts including server better.
- Fixes a few bugs in initialization of ipns name services due to asynchronous code in a synchronous context
- Cleans up some dev namespaces whose code no longer worked because of api changes
- Adds a method to notify connections of a new commit for a cached ledger by supplying a commit address instead of the commit map
- removes the `clj/` source path component since it's no longer necessary after removing `java/`


dpetran (approved):
:broom:

PR #987, opened 2025-03-12
use go-try block

kurtharriger:
Noticed that if an exception is thrown here  the promise will never resolve and the request will hang. 
This should use `go-try`


zonotope (approved):
🦀

PR #988, opened 2025-03-12
add log with-mdc

kurtharriger:
This allows attaching arbitrary key value pairs to be written in log files.  When logs are emitted as json then these properties will show up in the json and then search on these properties in cloudwatch log insights. 

To display these properties in not json log fromat you can use %X{propname} eg %X{tx.id} or %X to dispaly all mdc properties.  I use this pattern locally:

```
<appender name="CONSOLE_TRACE_COLORS" class="ch.qos.logback.core.ConsoleAppender">
    <encoder>
      <pattern>%d{HH:mm:ss.SSS} %highlight(%-5level) %cyan(%logger %t %X)\n  %msg%n</pattern>
    </encoder>
  </appender>
```
which renders as follows 
Screenshot 2025-03-12 at 11 58 41 AM

Note that mdc properties are stored on thread locals and thus not copied to go blocks or background threads.  For this to work in practice you also need to require this file to monkey-patch core.async thread executor so that it saves and restores the thread locals when the thread changes. 

https://github.com/fluree/nexus-server/commit/8ba2d7d2228151939d32c837950afcf09e22d28b#diff-684bac090a01548d7d3450bb82c9336370e3bba2b9b645a956208d35ef6827dc

Perhaps the above change should also be moved to db.  However it also captures the otel context and requires additional dependencies, but maybe this isn't so bad if we also want to add tracing and spans in db code as well.

### DIFF
--- a/src/fluree/db/util/log.cljc
+++ b/src/fluree/db/util/log.cljc
@@ -135,4 +135,27 @@
      [msg c]
      `(debug-async->vals ~c ~msg)))
 
+#?(:clj
+   (defn with-mdc* [context f]
+     (let [original (org.slf4j.MDC/getCopyOfContextMap)]
+       (try
+         (doseq [[k v] context]
+           (org.slf4j.MDC/put (name k) (str v)))
+         (f)
+         (finally
+           (org.slf4j.MDC/clear)
+           (when original
+             (doseq [[k v] original]
+               (org.slf4j.MDC/put k v)))))))
 
+   :cljs
+   (defn with-mdc* [_context f]
+     (f)))
+
+#?(:clj
+   (defmacro with-mdc [context & body]
+     `(with-mdc* ~context (fn [] ~@body)))
+
+   :cljs
+   (defmacro with-mdc [_context & body]
+     `(do ~@body)))


src/fluree/db/util/log.cljc
```diff
@@ -135,4 +135,27 @@
...
+           (when original
+             (doseq [[k v] original]
+               (org.slf4j.MDC/put k v)))))))
+
+   :cljs
+   (defn with-mdc* [_context f]
+     (f)))
+
+#?(:clj
+   (defmacro with-mdc [context & body]
```
zonotope, line 156: Could you post an example of how this should be used?

kurtharriger: https://github.com/fluree/nexus-server/pull/3/files#diff-fe046dc3eec2629e801daaa120b65f8049c49cc18ec59481e48ee8c23a89d1acR17

zonotope: In this example, does the `:ledger.id` and `:tx.id` values persist from that single call, or do we have to repeat it?

If we did this:
```clojure
(log/with-mdc {:ledger.id ledger-id :tx.id tx-id}
  (log/debug "begin ..."))

(log/debug "continuing ...")

(log/debug "end ...")
```
then will the `:ledger.id` and `:tx.id` values show up in the "continuing ..." and "end ..." logs?

kurtharriger: No.  It works more like binding when the with-mdc ends the values are removed. Since they are associated with the thread this ensures they get removed when the context such as a request is finished.

zonotope: sorry for losing track of this. So `log/with-mdc` is really meant to wrap all the top level api calls, and then they will automatically add the context to any log calls from any other functions within the call stack. Makes sense.

We probably want to wrap the functions in `fluree.db.api` with `log/with-mdc`, but that can come later.

zonotope (approved):
🍬

PR #990, opened 2025-03-17
Add cljfmt and clj-kondo

zonotope:
This patch adds CI steps for both [clj-kondo]() and [cljfmt](), and fixes all the errors and warnings flagged by those tools. 

Going forward after this patch is merged, CI will fail if either of those tools report any issues. [clj-kondo integrates with all the major Clojure editors](https://github.com/clj-kondo/clj-kondo/blob/master/doc/editor-integration.md), so setting it up provides the most seamless experience.

[cljfmt also integrates with the major editors](https://github.com/weavejester/cljfmt?tab=readme-ov-file#editor-integration) including [Intellij](https://github.com/eunmin/intellij-cljfmt-plugin), but I think it's easier to configure your editor to match cljfmt automatically I've pasted the configuration I use for Emacs below, and [here is a configuration for Intellij](https://clojureverse.org/t/cursive-how-to-match-cljfmt-formatting/7246)
```lisp
(define-clojure-indent
 ;; compojure
 (defroutes 'defun)
 (GET 2)
 (POST 2)
 (PUT 2)
 (DELETE 2)
 (HEAD 2)
 (ANY 2)
 (context 2)
 (go-try 0)
 (try* 0)
 (catch* 1))
```

I have also included an extra Make task to automatically reformat all Clojure code. Invoke it with `make cljfmt-fix`.

There are a lot of changes here and they could be disruptive, so I will hold off on merging this until I get explicit approval from @bplatz  and @dpetran as they may have unmerged branches which will have many conflicts after this branch is merged.


src/fluree/db/datatype.cljc
```diff
@@ -192,12 +195,14 @@
   - JS: a Javascript Date object. NB: If you don't supply a timezone JS will
         assume it's in your current, local timezone according to your device."
   [s]
-  (when-let [matches (re-matches iso8601-datetime-re s)]
-    #?(:clj
-       (if-let [offset         (peek matches)]
+  #?(:clj
```
dpetran, line 198: Is top-level-only reader conditionals configurable? My instinct is to push the reader conditionals as far down as we can so as much of the AST is shared between platforms as possible. Since we already have introduced bugs where they diverge, I'd like to minimize the surface area.

src/fluree/sdk/webworker.cljs
```diff
@@ -1,417 +1,385 @@
...
-
-(defn init
-  []
-  ;; when we detect that we are in a web worker, register the onmessage handler
-  (when-let [worker? (not (.-document js/self))]
-    (set! (.-onerror js/self) log-error)
-    (set! (.-onmessage js/self) decode-message))
-  ;; let server know we're alive
-  (worker-action nil :connInit))
+(ns fluree.sdk.webworker)
```
dpetran, line 1: Are we done with the webworker for good or just temporarily?

dpetran:
This comes at a good time - my work is confined to three files, one of which isn't clojure code. So keeping up with the linter will be easy for me.

dpetran (approved):
:grapes:

src/fluree/sdk/webworker.cljs
zonotope: I think just temporarily. I don't think it was ever updated post v3, so there were many out of data apis and lots of functions that failed the linter because they just returned `:TODO`. I decided to comment out the code instead of totally deleting it so we'd have a reference whenever we get around to re implementing this.

Makefile
```diff
@@ -116,12 +118,24 @@ pending-tests:
...
+cljfmt-fix:
+	cljfmt fix src dev test build.clj
+
 test: cljtest cljstest nodejs-test browser-test
 
 eastwood:
 	clojure -M:dev:cljtest:eastwood
 
-ci: test eastwood
+ci: test eastwood clj-kondo-lint-ci cljfmt-check
```
dpetran, line 138: We should add `cljfmt` and `clj-kondo` to the Prerequisites section of the readme so others will be able to contribute.

bplatz (approved):

PR #991, opened 2025-03-18
properly format json-ld values with metadata

dpetran:
Fixes the return value of a `construct` query to properly format json-ld values with metadata.


zonotope (approved):
🚘

PR #992, opened 2025-03-18
Feature: SPARQL Graph Update

dpetran:
Add support for SPARQL Graph Update syntax: https://www.w3.org/TR/2013/REC-sparql11-update-20130321/#graphUpdate

Since we only allow transacting to a single ledger at a time, we do not support GRAPH patterns or multiple USING/USING NAMED clauses.

https://github.com/fluree/core/issues/175


dpetran:
Converting to draft while I add the bits necessary to integrate with server.

dpetran:
This is now ready for review with all the server bits finished.

src/fluree/db/api/transact.cljc
```diff
@@ -28,11 +29,14 @@
 (defn stage
   [db txn opts]
   (go-try
-   (let [txn-context (or (ctx-util/txn-context txn)
-                         (:context opts))
-         parsed-opts (parse-opts txn opts txn-context)
-         parsed-txn  (q-parse/parse-txn txn txn-context)]
-     (<? (connection/stage-triples db parsed-txn parsed-opts)))))
+    (let [txn*        (if (= :sparql (:format opts))
```
zonotope, line 32: i think it makes sense to convert this check into a function that takes the transaction and the format since it's used multiple times.

zonotope (approved):
🥨

PR #993, opened 2025-03-19
Feature/bm25 persist

bplatz:
This builds on the persistence work done in #942 by @zonotope 

This now is fully working with tests that check persistence state on load, and some other improvements.

BM25 can be used with this PR, but there is still work to be done:
1) Deleting a BM25 index
2) Preventing updates to BM25 index (in favor of delete/re-create)
3) additional stop word support for different languages
4) additional filters in addition to stemmer (e.g. convert to ascii characters before stemming)


src/fluree/db/query/exec/where.cljc
```diff
@@ -629,14 +629,21 @@
...
-      (match-clause graph fuel-tracker solution clause error-ch)
-      nil-channel)
-    (catch* e (async/offer! error-ch (ex-info (str "Error activating alias: " alias
-                                                   " due to exception: " (ex-message e))
-                                              {:status 400
-                                               :error  :db/invalid-query} e)))))
+  (let [res-ch (async/chan)]
+    (go
+      (try*
+        (if-let [graph (async/<! (-activate-alias ds alias))]
```
zonotope, line 635: I think we should use `<?` here just in case `-activate-alias` throws

src/fluree/db/query/exec/where.cljc
```diff
@@ -629,14 +629,21 @@
...
+          (-> (match-clause graph fuel-tracker solution clause error-ch)
+              (async/pipe res-ch))
+          (async/close! res-ch))
+        (catch* e
+                (log/error e "Error activating alias" alias)
+          (async/>! error-ch (ex-info (str "Error activating alias: " alias
+                                           " due to exception: " (ex-message e))
+                                      {:status 400, :error :db/invalid-query}
+                                      e))
+          (async/close! res-ch))))
```
zonotope, line 645: I don't think closing this channel matters because any put onto the error channel should short-circuit the whole operation.

src/fluree/db/virtual_graph.cljc
```diff
@@ -0,0 +1,36 @@
...
+  [vg]
+  (keyword (vg-type-name vg)))
+
+(defmulti write-vg
+          (fn [_index-catalog vg]
+            (vg-type-kw vg)))
+
+(defn storage-path
+  [vg-type-kw db-alias vg-alias]
+  (str/join "/" [db-alias (name vg-type-kw) vg-alias]))
```
zonotope, line 23: Why not just use `vg-type-name` here instead of taking that value, turning it into the keyword and then back into a string?

src/fluree/db/virtual_graph/bm25/index.clj
```diff
@@ -257,29 +256,27 @@
   (-match-class [_ _fuel-tracker _solution _s-mch _error-ch]
     where/nil-channel)
 
-  (-activate-alias [this _]
```
zonotope, line 260: I don't think it's a good idea to have partial protocol implementations. In this case, we should probably return an empty channel if the alias the caller is trying to activate doesn't match the alias of this graph.

zonotope (approved):
I have a few suggestions, but 👍🏾

src/fluree/db/virtual_graph.cljc
bplatz: In this case it is an internal VG type that is being used for a subdirectory... e.g. :bm25, :hnsw, etc. 

I renamed the arg keyword to just `vg-type` to minimize confusion with the function names above which this unintentionally was identical to.

vg-type-name is used when parsing original JSON-LD to extract a vg type.

PR #994, opened 2025-03-26
Notify addresses

zonotope:
This patch changes the `fluree/notify` api to take a commit address instead of a commit map. This allows us to keep commit documents outside of inter-server communication


src/fluree/db/commit/storage.cljc
```diff
@@ -34,28 +34,37 @@
...
-(defn read-data-jsonld
-  [storage address]
-  (go-try
-    (let [jsonld (<? (storage/read-json storage address))]
-      (-> jsonld
-          (assoc "f:address" address)
-          json-ld/expand))))
-
-(defn read-commit-jsonld
+(defn read-verified-commit
```
dpetran, line 37: What does "verified" mean in this context?

src/fluree/db/flake/flake_db.cljc
```diff
@@ -283,6 +284,8 @@
       (when (empty? all-flakes)
         (commit-error "Commit has neither assertions or retractions!"
                       commit-metadata))
+      (log/debug "Updating db" (str/join "@" [(:alias db) (:t db)])
```
dpetran, line 287: I'm not sure `str/join` really brings any benefit over `str` when we're always joining two strings.

dpetran (approved):
:bell: 

Looks good, looking forward to having all of these fixes land : )

src/fluree/db/commit/storage.cljc
zonotope: It means that the commit that is read has also been verified, and that the returned value includes the proof fro verification.

src/fluree/db/flake/flake_db.cljc
zonotope: They're about the same. `str/join` is just the function I prefer to use when joining any number of strings with the same delimiter.

src/fluree/db/commit/storage.cljc
dpetran: Does that mean we re-hash and compare? I'm not sure we actually do that anywhere currently.

PR #995, opened 2025-04-09
Fix/hydrate shape cache

dpetran:
When loading an existing ledger we need to hydrate the SHACL shape cache. If no cached shapes are found in a transaction that has no shape flakes, no validation will be performed.

### DIFF
--- a/src/fluree/db/flake/flake_db.cljc
+++ b/src/fluree/db/flake/flake_db.cljc
@@ -560,11 +560,11 @@
            indexed-db* (if (nil? (:schema root-map)) ;; needed for legacy (v0) root index map
                          (<? (vocab/load-schema indexed-db (:preds root-map)))
                          indexed-db)
-           index-t     (:t indexed-db*)]
-       (<? (shacl/hydrate-shape-cache!
-            (if (= commit-t index-t)
-              indexed-db*
-              (<? (load-novelty commit-catalog indexed-db* index-t commit-jsonld)))))))))
+           index-t     (:t indexed-db*)
+           loaded-db   (if (= commit-t index-t)
+                         indexed-db*
+                         (<? (load-novelty commit-catalog indexed-db* index-t commit-jsonld)))]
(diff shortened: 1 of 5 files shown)


dpetran:
Tests are currently in the server PR: I've got some tests in https://github.com/fluree/server/pull/130.

dpetran:
The CI linter revealed that we have no `-class-ids` implementation for an `AsyncDB`, so I added that. I've also added a `-contains-key?` implementation to `Flake` to flesh out the IAssociative protocol. 

I'm not sure why the linter started complaining now, but they were pretty straightforward.

src/fluree/db/flake/flake_db.cljc
```diff
@@ -560,9 +561,10 @@
                          (<? (vocab/load-schema indexed-db (:preds root-map)))
                          indexed-db)
            index-t     (:t indexed-db*)]
-       (if (= commit-t index-t)
-         indexed-db*
-         (<? (load-novelty commit-catalog indexed-db* index-t commit-jsonld)))))))
+       (<? (shacl/hydrate-shape-cache!
```
zonotope, line 564: I think you should add another `let` binding here instead of this deeply nested async call.

src/fluree/db/json_ld/shacl.cljc
```diff
@@ -1152,12 +1152,18 @@
...
     (= p-ns shacl-ns-code)))
 
-(defn extract-shapes
+(defn hydrate-shape-cache!
   [db]
   (go-try
     (let [new-shapes (<? (rebuild-shapes db))]
-      (reset-shape-cache! db new-shapes) ;; TODO - there is really no reason this needs to be an atom any longer
-      (vals new-shapes))))
+      (reset-shape-cache! db new-shapes)
```
zonotope, line 1159: I think you should change the `reset-shape-cache!` function to return the `db` it was passed in.

zonotope (approved):

PR #997, opened 2025-04-15
only store an unwrapped db in ledger state

dpetran:
Storing a policy wrapped db in the ledger branch state as :current-db can lead to situations where policy is applied to operations where it is not called for. In operations that should be wrapped we always call `wrap-policy` before conducting those operations, so clearing the policy before storage is just fine.

https://github.com/fluree/flhubee/issues/1432


zonotope (approved):
🎁

PR #998, opened 2025-04-17
Feature/policy fuel tracking

dpetran:
Adds fuel tracking to the many db lookups that didn't use a fuel-tracker:
- history queries
- policy targeting
- policy execution

Also removes some dead code ([f668bcc](https://github.com/fluree/db/pull/998/commits/f668bcc945d85389df4ad7f1ca1bc3ae1147e5e1)) and fixes what I think is a potential bug ([293fe10](https://github.com/fluree/db/pull/998/commits/293fe10945ea9311c32c6b51789a6e9bfba6c809))

Still remaining:
- query policy targeting
- SHACL

These two items will require more intensive work: 

For query policy targeting  we'll need to refactor how dbs are prepped for querying - right now they are restricted and then a fuel tracker is added, but in order to properly restrict we need the fuel tracker first. This points towards a need for a better way do all the db prep so the order can be composed more naturally.

For SHACL, there's just a million functions that will need the fuel-tracker argument added and I didn't want to hold up this work while I chug through them, so I think that can be a separate PR.


src/fluree/db/api.cljc
```diff
@@ -289,7 +288,7 @@
   ([db policy policy-values]
    (promise-wrap
     (let [policy* (json-ld/expand policy)]
-      (policy/wrap-policy db policy* policy-values)))))
+      (policy/wrap-policy db nil policy* policy-values)))))
```
zonotope, line 291: we should change `wrap-policy` (and any other function that takes optional arguments) into a multi-arity function so we don't have to pass `nil` like this.

PR #1000, opened 2025-04-21
SHACL Fuel Tracking

dpetran:
This adds fuel tracking to all SHACL validation queries. Rebuilding the shape cache when a new shape is transacted costs fuel and some SHACL constraints cost fuel because they require lookups. Validation is run against a root db so no policy execution costs are incurred.


kurtharriger (approved):

zonotope (approved):
📴

dpetran:
Rebase on main.

PR #1001, opened 2025-04-21
add fuel tracking to virtual graph search

dpetran:
I noticed we don't track fuel for vector searches, so I added support for that as well.


dpetran:
Rebased on main.

zonotope (approved):
🔦

PR #1002, opened 2025-04-22
Track metadata in result maps

zonotope:
This patch begins to unify fuel, policy, and file metadata tracking under a track- api (more to come on that in subsequent prs). It also cleans up some of the `commit!` process by removing the `apply-stage!` function which was made redundant by the new tracking apis. Now commits will contain any fuel, policy, or file metadata directly within the returned commit result if the tracking was turned on.


dpetran (approved):
:tractor: 

Do you want me to hold off on merging my fuel tracking PR and rebase on this? https://github.com/fluree/db/pull/1000

Issue #1004, opened 2025-04-22
Is 'a' as a short hand for 'rdf:type' available in SPARQL queries ?


acassaignemondeca:
If I execute the SPARQL CONSTRUCT query:
```
PREFIX rdf:<http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs:<http://www.w3.org/2000/01/rdf-schema#>
PREFIX skos:<http://www.w3.org/2004/02/skos/core#>
CONSTRUCT {
        ?s ?p ?o .
        ?s a skos:Concept .
} FROM <LEDGER_ID>
WHERE {
        VALUES ?p { rdfs:label skos:prefLabel skos:altLabel }
        ?s ?p ?o .
}
```

I get an empty result.

Replacing 'a' by 'rdf:type' yields the correct output.

dpetran:
I've checked how this query gets parsed, and 'a' and 'rdf:type' parse to the same internal query representation, so you should get the same result from either query. Can you share some data that can reproduce the result?

Issue #1003, opened 2025-04-22
JSON-LD: invalid value for @type


acassaignemondeca:
Exporting a ledger in JSON-LD with this SPARQL query:
```
PREFIX rdf:<http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs:<http://www.w3.org/2000/01/rdf-schema#>
PREFIX skos:<http://www.w3.org/2004/02/skos/core#>
CONSTRUCT {
        ?s ?p ?o .
} FROM <LEDGER_ID>
WHERE {
        VALUES ?p { rdf:type rdfs:label skos:prefLabel skos:altLabel }
        ?s ?p ?o .
}
```

gives an export JSON-LD. Here is one resource:
`{"@id":"http://www.eionet.europa.eu/gemet/concept/7793","@type":[{"@id":"skos:Concept"}],"skos:prefLabel":[{"@value":"dinámica social","@language":"es"},{"@value":"dynamique sociale","@language":"fr"},{"@value":"social dynamics","@language":"en"}]}`

When parsing with rdf4j, I get this error:
`JsonLdError[code=An invalid value for an @type entry has been detected [code=INVALID_TYPE_VALUE]., message=@type value is not valid [[{"@id":"skos:Concept"}]].]`

Maybe we should have "@type": ["skos:Concept"]

PR #1005, opened 2025-04-22
fix @type handling in construct selectors

dpetran:
@type values were being wrapped in id-maps like any other iri, instead of special cased. This adds the special casing for that predicate.

We translate "a" to "@type", which then gets interned on a flake as "rdf:type". So, when formatting as json-ld, we need to translate "rdf:type" to "@type".

Also addresses handling of `:id` and `:class` patterns in the construct clause.

Closes https://github.com/fluree/db/issues/1003
Closes https://github.com/fluree/db/issues/1004


dpetran:
Rebased on main.

Issue #1006, opened 2025-04-24
SPARQL CONSTRUCT Allow property paths with full IRIs


acassaignemondeca:
This SPARQL CONSTRUCT fails with an error on the property path:

```
CONSTRUCT {
<http://www.eionet.europa.eu/gemet/concept/4799> <http://www.w3.org/2004/02/skos/core#broader> ?directBroader.
?broader <http://www.w3.org/2000/01/rdf-schema#label> ?broaderLabel1.
?broader <http://www.w3.org/2000/01/rdf-schema#label> ?broaderLabel2.
?broader <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?broaderType.
?broader <http://www.w3.org/2004/02/skos/core#broader> ?broaderBroader.
}
FROM <gemet>
WHERE {
<http://www.eionet.europa.eu/gemet/concept/4799> <http://www.w3.org/2004/02/skos/core#broader> ?directBroader.
<http://www.eionet.europa.eu/gemet/concept/4799> <http://www.w3.org/2004/02/skos/core#broader>+ ?broader.
OPTIONAL { ?broader <http://www.w3.org/2004/02/skos/core#prefLabel> ?broaderLabel1. }
OPTIONAL { ?broader <http://www.w3.org/2000/01/rdf-schema#label> ?broaderLabel2. }
OPTIONAL { ?broader <http://www.w3.org/2004/02/skos/core#broader> ?broaderBroader. }
OPTIONAL { ?broader <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?broaderType. }
}
```

I think this property path is actually valid: `<http://www.w3.org/2004/02/skos/core#broader>+`
Short version `skos:broader+` works.

PR #1011, opened 2025-05-01
handle expanded IRIs in SPARQL property path expressions

dpetran:
Expanded IRIs need to wrapped in angle brackets to be valid SPARQL expressions. A bit awkward, since the rest of our translator is built to translate SPARQL into jld-query, and in this case we do not want to do that.

Closes https://github.com/fluree/db/issues/1006

### DIFF
--- a/test/fluree/db/query/property_path_test.clj
+++ b/test/fluree/db/query/property_path_test.clj
@@ -45,7 +45,7 @@
           (testing "transitive"
             (testing "without cycle"
               (is (= ["ex:b" "ex:c" "ex:d" "ex:e"]
-                     @(fluree/query db1 {"where" [{"@id" "ex:a" "<ex:knows+>" "?who"}]
+                     @(fluree/query db1 {"where" [{"@id" "ex:a" "<<ex:knows>+>" "?who"}]
                                          "select" "?who"}))))
             (testing "with cycle"
               (let [db2 @(fluree/stage db1 {"insert" {"@id" "ex:e" "ex:knows" {"@id" "ex:a"}}})]
--- a/src/fluree/db/query/fql/parse.cljc
+++ b/src/fluree/db/query/fql/parse.cljc
@@ -377,7 +377,7 @@
                      ;; translate back to string
                     (sparql.translator/parse-term)
                      ;; separate recursion modifier
-                    (split-at (dec (count path-expr)))
+                    (#(split-at (dec (count %)) %))
                      ;; turn back into strings
                     (map (partial apply str)))
         recur-mod   ({"+" :one+ "*" :zero+} mod)]
--- a/src/fluree/db/query/sparql/translator.cljc
+++ b/src/fluree/db/query/sparql/translator.cljc
@@ -610,7 +610,12 @@
 (defmethod parse-term :PathElt
   [[_ primary mod]]
   (if mod
-    (str "<" (parse-term primary) (parse-term mod) ">")
+    (let [term  (parse-term primary)
+          term* (if ((set (flatten primary)) :IRIREF)
+                  ;; expanded IRIs need to be wrapped in angle brackets in a transitive path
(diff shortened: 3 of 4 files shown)


test/fluree/db/query/property_path_test.clj
```diff
@@ -45,7 +45,7 @@
           (testing "transitive"
             (testing "without cycle"
               (is (= ["ex:b" "ex:c" "ex:d" "ex:e"]
-                     @(fluree/query db1 {"where" [{"@id" "ex:a" "<ex:knows+>" "?who"}]
+                     @(fluree/query db1 {"where" [{"@id" "ex:a" "<<ex:knows>+>" "?who"}]
```
zonotope, line 48: So the point of this pr is that we first need to wrap the property iri in angle brackets, and then wrap the whole property path in angle brackets as well?

dpetran: Yeah, the outer angle brackets signify that it's a property path, inside of the angle brackets needs to be a valid SPARQL expression, and non-compact IRIs in SPARQL need to be wrapped in angle brackets.

I concede that it is ugly as sin, but it is the syntax we decided on.

zonotope: Why does the inside of the angle brackets need to be a valid sparql expression? This is about parsing json-ld query.

dpetran: We had a very long [Slack thread](https://fluree-internal.slack.com/archives/C04AA19P3FW/p1719254665214139) exploring different syntax options, and then Andrew and I worked on a document honing that down with Brian, and in the end we decided that the jld-query syntax for property path expressions _is_ SPARQL property path expressions, but wrapped in angle brackets.

@aaj3f if you know where that document is that would be useful, I failed to link to it on the [PR](https://github.com/fluree/db/pull/952) where it was introduced.

aaj3f: Maybe my one question would be re: aliased property IRIs vs fully-qualified property IRIs. My understanding is that in sparql if you had `<http://example.org/knows>`, then the clause with (say, `+`) would either look like

```sparql
WHERE {
  ex:a <http://example.org/knows>+ ?who .
}
```
or, with prefixing
```sparql
WHERE {
  ex:a ex:knows+ ?who .
}
```

I wonder if we shouldn't accommodate the same thing, so instead of `{"@id" "ex:a" "<ex:knows+>" "?who"}` or `{"@id" "ex:a" "<<ex:knows+>>" "?who"}`, we'd have either

```json
{
  "where": {
    "@id": "ex:a",
    "ex:knows+": "?who"
  }
}
```
or
```json
{
  "where": {
    "@id": "ex:a",
    "<http://example.org/knows>+": "?who"
  }
}
```

Obviously, the use of just `"ex:knows": "?who"` or `"http://example.org/knows": "who"` would still be valid, but if you want to use the property path syntax and **_IF_** you're also using expanded/fully-qualified IRIs, then you **_HAVE_** to use `<...IRI...>` so we can disambiguate between the use of `+`/`|`/`*`/`^` etc as part of the IRI string vs as part of the property path operator syntax

dpetran: We use the outer `<` ... `>` angle bracket wrapping to signify a property path expression as a performance enhancement. If we don't have some cheap and easy way to check that a property may actually be a property path expression then we have to use a regex to check every property.

Which is maybe ok? This only happens during parsing a query, so parse time does not dominate total execution time.

I think in the end we had decided to basically punt on property path expression complexity and just use SPARQL inside of angle brackets. This is one corner case where it doesn't _look_ nice, but it _is_ consistent and already well documented. Once we step out of this lane, we are back to having to describe our own custom syntax here.

zonotope (approved):
🌙

PR #1012, opened 2025-05-05
Capture policy parsing/rule exceptions

bplatz:
Exceptions with policy rules were being swallowed, and would effect act like the policy doesn't exist - usually not what you'd want.

This introduces an `error-ch` strategy in policy parsing and execution which will now throw when calling `wrap-policy` or related APIs.

Also (fluree/query ...) will check that its DB object is not an exception before executing. When the db is an exception, it is difficult to know the issue as you'll get a cryptical 'cannot assoc ... to Exception' error.

### DIFF
--- a/src/fluree/db/json_ld/policy/rules.cljc
+++ b/src/fluree/db/json_ld/policy/rules.cljc
@@ -4,10 +4,9 @@
             [fluree.db.dbproto :as dbproto]
             [fluree.db.json-ld.iri :as iri]
             [fluree.db.json-ld.policy :as policy]
-            [fluree.db.util.async :refer [go-try <?]]
+            [fluree.db.util.async :refer [<?]]
             [fluree.db.util.core :as util :refer [try* catch*]]
-            [fluree.db.util.log :as log]
-            [fluree.json-ld :as json-ld]))
+            [fluree.db.util.log :as log]))
 
 #?(:clj (set! *warn-on-reflection* true))
 
--- a/src/fluree/db/json_ld/policy/rules.cljc
+++ b/src/fluree/db/json_ld/policy/rules.cljc
@@ -5,7 +5,8 @@
             [fluree.db.json-ld.iri :as iri]
             [fluree.db.json-ld.policy :as policy]
             [fluree.db.util.async :refer [go-try <?]]
-            [fluree.db.util.core :as util]
+            [fluree.db.util.core :as util :refer [try* catch*]]
+            [fluree.db.util.log :as log]
             [fluree.json-ld :as json-ld]))
 
 #?(:clj (set! *warn-on-reflection* true))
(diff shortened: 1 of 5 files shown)


src/fluree/db/json_ld/policy/rules.cljc
```diff
@@ -78,22 +78,25 @@
...
-                                    (async/pipe (async/chan 2 (comp cat sid-xf)))
-                                    (async/pipe ch)))
-                              ;; non-maps are literals
-                              (async/onto-chan! ch [target-expr])))
+                            (async/go
+                              (try*
+                                (if (query-target? target-expr)
+                                  (let [target-q (cond-> (assoc target-expr
+                                                                "select" "?$target"
+                                                                :selection-context {}) ;; don't compact selection results
```
dpetran, line 92: What is `:selection-context` for?

test/fluree/db/policy/target_test.clj
```diff
@@ -484,3 +485,31 @@
...
+                       "f:targetSubject" {"@type"  "@json"
+                                          "@value" {"blah" {[] []}}}
+                       "f:action"        [{"@id" "f:view"}, {"@id" "f:modify"}]
+                       "f:query"         {"@type"  "@json"
+                                          "@value" {"@context" {"ex" "http://example.org/ns/"}
+                                                    "where"    [{"@id" "?$this" "ex:classification" "?c"}
+                                                                ["filter", "(< ?c 1)"]]}}}]
+          bad-db @(fluree/wrap-policy db1 bad-policy)]
+
+      (is (util/exception? bad-db)))))
```
dpetran, line 515: What sort of error message will the user see? Since policies may be parsed long after they are transacted, the user who sees this error may not be able to do anything to fix it.

src/fluree/db/json_ld/policy/rules.cljc
bplatz: This was introduced with other rules parsing.

It provides a separate context for selection results, different than the context used in the where clause.

The goal is the same as done here - not have to compact the select result, then expand it again - which is expensive.

test/fluree/db/policy/target_test.clj
bplatz: These errors are reported immediately when calling wrap-policy.

Do you mean if a user created policy rules in a DB but never used, them - then decided to use them later? To capture those errors we'd have to scan for policy transactions as they happen and try to parse them then.

I am not sure that is a good idea, but certainly debatable. Outside of the scope here - whose purpose is to fix the issue that policy errors were not getting reported at all.

src/fluree/db/json_ld/policy/rules.cljc
dpetran: That's a useful feature, I like it. We may want to call it something more generic, like `result-context`, since we use that context for `:construct` results in addition to regular `:select`/`:select-distinct`/`:select-one` results.

bplatz: I'd support that name change. It isn't documented at this point, just used internally in similar situations. I think outside the scope here as that would impact other non-policy code.

dpetran (approved):
:black_bird:

PR #1013, opened 2025-05-06
Policy nil query handled as false/deny

bplatz:
This builds on #1012 

This allows for `f:query` in policy to now be optional, and when not included it acts as a false/deny response if a query had existed.

Previously, if you wanted an 'always deny' policy, you'd have to write a query that always returned false. The common example is this one:
```
...
"f:query":  {"@type" : "@json"
             "@value": {"where": [["filter", "(not= 1 1)"]]}}
...
```

This is difficult to explain and didn't seem very natural. In addition it required a pointless query execution processing step.

I believe this updated approach is easier to explain and is more efficient. The change is backwards compatible, so an 'always false' query will still work.

### DIFF
--- a/src/fluree/db/json_ld/policy/enforce.cljc
+++ b/src/fluree/db/json_ld/policy/enforce.cljc
@@ -60,9 +60,8 @@
                  policy)))))
 
 (defn policy-query
-  [db sid policy]
+  [db sid query]
   (let [policy-values (-> db :policy :policy-values)
-        query         (:query policy)
         this-val      (iri/decode-sid db sid)
         values        (-> (util.parse/normalize-values policy-values)
                           (policy/inject-value-binding "?$this" {"@value" this-val "@type" const/iri-id}))]
@@ -73,6 +72,8 @@
   (ex-info (or (some :ex-message policies)
                "Policy enforcement prevents modification.")
            {:status 403 :error :db/policy-exception}))
+
+(def ^:const deny-query-result false)
 
 (defn policies-allow?
   "Once narrowed to a specific set of policies, execute and return
@@ -86,8 +87,11 @@
           (let [{exec-counter :executed
                  allowed-counter :allowed} (get tracer (:id policy))
 
-                query   (policy-query db sid policy)
-                result  (seq (<? (dbproto/-query (root db) fuel-tracker query)))]
+                query   (when-let [query (:query policy)]
+                          (policy-query db sid query))
+                result  (if query 
+                          (seq (<? (dbproto/-query (root db) fuel-tracker query)))
+                          deny-query-result)]
             (swap! exec-counter inc)
             (if result
               (do (swap! allowed-counter inc)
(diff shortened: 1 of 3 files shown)


dpetran (approved):
This looks good, but it naturally led to me to wonder how to write a policy that always allows. 

I've been thinking about, and maybe we should continue this to make allow/deny support more explicit, so policies don't just look incomplete, rather they can be sort of self-documenting.

My thought was this: support `f:when`, which accepts a boolean or a query (basically, extending `f:query` with boolean support and more semantically useful name):
```
| arg   | result |
|-------+--------|
| true  | allow  |
| false | deny   |
| query | check  |
```

bplatz:
> how to write a policy that always allows

You probably already know this, but for others the way this is done is a query without a where clause which ends up looking like this:
```
{"select": "?$this",
 "values": ["?$this", [<iri-here>]]}
```

But, because we "magically" inject both the `select` and `values` into the query, is actually defined like:
```
...
"f:query": {"@type": "@json", 
            "@value": {}}
...
```

To your point, we can do better.

I think your `f:when` sounds like a reasonable option. If we went that route, I'd sort of expect the query to be its own graph node as opposed to a JSON literal (although either could work).

But before you suggestion, I was thinking we might want to stop the "magic" and have people write their own `select` and `values` statements - and then we can verify they are formatted correctly. The same would apply to `?$target`.

dpetran:
Yeah, I think removing the "magic" `select` clause in `targetSubject` and `targetProperty` makes sense. Then it would match the way we do `f:query`. But I think we _have_ to inject the the `?$target` values pattern with the `targetValues` opt, that's meant to allow using the same policy with different inputs. 

Likewise, if people are going to transact a policy for reuse the query has to be a JSON literal, we can't store a query properly since it's not valid RDF.

Should I make a couple of tickets for this? I don't think this is difficult work and I think it would improve usability.

bplatz:
> But I think we _have_ to inject the the `?$target` values pattern with the `targetValues` opt, that's meant to allow using the same policy with different inputs.

We have other variables that are in consideration to inject. I'd almost rather see for those situations, which I'd think are much less common, that you specify other values to inject - and perhaps even specify the variable name you'd like them injected at. That way you can optionally inject a `value` into targetSubject or targetProperty query, and avoid the need for `targetValues`. But I don't know all the use cases we are trying to support.

> Likewise, if people are going to transact a policy for reuse the query has to be a JSON literal, we can't store a query properly since it's not valid RDF.

I meant the query is still JSON, but it could be a property of a subject of e.g. class `f:Query` where other things could be configured for the query as well - e.g. an `rdfs:label` to describe the query, and if we allowed custom specifying variables or other things as I mentioned above it could be handled there as well. A literal is fine too, as that's what we already do - just thinking ahead.

> Should I make a couple of tickets for this? I don't think this is difficult work and I think it would improve usability.

Sure thing, thanks!

PR #1014, opened 2025-05-07
Feature: Static bindings in :bind clause

dpetran:
Adds SPARQL and JLD-Query support for binding static values in a `:bind` clause, in addition to function calls. This is a commonly used idiom in SPARQL queries, though you can also use a `:values` clause to achieve the same effect.


zonotope (approved):
🖍️

PR #1016, opened 2025-05-09
Feature/drop

dpetran:
Add initial support for the `drop` api, which deletes all of the persistent artifacts created for a ledger: 
- all commits
- all indexes
- all garbage
- all nameservice records

### DIFF
--- a/src/fluree/db/connection.cljc
+++ b/src/fluree/db/connection.cljc
@@ -447,7 +447,7 @@
     (load-ledger-alias conn alias-or-address)))
 
 (defn drop-commit-artifacts
-  [{:keys [commit-catalog] :as conn} latest-commit]
+  [{:keys [commit-catalog] :as _conn} latest-commit]
   (let [error-ch  (async/chan)
         commit-ch (commit-storage/trace-commits commit-catalog latest-commit 0 error-ch)]
     (go-loop []
@@ -457,7 +457,7 @@
               data-address        (-> (util/get-first commit const/iri-data)
                                       (util/get-first-value const/iri-address))]
           (log/debug "Dropping commit" (-> (util/get-first commit const/iri-data)
-                                          (util/get-first-value const/iri-fluree-t)))
+                                           (util/get-first-value const/iri-fluree-t)))
           (when data-address
             (log/debug "Deleting data" data-address)
             (storage/delete commit-catalog data-address))
@@ -474,7 +474,7 @@
   [storage node-address]
   (go-try
     (loop [[address & r] [node-address]
-           addresses     (list )]
+           addresses     (list)]
       (if address
         (if-let [children (->> (:children (<? (storage/read-json storage address true)))
                                (mapv :id))]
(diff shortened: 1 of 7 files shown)


src/fluree/db/connection.cljc
```diff
@@ -445,6 +446,88 @@
...
+                  (nameservice/address-path alias)
+                  alias)]
+      (loop [[publisher & r] (publishers conn)]
+        (when publisher
+          (let [ledger-addr   (<? (nameservice/publishing-address publisher alias))
+                latest-commit (-> (<? (nameservice/lookup publisher ledger-addr))
+                                  json-ld/expand)]
+            (log/warn "Dropping ledger" ledger-addr)
+            (drop-index-artifacts conn latest-commit)
+            (drop-commit-artifacts conn latest-commit)
```
bplatz, line 525: This and the prior line never take `<?` and therefore despite catching exceptions in those functions they will never pull through.

That might be OK here if you decide you want everything else deleted regardless of a thrown exception, but then you could assign the result channel to a var, then before you `recur` below, do a `<?` on the var.

Also need to decide if that is OK when multiple publishers exist... should an exception in one publisher stop deletions for the others? Probably not.

Maybe need to reevaluate exception handling / reporting here. As it stands, if an exception is raised, it will stop all work for remaining publishers.

src/fluree/db/connection.cljc
```diff
@@ -445,6 +446,88 @@
...
+          (log/debug "Dropping node" address)
+          (storage/delete storage address))))
+    :nodes-dropped))
+
+(defn drop-index-artifacts
+  [{:keys [index-catalog] :as _conn} latest-commit]
+  (go-try
+    (let [storage       (:storage index-catalog)
+          index-address (some-> (util/get-first latest-commit const/iri-index)
+                                (util/get-first-value const/iri-address))]
```
bplatz, line 494: I don't think this will capture a virtual graph index (BM25), but hard to tell from first glance.

src/fluree/db/connection.cljc
dpetran: I deliberately swallowed exceptions here since multiple nameservice records will point to different places on the commit chain, and so lower parts of the chain may already be deleted, which I think is ok. This whole operation is meant to be idempotent, so if you lose power halfway through you should be able to try again when you're back up and still delete all the artifacts.

bplatz: Then not a huge issue, but your use of `<?` implies an intention to stop processing. Maybe you just want to `<!`, or still <? but have an upper try/catch to log the exception with a `go` instead of a `go-try`.

src/fluree/db/connection.cljc
bplatz: I think we should move forward with this PR and deal with the Virtual Graph stuff later - it is a moving target regardless. @zonotope do you agree, if so please review as well - thanks!

PR #1017, opened 2025-05-12
Expand literal nodes

zonotope:
This patch adds initial support for expanding literal nodes in queries. `@value`, `@type`, and `@language` are treated as "virtual properties" and can be queried for in the same way as properties can be queried for on subject nodes. Two examples can be found in the additions to the json-ld-basic-test namespace

This patch only enables treating literal nodes in the same way as subject nodes. It does not include the extension of the graph subselection query language to support wildcards for all properties like `{:select {:ex/foo {:* [:value :language]}}}`. That will come in another pr.


bplatz:
I'm not sure I agree with the `:depth x` flag crawling the metadata. I think that would not be what folks would expect.

I do think we should introduce (at some point) an `:expanded-result true` or similar flag that will return all values as maps with `@type` or `@language` included for all results, but I think of that is separate from 'depth' which to me means crawling connections deeper.

zonotope:
> I'm not sure I agree with the `:depth x` flag crawling the metadata. I think that would not be what folks would expect.
> 
> I do think we should introduce (at some point) an `:expanded-result true` or similar flag that will return all values as maps with `@type` or `@language` included for all results, but I think of that is separate from 'depth' which to me means crawling connections deeper.

I can't say what users will expect, but my mental model when starting these features was to treat literal nodes just as we do subject nodes. I had always thought of depth as "expand the connected nodes". In that framing, the literal nodes are connected too, so they too are expanded. 

I do think the "traverse connected subjects" interpretation also makes sense; it's just not how I thought of it. That does break the mental model of "treat literal nodes like subject nodes" which underpins this feature though.

I don't think we should add the `:expand-result` option if we can avoid it. I think special options muddle the query language by adding more complexity. That complexity doesn't seem to matter when considering the individual feature, but the complexity from each special option adds up when trying to learn or explain the language. I'd rather we come up with an intuitive syntax that maps to _some_ consistent mental model (not necessarily the one I have) because I think that would make the query language as a whole simpler to understand, explain, and implement.

bplatz:
> I can't say what users will expect, but my mental model when starting these features was to treat literal nodes just as we do subject nodes. I had always thought of depth as "expand the connected nodes". In that framing, the literal nodes are connected too, so they too are expanded.

If you wrote out the query instead of using the 'depth' shortcut, you'd do something like:

{:select {?s ["ex:name" {"ex:friend": ["ex:name" {"ex:friend": ["ex:name" {"ex:friend": ["ex:name*"]]]}]}

So the 'depth' shortcut allows you to not have to do that. But now, as it appears anyhow, if I use the depth shortcut the query results would not be the same.

zonotope:
@bplatz I decided to remove depth-related literal node expansion because there were just too many inconsistencies, even when giving it a new keyword instead of being under the select clause. 

I think there might be inconsistencies with depth in general. Even if it is consistent, I think it's a blunt instrument that isn't well suited for nodes with a lot of properties whose objects are references. I think we should revisit depth with a consistent property path syntax for both edn and json queries.

In any case @fluree/core this is ready for review.

PR #1018, opened 2025-05-13
Fix bm25 multi-cardinality strings issue

bplatz:
This fixes an issue where a multi-cardinality string in the bm25 index can create an issue.

The problem was a syntax error, no changes in logic.

Closes https://github.com/fluree/core/issues/182


dpetran (approved):
Looks good, do you have a test case with multicardinal string values that demonstrate the issue?

PR #1020, opened 2025-05-13
Fix/bm25 empty string

bplatz:
This address a BM25 issue when there is only an empty string `""` to index.

Now it will skip this scenario.


dpetran (approved):
:space_invader:

PR #1021, opened 2025-05-13
Feature/sparql group concat

dpetran:
Adds [GROUP_CONCAT](https://www.w3.org/TR/sparql11-query/#defn_aggGroupConcat) support for SPARQL queries and fixes our non-functional `groupconcat` function in JLD-Query.

Closes https://github.com/fluree/core/issues/186


bplatz (approved):
This is great to have (for LLMs and otherwise)

PR #1022, opened 2025-05-14
speed up commits with transient

bplatz:
Small change to speed up commits a little (at least larger commits)


dpetran (approved):
:runner:

PR #1024, opened 2025-05-16
offset and limit after projection

dpetran:
We were applying `distinct` after offset and limit, which means users would have cases where they wouldn't receive desired results.

Closes https://github.com/fluree/core/issues/188

### DIFF
--- a/src/fluree/db/query/exec.cljc
+++ b/src/fluree/db/query/exec.cljc
@@ -63,14 +63,14 @@
         (group/combine q)
         (having/filter q error-ch)
         (select/modify q)
-        (order/arrange q)
-        (drop-offset q)
-        (take-limit q))))
+        (order/arrange q))))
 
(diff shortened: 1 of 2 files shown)


src/fluree/db/query/exec.cljc
```diff
@@ -80,7 +80,9 @@
   [subquery]
   (fn [ds fuel-tracker error-ch]
     (->> (execute* ds fuel-tracker subquery error-ch)
-         (select/subquery-format ds subquery fuel-tracker error-ch))))
+         (select/subquery-format ds subquery fuel-tracker error-ch)
+         (drop-offset subquery)
```
zonotope, line 84: Tiny suggestion that you can take or leave, but it might be better to group these `drop-offset` and `take-limit`  into a `page` function since they will always be called together and need to be called in two different places at least.

PR #1025, opened 2025-05-19
Predicates for tracking queries and transactions

zonotope:
This patch adds two predicate functions for determining if a query or transactions should be tracked instead of having one function for both cases. It also adds a new option for tracking the time a query or transaction took, and explicitly removes identity from the create options.


dpetran (approved):
:tractor:

PR #1026, opened 2025-05-22
fix logical and and or

dpetran:
The macro should just return the result of first/last form that evaluates to truthy/falsy, which will be a TypedValue, no need to create a new one.


zonotope (approved):
🪨

PR #1029, opened 2025-05-30
Consolidate tracking

zonotope:
This patch consolidates fuel and time tracking into a single tracker instance that maintains stats and generates the tally in one place. I opted to leave policy tracking outside of the tracker instance because it wasn't strictly necessary as I initially thought, but I would like to eventually move policy tracking to the tracker instance as well, barring any resource contention issues.

I also ensured that the policy tracking opts were always respected whether the meta option was `true`, or the policy meta option was also true.

Lastly, I did a little cleanup and simplification of the internal policy api to avoid passing boolean options meant solely to facilitate control flow.


src/fluree/db/track/fuel.cljc
```diff
@@ -4,38 +4,36 @@
...
 
-(defn tracker
-  "Creates a new fuel tracker w/ optional fuel limit (0 means unlimited)."
-  ([] (tracker nil))
-  ([limit]
-   {:limit    (or limit 0)
-    :counters (atom [])}))
+(defn init
+  [max-fuel]
+  (atom {:limit    (or max-fuel 0)
```
dpetran, line 9: Just curious about this - why move the whole fuel tracking map into the atom?

PR #1032, opened 2025-06-05
Add basic insert API

bplatz:
This adds a basic `insert` API that works much like the current `stage` however it accepts a pure JSON-LD document.

This means a JSON-LD document doesn't need to be modified to be inserted into a Fluree db - today you must wrap a JSON-LD document in an `{"insert" <jld>}` declaration for it to work.

I'd like to see these features added to `import`, either in `fluree/db`, or perhaps more appropriately `fluree/server` in the future:
a) support other RDF formats (e.g. TTL) - there is already some work for this done in `fluree/server`
b) support very large documents - by breaking them up for the user and processing+indexing incrementally. Ideally this would be combined with work in the nameservice such that once started, a new query/txt to the ledger being processed would return an exception until it is finished. It probably also needs restart/retry logic to address power failure scenarios and ideally track some processing status to report back to user how much longer it is likely to take.

### DIFF
--- a/src/fluree/db/api/transact.cljc
+++ b/src/fluree/db/api/transact.cljc
@@ -16,10 +16,8 @@
 (defn insert
   [db txn override-opts]
   (go-try
-    (let [context    (or (ctx-util/txn-context txn)
-                         (:context override-opts))
-          parsed-txn {:insert (parse/jld->parsed-triples txn nil context)}]
-      (<? (connection/stage-triples db parsed-txn)))))
+    (let [parsed-triples (parse/jld->parsed-triples txn nil (:context override-opts))]
+      (<? (connection/stage-triples db {:insert parsed-triples})))))
 
 (defn stage
   [db txn override-opts]
--- a/src/fluree/db/api/transact.cljc
+++ b/src/fluree/db/api/transact.cljc
@@ -13,6 +13,14 @@
     (sparql/->fql txn)
     txn))
 
+(defn insert
+  [db txn override-opts]
+  (go-try
+    (let [context    (or (ctx-util/txn-context txn)
+                         (:context override-opts))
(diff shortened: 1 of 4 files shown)


src/fluree/db/api/transact.cljc
```diff
@@ -13,6 +13,14 @@
     (sparql/->fql txn)
     txn))
 
+(defn insert
+  [db txn override-opts]
+  (go-try
+    (let [context    (or (ctx-util/txn-context txn)
+                         (:context override-opts))
+          parsed-txn {:insert (parse/jld->parsed-triples txn nil context)}]
```
zonotope: Why not use the new `jld->parsed-triples` function directly here?

zonotope (approved):

src/fluree/db/api/transact.cljc
bplatz: It is using that function, but I might misunderstand what you mean.

zonotope: sorry, i didn't see that. i just noticed that you constructed a new map with just an insert key instead of using the parsed triples. i thought you had done this for parsing, but now i see it's to be compatible with the `stage-triples` function. perhaps we can add a function in a later refactor to take the triples directly without needing the map, but it isn't a big deal for now.

zonotope (approved):

src/fluree/db/api/transact.cljc
bplatz: > perhaps we can add a function in a later refactor to take the triples directly without needing the map, but it isn't a big deal for now.

Currently `-stage-txn` protocol expects the map syntax, so this would likely involve adding another protocol method if we were to change the input format of the current staging process.

PR #1033, opened 2025-06-09
use a single atom to manage fuel

zonotope:
This patch simplifies and dramatically optimizes fuel tracking by using a single atom to track the total fuel used for a query or transaction instead of keeping track of an atom of volatiles for each parallel thread of execution. 

We initially chose to use the atom of volatiles to avoid the contention of using a single atom shared across multiple parallel execution threads.  However this later created a tradeoff with the additional requirement of adding fuel limits and short circuiting the computation when that is exceeded. This required that all the volatiles be constantly summed to keep track of the current global fuel total to decide if we need to short-circuit the computation. The constant repeated summing of the distinct volatiles has much worse computational complexity than the contention on the single atom among multiple threads.

I did some quick benchmarks using a relatively large database with and without tracking. The single atom approach does not add a significant amount of overhead to my test queries, where summing the volatiles increased the query time by an order of magnitude:

| Branch | Tracking Off | Tracking On |
| -----------|-------------------|-------------------|
| main |  2,374 msecs | 36,789 msecs |
| refactor/fuel-atom |  2,708 msecs |  2,634 msecs |

### DIFF
--- a/src/fluree/db/track/fuel.cljc
+++ b/src/fluree/db/track/fuel.cljc
@@ -6,39 +6,27 @@
 
 (defn init
   [max-fuel]
-  (atom {:limit    (or max-fuel 0)
-         :counters []}))
+  (atom {:limit (or max-fuel 0)
+         :total 0}))
 
 (defn tally
   [fuel]
-  (let [{:keys [counters]} @fuel]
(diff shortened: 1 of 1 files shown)


dpetran (approved):
:racehorse:

PR #1034, opened 2025-06-09
log with mdc context and propagate locals for mdc and otel in go blocks

kurtharriger:
Update core async to support mdc and otel context propagation

### DIFF
--- a/src/fluree/db/util/async.cljc
+++ b/src/fluree/db/util/async.cljc
@@ -1,9 +1,12 @@
 (ns fluree.db.util.async
   (:require
+   #?(:clj [clojure.core.async.impl.dispatch :as dispatch])
    [clojure.core.async :as async]
    [clojure.core.async.impl.protocols :as async-protocols]
    [fluree.db.util.core :as util #?(:clj :refer :cljs :refer-macros) [try* catch*]])
-  #?(:cljs (:require-macros [fluree.db.util.async :refer [<? go-try]])))
+  #?(:cljs (:require-macros [fluree.db.util.async :refer [<? go-try]]))
+  #?(:clj (:import [java.util.concurrent Executor]
+                   [org.slf4j MDC])))
 
 #?(:clj (set! *warn-on-reflection* true))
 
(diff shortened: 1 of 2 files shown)


src/fluree/db/util/async.cljc
```diff
@@ -105,3 +108,57 @@
...
+;;   is propagated to threads when desired.
+;;   See https://github.com/open-telemetry/opentelemetry-java/blob/main/context/src/main/java/io/opentelemetry/context/Context.java
+;;   core.async does not expose the thread pool direclty so we need to monkey patch it.
+
+;;   core.async will propagate clojure var bindings, so alternatively it might be possible
+;;   to implement a ContextStorageProvider that stores state in a var instead of thread locals.
+;;   I spent a bit of time trying to figure this out but it seems quite a bit more involved than
+;;   patching core.aysnc"
+#?(:clj
+   (do
```
zonotope, line 123: I think it's better to have multiple calls to the reader conditional for each top-level form instead of wrapping everything with a `do`. `def`, `defn`, and `defonce` should always be at the top level.

src/fluree/db/util/async.cljc
```diff
@@ -105,3 +108,57 @@
...
+         (catch ClassNotFoundException _
+           (fn wrap-otel ^Runnable [^Runnable r] r))))
+
+     (defn wrapped-runable ^Runnable [^Runnable r] (wrap-otel (wrap-mdc r)))
+     (defn patched-run
+       "Runs Runnable r with OpenTelemetry context propagation."
+       [^Runnable r]
+       (original-run (wrapped-runable r)))
+
+     (alter-var-root #'dispatch/run (constantly patched-run))
```
zonotope, line 155: I get really squeamish about patching external function implementations we don't control, especially non-public implementation internals. Not only can it lead to subtle bugs that are difficult to diagnose, but it also relies on a non-contracted implementation that could change at any time either breaking everything all together or, worse, leading to undefined behavior and more difficult to diagnose subtle bugs. 

I think this is especially true in this case because `clojure.core.async` is about to totally change its implementation to move from their own async thread pool to using JVM native virtual threads provided by [project loom](https://openjdk.org/projects/loom/). That new implementation is already in alpha, and we plan on switching to it as soon as it's marked stable.

zonotope (commented):
Did you evaluate the [`clj-otel` library](https://github.com/steffan-westcott/clj-otel)? They have an [example utility namespace for working with core.async](https://github.com/steffan-westcott/clj-otel/blob/master/examples/common/core-async.utils/src/example/common/core_async/utils.clj). Does that not work for us?

PR #1036, opened 2025-06-10
Turtle (TTL) insert + upsert supprt

bplatz:
This adds Turtle file format (.ttl) support for `insert` and `upsert` APIs that were recently introduced.

I originally built this support into a fluree/server PR for a new `import` API here: https://github.com/fluree/server/pull/88 - but that PR never got merged. I'll close that PR, and the fluree/server work to support this will need to be done over to now leverage this capability directly in fluree/db.

This PR (as did the original fluree/server PR) leverages the great library from @quoll to assist with the Turtle parsing.

### DIFF
--- a/src/fluree/db/query/turtle/parse.cljc
+++ b/src/fluree/db/query/turtle/parse.cljc
@@ -1,0 +1,53 @@
+(ns fluree.db.query.turtle.parse
+  (:require [fluree.db.constants :as const]
+            [fluree.db.json-ld.iri :as iri]
+            [fluree.db.query.exec.where :as exec.where]
+            [quoll.raphael.core :as raphael]))
+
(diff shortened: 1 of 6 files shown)


dpetran (approved):
:turtle:

src/fluree/db/query/turtle/parse.cljc
```diff
@@ -0,0 +1,53 @@
+(ns fluree.db.query.turtle.parse
```
zonotope, line 1: Definitely not for this pr, but I think we should eventually reorganize these namespaces. The `fluree.db.query.*` fragment is dealing with much more than queries.

zonotope (commented):
🌸

PR #1037, opened 2025-06-11
Remove 'conn' dependency on Ledger record

bplatz:
This removes any final dependencies on 'conn' from the Ledger record - ledgers no longer need to retain a reference to conn.


src/fluree/db/connection.cljc
```diff
@@ -587,9 +589,11 @@
 
 (defn publish-commit
   "Publishes commit to all nameservices registered with the ledger."
-  [{:keys [primary-publisher secondary-publishers] :as _conn} commit-jsonld]
+  [{:keys [publishers] :as _ledger} commit-jsonld]
   (go-try
-    (let [result (<? (nameservice/publish primary-publisher commit-jsonld))]
+    (let [primary-publisher (first publishers)
```
zonotope: I'd rather we were explicit about which publisher is the primary because the semantics of how we publish to them are different.

bplatz: They are different for commits, but not different for other operations (e.g. index updates) - but maybe they should be.

Those other operations expect 'publishers' - so we'll have multiple places now where we treat them separate but then recombine them - one in the 'conn', and another in indexing. I'll put up the change and you can decide if you prefer it or I can revert.

Because there were downstream operations expecting 'publishers' I settled in on just using it.

bplatz: OK, pushed. I think I prefer it the way you suggested so good to go unless you have any additional comments.

zonotope (approved):
👍🏾

PR #1039, opened 2025-06-12
Refactor/transact ns

bplatz:
Building on #1037 and #1038, this primarily focuses on moving non-connection related items out of the connection namespace.

The motivation for this change is to have a database-managed nameservice, but for that we need a ledger object that can exist outside of a conn and nameservice-related messaging separated from the conn. With the `conn` now removed from `ledger` in #1037 this becomes more possible, but not without starting to remove all the non-connection items out of the bloated connection namespace.

This code doesn't add any new functionality or logic changes, it is only movement.

Primary changes involve:
- commit-data related items in connection ns moved to commit, commit-data namespaces
- transaction related items in connection ns moved to transact / ledger namespaces
- publish/notification related items in connection ns moved to new `fluree.db.nameservices.subscribe` namespace

Besides that the other changes were side-effects, or to deal with cyclic dependency issues with some of the above.

There is definitely more opportunity for reorganization, but this seems like a decent place to stop without making too much change at once and mostly satisfies what is needed to move forward on more nameservice work.


src/fluree/db/api/transact.cljc
```diff
@@ -1,13 +1,16 @@
 (ns fluree.db.api.transact
   (:refer-clojure :exclude [update])
-  (:require [fluree.db.connection :as connection]
+  (:require [clojure.core.async :as async]
```
zonotope, line 3: Not for this pr, but I feel like we're close to getting rid of this whole namespace.

src/fluree/db/flake/transact.cljc
```diff
@@ -16,6 +16,10 @@
 
 #?(:clj (set! *warn-on-reflection* true))
 
+(defprotocol Transactable
```
zonotope, line 19: I think this protocol should go in the more general `fluree.db.transact` namespace because I think we could at least theoretically transact json-ld without turning it into flakes.

I think the `fluree.db.flake.*` hierarchy should be reserved for things that specifically involve a `FlakeDB`, and I think the `Transactable` protocol is more general than that.

Was there a cyclic dependency or something?

src/fluree/db/json_ld/commit_data.cljc
```diff
@@ -530,3 +490,157 @@
     (-> db*
         (update-novelty commit-flakes)
         add-tt-id)))
+
+(defn- handle-list-values
+  [objs]
+  {"@list" (->> objs (sort-by :i) (map #(dissoc % :i)))})
```
zonotope: Tiny thing, and it may have already been this way, but this will serialize the actual list as a sequence, which is inconsistent with how we serialize other json-ld data. I think we should use `mapv` here instead.

src/fluree/db/json_ld/commit_data.cljc
```diff
@@ -530,3 +490,157 @@
...
+                                     p-flakes)
+          objs*  (cond-> objs
+                   list? handle-list-values
+                   (= 1 (count objs)) first)
+          acc'   (assoc acc p-iri objs*)]
+      (if (seq r)
+        (recur r acc')
+        acc'))))
+
+(defn- commit-flakes
```
zonotope, line 556: I'm not sure that this belongs here. This function and the other functions below could only ever operate on a `FlakeDB`, so I think it would be a better fit in the `fluree.db.flake.*` hierarchy.

src/fluree/db/flake/transact.cljc
bplatz: Yes, this was a cyclic dependency issue. I actually thought it would work well in fluree.db.ledger, but that had issues as well. Finally stuck it here.

src/fluree/db/json_ld/commit_data.cljc
bplatz: The reason I consolidated this is because we had commit code in multiple places for no logical reason I could discern - so moving just this doesn't help that much. e.g. now `transact` will have to require another commit ns to make calls interspersed with this one for the same processing.

If we move it, which I'm good with, I'd prefer to move the whole namespace. If we ever have more than a FlakeDB then we can move the parts that don't deal with that out of that namespace, but as of right now everything revolves around FlakeDB.

zonotope: I think this points to the need for a db protocol for generating the json-ld commit document. Since we currently only ever generate commits with the `FlakeDB`, the priority for adding that is and has been lower. 

For now, I'd support us moving it to `fluree.db.flake.commit-data` or something similar.

src/fluree/db/flake/transact.cljc
zonotope: Ok, I pulled the branch down and see that the culprit is the `commit!` function in this transact namespace. i think it's doing too much and is too stateful, so it should be refactored. i think refactoring that function along with adding a `generate-commit` method to the `Transactable` protocol will fix things.

Those are going to be more involved refactors and I don't want to hold this pr up any longer since it has a lot of good things, so I'm good with moving forward with it as is for now.

zonotope (approved):
💮

PR #1044, opened 2025-07-01
Fix/json data type

bplatz:
When retrieving `@json` data types, internally we converted this to rdf:JSON iri. When querying and extracting `@type` we will return `rdf:JSON` which is not ideal... both because the user input `@json`, and second because `rdf:JSON` is not part of the ref-syntax spec at all.

This will now return `@json` in queries that try to bind datatype to a variable.


dpetran (approved):
Good catch!

PR #1046, opened 2025-07-02
Ensure `@type` query var binding works

bplatz:
A direct variable binding for `@type` in a where clause was not working correctly nor was `@language` bindings.

It was working when querying in EDN using a symbol as the variable, but there was an issue an no tests when string variables / JSON was used. This fixes that issue.

@aaj3f mentioned there was a ticket for this issue however I was unable to find it.


zonotope (approved):
⛵

PR #1048, opened 2025-07-02
fix shacl property shape path segment handling

dpetran:
When we reconstruct a shape in order to evaluate it we do so greedily, pulling in all the predicates on each subject. When we are resolving a property shape path, we were assuming that a regular property would not have any additional data associated with it, and that if the path segment's map didn't contain one of our supported path modifiers then it was invalid.

This was an erroneous assumption. Now I've added an explicit check for unsupported property paths so we can correctly handle path segments with arbitrary data.


zonotope (approved):
🏹

PR #1050, opened 2025-07-04
Fix async error handling that could cause indefinite hangs

bplatz:
## Summary
- Fixes a critical issue where fluree/db could hang indefinitely when exceptions occur in async commit processing
- Adds proper exception handling in `with-commit-data` function to check if values from the commits channel are exceptions before attempting to destructure them

## Problem
The `with-commit-data` function expected all values from the commits channel to be `[commit-jsonld commit-proof]` tuples. However, when `trace-commits` encounters an error, it puts the exception directly on the channel. This caused a "nth not supported on this type: ExceptionInfo" error when trying to destructure the exception as a vector, which could result in the system hanging indefinitely.

## Solution
Added exception checking using `util/exception?` before destructuring. If an exception is detected:
1. Forward it to the error channel
2. Close the processing channel properly

This ensures errors are propagated correctly and prevents hangs.

## Test plan
- [x] Verify that commit processing works normally with valid data
- [x] Verify that exceptions in the commit pipeline are properly handled without hanging
- [x] Run existing test suite to ensure no regressions

### DIFF
--- a/src/fluree/db/flake/flake_db.cljc
+++ b/src/fluree/db/flake/flake_db.cljc
@@ -464,14 +464,21 @@
 (defn with-commit-data
   [commit-storage error-ch commits-ch]
   (let [to (async/chan)
-        af (fn [[commit-jsonld _commit-proof] ch]
+        af (fn [input ch]
              (go
-               (let [db-address (-> commit-jsonld
-                                    (get-first const/iri-data)
-                                    (get-first-value const/iri-address))]
-                 (-> commit-storage
-                     (read-commit-data commit-jsonld db-address error-ch)
-                     (async/pipe ch)))))]
+               ;; Check if input is an exception
+               (if (util/exception? input)
+                 (do
+                   (>! error-ch input)
+                   (async/close! ch))
+                 ;; Normal processing path
+                 (let [[commit-jsonld _commit-proof] input
+                       db-address (-> commit-jsonld
+                                      (get-first const/iri-data)
+                                      (get-first-value const/iri-address))]
+                   (-> commit-storage
+                       (read-commit-data commit-jsonld db-address error-ch)
+                       (async/pipe ch))))))]
     (async/pipeline-async 2 to af commits-ch)
     to))
 


zonotope (commented):
I don't think this is the correct solution. I would rather we fix `trace-commits` so that it puts any errors it encounters on the error channel that's passed to it instead of checking if each item we take off of the response channel is an error or not. This solution leads to ad-hoc error handling that will introduce much more complexity in the long term.

src/fluree/db/flake/flake_db.cljc
```diff
@@ -464,14 +464,21 @@
...
+        af (fn [input ch]
              (go
-               (let [db-address (-> commit-jsonld
-                                    (get-first const/iri-data)
-                                    (get-first-value const/iri-address))]
-                 (-> commit-storage
-                     (read-commit-data commit-jsonld db-address error-ch)
-                     (async/pipe ch)))))]
+               ;; Check if input is an exception
+               (if (util/exception? input)
```
zonotope: This should not be necessary and will cover up bugs in other parts of the implementation if it's left here.

src/fluree/db/commit/storage.cljc
```diff
@@ -121,10 +121,13 @@
                                       (<? (read-verified-commit storage prev-commit-addr))
                                       (catch* e
                                         (log/error e "Error tracing commits")
-                                        (>! error-ch e)))]
-                (recur verified-commit commit-t commit-tuples*)))))
+                                        (>! error-ch e)
```
zonotope: We have redundant `try`/`catch` blocks here. There's one starting at line 120, and another one wrapping the whole function starting at line 106.

I think we should remove the `try`/`catch` at 120 and rely on the one that's wrapping the whole function instead. 

We should replace the `try`/`catch` starting at line 120 with just:
```clojure
(<? (read-verified-commit storage prev-commit-addr))
```

And we should replace the `catch` starting at 129 with
```clojure
(catch* e
  (log/error e "Error tracing commits")
  (>! error-ch e))
```

src/fluree/db/commit/storage.cljc
```diff
@@ -121,10 +121,13 @@
...
+                                        (async/close! resp-ch)
+                                        nil))]
+                (when verified-commit
```
zonotope: We should prefer `(when-let ...` over `(let ... (when ...`

PR #1052, opened 2025-07-07
Fix S3 storage implementation and add documentation

bplatz:
## Summary

This PR fixes several issues with the S3 storage implementation and adds documentation.

### Bug Fixes
- **Fixed index loading bug**: Resolved "Error resolving index node" when loading ledgers from S3 cold start by properly extracting local paths from fluree addresses in S3Store's `-read-json` method
- **API parameter validation**: Made `s3-endpoint` required in `connect-s3` with proper validation and error messages

### Test Organization
- **Reorganized S3 tests** into three categories:
  - `s3_unit_test.clj`: Pure unit tests (6 tests, 16 assertions) - no external dependencies
  - `s3_test.clj`: Integration tests (2 tests, 6 assertions) - requires LocalStack  
  - `s3_indexing_test.clj`: Indexing tests (1 test, 12 assertions) - workflow validation
- **Moved shared utilities**: `s3-available?` function moved to `fluree.db.test-utils` namespace
- **Updated logging**: Replaced `println` with `fluree.db.util.log` calls in tests
- **Removed redundant tests**: Consolidated duplicate test scenarios

### Documentation
- **Added S3 Storage Configuration Guide** (`docs/S3_STORAGE_GUIDE.md`)
- Covers configuration, authentication, API usage, and troubleshooting
- Includes examples for AWS and LocalStack usage

## Test Results

All S3 tests pass:
- Unit tests: 6 tests, 16 assertions ✅
- Integration tests: 2 tests, 6 assertions ✅ (requires LocalStack)
- Indexing tests: 1 test, 12 assertions ✅ (requires LocalStack)

**Note**: Full S3 integration testing requires LocalStack running at `localhost:4566`. Unit tests run without external dependencies.

## API Changes

The `s3-endpoint` parameter is now required in `connect-s3`:

```clojure
(fluree/connect-s3 {:s3-bucket "my-bucket"
                    :s3-endpoint "https://s3.us-east-1.amazonaws.com"})
```

### DIFF
--- a/src/fluree/db/api.cljc
+++ b/src/fluree/db/api.cljc
@@ -74,8 +74,9 @@
 
 (defn disconnect
   [conn]
-  (go-try
-    (-> conn ::system-map system/terminate)))
+  (promise-wrap
+   (go-try
+     (-> conn ::system-map system/terminate))))
 
 (defn convert-config-key
   [[k v]]
@@ -137,6 +138,47 @@
                                                                "storage" {"@id" "fileStorage"}}}]}
                        defaults (assoc-in ["@graph" 1 "defaults"] (convert-keys defaults)))]
      (connect file-config))))
+
+#?(:clj
+   (defn connect-s3
+     "Forms a connection backed by S3 storage.
+     
+     Options:
+       - s3-bucket (required): The S3 bucket name
+       - s3-endpoint (required): S3 endpoint URL
+         * For AWS S3: 'https://s3.<region>.amazonaws.com' (e.g., 'https://s3.us-east-1.amazonaws.com')
+         * For LocalStack: 'http://localhost:4566'
+         * For MinIO: 'http://localhost:9000'
+       - s3-prefix (optional): The prefix within the bucket
+       - parallelism (optional): Number of parallel operations (default: 4)
+       - cache-max-mb (optional): Maximum memory for caching in MB (default: 1000)
+       - defaults (optional): Default options for ledgers created with this connection"
+     ([{:keys [s3-bucket s3-prefix s3-endpoint parallelism cache-max-mb defaults],
+        :or   {parallelism 4, cache-max-mb 1000}}]
+      (when-not s3-bucket
+        (throw (ex-info "S3 bucket name is required for S3 connection"
+                        {:status 400 :error :db/invalid-config})))
+      (when-not s3-endpoint
+        (throw (ex-info "S3 endpoint is required for S3 connection. Examples: 'https://s3.us-east-1.amazonaws.com' for AWS, 'http://localhost:4566' for LocalStack"
+                        {:status 400 :error :db/invalid-config})))
+      (let [s3-config (cond-> {"@context" {"@base"  "https://ns.flur.ee/config/connection/"
+                                           "@vocab" "https://ns.flur.ee/system#"}
+                               "@id"      "s3"
+                               "@graph"   [(cond-> {"@id"      "s3Storage"
+                                                    "@type"    "Storage"
+                                                    "s3Bucket" s3-bucket}
+                                             s3-prefix   (assoc "s3Prefix" s3-prefix)
+                                             s3-endpoint (assoc "s3Endpoint" s3-endpoint))
+                                           {"@id"              "connection"
+                                            "@type"            "Connection"
+                                            "parallelism"      parallelism
+                                            "cacheMaxMb"       cache-max-mb
+                                            "commitStorage"    {"@id" "s3Storage"}
+                                            "indexStorage"     {"@id" "s3Storage"}
+                                            "primaryPublisher" {"@type"   "Publisher"
+                                                                "storage" {"@id" "s3Storage"}}}]}
+                        defaults (assoc-in ["@graph" 1 "defaults"] (convert-keys defaults)))]
+        (connect s3-config)))))
 
 (defn address?
   "Returns true if the argument is a full ledger address, false if it is just an
--- a/src/fluree/db/storage/s3.clj
+++ b/src/fluree/db/storage/s3.clj
@@ -17,8 +17,17 @@
 (defn handle-s3-response
   [resp]
   (if (:cognitect.anomalies/category resp)
-    (if (:cognitect.aws.client/throwable resp)
+    (cond
+      ;; Handle NoSuchKey as nil (file doesn't exist)
+      (= "NoSuchKey" (get-in resp [:Error :Code]))
+      nil
+
+      ;; Return throwables as-is
+      (:cognitect.aws.client/throwable resp)
       resp
+
+      ;; Other errors become excepftions
+      :else
       (ex-info "S3 read failed"
                {:status 500, :error :db/unexpected-error, :aws/response resp}))
     (let [{in :Body} resp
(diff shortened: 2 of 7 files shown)


src/fluree/db/api.cljc
```diff
@@ -138,6 +139,47 @@
...
+        (throw (ex-info "S3 endpoint is required for S3 connection. Examples: 'https://s3.us-east-1.amazonaws.com' for AWS, 'http://localhost:4566' for LocalStack"
+                        {:status 400 :error :db/invalid-config})))
+      (let [s3-config (cond-> {"@context" {"@base"  "https://ns.flur.ee/config/connection/"
+                                           "@vocab" "https://ns.flur.ee/system#"}
+                               "@id"      "s3"
+                               "@graph"   [(cond-> {"@id"      "s3Storage"
+                                                    "@type"    "Storage"
+                                                    "s3Bucket" s3-bucket}
+                                             s3-prefix   (assoc "s3Prefix" s3-prefix)
+                                             s3-endpoint (assoc "s3Endpoint" s3-endpoint))
```
zonotope: If s3Endpoint is required, this should never be `nil` and therefore should go in the map in the first argument of the `cond->` form.

src/fluree/db/api.cljc
```diff
@@ -138,6 +139,47 @@
...
+                                           {"@id"              "connection"
+                                            "@type"            "Connection"
+                                            "parallelism"      parallelism
+                                            "cacheMaxMb"       cache-max-mb
+                                            "commitStorage"    {"@id" "s3Storage"}
+                                            "indexStorage"     {"@id" "s3Storage"}
+                                            "primaryPublisher" {"@type"   "Publisher"
+                                                                "storage" {"@id" "s3Storage"}}}]}
+                        defaults (assoc-in ["@graph" 1 "defaults"] (convert-keys defaults)))]
```
zonotope: I think this binding would be much more readable by a human if we removed the top level `cond->` form and replaced it with a `cond->` form explicitly targeting the second entry from the `"@graph"` value vector:

```clojure
...
"@graph" [...
          (cond-> {"@id" "connection"
                   ...}
            defaults (assoc "defaults" (convert-keys defaults)))]
```

src/fluree/db/storage/s3.clj
```diff
@@ -17,8 +17,17 @@
...
+    (cond
+      ;; Handle NoSuchKey as nil (file doesn't exist)
+      (= "NoSuchKey" (get-in resp [:Error :Code]))
+      nil
+
+      ;; Return throwables as-is
+      (:cognitect.aws.client/throwable resp)
       resp
+
+      ;; Other errors become excepftions
```
zonotope: Spelling of "exceptions" is wrong.

src/fluree/db/storage/s3.clj
```diff
@@ -33,13 +42,16 @@
 
 (defn read-s3-data
   [s3-client s3-bucket s3-prefix path]
-  (let [ch        (async/promise-chan (map handle-s3-response))
+  (let [ch        (async/promise-chan)
```
zonotope: Was there an issue with the transducer supplied to the `promise-chan` function? The code as it was seemed like it should work unless I'm missing something.

src/fluree/db/storage/s3.clj
```diff
@@ -33,13 +42,16 @@
...
         full-path (str s3-prefix "/" path)
         req       {:op      :GetObject
                    :ch      ch
                    :request {:Bucket s3-bucket, :Key full-path}}]
     (aws/invoke-async s3-client req)
-    ch))
+    (go
```
zonotope: I would prefer to keep the previous code with the transducer directly attached to the promise channel unless there was a problem with it that I can't recognize.

The updated code now creates a new channel with the extra (albeit minimal) latency this entails to essentially reimplement the functionality of the transducer we had before.

test/fluree/db/storage/s3_indexing_test.clj
```diff
@@ -0,0 +1,134 @@
...
+                  data1 [{"@id" "ex:alice" "ex:name" "Alice" "ex:age" 30}
+                         {"@id" "ex:bob" "ex:name" "Bob" "ex:age" 25}]
+                  db1 @(fluree/stage db0 {"@context" context "insert" data1})]
+
+              (is (some? ledger) "Ledger should be created successfully")
+              (is (some? db1) "Data should stage successfully")
+
+              @(fluree/commit! ledger db1)
+
+              ;; Wait for async indexing
```
zonotope, line 65: I know all too well the difficulties of dealing with localstack so I'm not going to ask that this sleep be removed, but could you add a FIXME comment to find a more deterministic solution to waiting for the indexes to kick off and finish?

test/fluree/db/storage/s3_indexing_test.clj
```diff
@@ -0,0 +1,134 @@
...
+          (testing "Index generation after threshold exceeded"
+            (let [ledger @(fluree/load conn ledger-name)
+                  db-current (fluree/db ledger)
+                  data2 [{"@id" "ex:carol" "ex:name" "Carol" "ex:age" 35}
+                         {"@id" "ex:dave" "ex:name" "Dave" "ex:age" 40}]
+                  db2 @(fluree/stage db-current {"@context" context "insert" data2})]
+
+              @(fluree/commit! ledger db2)
+
+              ;; Wait for indexing to complete
```
zonotope, line 82: Could you also add a FIXME comment here to replace the sleep with a more deterministic solution?

test/fluree/db/storage/s3_unit_test.clj
```diff
@@ -0,0 +1,50 @@
...
+
+(deftest s3-storage-creation-test
+  (testing "S3 storage can be created with valid parameters"
+    (let [store (s3-storage/open "test-s3" "test-bucket" "test-prefix")]
+      (is (some? store) "S3Store should be created")
+      (is (= "test-s3" (:identifier store)) "Identifier should match")
+      (is (= "test-bucket" (:bucket store)) "Bucket should match")
+      (is (= "test-prefix" (:prefix store)) "Prefix should match"))))
+
+(deftest s3-storage-protocols-test
```
zonotope: I don't think this tests is doing any good. The s3 record already implements these protocols and if that changes in the future, it will most likely be intentional.

src/fluree/db/storage/s3.clj
bplatz: There was, with the transducer there was a hang. I was going to spend some more time on that, but the presence of it was, in my opinion, not ideal and a bit misleading. I mean, a mapping transducer implies there will be a sequence of items on the channel - here it is only ever a single item that is ever read, not a sequence. I think its presence makes the code more confusing, and it holds no performance value.

src/fluree/db/storage/s3.clj
bplatz: I left my prior comment before reading this one... and I think this is not a high frequency operation that latency will be of concern but if you'd prefer it the other way I can definitely get the transducer working, just need to debug a bit more.

src/fluree/db/storage/s3.clj
bplatz: I ended up reverting to this syntax per your suggestion, whatever bug I had where I initially change it for I must have fixed so it worked fine

zonotope (approved):
📏

PR #1053, opened 2025-07-08
Replace test-with-files with babashka.fs

bplatz:
## Summary
- Replaced test-with-files library with babashka.fs for better maintenance and active support
- Updated all test files to use babashka.fs's `with-temp-dir` macro
- Fixed Path to string conversion issues where needed

## Details
The test-with-files library we were using is a fork that hasn't been actively maintained. This PR switches to babashka.fs which:
- Is actively maintained and widely used in the Clojure community
- Provides the same functionality we need (temporary directory creation)
- Has better documentation and community support

### Changes made:
1. Replaced dependency in deps.edn
2. Updated 7 test files to use `babashka.fs`
3. Changed `with-tmp-dir` to `with-temp-dir` with proper syntax
4. Added string conversion for Path objects where needed
5. Used `:refer` instead of namespace aliases for cleaner code

## Test plan
- [x] Run all tests with `make test`
- [x] Verify all 201 tests pass
- [x] Check that temporary directories are properly cleaned up

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -1,8 +1,8 @@
-{:source-paths   #{"src"}
+{:source-paths   #{"src" "test"}
  :linters        {:unresolved-symbol {:exclude [(fluree.db.util.core/catch*)
-                                                (test-with-files.tools/with-tmp-dir)
                                                 goog.isObject]}}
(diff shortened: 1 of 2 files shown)


zonotope (approved):
🧀

PR #1056, opened 2025-07-09
Sorted where execution

zonotope:
This patch is the beginning of a query optimization framework by introducing a heuristic to just sort the triples resulting from parsing a single node map by prioritizing triples  with specified subject, predicate, and object values, in that order. This still relies on the user to order their node maps in the most optimal way, but it avoids edge cases where the non-deterministic order of parse results from maps could lead to expensive queries involving full index scans followed by filtering instead of the other way around.

### DIFF
--- a/src/fluree/db/query/exec/where.cljc
+++ b/src/fluree/db/query/exec/where.cljc
@@ -451,14 +451,16 @@
            (nil? p)
            s
            o)
-    (let [f (if o-fn
-              (fn [mch]
-                (and (#{o} (or (get-value mch)
-                               (get-sid mch db)))
-                     (o-fn mch)))
-              (fn [mch]
-                (#{o} (get-value mch))))]
-      [nil f])
+    (let [match-fn (fn [mch]
+                     (when-let [v (or (get-value mch)
+                                      (get-sid mch db))]
+                       (= o v)))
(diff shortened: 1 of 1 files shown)


src/fluree/db/query/optimize.cljc
```diff
@@ -0,0 +1,38 @@
...
+  (if (where/matched-value? cmp-a)
+    (if (where/matched-value? cmp-b)
+      0
+      -1)
+    (if (where/matched-value? cmp-b)
+      1
+      0)))
+
+(def triple-pattern-types
+  #{:tuple :class})
```
dpetran, line 15: Could we do `:id` patterns as well?

zonotope: I went back and forth on it, but I think :id patterns are a special case that don't need to be handled immediately. They'll only show up in a *top level* node map with out any other properties (like `{:id ?s}`). That node only creates a single triple, so nothing will be sorted anyway.

I think when we extend the optimization system, we'll have to handle :id patterns to decide where they should fall in the sorting order, but I feel like we want them to show up last in all cases anyway, even after things like :optional and :union, so I don't think they should be handled with this comparator since we most likely want triple type patterns to appear early on, just after :values and :bind.

PR #1057, opened 2025-07-09
Update documentation and API docstrings

bplatz:
## Summary

This PR updates the project documentation and API docstrings to improve clarity and accuracy:

### API Documentation Updates
- Updated function docstrings for better clarity and accuracy
- Removed marketing language in favor of technical descriptions
- Added detailed parameter descriptions and return value documentation
- Fixed syntax errors in docstrings with proper quote escaping

### API Improvements
- Fixed bug in `reasoned-facts` function to properly extract rule metadata
- Added deprecation metadata for `transact\!` and `stage` functions
- Created `credential-update\!` as replacement for `credential-transact\!`

### Documentation Updates
- Added quick start example to README
- Added API reference section with function categories
- Updated build instructions to reflect current Makefile targets
- Added Makefile help system with descriptions for all targets
- Updated prerequisites with Java version requirements and `.java-version` note

### Test Plan
- [x] All reasoner tests pass after `reasoned-facts` bug fix
- [x] All API tests pass with updated docstrings
- [x] Makefile help system works correctly
- [x] README examples are accurate

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,13 @@
-.PHONY: all deps jar install deploy nodejs browser webworker cljtest	\
+.PHONY: help all deps jar install deploy nodejs browser webworker cljtest	\
 	cljs-browser-test cljs-node-test cljstest test eastwood ci clean	\
 	js-packages sync-package-json publish-nodejs publish-browser		\
 	publish-webworker publish-js pending-tests pt clj-kondo-lint            \
-	clj-kondo-lint-ci cljfmt-check
+	clj-kondo-lint-ci cljfmt-check cljfmt-fix
 
+.DEFAULT_GOAL := help
+
+help: ## Describe available tasks
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
 
 DOCS_MARKDOWN := $(shell find docs -name '*.md')
 DOCS_TARGETS := $(DOCS_MARKDOWN:docs/%.md=docs/%.html)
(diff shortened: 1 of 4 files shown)


docs/semantic_web_guide.md
```diff
@@ -0,0 +1,715 @@
...
+                         
+                         ex:bob a foaf:Person ;
+                                foaf:name \"Bob\" ;
+                                foaf:age 25 ;
+                                schema:email \"bob@example.org\" .
+                         
+                         ex:CompanyA a schema:Organization ;
+                                     schema:name \"Company A\" ;
+                                     schema:employee ex:alice, ex:bob ."
+                        {"format" "turtle"}))
```
dpetran: These `opts` maps can use keywords for the keys and vals, I think that would be a bit better in a clojure api demo.

docs/semantic_web_guide.md
```diff
@@ -0,0 +1,715 @@
...
+                           FILTER (?age >= 65)
+                         }"
+                        {"format" :sparql}))
+```
+
+### Notes on SPARQL UPDATE Support
+
+- All standard SPARQL UPDATE operations are supported: INSERT DATA, DELETE DATA,
+    DELETE WHERE, DELETE/INSERT WHERE
+- The WITH clause is supported for specifying the target graph
```
dpetran, line 254: I might add a point here:

> An UPDATE operation can only create triples for a single graph, using WITH or a single USING clause.

Just to give a little bit of reasoning about why this restriction exists.

docs/semantic_web_guide.md
```diff
@@ -0,0 +1,715 @@
...
+```turtle
+@prefix : <http://example.org/vocab/> .
+@prefix data: <http://example.org/data/> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+```
+
+### 2. SHACL Design Patterns
+- Use SHACL for data validation
+- Define reusable shapes with `sh:node`
```
dpetran: I think it would be clearer to say "Define reusable shapes with type `sh:NodeShape`"

That's how the shapes are defined, they can then be _referenced_ using `sh:node` or any of the logical or shape constraints.

docs/semantic_web_guide.md
```diff
@@ -0,0 +1,715 @@
...
+- Leverage SPARQL-based constraints for complex validation rules
+
+### 3. Reasoning Performance
+- Apply reasoning selectively using graph patterns
+- Pre-compute common inferences
+- Use property paths in SPARQL instead of reasoning when possible
```
dpetran: Our SPARQL (and JLD-query) path support is limited: we support sequences of predicate path segments, inverse path segments, and we only support a single segment of one-or-more, zero-or-more, or zero-or-one paths. We do not negated paths at all.

docs/semantic_web_guide.md
bplatz: Addressed in 970fd629a

docs/semantic_web_guide.md
bplatz: Addressed in 970fd629a

docs/semantic_web_guide.md
bplatz: Addressed in 6de06d2e3

docs/semantic_web_guide.md
bplatz: Addressed in 6de06d2e3

I think this is the last feedback item you had @dpetran

zonotope (approved):
🗃️

PR #1058, opened 2025-07-09
Fix JavaScript SDK browser commit function and add documentation

bplatz:
## Summary
- Fix browser SDK commit function to properly handle the 3-arity case when options are provided
- Add JavaScript developer guide for both Node.js and browser usage
- Create working examples with documentation

## Problem
The browser SDK's commit function was not correctly handling the case where options are passed as a third argument. When options are provided, the commit function returns a map with metadata and the database under the `:db` key, rather than just returning the database directly. This caused "should be a map" errors when the JavaScript object was passed to ClojureScript APIs expecting a proper database object.

## Solution
Modified the browser SDK commit function to:
1. Check if the result is a map (indicating options were provided)
2. Extract the database from the `:db` key
3. Convert only the metadata to JavaScript while keeping the database as a ClojureScript object
4. Return a JavaScript object with the metadata and the ClojureScript database

## Additional Changes
- Added JavaScript developer guide in `/docs/javascript-developers-guide.md`
- Created working examples:
  - `/examples/nodejs-demo/` - Node.js example with all major operations
  - `/examples/browser-demo/` - Interactive browser example
- Updated main README to include link to JavaScript guide

## Test Plan
- [x] Browser SDK builds successfully
- [x] Node.js SDK builds successfully  
- [x] All existing tests pass
- [x] Node.js example runs without errors
- [x] Browser example works with both memory and localStorage connections
- [x] Queries execute successfully after commit with options


zonotope (approved):
🛳️

PR #1060, opened 2025-07-11
util Cleanup

zonotope:
This patch cleans up the `fluree.db.util.core` namespace renaming it to `fluree.db.util` and removing some unused functions/macros.


src/fluree/db/time_travel.cljc
```diff
@@ -1,7 +1,7 @@
 (ns fluree.db.time-travel
   (:require [clojure.core.async :as async]
+            [fluree.db.util :as util #?(:clj :refer :cljs :refer-macros) [try* catch*]]
```
dpetran: While we're doing cleanup, I think we can remove these reader conditionals, I think `:refer` works for both clj and cljs.

dpetran (approved):
:bucket:

PR #1062, opened 2025-07-11
fix handling of GRAPH patterns in INSERT DATA operations

dpetran:
We were losing all but the first triple produced from an INSERT DATA operation when they were nested under a GRAPH clause, this ensures that all the triples are preserved.


zonotope (approved):
💅🏾

PR #1065, opened 2025-07-14
fix translation of multi-pattern OPTIONAL sparql patterns

dpetran:
We were translating multi-pattern SPARQL OPTIONAL patterns into jld-query patterns that look like this:

```[:optional [<p1> <p2>]]```

Which gets groups both patterns into a block when parsed:

```:where [... [:optional [<p1> <p2>]]]```

This produces a different result from the intended semantics of OPTIONAL. This commit changes the translation into this:

```[:optional <p1> <p2>]```

Which gets parsed to this:

```:where [... [:optional <p1>] [:optional <p2>]]```

Which provides the semantics that SPARQL expects.


zonotope (approved):
🏛️

PR #1066, opened 2025-07-15
Migrate to fluree.crypto with Ed25519 and standards-based verifiable credentials

bplatz:
## Summary

This PR migrates the codebase to use the updated fluree.crypto library, focusing on standards-based verifiable credentials support and modernizing cryptographic implementations.

#### Cryptographic Updates
- **Ed25519 Support**: Migrated from Secp256k1 to Ed25519 cryptographic signatures for better standards compliance
- **DID Format**: Changed from proprietary `did:fluree:` to standards-based `did:key:` format
- **JWS Algorithm**: Updated from `ES256K-R` to `EdDSA` for JSON Web Signatures
- **Proof Type**: Changed from `EcdsaSecp256k1RecoverySignature2020` to `Ed25519Signature2018`

#### Account ID Format Changes
- Account IDs now use Ed25519 multicodec prefix (`ed01`) instead of Secp256k1 (`e7`)
- Account ID generation now uses multibase encoding with `did:key:` prefix
- All test fixtures updated to reflect new Ed25519 key format

#### API Updates
- Updated `crypto/pub-key-from-private` → `crypto/public-key-from-private`
- Added support for `crypto/account-id-from-private` function
- All `crypto/create-jws` calls now include `{:include-pubkey true}` parameter
- Updated DID namespace constant from `"did:fluree:"` to `"did:key:"`

#### Dependency Updates
- **fluree.crypto**: Updated to commit `0db37977cccfcbe7f07ee8f43ab7979881f0154d`
- **Clojure**: Updated to 1.12.1
- **Shadow-cljs**: Updated to 2.28.15
- **Removed**: Obsolete crypto dependencies (`@fluree/sjcl`, `js-sha3`, `scrypt-js`)
- **Added**: Modern crypto libraries (`@noble/ed25519`, `@noble/hashes`)

#### Test Updates
- All test fixtures migrated from Secp256k1 to Ed25519 keys
- Updated expected commit hashes and DID values
- All 207 tests passing with new cryptography

#### Java Version Requirement
**⚠️ BREAKING**: This update requires **Java 17 minimum** (previously Java 11)

#### Account ID Format
Account IDs generated from private keys now use a different format:
- **Before**: Secp256k1-based with `did:fluree:` prefix
- **After**: Ed25519-based with `did:key:` prefix

#### Verifiable Credentials
- Signature algorithm changed from `ES256K-R` to `EdDSA`
- Proof type changed from `EcdsaSecp256k1RecoverySignature2020` to `Ed25519Signature2018`
- JWS creation now requires `{:include-pubkey true}` parameter

### Impact
- **Bundle Size**: No significant change (browser build remains 5.5MB)
- **Performance**: Ed25519 signatures are generally faster than Secp256k1
- **Standards Compliance**: Better alignment with W3C DID and Verifiable Credentials standards
- **Security**: Ed25519 provides equivalent security with smaller key sizes

### Testing
All 207 tests are passing with the new cryptographic implementation. The migration maintains backward compatibility for existing data while using the new crypto format for new operations.

## Test plan
- [x] All existing tests pass (207 tests, 1492 assertions)
- [x] Browser build compiles successfully
- [x] Crypto function migrations verified
- [x] DID format changes tested
- [x] JWS operations with new parameters confirmed


src/fluree/db/flake/commit_data.cljc
```diff
@@ -326,15 +326,16 @@
...
-    (cond-> commit
-      txn-id (assoc :txn txn-id)
-      author (assoc :author author)
-      issuer (assoc :issuer {:id issuer})
-      prev-commit (assoc :previous prev-commit)
-      message (assoc :message message)
-      annotation (assoc :annotation annotation)
-      tag (assoc :tag tag))))
+                               :time time))
+        result (cond-> commit
```
zonotope: Why is this extra binding necessary when it's immediately returned?

deps.edn
```diff
@@ -1,4 +1,4 @@
-{:deps {org.clojure/clojure              {:mvn/version "1.11.3"}
+{:deps {org.clojure/clojure              {:mvn/version "1.11.4"}
```
zonotope, line 1: Does 1.12 not work? I remember there being an issue with updating to it on this codebase when it was first released. Is that still the case?

bplatz: I think we need to go to 1.12.1 probably sooner rather than later. The only issue for us is Eastwood breaks, but only for defrecord linting so we could exclude those namespaces and still use Eastwood.

The maintainer of Eastwood indicates little time to spend on it, and it in turn depends on another library that needs to be updated which isn't yet. There is no Eastwood alternative that I can find and there doesn't seem to be anyone working on one. clj-kondo is great but designed to be fast and therefore doesn't do the level of introspection Eastwood does (e.g. reflection and other deeper issues).

I actually have some work on a different branch related to GraalVM where I used 1.12.1 and was thinking of proposing the move... but been trying to avoid it for the above reason.

src/fluree/db/flake/commit_data.cljc
bplatz: I was debugging the 'result' and logging it out, but forgot to put it back. Will address now.

deps.edn
zonotope: Yes, it was eastwood. That was the same issue as before.

zonotope (approved):

PR #1067, opened 2025-07-15
fix blank node variable grammar

dpetran:
Our grammar for blank node variables was underspecifying the blank node prefix, checking for `_` instead of `_:`. This meant string literals that began with an underscore were being parsed as variables, which lead to unexpected query results.


zonotope (approved):
🍧

PR #1068, opened 2025-07-16
fix xhttp response handling

dpetran:
You can't slurp a string, and body is sometimes a string depending on how the request is created. This change was needed in order to get tests passing in server.


bplatz (approved):
⚡

PR #1072, opened 2025-07-22
Add GraalVM native image support

bplatz:
## Summary

This PR adds support for building GraalVM native images from Fluree DB, enabling ahead-of-time compilation for applications that depend on fluree/db.

### GraalVM Support Infrastructure
- Added `graalvm/` directory with build scripts and test suite as reference implementation for downstream projects
- Introduced conditional compilation support with compile-time macros in `eval.cljc`
- Added SCI (Small Clojure Interpreter) for runtime code evaluation in native images

### Dependency Migrations
- **Replaced http-kit with Java 11's native HTTP client** - Removes incompatible dependency and uses built-in HTTP functionality
- **Replaced Cheshire with jsonista** - We already include jsonista elsewhere, so this removes a redundant JSON library and improves performance

### Code Changes
- Modified `fluree.db.query.exec.eval` to use SCI for query evaluation when compiled with GraalVM
- Updated `fluree.db.util.xhttp` to use Java 11 HttpClient instead of http-kit
- Migrated `fluree.db.util.json` from Cheshire to jsonista
- Fixed reflection warnings in `fluree.db.flake`

### Build Configuration
- Removed eastwood from CI/CD checks (temporarily disabled until it supports Clojure 1.12)
- Updated dependencies: json-ld and crypto libraries to latest versions

## Notes

- The `graalvm/` directory contains reference build scripts and tests for downstream applications building native images with Fluree DB
- Native image builds are not included in CI/CD as they take 5+ minutes and would typically be built by downstream projects
- All existing tests pass with these changes

## Testing

Run the GraalVM test suite:
```bash
cd graalvm
./build-test.sh
```

Run standard tests:
```bash
make test
```

### DIFF
--- a/script/sync_package_json.clj
+++ b/script/sync_package_json.clj
@@ -1,17 +1,17 @@
 (ns sync-package-json
-  (:require [jsonista.core :as j]))
+  (:require [jsonista.core :as json]))
 
 (defn js-deps
   []
-  (-> "package.json" slurp (j/read-value (j/object-mapper {:decode-key-fn true})) :dependencies))
+  (-> "package.json" slurp (json/read-value (json/object-mapper {:decode-key-fn true})) :dependencies))
 
 (defn run
   [& args]
   (let [version (first args)
         target-package-json-file (second args)
-        target-package-json (-> target-package-json-file slurp (j/read-value (j/object-mapper {:decode-key-fn true})))
+        target-package-json (-> target-package-json-file slurp (json/read-value (json/object-mapper {:decode-key-fn true})))
         write-package-json #(spit target-package-json-file %)
-        pretty-mapper (j/object-mapper {:pretty true})
+        pretty-mapper (json/object-mapper {:pretty true})
         sync-js-deps? (and (> 2 (count args)) (= "--node" (nth args 2)))
         sync-js-deps #(if sync-js-deps?
                         (assoc % :dependencies (js-deps))
@@ -20,5 +20,5 @@
     (-> target-package-json
         (assoc :version version)
         sync-js-deps
-        (j/write-value-as-string pretty-mapper)
+        (json/write-value-as-string pretty-mapper)
         write-package-json)))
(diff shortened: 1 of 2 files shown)


script/sync_package_json.clj
```diff
@@ -1,19 +1,17 @@
 (ns sync-package-json
-  (:require [clojure.edn :as edn]
-            [cheshire.core :as json]))
+  (:require [jsonista.core :as j]))
```
zonotope: Tiny nitpick, but I think single letter and heavily abbreviated namespace aliases should be avoided. `json` carries way more information than `j`. Those single letter aliases always mean multiple scrolls back to the top of the file to figure out what that namespace does. I will make an exception for extremely common libraries with well established aliases like `str` for `clojure.string`, and `ig` for `integrant`.

src/fluree/db/flake.cljc
```diff
@@ -497,23 +497,22 @@
...
-                                      const/$xsd:string (* 2 (count o))
-                                      const/$id 30
-                                      const/$xsd:boolean 1
-                                      const/$xsd:long 8
-                                      const/$xsd:int 4
-                                      const/$xsd:short 2
-                                      const/$xsd:double 8
-                                      const/$xsd:float 4
-                                      const/$xsd:byte 1
+        o-size  (condp = (dt f)
```
zonotope, line 500: Why did you remove `case+` on clj? I was under the impression that it had dramatically better performance, but I never tested it.

zonotope (approved):
🧑‍🎄

src/fluree/db/flake.cljc
bplatz: I tried a few things and was unable to come up with an alternative and still get GraalVM to work with it, but that isn't saying there isn't an alternative. 

The main reason I didn't sweat it too much is I still think we should eliminate this whole code path in lieu of returning size from writing of the data (and possibly commit) files on disk. It is more accurate in terms of actual size and it is something that shouldn't require a bunch of extra processing.

That would give us more speed and better accuracy, and less code. For now this still exists though and I haven't thought though if the removal of this might affect other things I am not considering.

script/sync_package_json.clj
bplatz: @zonotope Addressed this in a new commit FYI.

PR #1073, opened 2025-07-24
Add nameservice query functionality and separate nameservice storage

bplatz:
## Summary

- Moves nameservice records to dedicated `ns@v1/` folder separate from ledger data
- Adds `query-nameservice` API function to query nameservice records using standard Fluree query syntax
- Implements migration check that has been tested to ensure compatibility

### Nameservice Storage Separation
- Nameservice records now stored in `ns@v1/` directory instead of being co-located with ledger data
- Maintains backwards compatibility through migration logic

### New Query API
- `fluree.db.api/query-nameservice` - Query all nameservice records using standard Fluree query syntax
- Creates temporary in-memory ledger from aggregated nameservice records
- Supports full query capabilities including filters, projections, and complex patterns

### Storage Abstraction
- Added `ListableStore` protocol to enable listing files across different storage backends
- Implemented for `FileStore`, `MemoryStore`, and `S3Store`
- Extended `iNameService` protocol with `all-records` function

### Implementation Notes
- Current implementation aggregates all nameservice records into memory for querying
- Not optimized for performance but provides working foundation for future improvements
- Temporary ledger is created per query and disposed after execution
- S3 implementation uses existing list operations with proper pagination handling

### Testing
- Comprehensive test suite for both memory and file-based storage
- Tests verify query functionality across different storage backends
- File storage tests use `with-temp-dir` for proper cleanup

## Examples

Query all nameservice records:
```clojure
(query-nameservice conn 
  {"@context" {"f" "https://ns.flur.ee/ledger#"}
   "select" ["?s" "?p" "?o"]
   "where" [{"@id" "?s" "?p" "?o"}]})
```

Find ledgers on specific branch:
```clojure
(query-nameservice conn
  {"@context" {"f" "https://ns.flur.ee/ledger#"}
   "select" ["?ledger"]
   "where" [{"@id" "?ns"
             "f:ledger" "?ledger"
             "f:branch" "main"}]})
```

### DIFF
--- a/test/fluree/db/nameservice_query_test.clj
+++ b/test/fluree/db/nameservice_query_test.clj
@@ -43,8 +43,8 @@
           (let [query {"select" ["?s" "?p" "?o"]
                        "where" [{"@id" "?s" "?p" "?o"}]}
                 result @(fluree/query-nameservice conn query {})]
-            ;; Should have data from all ledgers plus temporary ledger metadata
-            (is (> (count result) 10) "Should have multiple records")
+            ;; Should have data from all ledgers plus nameservice metadata (at least 3 ledgers created)
+            (is (>= (count result) 9) "Should have at least 9 records from 3 ledgers and their metadata")
 
             ;; Check that we have nameservice-specific data by looking for ledger references
             (let [ledger-records (filter (fn [[_ p _]]


src/fluree/db/api.cljc
```diff
@@ -662,6 +663,31 @@
...
+
+  Returns promise resolving to query results."
+  ([conn query] (query-nameservice conn query {}))
+  ([conn query opts]
+   (validate-connection conn)
+   (promise-wrap
+    (go-try
+      ;; Get the nameservice from the connection's primary publisher
+      (let [primary-publisher (:primary-publisher conn)]
+        (if primary-publisher
```
zonotope: you can combine these two lines with `if-let` or, better yet, `if-some`

src/fluree/db/connection.cljc
```diff
@@ -307,21 +307,30 @@
...
+      (let [;; Extract minimal data from nameservice
+            commit-address (get-in ns-record ["f:commit" "@id"])
+            index-address  (get-in ns-record ["f:index" "@id"])
+
+            ;; Load full commit from disk
+            commit         (<? (commit-storage/load-commit-with-metadata commit-catalog
+                                                                         commit-address
+                                                                         index-address))
+
+            _              (log/debug "Attempting to load from address:" address
```
zonotope: I think this log should come before the previous line. Right now it wouldn't be printed if there was an error loading the commit.

src/fluree/db/nameservice/ipns.cljc
```diff
@@ -69,7 +69,10 @@
       (ipfs/read ipfs-endpoint address)))
   (alias [_ ledger-address]
     (let [[_ _ alias] (address-parts ledger-address)]
-      alias)))
+      alias))
+  (all-records [_]
+    ;; IPNS doesn't support listing all records
```
zonotope: I think this means that `all-records` should live in a different protocol that ipns doesn't implement instead of having an incomplete implementation.

bplatz: IPNS will support it, but it is broken anyhow and clj-kondo won't pass with something being here. We need to revisit IPFS/IPNS and I think the best time will be when some of these more core changes are in place we can do it once.

zonotope: Remote also doesn't support it right now, although we could enable it there too if we wanted

src/fluree/db/connection.cljc
bplatz: This is resolved...

src/fluree/db/api.cljc
bplatz: Address this in updated commit.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -19,62 +20,93 @@
...
-   (log/info "Fetching commit from record:" record)
-   (let [branch-iri (if branch
-                      (str (get record "@id") "(" branch ")")
-                      (get record "defaultBranch"))]
-     (some #(when (= (get % "@id") branch-iri)
-              (get % "commit"))
-           (get record "branches")))))
+  ([record _branch]
+   ;; Always return the record itself for new format
+   record))
```
zonotope: Shouldn't we just remove this `get-commit` function all together? It isn't used anywhere outside this namespace, and the usage in here could easily also be removed to just use the record directly.

src/fluree/db/remote_system.cljc
```diff
@@ -138,6 +138,9 @@
     (remote-lookup system-state ledger-address))
   (alias [_ ledger-address]
     ledger-address)
+  (all-records [_]
+    ;; Remote systems don't support querying all records
```
zonotope: I think we should put an explicit TODO here to implement this.

src/fluree/db/nameservice/ipns.cljc
zonotope: I think we should put a TODO here about implementing it eventually then.

src/fluree/db/migrations/nameservice_v1.cljc
```diff
@@ -0,0 +1,160 @@
...
+(defn find-old-nameservice-files
+  "Find old nameservice files in ledger directories - handles both root level and nested paths"
+  [file-store]
+  (go-try
+    (let [root-path   (:root file-store)
+          ;; Find all .json files recursively, excluding ns@v1 and commit directories
+          all-json-files (find-json-files-recursively root-path)
+          ;; Filter to only nameservice files (those that match ledger naming pattern)
+          ;; clj-kondo false positive - ledger-path is used in the clj branch  
+          ns-files #_{:clj-kondo/ignore [:unused-binding]}
```
zonotope: I think this would be better in the clj-kondo configuration files to avoid adding noise to the code.

test/fluree/db/nameservice_query_test.clj
```diff
@@ -0,0 +1,133 @@
...
+                            "@graph" [{"@id" "test:person4"
+                                       "@type" "Person"
+                                       "name" "David"}]}))
+
+        (testing "Query all nameservice records"
+          (let [query {"select" ["?s" "?p" "?o"]
+                       "where" [{"@id" "?s" "?p" "?o"}]}
+                result @(fluree/query-nameservice conn query {})]
+            ;; Should have data from all ledgers plus temporary ledger metadata
+            (is (> (count result) 10) "Should have multiple records")
```
zonotope: I think the annotation here should be more precised. 9 records returned still counts as "multiple" even though it would cause the test to fail

src/fluree/db/migrations/nameservice_v1.cljc
bplatz: I just implemented a full javascript version here so the linting message no longer applied.

Because this is just temporary code, I specifically didn't want to made changes to the main repo for something that will be short-lived and create confusion later on.

Because I know there are not people running transaction servers of the new version in JavaScript I didn't fully implement it, but now I just implemented it so this is resolved.

zonotope (approved):
🦈

PR #1074, opened 2025-07-25
Replace Cognitect AWS client with direct S3 implementation

bplatz:
## Summary
This PR replaces the Cognitect AWS client with a direct S3 implementation to resolve HTTP client conflicts in upstream projects.

## Changes
- Implement S3 REST API client using Java 11 HTTP client
- Add AWS Signature V4 authentication
- Support all required S3 operations: GET, PUT, DELETE, LIST
- Add PUT method to xhttp for S3 uploads
- Update testcontainers tests to work with LocalStack
- Remove unused parameters from internal S3 functions
- Remove Cognitect AWS dependencies from deps.edn

## Testing
- All unit tests pass (215 tests, 0 failures)
- Docker/testcontainers tests pass (2 tests, 0 failures)
- clj-kondo: 0 errors, 0 warnings
- cljfmt: All source files formatted correctly

## Benefits
- Eliminates HTTP client dependency conflicts in upstream projects
- Reduces dependency footprint
- Maintains full S3 functionality for Fluree database storage
- Compatible with LocalStack for local testing

### DIFF
--- a/src/fluree/db/storage/s3.clj
+++ b/src/fluree/db/storage/s3.clj
@@ -252,8 +252,7 @@
                              :path full-path
                              :body data
                              :credentials credentials})
-                ch)
-    ch))
+                ch)))
 
 (defn- tag-matches?
   "Check if an XML element's tag matches the given name, ignoring namespace"
@@ -274,14 +273,14 @@
   [xml-str]
   (let [parsed (xml/parse-str xml-str)
         contents (xml-seq parsed)]
-    {:IsTruncated (= "true" (get-xml-text "IsTruncated" contents))
-     :NextContinuationToken (get-xml-text "NextContinuationToken" contents)
-     :Contents (for [x contents
+    {:truncated? (= "true" (get-xml-text "IsTruncated" contents))
+     :next-continuation-token (get-xml-text "NextContinuationToken" contents)
+     :contents (for [x contents
                      :when (tag-matches? "Contents" x)]
                  (let [obj-content (:content x)]
-                   {:Key (get-xml-text "Key" obj-content)
-                    :Size (get-xml-text "Size" obj-content)
-                    :LastModified (get-xml-text "LastModified" obj-content)}))}))
+                   {:key (get-xml-text "Key" obj-content)
+                    :size (get-xml-text "Size" obj-content)
+                    :last-modified (get-xml-text "LastModified" obj-content)}))}))
 
 (defn s3-list*
   "List objects in S3 with optional continuation token"
@@ -305,9 +304,9 @@
                                          :credentials credentials
                                          :query-params query-params}))
                parsed (parse-list-objects-response response)]
-           (>! ch (update parsed :Contents
+           (>! ch (update parsed :contents
                           (fn [contents]
-                            (mapv #(select-keys % [:Key]) contents)))))
+                            (mapv #(select-keys % [:key]) contents)))))
          (catch Exception e
            (>! ch e))))
      ch)))
@@ -319,9 +318,9 @@
     (go-loop [continuation-token nil]
       (let [results (<! (s3-list* client path continuation-token))]
         (>! ch results)
-        (if (and (:IsTruncated results)
-                 (:NextContinuationToken results))
-          (recur (:NextContinuationToken results))
+        (if (and (:truncated? results)
+                 (:next-continuation-token results))
+          (recur (:next-continuation-token results))
           (async/close! ch))))
     ch))
 
--- a/test/fluree/db/storage/s3_testcontainers_test.clj
+++ b/test/fluree/db/storage/s3_testcontainers_test.clj
@@ -178,7 +178,7 @@
                   store (s3/->S3Store nil (s3/get-credentials) bucket "us-east-1" "indexing")
                   list-ch (s3/s3-list store "")
                   objects-resp (<!! list-ch)
-                  object-keys (map :Key (:Contents objects-resp))]
+                  object-keys (map :key (:contents objects-resp))]
 
               (is (= [50] count-result) "Should have 50 persons")
               (is (pos? (count object-keys)) "Should have objects in S3")


src/fluree/db/storage/s3.clj
```diff
@@ -1,113 +1,339 @@
...
+      (URLEncoder/encode "UTF-8")
+      ;; AWS requires specific encoding
+      (str/replace "+" "%20")
+      (str/replace "*" "%2A")
+      (str/replace "%7E" "~")))
+
+(defn canonical-uri
+  "Create canonical URI for AWS signature"
+  [path]
+  (if (empty? path)
```
zonotope: It looks like this `if` is unnecessary as every case would be covered by the else block.

src/fluree/db/storage/s3.clj
```diff
@@ -1,113 +1,339 @@
...
-          _        (log/debug "S3 response:" resp)
-          body-str (when in
-                     (with-open [out (ByteArrayOutputStream.)]
-                       (io/copy in out)
-                       (.close ^Closeable in)
-                       (String. (.toByteArray out))))]
-      (cond-> resp
-        body-str (assoc :Body body-str)))))
+;; AWS Signature V4 constants
+(def algorithm "AWS4-HMAC-SHA256")
```
zonotope: I think these constants should be prefixed. `aws-signing-algorithm`, and `aws-service`, for example

src/fluree/db/storage/s3.clj
```diff
@@ -1,113 +1,339 @@
...
+
+(defn create-canonical-request
+  "Create the canonical request for AWS Signature V4"
+  [method uri query-params headers payload-hash]
+  (str method "\n"
+       uri "\n"
+       (or query-params "") "\n"
+       (canonical-headers headers) "\n"
+       (signed-headers headers) "\n"
+       payload-hash))
```
zonotope:
```suggestion
  (str/join \newline [method
                      uri 
                      (or query-params "")
                      (canonical-headers headers)
                      (signed-headers headers)
                      payload-hash]))
```

src/fluree/db/storage/s3.clj
```diff
@@ -1,113 +1,339 @@
...
+
+(defn create-string-to-sign
+  "Create the string to sign for AWS Signature V4"
+  [amz-date credential-scope canonical-request-hash]
+  (str algorithm "\n"
+       amz-date "\n"
+       credential-scope "\n"
+       canonical-request-hash))
```
zonotope:
```suggestion
  (str/join \newline [algorithm 
                      amz-date 
                      credential-scope 
                      canonical-request-hash]))
```

src/fluree/db/storage/s3.clj
```diff
@@ -1,113 +1,339 @@
...
+
+(defn sign-request
+  "Sign an S3 request using AWS Signature V4"
+  [{:keys [method path headers payload region bucket credentials query-params]}]
+  (let [{:keys [access-key secret-key session-token]} credentials
+        now (Instant/now)
+        amz-date (.format ^DateTimeFormatter amz-date-formatter now)
```
zonotope: I think these type hints for the `amz-date-formatter` and `date-stamp-formatter` should go on the definition.

src/fluree/db/storage/s3.clj
```diff
@@ -1,113 +1,339 @@
...
-                   :request {:Bucket s3-bucket, :Key full-path, :Body data}}]
-    (aws/invoke-async s3-client req)
+  "Write an object to S3"
+  [client path data]
+  (let [{:keys [credentials bucket region prefix]} client
+        ch (async/promise-chan)
+        full-path (str prefix "/" path)]
+    (go
+      (try
+        (let [response (<? (s3-request {:method "PUT"
```
zonotope: I think this function would have the same effect with `(async/pipe (s3-request {...}) ch)` instead of using `(go ... try... let ...)`

src/fluree/db/storage/s3.clj
```diff
@@ -1,113 +1,339 @@
...
+          (>! ch e))))
     ch))
 
+(defn parse-list-objects-response
+  "Parse XML response from S3 ListObjectsV2"
+  [xml-str]
+  (let [parsed (xml/parse-str xml-str)
+        contents (xml-seq parsed)
+        ;; Helper to get text content regardless of namespace
+        get-text (fn [tag-name elements]
```
zonotope: this function should be defined on it's own. It doesn't close over any local bindings, so there's no need to re-define it on each invocation

src/fluree/db/storage/s3.clj
```diff
@@ -1,113 +1,339 @@
...
+                   (first (for [x elements
+                                :when (and (:tag x)
+                                           (= tag-name (name (:tag x))))]
+                            (first (:content x)))))
```
zonotope:
```suggestion
                   (some (fn [element]
                           (when (some-> element :tag name (= tag-name))
                             (first (:content element))))
                         elements)))
```

src/fluree/db/storage/s3.clj
```diff
@@ -1,113 +1,339 @@
...
+        ;; Helper to check tag name ignoring namespace
+        tag-matches? (fn [tag-name elem]
```
zonotope: This function should have its own definition. It's implementation shares a lot with the other function that should be defined on its own, so that one should probably be defined in terms of this one.

src/fluree/db/storage/s3.clj
```diff
@@ -1,113 +1,339 @@
...
+                              (not= full-path "/") (assoc "prefix" full-path)
+                              continuation-token (assoc "continuation-token" continuation-token))
+               response (<? (s3-request {:method "GET"
+                                         :bucket bucket
+                                         :region region
+                                         :path ""
+                                         :credentials credentials
+                                         :query-params query-params}))
+               parsed (parse-list-objects-response response)]
+           (>! ch {:IsTruncated (:truncated? parsed)
```
zonotope: It looks like there's a lot of wasteful conversion here. For example, the key indicating if the response was truncated goes from `"IsTruncated"` -> `:truncated?` -> `:IsTruncated`. The same is tru for the continuation token and contents, as well as the keys inside of the maps in the contents list.

bplatz:
@zonotope I've addressed all the suggestions here. I did rush to get this out originally as it was holding up the serverless deployment but I probably should have kept it as Draft until I had a chance to go through it a bit more.

src/fluree/db/storage/s3.clj
```diff
@@ -1,113 +1,335 @@
...
+        ch (async/promise-chan)
+        full-path (str prefix "/" path)]
+    (async/pipe (s3-request {:method "PUT"
+                             :bucket bucket
+                             :region region
+                             :path full-path
+                             :body data
+                             :credentials credentials})
+                ch)
     ch))
```
zonotope: line 256 is now redundant. `async/pipe` should already return the `to` channel, which is `ch` in this case, so there should be no need to return it explicitly.

src/fluree/db/storage/s3.clj
```diff
@@ -1,113 +1,335 @@
...
+          (when (tag-matches? tag-name element)
+            (first (:content element))))
+        elements))
+
+(defn parse-list-objects-response
+  "Parse XML response from S3 ListObjectsV2"
+  [xml-str]
+  (let [parsed (xml/parse-str xml-str)
+        contents (xml-seq parsed)]
+    {:IsTruncated (= "true" (get-xml-text "IsTruncated" contents))
```
zonotope: since we're converting these strings to keyword anyway, we might as well make them idomatic with `:truncated?`, `:next-continuation-token`, `:contents`, `:key`, `:size`, `:last-modified`.

zonotope (approved):
🛻

PR #1075, opened 2025-07-28
Add manual indexing API for external indexing processes

bplatz:
## Summary

Add manual indexing API that allows external processes to trigger indexing without creating transactions. This enables indexing to happen on separate servers and was specifically designed to support AWS Lambda indexing processes.

## Key Features

- **Manual Indexing API**: New `fluree/trigger-index` function for triggering indexing on demand
- **Blocking Mode**: Wait for indexing completion - required for AWS Lambda since background tasks don't work and Lambda shuts down after function returns
- **Non-blocking Mode**: Queue indexing and return immediately
- **Indexing Control**: `:indexing-disabled` flag to disable automatic indexing via connection configuration

## API Usage

```clojure
;; Queue indexing (returns immediately)
@(fluree/trigger-index conn "my-ledger")

;; Wait for indexing to complete (for Lambda)
@(fluree/trigger-index conn "my-ledger" {:block? true})

;; With timeout and branch
@(fluree/trigger-index conn "my-ledger" {:block? true 
                                         :timeout 600000
                                         :branch "main"})
```

## Configuration

```clojure
;; Disable automatic indexing when using external indexing
(fluree/connect-memory {:defaults {:indexing {:indexing-disabled true}}})
```

## Changes to commit\! Response

Modified `fluree/commit\!` to include additional metadata when `:meta true` option is used:
- `:novelty-size` - Current novelty size in bytes
- `:indexing-disabled` - Whether automatic indexing is disabled
- `:indexing-needed` - Whether novelty size exceeds reindex threshold
- `:index-t` - The t value of the last indexed commit

## Testing

Tests are located in the new `fluree.db.indexing-test` namespace, covering:
- Manual indexing API functionality
- Blocking and non-blocking modes
- Disabled indexing behavior
- File-based indexing and loading from disk
- Branch state updates after indexing

## Implementation Details

- Fixed `newer-commit?` function to handle nil t values correctly
- Fixed FlakeDB indexer to always return database instead of nil
- Added `connection/trigger-ledger-index` for core implementation
- Integrated with existing branch indexing queue infrastructure

### DIFF
--- a/src/fluree/db/api.cljc
+++ b/src/fluree/db/api.cljc
@@ -808,7 +808,7 @@
      (reasoner/reasoned-facts db grouping))))
 
 (defn trigger-index
-  "Manually triggers indexing for a ledger without requiring a transaction.
+  "Manually triggers indexing for a ledger and waits for completion.
   
   This is useful for external indexing processes (e.g., AWS Lambda) that need
   to ensure a ledger is indexed without creating new transactions.
@@ -818,19 +818,16 @@
     ledger-alias - The alias/name of the ledger to index
     opts - (optional) Options map:
       :branch - Branch name (defaults to main branch)
-      :block? - If true, waits for indexing to complete before returning
-      :timeout - Max wait time in ms when blocking (default 300000 / 5 minutes)
+      :timeout - Max wait time in ms (default 300000 / 5 minutes)
   
-  Returns a promise that resolves to:
-    - If :block? is false: {:status :queued}
-    - If :block? is true: {:status :success :db <indexed-db>} or {:status :error :error <error>}
+  Returns a promise that resolves to the indexed database object.
+  Throws an exception if indexing fails or times out.
   
   Example:
-    ;; Queue indexing (returns immediately)
-    @(trigger-index conn \"my-ledger\")
-    
-    ;; Wait for indexing to complete (useful in Lambda)
-    @(trigger-index conn \"my-ledger\" {:block? true})"
+    ;; Trigger indexing and wait for completion
+    (let [indexed-db @(trigger-index conn \"my-ledger\")]
+      ;; Use indexed-db...
+      )"
   ([conn ledger-alias]
    (trigger-index conn ledger-alias nil))
   ([conn ledger-alias opts]
--- a/src/fluree/db/api.cljc
+++ b/src/fluree/db/api.cljc
@@ -806,3 +806,34 @@
   ([db opts]
    (let [grouping (:group-by opts)]
      (reasoner/reasoned-facts db grouping))))
+
+(defn trigger-index
+  "Manually triggers indexing for a ledger without requiring a transaction.
+  
+  This is useful for external indexing processes (e.g., AWS Lambda) that need
+  to ensure a ledger is indexed without creating new transactions.
+  
+  Parameters:
+    conn - Database connection
+    ledger-alias - The alias/name of the ledger to index
+    opts - (optional) Options map:
+      :branch - Branch name (defaults to main branch)
+      :block? - If true, waits for indexing to complete before returning
+      :timeout - Max wait time in ms when blocking (default 300000 / 5 minutes)
+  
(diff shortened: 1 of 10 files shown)


src/fluree/db/branch.cljc
```diff
@@ -192,11 +201,12 @@
   then the ledger's latest commit t (in branch-data). The db 't' and db commit 't'
   should be the same at this point (just after committing the db). The ledger's latest
   't' should be the same (if just updating an index) or after the db's 't' value."
-  [{:keys [state index-queue] :as branch-map} new-db index-files-ch]
+  [{:keys [state index-queue indexing-opts] :as branch-map} new-db index-files-ch]
   (let [updated-db (-> state
                        (swap! update-commit (policy/root-db new-db))
                        :current-db)]
-    (enqueue-index! index-queue updated-db index-files-ch)
+    (when-not (:indexing-disabled indexing-opts)
```
zonotope, line 208: What sets the `:indexing-disabled` option?

src/fluree/db/connection.cljc
```diff
@@ -457,3 +458,41 @@
...
+                  timeout-ch (async/timeout timeout)]
+              (branch/enqueue-index! index-queue current-db nil complete-ch)
+              (async/alt!
+                complete-ch ([result] result)
+                timeout-ch {:status :error
+                            :error (ex-info "Indexing timeout" {:timeout timeout})}))
+            ;; Non-blocking mode - just enqueue
+            (do
+              (branch/enqueue-index! index-queue current-db nil)
+              {:status :queued})))))))
```
zonotope: This needs to be wrapped in a channel for the `promise-wrap` in the `fluree.db.api/trigger-index` function to work.

src/fluree/db/branch.cljc
zonotope, line 208: What sets the `:indexing-disabled` option?

src/fluree/db/api.cljc
```diff
@@ -806,3 +806,34 @@
...
+  
+  This is useful for external indexing processes (e.g., AWS Lambda) that need
+  to ensure a ledger is indexed without creating new transactions.
+  
+  Parameters:
+    conn - Database connection
+    ledger-alias - The alias/name of the ledger to index
+    opts - (optional) Options map:
+      :branch - Branch name (defaults to main branch)
+      :block? - If true, waits for indexing to complete before returning
```
zonotope: I think having boolean options for control flow increases code complexity. I think things would be simpler with two different functions in each place where the `:block?` option is read instead.

bplatz: @zonotope I've updated it so it always blocks, and then the caller always has an option to wrap it in a (future ...) or similar if they choose to have it non-blocking. It therefore now just returns the indexed db once complete (or exception if there is one).

src/fluree/db/branch.cljc
bplatz: @zonotope this is set with the connection config the same way other indexing options are set (like min/max reindex values).

src/fluree/db/connection.cljc
```diff
@@ -457,3 +458,30 @@
...
+   Returns the indexed database object or throws an exception on failure/timeout."
+  [conn ledger-alias opts]
+  (go-try
+    (let [{:keys [branch timeout]
+           :or {timeout 300000}} opts
+          ledger (<? (load-ledger-alias conn ledger-alias))
+          current-db (ledger/current-db ledger branch)
+          index-queue (-> ledger
+                          (ledger/get-branch-meta branch)
+                          :index-queue)
```
zonotope: I don't think we should be reaching into the index queue from here as that should be a hidden implementation detail that should change. I think we should add `trigger-index` apis to both the ledger and branch so we don't break the abstraction barrier.

bplatz: @zonotope addressed in 4817326feb0cf6f41fee869acdc647668c5e0795

PR #1076, opened 2025-07-28
Handle authorization checks on reversed properties in subgraph selector

dpetran:
When do formatting a SubgraphSelector with a reversed property, we were using a flake-xf that changed the result from a flake to a sid, which cannot be handled by the authorize-flakes transducer.

This PR moves the subject extraction out of the flake-xf and only applies it afterward as part of the formatting.

Fixes https://github.com/fluree/core/issues/161


zonotope (approved):
🩻

PR #1078, opened 2025-07-31
Fix S3 storage path handling

bplatz:
## Summary
- Fixes double slash issue in S3 file paths (e.g., `prefix//file.json`)
- Normalizes S3 prefix to always end with `/` when non-empty
- Simplifies S3 address format to `fluree:s3://path`

## Changes
- Modified `s3/open` to normalize prefix during store creation
- Removed conditional path concatenation logic from all S3 functions
- Updated `s3-address` to exclude identifier from path construction
- Added tests to verify correct S3 path structure

## Test plan
- [x] Unit tests pass
- [x] Docker/LocalStack integration tests pass
- [x] Verified S3 paths with AWS CLI listing

### DIFF
--- a/src/fluree/db/storage/s3.clj
+++ b/src/fluree/db/storage/s3.clj
@@ -81,9 +81,16 @@
       (str/replace "*" "%2A")
       (str/replace "%7E" "~")))
 
+(defn encode-s3-path
+  "Encode S3 path segments individually to match S3's automatic encoding"
+  [path]
+  (let [segments (str/split path #"/")]
+    (str/join "/" (map url-encode segments))))
+
 (defn canonical-uri
   "Create canonical URI for AWS signature"
   [path]
+  ;; Path must already be encoded when passed to this function
   (str "/" path))
 
 (defn canonical-query-string
@@ -195,15 +202,17 @@
     :or {method "GET"
          headers {}}}]
   (go-try
-    (let [query-string (canonical-query-string query-params)
-          url (str (build-s3-url bucket region path)
+    (let [;; Encode path segments for both URL and signature to match S3's encoding
+          encoded-path (encode-s3-path path)
+          query-string (canonical-query-string query-params)
+          url (str (build-s3-url bucket region encoded-path)
                    (when query-string (str "?" query-string)))
           headers-with-content-type (if (and (= method "PUT") body)
                                       (assoc headers "Content-Type" "application/octet-stream")
                                       headers)
           signed-headers (sign-request
                           {:method method
-                           :path path
+                           :path encoded-path  ;; Use encoded path for signature
                            :headers headers-with-content-type
                            :payload body
                            :region region
--- a/src/fluree/db/storage/s3.clj
+++ b/src/fluree/db/storage/s3.clj
@@ -225,7 +225,7 @@
   [client path]
   (let [{:keys [credentials bucket region prefix]} client
         ch (async/promise-chan)
-        full-path (str prefix "/" path)]
+        full-path (str prefix path)]
     (go
       (try
         (let [response (<? (s3-request {:method "GET"
@@ -245,7 +245,7 @@
   [client path data]
   (let [{:keys [credentials bucket region prefix]} client
         ch (async/promise-chan)
-        full-path (str prefix "/" path)]
+        full-path (str prefix path)]
     (async/pipe (s3-request {:method "PUT"
                              :bucket bucket
                              :region region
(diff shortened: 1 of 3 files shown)


src/fluree/db/storage/s3.clj
```diff
@@ -325,13 +323,13 @@
     ch))
 
 (defn s3-address
-  [identifier s3-bucket s3-prefix path]
-  (storage/build-fluree-address identifier method-name path [s3-bucket s3-prefix]))
+  [path]
```
zonotope: The identifier allows users to use two different s3 storage instances on the same connection. Without the identifier, there would be no way to differentiate them.

bplatz: If we support the identifier I think it should be after, e.g. fluree:s3,  not before as it was. But I am not sure it works how you envision. The storage is multi-purpose for different components independent of this, so this would only apply within a single component (e.g. ledger storage for the same ledger). I don't see how this would be used (or how the system supported it), but maybe I misunderstand it.

I do understand having unique storage identifiers to target e.g. nameservice storage vs ledger storage, but not sure how these would be used or differentiate within a single, e.g. nameservice storage. Unless they are doing that I don't think they do anything where their names should be embedded in IRIs.

zonotope: Connections have both index and commit storage attached to them. These are not storage records directly, but [`Catalog`s, which are groups of storage records](https://github.com/fluree/db/blob/main/src/fluree/db/storage.cljc#L210-L226). A catalog can contain any combination of storage records, including multiple records of the same type. A user could, for example, store indexes in two different s3 buckets by including both buckets in its index catalog. 

The catalog uses the storage _location_, which includes the address identifier if present, to route requests to read/write to storage to the right storage record.

The catalog could also include remote storage locations managed by another server. The identifier is how the catalog knows that it should look up a file address from it's own file storage vs the file storage managed by the other server. The identifier is included in the address, and each server is configured to tell it where to find data based on the identifier and storage type. 

Remote systems just look like any other storage to each server, so there could be fluree:s3 , or fluree:file, or whatever addresses that a server needs to look up and multiple "s3" or "file" storages it has access to, one on the local server, and another in the remote system. With an identifier in the address, each server knows from which storage to look for the data

Look at the [server remote system test](https://github.com/fluree/server/blob/main/test/fluree/server/integration/remote_system_test.clj) for an example of this. Each server has 3 "memory" storages: one that it manages and two managed by other servers that it can read from.

bplatz: I'll add it back but still unsure how it is used, I think a discussion will help me understand. Your example is for a remote system, but a remote system has to have its data served by the remote ledger (we have an API for this to retrieve index segments, etc.) This is by necessity because data must be filtered per policy enforcement. So I think this would only apply to a local ledger, and that I don't understand how one would use it yet.

zonotope: Let's take the example from the remote system test. There are three systems, one managing author data, another book, and another movie. Let's say they all manage their index data in s3 buckets. Without the identifier, the addresses of their index files would all look like `fluree:s3:...`. 

When the book system processes a query that involves authors and movies, it needs to fetch data from the remote systems. The book system doesn't use that api you mentioned to fetch index data directly. There is an [implementation of the `JsonArchive` storage protocol for remote systems](https://github.com/fluree/db/blob/22fb2334a8b12ea3f814c893857f29f31f52e686/src/fluree/db/remote_system.cljc#L128-L130) that wraps that api and allows us to treat remote systems like read only storage, and allows the query engine to process queries involving data from remote systems normally without any concern for where the data comes from since it only uses the `JsonArchive` protocol.

That means the author system has storage records corresponding to the book and movie systems, and both of those records are in its index catalog. If the author system processes a query involving the book system, then it will come across addresses for index files from the book system, and it will come across addresses for index files from its own local s3 bucket. If neither system uses storage identifiers, then these addresses would just look like `fluree:s3:...`, and there's no way for the author system to tell which storage record to use to look up the data.

If all the systems use storage identifiers, then the addresses for the author data look like `fluree:s3:authors:...` and book looks like `flure:s3:books:...`. When the query engine needs to load data from one of those addresses, it uses the same `JsonArchive` api, but [implemented for the Catalog](https://github.com/fluree/db/blob/22fb2334a8b12ea3f814c893857f29f31f52e686/src/fluree/db/storage.cljc#L211-L215). The Catalog implementation uses the [`locate-address` function](https://github.com/fluree/db/blob/22fb2334a8b12ea3f814c893857f29f31f52e686/src/fluree/db/storage.cljc#L197-L201) to figure out which storage record to use to load the data. That function returns the `location` of an address, which is everything in front of the last ':', so it includes the identifier if it's present. This is how the catalog knows to fetch the book data from the book system, but the author data from it's local storage record.

The same process is how a single system can have two different s3 storage records pointing at two different buckets where it stores indexes. It uses identifiers for at least one of the bucket, and then it's index catalog knows which storage record to use to look up index data based on the address alone.

bplatz: OK, support is back and ready for approval if ok.

zonotope (approved):
🫘

PR #1080, opened 2025-08-01
Optimize subject joins

zonotope:
This patch adds an optimization for queries containing node maps with variable subject ids and more than two properties with variable objects. Instead of resolving matching flakes serially for each property, we resolve the matching flakes  in parallel and rely on the underlying index ordering to perform an inner join on the flake channels for each property. 

We also had to reintroduce the `psot` index to make this possible so we can rely on it's sorting order to group flakes for each property by subject and join each subject group in-line as we exaust each property channel.


src/fluree/db/query/fql/parse.cljc
```diff
@@ -493,16 +493,32 @@
...
-                  (get const/iri-id)
-                  (parse-subject context))
+  (let [id    (get m const/iri-id)
+        s-mch (parse-subject id context)
         attrs (dissoc m const/iri-id)]
     (if (empty? attrs)
       [(where/->pattern :id s-mch)]
       (let [statements (parse-statements s-mch attrs vars context)]
-        (sort optimize/compare-triples statements)))))
+        (if (simple-property-join? id attrs)
```
dpetran, line 519: Since this optimization only applies to node maps with multiple properties, do we plan to do a pre-parse step of merging node patterns with the same subject value into one map? Right now our SPARQL parser will never produce node maps that satisfy this requirement.

src/fluree/db/flake/match.cljc
```diff
@@ -200,6 +201,50 @@
...
+(defn match-properties
+  [db tracker solution triples error-ch]
+  (let [triples* (->> triples
+                      (map (fn [triple]
+                             (where/assign-matched-values triple solution)))
+                      (map (partial where/compute-sids db)))]
+    (if (every? some? triples*)
+      (let [property-ranges (->> triples*
+                                 (map (fn [triple]
+                                        (where/resolve-flake-range db tracker error-ch triple :psot)))
```
dpetran, line 239: Just checking to make sure I understand - this is the only circumstance where we will use the `:psot` index, right? At least, in the current implementation.

zonotope: Yes

src/fluree/db/query/fql/parse.cljc
zonotope: Yes, this is one component of many for query optimization. We need to build on the previous pr related to sorting triples by the fewest unspecified variables to also take into account variables that have been resolved in previous triples in the query. Then we can merge contiguous triples from the sorted list with the same subject at that point.

Your point about the SPARQL parser is valid though, and it probably bumps up the priority of including that merging in the optimization.

PR #1083, opened 2025-08-06
Refactor API to remove ledger as end-user concern

bplatz:
## Summary

This PR is open to debate and discussion - please take time to react to and digest these changes.

This refactors the Fluree API to remove 'ledger' as an end-user concern and makes the API more consistent by adopting a connection-centric approach throughout.

## Key Changes

- **`create` and `load` now return database objects**: Previously returned ledger objects, now return database objects directly
- **`db` function signature changed**: Now requires `(conn, ledger-id)` instead of accepting a ledger object
- **`status` function signature changed**: Now requires `(conn, ledger-id)` instead of accepting a ledger object  
- **`history` function signature changed**: Now requires `(conn, ledger-id, query)` instead of accepting a ledger object
- **`credential-history` function signature changed**: Now requires `(conn, ledger-id, cred-query)` instead of accepting a ledger object

## Breaking Changes

Initially, I intended to make these changes backward compatible. However, with `fluree/create` and `fluree/load` now returning database objects instead of ledger objects, maintaining backward compatibility became less feasible.

## Impact

- **fluree/server** will need to be updated to work with these API changes
- Most of Fluree's usage is not directly through the API, therefore it is unlikely to present issues to many end-users, if any
- I believe this is an important change to make now, ahead of a formal 3.0 release

## Test Updates

All tests have been updated to work with the new API signatures, including:
- Clojure tests
- ClojureScript tests  
- JavaScript SDK tests

The changes ensure a more consistent and intuitive API surface for users working with Fluree databases.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,6 +116,20 @@ jobs:
           clj-kondo: 'latest'
       - run: make clj-kondo-lint-ci
 
+  eastwood-lint:
+    name: eastwood lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '21'
+      - uses: DeLaGuardo/setup-clojure@12.5
+        with:
+          cli: 'latest'
+      - run: make eastwood
+
   notifications:
     name: send notifications
     if: always()
--- a/Makefile
+++ b/Makefile
@@ -149,7 +149,7 @@ test: cljtest cljstest nodejs-test browser-test ## Run all tests
 eastwood: ## Run Eastwood linter
 	clojure -M:dev:cljtest:eastwood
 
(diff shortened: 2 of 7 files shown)


src/fluree/db/api.cljc
```diff
@@ -234,9 +235,10 @@
   ([conn ledger-alias opts]
    (validate-connection conn)
    (promise-wrap
-    (do
+    (go-try
       (log/info "Creating ledger" ledger-alias)
-      (connection/create-ledger conn ledger-alias opts)))))
+      (let [ledger (<? (connection/create-ledger conn ledger-alias opts))]
```
zonotope, line 240: Definitely doesn't have to happen in this pr, and I might also go either way on this, but I think it might be a good idea to remove the concept of `ledger` as a first class object within internal apis as well. We could just pass around a conn which has cached ledgers in it's internal state, but we wouldn't need to deal with them directly within our internal apis.

There could be negative implications of this idea that we should think through, but it could also simplify our internal apis.

zonotope (approved):
🫓

PR #1084, opened 2025-08-06
Re-enable Eastwood linter with namespace exclusions

bplatz:
## Summary

This PR re-enables Eastwood in CI/CD by excluding incompatible namespaces rather than removing the linter entirely.

## Details

Clojure 1.12 has compatibility issues with Eastwood, but only for a few specific namespaces. This PR excludes those problematic namespaces while keeping Eastwood active for the rest of the codebase, preserving the linting benefits for the majority of our code.

## Changes

- Excluded fluree.db.nameservice.ipns from Eastwood linting (joins already excluded fluree.db.storage.ipfs and fluree.db.storage.file)
- Re-added eastwood target to make ci
- Added eastwood-lint job to GitHub Actions workflow

## Testing

Verified that make eastwood runs successfully with 0 warnings and 0 exceptions.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -123,7 +123,8 @@
                  :exclude-linters [:implicit-dependencies]
                  ;; Exclude IPFS namespace due to Clojure 1.12.1 compatibility issues
                  :exclude-namespaces [fluree.db.storage.ipfs
-                                      fluree.db.storage.file]}]}
+                                      fluree.db.storage.file
+                                      fluree.db.nameservice.ipns]}]}
 
   :ancient
   {:extra-deps {com.github.liquidz/antq {:mvn/version "RELEASE"}}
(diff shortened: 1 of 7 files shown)


deps.edn
```diff
@@ -123,7 +123,8 @@
                  :exclude-linters [:implicit-dependencies]
                  ;; Exclude IPFS namespace due to Clojure 1.12.1 compatibility issues
                  :exclude-namespaces [fluree.db.storage.ipfs
-                                      fluree.db.storage.file]}]}
+                                      fluree.db.storage.file
```
zonotope, line 126: Is there a github issue or similar tracking the eastwood 1.12 bug? If there is, I think it would be useful to put it's url here in a comment so we can check the status later to know when we can re-enable it.

bplatz: Issue that discussed problem and also its reliance on tools.analyzer: https://github.com/jonase/eastwood/issues/457

PR #1086, opened 2025-08-11
Fix/optional vars

dpetran:
We had a user with a query utilizing several nested `:optional` patterns that was returning several orders of magnitude more results than it ought. While fixing it I found several issues.

The first, if a triple pattern matches on a literal (non-iri) value in the object position and then uses that variable in the subject (or predicate) position, that value will not constrain the results at all:
```
[?s1 ?p1 ?o1]  ;; where ?o1 matches a literal non-iri value
[?o1 ?p2 ?o2] ;; ?o1 is treated as a wildcard and matches everything, resulting in a full index scan
```
The fix for this is to not conduct an index scan at all if either the subject or the predicate pattern component are bound to a literal (non-iri) value, since a non-SID cannot be compared with a SID at all and cannot match anything.

The next problem was nested optional patterns.
```
[?s1 "@type" {"@id" "ex:foo"}]
[?s1 ?p1 ?o1]  ;; generate bindings for ?s1 ?p1 ?o1
[:optional 
  [?o1 ?p2 ?o2] ;; returns no results because ?o1 is literal value, no new bindings
  [:optional
     [?o2 ?p3 ?o3]  ;; ?o2 is "fresh", so this turns into a full index scan
     [:optional
        [?o3 ?p4 ?p5]]]]
```
According to the spec, in an `:optional` query pattern, if the optional part does not match, it creates no bindings but does not eliminate the solution. But we need to track that an optional clause with no resulting solutions actually produces a solution with all the vars in the `:optional` pattern marked as "not fresh" in some sense. If one of these "not fresh" variables is referenced in later patterns we can skip the index scan for those patterns as they shouldn't match anything. I implemented this by adding optional vars to the solution and marking the resulting matches as `::where/optional`, which I use to short circuit index scanning in `resolve-flake-range`.

The third problem had to do with alternating duplicate values in projecting `construct` results, which was a straightforward fix.

### DIFF
--- a/src/fluree/db/query/exec/where.cljc
+++ b/src/fluree/db/query/exec/where.cljc
@@ -465,9 +465,11 @@
 
 (defn resolve-flake-range
   [{:keys [t] :as db} tracker error-ch [s-mch p-mch o-mch]]
-  (let [s    (get-sid s-mch db)
+  (let [s    (or (get-sid s-mch db)
+                 (get-value s-mch))
         s-fn (::fn s-mch)
-        p    (get-sid p-mch db)
+        p    (or (get-sid p-mch db)
+                 (get-value p-mch))
         p-fn (::fn p-mch)
         o    (or (get-value o-mch)
                  (get-sid o-mch db))
@@ -505,7 +507,12 @@
                      :start-flake start-flake
                      :end-flake   end-flake
                      :flake-xf    flake-xf}]
-    (query-range/resolve-flake-slices db tracker idx error-ch opts)))
+    (if (and (or (iri/sid? s) (nil? s))
+             (or (iri/sid? p) (nil? p)))
+      ;; check for matching flake
+      (query-range/resolve-flake-slices db tracker idx error-ch opts)
+      ;; cannot return any flakes
+      (async/onto-chan (async/chan) []))))
 
 (defn compute-sid
   [s-mch db]
--- a/src/fluree/db/query/exec/where.cljc
+++ b/src/fluree/db/query/exec/where.cljc
@@ -473,11 +473,13 @@
     [o o-fn]))
 
 (defn comparable-iri?
-  "When matching against"
+  "When matching against an all-iri index (s or p - SIDs), the only values that can be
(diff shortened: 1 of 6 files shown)


src/fluree/db/query/exec/where.cljc
```diff
@@ -463,49 +472,66 @@
...
-                         (apply comp))
-        opts        {:idx         idx
-                     :to-t        t
-                     :start-flake start-flake
-                     :end-flake   end-flake
-                     :flake-xf    flake-xf}]
-    (query-range/resolve-flake-slices db tracker idx error-ch opts)))
+  [{:keys [t] :as db} tracker error-ch [s-mch p-mch o-mch :as triple-pattern]]
+  (let [s    (or (get-sid s-mch db)
+                 (get-value s-mch))
```
zonotope, line 489: anything in the `s` or `p` position should only be sids, so we should never consider `get-value` for `s` or `p`. i think we should instead short circuit if either `s` or `p` returns `nil` for `get-sid` and non `nil` for `get-value`.

dpetran: That's basically what I'm checking here, I just use `get-value` here so I can deal with all the short-circuiting logic in one place, in the `if` below..

zonotope: I think I would prefer line 488 be just `s (get-sid s-mch db)` and lin 495 be `(and (nil? s) (some? get-value s-mch))` to make it clear that there should only be an sid in the `s` position or we short circuit.

I think this is the less confusing thing to my eyes when scanning the file, but if you disagree strongly, then feel free to leave this as is.

zonotope (approved):

src/fluree/db/query/exec/where.cljc
dpetran: My thought is that it is perfectly valid to have a `get-value` match component for `s` or `p` - that is explicitly ok with the query language semantics and I don't want to give any impression that a non-iri value there is anomalous in any way. A non-iri value will generate no new matches, though, so that's why we check for comparability below.

PR #1087, opened 2025-08-11
Refactor: Consistent branch naming with ledger:branch format

bplatz:
### Summary
Standardizes branch naming to `ledger:branch`, removes branch segments from content‑addressed paths, and scopes nameservice records by branch. Consolidates branch handling and reserves `@` for time travel (PR #1088).

### Motivation
- Support upcoming branching features with a consistent, parseable `ledger:branch` format.
- Keep ledger names URL‑safe so ledgers can be used as full IRIs when hosted.
- Reserve `@` exclusively for time travel to align with forthcoming query semantics.
- Streamline branch handling across db, ledger objects for more consistency

### Key Changes
- **Branch format**
  - `ledger-name:branch-name` (e.g., `my-ledger:main`)
  - `:` reserved from ledger names; `@` reserved for time travel (e.g., `my-ledger:main@t:42`)
  - Branch names disallow `/` to avoid unintended subdirectories
- **Paths**
  - Content‑addressed artifacts (commit/txn/index) no longer include branch
    - Example: `/my-ledger/commit/[hash].json` (deduped across branches)
  - Nameservice files are per‑branch subfiles
    - Example: `/ns@v1/my-ledger/main.json` (supports ledger names with `/`)
- **API and internals**
  - `fluree.db.api/create` rejects `:` in ledger names (`:db/invalid-ledger-name`)
  - New `fluree.db.util.ledger` with `ledger-base-name`, `ledger-branch`, `ledger-parts`
  - Removed redundant branch tracking from `Ledger`; fixed circular deps
  - Commit storage now uses ledger-name (not alias)

### Impact
- Ledger names cannot contain `:`
- Branch names cannot contain `/`
- External tools relying on branch-in-path or old nameservice locations must update
- No data re-encoding; hashes remain stable and deduplicated

### Testing
- New/updated tests for name validation, nameservice layout, and path conventions
- All tests passing; no lint warnings

### Related Work
- Builds on time travel proposal in [PR #1088](https://github.com/fluree/db/pull/1088), which adds `@`-based query time travel:
  - Supports `@t:<tx>`, `@iso:<timestamp>`, `@sha:<commit>`, works with `ledger:branch`
  - `@` takes precedence over legacy `?t=...`; supports git‑style SHA prefixes

### DIFF
--- a/src/fluree/db/connection.cljc
+++ b/src/fluree/db/connection.cljc
@@ -503,4 +503,4 @@
         timeout-ch (ex-info "Indexing wait timeout, but assume indexing is proceeding in the background."
                             {:status 408
                              :error :db/timeout
-                             :timeout timeout})))) )
+                             :timeout timeout})))))


src/fluree/db/api.cljc
```diff
@@ -248,6 +249,13 @@
   ([conn ledger-alias] (create conn ledger-alias nil))
   ([conn ledger-alias opts]
    (validate-connection conn)
+   ;; Disallow branch specification in ledger name during creation
+   (when (str/includes? ledger-alias ":")
```
dpetran: I think pulling this out to `validate-ledger-name` would be nice so we don't have to duplicate this logic.

src/fluree/db/api/transact.cljc
```diff
@@ -130,6 +131,13 @@
                                                                            ; are no policies
                                                                            ; to check.
            ledger-opts (-> parsed-txn :opts syntax/coerce-ledger-opts)
+           ;; Disallow branch specification in ledger name during creation
+           _           (when (str/includes? ledger-id ":")
```
dpetran: We should use a common `validate-ledger-name` fn here.

src/fluree/db/util/ledger.cljc
```diff
@@ -0,0 +1,25 @@
+(ns fluree.db.util.ledger
+  "Utility functions for working with ledger names and branches."
+  (:require [clojure.string :as str]))
+
+(defn ledger-base-name
```
dpetran, line 5: Just want to check if this is correct:

`<ledger-name>:<branch>`
`^---ledger-alias-----^`
(branch is optional, but is always present once normalized)

`fluree:file://<ledger-alias>`
`^----ledger-address--------^`

src/fluree/db/connection.cljc
```diff
@@ -337,16 +323,12 @@
...
             expanded-commit (json-ld/expand commit)
-            ledger-alias    (commit->ledger-alias conn address expanded-commit)
-            ;; Determine branch using helpers in priority order
-            branch           (or (branch-from-commit expanded-commit)
-                                 (branch-from-ns ns-record)
-                                 (branch-from-alias ledger-alias))
-
-            {:keys [did branch indexing]} (parse-ledger-options conn {:branch branch})
-            ledger (ledger/instantiate ledger-alias address branch commit-catalog index-catalog
+            combined-alias  (commit->ledger-alias conn address expanded-commit)
```
dpetran: Why is this called the `combined-alias`, I don't see what we are combining? 

And why wouldn't the commit have the ledger alias in it, is that a backwards compatibility thing? (`commit->ledger-alias` falls back to the nameservices to return the alias if it's not in the commit, but I think we've always put the ledger alias in the commit.)

src/fluree/db/connection.cljc
```diff
@@ -458,7 +444,10 @@
     (try*
       (let [alias (if (fluree-address? alias)
                     (nameservice/address-path alias)
-                    alias)]
+                    ;; Normalize alias to include branch if not present
+                    (if (str/includes? alias ":")
```
dpetran: There is a function for this, `normalize-ledger-alias`.

src/fluree/db/connection.cljc
```diff
@@ -498,16 +487,15 @@
...
 
    Returns the indexed database object or throws an exception on failure/timeout."
   [conn ledger-alias opts]
   (go-try
-    (let [{:keys [branch timeout]
+    (let [{:keys [timeout]
            :or {timeout 300000}} opts
           ledger (<? (load-ledger-alias conn ledger-alias))
-          complete-ch (ledger/trigger-index! ledger branch)
+          complete-ch (ledger/trigger-index! ledger nil)
```
dpetran: There's a single-arg arity for `trigger-index`, so you shouldn't have to specify `nil` here.

src/fluree/db/flake/commit_data.cljc
```diff
@@ -223,12 +221,9 @@
 
 (defn blank-commit
   "Creates a skeleton blank commit map."
-  [alias branch publish-addresses init-time]
+  [alias publish-addresses init-time]
   (let [commit-json  (->json-ld {:alias  alias
                                  :v      0
-                                 :branch (if branch
```
dpetran, line 229: Since we're changing the commit structure we should update the `commit-version`. And I've just noticed that this function hardcodes a `v` of `0`, it should reference the `commit-version`.

src/fluree/db/flake/flake_db.cljc
```diff
@@ -546,21 +546,12 @@
                          (<? (index-storage/read-db-root index-catalog address))
                          (genesis-root-map ledger-alias))
            max-ns-code (-> root-map :namespace-codes iri/get-max-namespace-code)
-           ;; Ensure commit map reflects loaded index t when an index address exists
-           commit-map* (if (get-in commit-map [:index :address])
```
dpetran, line 550: Why is it that we don't need to update the index `t` here anymore?

src/fluree/db/ledger.cljc
```diff
@@ -129,15 +129,19 @@
 (defn instantiate
   "Creates a new ledger, optionally bootstraps it as permissioned or with default
   context."
-  [ledger-alias ledger-address branch commit-catalog index-catalog primary-publisher secondary-publishers
+  [combined-alias ledger-address commit-catalog index-catalog primary-publisher secondary-publishers
    indexing-opts did latest-commit]
-  (let [publishers (cons primary-publisher secondary-publishers)
-        branches {branch (branch/state-map ledger-alias branch commit-catalog index-catalog
+  (let [;; Parse ledger name and branch from combined alias
+        [_ branch] (if (str/includes? combined-alias ":")
```
dpetran: I would rather use `normalize-alias` to ensure consistent structure and then `util.ledger/address-parts` to destructure. As I see it the only way we wouldn't have a normalized alias already is in the case of legacy ledger aliases.

src/fluree/db/ledger.cljc
```diff
@@ -160,20 +164,24 @@
   context."
   [{:keys [alias primary-address publish-addresses commit-catalog index-catalog
            primary-publisher secondary-publishers]}
-   {:keys [did branch indexing] :as _opts}]
+   {:keys [did indexing] :as _opts}]
   (go-try
-    (let [ledger-alias*  (normalize-alias alias)
+    (let [normalized-alias  (normalize-alias alias)
+          ;; Add :main if no branch is specified
+          ledger-alias   (if (str/includes? normalized-alias ":")
```
dpetran: Isn't this redundant because we just called `normalize-alias`?

dpetran: Ah, I see, we have `connection/normalize-ledger-alias` and `ledger/normalize-alias`. `normalize-alias` just removes a leading `#` or `/` character, I think we should just disallow those constructions in `validate-ledger-name` - that's probably better manners towards the user than just modifying the name they've given to us.

Also, we _do_ call `connection/normalize-ledger-alias` inside `connection/create-ledger` before calling `ledger/create`, so we definitely do have a normalized `ledger-alias` already - this check is not necessary.

src/fluree/db/migrations/nameservice_v1.cljc
```diff
@@ -142,9 +142,10 @@
             (extract-commit-metadata old-data)
 
             ;; Create new minimal record using existing function
-            new-record (ns-storage/ns-record ledger-alias branch commit-address t-value index-address)
+            combined-alias (str ledger-alias ":" branch)
```
dpetran: Can we use the `connection/normalize-ledger-alias` function here to keep the implementation consistent?

dpetran: I'm also still not clear on what's being combined in the alias to warrant it being called the "combined-alias".

src/fluree/db/nameservice/storage.cljc
```diff
@@ -9,25 +9,22 @@
...
-  [store ledger-alias]
-  (-> store
-      storage/location
-      (storage/build-address ledger-alias)))
+  "Returns the local filename for a ledger's nameservice record.
+   Expects ledger-alias to be in format 'ledger:branch'.
+   Returns path like 'ns@v1/ledger-name/branch.json'."
+  [ledger-alias]
+  (let [[ledger-name branch] (str/split ledger-alias #":" 2)]
+    (str "ns@v1/" ledger-name "/" branch ".json")))
```
dpetran: I know this isn't new but `"ns@v1"` should be a constant.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -9,25 +9,22 @@
...
 (defn ns-record
-  "Generates nameservice metadata map for JSON storage using new minimal format"
-  [ledger-alias branch commit-address t index-address]
-  (let [branch (or branch "main")]
+  "Generates nameservice metadata map for JSON storage using new minimal format.
+   Expects ledger-alias to be in format 'ledger:branch'."
+  [ledger-alias commit-address t index-address]
+  (let [[alias branch] (str/split ledger-alias #":" 2)]
     (cond-> {"@context"     {"f" iri/f-ns}
-             "@id"          (str ledger-alias "@" branch)
```
dpetran, line 28: Is this a backwards incompatible change? Will we need another nameservice migration?

src/fluree/db/nameservice/storage.cljc
```diff
@@ -63,14 +55,16 @@
       (storage/delete store address)))
 
   (publishing-address [_ ledger-alias]
-    (go (publishing-address* store ledger-alias)))
+    ;; Just return the alias - lookup will handle branch extraction via local-filename
+    (go ledger-alias))
```
dpetran, line 60: I don't think the publishing address and the ledger alias are the same thing semantically, is this a compatible change?

src/fluree/db/storage.cljc
```diff
@@ -48,8 +48,14 @@
 (defn split-address
   "Splits `address` into the fully qualified storage method and local path."
   [address]
-  (let [i (str/last-index-of address ":")]
-    [(subs address 0 i) (subs address (inc i))]))
+  (let [i (str/index-of address "://")]
```
dpetran, line 57: It occurs to me that we should validate that branch names do not contain `//`, or probably anything besides alphanumeric characters.

src/fluree/db/transact.cljc
```diff
@@ -144,20 +145,22 @@
...
 (defn save-txn!
-  ([{:keys [commit-catalog] ledger-alias :alias :as _ledger} txn]
-   (save-txn! commit-catalog ledger-alias txn))
-  ([commit-catalog ledger-alias txn]
-   (let [path (str/join "/" [ledger-alias "txn"])]
+  ([{:keys [commit-catalog alias] :as _ledger} txn]
+   (let [ledger-name (util.ledger/ledger-base-name alias)]
+     (save-txn! commit-catalog ledger-name txn)))
+  ([commit-catalog ledger-name txn]
+   (let [path (str/join "/" [ledger-name "txn"])]
```
dpetran, line 152: We should also validate that ledger names don't end in a `/`, I imagine it could cause problems in some storage systems.

test/fluree/db/api/create_test.clj
```diff
@@ -0,0 +1,140 @@
...
+             #"Ledger name cannot contain ':' character"
+             (fluree/create conn ":branch"))
+            "Should reject name starting with colon")
+
+        (is (thrown-with-msg?
+             clojure.lang.ExceptionInfo
+             #"Ledger name cannot contain ':' character"
+             (fluree/create conn "ledger:"))
+            "Should reject name ending with colon"))
+
```
dpetran, line 108: Some additional ledger validation checks:

I think we should also disallow these, which are currently allowed:
```
"///"
"ledger/"
"#ledger"
"/ledger"
"@ledger"
"-._._.-"
"!@#$%^&*()"
"../../../Downloads/taxes.pdf"
"/var/data/postgres"
```

test/fluree/db/query/property_path_test.clj
```diff
@@ -4,11 +4,11 @@
 
 (deftest transitive-paths
   (let [conn   @(fluree/connect-memory)
-        db0 @(fluree/create conn "property/path")
-        db0    db0]
+        db0 @(fluree/create conn "property/path")]
     (testing "one+"
       (testing "no variables"
-        (let [db1 @(fluree/update db0 {"insert"
+        (let [db1 @(fluree/update db0 {"@context" {"ex" "http://example.org/"}
```
dpetran, line 10: Just curious, do these tests need an `@context`? `ex:a` is a perfectly valid IRI.

test/fluree/db/query/stable_hashes_test.clj
```diff
@@ -29,10 +29,10 @@
                       :schema/age   30}]})
           db1    @(fluree/commit! conn db0)]
       (testing "stable commit id"
-        (is (= "fluree:commit:sha256:bbdainpfs7v2pg2yj76uzpybdfdldvvt5idlbasisuhwlrxbiqhii"
+        (is (= "fluree:commit:sha256:bb7xyuvqhe2wgobfj7ylsaoyuh3nucqcichfblkkl6zfirha5pvlh"
                (get-in db1 [:commit :id]))))
       (testing "stable commit address"
-        (is (= "fluree:memory://bdainpfs7v2pg2yj76uzpybdfdldvvt5idlbasisuhwlrxbiqhii"
+        (is (= "fluree:memory://b7xyuvqhe2wgobfj7ylsaoyuh3nucqcichfblkkl6zfirha5pvlh"
```
dpetran: Just reaffirming that if commit shas change, the commit v needs to be updated.

src/fluree/db/util/ledger.cljc
bplatz: Yes, but ledger address I'm not sure is super useful here, but it does form a URI which I think will be useful. So a ledger address, if hosted on a web server, could also be http://my.comain/my/ledger:main

src/fluree/db/api.cljc
bplatz: Addressed in 24845d3668b8256da20cfe83b143cac41f6a1f05

src/fluree/db/api/transact.cljc
bplatz: addressed in 24845d3668b8256da20cfe83b143cac41f6a1f05

src/fluree/db/connection.cljc
bplatz: addressed in 5191a7bfad0be9ea93ed671cacfe21acf9705bf5

src/fluree/db/connection.cljc
bplatz: addressed in bd949d6cecbee623a3b2330abd123152902f191f

src/fluree/db/flake/commit_data.cljc
bplatz: addressed in 61f9b1fc68b5a97f1bfe9785342a058fc13f1c12

src/fluree/db/nameservice/storage.cljc
bplatz: Addressed in 06fb2ecc9a29c42cafc907689d7d3760ca23276c

src/fluree/db/storage.cljc
bplatz: addressed with other ledger name validations

src/fluree/db/transact.cljc
bplatz: addressed with other ledger name validations

test/fluree/db/api/create_test.clj
bplatz: addressed with other ledger name validations

test/fluree/db/query/stable_hashes_test.clj
bplatz: done

src/fluree/db/migrations/nameservice_v1.cljc
bplatz: Addressed in 897ae5444e3baf55f9b59df533a5039a842cbd03

src/fluree/db/ledger.cljc
bplatz: Addressed in 897ae5444e3baf55f9b59df533a5039a842cbd03

src/fluree/db/ledger.cljc
bplatz: addressed in 897ae5444e3baf55f9b59df533a5039a842cbd03

src/fluree/db/connection.cljc
bplatz: Addressed in 897ae5444e3baf55f9b59df533a5039a842cbd03

test/fluree/db/query/property_path_test.clj
bplatz: Since the intention with ex is that it is a 'normal' IRI I think it should.

I have mentioned before that if we see something like 'ex:a' that we should even consider throwing as it is almost certainly a user mistake... and this is probably one of my largest sources of unintended bugs when I query / transact and accidentally forgot the context in one of them and it looks like there is no data.

So I think if we ever use anything that looks like a prefix we should always include a context for good practice.

src/fluree/db/nameservice/storage.cljc
bplatz: This is likely only needed for IPNS now which we need to get back to. The nameservice is more capable now and relies on its internal record to extract an address, so the nameservice is now in charge of resolving this for you. This was not the case previously.

I think we should leave this here for now until we refactor IPFS/IPNS and understand the needs there as I suspect we will have the requirement (as the nameservice request will no longer come through a Fluree nameservice).

src/fluree/db/api.cljc
```diff
@@ -248,6 +250,8 @@
   ([conn ledger-alias] (create conn ledger-alias nil))
   ([conn ledger-alias opts]
    (validate-connection conn)
+   ;; Validate ledger name
```
zonotope: this comment is unnecessary

src/fluree/db/api/transact.cljc
```diff
@@ -130,6 +131,8 @@
                                                                            ; are no policies
                                                                            ; to check.
            ledger-opts (-> parsed-txn :opts syntax/coerce-ledger-opts)
+           ;; Validate ledger name
```
zonotope: unnecessary comment

src/fluree/db/flake/index/storage.cljc
```diff
@@ -38,8 +39,10 @@
 
 (defn write-index-file
   [storage ledger-alias index-type serialized-data]
-  (let [index-name (name index-type)
-        path       (str/join "/" [ledger-alias "index" index-name])]
+  (let [;; Extract base ledger name from alias for storage path
```
zonotope: unnecessary comment

src/fluree/db/transact.cljc
```diff
@@ -144,20 +145,22 @@
...
      (storage/content-write-json commit-catalog path txn))))
 
 ;; TODO - as implemented the db handles 'staged' data as per below (annotation, raw txn)
 ;; TODO - however this is really a concern of "commit", not staging and I don't think the db should be handling any of it
 (defn write-transaction!
-  [ledger staged]
+  [ledger db-alias staged]
```
zonotope: Is a `db-alias` conceptually different from a ledger alias? If not, I think this should be `ledger-alias` here.

zonotope (approved):
🍶

src/fluree/db/flake/flake_db.cljc
bplatz: The old code was a band‑aid to keep the commit’s embedded index metadata in sync with the actual index root. We don’t need that mutation anymore because we now treat the index root as the source of truth and reconcile index freshness via the normal index-update flow.

src/fluree/db/nameservice/storage.cljc
dpetran: My issue is that a `ledger-address` and a `ledger-alias` have different structures and none of the alias parsing functions will correctly handle an address.

src/fluree/db/nameservice/storage.cljc
bplatz: You are correct, addressed in 684fc035f4a9302679afbebace02a309a467ee01

src/fluree/db/transact.cljc
bplatz: Addressed in 7d9819fce979c70e2d2a7102a6e597c13ca2f6f7

src/fluree/db/flake/index/storage.cljc
bplatz: Addressed in 7d9819fce979c70e2d2a7102a6e597c13ca2f6f7

src/fluree/db/api/transact.cljc
bplatz: Addressed in 7d9819fce979c70e2d2a7102a6e597c13ca2f6f7

src/fluree/db/migrations/nameservice_v1.cljc
```diff
@@ -172,16 +179,188 @@
...
+              (let [res (<? (migrate-v1-file file-store info))]
+                (recur (rest files) (conj results res)))
+              (do
+                (log/info "ns@v1 ->" const/ns-version "migration completed. Migrated"
+                          (count (filter :migrated results))
+                          "files, skipped" (count (filter :skipped results)) "already updated files")
+                results))))
+        []))))
+
 (defn run-migration-if-needed
```
dpetran: If a fluree db is configured to use a non-file store backend (s3 or ipfs), will _not_ running this migration have the system be in a usable state? Or are those configurations not supported until we add migrations for them?

src/fluree/db/migrations/nameservice_v1.cljc
```diff
@@ -8,12 +8,12 @@
   - Detects old nameservice files stored at root level (e.g., ledger-name.json)
   - Extracts essential metadata from full commit JSON-LD records
   - Creates new minimal records with only: commit address, t value, index address
-  - Stores them in the new ns@v1/ directory with branch-aware naming (ledger-name@branch.json)
+  - Stores them in the new ns@v1/ directory with branch-aware naming (ledger-name/branch.json)
```
dpetran: We should also update the ns docstring, which claims migration to ns@v1 format. And the namespace itself, maybe we should just call it `fluree.db.migrations.nameservice`.

src/fluree/db/migrations/nameservice_v1.cljc
bplatz: IPFS isn't working in this version, and S3 was just added and I'm quite certain I'm the only one using it.

src/fluree/db/migrations/nameservice_v1.cljc
bplatz: Addressed in 44c7c3c7f9d7686ff3d77da5cdd11878ae1cf2ae

src/fluree/db/nameservice/storage.cljc
bplatz: They can have a different structure. `publishing-address` effectively means “the identifier this nameservice wants persisted in commit ns and later used for its own lookup.” For Storage this is the alias; for IPNS it’s the IPNS IRI.

Callers only invoke publishing-address with a relative alias. connection/publishing-addresses explicitly guards on relative aliases, and primary-address normalizes an alias before calling the publisher.

The returned value is not re-parsed as an alias; it’s written into the commit’s ns field and passed back to the same nameservice’s lookup. StorageNameService/lookup expects an alias and resolves the on-disk path via local-filename.

IPNS still returns a full fluree:ipns://... IRI; its lookup accepts that IRI and resolves it via IPNS. So heterogeneous return types are already handled by each nameservice.

I adjusted the docstring to better reflect the current implementation here: 6b242ac9d87e23259ab7f3da165cef28b04f5d74

PR #1088, opened 2025-08-12
Add time travel capability using @ syntax in query URIs

bplatz:
## Time travel queries via @ syntax in FQL graph URIs

Sits on top of: https://github.com/fluree/db/pull/1087 (branch name consistency)
Prerequisite of: https://github.com/fluree/db/pull/1096 (branch implementation)
Prerequisite of: https://github.com/fluree/db/pull/1095 (brach merge, rebase, reset)

### Summary
Adds time travel to database queries using `@` in `from` and `from-named` to select historical states by transaction number, ISO-8601 timestamp, or commit SHA.

### Motivation
- Clear, URL/IRI-friendly syntax for time travel queries with a concrete resource name (not requiring query params)
- Consistent with reserved `@` semantics and upcoming branching support.

### Supported formats
- `@t:<tx>`: transaction number
  - Example: `my-ledger@t:42`
- `@iso:<timestamp>`: ISO-8601 wall-clock time
  - Example: `my-ledger@iso:2025-07-01T00:00:00Z`
- `@sha:<commit>`: commit SHA (base32 SHA-256 with leading `b`)
  - Examples: `my-ledger@sha:bb6rql...` (full), `my-ledger@sha:bb6rql` (prefix, ≥6 chars)

Works with branches:
- `my-ledger:main@t:42`

Works with federated datasets:
- `from-named: ["g1@t:10", "g2"]`

### Behavior
- SHA handling:
  - Accepts full base32 SHA-256 with leading `b` (51–52 chars).
  - Partial prefixes allowed (minimum 6 chars).
  - Ambiguous prefixes return 400 `:db/ambiguous-commit-sha`.
  - Not found returns 400 `:db/invalid-commit-sha`.
- Missing values (e.g., `@t:`, `@iso:`, `@sha:`) return 400 `:db/invalid-time-travel`.
- ISO lookup returns the most recent commit at or before the provided time; if none exists, returns 400 “no data as of …”.
- Federated queries:
  - Top-level `t` must be ISO-8601 wall-clock time.
  - Per-graph time travel via `@` on each alias.

### Implementation details
- Time travel protocol extended:
  - `sha->t` for SHA-based lookups.
- Parsing:
  - `extract-query-string-t` in `fluree.db.query.api` parses `@` syntax on `from`/`from-named`.
- SHA normalization:
  - Supports only `fluree:commit:sha256:b` prefix form (when present) or a value starting with `b`.
  - Full SHAs use a direct query; prefixes use an index range scan.
  - Enforces minimum 6-char prefix.
- Creation-time validation:
  - Ledger names reject `:` and `@` to avoid ambiguity and reserve `@` for time travel.

### Testing
- Coverage for all three formats: `@t`, `@iso`, `@sha`.
- Controlled ISO tests to ensure stable comparisons.
- SHA tests include full hashes and prefix matching.
- Negative cases: missing spec values, invalid formats, not found, and ambiguous prefixes.

### Backward compatibility
- Existing queries continue to work. `@`-based time travel is additive; legacy `?t=` remains supported but is superseded when `@` is present.


src/fluree/db/flake/flake_db.cljc
```diff
@@ -379,6 +379,110 @@
...
+              (let [found-commit-sid (first commit-sids)
+                    found-commit-id (iri/decode-sid db found-commit-sid)
+                    ;; Use values to directly query for this specific commit's t value
+                    prefix-query {:select ["?t"]
+                                  :values ["?commit" [{"@type" "@id" "@value" found-commit-id}]]
+                                  :where [{"@id" "?commit"
+                                           const/iri-data "?data"}
+                                          {"@id" "?data"
+                                           const/iri-fluree-t "?t"}]}
+                    prefix-result (<? (fql/query (policy/root db) nil prefix-query))]
```
dpetran: Shouldn't we have access to the `t` value on the flake with commit-sid? I don't think we need to do a query here.

src/fluree/db/flake/flake_db.cljc
```diff
@@ -379,6 +379,110 @@
...
+                                       const/iri-fluree-t "?t"}]}
+                direct-result (<? (fql/query (policy/root db) nil direct-query))]
+            (if (seq direct-result)
+              (do
+                (log/debug "sha->t found exact match for" commit-id)
+                (ffirst direct-result))
+              (throw (ex-info (str "No commit found with SHA: " sha-normalized)
+                              {:status 400 :error :db/invalid-commit-sha :sha sha}))))
+
+          ;; Partial SHA - use prefix matching (< 52 chars)
```
dpetran: I feel pretty strongly that we shouldn't support partial SHAs. At least, not until someone begs us for it. It's slower, much more complicated, and in the only place a user will ever be able to find the SHA (a commit), it will not be truncated, so they'd have to manually truncate it anyways. 

I know that partial SHAs in git and in clojure deps.edn are a pain because they are potentially ambiguous and collisions do happen. And in that case we are just adding an error that users will have to deal with, and they'll deal with it by getting the non-truncated SHA from the same place they went to get the truncated one. So why not just skip it and save them the trouble?

test/fluree/db/query/time_travel_test.clj
```diff
@@ -1,240 +1,364 @@
...
+          ;; Get commit SHA for t1 to test SHA-based time travel
+          commit-query {"select" ["?commit"]
+                        "where" [{"@id" "?commit"
+                                  "@type" "fluree:Commit"
+                                  "https://ns.flur.ee/ledger#v" t1}]}
+          commit-result @(fluree/query db-t3 commit-query {})
+          commit-id (ffirst commit-result)
+
+          ;; Extract just the hash part after "fluree:commit:sha256:b"
+          commit-sha (when (and commit-id (string/starts-with? commit-id "fluree:commit:sha256:b"))
```
dpetran: This is a deterministic case so I don't think we need the `when` conditional - it will always be there.

test/fluree/db/query/time_travel_test.clj
```diff
@@ -1,240 +1,364 @@
...
+              query-now {"select" ["?name"]
+                         "where" [{"@id" "?s" "name" "?name"}]
+                         "from" [(str ledger-name "@iso:" iso-now)]
+                         "orderBy" ["?name"]}
+              result-now @(fluree/query-connection conn query-now {})]
+          (is (= [["Alice"] ["Bob"] ["Carol"]] result-now)
+              "At current ISO time: Should see all three people")))
+
+      ;; Test SHA-based time travel
+      (when commit-sha
```
dpetran: Are these tests flaky? It makes me nervous to only run them when some condition is met, I worry we might miss test failures if we have regression.

src/fluree/db/query/api.cljc
```diff
@@ -172,30 +172,81 @@
...
-    (if query-str
-      [alias (-> query-str
-                 query-str->map
-                 (get "t")
-                 parse-t-val)]
-      [alias nil])))
+  (cond
+    ;; Check for @ syntax first (takes precedence)
+    (str/includes? alias "@")
+    (let [at-idx (str/index-of alias "@")
```
dpetran: I'd like to see this logic live with the other ledger alias parsing functions in `util.ledger`, and then utilize `ledger-parts` to pick out the `t` data (maybe it would be better for it to return a map instead of a vector).

dpetran (commented):
Could you expound a bit on why we need the new syntax instead of query params? I think we can still support `sha` `t` and `iso` via that syntax, and having two ways to do it makes users have to decide which to use, and I'm not clear that there's a technical benefit of one over the other.

src/fluree/db/flake/flake_db.cljc
bplatz: > I feel pretty strongly that we shouldn't support partial SHAs. At least, not until someone begs us for it. It's slower, much more complicated, and in the only place a user will ever be able to find the SHA (a commit), it will not be truncated, so they'd have to manually truncate it anyways.

We plan to use them in the admin/nexus UI for the same reason Git does, so we do have a plan for them already... and even with 10s of millions of commits even 64 bits would be nearly impossible to have a collision. Yes < 256 bits has higher risk of collision for every piece of information in the entire world, but that is not the scope here. So I just thing the collision possibility here is close of enough to zero that it is fine for this use.

dpetran: I suppose that's fair, if it's for a user interface.

src/fluree/db/flake/flake_db.cljc
bplatz: Refactored a bit in 5c3a314ee2602991537ec37def928c0e9640ca51

src/fluree/db/nameservice/storage.cljc
```diff
@@ -1,33 +1,34 @@
...
-  [store ledger-alias]
-  (-> store
-      storage/location
-      (storage/build-address ledger-alias)))
+  "Returns the local filename for a ledger's nameservice record.
+   Expects ledger-alias to be in format 'ledger:branch'.
+   Returns path like 'ns@v2/ledger-name/branch.json'."
+  [ledger-alias]
+  (let [[ledger-name branch] (util.ledger/ledger-parts ledger-alias)
+        branch (or branch "main")]
```
zonotope: tiny nitpick, but i think `"main"` should be a constant somewhere that we can refer to everywhere we use it explicitly.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -38,21 +39,16 @@
   nameservice/Publisher
   (publish [_ data]
     (let [;; Extract data from compact JSON-LD format (both genesis and regular commits now use this)
-          ledger-alias   (get data "alias")
-          branch         (or (get data "branch")
-                             (when (and (string? ledger-alias)
-                                        (str/includes? ledger-alias "@"))
-                               (subs ledger-alias (inc (str/last-index-of ledger-alias "@"))))
-                             "main")
+          ledger-alias   (get data "alias")  ;; Already includes @branch
```
zonotope: another nitpick, but i think this comment should end with `:branch` instead of `@branch`.

src/fluree/db/flake/flake_db.cljc
```diff
@@ -379,6 +379,75 @@
   (latest-t [_]
     t)
 
+  (sha->t [db sha]
+    (go-try
+      (log/debug "sha->t looking up commit SHA:" sha)
+      ;; Normalize the input - use only 'fluree:commit:sha256:b' prefix when present,
+      ;; otherwise ensure the value starts with 'b'
+      (let [prefix-b "fluree:commit:sha256:b"
```
zonotope: this should be an explicit constant defined with `def` outside of this method definition instead of this `let` here.

src/fluree/db/flake/flake_db.cljc
```diff
@@ -379,6 +379,75 @@
...
+                           :sha sha :normalized sha-normalized :length sha-length}))
+
+          ;; Too short to be a useful/efficient prefix (minimum 6)
+          (< sha-length 6)
+          (throw (ex-info "SHA prefix must be at least 6 characters"
+                          {:status 400 :error :db/invalid-commit-sha :min 6}))
+
+          :else
+          (let [;; sha-normalized already has 'b' prefix from normalization
+                commit-id-prefix (str "fluree:commit:sha256:" sha-normalized)
```
zonotope: the `fluree:commit:sha256:` should also be an explicit constant.

zonotope (approved):
🏷️

PR #1089, opened 2025-08-13
Use valid expanded json ld

zonotope:
This patch updates our json-ld dependency to the version from fluree/json-ld#44 which changes the behavior of the json-ld library to return valid expanded json-ld data from the `fluree.json-ld/expand` function, and makes all the necessary changes to use the new format for expanded json-ld data.

### DIFF
--- a/src/fluree/db/commit/storage.cljc
+++ b/src/fluree/db/commit/storage.cljc
@@ -54,7 +54,7 @@
     (when-let [[commit _proof] (<? (read-verified-commit storage commit-address))]
       (let [commit-id (commit-data/hash->commit-id commit-hash)]
         (assoc commit
-               "@id" commit-id
+               const/iri-id commit-id
                const/iri-address commit-address)))))
 
 (defn read-data-jsonld


src/fluree/db/connection/config.cljc
```diff
@@ -257,7 +251,7 @@
 (defn keywordize-node-id
   [node]
   (if (subject-node? node)
-    (update node :id iri->kw)
+    (update node "@id" iri->kw)
```
dpetran: Can we use the `const/iri-id` and friends in this file instead of the string literal versions?

src/fluree/db/flake/flake_db.cljc
```diff
@@ -148,41 +149,41 @@
...
+    (contains? mapx "@value")
     false
 
     (list-value? mapx)
     false
 
     (and
-     (contains? mapx :set)
-     (= #{:set :idx} (set (keys mapx))))
+     (contains? mapx "@set")
```
dpetran: Can we make a const for this?

src/fluree/db/flake/flake_db.cljc
```diff
@@ -148,41 +149,41 @@
...
-(defn node?
-  "Returns true if a nested value is itself another node in the graph.
+(defn subject-node?
+  "Returns true if a nested value is itself another subject node in the graph.
   Only need to test maps that have :id - and if they have other properties they
   are defining then we know it is a node and have additional data to include."
   [mapx]
   (cond
-    (contains? mapx :value)
+    (contains? mapx "@value")
```
dpetran: And use a `const` for this as well.

src/fluree/db/query/fql/parse.cljc
```diff
@@ -856,8 +860,8 @@
                          " Please use the JSON-LD \"@type\" keyword instead.")
                     {:status 400 :error :db/invalid-predicate}))
 
-    (= :type pred)
-    (let [values*  (map (fn [typ] {:id typ})
+    (= "@type" pred)
```
dpetran: `const` here

src/fluree/db/util.cljc
```diff
@@ -335,7 +338,6 @@
 
 (defn get-graph
   [jsonld]
-  (cond
-    (contains? jsonld :graph)   (:graph jsonld)
-    (contains? jsonld "@graph") (get jsonld "@graph")
-    :else                       jsonld))
+  (if (contains? jsonld "@graph")
+    (get jsonld "@graph")
+    jsonld))
```
dpetran, line 343: And `const`s in this file as well.

dpetran (approved):
:coconut:

src/fluree/db/util.cljc
zonotope: We can't use a const here because of a circular dependency in the const namespace. At least, I'd rather not hold up this pr to resolve the circular dependency. I'll leave a TODO for that.

PR #1091, opened 2025-08-19
test: stabilize drop-test by increasing max-old-indexes to 1000

bplatz:
Fix intermittent drop-test failure caused by background index garbage collection pruning older index roots/segments during pre-drop file count assertions. Increase max-old-indexes for this test to prevent GC from running during the assertions, making the test consistent across runs.


dpetran (approved):
:wastebasket:

PR #1093, opened 2025-08-20
Feature/bm25 ns

bplatz:
### Summary
This PR promotes virtual graphs to first-class, independent databases and removes the prior “##” prefix workaround. BM25 virtual graphs now maintain their own state and stay eventually consistent with their source ledgers via dependency tracking and background indexing.

### Why
- Eliminate special-casing and naming hacks.
- Make virtual graphs composable, testable, and life-cycle managed like any other DB.
- Provide a clear, extensible path for additional virtual graph types.

### What’s Changed
- Virtual graphs are now independent database objects with their own storage and lifecycle.
- BM25 indices update asynchronously as dependent ledgers change (eventual consistency).
- System tracks ledger→VG dependencies to prevent unsafe operations (e.g., blocking ledger drop when VGs depend on it).

### API and Behavior
- New: drop-virtual-graph API for clean deletion (nameservice record, index data, dependency entries).
- Creation returns a VG object (DB-like) that can be queried and synced.
- SyncableVirtualGraph protocol added to wait for indexing (e.g., for CI/ingestion flows).
- Dependency checks enforced when dropping ledgers.

### Nameservice and Storage
- Nameservice refactor to a multimethod-based architecture for JSON-LD and dependency registration.
- Clearer parameter names, reduced duplication, and improved async error propagation.
- Uses existing UTF-8 helpers instead of ad-hoc conversions.

### BM25 Improvements
- Creation logic moved to `fluree.db.virtual-graph.create`.
- Old FlatRank implementation removed.
- Cleaner indexing pipeline with status reporting for long-running syncs.

### Tests
- Expanded BM25 coverage: memory, filesystem, and federated datasets.
- Drop path tests validate dependency enforcement and full cleanup.
- Existing tests continue to pass.

### Breaking Changes
- “##” prefix for virtual graphs is no longer supported; virtual graphs are resolved via nameservice as first-class resources.
- Virtual graph creation now returns a VG database object rather than just a name.
- Some internal APIs are streamlined for consistency.

### Migration Notes
- Replace any “##name” usage with the virtual graph’s registered name.
- Where callers previously expected a string from VG creation, update them to accept/use the returned VG object.
- If you scripted ledger drops, ensure you drop dependent VGs first or handle the new dependency error.

### Operational Notes
- Eventual consistency by design: use the sync protocol to await index completion when deterministic reads are required (e.g., CI).
- Dependency maps are initialized at startup and updated on publish/retract; drop operations clean up both storage and nameservice entries.

### Files of Interest
- New: `fluree.db.virtual-graph.create`, `fluree.db.virtual-graph.drop`
- Refactors: `fluree.db.nameservice.storage`, `fluree.db.connection`, `fluree.db.api`
- BM25: multiple files updated for the new architecture and sync protocol

### DIFF
--- a/src/fluree/db/api.cljc
+++ b/src/fluree/db/api.cljc
@@ -1039,6 +1039,22 @@
    (promise-wrap
     (connection/trigger-ledger-index conn ledger-alias opts))))
 
+(defn reindex
+  "Offline index rebuild from commit history (regenerates stats).
+
+  Options:
+  - :from-t - Start from this transaction t (default: 1; t=0 is genesis)
+  - :batch-bytes - Novelty threshold per batch (default: reindex-max-bytes)
+  - :index-files-ch - Optional channel for index file notifications"
+  ([conn ledger-alias]
+   (reindex conn ledger-alias {}))
+  ([conn ledger-alias opts]
+   (validate-connection conn)
+   (promise-wrap
+    (go-try
+      (let [ledger (<? (connection/load-ledger-alias conn ledger-alias))]
+        (<? (ledger/reindex! ledger opts)))))))
+
 ;; Full-Text Search APIs (JVM only)
 
 #?(:clj
--- a/src/fluree/db/nameservice/storage.cljc
+++ b/src/fluree/db/nameservice/storage.cljc
@@ -89,12 +89,14 @@
 
 (defn merge-index-into-record
   "Merges index info from the separate index file into a main NS record.
-   Index file takes precedence if it has a higher index-t."
+   Index file takes precedence if it has an equal or higher index-t.
+   Equal t values favor the index file since it's written by reindex operations
+   and may point to a newer index root (e.g., v2 with stats vs v1 without)."
   [main-record index-record]
   (if index-record
     (let [main-index-t  (get-in main-record ["f:index" "f:t"] 0)
           file-index-t  (get-in index-record ["f:index" "f:t"] 0)]
-      (if (> file-index-t main-index-t)
+      (if (>= file-index-t main-index-t)
         (assoc main-record "f:index" (get index-record "f:index"))
         main-record))
     main-record))
--- a/resources/META-INF/native-image/com.fluree/db/native-image.properties
+++ b/resources/META-INF/native-image/com.fluree/db/native-image.properties
@@ -1,6 +1,21 @@
 # GraalVM Native Image configuration for Fluree DB
-# S3 Express One Zone support - AWS SDK v2
-
+#
+# Caffeine Cache Support:
+# Caffeine dynamically generates two sets of classes based on configuration:
+#
+# Cache implementations (LocalCacheFactory):
+# - SSMS: Size-bounded cache with strong keys/values (used in fluree.db.cache)
+# - SSMSA: Size-bounded cache with access ordering
+# - SW: Weak-values cache (used in fluree.db.json-ld.iri for SID interning)
+# - SS: Strong keys/values cache
+#
+# Node implementations (NodeFactory) for storing cache entries:
+# - PSMS: Node for size-bounded cache
+# - PS: Basic node with strong references
+# - PSW: Node for weak values cache
+#
+# S3 Express One Zone support - AWS SDK v2:
 # AWS SDK classes used for S3 Express One Zone session management are imported
 # directly in fluree.db.storage.s3-express, allowing GraalVM to detect them.
+#
 # The reflect-config.json in this directory provides reflection metadata for runtime access.
--- a/resources/META-INF/native-image/com.fluree/db/reflect-config.json
+++ b/resources/META-INF/native-image/com.fluree/db/reflect-config.json
@@ -1,5 +1,54 @@
 [
   {
+    "name": "com.github.benmanes.caffeine.cache.SSMS",
+    "allDeclaredConstructors": true,
+    "allPublicConstructors": true,
+    "allDeclaredMethods": true,
+    "allPublicMethods": true
+  },
+  {
+    "name": "com.github.benmanes.caffeine.cache.SSMSA",
+    "allDeclaredConstructors": true,
+    "allPublicConstructors": true,
+    "allDeclaredMethods": true,
+    "allPublicMethods": true
+  },
+  {
+    "name": "com.github.benmanes.caffeine.cache.SW",
+    "allDeclaredConstructors": true,
+    "allPublicConstructors": true,
+    "allDeclaredMethods": true,
+    "allPublicMethods": true
+  },
+  {
+    "name": "com.github.benmanes.caffeine.cache.SS",
+    "allDeclaredConstructors": true,
+    "allPublicConstructors": true,
+    "allDeclaredMethods": true,
+    "allPublicMethods": true
+  },
+  {
+    "name": "com.github.benmanes.caffeine.cache.PSMS",
+    "allDeclaredConstructors": true,
+    "allPublicConstructors": true,
+    "allDeclaredMethods": true,
+    "allPublicMethods": true
+  },
+  {
+    "name": "com.github.benmanes.caffeine.cache.PS",
+    "allDeclaredConstructors": true,
+    "allPublicConstructors": true,
+    "allDeclaredMethods": true,
+    "allPublicMethods": true
+  },
+  {
+    "name": "com.github.benmanes.caffeine.cache.PSW",
+    "allDeclaredConstructors": true,
+    "allPublicConstructors": true,
+    "allDeclaredMethods": true,
+    "allPublicMethods": true
+  },
+  {
     "name": "software.amazon.awssdk.auth.credentials.AwsBasicCredentials",
     "allDeclaredConstructors": true,
     "allPublicConstructors": true,
--- a/src/fluree/db/flake.cljc
+++ b/src/fluree/db/flake.cljc
@@ -162,12 +162,16 @@
   [x]
   (instance? Flake x))
 
-(defn- equiv-flake
+(defn equiv-flake
+  "Returns true if two flakes represent the same fact. Compares subject, predicate,
+   object, datatype, and metadata (which includes list index and language tag).
+   Intentionally ignores transaction time (t) and operation (op)."
   [f other]
   (and (= (s f) (s other))
        (= (p f) (p other))
        (= (o f) (o other))
-       (= (dt f) (dt other))))
+       (= (dt f) (dt other))
+       (= (m f) (m other))))
 
 (defn parts->Flake
   "Used primarily to generate flakes for comparator. If you wish to
--- a/src/fluree/db/flake/flake_db.cljc
+++ b/src/fluree/db/flake/flake_db.cljc
@@ -294,8 +294,13 @@
       (when (empty? all-flakes)
         (commit-error "Commit has neither assertions or retractions!"
                       commit-metadata))
-      (log/debug "Updating db" (str/join "@" [(:alias db) (:t db)])
-                 "to t:" t-new "with new commit:" commit-metadata)
+      (log/trace "Updating db" (str/join "@" [(:alias db) (:t db)])
+                 "to t:" t-new "with new commit:"
+                 (assoc commit-metadata
+                        :index (some-> (:index commit-metadata)
+                                       (select-keys [:id :address :v :data])
+                                       (update :data #(when (map? %)
+                                                        (select-keys % [:t :address :id :v :flakes :size]))))))
       (-> db*
           (merge-flakes t-new all-flakes)
           (assoc :commit commit-metadata)))))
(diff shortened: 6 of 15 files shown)


src/fluree/db/api.cljc
```diff
@@ -257,6 +259,31 @@
       (let [ledger (<? (connection/create-ledger conn ledger-alias opts))]
         (ledger/current-db ledger))))))
 
+(defn create-virtual-graph
```
zonotope: I think this function is too general for a top level api like this. I think we'll eventually support a lot of virtual graphs, and the name "virtual graph" itself feels more like it's leaking an internal implementation detail.

I think this function should be called something like `create-full-text-index`.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -75,30 +80,148 @@
...
+
+(defn get-commit
+  "Returns the minimal nameservice record."
+  ([record]
+   (get-commit record nil))
+  ([record _branch]
+   ;; Always return the record itself for new format
+   record))
+
+;; Virtual Graph Dependency Tracking Functions
```
zonotope: I think these functions are in the wrong namespace. This namespace is for the storage-backed implementation of the nameservice protocols and supporting functions. These functions are more general and don't apply to the storage-backed implementation specifically, so I think they should instead go in the `fluree.db.nameservice` namespace.

src/fluree/db/connection.cljc
```diff
@@ -500,33 +507,70 @@
...
       :index-dropped)))
 
+(defn- check-ledger-dependencies
+  "Checks if ledger has dependent virtual graphs.
+   Expects ledger-alias in 'ledger:branch' format."
+  [primary-pub ledger-alias]
+  (when primary-pub
+    (ns-storage/check-vg-dependencies primary-pub ledger-alias)))
+
+(defn- throw-if-has-dependencies
```
zonotope: This is minor, but I think we should avoid functions defined specifically for throwing exceptions because they pollute the stack trace. Instead of the top frame of the stack trace being the site of the immediate problem, it's this function. You have to go into the subsequent frames to find out where the real problem is. It's not a big deal, but it's also not idea. 

I think we instead could have a function that constructs the exception object, and throw it at the site of the error. That way we can avoid repetition without polluting the stack trace. 

This function could instead be called `existing-dependency-error` and it would just build the `ex-info` itself, and then, later on when the error occurs, we could throw it with 

```(throw (existing-dependency-error alias dependent-vgs))```

File: src/fluree/db/flake/transact.cljc [hunk 4]
```diff
@@ -110,6 +109,7 @@
                      result
                      [@db-vol result]))))))
 
+;; Virtual graphs are now created through the nameservice API, not transactions
```
zonotope: I don't think this comment is necessary.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -75,30 +80,148 @@
...
+                               (update-ns-record ns-record ledger-alias commit-address commit-t
+                                                 index-address index-t))
+              res            (<? (storage/swap-json store filename record-updater))]
+          (log/debug "Nameservice published ledger:" {:ledger ledger-alias :filename filename})
+          res))))
+
+  (retract [this target]
+    (go-try
+      (let [;; Check if target is a ledger (contains :) or VG (no :)
+            ledger? (str/includes? target ":")
```
zonotope: I think this should be an explicit function to determine if an alias represents a ledger.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -133,19 +256,21 @@
...
+                        record (json/parse content-str false)]
                     (recur (rest remaining-paths) (conj records record)))
                   (recur (rest remaining-paths) records)))
               records))
           [])
-        ;; Fallback for stores that don't support ListableStore
+        ;; Fallback for stores that don't support RecursiveListableStore
         (do
-          (log/warn "Storage backend does not support RecursiveListableStore protocol")
+          (log/debug "Storage backend does not support RecursiveListableStore protocol")
```
zonotope: I think we should perform the `satisfies?` check on line 250 at nameservice initialization time instead and throw an error there.

src/fluree/db/nameservice/virtual_graph.cljc
```diff
@@ -0,0 +1,42 @@
+(ns fluree.db.nameservice.virtual-graph
```
zonotope: Why can't we use the same api we use for ledgers here?

src/fluree/db/nameservice/storage.cljc
```diff
@@ -6,18 +6,23 @@
...
-  (let [[ledger-name branch] (util.ledger/ledger-parts ledger-alias)
-        branch (or branch const/default-branch-name)]
-    (str const/ns-version "/" ledger-name "/" branch ".json")))
+   Can handle both resource names (for VG) and ledger aliases.
+   For ledgers, expects ledger-alias to be in format 'ledger:branch'.
+   Returns path like 'ns@v2/ledger-name/branch.json' for ledgers
+   or 'ns@v2/resource-name.json' for virtual graphs."
+  [resource-name]
+  (if (str/includes? resource-name ":")
+    (let [[ledger-name branch] (util.ledger/ledger-parts resource-name)
```
zonotope: I think beyond having an explicit function to determine if an alias corresponds to a ledger, we should have separate functions here for a ledger's local filename vs a bm25 index's.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -75,30 +80,148 @@
...
+   "@id" vg-name
+   "@type" (cond-> ["f:VirtualGraphDatabase"]
+             vg-type (conj vg-type))
+   "f:name" vg-name
+   "f:status" (or status "ready")
+   "f:dependencies" (mapv (fn [dep] {"@id" dep}) dependencies)
+   "fidx:config" {"@type" "@json"
+                  "@value" config}})
+
+;; NOTE: Primary StorageNameService defined later; this earlier definition was removed to avoid duplication
```
zonotope: i don't think this comment is necessary.

src/fluree/db/query/api.cljc
```diff
@@ -175,21 +177,70 @@
...
-            ledger       (<? (connection/load-ledger-alias conn base-alias))
-            db           (ledger/current-db ledger)
-            t*           (or explicit-t t)
-            query*       (-> sanitized-query
-                             (assoc :t t*)
-                             (ledger-opts-override db))]
-        (<? (restrict-db db tracker query* conn)))
+      (log/debug "load-alias called with:" alias)
+      ;; First try to load as a virtual graph
+      (if-let [vg (<? (load-virtual-graph conn alias))]
```
zonotope: Is there a way to dispatch on the nameservice record instead? If bm25 indexes have a different nameservice record format than graph ledgers, then we could call the same load function and instantiate the right kind of graph based on the format of the record it gets back.

src/fluree/db/virtual_graph.cljc
```diff
@@ -1,14 +1,47 @@
...
+      (and (vector? vg-type) (iri/sid? (first vg-type)))
+      (-> vg-type first iri/get-name str/lower-case)
+
+      ;; Handle string array type from nameservice
+      (and (vector? vg-type) (string? (first vg-type)))
+      (let [type-str (->> vg-type
+                          (filter #(str/starts-with? % "fidx:"))
+                          first)]
+        (when type-str
+          (-> type-str (str/split #":") second str/lower-case)))
```
zonotope: this should be an explicit function to parse type strings, and it should be reused on line 40

src/fluree/db/virtual_graph/bm25/index.clj
```diff
@@ -101,12 +104,15 @@
...
-    (try*
+    (try
       (let [{::vg-parse/keys [target limit timeout] :as search-params} (vg-parse/get-search-params solution)
             _ (when-not target
                 (throw (ex-info "No search target for virtual graph. Did you forget @context in your query?"
                                 {:status 400 :error
                                  :db/invalid-query})))
+            ;; Guard against empty/blank targets
+            _ (when (and (string? target) (str/blank? target))
+                (async/onto-chan! out-ch []))
```
zonotope: you can just call `close!` on `out-ch` here.

src/fluree/db/virtual_graph/bm25/index.clj
```diff
@@ -146,17 +152,30 @@
...
-                                 (assoc idx-state :index new-index
-                                        :pending-ch nil)))
-        (>! new-pending-ch new-index)))
+      (try
+        (let [latest-index  (if pending-ch
+                              (<! pending-ch)
+                              index)
+              status-update (fn [status]
+                              (swap! new-index-state assoc :pending-status status))
+              new-index     (<! (bm25.update/upsert-items bm25 latest-index items-count items-ch status-update))
```
zonotope: How are errors handled here?

src/fluree/db/virtual_graph/bm25/index.clj
```diff
@@ -146,17 +152,30 @@
...
+          (swap! new-index-state (fn [idx-state]
+                                   (assoc idx-state :index new-index
+                                          :pending-ch nil)))
+          (log/debug "BM25 index update complete, writing to storage...")
+          ;; Persist the updated BM25 index to storage
+          (when catalog
+            (let [updated-bm25 (assoc bm25 :index-state new-index-state :t t)]
+              (log/debug "Writing BM25 index to storage for VG:" (:alias updated-bm25))
+              (try
+                (<! (vg/write-vg catalog updated-bm25))
```
zonotope: this doesn't look like it would ever actually throw any errors.

src/fluree/db/virtual_graph/bm25/index.clj
```diff
@@ -146,17 +152,30 @@
...
+              new-index     (<! (bm25.update/upsert-items bm25 latest-index items-count items-ch status-update))
+              catalog       (:index-catalog bm25)]
+          ;; reset index state atom once index is complete, remove pending-ch
+          (swap! new-index-state (fn [idx-state]
+                                   (assoc idx-state :index new-index
+                                          :pending-ch nil)))
+          (log/debug "BM25 index update complete, writing to storage...")
+          ;; Persist the updated BM25 index to storage
+          (when catalog
+            (let [updated-bm25 (assoc bm25 :index-state new-index-state :t t)]
```
zonotope: `new-index-state` is a freshly created atom in this function on line 150. Is there a reason why you couldn't have reused the atom that was already on `bm25` under the `:index-state` key?

src/fluree/db/virtual_graph/bm25/index.clj
```diff
@@ -263,25 +335,38 @@
...
+  ;; return the VG alias
   (-aliases [_]
-    [db-alias]))
+    [alias])
+
+  dbproto/IFlureeDb
+  (-query [_this _tracker _query]
+    ;; BM25 virtual graphs handle queries through the where/Matcher protocol
+    ;; This method should not be called directly on BM25 instances
+    (throw (ex-info "BM25 virtual graphs do not support direct query execution"
```
zonotope, line 456: If this is the case, then this protocol shouldn't be implemented at all. I get that this particular protocol is a little thorny because there are still some circular dependencies, so it can wait for another pr.

src/fluree/db/virtual_graph/create.cljc
```diff
@@ -0,0 +1,117 @@
...
+  [{:keys [config]}]
+  (let [ledgers (get-in config [:ledgers] [])]
+    (when (not= 1 (count ledgers))
+      (throw (ex-info "BM25 virtual graphs currently support only a single ledger. Multi-ledger support is not yet implemented."
+                      {:error :db/invalid-config
+                       :type :bm25
+                       :ledgers ledgers
+                       :count (count ledgers)})))))
+
+(defn- ensure-string-name
```
zonotope: This function is not necessary. `clojure.core/name` can take keywords, symbols, or strings as input, and it will safely return the string representation of all three as its output.

src/fluree/db/virtual_graph/create.cljc
```diff
@@ -0,0 +1,117 @@
...
+  (go-try
+    (validate-common-config config)
+    (<? (create-vg conn config))))
+
+(defn- validate-bm25-config
+  "Validates BM25-specific configuration."
+  [{:keys [config]}]
+  (let [ledgers (get-in config [:ledgers] [])]
+    (when (not= 1 (count ledgers))
+      (throw (ex-info "BM25 virtual graphs currently support only a single ledger. Multi-ledger support is not yet implemented."
```
zonotope, line 48: I think if we only support a single ledger, then we should make it clear by only accepting a `:ledger` key in the config, and only loading a single ledger at a time in the implementation below. We are free to change that behavior through addition to the keys we accept in the config in a backwards compatible way when we implement multi ledger support. 

Having incomplete implementations for things we plan in the future leads to technical debt that makes it more difficult to modify the codebase going forward, and we will have a lot more information about what the final implementation of the feature is, and what shape it has to take, when we are ready to implement it.

src/fluree/db/virtual_graph/create.cljc
```diff
@@ -0,0 +1,117 @@
...
+     :vg-type "fidx:BM25"
+     :config config
+     :dependencies (or dependencies
+                       (mapv #(str % ":main") ledgers))}))
+
+(defn- load-and-validate-ledgers
+  "Loads all ledgers and validates they exist. Returns loaded ledgers."
+  [conn ledger-aliases]
+  (go-try
+    (loop [remaining ledger-aliases
```
zonotope, line 69: I think we should remove this loop and load a single ledger alias since that's all we support. We should also make the function name singular.

src/fluree/db/virtual_graph/create.cljc
```diff
@@ -0,0 +1,117 @@
...
+           loaded {}]
+      (if-let [ledger-alias (first remaining)]
+        (if-let [ledger (<? (connection/load-ledger-alias conn ledger-alias))]
+          (recur (rest remaining) (assoc loaded ledger-alias ledger))
+          (throw (ex-info (str "Ledger does not exist: " ledger-alias)
+                          {:error :db/invalid-config
+                           :ledger ledger-alias})))
+        loaded))))
+
+(defn- initialize-bm25-for-ledgers
```
zonotope, line 79: We should make this function name singular.

src/fluree/db/virtual_graph/drop.cljc
```diff
@@ -0,0 +1,46 @@
...
+          (log/debug "Found" (count vg-files) "VG files to delete")
+          (doseq [file-path vg-files]
+            (log/debug "Deleting VG file:" file-path)
+            (<? (storage/delete storage file-path)))
+          ;; Try to delete the directory itself (may not work on all storage types)
+          (try*
+            (<? (storage/delete storage vg-path))
+            (catch* e
+              nil)))
+        (log/warn "Storage backend does not support listing files, cannot clean up VG artifacts"))
```
zonotope: This should return a different keyword here indicating that the artifacts were not dropped in this case.

src/fluree/db/virtual_graph/drop.cljc
```diff
@@ -0,0 +1,46 @@
...
+            [fluree.db.storage :as storage]
+            [fluree.db.util :refer [try* catch*]]
+            [fluree.db.util.async :refer [<? go-try]]
+            [fluree.db.util.log :as log]))
+
+(defn drop-artifacts
+  "Deletes all storage artifacts for a virtual graph"
+  [{:keys [index-catalog] :as _conn} vg-name]
+  (go-try
+    (let [storage (:storage index-catalog)
```
zonotope: As far as I know, [there is no :storage key within a Catalog record](https://github.com/fluree/db/blob/62397470eecae3a316224f34c8add7b2abf5a364/src/fluree/db/storage.cljc#L195), so this will always be `nil`. [The catalog _is_ storage](https://github.com/fluree/db/blob/62397470eecae3a316224f34c8add7b2abf5a364/src/fluree/db/storage.cljc#L272-L299), it doesn't have storage, so you should write to it directly.

src/fluree/db/virtual_graph/nameservice_loader.cljc
```diff
@@ -0,0 +1,91 @@
...
+
+(defn load-vg-config-from-nameservice
+  "Loads a virtual graph configuration from the nameservice"
+  [nameservice vg-name]
+  (go-try
+    (let [vg-record (<? (ns-vg/get-virtual-graph nameservice vg-name))]
+      (when (= :not-found vg-record)
+        (throw (ex-info (str "Virtual graph not found in nameservice: " vg-name)
+                        {:status 404
+                         :error :db/invalid-query})))
```
zonotope: I don't think this is the correct error here. The query could still be valid if the graph is missing or removed, just like a query against a non existing ledger.

src/fluree/db/virtual_graph/parse.cljc
```diff
@@ -31,6 +32,13 @@
   [triple]
   (-> triple (nth 2) where/get-iri))
 
+(defn- obj-node-ref
+  "Returns the object as a map if it's a node reference"
+  [triple]
+  (let [obj (nth triple 2)]
+    (when (and (map? obj) (::where/var obj))
```
zonotope: You should use the `where/get-variable` function instead of `::where/var` directly.

src/fluree/db/virtual_graph/parse.cljc
```diff
@@ -52,12 +60,25 @@
         (assoc-in solution [::virtual-graph ::timeout] (obj-var triple))
 
         (= const/iri-index-result p-iri)
-        (assoc-in solution [::virtual-graph ::result ::id] (obj-var triple))
+        ;; idx:result can have either a simple variable or a node reference
+        ;; Store the node var so we can match nested properties later
+        (if-let [node-ref (obj-node-ref triple)]
+          (assoc-in solution [::virtual-graph ::result-node-var] (::where/var node-ref))
```
zonotope: `where/get-variable` here too.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -75,30 +80,148 @@
...
+  "Returns the minimal nameservice record."
+  ([record]
+   (get-commit record nil))
+  ([record _branch]
+   ;; Always return the record itself for new format
+   record))
+
+;; Virtual Graph Dependency Tracking Functions
+
+(defn is-virtual-graph-record?
```
zonotope: it's more idiomatic to drop the `is-` here.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -75,30 +80,148 @@
...
-  (retract [_ ledger-alias]
-    (let [filename (local-filename ledger-alias)
-          address  (-> store
-                       storage/location
-                       (storage/build-address filename))]
-      (storage/delete store address)))
+  (publish [this record]
+    (go-try
+      (if-let [vg-name (:vg-name record)]
+        ;; Virtual Graph: use write-bytes (non-atomic, but VGs aren't concurrently updated)
```
zonotope: Is this a guarantee, even with shared storage and rotating transactors?

src/fluree/db/nameservice/storage.cljc
bplatz: In 8e3b2ad5667968aa1df65f3abb75fed567ea080d

src/fluree/db/connection.cljc
bplatz: In fd48cfd69ff1030deed0a35dd0c3e5e04fa522f3

src/fluree/db/nameservice/storage.cljc
bplatz: In ddde16c3518fd032f1e2c170e4f6a3044884f9b2

src/fluree/db/nameservice/storage.cljc
bplatz: In ddde16c3518fd032f1e2c170e4f6a3044884f9b2

src/fluree/db/nameservice/storage.cljc
bplatz: In ddde16c3518fd032f1e2c170e4f6a3044884f9b2

src/fluree/db/flake/transact.cljc
bplatz: In ddde16c3518fd032f1e2c170e4f6a3044884f9b2

src/fluree/db/virtual_graph/create.cljc
bplatz: In d976b3535fdfeb011e0fbc2e1e9f1c812f34645b

src/fluree/db/virtual_graph/bm25/index.clj
bplatz: In d976b3535fdfeb011e0fbc2e1e9f1c812f34645b

src/fluree/db/virtual_graph.cljc
bplatz: In d976b3535fdfeb011e0fbc2e1e9f1c812f34645b

src/fluree/db/nameservice/storage.cljc
bplatz: In d976b3535fdfeb011e0fbc2e1e9f1c812f34645b

src/fluree/db/virtual_graph/drop.cljc
bplatz: In 24136416f762eb25a7c986935e21ecbc3d551f3f

src/fluree/db/virtual_graph/drop.cljc
bplatz: In 24136416f762eb25a7c986935e21ecbc3d551f3f

src/fluree/db/virtual_graph/parse.cljc
bplatz: In 24136416f762eb25a7c986935e21ecbc3d551f3f

src/fluree/db/nameservice/storage.cljc
bplatz: In 24136416f762eb25a7c986935e21ecbc3d551f3f

src/fluree/db/virtual_graph/parse.cljc
bplatz: In 24136416f762eb25a7c986935e21ecbc3d551f3f

src/fluree/db/virtual_graph/create.cljc
bplatz: Supporting multiple ledgers is an important requirement and something we should implement sooner rather than later - We'd have a deprecated API as soon as it supported, or have to support both APIs which I think is worse technical debt. I would like to have put support in here already for it, but it would have made this PR even larger so I focused on first what we had current support for.

src/fluree/db/virtual_graph/create.cljc
bplatz: Likewise, multi-ledger support forthcoming and this is working code that would have to get changed. I think we should keep it as-is and I'll do another PR shortly for multiple ledger support.

src/fluree/db/virtual_graph/create.cljc
bplatz: Same API comment as above.

src/fluree/db/virtual_graph/create.cljc
zonotope: Ok, if it's immediately forthcoming then I think it's fine.

src/fluree/db/virtual_graph/nameservice_loader.cljc
bplatz: In 522e1c0b5e05120dbcf3a8cd6a60a6397f8d559e

src/fluree/db/virtual_graph/bm25/index.clj
bplatz: In 522e1c0b5e05120dbcf3a8cd6a60a6397f8d559e

src/fluree/db/virtual_graph/bm25/index.clj
bplatz: In 522e1c0b5e05120dbcf3a8cd6a60a6397f8d559e

src/fluree/db/virtual_graph/bm25/index.clj
bplatz: In 522e1c0b5e05120dbcf3a8cd6a60a6397f8d559e

src/fluree/db/nameservice/virtual_graph.cljc
bplatz: In 2ce433584b5908c800ecf33d5dee549cfa2c412e

src/fluree/db/api.cljc
bplatz: Addressed in 667805a0bc437a5c67d183bc1cfd227807baa82f

src/fluree/db/nameservice/storage.cljc
bplatz: The code has been updated to use storage/swap-json for virtual graphs (line 195-202 now in the code)

src/fluree/db/query/api.cljc
```diff
@@ -195,22 +197,48 @@
...
+      (if (virtual-graph-record? ns-record)
+        ;; Virtual graph - load via VG loader (JVM only)
+        #?(:clj
+           (let [;; For VG queries, we need a source ledger's db to initialize the VG
+                 ;; Get it from the VG's dependencies
+                 deps          (get ns-record "fidx:dependencies")
+                 source-alias  (first deps)
+                 source-ledger (<? (connection/load-ledger-alias conn source-alias))
+                 source-db     (ledger/current-db source-ledger)
+                 vg            (<? (vg-loader/load-virtual-graph-from-nameservice
```
zonotope, line 225: I don't think this `vg` binding is necessary. We can just return the `(<? (vg-loader/load-virtual-graph-from-nameservice ...))` call directly

zonotope (approved):
🕹️

PR #1095, opened 2025-08-21
Add merge, rebase and reset operations for branch management

bplatz:
## Summary

This sits on top of PR https://github.com/fluree/db/pull/1095 which in turn sits on top of several others. The final part of the full branching support which sits on top of this branch is https://github.com/fluree/db/pull/1102 which adds cuckoo filters for index garbage collection.

  This PR implements core branch management operations for Fluree, providing Git-like capabilities for database version control:

  - **Merge with fast-forward and squash modes** - Integrate changes from source into target branch
  - **Rebase with squash mode** - Replay source branch onto target (true git-style rebase)
  - **Safe reset** - Non-destructively revert branches to previous states
  - **Branch divergence analysis** - Determine relationships between branches
  - **Branch graph visualization** - ASCII and JSON representations of branch relationships

  ## Features Implemented

  ### 1. Merge Operations (Updates Target Branch)

  #### Fast-forward
  Automatically moves target branch pointer when no divergence exists:
  ```clojure
  @(fluree/merge! conn "mydb:feature" "mydb:main")
  ;; Returns: {:status :success :strategy "fast-forward" ...}

  Squash Merge

  Combines multiple commits into a single commit on target, with intelligent cancellation of assert/retract pairs:
  @(fluree/merge! conn "mydb:feature" "mydb:main"
    {:squash? true
     :message "Feature complete: Add user authentication"})

  2. Rebase Operations (Updates Source Branch)

  True git-style rebase that updates the source branch by replaying it onto target:
  @(fluree/rebase! conn "mydb:feature" "mydb:main"
    {:squash? true
     :message "Rebase feature onto latest main"})
  ;; Note: Updates feature branch, main remains unchanged

  3. Safe Reset

  Creates a new commit that reverts the branch to a previous state (non-destructive):
  ;; Reset to transaction number
  @(fluree/reset-branch! conn "mydb:main" {:t 90}
    {:message "Reverting to stable state"})

  ;; Reset to commit SHA
  @(fluree/reset-branch! conn "mydb:main" {:sha "abc123"}
    {:message "Reverting to release v1.0"})

  4. Branch Analysis & Visualization

  Divergence Analysis

  Check if branches can fast-forward before performing operations:
  (def divergence @(fluree/branch-divergence conn "mydb:feature" "mydb:main"))
  ;; Returns: {:can-fast-forward true 
  ;;           :fast-forward-direction :branch1-to-branch2
  ;;           :common-ancestor "sha..."}

  Branch Graph

  Visualize branch relationships:
  ;; ASCII visualization
  (println (<!! (merge/branch-graph conn "mydb" {:format :ascii})))

  ;; JSON data structure
  (def graph (<!! (merge/branch-graph conn "mydb" {:format :json})))

  Key Implementation Details

  - Correct git semantics: merge! updates target branch, rebase! updates source branch
  - Smart cancellation: When squashing, assert/retract pairs for the same value cancel out
  - Multi-cardinality support: Correctly handles both set and list semantics with meta field preservation
  - Namespace handling: Properly manages RDF namespace mappings across branch operations
  - Branch recreation: Rebase operations delete and recreate branches to maintain correct history
  - Comprehensive testing: All 246 tests passing, including cancellation, divergence, and file-based storage

  Architecture

  The implementation is modularized across focused namespaces:
  - fluree.db.merge - Public API (refactored for clarity)
  - fluree.db.merge.operations - Core operations (squash!, fast-forward!, safe-reset!)
  - fluree.db.merge.flake - Flake manipulation and cancellation logic
  - fluree.db.merge.branch - Branch analysis and LCA detection
  - fluree.db.merge.graph - Branch visualization
  - Supporting namespaces for commits, database prep, and responses

  Documentation

  See docs/branch-operations.md for:
  - Complete API reference with correct semantics
  - Usage examples and workflows
  - Clear distinction between merge and rebase operations
  - Implementation details for developers
  - Troubleshooting guide

  Testing

  Run tests with:
  make cljtest

  Test coverage includes:
  - Fast-forward merge scenarios
  - Squash merge with divergent branches
  - Rebase operations updating source branch
  - Assert/retract cancellation verification
  - Safe reset operations
  - Branch graph visualization
  - File and memory storage compatibility

  Next Steps

  - Implement commit-by-commit replay for rebase (without squash)
  - Add detailed anomaly reports to merge operations
  - Implement three-way merge with conflict resolution (when SHACL/policy validation is added)
  - Implement hard reset mode with branch pointer manipulation
  - Add cherry-pick functionality for selective commit application


test-branch-graph.clj
```diff
@@ -0,0 +1,14 @@
+(require '[fluree.db.api :as fluree])
```
zonotope: Was this file meant to be removed?

src/fluree/db/api.cljc
```diff
@@ -635,6 +636,216 @@
...
+      
+  Returns promise resolving to:
+    {:status :success|:conflict|:error
+     :operation :merge
+     :from '...' :to '...'
+     :strategy '3-way'
+     :commits {:merged [...] :conflicts [...]}
+     :new-commit 'sha'}
+     
+  Phase 2 - Not yet implemented."
```
zonotope: What does this mean?

src/fluree/db/api.cljc
```diff
@@ -635,6 +636,216 @@
...
+     :operation :reset
+     :branch '...'
+     :mode :safe|:hard
+     :reset-to {:t 90}|{:sha '...'}
+     :new-commit 'sha'  ; For safe mode
+     :archived {:type :tag :name '...'}  ; For hard mode
+     :previous-head 'sha'}
+     
+  Phase 1 implements: safe reset only.
+  Phase 2 will add: hard reset with archiving."
```
zonotope: I don't think we should document anything that hasn't been implemented yet, so we should remove the "Phase 2 ..." line here, as well as any references to things that don't already exist like "Hard mode ..."

We should also remove the "Phase 1" line in favor of explicit documentation of what exists.

src/fluree/db/api.cljc
```diff
@@ -635,6 +636,216 @@
...
+  ([conn branch to]
+   (reset-branch! conn branch to {}))
+  ([conn branch to opts]
+   (validate-connection conn)
+   (promise-wrap
+    (merge/reset-branch! conn branch to opts))))
+
+;; Legacy API - Deprecated, use rebase! instead
+(defn merge-branches!
```
zonotope: This should be marked with deprecated metadata.

src/fluree/db/commit/storage.cljc
```diff
@@ -38,13 +38,29 @@
...
+            addr-path     (conj subject-path "address")
+            subject-json  (if cred? (get commit-json "credentialSubject") commit-json)
+            ;; Prefer commit id from address/hash; fallback to computing from JSON only if needed
+            existing-id   (get subject-json "id")
+            ;; Extract base32 hash-like token from the address if present
+            addr-str      (str commit-address)
+            hash-token    (last (clojure.string/split addr-str #"/"))
+            normalized-b  (when (and hash-token (clojure.string/starts-with? hash-token "b")) hash-token)
+            id-from-addr  (when normalized-b (str "fluree:commit:sha256:" normalized-b))
+            computed-id   (or id-from-addr (commit-data/commit-json->commit-id subject-json))
```
zonotope: There should be a separate function to compute the id if necessary. This let form is hard to follow because there are so many bindings doing so many things in one place.

src/fluree/db/commit/storage.cljc
```diff
@@ -38,13 +38,29 @@
...
+      (log/trace "read-commit at:" commit-address "data:" commit-json)
+      (let [cred?         (contains? commit-json "credentialSubject")
+            subject-path  (if cred? ["credentialSubject"] [])
+            addr-path     (conj subject-path "address")
+            subject-json  (if cred? (get commit-json "credentialSubject") commit-json)
+            ;; Prefer commit id from address/hash; fallback to computing from JSON only if needed
+            existing-id   (get subject-json "id")
+            ;; Extract base32 hash-like token from the address if present
+            addr-str      (str commit-address)
+            hash-token    (last (clojure.string/split addr-str #"/"))
```
zonotope: There is a recently added [`get-hash` method on the `ContentArchive` protocol](https://github.com/fluree/db/blob/0ce80297e61f90cb59dffd4538b79e302fb48e60/src/fluree/db/storage.cljc#L130-L131)

src/fluree/db/commit/storage.cljc
zonotope: There is in fact a recently added [`hash->commit-id` function](https://github.com/fluree/db/blob/0ce80297e61f90cb59dffd4538b79e302fb48e60/src/fluree/db/flake/commit_data.cljc#L197-L199)

bplatz: This was addressed in the upstream branch.

src/fluree/db/commit/storage.cljc
bplatz: This was addressed in the upstream branch.

test-branch-graph.clj
bplatz: Addressed in e84ccafa84a4d19aaeec540ce075907129e9fbbf

src/fluree/db/api.cljc
bplatz: Addressed in e84ccafa84a4d19aaeec540ce075907129e9fbbf

src/fluree/db/api.cljc
bplatz: Addressed in e84ccafa84a4d19aaeec540ce075907129e9fbbf

src/fluree/db/api.cljc
bplatz: Addressed in e84ccafa84a4d19aaeec540ce075907129e9fbbf

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+           :or {ff :auto
+                squash? false}} opts
+
+          {:keys [source-db target-db source-branch-info target-branch-info]}
+          (<? (merge-branch/load-branches conn from to))
+
+          can-ff? (<? (merge-branch/can-fast-forward? conn source-db target-db
+                                                      source-branch-info target-branch-info))]
+
+      (when preview?
```
zonotope, line 48: This doesn't short circuit because of the `when`. The subsequent merging code here will always run whether or not the `preview?` option is set, which is the opposite of what we want when people set that option.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+        (reduced
```
zonotope, line 49: This doesn't look like it's taking place inside of a reduction. What is this `reduced` doing?

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+                        {:status 501 :error :not-implemented}))))))
+
+(defn- recreate-branch
+  "Deletes and recreates a branch pointing to a new commit."
+  [conn branch-spec temp-branch new-commit-id branch-metadata]
+  (go-try
+    ;; Delete the existing branch
+    (<? (api.branch/delete-branch! conn branch-spec))
+    ;; Recreate it pointing to the new commit
+    (<? (api.branch/create-branch!
```
zonotope, line 90: is it possible to do this in a non-destructive way? i'm worried about an error upon recreation after the original branch data has been deleted.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+    to - Target branch spec (e.g., 'ledger:main')
+    opts - Merge options:
+      :ff - Fast-forward behavior (default :auto)
+        :auto - Fast-forward when possible
+        :only - Only allow fast-forward
+        :never - Never fast-forward
+      :squash? - Combine all commits into one (default false)
+      :preview? - Dry run without changes (default false)
+      
+  Returns promise resolving to merge result with anomalies report."
```
zonotope, line 34: This function does not return a promise; it returns a core.async channel.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+  
+  Parameters:
+    conn - Connection object
+    from - Source branch to rebase (will be updated)
+    to - Target branch to rebase onto (unchanged)
+    opts - Rebase options:
+      :squash? - Combine all commits into one (default false)
+      :preview? - Dry run without changes (default false)
+      
+  Returns promise resolving to rebase result."
```
zonotope, line 110: This function does not return a promise.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+
+(defn branch-divergence
+  "Analyzes divergence between two branches.
+  
+  Parameters:
+    conn - Connection object
+    branch1-spec - First branch spec
+    branch2-spec - Second branch spec
+    
+  Returns promise resolving to divergence analysis including:
```
zonotope, line 226: This function does not return a promise.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+  
+  Parameters:
+    conn - Connection object
+    ledger-spec - Ledger specification (e.g., 'myledger')
+    opts - Options map:
+      :format - :json (default) or :ascii
+      :depth - Number of commits to show (default 20, :all for everything)
+      :branches - Specific branches to include (default: all)
+  
+  Returns promise resolving to graph data in requested format."
```
zonotope, line 260: This function does not return a promise

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+           :or {squash? false}} opts
+
+          ;; Load branches and find LCA
+          {:keys [source-db target-db source-branch-info target-branch-info]}
+          (<? (merge-branch/load-branches conn from to))
+
+          lca (<? (merge-branch/find-lca conn source-db target-db
+                                         source-branch-info target-branch-info))]
+
+      (when preview?
```
zonotope, line 124: This does not short circuit, so the rebasing code below will always execute whether or not the `preview?` option is set.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+        (reduced
```
zonotope, line 125: This is not taking place inside a reduction.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+                                      (:id (:commit result)))))
+                              (catch #?(:clj Exception :cljs js/Error) e
+                                ;; Clean up temp branch on error
+                                (try
+                                  (<? (api.branch/delete-branch! conn temp-branch-name))
+                                  (catch #?(:clj Exception :cljs js/Error) _ nil))
+                                (throw e)))
+
+              ;; Recreate source branch pointing to new commit
+              _ (<? (recreate-branch conn from
```
zonotope, line 178: This and the block below should happen outside of the `let` binding vector because the bindings are never used.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+                                     temp-branch-name
+                                     new-commit-id
+                                     (select-keys source-branch-info [:protected :description])))
+
+              ;; Clean up temp branch
+              _ (try
+                  (<? (api.branch/delete-branch! conn temp-branch-name))
+                  (catch #?(:clj Exception :cljs js/Error) _ nil))]
```
zonotope, line 186: I think we should at least log every exception that's caught. If there's an explicit list of possible errors during normal operations, like if the branch is already deleted, then we should re-throw any caught exception that falls out of that list.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+  [conn branch to opts]
+  (go-try
+    (let [{:keys [mode preview?]
+           :or {mode :safe}} opts]
+
+      (when (= mode :hard)
+        (throw (ex-info "Hard reset not yet implemented"
+                        {:status 501 :error :not-implemented})))
+
+      (when preview?
```
zonotope, line 207: Does not short circuit.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+          :to to
+          :lca lca
+          :strategy (if squash? "squash" "replay")}))
+
+      ;; Get commits from source since LCA
+      (let [source-commits (<? (merge-commit/extract-commits-since conn source-db lca))]
+
+        ;; Check if there's anything to rebase
+        (when (empty? source-commits)
+          (throw (ex-info "Nothing to rebase - source branch is already up to date"
```
zonotope, line 138: I don't think this is actually an error, so it shouldn't throw an exception.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+
+              [ledger-id _] (util.ledger/ledger-parts from)
+
+              ;; Create temporary branch manually (can't use with-temp-branch as we need it for recreate)
+              temp-branch-name (str ledger-id ":__rebase_temp_" (random-uuid))
+              _ (<? (api.branch/create-branch! conn temp-branch-name to nil))
+
+              ;; Create the rebased commit on temp branch
+              new-commit-id (try
+                              (<? (go-try
```
zonotope, line 160: This `go-try` here creates a redundant and unnecessary channel. Just wrapping the `let` in a `try` should have the same effect.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+                                          ;; Commit the rebased state
+                                          result (<? (transact/commit! temp-ledger staged-db
+                                                                       {:message message
+                                                                        :author (or (:author opts) "system/rebase")}))]
+                                      (:id (:commit result)))))
+                              (catch #?(:clj Exception :cljs js/Error) e
+                                ;; Clean up temp branch on error
+                                (try
+                                  (<? (api.branch/delete-branch! conn temp-branch-name))
+                                  (catch #?(:clj Exception :cljs js/Error) _ nil))
```
zonotope, line 174: We should at least log every exception caught.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+                                         source-branch-info target-branch-info))]
+
+      (when preview?
+        (reduced
+         {:status :preview
+          :operation :rebase
+          :from from
+          :to to
+          :lca lca
+          :strategy (if squash? "squash" "replay")}))
```
zonotope, line 131: I think the `squash?` case should be handled by a separate function. `squashed?` is referenced many times in subsequent code, which is a source of future bugs as these functions get modified.

src/fluree/db/merge/db.cljc
```diff
@@ -0,0 +1,103 @@
+(ns fluree.db.merge.db
```
zonotope, line 1: There needs to be a protocol for this functionality. A lot of the code here assumes we'll be operating on a `FlakeDB` explicitly, and there's a function that assumes that an `AsyncDB` will always wrap a `FlakeDB`. We should be able to use any dbs we receive interchangeably without ever having to explicitly check what kind of dbs they are, so new functionality on dbs should be defined in terms of new protocols.

src/fluree/db/merge.cljc
```diff
@@ -0,0 +1,262 @@
...
+      :squash? - Combine all commits into one (default false)
+      :preview? - Dry run without changes (default false)
+      
+  Returns promise resolving to merge result with anomalies report."
+  [conn from to opts]
+  (go-try
+    (merge-branch/validate-same-ledger! from to)
+    (let [{:keys [ff squash? preview?]
+           :or {ff :auto
+                squash? false}} opts
```
zonotope, line 40: If we are going to control the behavior of these functions through options maps, then we need to have an explicit options validation step. It's really easy for a user to use `{:preview true}` instead of `{:preview? true}` and perform some destructive process that they didn't intend.

This is part of the reason why I don't like options maps. I would rather have separate, explicit functions for all the legal options combinations we accept (like `preview-merge` ...). Those functions could share internal helper functions for shared functionality. This makes the api explicit, reduces the complexity of both the code and documentation, and makes it obvious what is and is not supported from what functions are available.

src/fluree/db/merge/branch.cljc
```diff
@@ -0,0 +1,121 @@
...
+(defn- same-commit?
+  "Check if two databases are at the same commit."
+  [source-db target-db]
+  (= (get-in source-db [:commit :id])
+     (get-in target-db [:commit :id])))
+
+(defn- branch-origin
+  "Get the commit ID a branch was created from."
+  [branch-info]
+  (or (get-in branch-info [:created-from "f:commit" "@id"])      ; nameservice expanded
```
zonotope, line 21: i think we should operate on a single format internally, and convert between external representations as necessary. Checking all these cases whenever we want to find information about a branch seems like a source of future bugs.

src/fluree/db/merge/branch.cljc
```diff
@@ -0,0 +1,121 @@
...
+    (= source-origin (branch-origin target-branch-info))))
+
+(defn- get-commit-chain
+  "Returns vector of commit maps from genesis to head for a db."
+  [conn db]
+  (go-try
+    (let [commit-catalog (:commit-catalog conn)
+          latest-expanded (<? (merge-commit/expand-latest-commit conn db))
+          error-ch (async/chan)
+          tuples (commit-storage/trace-commits commit-catalog latest-expanded 0 error-ch)]
```
zonotope, line 44: This doesn't handle any errors as nothing ever takes anything off of the error channel. It looks like this code will just hang, possibly locking the whole jvm process, if an error occurs.

docs/branch-operations.md
```diff
@@ -0,0 +1,540 @@
...
+##### Fast-forward Only
+```clojure
+(fluree/merge! conn "ledger:feature" "ledger:main"
+  {:ff :only})  ; Fail if fast-forward not possible
+```
+
+#### Response
+
+```clojure
+{:status :success  ; or :error
```
dpetran, line 165: I feel like the return should just be the same as a regular commit - return the db with the new HEAD.

docs/branch-operations.md
```diff
@@ -0,0 +1,540 @@
...
+
+### 3. RESET - Safe Rollback or Hard Reset
+
+Resets a branch to a previous state.
+
+#### Implementation Status
+
+| Mode | Status | Description |
+|------|--------|-------------|
+| Safe mode (`:mode :safe`) | ✅ Implemented | Creates a revert commit to target state |
```
dpetran, line 193: I think `revert` is a distinct operation from `reset`, I would vote we move this to a different api and leave `reset` unimplemented for now.

docs/branch-operations.md
```diff
@@ -0,0 +1,540 @@
...
+| Hard mode (`:mode :hard`) | ❌ Not Implemented | Will move branch pointer (rewrite history) |
+| Preview (`:preview? true`) | ✅ Implemented | Dry-run to see what would happen |
```
dpetran, line 195: Just a thought, but if we're copying `git`'s api we might want to call this `dry-run?` instead : )

docs/branch-operations.md
```diff
@@ -0,0 +1,540 @@
...
+Before performing branch operations, it's helpful to understand how branches relate:
+
+```clojure
+;; Check relationship between branches
+(def divergence @(fluree/branch-divergence conn "ledger:feature" "ledger:main"))
+
+;; Returns:
+{:common-ancestor "fluree:commit:sha256:..."
+ :can-fast-forward true  ; or false
+ :fast-forward-direction :branch1-to-branch2}  ; or :branch2-to-branch1, nil
```
dpetran, line 96: I feel like this datastructure can be made more useful:

```clojure
{:common-ancestor <lca>
 :target [<commita> <commitb>] ;; commits since lca
 :source [<commitc> <commitd> <commite>]} ;; commits since lca
```

`can-fast-forward` is implicit - only possible when `:target` is empty. This way we don't have to try to decipher what the direction is by parsing the `:fast-forward-direction` and we have the information we need to check for conflicts.

PR #1096, opened 2025-08-21
Add foundational branching support to Fluree

bplatz:
## Summary

This PR introduces foundational branching capabilities to Fluree, enabling parallel database development through branch isolation. This establishes the infrastructure for Git-like version control workflows, allowing multiple branches to coexist and evolve independently.

This PR sits on top of https://github.com/fluree/db/pull/1088 (branch and time travel naming)
This is a prerequisite to https://github.com/fluree/db/pull/1095 (merge, rebase, reset branch operations)
This is also a prerequisite to https://github.com/fluree/db/pull/1102 (cuckoo filter for safe index cleanup across branches)

### 1. Branch-Aware API

All Fluree APIs now support branch notation using the `:` separator:
```clojure
;; Load specific branches
@(fluree/load conn "mydb:main")
@(fluree/load conn "mydb:feature")

;; Write to specific branches
@(fluree/insert! conn "mydb:feature" {...})
@(fluree/update! conn "mydb:main" {...})

;; Query from branch-specific databases
(def feature-db @(fluree/load conn "mydb:feature"))
@(fluree/query feature-db {...})
```

### 2. Branch Creation

Create new branches from existing branches:
```clojure
;; Create a feature branch from main
@(fluree/create-branch! conn "mydb:feature" "mydb:main")

;; Create another branch from feature
@(fluree/create-branch! conn "mydb:feature-2" "mydb:feature")
```

### 3. Branch Isolation

Each branch maintains its own:
- Commit history
- Current state
- Independent evolution

Changes to one branch don't affect others until explicitly merged (merge operations coming in future PRs).

### 4. Time Travel with SHA Support

Navigate to specific commits using SHA:
```clojure
;; Time travel using SHA in branch URIs
@(fluree/load conn "mydb:branch@sha256:abc123...")
```

### Architecture Changes
- **Branch notation**: Standardized on `:` separator (e.g., `ledger:branch`)
- **Default branch**: All ledgers default to `main` branch
- **Nameservice**: Branch metadata stored in dedicated directory structure
- **Ledger validation**: Ledger names cannot contain `:` to avoid conflicts with branch notation

### Key Components Modified
- Connection and ledger loading made branch-aware
- API functions updated to parse and handle branch specifications
- Nameservice updated to store branch metadata separately
- File and memory storage adapted for branch support

## Testing

Tests added for:
- Branch creation and isolation
- Branch-aware API operations
- SHA-based time travel
- File and memory storage compatibility

Run tests with:
```bash
make cljtest
```

## Breaking Changes

- **Ledger naming**: Ledger names can no longer contain `:` character (reserved for branch notation)
- **Branch specification**: APIs now expect `ledger:branch` format where branch is specified

## Migration

- Existing ledgers work unchanged - they default to the `main` branch
- No data migration required
- Existing code continues to work by implicitly using the `main` branch

## What This Enables (Future Work)

This foundational work enables future PRs to add:
- Branch merge operations (rebase, squash, fast-forward)
- Branch divergence analysis
- Reset and rollback operations
- Conflict detection and resolution
- Cherry-pick functionality
- Branch protection rules

### Creating a Feature Branch Workflow
```clojure
;; Start with main branch
@(fluree/create conn "myapp" {})
@(fluree/insert! conn "myapp:main" 
  {"@context" {"ex" "http://example.org/"}
   "@id" "ex:product1"
   "ex:name" "Widget"
   "ex:price" 100})

;; Create feature branch for price update
@(fluree/create-branch! conn "myapp:update-prices" "myapp:main")

;; Work on feature branch
@(fluree/update! conn "myapp:update-prices"
  {"@context" {"ex" "http://example.org/"}
   "where" {"@id" "ex:product1"}
   "delete" {"ex:price" 100}
   "insert" {"ex:price" 110}})

;; Main branch remains unchanged
(def main-db @(fluree/load conn "myapp:main"))
(def main-price @(fluree/query main-db 
  {"select" "?price" 
   "where" {"@id" "ex:product1" "ex:price" "?price"}}))
;; Returns: 100

;; Feature branch has the update
(def feature-db @(fluree/load conn "myapp:update-prices"))
(def feature-price @(fluree/query feature-db
  {"select" "?price"
   "where" {"@id" "ex:product1" "ex:price" "?price"}}))
;; Returns: 110
```

## Next PR

The next PR will add merge operations (rebase, reset) to integrate changes between branches, building on this foundational branching support.


src/fluree/db/api/branch.cljc
```diff
@@ -0,0 +1,201 @@
...
+            source-commit-map (:commit source-db)
+            compact-commit (-> source-commit-map
+                               commit-data/->json-ld
+                               (assoc "alias" new-branch-spec
+                                      "branch" new-branch
+                                      "branchMetadata" branch-metadata))
+
+            ;; Publish to nameservice
+            primary-publisher (:primary-publisher conn)
+            _ (when primary-publisher
```
dpetran, line 57: I don't think it's possible to _not_ have a primary publisher, this condition is not necessary.

src/fluree/db/api/branch.cljc
```diff
@@ -0,0 +1,201 @@
...
+
+      ;; Load source ledger to get its current commit
+      (let [source-ledger (<? (connection/load-ledger conn from-branch-spec))
+            source-db (ledger/current-db source-ledger)
+            ;; Prefer commit ID (sha256 URI) for lineage as it's consistent across storage backends
+            source-commit-id (or from-commit (get-in source-db [:commit :id]))
+
+            ;; Create branch metadata
+            created-at (util/current-time-iso)
+            branch-metadata {:created-at created-at
```
dpetran: Just a consideration, but I think we should prefer flatter data structures for metadata, just so we don't have to do extra index scans to get them back:
```
  "f:createdAt" <>
  "f:sourceBranch" <>
  "f:sourceCommit" <>
```
And then have it live at the top level of the commit.

src/fluree/db/api/branch.cljc
dpetran: Also, should we attempt to publish to secondary publishers? We've got the handy `nameservice/publish-to-all` function for that.

src/fluree/db/api/branch.cljc
```diff
@@ -0,0 +1,201 @@
...
+  
+  Parameters:
+    conn - Connection object
+    ledger-alias - Ledger alias string (without branch)
+    
+  Returns a vector of branch names."
+  [conn ledger-alias]
+  (go-try
+    (log/info "list-branches for ledger:" ledger-alias)
+    (if-some [primary-publisher (:primary-publisher conn)]
```
dpetran, line 96: `primary-publisher` should always be present.

src/fluree/db/api/branch.cljc
```diff
@@ -0,0 +1,201 @@
...
+                            :let [;; The ledger field is an object with @id in the nameservice records
+                                  ledger-obj (get record "f:ledger")
+                                  ledger-name (if (map? ledger-obj)
+                                                (get ledger-obj "@id")
+                                                ledger-obj)
+                                  branch-name (get record "f:branch")]
+                            :when (and (= ledger-name ledger-alias) branch-name)]
+                        branch-name))]
+        (log/info "Found branches:" branches "for ledger:" ledger-alias)
+        (vec branches))
```
dpetran: We can skip the `(vec branches)` call by just not turning it into a lazy seq to begin with. I'd prefer something like this:
```
(defn same-ledger?
  [ns-record]
  (let [;; The ledger field is an object with @id in the nameservice records
        ledger-obj (get record "f:ledger")
        ledger-name (if (map? ledger-obj)
                      (get ledger-obj const/iri-id)
                      ledger-obj)
       branch-name (get record "f:branch")]
    (= ledger-name ledger-alias)))

;; or maybe combine into a transducer if we need to
(->> (<? (nameservice/all-records primary-publisher))
     (filter same-ledger?)
     (map #(get % "f:branch"))
     (distinct))

File: src/fluree/db/api/branch.cljc [hunk 5]
```diff
@@ -0,0 +1,201 @@
...
+      ;; Get all nameservice records and filter for this ledger's branches
+      (let [_ (log/info "Getting all nameservice records...")
+            records (<? (nameservice/all-records primary-publisher))
+            _ (log/info "Got" (count records) "nameservice records")
+            ;; Filter for this ledger's branches
+            branches (distinct
+                      (for [record records
+                            :let [;; The ledger field is an object with @id in the nameservice records
+                                  ledger-obj (get record "f:ledger")
+                                  ledger-name (if (map? ledger-obj)
```
dpetran: Does `ledger-obj` not have a consistent structure? I would expect that it would always be a map, or always not.

src/fluree/db/api/branch.cljc
```diff
@@ -0,0 +1,201 @@
...
+                                                (get ledger-obj "@id")
+                                                ledger-obj)
+                                  branch-name (get record "f:branch")]
+                            :when (and (= ledger-name ledger-alias) branch-name)]
```
dpetran: I think `f:branch` is always written in the ns-record, so we won't need to check if it exists.

src/fluree/db/connection.cljc
```diff
@@ -310,29 +310,38 @@
...
-            combined-alias  (commit->ledger-alias conn address expanded-commit)
 
             {:keys [did indexing]} (parse-ledger-options conn {})
-            ledger (ledger/instantiate combined-alias address commit-catalog index-catalog
-                                       primary-publisher secondary-publishers indexing did expanded-commit)]
-        (ns-subscribe/subscribe-ledger conn combined-alias)
+            ;; Extract the alias from the nameservice record, or use the address as the alias
+            ;; (for branch ledgers, the address IS the alias like "test-ledger:feature")
+            ledger-alias (or (get-first-value ns-record const/iri-alias)
+                             ledger-address)
```
dpetran: A `ledger-alias` and a `ledger-address` are two different things and none of the ledger parsing functions will properly handle a `ledger-address`. 

Also, looking at `fluree.db.nameservice.storage/ns-record`, we don't use the `const/iri-alias` at all (`const/iri-alias` is only ever used in commits), so the `ns-record` lookup will never return anything. You'd need to use `const/iri-id` to get the ledger alias from the `ns-record`.

src/fluree/db/ledger.cljc
```diff
@@ -120,27 +100,30 @@
...
   "Creates a new ledger, optionally bootstraps it as permissioned or with default
   context."
   [combined-alias ledger-address commit-catalog index-catalog primary-publisher secondary-publishers
-   indexing-opts did latest-commit]
+   indexing-opts did latest-commit & [branch-metadata]]
   (let [;; Parse ledger name and branch from combined alias
-        [_ branch] (if (str/includes? combined-alias ":")
-                     (str/split combined-alias #":" 2)
-                     [combined-alias "main"])
+        branch (or (util.ledger/ledger-branch combined-alias) "main")
```
dpetran: The alias here is always a normalized alias with a branch, so we don't need to use the `or` fallback.

src/fluree/db/ledger.cljc
```diff
@@ -185,14 +168,40 @@
...
+
+;; Branch operations are now handled at the connection/nameservice level
+;; Each branch is a separate ledger object
+
+(defn branch-info
+  "Returns detailed information about this ledger's branch"
+  [{:keys [primary-publisher alias state] :as _ledger}]
+  (go-try
+    ;; Extract branch from alias
+    (let [current-branch (or (util.ledger/ledger-branch alias) "main")]
```
dpetran, line 171: Same deal here, a Ledger record will always have a normalized alias at the `:alias` key, so we don't need to fall back here.

src/fluree/db/ledger.cljc
```diff
@@ -185,14 +168,40 @@
...
+      ;; Get nameservice record for branch if available, otherwise get from state
+      (if primary-publisher
```
dpetran, line 173: Ok, I had to try it out since I've seen this check in so many places. Turns out you _can_ create a Connection without a `primary-publisher`, but you cannot `create` a ledger - it throws an error, nor can you `load` any ledgers, so practically speaking this isn't possible. We should add some connection validation though, so you get a nicer error than `"No implementation of method: :publishing-address of protocol: #'fluree.db.nameservice/Publisher found for class: nil"` :D

src/fluree/db/nameservice/storage.cljc
```diff
@@ -18,34 +18,72 @@
...
+
+            ;; Preserve existing branch metadata if not provided in this publish
+            preserved-metadata (if (and existing-record (not branch-metadata))
+                                 {:created-at (get existing-record "f:createdAt")
+                                  :created-from (get existing-record "f:createdFrom")
+                                  :protected (get existing-record "f:protected")
+                                  :description (get existing-record "f:description")}
+                                 branch-metadata)
+
+            ;; Validation logic
```
dpetran: Let's move this validation logic into its own function.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -18,34 +18,72 @@
...
+                  ;; Allow same t for updates (e.g., index updates)
+                  (< t-value existing-t)
+                  (throw (ex-info (str "Cannot publish commit with t=" t-value
+                                       " - current HEAD is at t=" existing-t)
+                                  {:status 409 :error :db/invalid-commit-sequence
+                                   :alias ledger-alias :new-t t-value :existing-t existing-t}))))
+
+            ns-metadata    (ns-record ledger-alias commit-address t-value index-address preserved-metadata)
+            record-bytes   (json/stringify-UTF8 ns-metadata)
+            filename       (local-filename ledger-alias)]
```
dpetran: `filename` here supposes that we're writing to a file, but we don't actually know what storage backend we're using here. The `storage/write-bytes` api calls it `path`, we should probably keep it consistent.

dpetran: Also `local-path` instead of `local-filename`.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -77,29 +115,47 @@
...
+                                     (when-let [record-bytes (<? (storage/read-bytes store file-path))]
+                                       (let [record (json/parse record-bytes false)]
+                                         (async/>! out-ch record)))
+                                     #?(:clj (catch Exception e
+                                               (log/error e "Error reading/parsing file:" file-path))
+                                        :cljs (catch :default e
+                                                (log/error e "Error reading/parsing file:" file-path))))
+                                   (async/close! out-ch)))
+                ;; Set up pipeline with parallelism
+                _ (async/pipeline-async 10 ; parallelism level
```
dpetran: This can go outside the `let` block since we don't need the return value.

src/fluree/db/ledger.cljc
```diff
@@ -9,70 +9,51 @@
...
-  ([ledger branch]
-   (if-let [branch-meta (get-branch-meta ledger branch)] ; if branch is nil, will return default
-     (branch/current-db branch-meta)
-     (throw (ex-info (str "Invalid branch: " branch " is not one of:"
-                          (available-branches ledger))
-                     {:status 400, :error :db/invalid-branch})))))
+  "Returns the current database for this ledger.
+  Since each ledger now represents a single branch, no branch parameter is needed."
+  [ledger]
+  (when-let [state (:state ledger)]
```
dpetran: `:state` is always present on the Ledger record.

src/fluree/db/ledger.cljc
```diff
@@ -9,70 +9,51 @@
...
-                            "Did it just get deleted? Branches that exist are: " (keys (:branches @state)))
-                       {:status 400, :error :db/invalid-branch})))
-     (-> branch-meta
-         (branch/update-commit! db index-files-ch)
-         branch/current-db))))
+  ([ledger db]
+   (update-commit! ledger db nil))
+  ([{:keys [state alias] :as _ledger} db index-files-ch]
+   (log/debug "Attempting to update ledger:" alias "with new commit to t" (:t db))
+   (when-not state
```
dpetran: Ledgers always have state.

src/fluree/db/ledger.cljc
```diff
@@ -9,70 +9,51 @@
...
-      :flakes  flakes
-      :commit  commit})))
+  "Returns current commit metadata for this ledger"
+  [{:keys [address alias state]}]
+  (when state
+    (let [current-db  (branch/current-db @state)
+          {:keys [commit stats t]} current-db
+          {:keys [size flakes]} stats
+          ;; Extract branch from alias
+          branch (or (util.ledger/ledger-branch alias) "main")]
```
dpetran: Always have a normalized alias on a ledger.

src/fluree/db/ledger.cljc
```diff
@@ -9,70 +9,51 @@
...
-     {:address address
-      :alias   alias
-      :branch  branch
-      :t       t
-      :size    size
-      :flakes  flakes
-      :commit  commit})))
+  "Returns current commit metadata for this ledger"
+  [{:keys [address alias state]}]
+  (when state
```
dpetran: Always have state on a ledger.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -18,34 +18,72 @@
...
+      (let [;; Extract data from compact JSON-LD format (both genesis and regular commits now use this)
+            ledger-alias   (get data "alias")  ;; Already includes :branch
+            commit-address (get data "address")
+            t-value        (get-in data ["data" "t"])
+            index-address  (get-in data ["index" "address"])
+            branch-metadata (get data "branchMetadata")
+
+            ;; Check if this branch already exists
+            existing-record (<? (nameservice/lookup this ledger-alias))
+            existing-t      (when existing-record (get existing-record "f:t"))
```
dpetran: This `when` guard isn't necessary, `get` can handle `nil` values for the map.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -77,29 +115,47 @@
...
-        ;; Fallback for stores that don't support ListableStore
-        (do
-          (log/warn "Storage backend does not support RecursiveListableStore protocol")
+      ;; Read all JSON files from ns@v1 directory structure
+      (let [base-dir "ns@v1/"
+            all-files (<? (storage/list-paths-recursive store base-dir))]
+        ;; Use pipeline for parallel reading and parsing
+        (if (seq all-files)
+          (let [input-ch (async/chan 100 (filter #(str/ends-with? % ".json")))
+                _ (async/onto-chan! input-ch all-files)
```
dpetran: We can move this out of the `let` since we don't use the return value.

src/fluree/db/transact.cljc
```diff
@@ -205,10 +207,21 @@
 
 (defn publish-commit
   "Publishes commit to all nameservices registered with the ledger."
-  [{:keys [primary-publisher secondary-publishers] :as _ledger} commit-jsonld]
+  [{:keys [primary-publisher secondary-publishers alias] :as _ledger} commit-jsonld]
   (go-try
-    (let [result (<? (nameservice/publish primary-publisher commit-jsonld))]
-      (nameservice/publish-to-all commit-jsonld secondary-publishers)
+    (let [existing-record (when primary-publisher
```
dpetran: Always have `primary-publisher`.

src/fluree/db/transact.cljc
```diff
@@ -205,10 +207,21 @@
...
+                            (<? (nameservice/lookup primary-publisher alias)))
+          branch-metadata (when existing-record
```
dpetran: And the `primary-publisher` should always have an existing record for the ledger, I think `lookup` will throw if it doesn't find anything.

src/fluree/db/transact.cljc
```diff
@@ -205,10 +207,21 @@
...
+                            {:created-at (get existing-record "f:createdAt")
+                             :created-from (get existing-record "f:createdFrom")
+                             :protected (get existing-record "f:protected")
+                             :description (get existing-record "f:description")})
+          ;; Add branch metadata to commit if it exists
+          commit-with-metadata (if branch-metadata
+                                 (assoc commit-jsonld "branchMetadata" branch-metadata)
```
dpetran: It feels weird to add information to the commit that isn't actually persisted on the commit - do we need to do this when we publish, since it's already going to live on the ns-record?

test/fluree/db/branch_test.clj
```diff
@@ -0,0 +1,280 @@
...
+                results @(fluree/query main-db {"@context" {"ex" "http://example.org/"}
+                                                "select" "?name"
+                                                "where" {"@id" "ex:alice" "ex:name" "?name"}})]
+            (is (= ["Alice"] results) "Should find Alice in main branch")))
+
+        (testing "Create feature branch from main"
+          (let [branch-result @(fluree/create-branch! conn "test-ledger:feature-branch" "test-ledger:main")]
+            (is (= "feature-branch" (:name branch-result)))
+            (is (= "main" (get-in branch-result [:created-from :branch])))
+            (is (some? (:created-at branch-result)))
```
dpetran: We can use `with-redefs` to make `current-time-iso` deterministic, I prefer that approach to just checking that it's present.

test/fluree/db/branch_test.clj
```diff
@@ -0,0 +1,280 @@
...
+                                                         "select" "?name"
+                                                         "where" {"@id" "?id" "ex:name" "?name"}})]
+                (is (= ["Alice"] main-results) "Main branch should only have Alice"))
+
+              ;; Feature branch should have both Alice and Bob
+              (let [feature-db @(fluree/load conn "test-ledger:feature-branch")
+                    feature-results @(fluree/query feature-db {"@context" {"ex" "http://example.org/"}
+                                                               "select" "?name"
+                                                               "where" {"@id" "?id" "ex:name" "?name"}})]
+                (is (= (set feature-results) #{"Alice" "Bob"})
```
dpetran: This one has the expect/actual args backwards, and the results should be deterministic so no need for the `set` call.

test/fluree/db/branch_test.clj
```diff
@@ -0,0 +1,280 @@
...
+                (is (= "feature-branch" (:name feature-info)))
+                (is (some? (:head feature-info)))
+                (is (some? (:created-at feature-info)))
+                (is (= "main" (get-in feature-info [:created-from "f:branch"]))))))
+
+          (testing "Branch deletion"
+            @(fluree/delete-branch! conn "test-ledger:feature-branch")
+            ;; Verify branch is deleted by attempting to load it
+            (let [result @(fluree/load conn "test-ledger:feature-branch")]
+              (is (util/exception? result) "Should return an exception when loading deleted branch")
```
dpetran, line 76: Do we get a pretty error here? I'd prefer to at least extract the message and test for it explicitly.

test/fluree/db/branch_test.clj
```diff
@@ -0,0 +1,280 @@
...
+              (when (util/exception? result)
+                (is (= :db/unknown-ledger (:error (ex-data result))) "Should have correct error code"))))
+
+          (testing "Cannot delete main branch"
+            (let [result @(fluree/delete-branch! conn "test-ledger:main")]
+              (is (util/exception? result))
+              (is (re-find #"Cannot delete the main branch" (ex-message result))))))
```
dpetran: We should perhaps mention in the error that users can use the `drop` api to get rid of it.

test/fluree/db_test.cljc
```diff
@@ -1075,68 +1075,81 @@
...
+                  (count (async/<!! (fs/list-files (str primary-path "/" ledger-alias "/txn"))))))
+           ;; Each transaction creates one data file and one commit file
+           ;; But the initial creation doesn't create a data file (just the commit)
+           ;; So we have: 1 commit (initial) + 3 * 2 (data + commit for each tx) = 7 files
+           ;; But actually, it seems we're getting 5 files, which might be:
+           ;; 1 initial commit + 2 files per tx (data and commit) = 1 + 3*2 = 7 
+           ;; Let's check what we're actually getting
+           (let [commit-files (async/<!! (fs/list-files (str primary-path "/" ledger-alias "/commit")))]
+             ;; For now, just check we have some commit files
+             (is (pos? (count commit-files))
```
dpetran: I know it's fiddly, but I think we should have concrete expectations on how many files there are after we perform operations.

src/fluree/db/api/branch.cljc
zonotope: We probably should publish to every listed name service.

src/fluree/db/api/branch.cljc
```diff
@@ -0,0 +1,201 @@
...
+          ;; Get branch info to check protection status
+          branch-info (<? (ledger/branch-info ledger))
+          _ (when (:protected branch-info)
+              (throw (ex-info (str "Cannot delete protected branch: " branch)
+                              {:status 400 :error :db/cannot-delete-protected-branch})))
+          ;; Now delete the branch from nameservice
+          primary-publisher (:primary-publisher conn)]
+      (if primary-publisher
+        (do
+          (<? (nameservice/retract primary-publisher branch-spec))
```
zonotope: This also ignores secondary publishers.

src/fluree/db/api/branch.cljc
```diff
@@ -0,0 +1,201 @@
...
+  [conn old-branch-spec new-branch-spec]
+  (go-try
+    (let [[old-ledger-id old-branch] (util.ledger/ledger-parts old-branch-spec)
+          [new-ledger-id new-branch] (util.ledger/ledger-parts new-branch-spec)]
+
+      (when (not= old-ledger-id new-ledger-id)
+        (throw (ex-info "Cannot rename branch across different ledgers"
+                        {:status 400 :error :db/invalid-branch-operation})))
+
+      (when (or (= old-branch "main") (nil? old-branch))
```
zonotope: I think this check deserves to be it's own function called `main-branch?` or similar.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -77,29 +115,47 @@
...
+                output-ch (async/chan 100)
+                ;; Function to read and parse a single file
+                read-and-parse (fn [file-path out-ch]
+                                 (go
+                                   (try
```
zonotope: Why didn't you use `try*` for this?

src/fluree/db/nameservice/storage.cljc
```diff
@@ -77,29 +115,47 @@
...
+                                     (when-let [record-bytes (<? (storage/read-bytes store file-path))]
+                                       (let [record (json/parse record-bytes false)]
+                                         (async/>! out-ch record)))
+                                     #?(:clj (catch Exception e
```
zonotope: I think any errors should be propagated to the calling function. A user has to be watching the logs to know something went wrong as opposed to there being less or no records.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -77,29 +115,47 @@
...
+                                        :cljs (catch :default e
+                                                (log/error e "Error reading/parsing file:" file-path))))
+                                   (async/close! out-ch)))
+                ;; Set up pipeline with parallelism
+                _ (async/pipeline-async 10 ; parallelism level
+                                        output-ch
+                                        read-and-parse
+                                        input-ch)]
+            ;; Collect all results into a vector
+            (<? (async/into [] output-ch)))
```
zonotope: This `<?` does nothing here as there will never be an exception on the channel returned by `async/into`.

zonotope: There should be an error channel or some other construct to propagate errors to the callers.

src/fluree/db/api/branch.cljc
bplatz: Addressed in 555a577dbfdc8dbffafd473e1b83233bcf20b858

src/fluree/db/api/branch.cljc
bplatz: Good idea! Addressed in 555a577dbfdc8dbffafd473e1b83233bcf20b858

src/fluree/db/api/branch.cljc
bplatz: Addressed in b57b772330e36ad0430955fa2173a6e1a18b3a17

src/fluree/db/api/branch.cljc
bplatz: Addressed in b57b772330e36ad0430955fa2173a6e1a18b3a17

src/fluree/db/connection.cljc
bplatz: This was addressed in 'main' and merged here.

src/fluree/db/nameservice/storage.cljc
bplatz: Addressed in eb0a9720687ee2bfd881755e4b1cd2674b3ab34b

src/fluree/db/nameservice/storage.cljc
bplatz: Addressed in eb0a9720687ee2bfd881755e4b1cd2674b3ab34b

src/fluree/db/nameservice/storage.cljc
bplatz: This had already been addressed previously

src/fluree/db/nameservice/storage.cljc
bplatz: This code had been removed previously.

src/fluree/db/nameservice/storage.cljc
bplatz: This code had been removed previously.

src/fluree/db/transact.cljc
bplatz: Addressed in a161d37baca16ba5caefc3e0f6234f1b39351949

src/fluree/db/ledger.cljc
bplatz: Addressed in a161d37baca16ba5caefc3e0f6234f1b39351949

src/fluree/db/ledger.cljc
bplatz: Addressed in a161d37baca16ba5caefc3e0f6234f1b39351949

src/fluree/db/ledger.cljc
bplatz: Addressed in a161d37baca16ba5caefc3e0f6234f1b39351949

src/fluree/db/ledger.cljc
bplatz: Addressed in a161d37baca16ba5caefc3e0f6234f1b39351949

test/fluree/db_test.cljc
bplatz: This test had previously been updated.

test/fluree/db/branch_test.clj
bplatz: Addressed in 86bc49b85873523bfd5665305ae200602a6a64d1

test/fluree/db/branch_test.clj
bplatz: Addressed in 86bc49b85873523bfd5665305ae200602a6a64d1

test/fluree/db/branch_test.clj
bplatz: Addressed in 86bc49b85873523bfd5665305ae200602a6a64d1

test/fluree/db/branch_test.clj
bplatz: Addressed in 86bc49b85873523bfd5665305ae200602a6a64d1

src/fluree/db/api/branch.cljc
bplatz: Addressed in 555a577db

src/fluree/db/nameservice/storage.cljc
bplatz: This had been addressed previously

src/fluree/db/nameservice/storage.cljc
bplatz: This had already been refactored previously.

src/fluree/db/nameservice/storage.cljc
bplatz: This had already been refactored previously.

src/fluree/db/api/branch.cljc
bplatz: Addressed in 146c07f2b68e93971d8ed41eb00db88cefaf7cdd

src/fluree/db/api/branch.cljc
bplatz: primary-publisher should be optional - I think there may be parts in the legacy code where it still creates an issue but trying to extract it a bit.

I also think there should only be one nameservice, and optionally your primary nameservice might push/notify to other (secondary) name services... but that will need some more change too.

Concrete use case, we'd like to use a fluree/db as a nameservice - but if so it wouldn't make sense for the nameservice to have a nameservice.

src/fluree/db/ledger.cljc
bplatz: Refer to prior nameservice comments.

src/fluree/db/transact.cljc
bplatz: Refer to prior nameservice comments.

src/fluree/db/transact.cljc
bplatz: Refer to prior nameservice comments.

src/fluree/db/connection.cljc
```diff
@@ -273,9 +274,11 @@
     (loop [[nsv & r] (publishers conn)
            addrs     []]
       (if nsv
-        (if (<? (nameservice/published-ledger? nsv ledger-alias))
-          (recur r (conj addrs (<? (nameservice/publishing-address nsv ledger-alias))))
-          (recur r addrs))
+        (let [published? (<? (nameservice/published-ledger? nsv ledger-alias))]
+          (log/info "published-addresses: checking" ledger-alias "published?" published?)
```
zonotope, line 278: I don't think this should log at the info level.

src/fluree/db/connection.cljc
```diff
@@ -273,9 +274,11 @@
     (loop [[nsv & r] (publishers conn)
            addrs     []]
       (if nsv
-        (if (<? (nameservice/published-ledger? nsv ledger-alias))
-          (recur r (conj addrs (<? (nameservice/publishing-address nsv ledger-alias))))
-          (recur r addrs))
+        (let [published? (<? (nameservice/published-ledger? nsv ledger-alias))]
```
zonotope, line 277: I think the `nameservice/published-ledger?` call would go better in the `if` condition directly on line 279.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -20,98 +21,121 @@
...
 
-(defn get-index-t
-  [ns-record]
-  (get-in ns-record ["f:index" "f:t"] 0))
-
 (defn update-commit-address
-  [ns-record commit-address commit-t]
+  "Updates commit address if new t is greater than existing t.
+   For branch creation, also validates that branch doesn't already exist."
+  [ns-record ledger-alias commit-address commit-t is-branch-creation]
```
zonotope, line 47: I think this should be split into two different functions instead of including the `is-branch-creation` boolean option here.

src/fluree/db/nameservice/storage.cljc
```diff
@@ -20,98 +21,121 @@
...
+          path           (local-path ledger-alias)]
+      (log/debug "nameservice.storage/publish start" {:ledger ledger-alias :path path})
       (let [commit-address (get data "address")
             commit-t       (get-in data ["data" "t"])
             index-address  (get-in data ["index" "address"])
-            index-t        (get-in data ["index" "data" "t"])
+            ;; Extract metadata from incoming data
+            new-metadata   (util.branch/extract-branch-metadata data)
+            ;; Check if this is a branch creation (has source metadata)
+            is-branch-creation (and (:source-branch new-metadata) (:source-commit new-metadata))
```
zonotope, line 107: I think this is clearer as a separate `branch-creation?` function

src/fluree/db/api/branch.cljc
```diff
@@ -0,0 +1,196 @@
...
+
+            ;; Prepare commit for new branch with flat metadata fields
+            source-commit-map (:commit source-db)
+            compact-commit (-> source-commit-map
+                               commit-data/->json-ld
+                               (assoc "alias" new-branch-spec
+                                      "branch" new-branch)
+                               (util.branch/augment-commit-with-metadata metadata))
+
+            primary-publisher (:primary-publisher conn)
```
zonotope, line 53: This is not for this pr as I think we need an explicit way to connect to a nameservice and provide an address for it, but if the primary publisher (or any other publishers for that matter) should be optional, then it shouldn't be tied to the connection. We should be able to use the same connection to create multiple branches, some of which do have publishers configured, and some without.

src/fluree/db/api/branch.cljc
```diff
@@ -0,0 +1,196 @@
...
+                           "address:" (get compact-commit "address")
+                           "t:" (get-in compact-commit ["data" "t"]))
+                (<? (nameservice/publish primary-publisher compact-commit))
+                (log/debug "Published commit for" new-branch-spec)
+                ;; Also publish to secondary publishers asynchronously
+                (nameservice/publish-to-all compact-commit secondary-publishers))]
+
+        (util.branch/branch-creation-response new-branch metadata source-commit-id)))))
+
+(defn- same-ledger?
```
zonotope, line 68: This terminology is confusing to me. I had been using `ledger-alias` as the full name including the branch, and `ledger-name` as the name without the branch, and `ledger` as the catch all for all of a ledgers branches treated as a single unit.

The name of this function and the intermediate bindings used within it do not follow that model. I'm fine with changing my internal terminology, but I do think we should pick consistent terms for those concepts and stick to them.

PR #1099, opened 2025-08-25
Add OWL-Datalog reasoner: Horn fragment approximation of SROIQ(D)

bplatz:
## Summary

Implements OWL-Datalog reasoner, extending OWL 2 RL with additional constructs while maintaining Datalog compatibility. This provides a rule-materialization approach approximating parts of SROIQ(D) - more expressive than OWL 2 RL but incomplete for full OWL DL.

### Extensions Beyond OWL 2 RL

- **Complex intersections with restrictions**: Full support for equivalentClass definitions containing multiple restrictions in intersections
- **Unions compiled as rule alternatives**: Union classes generate separate Datalog rules for each disjunct
- **Property chains with inverses**: Supports chains including inverse properties in any position with double-inverse normalization
- **Forward entailment for hasValue/allValuesFrom**: Derives property values from class membership and types from universal restrictions
- **Multiple restrictions on same property**: Handles complex patterns like `Item ∩ ∃p.A ∩ ∃p.B ∩ ∃p.{v}`

### Technical Implementation

- Queries ontology with depth-6 traversal to capture nested structures
- Compiles OWL constructs to Datalog rules for forward-chaining inference
- Handles inline property chains in restrictions (e.g., `∃(R₁ ∘ R₂⁻).C`)
- Uses `util/sequential` to handle JSON-LD list collapsing for single-item chains

## Current Limitations

1. **Data property hasValue with typed literals**: Backward inference from typed literal values to class membership not yet implemented
2. **Property chains with separate axioms**: allValuesFrom restrictions don't work when property has chain axiom defined separately (works for inline chains)
3. **Qualified cardinalities**: Parsed but not reasoned over

## Testing

- 19 tests, 50 assertions, 0 failures
- Comprehensive test suites for core functionality, edge cases, and restriction features
- Known limitations documented with test cases

## Files Changed

- `src/fluree/db/reasoner/owl_datalog.cljc`: Core reasoner implementation
- `src/fluree/db/flake/reasoner.cljc`: Integration with reasoning framework
- `test/fluree/db/reasoner/owl_datalog_test.clj`: Core functionality tests
- `test/fluree/db/reasoner/owl_datalog_edge_cases_test.clj`: Complex scenario tests
- `test/fluree/db/reasoner/owl_datalog_restrictions_test.clj`: Restriction feature tests
- `docs/owl-datalog.md`: Comprehensive documentation

## Usage

```clojure
(def reasoned-db @(fluree/reason db :owl-datalog))
```

The reasoner generates materialized inferences stored alongside original data, providing fast query performance over derived triples.

## Documentation

See `docs/owl-datalog.md` for detailed documentation including examples, architecture details, and comparison with other reasoners.

### DIFF
--- a/docs/owl-datalog.md
+++ b/docs/owl-datalog.md
@@ -14,12 +14,12 @@ OWL 2 RL has limited support for complex class expressions in intersections. OWL
 
 ```turtle
 # Complex equivalentClass with multiple restrictions
-DrugFormulation ≡ Specification ∩ ∀(isMemberOf)⁻.Ingredient ∩ ∃hasUnit.{mg}
+ElectricVehicle ≡ Vehicle ∩ ∀hasPowerSource.{electricity} ∩ ∃hasRange.LongRange
 ```
 
 This generates rules for:
-- Forward entailment: If x is DrugFormulation → x has the required properties
-- Backward inference: If x meets all intersection criteria → x is DrugFormulation
+- Forward entailment: If x is ElectricVehicle → x has the required properties
+- Backward inference: If x meets all intersection criteria → x is ElectricVehicle
 
 ### Union Classes as Multiple Rules
 
@@ -27,12 +27,12 @@ While OWL 2 RL supports basic unions, OWL-Datalog compiles unions into multiple 
 
 ```turtle
 # Union in equivalentClass
-DrugTarget ≡ Protein ∪ Receptor ∪ Enzyme
+PaymentMethod ≡ CreditCard ∪ DebitCard ∪ DigitalWallet
 
 # Generates separate rules:
-# Rule 1: Protein(?x) → DrugTarget(?x)
-# Rule 2: Receptor(?x) → DrugTarget(?x) 
-# Rule 3: Enzyme(?x) → DrugTarget(?x)
+# Rule 1: CreditCard(?x) → PaymentMethod(?x)
+# Rule 2: DebitCard(?x) → PaymentMethod(?x) 
+# Rule 3: DigitalWallet(?x) → PaymentMethod(?x)
 ```
 
 ### Property Chains with Inverse Properties
@@ -51,20 +51,20 @@ ChainedClass ≡ ∃(hasParent ∘ hasChild⁻ ∘ hasSibling).Person
 
 #### hasValue Forward Entailment
 ```turtle
-# If x is KilogramMagnitude, infer hasUnit kg
-KilogramMagnitude ≡ Magnitude ∩ ∃hasUnit.{kg}
+# If x is KilogramMeasurement, infer hasUnit kg
+KilogramMeasurement ≡ Measurement ∩ ∃hasUnit.{kg}
 ```
 
 #### allValuesFrom Forward Entailment
 ```turtle
-# If x is SafeContainer and x contains y, then y is SafeItem
-SafeContainer ≡ Container ∩ ∀contains.SafeItem
+# If x is SecureFolder and x contains y, then y is SecureDocument
+SecureFolder ≡ Folder ∩ ∀contains.SecureDocument
 ```
 
 #### Multiple Restrictions on Same Property
 ```turtle
 # Multiple constraints on the same property
-ComplexProduct ≡ Item ∩ ∃hasProperty.PropertyA ∩ ∃hasProperty.PropertyB ∩ ∃hasProperty.{specificValue}
+LuxuryCar ≡ Car ∩ ∃hasFeature.LeatherSeats ∩ ∃hasFeature.NavigationSystem ∩ ∃hasFeature.{sunroof}
 ```
 
 ## Supported Constructs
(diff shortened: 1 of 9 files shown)


docs/owl-datalog.md
```diff
@@ -0,0 +1,331 @@
...
+
+```turtle
+# Forward entailment works:
+KilogramMeasurement(?x) → hasUnit(?x, kg)
+
+# Backward inference doesn't work:
+hasQualityScore(?x, "95"^^xsd:int) ↛ HighQuality(?x)
+```
+
+**Workaround**: Use untyped literals or object properties where possible.
```
dpetran: What is an "untyped literal" in this context?

src/fluree/db/util.cljc
```diff
@@ -254,9 +254,10 @@
 (defn of-type?
   "Returns true if the provided json-ld node is of the provided type."
   [jsonld rdf-type]
-  (->> jsonld
-       get-types
-       (some #(= % rdf-type))))
+  (let [types (get-types jsonld)]
+    (if (sequential? types)
```
dpetran: I think it would be simpler just to use `sequential` here:

```
(->> jsonld
       get-types
       sequential
       (some #(= % rdf-type)))
```

src/fluree/db/reasoner/owl_datalog.cljc
```diff
@@ -607,11 +1051,14 @@
 (defmethod to-datalog ::cax-eqc
   [_ inserts owl-statement all-rules]
   (let [c1 (util/get-id owl-statement) ;; the class which is the subject
+        equiv-class-val (get owl-statement const/iri-owl:equivalentClass)
+        unwrapped (util/unwrap-list equiv-class-val)
+        ;; Ensure unwrapped is always a sequence for group-by
+        unwrapped-seq (if (sequential? unwrapped) unwrapped [unwrapped])
```
dpetran: We have `util/sequential` for this, though I think it would be better for `unwrap-list` can be refactored to always return a sequence, since other callsites assume that it does.

src/fluree/db/reasoner/owl_datalog.cljc
```diff
@@ -607,11 +1051,14 @@
...
         ;; combine with all other equivalent classes for a set of 2+ total classes
         {:keys [classes intersection-of union-of one-of
                 has-value some-values all-values
-                max-cardinality max-qual-cardinality]} (->> (get owl-statement const/iri-owl:equivalentClass)
-                                                            util/unwrap-list
+                max-cardinality max-qual-cardinality]} (->> unwrapped-seq
```
dpetran: No need for the thread when we only have a single expression.

src/fluree/db/reasoner/owl_datalog.cljc
```diff
@@ -552,7 +976,28 @@
...
+                                                       acc**))
+                                                   acc*
+                                                   nested-unions))
+
+                                         :else acc*))
+                                     []
+                                     unions)
+           {:keys [has-value]} (group-by equiv-class-type unions)
            restrictions      (cond->> []
                                has-value (into (has-value-condition "?y" has-value)))
```
dpetran: I'm not sure why we need to call `into` here with an empty vector, we'll never add any data to the result of `has-value-condition` regardless of the value of `has-value`.

src/fluree/db/reasoner/owl_datalog.cljc
```diff
@@ -526,6 +888,66 @@
...
+                                                                   property "?y"}]
+                                                        "insert" {"@id"   "?y"
+                                                                  "@type" target-class}})]
+                                            (conj acc* [(str rule-class "(owl:allValuesFrom-forward-" property ")") rule]))
+                                          acc*)))
+                                    []
+                                    all-values)
+
+           ;; Generate forward entailment rules for hasValue in intersections
+           has-value-rules (reduce (fn [acc* has-val-restriction]
```
dpetran: I think this function would be more readable if we extract the `has-value-rules` and `all-values-rules` into their own functions.

test/fluree/db/reasoner/owl_datalog_test.clj
```diff
@@ -0,0 +1,401 @@
...
+                                                      "@type" "ex:Formulation"}}}
+
+          db-with-data @(fluree/update db-with-ontology instance-data)
+
+          ;; Apply owl-datalog reasoning
+          db-reasoned @(fluree/reason db-with-data :owl-datalog)]
+
+      (testing "Instance with all intersection components should be inferred as DrugProduct"
+        (is (= #{"ex:DrugProduct" "ex:ManufacturedItem" "ex:SubstanceDefinition"}
+               (set @(fluree/query db-reasoned
```
dpetran: Do we need to coerce the results into a set? By my read the output should be deterministic.

test/fluree/db/reasoner/owl_datalog_test.clj
```diff
@@ -0,0 +1,401 @@
...
+                         "insert"   [{"@id"        "ex:john"
+                                      "ex:hasChild" {"@id" "ex:mary"}}
+                                     {"@id"   "ex:mary"
+                                      "@type" "ex:Person"}]}
+
+          db-with-data @(fluree/update db-with-ontology instance-data)
+          db-reasoned @(fluree/reason db-with-data :owl-datalog)]
+
+      (testing "Individual with property pointing to correct type should be inferred as Parent"
+        (is (contains? (set @(fluree/query db-reasoned
```
dpetran: Same thing here, there should only be a single result type, so no need to put it in a set.

dpetran (approved):
:owl:

docs/owl-datalog.md
bplatz: Update doc to clarify.

PR #1100, opened 2025-08-25
Move blocking HTTP operations to thread pool

bplatz:
## Summary
- Replace async/go with async/thread in xhttp.cljc for blocking HTTP operations
- Add debug logging in critical I/O paths for S3 operations
- Prevents blocking I/O from monopolizing limited go thread pool

## Changes
- HTTP operations (post/get/put/delete) now use async/thread instead of async/go
- Added logging in storage, nameservice, and transact operations
- Should improve resource utilization in Lambda/serverless environments

### DIFF
--- a/src/fluree/db/nameservice/storage.cljc
+++ b/src/fluree/db/nameservice/storage.cljc
@@ -50,7 +50,10 @@
           ns-metadata    (ns-record ledger-alias branch commit-address t-value index-address)
           record-bytes   (json/stringify-UTF8 ns-metadata)
           filename       (local-filename ledger-alias branch)]
-      (storage/write-bytes store filename record-bytes)))
+      (log/debug "nameservice.storage/publish start" {:ledger ledger-alias :branch branch :filename filename})
+      (let [res (storage/write-bytes store filename record-bytes)]
+        (log/debug "nameservice.storage/publish enqueued" {:ledger ledger-alias :branch branch :filename filename})
+        res)))
 
   (retract [_ ledger-alias]
     (let [filename (local-filename ledger-alias)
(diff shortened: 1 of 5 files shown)


src/fluree/db/nameservice/storage.cljc
```diff
@@ -50,7 +50,10 @@
           ns-metadata    (ns-record ledger-alias branch commit-address t-value index-address)
           record-bytes   (json/stringify-UTF8 ns-metadata)
           filename       (local-filename ledger-alias branch)]
-      (storage/write-bytes store filename record-bytes)))
+      (log/debug "nameservice.storage/publish start" {:ledger ledger-alias :branch branch :filename filename})
```
zonotope, line 53: I've always had a dream to change all of our logging to logging data instead of English, so it would be easier to get insights from tools like [Riemann](https://riemann.io/) in line with the [recommendations from juxt](https://www.juxt.pro/blog/logging/) . I never made the change because of the low priority to effort ratio, but this log statement is so close that we might as well do it here. All we would have to do is change the first phrase of this log message to `::publish-start`.

zonotope: Kurt also did some open telemetry work that has similar ideas to Riemann, and he has a [pull request against db](https://github.com/fluree/db/pull/1034) to add a common context to log messages. I think we could use some of those ideas to automatically include information like ledger alias and branch to each message.

zonotope (approved):
I agree with the core changes here. It also might be time to at least gradually formalize our logging practices so we can get more insights from our logs using tools like Riemann (or something else).

src/fluree/db/nameservice/storage.cljc
bplatz: Kurt's PR was for OpenTelemetry - a little out of my zone but from what I understand it is the main standard and most major tools support it. I'm using AWS X-Ray in the serverless version (which supports Otel) and it is great for troubleshooting. They all expect JSON logs though, so EDN I don't thin will work if we want folks to be able to plug it into their tracing tools of choice.

PR #1101, opened 2025-08-25
Fix/upsert multiple multicardinal values

dpetran:
We were creating a new variable for each object, instead of for each subject+predicate pair, which lead to as each object for each subject+predicate pair generating a semantically identical triple pattern with a different var for each. This produced an overwhelming amount of data so that even with small cases (like the test case in this commit) upserts would take tens of minutes to complete or run out of memory.


zonotope (approved):
🥟

PR #1102, opened 2025-08-26
Add per-branch cuckoo filters for cross-branch-safe index GC

bplatz:
This sits on top of this PR: https://github.com/fluree/db/pull/1095
Which sits on top of this PR: https://github.com/fluree/db/pull/1096

All three PRs should be thought of as one package, as without this PR we might garbage collect index segments still relied on by other branches.

  ## Summary

  Introduces per-branch cuckoo filters to prevent deletion of index nodes still referenced by other branches during garbage collection. Uses 16-bit fingerprints from SHA-256 hashes with a chain design for dynamic growth. Integrates seamlessly with indexing and GC workflows.

  ## Problem

  - Index segments are shared across branches via content-addressing
  - GC could delete segments still needed by other branches
  - Need fast, memory-efficient way to check cross-branch usage

  ## Solution

  **Cuckoo filter implementation** (`fluree.db.indexer.cuckoo`):
  - 16-bit fingerprints extracted from first 2 bytes of decoded base32 SHA-256
  - Chain growth: 100K segment capacity per filter, adds new filters at 90% full
  - FNV-1a 32-bit hash ensures CLJ/CLJS compatibility
  - CBOR binary storage at `ledger/index/cuckoo/<branch>.cbor`

  **Integration**:
  - **Indexing**: Updates branch filter with new segments after index refresh
  - **GC**: Removes garbage from current branch filter, checks other branches before deletion
  - **Caching**: Loads other-branch filters once per GC run to minimize I/O

  ## Performance

  - **False positive rate**: ~0.012% (1 in ~8,200) with 16-bit fingerprints
  - **Memory efficiency**: ~4.6 bytes per segment with realistic hash distribution
  - **Scalability**: Chain design avoids expensive rebuilds; batch operations minimize overhead
  - **Filter size examples**:
    - 10GB database (~50K segments): ~224KB
    - 1TB database (~5M segments): ~22MB

  ## Documentation

  See [docs/cuckoo-filter-gc-strategy.md](docs/cuckoo-filter-gc-strategy.md) for detailed implementation notes.


bplatz:
> > See [docs/cuckoo-filter-gc-strategy.md](docs/cuckoo-filter-gc-strategy.md) for detailed implementation notes.
> 
> This link is broken.

Not really, it will work once the PR is merged and it is on the main branch... but the doc is part of the PR.

PR #1103, opened 2025-08-26
add time support for shacl lessThan, lessThanOrEqual constraints

dpetran:
We were trying to compare all values with `<`, `<=`, time values need to use `compare`.


zonotope (approved):
♠️

PR #1105, opened 2025-08-28
Enable index-only updates via nameservice notifications

bplatz:
## Summary
- Previously, the `notify` function only handled commit updates and ignored standalone index updates
- Added support for notifying ledgers of index updates without requiring new commits
- Leveraged nameservice records which track indexing state independently from commits

## Changes
- Extended `fluree.db.api/notify` to accept nameservice records (2-arity) containing index information
- Added `notify-index` function in ledger.cljc to apply index-only updates when appropriate
- Implemented unified notification logic that determines minimal update action (noop, index, commit, or stale)

## Key Benefits
- Branches can now receive updated indexes without new commits
- Minimizes I/O by only reading index files when addresses change
- Supports the new nameservice architecture where indexing state is tracked independently

## Test Plan
- [x] All existing tests pass
- [x] file-based-indexing-test specifically verified
- [x] transact tests verified

### DIFF
--- a/src/fluree/db/ledger.cljc
+++ b/src/fluree/db/ledger.cljc
@@ -1,6 +1,5 @@
 (ns fluree.db.ledger
   (:require [clojure.string :as str]
-            [fluree.db.async-db :as async-db]
             [fluree.db.branch :as branch]
             [fluree.db.commit.storage :as commit-storage]
             [fluree.db.constants :as const]
@@ -202,11 +201,8 @@
     (let [data      (-> db :commit :data)
           index-id  (<? (idx-address->idx-id index-catalog index-address))
           index-map (commit-data/new-index data index-id index-address
-                                           (select-keys root [:spot :post :opst :tspo]))
-          res       (dbproto/-index-update db index-map)]
-      (if (async-db/db? res)
-        (do (<? (async-db/deref-async res)) res)
-        (<? res)))))
+                                           (select-keys root [:spot :post :opst :tspo]))]
+      (<? (dbproto/-index-update db index-map)))))
 
 (defn- update-branch-with-index
   "Updates branch state with new index. If use-update-commit? is true,


src/fluree/db/ledger.cljc
```diff
@@ -117,6 +129,128 @@
...
+    (when (and index-node (nil? index-addr))
+      (log/warn "notify: nameservice record f:index present but missing @id" {:ledger-alias ledger-alias :index index-node}))
+    {:ledger-alias   ledger-alias
+     :branch         (or branch-val
+                         (second (str/split ledger-alias #":" 2)))
+     :ns-t           ns-t
+     :commit-address commit-addr
+     :index-address  index-addr}))
+
+(defn idx-address->idx-id
```
zonotope, line 187: I think this should be adapted to use the new `get-hash` protocol method introduced in https://github.com/fluree/db/pull/1111. Each storage type encodes the hash within the index address slightly differently, so this function won't work in all cases.

src/fluree/db/connection.cljc
```diff
@@ -100,31 +101,153 @@
...
+   :branch         (get-first-value expanded-commit const/iri-branch)
+   :ns-t           (-> expanded-commit
+                       (get-first const/iri-data)
+                       (get-first-value const/iri-fluree-t))
+   :commit-address (get-first-value expanded-commit const/iri-address)
+   :index-address  (-> expanded-commit
+                       (get-first const/iri-index)
+                       (get-first-value const/iri-id))})
+
+(defn notify
```
zonotope: I think this should be two functions: `notify-commit` and `notify-index`. Overloading the same argument for different concepts just adds complexity in my opinion. I think the same goes for the external api.

src/fluree/db/ledger.cljc
```diff
@@ -117,6 +130,101 @@
...
+          (log/debug "notify-index loaded root" {:root-t root-t :cur-t cur-t})
+          (cond
+            (flake/t-after? root-t cur-t)
+            (do (log/debug "notify-index: root ahead of current commit; marking stale"
+                           {:root-t root-t :cur-t cur-t})
+                ::stale)
+
+            (flake/t-before? root-t cur-t)
+            (if (some? cur-idx)
+              (do (log/debug "notify-index: root behind current commit; ignoring"
```
zonotope: What happens for a long-running index process where new commits come in before it has a chance to finish?

Say the current index is from t=4975, a new index kicks off at t=5000 but the database is large so it takes some time to finish. If 10 commits come in during before the index finishes, then it looks like this code will ignore the index from 5000 in favor of the index from 4975 because the current commit is t=5010.

src/fluree/db/ledger.cljc
```diff
@@ -117,6 +130,101 @@
...
+                      (swap! (:state branch-meta) branch/update-index updated-db))
+                    ::index-updated)))
+
+            :else
+            (let [data       (-> db :commit :data)
+                  index-id   (idx-address->idx-id index-address)
+                  index-map  (commit-data/new-index data index-id index-address
+                                                    (select-keys root [:spot :post :opst :tspo]))
+                  res        (dbproto/-index-update db index-map)
+                  updated-db (if (async-db/db? res)
```
zonotope: This should not be necessary. The need for it either points to a missing/incomplete protocol or other code in the system that isn't using the protocol correctly or at all. This pr is probably not the place to identify and fix the problem, but we should make note of it.

src/fluree/db/connection.cljc
zonotope: But, this is minor and we can revisit it later.

bplatz: @zonotope It is more a notify with a nameservice record vs a commit address at this point (it wasn't overload until you changed the latter to a 2-arity fn). 

Ideally the nameservice record notify would be the primary, and commit notify would be one which I'd see more use of with IPFS or web3 where you just have a SHA.

My preference would be 'notify' only uses nameservice records, and if you only have a SHA I would almost prefer to you use 'load' and not notify at all, but there might be a need for notify still at least for legacy support. If so, then I'd suggest we have 'notify-commit'. This would be a breaking change though for likely Nexus.

src/fluree/db/ledger.cljc
bplatz: Addressed in fd074871fa0296034e450462ca0232fe7f3d1d55

src/fluree/db/ledger.cljc
bplatz: Addressed in 1e6b8875d642b67424ce6a5d6205dc664925e4a7

src/fluree/db/ledger.cljc
bplatz: Addressed in 4a521d83e8f5844c54c00b282d5c5f661800fa5b

PR #1107, opened 2025-08-28
only use shacl flakes when rebuilding shapes

dpetran:
The build-shape-node function was overly eager, following every ref flake and constructing a node out of it.

For example, in this shape:
```
{"@id"   "ex:shape1",
 "@type" "sh:NodeShape",
 "sh:targetClass" {"@id" "ex:User"},
 "sh:property" [{"sh:path" {"@id" "ex:name"},
                 "sh:maxCount" 1}]}
```
When rebuilding, would produce something like this:
```
{#SID[1 "id"] #SID[101 "shape1]
 #SID[3 "type"] [#SID[5 "NodeShape"]]
 #SID[5 "targetClass"] [{#SID[1 "id"] #SID[101 "User"]
                         #SID[3 "type"] [#SID[106 "Class"]]
                         ,,, <any other flakes associated with ex:User>,,,}]
 #SID[5 "property"]
   [{#SID[1 "id"] #SID[24 "fdb1"]
     #SID[5 "path"] [{#SID[1 "id"] #SID[101 "name"]
                      #SID[3 "type"] [#SID[107 "Property"]]
                      ,,,<any other properties associated with ex:name>,,,}]
     #SID[5 "maxCount"] [1]}]}
```
And so on recursively until the hardcoded and arbitrary depth limit of 10.

Now we only recursively build shacl ref flakes, so we aren't pulling in any extraneous data (and charging the user fuel for it). I don't believe we need the depth limit any more, but I'm retaining for now as a safety.

### DIFF
--- a/src/fluree/db/json_ld/shacl.cljc
+++ b/src/fluree/db/json_ld/shacl.cljc
@@ -34,6 +34,19 @@
      #fluree/SID [5 "datatype"] [#fluree/SID [2 "string"]]
      #fluree/SID [5 "path"] [#fluree/SID [17 "email"]]}]})
 
+(defn property-ns-code
+  "Returns namespace code for the property of the flake"
+  [flake]
+  (-> flake
+      flake/p
+      iri/get-ns-code))
+
+(def ^:const shacl-ns-code (get iri/default-namespaces iri/shacl-ns))
+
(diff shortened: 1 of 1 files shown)


bplatz (approved):
LGTM

PR #1108, opened 2025-08-30
Add objectVarParsing option, parser changes, tests

bplatz:
This PR introduces a configurable objectVarParsing option for transactions and queries to control variable parsing for object values, always honors explicit {\"@variable\": \"?x\"}, and sets sensible defaults (insert/upsert=false, update=true). It includes:

- Parser updates in fql/parse.cljc and validation tweaks to variable detection
- Tests for insert/upsert/update and query behavior with the new option
- Documentation updates: json_ld_query_syntax.md (opts) and new transactions.md guide focused on JSON-LD

### DIFF
--- a/src/fluree/db/query/fql/parse.cljc
+++ b/src/fluree/db/query/fql/parse.cljc
@@ -22,6 +22,12 @@
 
 #?(:clj (set! *warn-on-reflection* true))
 
+(def ^:dynamic *object-var-parsing*
+  "When true (default), bare object values like \"?x\" are parsed as variables.
+  When false, only an explicit object value map with '@variable' is parsed as
+  a variable. Keys and @id variable parsing are unaffected."
+  true)
+
 (defn parse-var-name
   "Returns x as a symbol if x is a valid variable, or nil otherwise. A valid
   variable is a string, symbol, or keyword whose name starts with '?'."
@@ -437,14 +443,25 @@
 
 (defn parse-object-map
   [s-mch p-mch o vars context]
-  (let [o* (expand-keys o context)]
-    (if-let [v (get o* const/iri-value)]
-      (let [attrs (dissoc o* const/iri-value)
-            o-mch (if-let [var (parse-var-name v)]
-                    (parse-variable-attributes var attrs vars context)
+  (let [o* (expand-keys o context)
+        explicit-var (get o "@variable")]
+    (cond
+      explicit-var
+      (let [attrs (-> o (dissoc "@variable") (expand-keys context))
+            var   (parse-var-name explicit-var)
+            o-mch (parse-variable-attributes var attrs vars context)]
+        [(flip-reverse-pattern [s-mch p-mch o-mch])])
+
+      (contains? o* const/iri-value)
+      (let [v     (get o* const/iri-value)
+            attrs (dissoc o* const/iri-value)
+            o-mch (if (and (parse-var-name v)
+                           *object-var-parsing*)
+                    (parse-variable-attributes (parse-var-name v) attrs vars context)
                     (parse-value-attributes v attrs context))]
         [(flip-reverse-pattern [s-mch p-mch o-mch])])
-      ;; ref
+
+      :else ;;ref
       (let [id-map  (with-id o context) ; not o*, we can't use expanded or we'll lose @reverse
             o-mch   (-> id-map
                         (get const/iri-id)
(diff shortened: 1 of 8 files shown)


buckflannigan:
This looks very cool and huge thanks for going to extra mile on the weekend with this corner case for this toggle. Would the toggle be "opts": { "objectVarParsing": false }, as a directive in the /transact API included in the fluree json headers in between the `ledger:` and `insert:` ?
For backtesting this on the text-chunk-loader which accidentally tripped the "Parsing failure due to: variable ? is not bound in where nor values clause", would this be the semi-correct constructor to the Nexus API where the modifications highlighted in bold italic below (ie. the combination of `/insert` endpoint + including `fluree-ledger:` explicitly in the Header are the effective equivalent of "opts": { "objectVarParsing": false } ?
curl -sS -i "https://data.flur.ee/**_insert_**" \
  **_-H "fluree-ledger: buckflannigan/freddy-test" \_**
  -H "Content-Type: application/ld+json" \
  -H "Authorization: Bearer YadaYada" \
  --data-binary @- <<'JSON-LD'
{json-ld}
JSON-LD

bplatz:
/insert API doesn't look for variables, so if you are just inserting then there is no need to do anything special. If you want to use variables, but are also inserting data that looks like a variable but isn't really a variable then you would use this new option but that would be for the /update (formerly /transact) API.

src/fluree/db/query/fql/parse.cljc
```diff
@@ -437,14 +443,25 @@
 
 (defn parse-object-map
   [s-mch p-mch o vars context]
-  (let [o* (expand-keys o context)]
-    (if-let [v (get o* const/iri-value)]
-      (let [attrs (dissoc o* const/iri-value)
-            o-mch (if-let [var (parse-var-name v)]
-                    (parse-variable-attributes var attrs vars context)
+  (let [o* (expand-keys o context)
+        explicit-var (get o "@variable")]
```
dpetran: We should define a const for `@variable` instead of just using a string literal.

src/fluree/db/validation.cljc
```diff
@@ -22,7 +22,10 @@
 (defn variable?
   [x]
   (and (or (string? x) (symbol? x) (keyword? x))
-       (-> x name first (= \?))))
+       (-> x name first (= \?))
+       (if (string? x)
+         (re-matches #"^\?\S+$" x)
```
dpetran: What does this regex match that's not matched by `(-> x name first (= \?))`?

src/fluree/db/query/fql/parse.cljc
```diff
@@ -437,14 +443,25 @@
...
+      (let [attrs (-> o (dissoc "@variable") (expand-keys context))
+            var   (parse-var-name explicit-var)
+            o-mch (parse-variable-attributes var attrs vars context)]
+        [(flip-reverse-pattern [s-mch p-mch o-mch])])
+
+      (contains? o* const/iri-value)
+      (let [v     (get o* const/iri-value)
+            attrs (dissoc o* const/iri-value)
+            o-mch (if (and (parse-var-name v)
+                           *object-var-parsing*)
```
dpetran: I think we should check `*object-var-parsing*` first, since it's less work than `parse-var-name`.

src/fluree/db/query/fql/parse.cljc
```diff
@@ -22,6 +22,12 @@
 
 #?(:clj (set! *warn-on-reflection* true))
 
+(def ^:dynamic *object-var-parsing*
```
dpetran, line 25: I think we can achieve this functionality without resorting to a dynamic var. We're already passing around `bound-vars`/`allowed-vars` in data parsing and `vars` in where-clause parsing, and I think we can add more structure to those arguments to handle this. That would keep our parser using nice pure functions instead of global state.

dpetran (approved):
:grey_question: 

I'd like to do a quick refactor in the near future to remove the dynamic var, but other than that this looks good.

src/fluree/db/validation.cljc
bplatz: It previously would match anything that started with a '?' regardless of what came after it (or not). It now only matches if there is at least one character after the '?' and no space-type characters anywhere in the string. So it will no longer match '?' nor '? foo'.

src/fluree/db/query/fql/parse.cljc
dpetran: Whipped this up earlier today: https://github.com/fluree/db/pull/1109

PR #1109, opened 2025-09-03
Feature/object var parsing pure

dpetran:
A quick refactor of the object-var-parsing branch to remove the dynamic variable.

Uses a more complex structure (`var-config`) where before we were passing `bound-vars`, a set of variables. `var-config` is a map `:bound-vars` and `:parse-object-vars?` which affect how we parse and validate variables.


bplatz (approved):
👍

PR #1112, opened 2025-09-05
Fix/fuel limit logging

dpetran:
This cleans up logging around fuel limits and removes potential channel overflows, by 1) not putting a new error on the error chan for every additional fuel spent over the limit and 2) closing the error chan after the first error.


zonotope (approved):
🪮

PR #1113, opened 2025-09-06
Load commits with a single function

zonotope:
This patch consolidates the common commit loading behavior of both the `read-verified-commit` and `load-commit-with-metadata` functions.


src/fluree/db/nameservice/storage.cljc
```diff
@@ -39,15 +39,15 @@
...
+          ledger-alias (get data "alias")  ;; Already includes :branch
+          filename     (local-filename ledger-alias)]
       (log/debug "nameservice.storage/publish start" {:ledger ledger-alias :filename filename})
-      (let [res (storage/write-bytes store filename record-bytes)]
+      (let [commit-address (get data "address")
+            t-value        (get-in data ["data" "t"])
+            index-address  (get-in data ["index" "address"])
+            ns-metadata    (ns-record ledger-alias commit-address t-value index-address)
+            record-bytes   (json/stringify-UTF8 ns-metadata)
+            res            (storage/write-bytes store filename record-bytes)]
```
dpetran, line 50: I know you didn't change this, but we're not necessarily writing to a file here so we should probably rename the `filename` and the `local-filename` fn.

zonotope: You know, I agreed with you until I went to change it and tried to think of a better name. We _are_ storing files, but we might not necessarily be storing those files in a file store. We put files on s3, in ipfs, and even in memory (although you could argue that the memory example is closer to a value a la key value store, but I think file still fits). I think if there's a name that should be changed, it should be the name of the `FileStore` because that's pretty generic, but I think that's well beyond the scope of this pr.

PR #1114, opened 2025-09-09
Fix @list container serialization for single-value lists

bplatz:
## Summary
Fixed incorrect serialization of single-value @list containers that prevented databases from loading from disk.

## Problem
When a property with `@container: @list` had a single value, the commit data was incorrectly serialized as a vector `["@list", [...]]` instead of a map `{"@list": [...]}`. This caused JSON-LD expansion errors when loading the database from disk.

## Root Cause
The code used `cond->` which applied transformations sequentially:
1. Applied `handle-list-values` creating `{"@list": [value]}`
2. Then applied `first` (when count was 1), which on a map returns a MapEntry as a vector

## Fix
Changed from `cond->` to `cond` to ensure list containers are never passed to `first`, preserving the map structure.

## Test plan
- Added tests for single-value, multi-value, and object reference @list containers
- All new tests pass
- Existing tests continue to pass

### DIFF
--- a/src/fluree/db/flake/commit_data.cljc
+++ b/src/fluree/db/flake/commit_data.cljc
@@ -550,9 +550,10 @@
           p-iri  (get-s-iri db pid compact-fn)
           objs   (subject-block-pred db compact-fn list?
                                      p-flakes)
-          objs*  (cond-> objs
-                   list? handle-list-values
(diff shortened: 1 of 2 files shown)


dpetran (approved):
:1234:

PR #1115, opened 2025-09-14
Positive configuration setting for indexing

zonotope:
This patch changes the indexing options configuration to use `indexingEnabled` instead of the inverted `indexingDisabled`. I think this makes for a friendlier api because it requires less boolean inversions to reason about whether or not indexing will run.


bplatz (approved):
🚥

PR #1116, opened 2025-09-15
Refactor/async db indexing opts

dpetran:
This is a refactor that ensures that the static data that doesn't change over the lifetime of a DB record (the indexing opts) is always present, regardless of whether we have an `AsyncDB` or `FlakeDB`. 

A FlakeDB carries the indexing thresholds around under the keys `:reindex-min-bytes`, `:reindex-max-bytes`, and `:max-old-indexes`. These were not carried over to an `AsyncDB`, which caused a bug when we started reporting indexing data for txn tracking - we would compare the novelty size to `:reindex-min-bytes` to determine the result for the `:indexing-needed` key, but you can't compare a number with `nil`. This was fixed by just falling back to the default `:reindex-min-bytes` setting (1000000), but that was inaccurate. This PR refactors our `AsyncDB` initialization so that we always have that data, and any other static data.

### DIFF
--- a/src/fluree/db/async_db.cljc
+++ b/src/fluree/db/async_db.cljc
@@ -36,7 +36,14 @@
         (<? (dbproto/-class-ids db tracker subject)))))
   (-index-update [_ commit-index]
     (let [commit* (assoc commit :index commit-index)
-          updated-db (->async-db alias branch commit* t)]
+          updated-db (map->AsyncDB {:alias alias
+                                    :branch branch
+                                    :commit commit*
+                                    :t t
+                                    :db-chan (async/promise-chan)
+                                    :reindex-min-bytes reindex-min-bytes
+                                    :reindex-max-bytes reindex-max-bytes
+                                    :max-old-indexes max-old-indexes})]
       (go-try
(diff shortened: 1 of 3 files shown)


zonotope (commented):
The indexing thresholds should only be a concern for systems actually performing indexing. We already have [a protocol for indexing](https://github.com/fluree/db/blob/c81d22946d4fa37f796850d0150eb02f9ab83c16/src/fluree/db/indexer.cljc#L6-L7), and both [FladeDB](https://github.com/fluree/db/blob/c81d22946d4fa37f796850d0150eb02f9ab83c16/src/fluree/db/flake/flake_db.cljc#L351-L355) and [AsyncDB](https://github.com/fluree/db/blob/c81d22946d4fa37f796850d0150eb02f9ab83c16/src/fluree/db/async_db.cljc#L147-L151) already implement this protocol. 

Anything concerning the indexing thresholds should be the responsibility of the protocol, so adding these thresholds should not be necessary because AsyncDBs only dispatch indexing to whatever dbs they wrap. 

If there is something that is accessing these thresholds outside the context of this protocol, then that's what should change.

PR #1118, opened 2025-09-17
refactor: streamline ledger alias handling and ensure branch inclusion

bplatz:
## Summary
- Consolidated ledger name and branch handling into `fluree.db.util.ledger` namespace
- Moved `normalize-ledger-alias` → `ensure-ledger-branch` for better naming consistency
- Fixed branch handling in `query/api` to ensure consistent branch inclusion

### DIFF
--- a/src/fluree/db/connection.cljc
+++ b/src/fluree/db/connection.cljc
@@ -13,6 +13,7 @@
             [fluree.db.storage :as storage]
             [fluree.db.util :as util :refer [get-first get-first-value try* catch*]]
             [fluree.db.util.async :refer [<? go-try]]
+            [fluree.db.util.ledger :as util.ledger]
             [fluree.db.util.log :as log :include-macros true]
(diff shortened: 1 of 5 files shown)


src/fluree/db/connection.cljc
```diff
@@ -450,7 +443,7 @@
       (let [alias (if (fluree-address? alias)
                     (nameservice/address-path alias)
```
dpetran: Now that I think about it, we should call `ensure-ledger-branch` on this branch as well, I don't know that an address is guaranteed to have a branch component.

dpetran: And a quick look shows that `nameservice/address-path` is just a wrapper over `storage/get-local-path`, I think we should just delete `nameservice/address-path` and use `storage/get-local-path`, this file is the only place the nameservice wrapper is used.

dpetran (approved):
:cowboy_hat_face:

PR #1120, opened 2025-09-19
Dedupe identical retract/assert pairs at same t during update/upsert

bplatz:
We allowed update / upset to retract and then re-assert the identical flake. While I thought this should not matter except that it takes extra space unnecessarily (but performs less logic during transaction time), it turns out how we replay flakes it can matter and creates a situation where it seems data isn't there - when it actually is there in the commit correctly.

This results in flakes that look like:
[ex:alice ex:name "Bob" 1 true]
[ex:alice ex:name "Bob" 2 false]
[ex:alice ex:name "Bob" 2 true]

And this creates some unexpected problems.

### Summary
- Fixes a bug where updates/upserts that retract and then re-assert the exact same triple at the same t produced redundant flakes.
- Adds single-pass cancellation so identical retract+assert pairs at the same t are dropped from novelty.
- Keeps insert-only performance unchanged by only doing extra work when retractions occur.

### Problem
- Update/upsert pipeline produce retractions for current values and assertions for new values.
- When a value is “updated” to the same value (or passes through an update that retracts then re-asserts the same triple), both the retract and the assert were included in the resulting commit/novelty.
- This leads to unnecessary work and larger novelty/commit output, especially painful on large transactions.

### Approach
- Implement cancellation in the existing streaming accumulation step (no extra pass over all flakes):
  - While transducing the flake stream into the AVL-backed `:spot` set, we capture retractions in a local volatile.
  - In the reducer’s completion arity, if any retractions were seen, we sweep just those retractions:
    - For each retract r, compute the corresponding assertion a via `flake/flip-flake`.
    - If a is present in the set, remove both a and r (they cancel).
- This is order-independent and runs inside the single pass we already perform to build the set.

### Why this is efficient
- Insert-only transactions: unchanged. We pay zero additional lookups because we only sweep if retractions exist.
- Updates/upserts: extra work is proportional to the number of retractions R, not total flakes N.
  - Time: O(R log N) (contains?/disj on the AVL set)
  - Memory: Should not be significant as we use the same underlying Flake when tracking retractions, so I is the list and pointers only

### Complexity
- Base accumulation (unchanged): O(N log N) into the AVL set.
- Additional cancellation work (only when retractions present): O(R log N).
- Net: O(N log N + R log N), with R << N in typical workloads.

### Correctness details
- Cancellation only occurs for truly identical triples at the same t (same s, p, o, dt, m and flipped op), which matches the desired behavior.
- Non-identical changes (e.g., name “Alice” → “Alice2”, or list differences) are preserved: only new assertions and actual retractions remain.
- Works regardless of assertion/retraction arrival order from the merged channels.

### DIFF
--- a/src/fluree/db/flake/transact.cljc
+++ b/src/fluree/db/flake/transact.cljc
@@ -61,12 +61,33 @@
 
 (defn into-flakeset
   [tracker error-ch flake-ch]
-  (let [flakeset (flake/sorted-set-by flake/cmp-flakes-spot)
-        error-xf (halt-when util/exception?)
-        flake-xf (if-let [track-fuel (track/track-fuel! tracker error-ch)]
-                   (comp error-xf track-fuel)
-                   error-xf)]
-    (async/transduce flake-xf (completing conj) flakeset flake-ch)))
+  (let [flakeset    (flake/sorted-set-by flake/cmp-flakes-spot)
+        error-xf    (halt-when util/exception?)
+        flake-xf    (if-let [track-fuel (track/track-fuel! tracker error-ch)]
+                      (comp error-xf track-fuel)
+                      error-xf)
+        retractions (volatile! [])]
+    (async/transduce
+     flake-xf
+     (completing
+      (fn [acc f]
+        (when (false? (flake/op f))
+          (vswap! retractions conj f))
+        (conj acc f))
+      (fn [acc]
+        (if (seq @retractions)
+          (reduce (fn [s r]
+                    (let [a (flake/flip-flake r)]
+                      (if (contains? s a)
+                        (-> s
+                            (disj a)
+                            (disj r))
+                        s)))
+                  acc
+                  (distinct @retractions))
+          acc)))
+     flakeset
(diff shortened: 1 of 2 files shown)


src/fluree/db/flake/transact.cljc
```diff
@@ -61,12 +61,33 @@
...
+  (let [flakeset    (flake/sorted-set-by flake/cmp-flakes-spot)
+        error-xf    (halt-when util/exception?)
+        flake-xf    (if-let [track-fuel (track/track-fuel! tracker error-ch)]
+                      (comp error-xf track-fuel)
+                      error-xf)
+        retractions (volatile! [])]
+    (async/transduce
+     flake-xf
+     (completing
+      (fn [acc f]
```
zonotope, line 87: This is minor so you can take or leave it, but I think `acc` is too generic. The completion function assumes that `acc` is going to be a set of flakes, so `flakeset` makes that more explicit I think.

zonotope (approved):
🍎

PR #1122, opened 2025-09-26
Atomically update nameservice records

zonotope:
This patch makes it possible for storage backed name services to atomically update their ledger records.

First, it adds a new `swap-bytes` method to the `fluree.db.storage/ByteStore` protocol which takes a path and single argument function, and atomically reads the data stored at the path, applies the function, and writes the result back to the same path replacing what was there previously.

There are also implementations of the `swap-bytes` method for the file, memory, and s3 storage record.

The implementation for the file store uses both os level and process level locks. The os level locks ensure that other processes (or even processes on other servers in the case of a shared filesystem) can't access the same file at the same time, but os level locks do not prevent multiple threads started by the same process from accessing a file at the same time. The process level locks only work within the same process and have no effect on other processes or servers, but they do prevent multiple threads in the same process from accessing the same file. That's why both locking strategies are necessary. 

The implementation for the s3 store relies on the `ETAG` header returned by the AWS api for any s3 requests. The etag is based on a hash of the file and will change if the file has been modified. `swap-bytes` for s3 storage first does a `GET` request to read the data and keeps track of the etag. Then it issues a `PUT` request to write the data and includes the etag in an `IF-MATCH` header that will cause the request to fail if the etag has changed. I also had to make some modifications to the xhttp util namespace to be able to access the headers returned from an http request. 

Besides `swap-bytes` I've added the index t to the nameservice record and reimplemented storage-backed nameservice updates with `swap-bytes` by first validating that the commit and index t is newer before overwriting the previous record.


dpetran (approved):
:lock: 

Are the changes in `xhttp` going to break server/nexus-server/raft?

PR #1123, opened 2025-09-26
add support for SPARQL empty prefix

dpetran:
Also added a test for the already-supported BASE clause.

### DIFF
--- a/src/fluree/db/query/sparql/translator.cljc
+++ b/src/fluree/db/query/sparql/translator.cljc
@@ -362,15 +362,15 @@
   ;; PrefixDecl ::= <'PREFIX'> WS PNAME_NS WS IRIREF
   ;; [:PrefixDecl "e" "x" ":" [:IRIREF "<http://example.com/>"]]
   [[_ & prefix-decl]]
-  (let [prefix (->> (drop-last 2 prefix-decl) vec str/join not-empty
-                    ;; sometimes no prefix is specified: PREFIX : <my:IRI> and all IRIs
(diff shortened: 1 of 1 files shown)


src/fluree/db/query/sparql/translator.cljc
```diff
@@ -362,12 +362,14 @@
   ;; PrefixDecl ::= <'PREFIX'> WS PNAME_NS WS IRIREF
   ;; [:PrefixDecl "e" "x" ":" [:IRIREF "<http://example.com/>"]]
   [[_ & prefix-decl]]
-  (let [prefix (->> (drop-last 2 prefix-decl) vec str/join)
+  (let [prefix (->> (drop-last 2 prefix-decl) vec str/join not-empty)
```
zonotope:
```suggestion
  (let [prefix (->> (drop-last 2 prefix-decl) vec str/join not-empty (or ":")
```

src/fluree/db/query/sparql/translator.cljc
```diff
@@ -362,12 +362,14 @@
...
         iri    (->> prefix-decl
                     (drop-while (comp not sequential?))
                     first
                     parse-term)]
-    [[prefix iri]]))
+    ;; sometimes no prefix is specified: PREFIX : <my:IRI>
+    ;; and all IRIs are just prefixed with a colon
+    [[(or prefix ":") iri]]))
```
zonotope: I think you could remove this change if you accept my suggested change above ☝🏾 (and move the descriptive comment there).

PR #1124, opened 2025-09-30
Ensure alias includes branch name for new commits

zonotope:
This patch primarily ensures that the ledger alias for new commits includes the branch name. If the previous commit's ledger alias did not include a branch name, then "main" is assumed and appended to the existing alias. The branch name is also ensured in the ledger alias stored in a connection's ledger cache.

This patch also modifies the `fluree.db.util.ledger/ensure-ledger-branch` function was modified so that when `""` or `nil` was passed in as a ledger name, no branch would be appended.

The `fluree.db.nameservice.sub` namespace was removed with it's functionality consolidated with the `fluree.db.connection` namespace. First, there were duplicate `notify` functions both in the `fluree.db.connection` and `fluree.db.nameservice.sub` namespaces, but different code paths used each one. This would have led to bugs as each function was modified and diverged. Next, all of the functions in this namespace operated on connection records, so they should be in the connection hierarchy. If the connection namespace is too big, then we should refactor connections into discrete components with limited functionality, and pass those components to other namespaces if necessary (like a `subscriber` component within a connection, for example). 

Lastly, I removed some unused functions.


src/fluree/db/util/ledger.cljc
```diff
@@ -32,9 +32,11 @@
    e.g., 'my-ledger' -> 'my-ledger:main'
          'my-ledger:branch' -> 'my-ledger:branch'"
   [ledger-alias]
-  (if (ledger-branch ledger-alias)
-    ledger-alias
-    (str ledger-alias ":" const/default-branch-name)))
+  (if (not-empty ledger-alias)
```
dpetran, line 35: Where does this get called from where we can receive a `nil` or `""` `ledger-alias`?

zonotope: I don't think it does, but it could. We never want to get a ledger that looks like ":main", so this is a defense against that future bug.

dpetran: Ok, I hope it doesn't come up, because in that case this function won't do what it says on the tin. Maybe we should throw in that case, since we've clearly got some bad inputs and we don't want to propagate a malformed ledger alias.

PR #1126, opened 2025-10-02
Move ledger operations to the ledger namespace

zonotope:
This is a small patch to move stateful operations on ledger records to the `fluree.db.ledger` namespace. After moving these operations, we were finally able to move the `Transactable` protocol from `fluree.db.flake.transact` to `fluree.db.transact` without any circular namespace dependencies.

There's a lot more work to be done to simplify and clean up our namespace hierarchy that I've left for later. I noticed these fixes while tracking down an unrelated issue.


dpetran (approved):
:raccoon:

PR #1127, opened 2025-10-03
properly handle whitespace after composed expressions

dpetran:
This uses ROUND to test, but it would occur after any :UnaryExpression.


zonotope (approved):
🦞

PR #1128, opened 2025-10-03
Revert "Optimize subject joins"

zonotope:
This patch reverts #1080 because the optimizations there required a `psot` index which would have been onerous to build for large existing ledgers. There are also some other proposals or changing our indexing and serialization approach which may make those optimization unnecessary, and, at the very least, would also require migrations, so it would be better to do those all at once.


dpetran (approved):
:pager:

PR #1130, opened 2025-10-07
Remove resources/logback.xml and add LOG_LEVEL env variable support

bplatz:
## Summary
- Removed `resources/logback.xml` to avoid shipping logging config with the library (consuming apps should provide their own)
- Updated `dev-resources/logback.xml` and `test-resources/logback-test.xml` to support dynamic log level via `LOG_LEVEL` environment variable

## Usage
Set log level via environment variable:
```bash
export LOG_LEVEL=DEBUG
# or
LOG_LEVEL=DEBUG clojure -M:cljtest ...
```

Defaults to INFO (dev) and DEBUG (test) if not specified.


dpetran (approved):
:deciduous_tree:

PR #1133, opened 2025-10-10
Add streaming query API with metadata tracking support

bplatz:
## Summary
Adds streaming query API that emits individual results as they are produced, rather than collecting all results into a vector. Supports all query types (SELECT, SELECT ONE, CONSTRUCT) and formats (FQL, SPARQL).

## Key Features
- **Memory efficient streaming**: Results emitted individually for large result sets
- **Metadata tracking**: Optional `:meta` tracking emits final metadata map after results
- **CONSTRUCT streaming**: Emits ungrouped graph nodes without @graph wrapper
- **Full query support**: ORDER BY, GROUP BY, and aggregates (collect before emitting)
- **Cache validation**: Early validation rejects :cache option for streaming queries

## API Functions
- `fluree.db.api/query-stream` - Stream query against database
- `fluree.db.api/query-connection-stream` - Stream query via connection

## Metadata Format
When `:meta true` is enabled, final result is:
```clojure
{:_fluree-meta {:fuel 123, :time "5ms", :policy {...}, :status 200}}
```

## Test Coverage
10 tests, 39 assertions covering:
- Basic streaming (SELECT, LIMIT, SELECT ONE)
- Metadata tracking (FQL and SPARQL)
- Connection-based queries
- CONSTRUCT queries (FQL and SPARQL)

## Implementation Notes
- Cache validation moved to public API layer
- Internal functions simplified (validation upstream)
- CONSTRUCT uses `cat` transducer for streaming ungrouped nodes
- Metadata calculated from tracker counters without forcing realization

### DIFF
--- a/src/fluree/db/api.cljc
+++ b/src/fluree/db/api.cljc
@@ -849,6 +849,53 @@
         (throw (ex-info "No nameservice available for querying"
                         {:status 400 :error :db/no-nameservice})))))))
 
+;; Streaming Query APIs
+
+(defn query-stream
+  "Streaming version of `query`. Returns a core.async channel that emits
+  individual results as produced, rather than collecting into a vector.
+
+  See `query` for query syntax and options. Key differences:
+    - Returns core.async channel instead of promise
+    - Emits results individually (reduces memory for large result sets)
+    - Caching not supported
+    - CONSTRUCT queries emit individual graph nodes (not grouped by subject)
+    - When :meta enabled, final item is {:_fluree-meta {...}}"
+  ([ds q]
+   (query-stream ds q {}))
+  ([ds q opts]
+   (cond
+     (util/exception? ds)
+     (async/to-chan! [ds])
+
+     (:cache opts)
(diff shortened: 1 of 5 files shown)


src/fluree/db/connection.cljc
```diff
@@ -500,33 +506,70 @@
...
-      (catch* e (log/debug e "Failed to complete ledger deletion")))))
+  (go-try
+    (let [alias* (cond-> alias
+                   (fluree-address? alias) storage/get-local-path
+                   true util.ledger/ensure-ledger-branch)
+          primary-pub   (:primary-publisher conn)
+          dependent-vgs (check-ledger-dependencies primary-pub alias*)]
+
+      (when (seq dependent-vgs)
+        (throw (existing-dependency-error alias* dependent-vgs)))
```
zonotope: It might be useful to have forced/cascading drops too.

bplatz:
Closed in favor of https://github.com/fluree/db/pull/1186

PR #1146, opened 2025-10-21
Consolidate index metadata extraction and prepare for nameservice-based index tracking

bplatz:
Centralizes index metadata access patterns that were previously scattered across multiple modules and removes dead code from the commit workflow.

Removes 'index' from being written to commits going forward to focus on a single source of truth in nameservice. Currently index-t/address are embedded in commit so just dissoc from write for now, but in the future it should be held in a separate data structure outside of the commit that will hold other nameservice info.

Key changes:
  - New index.metadata namespace: Provides canonical functions (index-metadata, index-address, index-t) for extracting index metadata from commits or JSON-LD
  - Removed dead code: Deleted unused use-latest-index function from commit_data.cljc
  - Removed duplicate index-t: Eliminated duplicate function from commit_data.cljc in favor of centralized implementation

Foundation for future work:
This refactoring prepares the codebase to migrate from storing index metadata in commits to using the nameservice as the system of record. The nameservice already tracks index address, index t, and status fields and is the only thing used, index metadata is already ignored in commits.

Future work will update callers to accept nameservice metadata to extract latest index, eliminating the need to duplicate index metadata in commit structures.

### DIFF
--- a/src/fluree/db/ledger.cljc
+++ b/src/fluree/db/ledger.cljc
@@ -6,6 +6,7 @@
             [fluree.db.did :as did]
             [fluree.db.flake :as flake]
             [fluree.db.flake.commit-data :as commit-data]
+            [fluree.db.index.metadata :as index.meta]
             [fluree.db.json-ld.credential :as credential]
             [fluree.db.json-ld.iri :as iri]
             [fluree.db.nameservice :as nameservice]
@@ -273,7 +274,8 @@
 (defn write-commit
   [commit-storage alias {:keys [did private]} commit]
   (go-try
-    (let [commit-jsonld (commit-data/->json-ld commit)
+    (let [commit-jsonld (-> (commit-data/->json-ld commit)
+                            (dissoc "index"))
           ;; For credential/generate, we need a DID map with public key
           did-map (when (and did private)
                     (if (map? did)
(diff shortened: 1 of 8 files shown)


src/fluree/db/ledger.cljc
```diff
@@ -273,7 +274,8 @@
 (defn write-commit
   [commit-storage alias {:keys [did private]} commit]
   (go-try
-    (let [commit-jsonld (commit-data/->json-ld commit)
+    (let [commit-jsonld (-> (commit-data/->json-ld commit)
+                            (dissoc "index"))
```
zonotope, line 278: Right now we broadcast new indexes by publishing a [new commit that includes an updated index map](https://github.com/fluree/db/blob/fdb2a2320b0ce172f43d87bbf0deb309ff386ac4/src/fluree/db/nameservice/storage.cljc#L58-L76). I agree that we shouldn't overload these methods and that the latest commit and latest index should live in different data structures on the nameservice. 

Does this pr add another way to publish new indexes since the index information is now stripped from the commit document?

bplatz: It was already ignored in the commit -- during load it gets it from the nameservice. So the commit index reference is not used at all.

PR #1169, opened 2025-12-08
Feature/structured logging

dpetran:


dpetran:
This is just a work-in-progress, the otel stuff is just for improved iteration while we're figuring out how to plumb everything together. Likewise, the separate logging api (with ! suffixes) are to make iterating easier, the plan is to integrate it properly once the basic tracing functionality is working.

bplatz:
One thing to consider in making it work correctly is making it work like an actual app would - so I think the goal would be to configure the telemere backend there, since we cannot ship telemere as a backend with fluree/db. That way you can make sure that the fluree/db loging flows through to the telemere backend (e.g. in fluree/server) as you expect... and I think the key here is MDC handling which is the tricky part, especially across threads (including core.async which was our issue before). If we want to adopt telemere as the backend logger in fluree/server that is fine, but that is where the backend logger code needs to live. I presume we'd remain with SLF4J as the facade here, but maybe you had a different idea in mind.

PR #1178, opened 2025-12-16
Streaming Aggregates for Queries

bplatz:
### Summary

Implements incremental streaming aggregation for queries, reducing memory overhead for aggregate queries over large result sets. Instead of collecting all solutions into groups before computing aggregates, this approach updates aggregate state incrementally as solutions flow through the query pipeline.

### Key Changes

- **Streaming aggregate descriptors** (`src/fluree/db/query/exec/eval.cljc`) - Registry of `{:init :step! :final}` functions for `count`, `count-distinct`, `sum`, `avg`, `min`, `max`
- **Eligibility detection** (`src/fluree/db/query/exec/group.cljc`) - Automatically uses streaming path when query structure permits (no HAVING clause, all selectors are group vars or streamable aggregates)
- **Selector updates** (`src/fluree/db/query/exec/select.cljc`) - `AggregateSelector` and `AsSelector` now carry optional `streaming-agg` descriptor
- **Parse-time detection** (`src/fluree/db/query/fql/parse.cljc`) - Builds streaming descriptors for simple aggregate forms like `(count ?x)`, `(sum ?y)`

### How It Works

Traditional grouping collects all solutions, then applies aggregate functions:
solutions → [collect all] → group → apply agg-fn → results

Streaming aggregation updates state incrementally:
solutions → step!(state, value) per solution → final(state) → results

### Supported Aggregates

| Function | Streaming Support |
|----------|-------------------|
| `count` | ✅ |
| `count` with `*` | ✅ |
| `count-distinct` | ✅ |
| `sum` | ✅ |
| `avg` | ✅ |
| `min` | ✅ |
| `max` | ✅ |

### Limitations

Falls back to traditional grouping when:
- Query has a `HAVING` clause
- Aggregate uses complex expressions (e.g., `(sum (+ ?x ?y))`)
- Non-streamable aggregate functions (e.g., `sample`, `group_concat`)


src/fluree/db/query/exec/eval.cljc
```diff
@@ -90,6 +90,87 @@
...
   (where/->typed-val (count coll)))
 
+;; Streaming aggregate descriptors (incremental group aggregation)
+
+(declare compare*)
+
+(def streaming-aggregate-registry
+  "Streaming aggregate registry: op -> {:init :step! :final}."
+  {'count        {:init  (fn [] 0)
+                  :step! (fn [state tv]
```
zonotope: Tiny nitpick, but the value of `:step!` here is a pure function, so I don't think we should end the key with "!" and `:step` is more appropriate.

src/fluree/db/query/exec/select.cljc
```diff
@@ -60,53 +60,73 @@
       :sparql (with-meta selector {`format-value select.sparql/format-wildcard-selector-value})
       (with-meta selector {`format-value select.fql/format-wildcard-selector-value}))))
 
-(defrecord AggregateSelector [agg-fn]
+(defrecord AggregateSelector [agg-fn streaming-agg]
```
zonotope: I think I would prefer a `StreamingAggregateSelector` record with its own separate `ValueSelector` implementation. Having this opaque map complects multiple behavior traits into the same record.

src/fluree/db/query/exec/eval.cljc
```diff
@@ -90,6 +90,87 @@
   [coll]
   (where/->typed-val (count coll)))
 
+;; Streaming aggregate descriptors (incremental group aggregation)
+
+(declare compare*)
+
+(def streaming-aggregate-registry
```
zonotope: Having these grouped in a map like this implies that they are dynamic and will change, but these are a static, closed list as far as I can tell. 

I think we're reinventing facilities the language already has. I think we could implement the same functionality with a `StreamingAggregator` protocol whose methods are `initialize!`, `step!`, and `finalize`, and then records that implement that protocol for all the keys in this registry map. Those records could also encapsulate their state .

zonotope (commented):
I'd like to start a discussion about the approach and whether or not my suggestions are feasible.

src/fluree/db/query/exec/select.cljc
bplatz: Addressed in 7c57470b9a52c69e9ab2b6f69a0219e4ab4fa64f

src/fluree/db/query/exec/eval.cljc
bplatz: Addressed in 7c57470b9a52c69e9ab2b6f69a0219e4ab4fa64f

src/fluree/db/query/exec/eval.cljc
bplatz: @zonotope take a look at commit 22d94e64af5af9c6e7800f68d7f10d75e337ad49 -- let me know if you prefer that.

src/fluree/db/query/exec/aggregate.cljc
```diff
@@ -0,0 +1,219 @@
...
+          (.atOffset datetime ZoneOffset/UTC)))
+#?(:clj (defmethod ->offset-date-time LocalDate
+          [^LocalDate date]
+          (.atOffset (.atStartOfDay date) ZoneOffset/UTC)))
+#?(:clj (defmethod ->offset-date-time :default
+          [x]
+          (throw (ex-info "Cannot convert value to OffsetDateTime."
+                          {:value  x
+                           :status 400
+                           :error  :db/invalid-fn-call}))))
```
zonotope: I think this fits better as a protocol since we're dealing with predefined types directly. I like to reserve multimethods for flat clojure data like maps and vectors, but protocols get much better performance if we have the option to let the jvm dispatch on type.

Here's how I would define and implement a protocol to convert to an offset datetime. I haven't run this code, so there might be some typos. 

```suggestion
#?(:clj
   (defprotocol OffsetDateTimeConverter
     (->offset-date-time [this])))
 
#?(:clj
   (extend-protocol OffsetDateTimeConverter
     OffsetDateTime
     (->offset-date-time
       [^OffsetDateTime this]
       this)

     LocalDateTime
     (->offset-date-time
       [^LocalDateTime this]
       (.atOffset this ZoneOffset/UTC))

     LocalDate
     (->offset-date-time
       [^LocalDate this]
       (-> this .atStartOfDay (.atOffset ZoneOffset/UTC)))

     Object
     (->offset-date-time
       [this]
       (throw (ex-info "Cannot convert value to OffsetDateTime."
                       {:value  this
                        :status 400
                        :error  :db/invalid-fn-call})))))
```

src/fluree/db/query/exec/aggregate.cljc
```diff
@@ -0,0 +1,219 @@
...
+          time))
+#?(:clj (defmethod ->offset-time LocalTime
+          [^LocalTime time]
+          (.atOffset time ZoneOffset/UTC)))
+#?(:clj (defmethod ->offset-time :default
+          [x]
+          (throw (ex-info "Cannot convert value to OffsetTime."
+                          {:value  x
+                           :status 400
+                           :error  :db/invalid-fn-call}))))
```
zonotope: Similarly, here's how I would define and implement an offset time converter protocol.

```suggestion
#?(:clj
   (defprotocol OffsetTimeConverter
     (->offset-time [this])))

#?(:clj
   (extend-protocol OffsetTimeConverter
     OffsetTime
     (->offset-time
       [^OffsetTime this]
       this)

     LocalTime
     (->offset-time
       [^LocalTime this]
       (.atOffset time ZoneOffset/UTC))

     Object
     (->offset-time
       [this]
       (throw (ex-info "Cannot convert value to OffsetTime."
                       {:value  this
                        :status 400
                        :error  :db/invalid-fn-call})))))
```

src/fluree/db/query/exec/group.cljc
```diff
@@ -95,17 +101,103 @@
   [match compact]
   (display-sparql-aggregate match compact))
 
+(defn- streaming-agg-selector?
+  "Returns true if selector supports streaming aggregation."
+  [sel]
+  (or (instance? fluree.db.query.exec.select.StreamingAggregateSelector sel)
```
zonotope: You could require this type to avoid this long line

src/fluree/db/query/exec/group.cljc
```diff
@@ -95,17 +101,103 @@
...
+  - Either implicit grouping with all aggregate selectors,
+    or explicit group-by with all selectors being group vars or aggregates"
+  [having streaming-aggs implicit? selectors group-vars group-vars-set]
+  (and (nil? having)
+       (seq streaming-aggs)
+       (if implicit?
+         (every? streaming-agg-selector? selectors)
+         (and (seq group-vars)
+              (every? (fn [sel]
+                        (or (and (instance? fluree.db.query.exec.select.VariableSelector sel)
```
zonotope: You could require this type to avoid this long line

src/fluree/db/query/exec/group.cljc
```diff
@@ -95,17 +101,103 @@
...
+                                 (contains? group-vars-set (:var sel)))
+                            (streaming-agg-selector? sel)))
+                      selectors)))))
+
+(defn- update-streaming-groups
+  "Reducer function that updates streaming aggregate states for each solution.
```
zonotope: When I read "reducer function", I think a function you pass to `reduce`. Then the function signature with four arguments is immediately confusing. Perhaps the docstring could be tweaked so that the first sentence is "Updates streaming aggregae states for a solution". I think that's clearer.

src/fluree/db/query/exec/aggregate.cljc
```diff
@@ -0,0 +1,219 @@
...
+
+(defrecord CountAggregator []
+  StreamingAggregator
+  (initialize [_] 0)
+  (step [_ state tv]
+    (if (some-> tv :value some?)
+      (inc state)
+      state))
+  (finalize [_ state]
+    (where/->typed-val state)))
```
zonotope: I think the code would be simpler if the aggregator records encapsulated their state. Then you could use the aggregators themselves to keep track of the running tally. I also would remove the `initialize` method in favor of a constructor. Here's how I'd change the `CountAggregator`:

```suggestion
(defrecord CountAggregator [tally]
  StreamingAggregator
  (step [this tv]
    (if (some-> tv :value some?)
      (update this :tally inc)))
  (finalize [_]
    (where/->typed-val tally)))

(defn count-aggregator
  []
  (->CountAggregator 0))
```

I'd change the other records similarly.

src/fluree/db/query/exec/select.cljc
```diff
@@ -63,50 +63,82 @@
...
+  Optional parameters:
+    `streaming-agg` - descriptor map for incremental aggregates
+    `agg-info` - {:fn-name sym, :vars #{?v ...}}"
   ([agg-function]
-   (aggregate-selector agg-function nil))
-  ([agg-function agg-info]
-   (cond-> (->AggregateSelector agg-function)
+   (->AggregateSelector agg-function))
+  ([agg-function streaming-agg]
+   (if streaming-agg
```
zonotope: I think this conditional should be lifted to the caller of this constructor function, and this function should be replaced with two different functions. In general I think conditionals should be evaluated as soon as possible. This simplifies the code by making clear execution branches instead of putting everything together and transmitting conditionals without context. If there's shared functionality between the two branches, then that could still be implemented with helper functions.

src/fluree/db/query/fql/parse.cljc
```diff
@@ -666,30 +667,68 @@
...
-        (eval/compile context)
-        (select/aggregate-selector agg-info))))
+  (let [code          (parse-code f)
+        fn-name       (when (seq? code) (first code))
+        agg-vars      (variables code)
+        agg-fn        (eval/compile code context)
+        streaming-agg (build-streaming-agg code)
+        agg-info      {:fn-name fn-name
+                       :vars    agg-vars}]
+    (select/aggregate-selector agg-fn streaming-agg agg-info)))
```
zonotope: With respect to my previous comment about lifting the conditional, I would rewrite this function this way:

```suggestion
  (let [code          (parse-code f)
        fn-name       (when (seq? code) (first code))
        agg-vars      (variables code)
        agg-fn        (eval/compile code context)
        agg-info      {:fn-name fn-name
                       :vars    agg-vars}]
    (if-let [streaming-agg (build-streaming-agg code)]
      (select/streaming-aggregate-selector agg-fn streaming-agg agg-info)
      (select/aggregate-selector agg-fn agg-info))))
```

I'd do the same thing to `parse-select-as-fn` too, although a lot of what I'd change was here before this pr.

src/fluree/db/query/exec/group.cljc
```diff
@@ -95,17 +101,103 @@
...
-                           (completing group-solution
-                                       (partial unwind-groups grouping))
+  [{:keys [group-by select having]} solution-ch]
+  (let [selectors      (util/sequential select)
+        group-vars     (vec group-by)
+        group-vars-set (set group-vars)
+        streaming-aggs (->> selectors
+                            (filter streaming-agg-selector?)
+                            (mapv :streaming-agg))
+        implicit?      (and (empty? group-vars)
```
zonotope, line 183: If this is necessary, then I think it's a bug unless I'm missing something. `select/implicit-grouping?`. If I remember correctly, `select/implicit-grouping?` should only return a truthy value if there are no grouped variables.

src/fluree/db/query/exec/group.cljc
```diff
@@ -95,17 +101,103 @@
...
+                            (some select/implicit-grouping? selectors))
+        streaming?     (streaming-eligible? having streaming-aggs implicit?
```
zonotope: I don't think this needs to be a separate binding. I think `(if (streaming-eligible? having ...)` is just as if not more readable.

src/fluree/db/query/exec/group.cljc
bplatz: select/implicit-grouping? only checks if a selector is an aggregate type, not whether there are group variables. Both conditions are needed for implicit grouping detection

bplatz:
@zonotope ready for re-review:

Changes Made
1. aggregate.cljc - Protocol conversions
Converted ->offset-date-time multimethod to OffsetDateTimeConverter protocol for better JVM dispatch performance
Converted ->offset-time multimethod to OffsetTimeConverter protocol
2. aggregate.cljc - Aggregator state encapsulation
Refactored StreamingAggregator protocol to encapsulate state in records
Removed initialize method; aggregators now start with initial state via constructor functions
Changed step signature from (step this state tv) to (step this tv) - returns new aggregator
Renamed finalize to complete (to avoid conflict with Java's Object.finalize())
Updated all aggregator records: CountAggregator, CountStarAggregator, CountDistinctAggregator, SumAggregator, AvgAggregator, MinAggregator, MaxAggregator
streaming-aggregators map now stores constructor functions
3. group.cljc
Added imports for AsSelector, StreamingAggregateSelector, VariableSelector to avoid long fully-qualified names
Updated streaming-agg-selector? and streaming-eligible? to use short names
Updated docstring for update-streaming-groups
Updated update-streaming-groups and finalize-streaming-groups to work with new aggregator interface
Inlined the streaming? binding into the if condition
4. select.cljc
Split aggregate-selector into two separate functions:
aggregate-selector - for non-streaming aggregates
streaming-aggregate-selector - for streaming aggregates
Removed conditional logic from the function (lifted to caller)
5. parse.cljc
Updated parse-select-aggregate to explicitly choose between aggregate-selector and streaming-aggregate-selector

bplatz:
Note on ci/cd errors, this is fixed with https://github.com/fluree/db/pull/1190 -- we use 'latest' version of clj-kondo and I think a new version is flagging these.

src/fluree/db/query/exec/aggregate.cljc
```diff
@@ -0,0 +1,256 @@
...
+  (complete [this] "Convert accumulated state to result typed-val"))
+
+(defrecord CountAggregator [tally]
+  StreamingAggregator
+  (step [this tv]
+    (if (some-> tv :value some?)
+      (update this :tally inc)
+      this))
+  (complete [this]
+    (where/->typed-val (:tally this))))
```
zonotope, line 25: Small nit, but tally is already bound, so you can just do this:

```suggestion
    (where/->typed-val tally)))
```

src/fluree/db/query/exec/aggregate.cljc
```diff
@@ -0,0 +1,256 @@
...
+
+     Object
+     (->offset-time
+       [this]
+       (throw (ex-info "Cannot convert value to OffsetTime."
+                       {:value  this
+                        :status 400
+                        :error  :db/invalid-fn-call})))))
+
+(defn compare*
```
zonotope, line 180: I think this could have a slightly more descriptive name like `compare-values`.

src/fluree/db/query/exec/aggregate.cljc
```diff
@@ -0,0 +1,256 @@
...
+      (compare val-a val-b)
+
+      ;; datetimes need to be converted to OffsetDateTimes for proper comparison
+      (and (contains? comparable-time-datatypes dt-a)
+           (contains? comparable-time-datatypes dt-b))
+      #?(:clj (compare (->offset-date-time val-a) (->offset-date-time val-b))
+         :cljs (compare val-a val-b))
+
+      ;; same types compare
+      (= dt-a dt-b)
```
zonotope, line 200: this might not matter that much in practice, but I feel like this case should come first since equality is the least expensive check we do in this function.

src/fluree/db/query/exec/aggregate.cljc
```diff
@@ -0,0 +1,256 @@
...
+                       :b      val-b :b-dt dt-b
+                       :status 400
+                       :error  :db/invalid-query})))))
+
+(defrecord MinAggregator [current]
+  StreamingAggregator
+  (step [this tv]
+    (let [cur (:current this)]
+      (cond
+        (nil? (some-> tv :value)) this
```
zonotope, line 215: The `nil?` and `some->` here is redundant. We can just do this:

```suggestion
        (-> tv :value nil?) this
```

src/fluree/db/query/exec/aggregate.cljc
```diff
@@ -0,0 +1,256 @@
...
+(defn min-aggregator
+  []
+  (->MinAggregator nil))
+
+(defrecord MaxAggregator [current]
+  StreamingAggregator
+  (step [this tv]
+    (let [cur (:current this)]
+      (cond
+        (nil? (some-> tv :value)) this
```
zonotope, line 231: Same here for the redundant `nil?` and `some->`

```suggestion
        (-> tv :value nil?) this
```

src/fluree/db/query/exec/aggregate.cljc
```diff
@@ -0,0 +1,256 @@
...
+        (nil? cur) (assoc this :current tv)
+        (neg? (compare* tv cur)) (assoc this :current tv)
+        :else this)))
+  (complete [this]
+    (or (:current this) (where/->typed-val nil))))
```
zonotope, line 220: This `or` is redundant. `current` should never be false, so the first branch of the `or` is falsey only if `current` is `nil`, so the second branch shouldn't be necessary. This should work:

```suggestion
    (where/->typed-val current)))
```

src/fluree/db/query/exec/aggregate.cljc
```diff
@@ -0,0 +1,256 @@
...
+  StreamingAggregator
+  (step [this tv]
+    (let [cur (:current this)]
+      (cond
+        (nil? (some-> tv :value)) this
+        (nil? cur) (assoc this :current tv)
+        (pos? (compare* tv cur)) (assoc this :current tv)
+        :else this)))
+  (complete [this]
+    (or (:current this) (where/->typed-val nil))))
```
zonotope, line 236: Same here

```suggestion
    (where/->typed-val current)))
```

src/fluree/db/query/exec/group.cljc
```diff
@@ -95,17 +102,102 @@
...
+  Returns updated groups map with accumulated aggregate states per group key."
+  [group-vars streaming-aggs groups solution]
+  (let [group-key (extract-group-key group-vars solution)
+        {:keys [group-vars-map agg-states]}
+        (get groups group-key
+             {:group-vars-map (zipmap group-vars group-key)
+              :agg-states     {}})
+        agg-states'
+        (reduce (fn [states {:keys [arg-var result-var aggregator]}]
+                  (let [agg    (get states result-var (aggregator))
```
zonotope, line 142: the spacing here is off

src/fluree/db/query/exec/group.cljc
```diff
@@ -95,17 +102,102 @@
...
+                        tv     (when arg-var
+                                 (some-> solution
+                                         (get arg-var)
+                                         where/mch->typed-val))
+                        agg'   (agg/step agg tv)]
+                    (assoc states result-var agg')))
```
zonotope, line 148: I think you can get rid of both the `agg` and `agg'` bindings above and just do

```clojure
(update states result-var (fnil agg/step (aggregator)) tv)
```

you can take or leave that. I prefer it since it's less code, but the `fnil` makes it a little more complex so i get if you prefer how it is now.

zonotope (approved):
🎒

PR #1182, opened 2026-01-05
Feature/inline multivariable filters

zonotope:
This patch extends the existing in-line filter optimization for single variable filter functions to multi-variable functions.  It also introduces a clause-oriented query reordering flow. 

Filters are separated from binding patterns at the beginning of the optimization step, the binding patterns are then reordered, and then both single- and multi- variable filters are inlined onto the sorted binding patterns when safe. The same separation, reordering, and inlining now occur within nested clauses, improving performance while preserving correctness and determinism.

Filters are inlined onto the sorted list of binding patterns at the point when exactly one referenced variable becomes newly bound by a pattern, and all other referenced variables were bound by a previous pattern. Filters that reference multiple variables and would have more than one newly bound in the same pattern are not inlined; they are emitted immediately  after that pattern as a (not inline) `:filter` pattern.


bplatz (approved):
Looks good other than the minus, et.c issues I brought up...

PR #1191, opened 2026-01-16
Fix/remote ledger not found

dpetran:
This fixes a problem introduced on `main` where we only check the primary publisher for a ledger alias before returning a 404 Not Found error. However this is handled correctly by the `connection/ledger-ledger-alias` function, so this PR restores the old, correct behavior.


zonotope (approved):
⚙️

PR #1193, opened 2026-04-23
fix: resolve fulltext config against state.t(), not snapshot.t

bplatz:
On a first-ever index build the ledger has no prior index, so LedgerState::load returns a genesis snapshot with snapshot.t == 0 and puts every committed flake in novelty. ApiFulltextConfigProvider::resolve (and the mirror call sites in admin::reindex and rebase) was passing state.snapshot.t as to_t to resolve_ledger_config, but Novelty::for_each_overlay_flake filters with flake.t <= to_t, so the genesis to_t = 0 drops every config flake. The provider silently returned an empty Vec, the indexer left FulltextHookConfig empty, FulltextHook::on_op never matched configured predicates, and build_and_upload_fulltext_arenas short-circuited — no BM25 arena was ever built for properties declared via f:fullTextDefaults, and fulltext(?v, "…") returned unbound.

Fix is a three-line change: pass state.t() (= max(novelty.t, snapshot.t)) instead  of state.snapshot.t in indexer_fulltext_provider.rs, admin.rs, and rebase.rs. Existing tests masked this by using a two-reindex pattern where the first build still wrote raw config flakes to the binary index, letting the second build find them at a non-zero snapshot.t. 

Added four regression tests covering the single-pass scenarios (first-build via provider, language-tagged literal, incremental extension, two-commits full rebuild) — all fail on main without the fix, pass with it.

### DIFF
--- a/fluree-db-api/tests/it_query_fulltext.rs
+++ b/fluree-db-api/tests/it_query_fulltext.rs
@@ -859,3 +859,99 @@ async fn fulltext_configured_property_picked_up_by_build_index_for_ledger() {
         "non-matching title should not score: {results:?}"
     );
 }
+
+/// Reproducer for the Solo bug report (2026-04-23): first-ever index build via
+/// the background / provider path must pick up `f:fullTextDefaults` from
+/// novelty. Before the fix, `ApiFulltextConfigProvider::resolve()` called
+/// `resolve_ledger_config(snapshot, novelty, snapshot.t)` with
+/// `snapshot.t == 0` (genesis), which filtered out all novelty flakes and
+/// returned an empty configured-property list.
+#[tokio::test]
+async fn fulltext_configured_property_first_build_via_provider() {
+    use fluree_db_transact::{CommitOpts, TxnOpts};
+
+    let fluree = FlureeBuilder::memory().build_memory();
+    let ledger_id = "it/fulltext-config-first-build:main";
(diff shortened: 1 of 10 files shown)


fluree-db-api/tests/it_query_fulltext.rs
```diff
@@ -859,3 +859,482 @@ async fn fulltext_configured_property_picked_up_by_build_index_for_ledger() {
...
+    });
+    let result = support::query_jsonld(&fluree, &loaded, &query)
+        .await
+        .expect("query");
+    let json_rows = result.to_jsonld(&loaded.snapshot).expect("jsonld");
+    let rows = json_rows.as_array().expect("rows");
+    assert!(
+        !rows.is_empty(),
+        "incremental indexing must add new configured values to arena: {rows:?}"
+    );
```
aaj3f, line 1204: It may be worth extending this test to ensure **not only** that the new `performance` scores landed, but also that they didn't stomp the pre-existing `competencies` scores

aaj3f (approved):
I actually saw the symptoms of this w/ that use of `fluree-db-server` I got out for Adaptive, but I didn't know (or have time to investigate) what the cause was. Nice to have it fixed. Just one small note below:

fluree-db-api/tests/it_query_fulltext.rs
bplatz: @aaj3f addressed din 658c482ecab0cb626128b42ad17fcc7e0853b756

PR #1194, opened 2026-04-23
remote reindex support

bplatz:
## Summary                                                                         
  
  - Adds `fluree reindex --remote <name>` plus auto-routing through a locally-running
   server (same triad pattern as `fluree create` / `fluree branch`).
  - Adds `POST /reindex` to `fluree-db-server`, admin-auth-gated like `/create` and  
  `/drop`, returning `{ ledger_id, index_t, root_id, stats }`.                       
  - Documents the `/reindex` contract in `docs/cli/server-integration.md` so Solo and
   other server implementers have a fixed shape to target.                           
                                                            
  ## Motivation                                                                      
                                                            
  `fluree reindex` previously only worked against the local store — there was no CLI 
  flag, no auto-routing, and no server endpoint. That left a gap: once a ledger is
  being served by a Fluree HTTP server (or Solo), there was no CLI-compatible way to 
  trigger a full reindex remotely. This PR closes that gap end-to-end in `fluree/db`
  and defines the contract that downstream servers can implement to be compatible.

  ## What's in the request body

  ```json
  { "ledger": "mydb:main", "opts": {} }
  ```                                                                                
  
  `opts` is accepted but ignored today. It's reserved so that we can add per-request 
  indexer overrides later (e.g. leaflet tuning) without a breaking contract change.
  The server always reindexes using its own configured indexer settings for now — a  
  deliberate choice: CLI users against a remote server get the server's tuning, which
   is usually what you want.

  ## Response shape

  Mirrors `fluree_db_api::ReindexResult`. `ReindexResult` is the source of truth; new
   fields should be added additively.
                                                                                     
  ```json                                                                            
  {
    "ledger_id": "mydb:main",                                                        
    "index_t": 42,                                          
    "root_id": "fluree:index:sha256:...",
    "stats": { "flake_count": 0, "leaf_count": 0, "branch_count": 0, "total_bytes": 0
   }                                                                                 
  }                                                                                  
  ```                                                                                
                                                            
  ## Files changed

  **CLI** (`fluree-db-cli`)                                                          
  - `src/cli.rs` — `--remote <name>` on `Reindex`.
  - `src/commands/index.rs` — `run_reindex` routes via explicit `--remote` →         
  `try_server_route` auto-route → local.                                             
  - `src/lib.rs` — dispatch plumbing.                                                
  - `src/remote_client.rs` — new `RemoteClient::reindex(&alias)`.                    
                                                                                     
  **Server** (`fluree-db-server`)                                                    
  - `src/routes/mod.rs` — registered `POST /reindex` in `v1_admin_protected_routes`. 
  - `src/routes/ledger.rs` — `reindex` handler,                                      
  `ReindexRequest`/`ReindexResponse`/`ReindexStats`, peer-mode forwarding.           
                                                                                     
  **Docs**                                                                           
  - `docs/cli/server-integration.md` — added `## /reindex Contract` section.

### DIFF
--- a/fluree-db-cli/src/commands/index.rs
+++ b/fluree-db-cli/src/commands/index.rs
@@ -2,6 +2,7 @@ use crate::context::{self, build_fluree, LedgerMode};
 use crate::error::{CliError, CliResult};
 use colored::Colorize;
 use fluree_db_api::server_defaults::FlureeDir;
+use fluree_db_api::wire::ReindexResponse;
 use fluree_db_api::ReindexOptions;
 
 /// Run incremental indexing for a ledger.
@@ -127,19 +128,9 @@ pub async fn run_reindex(
     Ok(())
 }
 
-fn print_reindex_result(result: &serde_json::Value) {
-    let ledger_id = result
-        .get("ledger_id")
-        .and_then(|v| v.as_str())
-        .unwrap_or("(unknown)");
-    let index_t = result
-        .get("index_t")
-        .and_then(serde_json::Value::as_i64)
-        .unwrap_or(0);
-    let root_id = result
-        .get("root_id")
-        .and_then(|v| v.as_str())
-        .unwrap_or("(unknown)");
-
-    println!("Reindexed {ledger_id} to t={index_t} (root: {root_id})");
+fn print_reindex_result(result: &ReindexResponse) {
+    println!(
+        "Reindexed {} to t={} (root: {})",
+        result.ledger_id, result.index_t, result.root_id
+    );
 }
--- a/fluree-db-cli/src/commands/index.rs
+++ b/fluree-db-cli/src/commands/index.rs
@@ -1,4 +1,4 @@
-use crate::context::{self, build_fluree};
+use crate::context::{self, build_fluree, LedgerMode};
 use crate::error::{CliError, CliResult};
 use colored::Colorize;
 use fluree_db_api::server_defaults::FlureeDir;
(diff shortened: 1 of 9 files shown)


fluree-db-cli/src/remote_client.rs
aaj3f, line 243: I wonder if we need to either increase this or establish a separate timeout for remote reindex, as a valid remote reindex **could** extend beyond 5min

fluree-db-server/src/routes/ledger.rs
```diff
@@ -335,6 +335,116 @@ async fn drop_local(state: Arc<AppState>, request: Request) -> Result<Json<DropR
...
+                flake_count: result.stats.flake_count,
+                leaf_count: result.stats.leaf_count,
+                branch_count: result.stats.branch_count,
+                total_bytes: result.stats.total_bytes,
+            },
+        }))
+    }
+    .instrument(span)
+    .await
+}
```
aaj3f, line 405: I have been wondering even before this PR about whether we should have locks / mutual exclusion for invoking a reindex in the middle of another ongoing/in-process reindex. I think we have this at publish-time already, but it still currently (if I'm correct about the pathways and logic) would have the loser in an index publish race having already done the reindex work entirely first. We could likely add the responsibility for this (particularly on `--remote`) to `solo` instead, but we could also establish it here in `db` in ways that could abstract over `solo`'s patterns for dynamo nameservices etc.

fluree-db-cli/src/commands/index.rs
```diff
@@ -57,30 +57,89 @@ pub async fn run_index(ledger: Option<&str>, dirs: &FlureeDir) -> CliResult<()>
...
+        .and_then(|v| v.as_str())
+        .unwrap_or("(unknown)");
+    let index_t = result
+        .get("index_t")
+        .and_then(serde_json::Value::as_i64)
+        .unwrap_or(0);
+    let root_id = result
+        .get("root_id")
+        .and_then(|v| v.as_str())
+        .unwrap_or("(unknown)");
```
aaj3f: It might be nice to have a struct like `ReindexResult`, maybe even in a lower-level crate. I think the behavior is likely to be shared across at least `fluree-db-server` and rather than having them both do brittle `Value::get()` calls, we could strengthen the typing and avoid drift.

aaj3f (approved):
Glad to have this. In a few time crunch scenarios w/ Boomi I was having to take the remote auth token from fluree CLI and then do manual HTTP curls to reindex remote ledgers. A few notes below.

fluree-db-cli/src/remote_client.rs
bplatz: Addressed in 8d8605117d4603e63e4b9e37b5e05b4c393f95b1 moved to 1h

fluree-db-server/src/routes/ledger.rs
bplatz: I agree this would be useful but we don't have it today and it would be a larger effort so I'd like to defer.

fluree-db-cli/src/commands/index.rs
bplatz: Addressed in 8d8605117d4603e63e4b9e37b5e05b4c393f95b1

PR #1196, opened 2026-04-24
feat: Ontology imports via `f:schemaSource` + `owl:imports`

bplatz:
\
- Reasoning now sees class/property axioms that live outside the query's graph. Configure `f:schemaSource` on `f:reasoningDefaults` to point at an ontology graph, optionally set `f:followOwlImports true` to transitively resolve `owl:imports` across same-ledger graphs, and optionally provide an `f:ontologyImportMap` to bind external ontology IRIs to local graphs.
- Existing RDFS / OWL2-RL / OWL2-QL extraction code is unchanged. A new `SchemaBundleOverlay` projects whitelisted schema triples from every source graph onto `g_id=0`, so reasoners that scan the default graph pick them up automatically.
- End-to-end tests prove OWL2-RL rule firing (TransitiveProperty, inverseOf, rdfs:domain) when the axiom is declared in a transitively imported ontology.

## Problem

`f:schemaSource` was declared in the ledger-config schema but effectively ignored — `ReasoningDefaults.schema_source` was parsed into the type and never consulted. Reasoning always read schema from the query's graph (usually the default graph), so there was no way to:

- Keep instance data in one graph and the ontology in another.
- Share an ontology across multiple data graphs in a ledger.
- Use OWL's native `owl:imports` to compose ontologies.

The `reasoning_prep` / OWL2-RL code paths hardcoded `g_id=0` or `db.g_id`, so simply extending the config wiring wasn't enough — schema triples from other graphs had to be made visible to code that only looks at one graph.

## Fix

Three pieces, composed:

### 1. Resolver (`fluree-db-api/src/ontology_imports.rs`)

Given `ReasoningDefaults`, walk from `f:schemaSource` through the transitive `owl:imports` closure and return a `ResolvedSchemaBundle { ledger_id, to_t, sources: Vec<GraphId> }`.

Each `owl:imports <X>` triple is resolved in this order:

1. `<X>` is a named graph IRI in the current ledger → use that `GraphId`.
2. `<X>` has an entry in `f:ontologyImportMap` → resolve via the bound `GraphSourceRef`.
3. Otherwise, hard error with `ApiError::OntologyImport`. No silent skip.

Walk is BFS with dedup-by-resolved-`GraphId`, so cycles terminate naturally.

All resolutions flow through a single `resolve_local_graph_source` chokepoint that enforces:

- **Same-ledger.** Cross-ledger `GraphSourceRef` targets are rejected.
- **No reserved system graphs.** Resolving to `g_id=1` (txn-meta) or `g_id=2` (config) is always an error, regardless of resolution path (direct match, mapping table, or selector).
- **No unhonored `GraphSourceRef` fields.** `f:atT`, `f:trustPolicy`, `f:rollbackGuard` are rejected with clear errors rather than silently ignored — the bundle resolves at the query's single `to_t` with no trust verification.

Resolutions are cached in a process-global `SchemaBundleCache` (moka) keyed by `(ledger_id, to_t, starting_g_id, follow_imports)`. Because any config change in the ledger advances `t`, old entries naturally invalidate via LRU; no explicit invalidation needed.

### 2. Overlay (`fluree-db-query/src/schema_bundle.rs`)

`SchemaBundleOverlay` is an `OverlayProvider` wrapper that exposes whitelisted schema flakes from the bundle's source graphs **as if they lived in `g_id=0`**. This lets existing reasoning code run unchanged:

- For `g_id != 0`, delegates straight to the base overlay.
- For `g_id == 0`, linear-merges base flakes with bundle flakes in index order.

Flakes are materialized up-front via `build_schema_bundle_flakes`:

- One PSOT scan per schema predicate (`rdfs:subClassOf`, `rdfs:subPropertyOf`, `rdfs:domain`, `rdfs:range`, `owl:inverseOf`, `owl:equivalentClass`, `owl:equivalentProperty`, `owl:sameAs`, `owl:imports`).
- One OPST scan per schema class (`rdf:type ?` where the class is `owl:Class`, `owl:ObjectProperty`, `owl:DatatypeProperty`, `owl:SymmetricProperty`, `owl:TransitiveProperty`, `owl:FunctionalProperty`, `owl:InverseFunctionalProperty`, `owl:Ontology`, `rdf:Property`).

Results are sorted into the four index orderings and stored as `Arc<[Flake]>`.

### 3. Wiring (`fluree-db-api/src/view/{query,dataset_query}.rs`)

`build_executable_for_view` → `attach_schema_bundle` runs per query:

1. Short-circuits when `"reasoning": "none"` — a query that opted out of reasoning must not fail because of an unrelated broken import in the ledger's config.
2. Short-circuits when there's no `f:schemaSource` — legacy default-graph reasoning applies unchanged.
3. Otherwise calls `resolve_schema_bundle` (cached), then `build_schema_bundle_flakes`, then sets `executable.options.schema_bundle`.

`prepare_execution` in `fluree-db-query` detects `schema_bundle` on the options and wraps `db.overlay` in a `SchemaBundleOverlay` for the reasoning-prep block. Hierarchy extraction, OWL2-RL materialization, and OWL2-QL ontology loading all see the full closure.

## Schema-triple whitelist

Only entailment-relevant predicates are projected from import graphs:

- **RDFS:** `subClassOf`, `subPropertyOf`, `domain`, `range`
- **OWL:** `inverseOf`, `equivalentClass`, `equivalentProperty`, `sameAs`, `imports`
- **`rdf:type`** only when the object is one of: `owl:Class`, `owl:ObjectProperty`, `owl:DatatypeProperty`, `owl:SymmetricProperty`, `owl:TransitiveProperty`, `owl:FunctionalProperty`, `owl:InverseFunctionalProperty`, `owl:Ontology`, `rdf:Property`.

Anything else in an import graph — in particular, instance data — does not surface in the reasoner's view. Covered by the `instance_data_in_schema_graph_does_not_leak` regression test.

## Example config

```turtle
@prefix f:    <https://ns.flur.ee/db#> .
@prefix owl:  <http://www.w3.org/2002/07/owl#> .
@prefix rdf:  <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .

GRAPH <urn:fluree:myapp:main#config> {
  <urn:myapp:config> a f:LedgerConfig ;
    f:reasoningDefaults <urn:myapp:config:reasoning> .

  <urn:myapp:config:reasoning>
    f:reasoningModes ( "rdfs" "owl2-rl" ) ;
    f:schemaSource <urn:myapp:config:schema-ref> ;
    f:followOwlImports true ;
    f:ontologyImportMap <urn:myapp:config:bfo-binding> .

  <urn:myapp:config:schema-ref> a f:GraphRef ;
    f:graphSource [ f:graphSelector <http://example.org/ontology/core> ] .

  <urn:myapp:config:bfo-binding>
    f:ontologyIri <http://purl.obolibrary.org/obo/bfo.owl> ;
    f:graphRef [ a f:GraphRef ;
                 f:graphSource [ f:graphSelector <http://example.org/ontology/local/bfo> ] ] .
}
```

## Tests

`fluree-db-api/tests/it_reasoning_imports.rs` covers:

- Same-ledger auto resolution of a named schema source.
- Transitive `A → B` with a subclass edge in `B`.
- Mapping table fallback for external IRIs.
- Unresolved imports surface as `ApiError::OntologyImport`.
- Cycle `A → B → A` terminates and still yields the correct closure.
- Mapping entries that would target a reserved system graph are rejected.
- `"reasoning": "none"` queries skip resolution entirely (no spurious errors from unrelated config).
- `f:atT` on a `GraphRef` is rejected with a clear message.
- Instance data in the schema graph does **not** leak into query results.
- **End-to-end OWL2-RL rule firing through a transitive import:**
  - `owl:TransitiveProperty` in imported graph B → transitive closure computed correctly.
  - `owl:inverseOf` in imported graph B → inverse query returns entailed bindings (base data has zero direct matches).
  - `rdfs:domain` in imported graph B → instance typed with the target class even though the data has no `rdf:type` triple.

Module-level unit tests cover cache keys, empty-bundle passthrough, and non-default-graph delegation.

## Files

**Core**
- `fluree-db-core/src/ledger_config.rs` — `ReasoningDefaults` gets `follow_owl_imports`, `ontology_import_map: Vec<OntologyImportBinding>`.
- `fluree-db-core/src/namespaces.rs` — `is_schema_predicate`, `is_schema_class`, `is_owl_imports` helpers (+ re-exports in `lib.rs`).

**Vocab**
- `fluree-vocab/src/lib.rs` — `owl::{IMPORTS, ONTOLOGY, CLASS, OBJECT_PROPERTY, DATATYPE_PROPERTY}`, `rdf::PROPERTY`, predicate local names, `config_iris::{FOLLOW_OWL_IMPORTS, ONTOLOGY_IMPORT_MAP, ONTOLOGY_IRI, GRAPH_REF_PROP}`.

**API**
- `fluree-db-api/src/ontology_imports.rs` (new) — resolver, closure walk, cache.
- `fluree-db-api/src/config_resolver.rs` — parses the two new reasoning fields, new `read_ontology_import_map`.
- `fluree-db-api/src/error.rs` — `ApiError::OntologyImport` variant.
- `fluree-db-api/src/view/{query,dataset_query}.rs` — `build_executable_for_view` is async; calls `attach_schema_bundle`.

**Query**
- `fluree-db-query/src/schema_bundle.rs` (new) — `SchemaBundleFlakes`, `SchemaBundleOverlay`, `build_schema_bundle_flakes`.
- `fluree-db-query/src/options.rs` — `QueryOptions::schema_bundle` field + `with_schema_bundle`.
- `fluree-db-query/src/execute/runner.rs` — wraps base overlay with `SchemaBundleOverlay` when a bundle is attached.

**Tests & Docs**
- `fluree-db-api/tests/it_reasoning_imports.rs` (new) — 12 integration tests.
- `docs/design/ontology-imports.md` (new), linked from `docs/SUMMARY.md`, `docs/design/README.md`, `CLAUDE.md`.

### DIFF
--- a/fluree-db-api/src/view/dataset_query.rs
+++ b/fluree-db-api/src/view/dataset_query.rs
@@ -527,17 +527,13 @@ impl Fluree {
         else {
             return Ok(());
         };
-        let flakes = fluree_db_query::schema_bundle::build_schema_bundle_flakes(
+        let flakes = crate::ontology_imports::get_or_build_schema_bundle_flakes(
             db_ref.snapshot,
             db_ref.overlay,
-            db_ref.t,
-            &bundle.sources,
+            &bundle,
         )
-        .await
-        .map_err(|e| {
-            ApiError::OntologyImport(format!("failed to materialize schema bundle: {e}"))
-        })?;
-        executable.options.schema_bundle = Some(std::sync::Arc::new(flakes));
+        .await?;
+        executable.options.schema_bundle = Some(flakes);
         Ok(())
     }
 
--- a/fluree-db-api/src/view/dataset_query.rs
+++ b/fluree-db-api/src/view/dataset_query.rs
@@ -100,7 +100,7 @@ impl Fluree {
         super::query::maybe_wrap_for_graph_source(primary, &mut parsed);
 
         // 2. Build executable with optional reasoning override from primary view
-        let executable = self.build_executable_for_dataset(dataset, &parsed)?;
+        let executable = self.build_executable_for_dataset(dataset, &parsed).await?;
 
         // 3. Get tracker for fuel limits
         let tracker = match &input {
@@ -192,7 +192,7 @@ impl Fluree {
         super::query::maybe_wrap_for_graph_source(primary, &mut parsed);
 
         // 2. Build executable with optional reasoning override from primary view
-        let executable = self.build_executable_for_dataset(dataset, &parsed)?;
+        let executable = self.build_executable_for_dataset(dataset, &parsed).await?;
 
         // 3. Get tracker for fuel limits
         let tracker = match &input {
@@ -286,6 +286,7 @@ impl Fluree {
         // Build executable
         let executable = self
             .build_executable_for_dataset(dataset, &parsed)
+            .await
             .map_err(|e| {
                 crate::query::TrackedErrorResponse::new(400, e.to_string(), tracker.tally())
             })?;
@@ -399,6 +400,7 @@ impl Fluree {
 
         let executable = self
             .build_executable_for_dataset(dataset, &parsed)
+            .await
             .map_err(|e| {
                 crate::query::TrackedErrorResponse::new(400, e.to_string(), tracker.tally())
             })?;
@@ -465,8 +467,10 @@ impl Fluree {
 
     /// Build an ExecutableQuery for dataset queries.
     ///
-    /// Applies reasoning from the primary view if set.
-    fn build_executable_for_dataset(
+    /// Applies reasoning from the primary view if set. When reasoning config
+    /// on the primary view declares `f:schemaSource`, resolves the schema
+    /// bundle closure and attaches it to `executable.options.schema_bundle`.
+    async fn build_executable_for_dataset(
         &self,
         dataset: &DataSetDb,
         parsed: &fluree_db_query::parse::ParsedQuery,
@@ -485,9 +489,56 @@ impl Fluree {
                     executable.options.reasoning = effective.clone();
                 }
             }
+
+            // Resolve schema bundle against the primary view's ledger
+            // (same-ledger only). Mirrors the single-view path in
+            // `view/query.rs::attach_schema_bundle`; see that method for the
+            // reasoning-disabled short-circuit rationale.
+            Self::attach_dataset_schema_bundle(primary, &mut executable).await?;
         }
 
         Ok(executable)
+    }
+
+    async fn attach_dataset_schema_bundle(
+        primary: &crate::view::GraphDb,
+        executable: &mut ExecutableQuery,
+    ) -> Result<()> {
+        if executable.options.reasoning.is_disabled() {
+            return Ok(());
+        }
+        let Some(resolved) = primary.resolved_config() else {
+            return Ok(());
+        };
+        let Some(reasoning) = resolved.reasoning.as_ref() else {
+            return Ok(());
+        };
+        if reasoning.schema_source.is_none() {
+            return Ok(());
+        }
+        let db_ref = primary.as_graph_db_ref();
+        let Some(bundle) = crate::ontology_imports::resolve_schema_bundle(
+            db_ref.snapshot,
+            db_ref.overlay,
+            db_ref.t,
+            reasoning,
+        )
+        .await?
+        else {
+            return Ok(());
+        };
+        let flakes = fluree_db_query::schema_bundle::build_schema_bundle_flakes(
+            db_ref.snapshot,
+            db_ref.overlay,
+            db_ref.t,
+            &bundle.sources,
+        )
+        .await
+        .map_err(|e| {
+            ApiError::OntologyImport(format!("failed to materialize schema bundle: {e}"))
+        })?;
+        executable.options.schema_bundle = Some(std::sync::Arc::new(flakes));
+        Ok(())
     }
 
     /// Execute against dataset (multi-ledger).
--- a/fluree-db-query/src/schema_bundle.rs
+++ b/fluree-db-query/src/schema_bundle.rs
@@ -1,0 +1,407 @@
+//! Schema-bundle overlay that projects whitelisted ontology flakes into `g_id=0`.
+//!
+//! # Purpose
+//!
+//! When reasoning is configured with an `f:schemaSource` (plus, optionally, a
+//! transitive `owl:imports` closure), RDFS/OWL extraction needs to see
+//! schema triples regardless of which graph they physically live in.
+//!
+//! The existing reasoning extraction code only scans the default graph
+//! (`g_id=0`). Rather than teach every reasoner to scan multiple graphs,
+//! this module materializes the **relevant** triples from each source graph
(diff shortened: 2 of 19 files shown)


fluree-db-api/src/view/dataset_query.rs
```diff
@@ -485,11 +489,58 @@ impl Fluree {
...
+        let flakes = fluree_db_query::schema_bundle::build_schema_bundle_flakes(
+            db_ref.snapshot,
+            db_ref.overlay,
+            db_ref.t,
+            &bundle.sources,
+        )
+        .await
+        .map_err(|e| {
+            ApiError::OntologyImport(format!("failed to materialize schema bundle: {e}"))
+        })?;
```
aaj3f: I wonder if this is worth caching in the same way you also have, in this PR, caching for `SchemaBundleCache`. Otherwise it seems every query (when `schemaSource` is configured) would rebuild the same flakes for the same (static/immutable) `(ledger_id, to_t, sources)` flakes. 

If that's a sound decision pattern, I'm not sure how best to do it. Either by extending `SchemaBundleCache` so key-values are something like `(Arc<ResolvedSchemaBundle>, Arc<SchemaBundleFlakes>)` or by introducing a sibling cache for keys like `(ledger_id, to_t, Vec<GraphId>)`

Same comment / value-prop for `fluree-db-api/src/view/query.rs:564` where we also do `build_schema_bundle_flakes()`

fluree-db-query/src/schema_bundle.rs
```diff
@@ -0,0 +1,407 @@
...
+            return;
+        }
+
+        // Collect base flakes and bundle flakes separately (both are already
+        // in index order), then linear-merge.
+        let mut base_flakes: Vec<Flake> = Vec::new();
+        self.base
+            .for_each_overlay_flake(g_id, index, first, rhs, leftmost, to_t, &mut |f| {
+                base_flakes.push(f.clone());
+            });
```
aaj3f, line 339: I don't *think* this is a hot-path scenario (not reasoning itself, just reasoning-prep) so maybe this note doesn't matter so much, but I do wonder if we could avoid the `Vec` allocation and the `.clone()` of each base flake. I don't know if we could use a `RefCell` or `parking_lot::Mutex` on `SchemaBundleOverlay` so that it could still mutate those flakes while outside callers could easily reuse it? 

Or maybe when we compute `start..end` on bundle slices, if the bundle is empty (`start == end`), we could short-circuit straight to `self.base.for_each_overlay_flake(...)` so that for all the scans where the range doesn't intersect with the bundle we avoid the overhead?

aaj3f (approved):
This is really neat and I'm excited about it! Two notes below to take or leave:

fluree-db-api/src/view/dataset_query.rs
bplatz: Done in 2decc15b3aad7371240a47c9ab6d566ad4679fca

fluree-db-query/src/schema_bundle.rs
bplatz: Did your second option in 2decc15b3aad7371240a47c9ab6d566ad4679fca

PR #1197, opened 2026-04-24
fix: align JSON-LD `["optional", ...]` semantics with SPARQL `OPTIONAL`

bplatz:
Fixes a SPARQL-conformance bug where `OPTIONAL { ... FILTER(...) }` over a multi-valued predicate emitted a spurious null-extended row in addition to the matching row.

The root cause was JSON-LD-specific: the WHERE-clause parser treated each node-map object inside an `["optional", {a}, {b}, ...]` array as a separate `OPTIONAL` clause (`OPTIONAL { a } OPTIONAL { b }`). Per SPARQL 1.1 §8.1, an `OPTIONAL` block is a single conjunctive group graph pattern, so the array should lower to one `LeftJoin` over a conjunctive inner pattern.

## What was happening

The reported repro:

```jsonc
["optional",
  { "@id": "?child", "skosxl:prefLabel": { "@id": "?xlLn" } },
  { "@id": "?xlLn",  "skosxl:literalForm": "?xlLang" },
  ["filter", "(= (lang ?xlLang) \"en\")"]
]
```

…parsed to **two** sibling `Optional`s, producing `LeftJoin(LeftJoin(outer, P1), P2_with_filter)`. For a concept with three multi-language labels, the outer LeftJoin produced three candidate `?xlLn` bindings; for two of them the filter inside the second OPTIONAL rejected every inner row, so the operator emitted a null-extended row per non-match. After `SELECT DISTINCT` this surfaced as `[concept, "Spreadsheet Work"@en]` plus a spurious `[concept, null]`.

The OPTIONAL execution operator (`fluree-db-query/src/optional.rs`) was already correct — it tracks `matched` per required row and only null-extends when no inner row was compatible. The bug was purely upstream, in how the parser grouped the OPTIONAL block.

## Fix

`fluree-db-query/src/parse/where_clause.rs::parse_optional_patterns` now accumulates **all** items in the `["optional", ...]` array into one conjunctive group (one `Pattern::Optional`). To express two independent left joins, write two sibling arrays:

```jsonc
// Before (split into two OPTIONALs implicitly):
["optional", {a}, {b}]
// Now equivalent to SPARQL: OPTIONAL { a . b }  (one LeftJoin, conjunctive inner)

// To get the old "two independent OPTIONALs" behavior, write:
["optional", {a}],
["optional", {b}]
// Equivalent to SPARQL: OPTIONAL { a } OPTIONAL { b }
```

This matches SPARQL semantics exactly and incidentally fixes the secondary "cartesian-join" issue mentioned in the bug report (a fallback OPTIONAL adjacent to one with a filter was also being split into multiple LeftJoins).

## Behavior changes

| Form                                | Before                                  | After                                           |
|-------------------------------------|-----------------------------------------|-------------------------------------------------|
| `["optional", {a}]`                 | one OPTIONAL                            | one OPTIONAL (unchanged)                        |
| `["optional", {a}, ["filter",...]]` | one OPTIONAL with filter (unchanged)    | one OPTIONAL with filter (unchanged)            |
| `["optional", {a}, {b}]`            | **two** independent OPTIONALs (Fluree)  | **one** OPTIONAL with conjunctive inner (SPARQL)|
| `["optional", {a}], ["optional", {b}]` | two independent OPTIONALs            | two independent OPTIONALs (unchanged)           |

Tests updated:

- `fluree-db-query::parse::tests::test_optional_multiple_patterns` — now asserts one OPTIONAL with two triples.
- `it_query_jsonld::jsonld_optional_two_in_same_vector` → renamed to `jsonld_optional_conjunctive_inner`; expected output updated to reflect conjunctive semantics (subjects missing either var get null on both, not just the missing one).
- `it_query_jsonld::jsonld_optional_with_multiple_triples` → renamed to `jsonld_two_independent_optionals`; rewritten with two sibling arrays so the original "independent left joins" intent still has coverage under SPARQL semantics.

New regression test:

- `it_query_jsonld::jsonld_optional_with_filter_multivalue_skosxl` — direct port of the bug-report repro.

### DIFF
--- a/fluree-db-query/src/parse/where_clause.rs
+++ b/fluree-db-query/src/parse/where_clause.rs
@@ -452,16 +452,20 @@ pub fn parse_subquery_patterns(
     Ok(patterns)
 }
 
-/// Parse OPTIONAL patterns with Fluree's grouping semantics
+/// Parse OPTIONAL patterns with SPARQL-canonical grouping semantics
 ///
-/// Fluree semantics: each node-map OBJECT is a SEPARATE optional.
-/// - `["optional", {nodeA}, {nodeB}]` means two separate left joins
-/// - `["optional", {nodeWithMultipleProps}]` is a single conjunctive optional
-/// - `["optional", {node}, ["filter", ...]]` keeps filter with preceding node (conjunctive)
+/// The entire `["optional", ...]` array is a single OPTIONAL block — equivalent
+/// to one SPARQL `OPTIONAL { ... }`. All items inside become a conjunctive group
+/// (a single `LeftJoin` in the algebra).
 ///
-/// Array elements (filters, nested optionals, values, etc.) are accumulated
-/// with the preceding node-map patterns to form conjunctive groups.
-/// Array elements must follow a node-map; ["optional", ["filter", ...], {...}] is invalid.
+/// - `["optional", {a}, {b}]` ≡ `OPTIONAL { a . b }` (one left join, conjunctive inner)
+/// - `["optional", {a}, ["filter", ...]]` ≡ `OPTIONAL { a FILTER(...) }`
+/// - To get two independent left joins, use two sibling arrays:
+///   `["optional", {a}], ["optional", {b}]`.
+///
+/// Filters and binds require a preceding node-map in the group, since they
+/// constrain or compute from existing bindings. Other array forms (nested
+/// optional, values, query, etc.) are self-contained.
 fn parse_optional_patterns(
     items: &[JsonValue],
     ctx: &JsonLdParseCtx,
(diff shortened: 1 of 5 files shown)


fluree-db-query/src/parse/where_clause.rs
```diff
@@ -522,7 +510,6 @@ fn parse_optional_patterns(
                     }
                 }
```
aaj3f, line 514: One thing that might get silently missed here is a pattern like

```jsonc
["optional",
  ["values", ["?x", [1, 2, 3]]],
  ["filter", "(> ?x 0)"]
]
```

Where the `values` would make `current_group.is_empty() == false`. There may be a good reason to have relaxed the check to drop the `current_group.is_empty()` and I *may* be wrong about the above being valid vis-a-vis its SPARQL counterpart, but just wanted to check with you

aaj3f (approved):
Makes sense and good catch. One question in the comments below, and then perhaps we may want even the briefest of updates to `docs/query/jsonld-query.md` re: what "grouped optionals" mean and what the contract around their behavior is (i.e. how to sibling `["optional", ...], ["optional", ...]` clauses behave versus `["optional", ..., ...]`

fluree-db-query/src/parse/where_clause.rs
aaj3f: That's weird... it didn't give the full code context.

Old code:
```rust
if !has_node_map_anchor && current_group.is_empty() {
    // ... reject filter/bind
}
```

New code:
```rust
if !has_node_map_anchor {
    // ... reject filter/bind
}
```

bplatz: Addressed in 03106b63264ef6419efdb118a219d7d7e29ec67e

PR #1200, opened 2026-04-24
fix: branch-aware reads + opt-in default context

bplatz:
Two related cleanups in one PR — both started from observed bugs and both center on "what does the API quietly do that the caller didn't ask for".

1. **Branch reads were broken.** Commits on a branched ledger reference parent commits that live under the source branch's storage namespace, not the new branch's. The commit-chain walker, incremental indexer, pack/transfer, BM25 / vector graph-source refresh, and several index-attach paths all built a flat `backend.content_store(ledger_id)` and 404'd on every pre-fork ancestor. Every branched ledger that received a transaction was unreadable from the first write forward — query cache refresh and indexer retried in a loop.
2. **Default JSON-LD context was eagerly loaded onto every `LedgerState`.** Direct API consumers got silent prefix expansion they never asked for, with no clean way to express "no prefix". The intent was always opt-in at the boundary; the implementation had drifted.

## Branch-aware reads

The codebase already had `BranchedContentStore` and `LedgerState::build_branched_store`, but they were only used on the **initial load / full reload** path. Other call sites built a flat branch-scoped store and handed it to a function that walks history or loads index blobs.

**Helper added** — `fluree-db-nameservice/src/branched_store.rs`:
- `branched_content_store_for_record(backend, ns, &record) -> Arc<dyn ContentStore>`
- `branched_content_store_for_id(backend, ns, ledger_id) -> Arc<dyn ContentStore>`
- `build_branched_store(backend, ns, &record) -> BranchedContentStore` (the recursive ancestry walker, exposed so `fluree-db-ledger` can delegate)

For non-branched ledgers both helpers return the flat namespace store (single nameservice lookup in the `_for_id` variant). For branched ledgers they walk `record.source_branch` recursively to assemble the parent fallback chain.

Sits in `fluree-db-nameservice` (not `fluree-db-core`) so the indexer's `BackgroundIndexerWorker` can reach it without taking a `fluree-db-ledger` dep.

## Default-context opt-in

**Field removed** — `LedgerState.default_context` and `CachedLedgerState.default_context` are gone. The eager loads in `Fluree::ledger()`, `LedgerManager::load_and_attach_binary_store`, and `load_graph_db_impl` are removed.

**Opt-in surface preserved**:
- `Fluree::db(id)` returns a `GraphDb` with `default_context = None` → no auto-injection
- `Fluree::db_with_default_context(id)` and `db_at_with_default_context(id, spec)` fetch the context via the branch-aware helper and attach it to the returned `GraphDb` only
- CLI uses `db_with_default_context` (already did; unchanged)
- HTTP server query endpoint uses `db_with_policy → db()` (was already on the no-default path; unchanged)
- Storage layer (`NsRecord.default_context` CID, `LedgerConfig.default_context` CID, `Fluree::get_default_context` / `set_default_context` admin API) unchanged

**Behavior changes** for the leak readers (all now consistent with the boundary policy):
- `Fluree::query_graph_source` / `sparql_graph_source` (R2RML query path) — no longer auto-inject; only consumed by integration tests in production
- `CommitBuilder::execute` IRI compaction — no longer falls back to ledger default; only honors `user_context`; only consumed by API integration tests

The HTTP server's standard query endpoint never relied on the leak.

## Regression tests

- `it_branch::branch_incremental_index_resolves_pre_fork_parent` — verified to fail with `IndexTimeout(15000)` without the indexer fix; passes in 0.42s with it
- `it_notify_incremental::notify_branch_catch_up_resolves_pre_fork_parent` — verified to fail with the exact production error text (`"load commit during catch-up: Storage error: Failed to read commit envelope … Not found"`) without the catch-up fix
- `it_default_context::*` (3 cases) — verified that test 2 fails when the opt-in gate is removed; all pass with it in place

## Docs

- `docs/design/storage-traits.md` — added section on the new helpers and the rule that any commit-chain or index-blob walker must use them
- `docs/getting-started/quickstart-ledger.md` — corrected the "Setting Default Context" section (transactions never updated default context; only import + explicit `set_default_context` do; spelled out the API vs CLI/server distinction)


fluree-db-api/src/branched_store_helpers.rs
```diff
@@ -0,0 +1,73 @@
+//! Internal helpers that compose the nameservice's branched store helpers
```
zonotope: I think the functions defined here would be better placed at `fluree-db-nameservice::branched_store`.  This module only defines two functions and it already relies heavily on the nameservice crate to do its work, so I don't think the extra module is justified.

fluree-db-indexer/src/orchestrator.rs
```diff
@@ -1122,12 +1153,24 @@ pub struct PostCommitIndexResult {
...
-///
-/// - Uses the provided `storage` for index building and apply.
-/// - Uses `target_t` explicitly (use `CommitReceipt.t`).
-/// - Never fails the commit path; returns status + error string for logging.
-/// - **Applies index even if publish fails** for local correctness.
+/// Build a content store for the post-commit indexing path that walks
+/// branch ancestry on read miss. Uses the live `NsRecord` if the ledger
+/// state's cached record is in sync, otherwise falls back to a lookup by id.
+#[cfg(feature = "embedded-orchestrator")]
+async fn build_post_commit_store(
```
zonotope: I think this function could also move to the branched_store module, perhaps with a different name to both generalize it and differentiate it from the other branched content store helpers that return a flat content store on lookup failure.

Perhaps:
-  `content_store_for_record_or_id` -> returns a flat content store as a fallback to avoid an ns lookup
-  `branched_content_store_for_record_or_id` -> always returns a branched store, even if more ns lookups are necessary

zonotope (approved):
The code looks fine, but I think it could be consolidated. You can take or leave those suggestions as you see fit.

fluree-db-api/src/branched_store_helpers.rs
bplatz: Addressed in b92aca27e5238d54056220e016c5de8dd10e149c

fluree-db-indexer/src/orchestrator.rs
bplatz: addressed in b92aca27e5238d54056220e016c5de8dd10e149c

PR #1201, opened 2026-04-24
cache schema-bundle flakes + short-circuit overlay misses

bplatz:
Summary                                                                                                                                       
                                                            
  Follow-up to #1196 (owl-imports) — a small optimization that was intended to land with that PR but was never pushed before merge.             
   
  Short-circuits SchemaBundleOverlay::for_each_overlay_flake when the requested index range doesn't intersect the schema bundle: slice the bundle first, and if the sub-range is empty, delegate straight to the base provider instead of buffering all base flakes into a Vec for a merge that has nothing to merge against.                                                                                                      
                                                            
  Behavior is unchanged when the bundle does intersect the scan range. Net effect: arbitrary range scans unrelated to schema triples (the common case) skip an unnecessary allocation and copy.

### DIFF
--- a/fluree-db-query/src/schema_bundle.rs
+++ b/fluree-db-query/src/schema_bundle.rs
@@ -330,15 +330,10 @@ impl OverlayProvider for SchemaBundleOverlay<'_> {
             return;
         }
 
-        // Collect base flakes and bundle flakes separately (both are already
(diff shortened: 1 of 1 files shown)


aaj3f (approved):
👍

PR #1203, opened 2026-04-25
fix: history `@t`/`@op` for IRI-valued objects + boolean `@op`

bplatz:
This is based on #1198 branch and continues history query fixes

History queries on the `fix/history-ranges` branch only carried `@t`/`@op`
metadata onto literal-valued objects. Ref-valued predicates (`rdf:type`,
`skos:inScheme`, `skosxl:prefLabel`, etc.) were emitted with `?v` populated
but `?t` and `?op` `null`. The root cause was structural: `Binding::Sid(Sid)` had no metadata
channel, so the `flakes_to_bindings` ref arm and the
`{materialized,late_materialized}_object_binding` helpers all dropped
`flake.t` / `flake.op` whenever the object was a `FlakeValue::Ref`.

This PR fixes the asymmetry at the type layer rather than at individual
call sites, so the invariant *"object bindings carry assertion metadata"*
is uniform across the literal and ref paths. It also flips the on-the-wire
`@op` representation from string (`"assert"` / `"retract"`) to boolean
(`true` / `false`) to match `Flake.op` on disk and avoid a per-row Arc
allocation.

### Binding type — metadata-capable struct variants

`fluree-db-query/src/binding.rs`

- `Binding::Sid` is now `Sid { sid: Sid, t: Option<i64>, op: Option<bool> }`.
- `Binding::EncodedSid` gains the same `t`/`op` Option fields.
- Both fields are intentionally **excluded** from `PartialEq` and `Hash`,
  matching the discipline already in place for `Binding::Lit`. Set
  semantics (joins, DISTINCT, GROUP BY, hash maps) ignore the metadata, so
  a metadata-bearing object binding compares equal to a metadata-free one
  with the same SID.
- New constructors keep call sites concise:
  - `Binding::sid(sid)` — no metadata (subjects, predicates, constants,
    VALUES rows, expression results).
  - `Binding::sid_with_t(sid, t)` — current-state object binding from a
    flake.
  - `Binding::sid_with_t_op(sid, t, op)` — history-mode object binding.
  - `Binding::encoded_sid(...)` and `Binding::encoded_sid_with_t_op(...)`
    mirror these for the late-materialised path.
- Central accessors `Binding::t() -> Option<i64>` and
  `Binding::op() -> Option<bool>` cover all four metadata-bearing variants
  (`Lit`, `EncodedLit`, `Sid`, `EncodedSid`). `eval_t` / `eval_op` route
  through these so the variant list lives in exactly one place.

### Object-construction sites threading `t`/`op`

Object positions only — subject and predicate bindings still emit
`t: None, op: None`, so `T(?s)` / `OP(?p)` keep the previous "no metadata"
semantics rather than inventing a meaning.

- `binary_scan.rs::flakes_to_bindings` — ref arm now uses
  `sid_with_t(...)` for current-state scans (so `T(?v)` works for
  ref-valued predicates outside history mode too) and `sid_with_t_op(...)`
  in history mode.
- `object_binding.rs::{materialized_object_binding, late_materialized_object_binding}`
  — accept an `op: Option<bool>` parameter and propagate it to both the
  `Lit` arm and the `Sid` / `EncodedSid` arm. `IriRef` and `BlankNode`
  decode kinds now carry `t` like the literal kinds do.
- `binding.rs::from_object_with_t_op` — fixed; this is the originally
  reported zero-call-site constructor that motivated the bug write-up.
  Both the literal and ref arms now thread `t`/`op` through.

### `OP()` returns boolean (`true` / `false`)

`Flake.op` is a boolean on disk. The previous string surface required a
per-row `Arc::from("assert" | "retract")` allocation in expression
evaluation and didn't reflect storage. Now:

- `eval_op` returns `ComparableValue::Bool(op)`.
- The parser accepts:
  - `"@op": "?var"` — generates `BIND(op(?v) AS ?var)`.
  - `"@op": true` / `"@op": false` — generates
    `FILTER(op(?v) = true|false)`.
- String constants (`"assert"`, `"retract"`) are no longer accepted; the
  parser returns a clear error pointing users at the boolean form. (No
  backward-compatibility shim — product is pre-release.)

`fluree-db-query/src/expression/fluree.rs`,
`fluree-db-query/src/parse/node_map.rs` (new internal `OpAnnotation` enum
with `Variable(Arc<str>)` / `Constant(bool)` variants).

### Parser: `@t` / `@op` permitted with `@type: "@id"`

Removed the parser-level rejection that forbade `@t` / `@op` annotations on
explicitly IRI-typed value objects. Now that ref-valued bindings carry the
metadata too, the parser-generated `BIND(t(?v) AS ?t)` /
`BIND(op(?v) AS ?op)` resolve correctly.

### Tests

`fluree-db-api/tests/it_query_history_range.rs`

- All existing literal-valued cases migrated to bool form (`?op` is now
  `bool` in the flattened-row tuples, ordered as `false < true`).
- Two new IRI-object regression cases against `ex:knows`:
  - `history_range_iri_object_sidecar_plus_base` — verifies the
    sidecar+base path threads `t`/`op` through the ref arm.
  - `history_range_iri_object_novelty_only` — verifies the novelty branch
    of the history collector does too.
- Constant-filter tests now use `"@op": true` / `"@op": false`.

### DIFF
--- a/docs/api/endpoints.md
+++ b/docs/api/endpoints.md
@@ -989,16 +989,18 @@ Query the history of entities using the standard `/query` endpoint with `from` a
 ```
 
 The `@t` and `@op` annotations capture transaction metadata:
-- **@t** - Transaction time when the value was asserted or retracted
-- **@op** - Operation type: `"assert"` or `"retract"`
+- **@t** - Transaction time (integer) when the fact was asserted or retracted.
+- **@op** - Operation type as a boolean: `true` for assertions, `false` for retractions. (Mirrors `Flake.op` on disk; constants `"assert"` / `"retract"` are not accepted.)
+
+Both annotations work uniformly for literal-valued and IRI-valued objects.
 
 **Response:**
 
 ```json
 [
-  ["Alice", 30, 1, "assert"],
-  ["Alice", 30, 5, "retract"],
-  ["Alicia", 31, 5, "assert"]
+  ["Alice", 30, 1, true],
+  ["Alice", 30, 5, false],
+  ["Alicia", 31, 5, true]
 ]
 ```
 
--- a/docs/concepts/time-travel.md
+++ b/docs/concepts/time-travel.md
@@ -114,13 +114,13 @@ History queries capture both the retraction and assertion with `@op`:
 
 ```json
 [
-  [25, 1, "assert"],
-  [25, 5, "retract"],
-  [26, 5, "assert"]
+  [25, 1, true],
+  [25, 5, false],
+  [26, 5, true]
 ]
 ```
 
-Each row shows `[value, transaction_time, operation]`.
+Each row shows `[value, transaction_time, op]` where `op` is `true` for assertions and `false` for retractions.
 
 ### Valid Time vs Transaction Time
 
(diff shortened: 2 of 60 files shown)


zonotope (approved):
🔦

PR #1204, opened 2026-04-25
feat: read-only branch merge preview (`merge_preview` / `branch diff`)

bplatz:
Adds a read-only "what would a merge look like?" primitive across all three
surfaces — Rust API, HTTP, and CLI — that returns the rich diff between two
branches (ahead/behind commits, common ancestor, conflict keys, fast-forward
eligibility) **without mutating any state**. API callers can consumes it for
branch UI; the standalone server exposes it as an HTTP route; the CLI
ships a `branch diff` subcommand with `--remote` support.

The implementation reuses the same primitives `merge_branch` already calls,
factored into `fluree-db-core` so they're reusable for git-log-style
viewers, indexer status panels, peer-mode diagnostics, and other tooling.

### `fluree-db-core` — new reusable primitives

- **`CommitSummary`** struct: `t`, `commit_id`, `time`, `asserts`,
  `retracts`, `flake_count`, `message: Option<String>` (extracted from the
  `f:message` `txn_meta` entry when present).
- **`commit_to_summary(&Commit) -> CommitSummary`** — pure decoder, no I/O.
- **`walk_commit_summaries(store, head, stop_at_t, max) -> Result<(Vec<CommitSummary>, usize)>`**
  — newest-first DAG walk that returns both the (capped) summary list and
  the unbounded total count. Uses byte-range envelope reads for the BFS
  plus one full `load_commit_by_id` per summary returned.
- **`ConflictKey`** gained `Serialize`, `Ord`, `PartialOrd` derives so
  capped responses can be sorted before truncation (stable across builds /
  paginated requests — `HashSet::intersection` order is otherwise
  unspecified).

8 new unit tests cover counting, `f:message` extraction (including a
wrong-namespace negative case), walk ordering, `stop_at_t` exclusivity,
truncation, and merge-commit DAGs.

### `fluree-db-api` — new public surface

- `Fluree::merge_preview(ledger, source, target?) -> MergePreview`
- `Fluree::merge_preview_with(ledger, source, target?, opts) -> MergePreview`
- Types: `MergePreview`, `BranchDelta`, `ConflictSummary`, `AncestorRef`,
  `MergePreviewOpts`
- Constants: `DEFAULT_MAX_COMMITS = 500`, `DEFAULT_MAX_CONFLICT_KEYS = 200`
- Re-exports for embedded callers: `CommitSummary`, `CommonAncestor`,
  `ConflictKey`, `commit_to_summary`, `walk_commit_summaries`,
  `find_common_ancestor`

Mirrors `merge_branch_inner` for source/target resolution and the
fast-forward predicate, but skips publish/copy steps. Walks both sides
with `tokio::try_join!`.

16 integration tests covering: fast-forward, diverged-no-conflicts,
diverged-with-conflicts, equal heads, behind-only, default target,
self-merge rejection, truncation, `include_conflicts=false`, read-only
invariant (pre/post nameservice equality), `main`-as-source rejection,
nonexistent source, **sibling-branch ancestor lookup** (regression),
`max_commits=None` unbounded, default opts carry caps, conflict keys
sorted.

### `fluree-db-server` — HTTP route

`GET /fluree/merge-preview/*ledger?source=&target=&max_commits=&max_conflict_keys=&include_conflicts=`

- Auth pattern matches `list_branches` (Bearer required when
  `data_auth.required`; `404` not `403` when bearer can't read).
- Server-side hard caps (`MERGE_PREVIEW_HARD_MAX_COMMITS = 5_000`,
  `MERGE_PREVIEW_HARD_MAX_CONFLICT_KEYS = 5_000`) clamp caller values via
  `n.min(HARD_MAX)` so `max_commits=10000000` collapses to 5000.
- Defaults applied when query params absent (500 / 200).

### `fluree-db-cli` — `branch diff` subcommand

```
fluree branch diff <source> [--target <name>] [--max-commits N]
                            [--max-conflict-keys N] [--no-conflicts]
                            [--json] [--ledger <name>] [--remote <name>]
```

Mirrors the `rebase` / `merge` command structure: supports local-mode
direct execution, tracked-ledger auto-routing, and explicit `--remote`.
Default output is human-readable; `--json` emits the raw `MergePreview`.

`RemoteClient::merge_preview` issues `GET /merge-preview/{ledger}` via
`op_url` + `encode_ledger_path`, so ledger names with spaces / `?` / `#` /
`%` produce well-formed URLs (regression test included).

### Defaults vs. unbounded

`MergePreviewOpts::default()` carries `Some(500)` / `Some(200)` caps so
the no-arg `merge_preview()` matches the spec defaults. `None` is reserved
as an explicit "unbounded" signal for direct Rust callers (e.g., tooling
that needs the full divergence). The HTTP layer constructs from
`default()` and overrides only when the query param is present, so
unbounded responses cannot leak over the wire.

### Cross-branch ancestor lookup

`find_common_ancestor` needs to load commit envelopes from **both**
branches' namespaces. The first cut walked through `source_store` only,
which fails when the target is a sibling branch with commits in its own
namespace. Fixed by building a union view —
`BranchedContentStore::with_parents(source_store_inner, [target_branched])` —
used only for the ancestor walk; per-side walks and delta computation
still use their own dedicated stores.

### Stable conflict-key ordering

`HashSet::intersection` returns items in arbitrary order. After
truncation, capped responses would surface a non-deterministic subset.
Sort lexicographically by `(s, p, g)` before truncating. `ConflictKey`
gained `Ord`/`PartialOrd` derives.

### Cap semantics — what they protect

`max_commits` and `max_conflict_keys` bound **response size + per-summary
`load_commit_by_id` reads**, not the cost of the underlying envelope BFS
or `compute_delta_keys` walks. The unbounded `count` on each side is
still computed — a 1M-commit divergence costs 1M envelope reads
regardless of cap. Documented prominently in rustdoc, the HTTP doc, and
the server-integration contract; `include_conflicts=false` is the
fast-preview escape hatch for diverged branches.

This is intentional — actual divergence rejection (refusing to preview a
branch that diverged 1M commits ago) is a separate operational concern;
the contract calls out where to add such a guard if needed.

## Documentation

- **`docs/api/endpoints.md`** — new `GET /fluree/merge-preview/{ledger}`
  section with request/response schema, error codes, examples. Cap
  columns explicitly note "bounds response size, not divergence-walk
  cost."
- **`docs/cli/server-integration.md`** — new `## Merge Preview Contract`
  section under the existing custom-server contract doc, alongside
  `## Policy Enforcement Contract` and `## Replication Auth Contract`.
  Spells out the 9 required semantic rules custom servers must honor
  (cross-branch ancestor walk, sort-before-truncate, mandatory
  server-side cap, etc.) plus the wire encoding for `ConflictKey`.
- **`docs/getting-started/rust-api.md`** — new `### Branch Diff
  (Merge Preview)` section under "Advanced Usage" with embedded usage
  examples, response-shape table, and pointers to the reusable
  primitives.

### DIFF
--- a/docs/api/endpoints.md
+++ b/docs/api/endpoints.md
@@ -1910,6 +1910,85 @@ curl -X POST http://localhost:8090/v1/fluree/merge \
   -d '{"ledger": "mydb", "source": "dev", "target": "main"}'
 ```
 
+### GET /fluree/merge-preview/{ledger}
+
+Read-only preview of merging a source branch into a target branch. Returns the rich diff — ahead/behind commit summaries, conflict keys, and fast-forward eligibility — without mutating any nameservice or content store state.
+
+Bearer token required when `data_auth.mode = required`; reads are gated on `bearer.can_read(ledger)`.
+
+**URL:**
+```
+GET /fluree/merge-preview/{ledger-name}?source={source}&target={target}&max_commits={n}&max_conflict_keys={n}&include_conflicts={bool}
+```
+
+**Path / Query Parameters:**
+
+| Parameter | Type | Required | Description |
+|-----------|------|----------|-------------|
+| `ledger` (path) | string | Yes | Ledger name (e.g., "mydb") |
+| `source` | string | Yes | Source branch to merge from (e.g., "feature-x") |
+| `target` | string | No | Target branch (defaults to the source's parent branch) |
+| `max_commits` | number | No | Cap on per-side commit summaries returned (default 500). Server clamps to a hard maximum of 5,000 — values above are silently lowered. Bounds response size, **not** divergence-walk cost (the unbounded `count` is still computed). |
+| `max_conflict_keys` | number | No | Cap on conflict keys returned (default 200). Server clamps to a hard maximum of 5,000. Bounds response size, **not** the conflict-delta walks. |
+| `include_conflicts` | bool | No | When false, skips the conflict computation (default true). Use this to make the preview cheap on diverged branches. |
+
+**Response body (200 OK):**
+
+```json
+{
+  "source": "feature-x",
+  "target": "main",
+  "ancestor": { "commit_id": "bafy...", "t": 5 },
+  "ahead": {
+    "count": 3,
+    "commits": [
+      { "t": 8, "commit_id": "bafy...", "time": "2026-04-25T12:00:00Z",
+        "asserts": 2, "retracts": 0, "flake_count": 2, "message": null }
+    ],
+    "truncated": false
+  },
+  "behind": { "count": 1, "commits": [...], "truncated": false },
+  "fast_forward": false,
+  "conflicts": {
+    "count": 0,
+    "keys": [],
+    "truncated": false
+  }
+}
+```
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `source` | string | Source branch name |
+| `target` | string | Target branch name (resolved from default when not supplied) |
+| `ancestor` | object \| null | Common ancestor `{commit_id, t}`. `null` when both heads are absent |
+| `ahead` | object | Commits on source not on target (`count`, `commits`, `truncated`) |
+| `behind` | object | Commits on target not on source |
+| `fast_forward` | bool | True when target HEAD == ancestor (or both heads absent) |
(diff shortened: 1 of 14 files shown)


zonotope (approved):
🇼🇫

PR #1209, opened 2026-04-29
Fix: `rdfs:Class` COUNT Misses

### LINKED ISSUE #1208: COUNT `?s rdf:type rdfs:Class` returns 0 on indexed data when `rdfs:Class` is mid-leaflet
## Symptom

After bulk-importing an RDFS-style ontology, a `COUNT(?c) WHERE { ?c a rdfs:Class }` query returns **0** even though the triples are present in the index. The same data is reachable via other query shapes:

| # | Query | Result |
|---|-------|--------|
| A | `?c a rdfs:Class` (prefixed) | **0** ❌ |
| B | `?c <rdf:type-IRI> <rdfs:Class-IRI>` (full IRIs) | **0** ❌ |
| C | `?c rdf:type ?t . FILTER(?t = rdfs:Class)` (FILTER rewrite) | correct count ✅ |
| D | `?s rdf:type ?t GROUP BY ?t` (aggregate) | shows the correct rdfs:Class count ✅ |
| E | `<some-instance> ?p ?o` (describe a class instance) | yields `rdf:type rdfs:Class` ✅ |
| F | `?p a rdf:Property` (parallel of A for `rdf:Property`) | correct count ✅ |
| G | `?c a <user-class>` (parallel of A for a user class) | correct count ✅ |

`?s rdf:type rdf:Property` and arbitrary user classes work — only `rdfs:Class` is invisible to the bound-object form.

## Reproduction

Bulk-import an RDFS ontology declaring multiple classes via `a rdfs:Class` plus a moderate amount of instance data, then run a `COUNT` query against `?c a rdfs:Class`.

```turtle
@prefix ex:   <http://example.org/ns/> .
@prefix exd:  <http://example.org/data/> .
@prefix rdf:  <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix xsd:  <http://www.w3.org/2001/XMLSchema#> .

ex:ClassA  a rdfs:Class .
ex:ClassB  a rdfs:Class .
ex:ClassC  a rdfs:Class .
# … (more classes) …

ex:propX a rdf:Property ; rdfs:domain ex:ClassA ; rdfs:range xsd:string .
# … (more properties) …

# Plus several thousand instance triples
exd:item-1 a ex:ClassA ; ex:propX "value-1" .
# …
```

```bash
fluree create my-db --from path/to/ttl/dir
```

```sparql
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
SELECT (COUNT(?c) AS ?n) WHERE { ?c a rdfs:Class }
# expected: count of declared classes, actual: 0
```

A self-contained regression test that builds a small bulk-imported ledger and reproduces the asymmetry deterministically lives at `fluree-db-api/tests/it_query_rdfs_class_repro.rs::count_bound_object_mixed_leaflet_regression`.

## Workarounds (until fixed)

Any of these return the correct count:

1. Rewrite as a FILTER form: `?c rdf:type ?t . FILTER(?t = rdfs:Class)`
2. Use a plain `SELECT ?c` (no aggregate) — only the COUNT path is broken
3. Add `"reasoning": "none"` to a JSON-LD query

## Scope

- Affects bulk-imported / indexed datasets, not pure-novelty (memory-only) ledgers.
- Triggered for *any* class IRI whose first instance lands inside a POST leaflet rather than at its boundary; `rdfs:Class` happens to be a frequent victim because it sorts between `rdf:Property` and any user-namespaced class.
- Affects both JSON-LD and SPARQL forms of the query.
- Branch reproduced on: `main` (worktree at commit `efa05951`), Fluree CLI v4.0.1 release build.

## Notes

The data is correct on disk: `range_with_overlay(OPST, p=rdf:type, o=rdfs:Class)` returns the expected flakes, and `snapshot.encode_iri(...)` produces matching SIDs. The bug is in the COUNT fast-path's leaflet-skip heuristic — see linked PR for details.

aaj3f:
## Summary

Fixes #1208 — `COUNT { ?c rdf:type rdfs:Class }` returns 0 on indexed datasets, while the parallel `COUNT { ?p rdf:type rdf:Property }` and any user-class variant return the correct count. The data is on disk and reachable via FILTER / GROUP BY / non-COUNT shapes; only the COUNT-with-bound-object fast path drops the rows.

The bugs live in `count_bound_object_v6` and `group_count_v6` (`fluree-db-query/src/fast_group_count_firsts.rs`) — the fast paths that answer `SELECT (COUNT(?s) AS ?n) WHERE { ?s <p> <o> }` and `SELECT ?o (COUNT(?s) AS ?n) WHERE { ?s <p> ?o } GROUP BY ?o` from POST-leaflet directory metadata.

## Root cause

POST leaflets store one directory entry per leaflet whose `first_key` is the leaflet's first row. The fast path walks those entries and decides per-leaflet whether to skip, fast-count, or row-scan. POST sorts by `(p_id, o_type, o_key, …)`. Several compounding errors:

### 1. Skip-by-first-prefix

Old logic:

```rust
if prefix < target_prefix { continue; }
if prefix > target_prefix { break; }
// fall through: prefix == target → fast-count by row_count
```

A leaflet's rows span `[first_prefix, next_leaflet_first_prefix)`. A target value can sit *inside* a leaflet whose first row is some smaller value. In the reproducer the very first POST leaflet contains `rdf:Property` (the first row), then a few `rdfs:Class` rows, then a run of user-class rows; the loop saw `first_prefix < target` and `continue`d, then saw the next leaflet's `first_prefix > target` and broke — never decoding the leaflet that actually held `rdfs:Class`.

### 2. Mixed-predicate skip

Old logic:

```rust
if entry.p_const != Some(p_id) { continue; }
```

`p_const` is `None` for any leaflet whose rows don't all share a predicate — i.e. boundary leaflets that straddle a `p_id` change. Those can still contain rows for our predicate, but the check skipped them outright.

### 3. `prefix_v6_from_entry` strips `p_id`

POST's primary sort key is `p_id`. The fast path's `prefix_v6_from_entry` returns just `(o_type, o_key)`, which is a meaningful range bound *within* a predicate but unrelated to our range when the next leaflet starts a *different* predicate. With the predicate boundary fix above (point 2), this becomes a real footgun: the next leaflet's `(o_type, o_key)` for some other predicate could happen to be lexicographically less than the target's, falsely tripping the "leaflet ends before target → skip" condition. In the mixed-predicate regression test this fired exactly: a homogeneous `rdf:type` leaflet with `prefix == target` was skipped because the next leaflet started on `rdfs:label` with a smaller `(o_type, o_key)`.

### 4. Early break unsafe for mixed leaflets

`prefix > target_prefix → break` is sound only for homogeneous-predicate leaflets. In a mixed-predicate leaflet the first row may belong to a different predicate whose `(o_type, o_key)` exceeds the target while later rows in the leaflet (or in subsequent leaflets) still belong to our predicate.

### 5. Non-cached column projection misses `p_id`

`load_v6_batch`'s no-cache branch built a projection with only `OKey` (and optionally `OType`). The per-row `p_id` check inside mixed-predicate leaflets needs the `PId` column; without it `batch.p_id.get_or(row, 0)` returns `0` and every row gets silently rejected. The cached path always loads all columns so production was unaffected, but the no-cache path is reachable from test setups and is a footgun for any future caller.

### 6. Boundary-equality fast count needs homogeneous gate

The `next_prefix == Some(prefix) → += entry.row_count` shortcut is only valid when the leaflet is entirely target rows. In a mixed-predicate leaflet that condition can hold while the leaflet still contains rows for *other* predicates that must be excluded.

### Why the asymmetry (rdfs:Class fails, rdf:Property works)

In POST sort order, `rdf:type` rows are grouped by object IRI (`o_key`). In the reproducer, `rdf:Property` happens to be the **first** row of the leaflet (`first_prefix == target`), so the existing `prefix == target` branch catches it. `rdfs:Class` lives between `rdf:Property` and the first user class within the same leaflet — `first_prefix < target` — and gets dropped by error `#1` above. Any class that lands as the very first row of a POST leaflet works; any class that lands mid-leaflet silently returns 0.

### Why only on indexed/bulk-imported data

`count_bound_object_v6` and `group_count_v6` only fire when there's a binary index. Memory-only / pure-novelty queries never reach them.

## Fix

In `count_bound_object_v6` (and the same set of fixes mirrored into `group_count_v6`):

- **Mixed-predicate leaflets are no longer skipped on `p_const`.** Replace `entry.p_const != Some(p_id) { continue }` with: only skip when `p_const = Some(other_pid)`; verify `p_id` per-row inside mixed leaflets.
- **Skip a leaflet only when it ends strictly before the target.** Compute `next_prefix` and skip the leaflet only when `next_prefix < target_prefix`. Otherwise the target could be inside the leaflet — fall through to either the boundary-equality fast count or the row-level scan.
- **Predicate-qualified `next_prefix`.** Added `pid_prefix_v6_from_entry` returning `(p_id, o_type, o_key)`. When the next leaflet starts a different predicate, treat next_prefix as `None` (unknown upper bound) rather than comparing against an unrelated `(o_type, o_key)` from a different predicate's rows.
- **Gated early break.** `if prefix > target_prefix { break; }` is now `if prefix > target_prefix && entry.p_const == Some(p_id) { break; }`.
- **`load_v6_batch` non-cached projection includes `ColumnId::PId`** when `entry.p_const.is_none()`, so the per-row `p_id` check works regardless of cache state.
- **Boundary-equality fast count is gated on `prefix == target_prefix && next_prefix == Some(target_prefix) && entry.p_const == Some(p_id)`** — the entire leaflet is target rows for our predicate.

## Test plan

- [x] Add `count_bound_object_first_key_skip_regression` (`fluree-db-api/tests/it_query_rdfs_class_repro.rs`) — bulk-imports a small TTL where multiple distinct classes share a single homogeneous-`rdf:type` leaflet, then asserts `COUNT` for `rdf:Property`, `rdfs:Class`, and a 3000-instance user class. Exercises bug `#1`.
- [x] Add `count_mixed_predicate_leaflet_regression` — bulk-imports a tiny TTL designed to land in a single leaflet that straddles `rdf:type` and `rdfs:label`, then asserts `COUNT { ?c a rdfs:Class }` and the parallel `SELECT`.
- [x] All 12 existing tests in `fluree-db-api/tests/it_fast_group_count.rs` pass unchanged.
- [x] All 997 `fluree-db-query` library unit tests pass.
- [x] Verified end-to-end on the originally-reported dataset: COUNT for `rdfs:Class`, COUNT for `rdf:Property`, and the FILTER rewrite all agree, and the full-IRI form matches the prefixed form.

## Follow-ups (out of scope for this PR)

There are ~16 other call sites in the codebase using the same `entry.p_const != Some(p_id)` skip pattern (`fast_count.rs`, `fast_path_common.rs`, `fast_exists_join_count_distinct_object.rs`, `fast_sum_strlen_group_concat.rs`, `count_plan_exec.rs`, `join.rs`). They don't all share the prefix-skip variant of the bug, but the mixed-predicate-leaflet under-counting concern likely affects some of them. Worth a focused audit pass — tracking separately.

### DIFF
--- a/fluree-db-query/src/fast_group_count_firsts.rs
+++ b/fluree-db-query/src/fast_group_count_firsts.rs
@@ -495,10 +495,23 @@ impl Operator for PredicateObjectCountFirstsOperator {
 
 /// Extract the object prefix `(o_type, o_key)` from a V3 leaflet directory entry's
 /// `first_key` field, interpreted in POST order.
+///
+/// Strips `p_id`. Only meaningful as a bound for rows known to be within the
+/// same predicate range — see [`predicate_qualified_prefix`] for cross-predicate
+/// comparisons.
 #[inline]
 fn prefix_v6_from_entry(entry: &LeafletDirEntryV3) -> (u16, u64) {
     let rec = read_ordered_key_v2(RunSortOrder::Post, &entry.first_key);
     (rec.o_type, rec.o_key)
+}
+
+/// Read the `(p_id, o_type, o_key)` tuple from a V3 leaflet directory entry's
+/// `first_key`, interpreted in POST order. Includes `p_id` so callers can
+/// detect predicate boundaries.
+#[inline]
+fn pid_prefix_v6_from_entry(entry: &LeafletDirEntryV3) -> (u32, u16, u64) {
+    let rec = read_ordered_key_v2(RunSortOrder::Post, &entry.first_key);
+    (rec.p_id, rec.o_type, rec.o_key)
 }
 
 /// Load a V3 leaflet's columns, using the `LeafletCache` when available.
@@ -527,6 +540,12 @@ fn load_v6_batch(
         needed.insert(ColumnId::OKey);
         if entry.o_type_const.is_none() {
             needed.insert(ColumnId::OType);
+        }
+        // Mixed-predicate leaflets need the per-row `p_id` column to verify
+        // each row belongs to the queried predicate; the cached path always
+        // loads all columns, so this is only a concern for the no-cache path.
+        if entry.p_const.is_none() {
+            needed.insert(ColumnId::PId);
         }
         let projection = ColumnProjection {
             output: ColumnSet::EMPTY,
@@ -617,26 +636,49 @@ fn count_bound_object_v6(
             }
 
             let prefix = prefix_v6_from_entry(entry);
-            // The next leaflet's first prefix bounds this leaflet's range from
-            // above: rows live in `[prefix, next_prefix)`. Without that bound,
-            // a leaflet that *starts* below the target may still *contain* it.
-            let next_prefix = if i + 1 < dir.entries.len() {
-                Some(prefix_v6_from_entry(&dir.entries[i + 1]))
+            // The leaflet's rows span `[prefix, next_prefix)` half-open: the
+            // next leaflet's first key is the first row *not* in this leaflet.
+            // (A single key value can spill across leaflet boundaries — both
+            // leaflets then have rows for that value, and the gating below
+            // handles it correctly.)
+            //
+            // POST sorts by `(p_id, o_type, o_key, ...)`. The `(o_type, o_key)`
+            // pair is only a meaningful range bound *within* a single predicate;
+            // when the next leaflet starts a different predicate, its
+            // `(o_type, o_key)` tuple is unrelated to our range and cannot be
+            // used to skip this leaflet. Detect that case and treat next_prefix
+            // as `None` (unknown upper bound) — fall through to row-level scan.
+            let next_full = if i + 1 < dir.entries.len() {
+                Some(pid_prefix_v6_from_entry(&dir.entries[i + 1]))
             } else if leaf_idx + 1 < leaf_range.end {
-                let next_entry = &branch.leaves[leaf_idx + 1].first_key;
-                Some((next_entry.o_type, next_entry.o_key))
+                let next = &branch.leaves[leaf_idx + 1].first_key;
+                Some((next.p_id, next.o_type, next.o_key))
             } else {
                 None
             };
+            let next_prefix = next_full.and_then(|(np_p, np_ot, np_ok)| {
+                if np_p == p_id {
+                    Some((np_ot, np_ok))
+                } else {
+                    None
+                }
+            });
 
-            if prefix > target_prefix {
+            // `prefix` is computed from the leaflet's first row stripped of
+            // `p_id`. For a homogeneous-predicate leaflet (`p_const = Some(p_id)`)
+            // it's a sound sort key for early break, but in a mixed-predicate
+            // leaflet the first row may belong to a *different* predicate — its
+            // `(o_type, o_key)` could exceed `target_prefix` even when later
+            // rows in the leaflet (or in subsequent leaflets) belong to our
+            // predicate and are still on or before `target_prefix`. Restrict
+            // the break to homogeneous leaflets.
+            if prefix > target_prefix && entry.p_const == Some(p_id) {
                 break;
             }
-            // The leaflet's rows span `[prefix, next_prefix]` inclusive on both
-            // ends — `next_prefix` is the FIRST key of the *next* leaflet, but
-            // this leaflet's last row may equal target_prefix even when the
-            // next leaflet also starts at target_prefix. We can only skip when
-            // `next_prefix < target_prefix` (leaflet ends strictly before).
+            // Skip a leaflet only when it ends strictly before the target —
+            // i.e. the next leaflet's first row (within our predicate) also
+            // sorts before the target. Equality is not enough: this leaflet's
+            // last row can equal target if a key spills across the boundary.
             if let Some(np) = next_prefix {
                 if np < target_prefix {
                     continue;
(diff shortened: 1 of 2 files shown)


fluree-db-query/src/fast_group_count_firsts.rs
```diff
@@ -605,34 +624,75 @@ fn count_bound_object_v6(
...
+            if let Some(np) = next_prefix {
+                if np < target_prefix {
+                    continue;
+                }
+            }
+
+            // Boundary-equality fast count: only valid when the leaflet is
+            // entirely target rows (it starts at the target value AND the next
+            // leaflet also starts at the target value).
+            if prefix == target_prefix && next_prefix == Some(target_prefix) {
```
bplatz: Please add `&& entry.p_const == Some(p_id)` to this boundary-equality fast-count gate, matching `group_count_v6` and the PR description.

bplatz (approved):
Approving with one small consistency fix requested inline.

fluree-db-query/src/fast_group_count_firsts.rs
aaj3f: Addressed!

PR #1210, opened 2026-04-30
Revert Commits

zonotope:
## Summary

Adds a `revert` operation to the database, mirroring `git revert`. A revert produces a new commit on a target branch whose flakes invert one or more prior commits' flakes (assertion ⇄ retraction). Conflicts — `(s, p, g)` tuples touched by both the reverted commits and intervening non-reverted commits — are resolved per a chosen strategy. Three input shapes cover the natural use cases: a single commit, an explicit set (cherry-pick), or a git-style `A..B` range. A read-only `revert-preview` shows what a revert would do without writing a commit.

The user-supplied design choices:

1. Range syntax follows git: `A..B` is exclusive of `A`, inclusive of `B`, applied newest-first.
2. The result is bundled into a single revert commit (rather than one commit per reverted entry).
3. Conflict resolution reuses the existing merge `ConflictStrategy` enum (`Abort`, `TakeSource`, `TakeBranch`); `TakeBoth` and `Skip` are rejected as not meaningful for revert.
4. Merge commits inside the revert range are rejected for v1 (reverting merges with `-m`-style mainline selection is deferred).
5. The revert commit has a single parent (current HEAD); reverted commit IDs are recorded in `txn_meta` as `f:reverts` entries (one per reverted commit) for git-like provenance.

## Public Rust API

Three mutating entry points on `Fluree`, all returning `RevertReport` (branch, reverted commit IDs, conflict count, strategy applied, new HEAD t/id):

```rust
fluree.revert_commit (ledger, branch, commit:  CommitRef,                strategy)
fluree.revert_commits(ledger, branch, commits: Vec<CommitRef>,           strategy)
fluree.revert_range  (ledger, branch, from:    CommitRef, to: CommitRef, strategy)
```

Three read-only preview entry points returning `RevertPreview` (resolved commit summaries, conflict summary with optional cap, `revertable` verdict for the chosen strategy):

```rust
fluree.revert_commit_preview  (ledger, branch, commit:  CommitRef)
fluree.revert_commits_preview (ledger, branch, commits: Vec<CommitRef>)
fluree.revert_range_preview   (ledger, branch, from:    CommitRef, to: CommitRef)
```

Each preview method has a `*_with(opts: RevertPreviewOpts)` variant for tuning caps (`max_commits`, `max_conflict_keys`), opting out of conflict computation, or selecting a different strategy for the verdict.

`CommitRef` is the same user-facing reference type used by `branch create --at`: `Exact(CommitId)`, `Prefix(hex_digest)`, or `T(transaction_number)`. `CommitRef::parse` accepts the `t:N` / hex-prefix / full-CID string forms, so users can supply any of them at every external surface.

## External surfaces

**HTTP — mutating:** `POST /v1/fluree/revert` with body `{ ledger, branch, strategy?, commit | commits | range: { from, to } }`. Server validates exactly one of the three selection forms is supplied. Default strategy is `abort` (safer than merge's `take-both`, since revert is destructive). Forwarded in peer mode.

**HTTP — preview:** `GET /v1/fluree/revert-preview/*ledger?branch=&commit=|commits=|from=&to=&strategy=&max_commits=&max_conflict_keys=&include_conflicts=`. `commits` is comma-separated. Hard caps at 5,000 commits / 5,000 conflict keys, shared with merge-preview.

**CLI:** `fluree branch revert <commits>... [--from CID --to CID] [--branch B] [--strategy S] [--preview] [--json] [--ledger L] [--remote R]`. Positional commits and `--from`/`--to` are mutually exclusive. `--preview` switches to the read-only path; `--json` emits the raw `RevertPreview` JSON. Default strategy is `abort`.

**Remote client:** `RemoteLedgerClient::revert(...)` and `revert_preview(...)`, taking a `RevertPayload` enum (`Single` / `Set` / `Range`) so the CLI can pass through whichever form the user invoked.

## Conflict resolution

For each commit `C` being reverted, the inverse of `C`'s flakes is staged on top of the current HEAD. A conflict on key `(s, p, g)` arises when `C` modified that key **and** some intervening non-reverted commit between `C` and HEAD also modified it. Resolution mirrors merge:

- `Abort`: bail with `ApiError::BranchConflict` if any conflicts exist; nameservice state unchanged.
- `TakeSource`: revert wins — emit the inverse flakes plus retractions of HEAD's current values for each conflicting key.
- `TakeBranch`: HEAD wins — drop the inverse flakes for conflicting keys; HEAD's current values stay.

When `TakeBranch` drops every staged flake (e.g. all conflicts resolved in HEAD's favor), no commit is written; the report records the no-op with `new_head_t/id` equal to the existing HEAD.

## Implementation highlights

**`Flake::invert` / `Flake::invert_at(t)`** added to `fluree-db-core` (next to the existing `retract` / `retract_at`). Symmetric op-flip — distinct from `retract`, which only sets `op: false` regardless of input. Used by the revert flake-collection step.

**`f:reverts` predicate** added to the `fluree_vocab::db` namespace for revert provenance (one txn-meta entry per reverted commit).

**`NonEmpty<T> { head: T, tail: Vec<T> }`** introduced as a private newtype in `revert.rs` so the internal `RevertSource::Commits` and `RevertPlan.ordered_commits` cannot represent emptiness past the public-API boundary. Validation happens once at the entry point; downstream code relies on the structural invariant.

**Shared `apply_two_way_strategy` helper** moved to `rebase.rs` (next to `build_source_retractions`). Both merge and revert call it; the helper's docs note that rebase has the opposite polarity (`flakes` is the losing side under `TakeSource`) and so uses a separate resolver.

**Shared `commit_data` module** (`fluree-db-api/src/commit_data.rs`) exposes `CollectedCommitData` and `collect_from_commits<I, F>(commits, flake_transform)`. Merge passes `std::convert::identity` for the transform; revert passes `|f| f.invert_at(0)`. The namespace and graph delta accumulation (with earlier-wins `or_insert` semantics) is shared.

**`build_revert_context` helper** factored out of `revert_inner` into a `pub(crate)` method on `Fluree`. It performs every step the mutating path needs *before* writing — branch lookup, branched-store construction, `CommitRef` resolution, plan computation, conflict-key computation — so the preview path can call it without any duplicated orchestration.

**Shared preview defaults / hard caps**: `RevertPreviewOpts` reuses `merge_preview::DEFAULT_MAX_COMMITS` and `DEFAULT_MAX_CONFLICT_KEYS`. The server's per-request hard caps for both preview endpoints share renamed neutral constants `PREVIEW_HARD_MAX_COMMITS` / `PREVIEW_HARD_MAX_CONFLICT_KEYS`.

**`RevertPayload`** lives in `remote_client.rs` (the wire shape, alongside the methods that serialize it) so the dependency direction stays one-way: `commands → remote_client`, never the reverse.


bplatz (approved):
Awesome!

PR #158, opened 2022-03-03
Allow clj api to work with closed api ledger

dpetran:
A closed api ledger will not send avro/binary on the wire. This commit changes the
default clojure connection to speak json - the default-read-fn now requests json instead
of avro, and the serializer now expects json.

FC-1372
FC-1370

PR #1211, opened 2026-04-30
Fix: Reindex Rebuild Performance Regression

aaj3f:
Fixes the regression that pushed first-time reindex on the large dataset indexing Lambda past the 13-minute waiter timeout repeatedly. Two stacked slowdowns were hitting any reindex of a ledger with no prior index; this PR addresses both, plus a defensive fix for branched ledgers and regression guards so neither can re-emerge silently.

## Background

CloudWatch logs for a particular indexing lambda showed 787-commit reindexes of a dataset consistently timing out at 13 minutes:

- 24-30s of preamble (then silence in the default INFO log filter)
- 1574 = 2 × 787 `read_commit_v4` debug events when DEBUG was bumped
- 11+ minutes of total silence after that — no Phase A/B/C/D progress
- Repeated `Index trigger timed out` waits, identical retry pattern across cold starts

Two distinct regressions, both shipped via PR #1199 / commit `5985d0f01 fluree v4 baseline`:

1. **`Novelty::apply_commit` called per-commit during `LedgerState::load_novelty`.** The `merge_batch_into_index` two-way merge accrued `O(M·N̄)` over the chain. For 787 commits × ~9k ops this measured ~11 minutes on Lambda single-vCPU. Silent in logs because `apply_commit`'s span lives in `fluree_db_novelty`, not in any rebuild target most filters include.

2. **`ApiFulltextConfigProvider::resolve` always loaded full `LedgerState`.** Even when the ledger had never written `f:LedgerConfig`, `resolve` walked every commit via `load_commit_by_id` (`read_commit_v4`) just to discover the config graph was empty. ~30s of wasted full-blob reads + zstd decompresses + flake construction before the rebuild even started.

The provider regression originated in fluree/db-r commit `4d6ff8fc "fulltext fixes/enahnce"` on the unmerged `misc/fixes` branch (a fix for the "Solo bug report 2026-04-23" referenced in `fluree-db-api/tests/it_query_fulltext.rs:863-868`). The novelty regression has been there longer but only surfaced once the provider started forcing it on every reindex.

`BranchedContentStore` separately doesn't implement `get_range`, so PR fluree/db-r#158's envelope-only byte-range probe falls back to full-blob fetch via the trait default for branched ledgers. The dataset-in-question's main branch isn't branched so this isn't the blocking cause for them, but it's a regression vector waiting for the next branched ledger reindex.

### `Novelty::bulk_apply_commits` (commit 1, `51217f33b`)

`fluree-db-novelty/src/lib.rs` — new bulk path designed for first-load / catch-up. One ingest pass routes flakes per-graph, parallel sort by identity-then-t, linear set-semantics dedup walk (matches `apply_commit`'s `fact_currently_asserted_in_graph` skip rule for assertions; always keeps retractions), then four parallel sorts to build SPOT/PSOT/POST/OPST. `O(N log N)` total regardless of commit count.

`fluree-db-ledger/src/lib.rs::load_novelty` — switches the deferred-batch replay loop from per-commit `apply_commit` to a single `bulk_apply_commits` call. dict-novelty population still runs per-batch in commit order so overlay translation invariants are preserved.

### Range-aware `BranchedContentStore` (commit 1, `51217f33b`)

`fluree-db-core/src/storage.rs` — implements `ContentStore::get_range` on `BranchedContentStore`. Mirrors `get`'s ancestry-fallback semantics, calling `get_range` natively on each ancestor so range-capable backends keep their byte-range optimization across the entire chain.

### Envelope-only short-circuit (commit 2, `d5e03c27d`)

`fluree-db-core/src/commit.rs` — surfaces `graph_delta` on the public `CommitEnvelope` (the binary envelope already encoded it; the v4 reader was dropping it on the floor). Adds new public helper `first_t_where_graph_registered<C: ContentStore>(store, head, iri)` that walks the DAG via `load_commit_envelope_by_id` (byte-range ~128 KiB per envelope) and returns the lowest `t` where `iri` appears in any commit's `graph_delta`, or `None` if it never does. v3 chains return `None` (v3 envelopes never carried `graph_delta`) — callers fall through to a full load on `None`.

`fluree-db-api/src/indexer_fulltext_provider.rs::resolve` — runs the envelope-only probe before `LedgerState::load`. If `config_graph_iri(ledger_id)` is never registered in any commit's `graph_delta`, returns an empty configured-properties list immediately. No `LedgerState::load`, no novelty build, no full commit reads. If the IRI is found, falls through to the existing load + `resolve_ledger_config` path — which is itself fast now thanks to `bulk_apply_commits`.

### Regression tests (commit 3)

`fluree-db-api/tests/it_reindex_pre_build_perf.rs` (new):

- **`first_t_where_graph_registered_no_full_reads_when_iri_absent`** — deterministic `CountingContentStore` test asserting the envelope walk uses only `get_range`, never `get`, when the target IRI is missing.
- **`first_t_where_graph_registered_returns_lowest_t_when_iri_present`** — confirms the helper finds the earliest registration `t`.
- **`orchestrator_first_reindex_no_config_completes_quickly`** — end-to-end orchestrator-path reindex of a 200-commit no-config ledger via `BackgroundIndexerWorker` matching the AWS Lambda wiring (`with_fulltext_config_provider`), with a 60s wall-clock budget far below the 13-minute Lambda waiter that motivated the fix.

## Effect on the original reindex scenario

Pre-fix (CloudWatch evidence, requestId `6e44e74e-1c8e-411b-a8b7-b73d8b268db8`, cold start `2026-04-30T16:21:35`):

- `LedgerState::load_novelty` walked all 787 commits via `read_commit_v4` in 29s
- `apply_commit` × 787 in `O(N²)` replay then ran for 11+ minutes silently
- Lambda waiter timed out at 13 minutes
- Lambda function timed out at 15 minutes before any rebuild progress

Post-fix (projected from in-process timings + Lambda ratios):

- Envelope-only probe — ~787 byte-range reads, ~5-8s end-to-end on Lambda
- `first_t_where_graph_registered` returns `None`. It has no `f:LedgerConfig` — verified via DynamoDB query of nameservice, where `pk=dataset:main, sk=config` shows `config_v: 0` and no `config_id` field.
- `ApiFulltextConfigProvider::resolve` returns `Vec::new()` immediately. No `LedgerState::load`, no novelty build.
- Rebuild proceeds normally — Phase A 31s, Phase B ~10s with K=3, Phase C tiny, Phase D 12s. Total reindex 60-90s, well under 900s.

For ledgers with config in commits (test path `fulltext_configured_property_first_build_via_provider`, `fulltext_configured_property_picked_up_by_build_index_for_ledger`, etc.), the envelope walk falls through to the full load path — which is now fast thanks to `bulk_apply_commits`. Behavior unchanged from a correctness standpoint.

## Diagnosis methodology

Iteratively bumped Lambda `RUST_LOG` across cold starts to localize the silent stretch:

1. `INFO` baseline — only orchestrator + waiter messages visible, the 13-minute stretch invisible.
2. `info,fluree_db_indexer=debug,fluree_db_core::commit=debug,...` — surfaced the 1574 = 2 × 787 `read_commit_v4` events, bounded the catch-up at 29s. Span tag `name:"load_commit"` on every event traced the calls to `load_commit_by_id`.
3. Code audit (`fluree-db-indexer/src/lib.rs:122-126` → `fluree-db-api/src/indexer_fulltext_provider.rs:43-82` → `fluree-db-ledger/src/lib.rs::load_novelty:284`) located both regressions.
4. DynamoDB inspection of the symptomatic NsRecord confirmed `config_id: None`, validating the envelope-short-circuit fast path.

Audit of the proposed fix by an independent agent caught one early incorrect direction (skipping the provider on `index_head_id.is_none()` would have silently broken config-via-commits on first build, failing `fulltext_configured_property_first_build_via_provider`). The final shape preserves correctness for that path while still avoiding the slow walk for no-config ledgers.

## Test plan

- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy --workspace --all-features --all-targets -- -D warnings` passes
- [x] `cargo nextest run --workspace --all-features --no-fail-fast` — 6014/6014 tests pass (11 skipped — feature-gated S3 testcontainers paths run separately)
- [x] Existing `fulltext_configured_property_first_build_via_provider` and friends still pass — confirms config-via-commits fulltext path on first reindex is preserved
- [x] Existing `it_rebuild_fetch_counts.rs` still passes — confirms Phase A/B fetch contract from PR #158 is preserved
- [x] Existing `it_indexing_workflow.rs` passes — confirms orchestrator + manual-trigger paths unchanged
- [x] New `it_reindex_pre_build_perf.rs`: `first_t_where_graph_registered_no_full_reads_when_iri_absent`, `first_t_where_graph_registered_returns_lowest_t_when_iri_present`, `orchestrator_first_reindex_no_config_completes_quickly`
- [ ] **Lambda smoke test on original dataset** — after this PR merges and the indexing Lambda redeploys, confirm the reindex completes well under 900s

## Notes

Read-path only on the `BranchedContentStore` change. No on-disk format changes. `CommitEnvelope::graph_delta` is a new public field, but the binary encoding already carried it — only the v4 reader surfaced it on the public type.

v3 chains never encoded `graph_delta` on the envelope and will always return `None` from `first_t_where_graph_registered`. Callers that need v3 compatibility should fall back to a full load on `None`, which `ApiFulltextConfigProvider` already does.

Out of scope, follow-up candidates:

- Parallelize the envelope walk in `first_t_where_graph_registered` via `buffered(K)` for very long chains.
- Extend the orchestrator to expose its content store for test instrumentation so the wall-clock budget test can also assert fetch counts deterministically.
- Audit other `LedgerState::load` callers (`merge.rs:543`, `rebase.rs:348`, `ledger_info.rs:539`) for similar slow-path opportunities now that the bulk-load infrastructure exists.

### DIFF
--- a/.fluree-memory/repo.ttl
+++ b/.fluree-memory/repo.ttl
@@ -1269,6 +1269,39 @@ mem:fact-01kj8k5nzpe59p5vp7zrehbrr9 a mem:Fact ;
     mem:branch "feature/memory" ;
     mem:createdAt "2026-02-24T19:49:14.358637+00:00"^^xsd:dateTime .
 
+mem:fact-01kqfy6r0w86fcsd9km0707xzg a mem:Fact ;
+    mem:content "v3 commit envelope reader at `legacy_v3.rs:305-309` deliberately stubs `graph_delta = HashMap::new()` since v3 never encoded the field. Envelope-only walkers (`first_t_where_graph_registered`) therefore return `None` on any chain that includes v3 commits, regardless of whether the named graph was actually registered. Callers that rely on the result for short-circuiting must either fall through to a full load on `None`, detect v3 envelopes and signal \"inconclusive\", or accept silent correctness regressions on legacy chains." ;
+    mem:tag "commit-envelope" ;
+    mem:tag "gotcha" ;
+    mem:tag "graph-delta" ;
+    mem:tag "legacy-v3" ;
+    mem:tag "short-circuit" ;
+    mem:tag "v3-compat" ;
+    mem:scope mem:repo ;
+    mem:artifactRef "fluree-db-api/src/indexer_fulltext_provider.rs" ;
+    mem:artifactRef "fluree-db-core/src/commit.rs" ;
+    mem:artifactRef "fluree-db-core/src/commit/codec/legacy_v3.rs" ;
+    mem:branch "fix/reindex-pre-build-perf-regression" ;
+    mem:createdAt "2026-04-30T19:35:13.692947+00:00"^^xsd:dateTime ;
+    mem:rationale "Surfaced during PR #1211 review — code returned Vec::new() on None even though the PR description said it falls through to full load. Documenting so future envelope-only short-circuit work doesn't repeat the silent-fail shape." .
+
+mem:fact-01kqfy6txdrjppaf6756xzdz25 a mem:Fact ;
+    mem:content "`Novelty::bulk_apply_commits` is the O(N log N) bulk replacement for per-commit `apply_commit` on first-load / catch-up paths. Identity sort key is `(s, p, o, dt, m, t, op)`; dedup walk drops an assertion iff the prior kept flake for the same `(s, p, o, dt, m)` was also an assertion (mirrors `fact_currently_asserted_in_graph`). Used by `LedgerState::load_novelty` after reversing HEAD→oldest batches and populating dict_novelty in commit order. Delta vs apply_commit: deduped flakes remain in the FlakeStore arena and self.size over-reports — bounded for fresh-load callers but not appropriate for hot-path mutation." ;
+    mem:tag "bulk-apply" ;
+    mem:tag "dedup" ;
+    mem:tag "indexing" ;
+    mem:tag "lambda" ;
+    mem:tag "load-novelty" ;
+    mem:tag "novelty" ;
+    mem:tag "performance" ;
+    mem:scope mem:repo ;
+    mem:artifactRef "fluree-db-core/src/comparator.rs" ;
+    mem:artifactRef "fluree-db-ledger/src/lib.rs" ;
+    mem:artifactRef "fluree-db-novelty/src/lib.rs" ;
+    mem:branch "fix/reindex-pre-build-perf-regression" ;
+    mem:createdAt "2026-04-30T19:35:16.653062+00:00"^^xsd:dateTime ;
+    mem:rationale "Captures the design contract and arena-bloat semantics from PR #1211 so future callers of bulk_apply_commits don't accidentally wire it into a hot-path mutation site." .
+
 mem:constraint-01kjte29y3cp8mae56xy6yfv2n a mem:Constraint ;
     mem:content "SPARQL↔JSON-LD Parity: both compile to the same IR (fluree-db-query/src/ir.rs) and share the entire execution engine. When adding SPARQL features: (1) shared IR changes already affect JSON-LD — verify JSON-LD tests pass, (2) if expressible in JSON-LD, add parallel tests, (3) execution path missing in JSON-LD lower → track follow-up.\n\nSPARQL tests: it_query_sparql.rs | JSON-LD tests: it_query.rs, it_query_analytical.rs, it_query_grouping.rs\nValidate: `cd testsuite-sparql && make test-eval-cat CAT=<cat>` then `cargo test -p fluree-db-api --test it_query`" ;
     mem:tag "compliance" ;
--- a/fluree-db-api/src/indexer_fulltext_provider.rs
+++ b/fluree-db-api/src/indexer_fulltext_provider.rs
@@ -17,6 +17,7 @@
 use std::sync::Arc;
 
 use async_trait::async_trait;
+use fluree_db_core::{config_graph_iri, first_t_where_graph_registered};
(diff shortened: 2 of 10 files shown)


bplatz (approved):
When the config graph IS registered, ApiFulltextConfigProvider::resolve walks the DAG once via first_t_where_graph_registered (envelope-only), then LedgerState::load walks it again (full blobs). For the no-config case (the regressing one) this is fine — probe terminates and we return. For ledgers that DO use fulltext config it's additional latency.
                                                                                                                                                                                                        
The PR description defers this as item 3, but it's worth at least returning the discovered t from Found(t) so a future commit can use it as a stop_at_t hint or to seed a partial walk. Right now t is logged and discarded.

aaj3f:
> When the config graph IS registered, ApiFulltextConfigProvider::resolve walks the DAG once via first_t_where_graph_registered (envelope-only), then LedgerState::load walks it again (full blobs). For the no-config case (the regressing one) this is fine — probe terminates and we return. For ledgers that DO use fulltext config it's additional latency.
> 
> The PR description defers this as item 3, but it's worth at least returning the discovered t from Found(t) so a future commit can use it as a stop_at_t hint or to seed a partial walk. Right now t is logged and discarded.

Added a comment in the code and made a GH issue at #1218

PR #1212, opened 2026-04-30
refactor: Remove `history_mode` from `ExecutionContext` 

bplatz:
This PR follows up on the history-query planning feedback by removing `history_mode` from `ExecutionContext` and making temporal mode a planner-time concern. History-vs-current semantics are now decided before operator construction via `PrepareConfig`/`PlanningContext`, so the planner builds the appropriate operator tree once instead of asking runtime operators to branch on a context flag.

For history queries, scan construction now selects the dedicated history scan path up front, while current-state queries continue to use the normal binary scan operator. Fast paths that only have current-state semantics are gated during planning, so they are not constructed for history-range queries. Late-built subplans such as `OPTIONAL`, `UNION`, `FILTER EXISTS`, `MINUS`, and related operators capture the same planning context to avoid silently falling back to current-state behavior.

The refactor keeps `RangeOptions::history_mode`/`QueryBounds::history_mode` in the core range API, where they remain the lower-level instruction for `range_with_overlay`. The user-facing query execution path no longer carries a runtime history flag through `ExecutionContext`; prepared execution now captures temporal semantics in the planned operator tree.

Test coverage was expanded with history combination cases for `OPTIONAL`, `UNION`, `FILTER NOT EXISTS`, and count fast-path avoidance. These tests specifically guard against the old failure mode where nested or late-built subplans could accidentally plan as current-state and miss retract events.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,9 @@ __pycache__/
 # Test database (generated data)
 /test-database/
 
+# Benchmark inputs/outputs — never tracked
+/benchmarks/
+
 benchmarks/bsbm/bsbm100m_results_quick.xml
 benchmarks/bsbm/bsbm100m_results_v2.xml
 benchmarks/bsbm/bsbm100m_results.xml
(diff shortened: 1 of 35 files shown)


zonotope (approved):
🎒

PR #1213, opened 2026-05-01
Arch Linux PKGBUILD

zonotope:
## Summary

Adds an Arch Linux binary package (`fluree-bin`) for the AUR so Arch users can install Fluree with a standard pacman AUR helper (e.g. `paru -S fluree-bin`, `yay -S fluree-bin`). The package consumes the same prebuilt Linux tarballs that `dist` already publishes to GitHub Releases — no extra build artifacts required upstream. A new `publish-aur` job in the existing release workflow keeps the AUR copy in lockstep with each stable release alongside the existing Docker and Homebrew publish jobs.

## What's added

**`contrib/aur/fluree-bin/PKGBUILD`** — the AUR recipe. Pulls `fluree-db-cli-{x86_64,aarch64}-unknown-linux-gnu.tar.xz` from the matching GitHub release, installs `/usr/bin/fluree`, the BUSL-1.1 LICENSE, the README, and bash/zsh/fish shell completions (generated via the binary's own `fluree completions <shell>` command). Declares `glibc` and `gcc-libs` as runtime deps (verified via `ldd` — no other shared libs), and uses `provides=fluree` / `conflicts=fluree` so the package can coexist with a future source-build variant. Skips the strip step (`options=('!strip' '!debug')`) because `dist` already ships a stripped binary.

**`contrib/aur/fluree-bin/.SRCINFO`** — generated from the PKGBUILD via `makepkg --printsrcinfo`. Required by AUR alongside the PKGBUILD.

**`contrib/aur/publish.sh`** — the release-time updater. Given a tag (e.g. `v4.0.2`), it fetches both per-arch `.sha256` siblings from the GitHub release, sed-substitutes `pkgver`, `pkgrel`, and the two `sha256sums_*` lines into the in-tree PKGBUILD template, regenerates `.SRCINFO` by running `makepkg --printsrcinfo` inside a one-shot `archlinux:base-devel` container (so the runner doesn't need pacman/makepkg installed), then commits and pushes to `ssh://aur@aur.archlinux.org/fluree-bin.git`. No-ops cleanly when the AUR repo already matches the requested version.

**`.github/workflows/release.yml`** — new `publish-aur` job modeled on the existing `publish-docker` job. Runs after `host` (so the GitHub release and its `.sha256` files exist), gated on `!announcement_is_prerelease` so prereleases skip, sets up SSH with pinned AUR host keys (rotation will fail loudly rather than silent-TOFU), then invokes `bash contrib/aur/publish.sh "$TAG"`. Wired into `announce`'s `needs` and `if` clauses with the same `skipped || success` allowance the other publish jobs use.

## One-time setup required before merge takes effect (Done)

1. Generate a dedicated SSH key (ed25519, no passphrase) for the AUR release bot.
2. Register the public half on a Fluree-owned AUR account.
3. Add the private half to GitHub repo secrets as `AUR_SSH_PRIVATE_KEY` (`gh secret set` or web UI).
4. Bootstrap the AUR repo by pushing the initial PKGBUILD + .SRCINFO from a workstation — AUR creates the repo on first push, the workflow only updates it on subsequent releases.

## Notes

- The in-tree `contrib/aur/fluree-bin/PKGBUILD` will go stale relative to AUR after each release — the workflow only updates the AUR copy, not the in-tree template. This is harmless since users install via AUR rather than checking out a tag, but is worth knowing. If we want the in-tree file to always reflect the latest published version, a follow-up can add a step to `publish-aur` that opens a PR back to `main` with the bumped values.
- `.github/workflows/release.yml` is autogenerated by `dist`, but `allow-dirty = ["ci"]` is already set in `dist-workspace.toml` so `dist generate` will not clobber the new job. Anyone running `dist init` from scratch would need to re-add it.
- Validation done: PKGBUILD checksums match upstream's published `.sha256` files for v4.0.2; full `makepkg` build produces a `.pkg.tar.zst` with the expected layout (`/usr/bin/fluree`, completions in their standard system paths, LICENSE under `/usr/share/licenses/fluree-bin/`); workflow YAML parses cleanly; `publish.sh` passes `bash -n` and its sed substitutions produce the expected pinned values when dry-run against the live v4.0.2 release artifacts.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -378,16 +378,60 @@ jobs:
           done
           git push
 
+  # Publish the fluree-bin package to the Arch User Repository (stable releases only).
+  publish-aur:
+    needs:
+      - plan
+      - host
+    runs-on: "ubuntu-22.04"
+    if: ${{ !fromJson(needs.plan.outputs.val).announcement_is_prerelease }}
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - name: Configure SSH for AUR
+        env:
+          AUR_SSH_PRIVATE_KEY: ${{ secrets.AUR_SSH_PRIVATE_KEY }}
+        run: |
+          mkdir -p ~/.ssh
+          chmod 700 ~/.ssh
+          printf '%s\n' "$AUR_SSH_PRIVATE_KEY" > ~/.ssh/id_aur
+          chmod 600 ~/.ssh/id_aur
+          # Pinned AUR host keys. Refresh with `ssh-keyscan aur.archlinux.org`
+          # if upstream rotates them — a mismatch will fail this job loudly.
+          cat > ~/.ssh/known_hosts <<'EOF'
+          aur.archlinux.org ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEuBKrPzbawxA/k2g6NcyV5jmqwJ2s+zpgZGZ7tpLIcN
+          aur.archlinux.org ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBLMiLrP8pVi5BFX2i3vepSUnpedeiewE5XptnUnau+ZoeUOPkpoCgZZuYfpaIQfhhJJI5qgnjJmr4hyJbe/zxow=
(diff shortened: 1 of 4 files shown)


bplatz (approved):
💻 

Looks like we need a pre-merge step.

zonotope:
> 💻
> 
> Looks like we need a pre-merge step.

I already completed the AUR registration and the [`fluree-bin` package](https://aur.archlinux.org/packages/fluree-bin) has been published. I have also verified that it works on my own system by installing it with `paru -S fluree-bin`. 

Were you talking about another pre-merge step?

bplatz:
> > 💻
> > Looks like we need a pre-merge step.
> 
> I already completed the AUR registration and the [`fluree-bin` package](https://aur.archlinux.org/packages/fluree-bin) has been published. I have also verified that it works on my own system by installing it with `paru -S fluree-bin`.
> 
> Were you talking about another pre-merge step?

No, I meant the registration so perfect!

PR #1214, opened 2026-05-02
feat: Remote-source bulk import (`import_from_storage`)

bplatz:
Adds `fluree.create("ledger").import_from_storage(storage, source).execute()` so
consumers whose source data lives in `StorageRead`-backed storage (S3 today;
any backend tomorrow) can bulk-import directly into a new ledger without
first staging the input on local disk.

Motivation: a Lambda consumer needed to import large datasets from
S3. The local `.import(path)` requires every source byte on local disk before
import begins — wasted I/O at best, infeasible at worst in ephemeral
environments. This streams source bytes directly into the existing parse/commit
pipeline.

## What's in scope

- **`StorageRead::list_prefix_with_metadata`** — new trait method returning
  `(address, size_bytes)` pairs. Default impl errors loudly; `S3Storage`
  overrides using `list_objects_v2` (size is native); `MemoryStorage`
  overrides for tests.
- **`RemoteObject` + `RemoteSource`** — a typed `(address, size_bytes)` struct
  and a small enum (`OrderedObjects(Vec<RemoteObject>)` /
  `Prefix { prefix }`). `OrderedObjects` is the production-recommended path;
  `Prefix` is the convenience listing path.
- **`CreateBuilder::import_from_storage(storage, source)`** — new public API
  parallel to `.import(path)`, accepting `Arc<dyn StorageRead>`.
- **Remote chunk producer** — async tokio task fetches each object whole via
  `read_bytes`, sends `(idx, bytes)` into a bounded tokio mpsc; a small
  bridge thread forwards to a `std::sync::mpsc::SyncSender` that parser
  workers drain. Parser workers stay CPU-only — no `block_on` from a parser
  thread.
- **Producer error vs EOF** — explicit `oneshot::Receiver<Option<ImportError>>`.
  Channel close alone is not treated as success; the import driver awaits the
  signal so a mid-stream producer crash never silently truncates the import.
- **Two execution arms** chosen by source content:
  - **Parallel** for pure `.ttl` — same parser pool the local `Files` path
    uses. Hot path.
  - **Serial** for `.jsonld` — drains the channel sequentially and dispatches
    via `parse_jsonld_chunk` + `finalize_parsed_chunk`, mirroring the local
    serial fallback.
- **`.trig` is rejected** with an actionable error pointing to the upstream
  limitation in `fluree-db-transact/src/import.rs:452` (named-graph flakes
  aren't captured by the Tier 2 spool pipeline). The serial remote arm
  already dispatches `.trig` correctly, so enabling it here is a 5-line
  resolver change once the upstream issue is fixed.

## Local path is not modified

Bit-for-bit unchanged. No generics introduced into `StreamingTurtleReader`,
no shared traits, no `Box<dyn>` in any hot loop. The local `Files` and
`Streaming` arms compile to identical machine code.

The 12 existing local-import tests continue to gate the local-path
performance guarantee. None of them changed.

## Lambda runtime guidance

Streaming the input from S3 means input bytes never touch local `/tmp`.
Scratch (spool/runs/index) still uses local disk under `FLUREE_IMPORT_DIR`,
so Lambda viability is bounded by scratch ratio (~4× input, data-dependent),
not input size. Working envelope at 4× scratch is roughly a 2 GB input
ceiling for Lambda's 10 GB `/tmp` — but consumers should validate empirically
on a representative subset and enforce an upload-size cap.

For unbounded import sizes, Lambda is the wrong runtime regardless — use
Fargate or EC2.

## Files changed

| File | What |
|---|---|
| `fluree-db-core/src/storage.rs` | `RemoteObject` + `list_prefix_with_metadata` |
| `fluree-db-core/src/storage/memory.rs` | `MemoryStorage` override (used in tests) |
| `fluree-db-storage-aws/src/s3/mod.rs` | `S3Storage` override (`list_objects_v2`) |
| `fluree-db-api/src/import.rs` | `RemoteSource`, `RemoteChunkProducer`, async producer + bridge thread, parallel/serial remote arms, `import_from_storage` API |
| `fluree-db-api/src/lib.rs` | Re-exports `RemoteSource`, `RemoteObject` |
| `fluree-db-api/tests/it_import_remote.rs` | 7 integration tests (new file) |
| `docs/design/storage-traits.md` | Trait surface updated for new method |

## Future work (not this PR)

- **Single large remote `.ttl` streamed via S3 byte-range reads** with
  statement-boundary splitting. Lets a multi-GB remote `.ttl` be split
  without ever holding the whole object in memory. Materially more complex;
  deferred until a consumer needs it.
- **Remote `.trig` enablement** — pending fix to the Tier 2 spool pipeline
  (see `fluree-db-transact/src/import.rs:452`).
- **Object-storage-backed scratch** — would unlock Lambda for arbitrary
  import sizes by removing the `/tmp` scratch dependency. Larger separate
  workstream.

### DIFF
--- a/docs/design/storage-traits.md
+++ b/docs/design/storage-traits.md
@@ -111,14 +111,43 @@ pub trait StorageRead: Debug + Send + Sync {
     /// Check if an address exists
     async fn exists(&self, address: &str) -> Result<bool>;
 
+    /// Read a `[start, end)` byte range from an address.
+    /// Default implementation fetches the full object and slices; backends
+    /// that support native range reads (S3, HTTP) should override.
+    async fn read_byte_range(&self, address: &str, range: std::ops::Range<u64>)
+        -> Result<Vec<u8>>;
+
     /// List all addresses with a given prefix
     async fn list_prefix(&self, prefix: &str) -> Result<Vec<String>>;
+
+    /// List addresses under a prefix together with byte sizes.
+    /// Default implementation returns an error indicating the backend does
+    /// not support cheap metadata listing. Backends with native list+size
+    /// (S3 `list_objects_v2`, GCS, etc.) should override.
+    async fn list_prefix_with_metadata(&self, prefix: &str)
+        -> Result<Vec<RemoteObject>>;
+
+    /// Resolve a CAS address to a local filesystem path, if available.
+    fn resolve_local_path(&self, address: &str) -> Option<PathBuf> { None }
+}
+
+/// `(address, size)` pair returned by `list_prefix_with_metadata`.
+pub struct RemoteObject {
+    pub address: String,
+    pub size_bytes: u64,
 }
 ```
 
 **Design notes:**
 - `read_bytes_hint` enables optimizations like returning pre-encoded flakes for leaf nodes
+- `read_byte_range` allows partial reads against backends with native HTTP/S3 range support;
+  the default impl is correct but does N full-object fetches for N range reads
 - `list_prefix` is essential for garbage collection and administrative operations
+- `list_prefix_with_metadata` is used by the bulk-import remote-source path so the
+  importer can size each chunk before fetching. Backends without cheap size metadata
+  return an error; callers can fall back to caller-supplied object lists
+- `resolve_local_path` lets callers (e.g., import scratch staging) skip a copy when
+  the storage already exposes data on the local filesystem (`FileStorage`)
 - All methods return `fluree_db_core::Result<T>` (alias for `std::result::Result<T, Error>`)
 
 ### StorageWrite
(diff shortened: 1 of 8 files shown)


zonotope (approved):
Looks good. The only small issue is that the hard coded line number in the error message starting at fluree-db-api/src/import.rs:818 is going to be difficult to keep up to date as the file changes. I think it would be better to just give a blanket "not supported" message there.

bplatz:
> Looks good. The only small issue is that the hard coded line number in the error message starting at fluree-db-api/src/import.rs:818 is going to be difficult to keep up to date as the file changes. I think it would be better to just give a blanket "not supported" message there.

Addressed in e8349c8d7b3fbf86d45bf8ced0440d06654c20bd

PR #1215, opened 2026-05-04
Consolidate query execution and rewrite pipelines

zonotope:
## Summary

Consolidates the query-execution and pattern-rewrite code paths in `fluree-db-query` behind a single set of entry points, removes the wrapper functions that accumulated to thread varying combinations of optional context state, and replaces hand-rolled match-and-recurse blocks across four rewrite passes with a shared `Pattern::map_subpatterns` helper. These changes caused a net reduction of about 1000 lines of code.

## Motivation

The execution surface in `fluree-db-query` had grown a family of nearly-identical wrapper functions — `execute`, `execute_with_dataset`, `execute_with_dataset_and_bm25`, `execute_with_policy`, `execute_pattern_with_overlay_at`, and so on — each one differing only in which optional pieces of context state (tracker, policy enforcer, dataset, R2RML providers, BM25/vector providers, binary store, dict novelty, spatial/fulltext providers, remote service executor) it threaded through. New optional state required either a new wrapper or yet another positional parameter on an existing one. Callers had to pick the wrapper that matched their exact combination, and the lib.rs surface re-exported a tangle of overlapping entry points. Likewise, the pattern-rewrite passes (`rewrite.rs`, `geo_rewrite.rs`, `r2rml/rewrite.rs`, `rewrite_owl_ql.rs`) each contained the same 8-arm match block to recurse into `Pattern::Optional`, `Union`, `Minus`, `Exists`, `NotExists`, `Graph`, `Service`, and `Subquery` — modulo the per-pass leaf logic.

## Key changes

- **Single execution entry point.** `execute` and `execute_streaming` in `fluree-db-query/src/execute/runner.rs` are now the canonical paths for prepared and lazy execution. Both take a `ContextConfig<'a, 'b>` carrying every optional piece of context state.
- **`ContextConfig` replaces parameter sprawl.** Callers build a config with `ContextConfig { tracker: Some(&t), r2rml: Some((&p, &p)), ..Default::default() }` instead of selecting one of N positional-argument wrappers. `Default::default()` sets `strict_bind_errors: true` (the de-facto default at every prior call site).
- **Wrappers removed.** `execute_with_dataset`, `execute_with_dataset_and_bm25`, `execute_with_policy`, and the `execute_pattern_with_overlay_at` family are gone from `fluree-db-query/src/lib.rs` and `execute.rs`. Callers in `fluree-db-api` (BM25, vector, ontology imports, graph-source helpers, R2RML integration tests, transact staging) now construct a `ContextConfig` directly.
- **`Pattern::map_subpatterns`.** New method on `Pattern` (in `fluree-db-query/src/ir.rs`) that applies a closure to every immediate `Vec<Pattern>` container — Optional, Minus, Exists, NotExists, Graph, Service, Subquery, and each branch of Union — and reconstructs the surrounding variant. Leaf variants pass through. Doc comment is explicit that callers intercept variants their semantics shouldn't descend into (e.g. `encode_patterns_for_reasoning` and `geo_rewrite::rewrite_recursive` both carve out Service/Subquery before falling through; `dedup_exact_triples` recurses into them because triple dedup is idempotent under SPARQL set semantics).
- **Rewrite passes consolidated.** The repeated 8-arm match blocks in `rewrite.rs`, `geo_rewrite.rs`, `r2rml/rewrite.rs`, and `rewrite_owl_ql.rs` are replaced by `pattern.map_subpatterns(&mut recurse)` (or a small `match` for the variants that do need pass-specific carve-outs).
- **Test surface tightened.** `it_graph_source_r2rml.rs` consolidates four near-identical `ContextConfig { tracker: Some(&t), r2rml: Some((&p, &p)), ..Default::default() }` blocks behind a small generic `r2rml_test_config` helper. `groupby_aggregate_tests.rs`, `owl2rl_property_rules_integration_tests.rs`, and `values_bind_union_tests.rs` migrate from the removed wrappers to `execute(..., ContextConfig::default())`.

## Behavior

No semantic changes for callers. The `dedup_exact_triples` rewrite is now slightly more thorough — it descends into Service and Subquery scopes through `map_subpatterns`, where the original explicit match block stopped at the boundary. Dedup of identical `Pattern::Triple` nodes is idempotent in any scope under SPARQL set semantics, and the remote endpoint sees an equivalent (smaller) query, so this is a strict optimization.


bplatz (approved):
🧭

PR #1217, opened 2026-05-04
fix: SPARQL UPDATE commits referencing fresh namespaces aren't queryable post-publish

bplatz:
SPARQL `INSERT DATA` (or any `lower_sparql_update`-derived `Txn`) committed against a fresh ledger leaves committed flakes whose namespace_codes the post-commit snapshot can't resolve back to IRIs. A subsequent `SELECT` returns no rows — even though the flakes are physically in storage.

In production this is normally masked: JSON-LD inserts pre-allocate the namespaces before any SPARQL touches them, so the staging registry happens to land on the same codes. The bug surfaces when SPARQL is the *first* commit allocating a given namespace.

## Root cause

`lower_sparql_update` allocates IRIs against a **caller-owned** `NamespaceRegistry` to build the templates' `Sid`s (e.g. `ex:doc1` → `Sid{13, "doc1"}`). `stage_transaction_from_txn` then builds an **independent** registry from the ledger snapshot, never sees those allocations, and commits flakes carrying namespace_code 13. The staging registry's persistence delta is empty for `http://example.org/`, so the commit envelope omits the `(13, "http://example.org/")` mapping. On reload, the snapshot's namespace map doesn't contain code 13, and `sid_for_iri("http://example.org/...")` re-allocates fresh — predicate Sid mismatch, no rows.

Verified with diagnostic prints comparing SPARQL vs. JSON-LD commit state on a fresh ledger:

```
SPARQL: snapshot.namespaces() = {... 12: "urn:fluree:"}        ← missing 13
JSON-LD: snapshot.namespaces() = {... 12: "urn:fluree:", 13: "http://example.org/"}
```

## Fix

A two-step handoff so allocations made during lowering survive into the commit's persisted namespace map.

- **`fluree-db-transact/src/ir.rs`** — `Txn` gains a `namespace_delta: HashMap<u16, String>` field, documented as the lowering→staging contract. JSON-LD producers (`parse_transaction`) share the staging registry directly and leave it empty.
- **`fluree-db-transact/src/lower_sparql_update.rs`** — `lower_sparql_update` captures `ns.delta().clone()` into the resulting `Txn.namespace_delta` after the inner lowering call. All four inner Txn literals (`InsertData`, `DeleteData`, `DeleteWhere`, `Modify`) initialize the field.
- **`fluree-db-core/src/ns_encoding.rs`** — new `NamespaceCodes::adopt_delta_for_persistence` adds entries to both the lookup tables AND the persistence delta. Distinct from the existing `merge_delta`, which only updates lookup tables (it's used for post-publish sync from the shared allocator and intentionally skips delta to avoid re-persisting). Same conflict-checking semantics as `merge_delta`.
- **`fluree-db-transact/src/namespace.rs`** — `NamespaceRegistry::adopt_delta_for_persistence` thin wrapper.
- **`fluree-db-api/src/tx.rs`** — `stage_transaction_from_txn` calls `adopt_delta_for_persistence(&txn.namespace_delta)` immediately after building the staging registry. On conflict, surfaces as `ApiError::Internal` (would indicate the caller staged against a stale snapshot).

## Tests

- **`fluree-db-api/tests/it_vector_corruption_repro.rs`** — un-ignores `sparql_insert_data_embedding_vector_literal_round_trips_after_indexing`; tightens it to verify all 4 vector element values, not just length. Updates the module comment to reflect that only one deferred concern remains (vector retraction). The test exercises the fix on a vector typed literal, but the same fix applies to all SPARQL-lowered Txns regardless of value type.

  Pre-fix: `INSERT DATA { ex:doc1 ex:embedding "[0.1, 0.2, 0.3, 0.4]"^^f:embeddingVector }` on a fresh ledger → published commit has the flake → SELECT returns `[]`.
  Post-fix: same flow → SELECT returns `[[[0.1, 0.2, 0.3, 0.4]]]`.

The diagnostic baseline I built during investigation (showing the same bug for a plain string `INSERT DATA { ex:doc1 ex:name "Alice" }` on a fresh ledger) is not committed — the un-ignored vector test exercises the same code path and is sufficient regression coverage.

## Notes for reviewers

- The `merge_delta` vs. `adopt_delta_for_persistence` split is load-bearing. `merge_delta` is called post-publish to sync the serial registry with what the shared allocator already persisted — adding to delta there would double-persist on the next commit. The new method is for the inverse case: adopting allocations that *haven't* been persisted yet.
- The fix is agnostic to value type. The vector typed literal in the test is incidental — any SPARQL-only namespace allocation on a fresh ledger had the same bug. JSON-LD already shared a single registry between parse and stage and was unaffected.
- Existing SPARQL UPDATE tests (e.g. `sparql_update_supports_subquery_aggregate_and_bind` in `fluree-db-server/tests/integration.rs`) didn't catch this because they always seed via JSON-LD INSERT first, pre-allocating the relevant namespaces.

### DIFF
--- a/fluree-db-transact/src/lower_sparql_update.rs
+++ b/fluree-db-transact/src/lower_sparql_update.rs
@@ -234,20 +234,28 @@ pub fn lower_sparql_update(
     let mut vars = VarRegistry::new();
     let mut bnodes = BlankNodeCounter::new();
 
-    match op {
+    let mut txn = match op {
         UpdateOperation::InsertData(insert) => {
-            lower_insert_data(&insert.data, prologue, ns, &mut vars, &mut bnodes, opts)
+            lower_insert_data(&insert.data, prologue, ns, &mut vars, &mut bnodes, opts)?
         }
         UpdateOperation::DeleteData(delete) => {
-            lower_delete_data(&delete.data, prologue, ns, &mut vars, &mut bnodes, opts)
+            lower_delete_data(&delete.data, prologue, ns, &mut vars, &mut bnodes, opts)?
         }
         UpdateOperation::DeleteWhere(delete_where) => {
-            lower_delete_where(&delete_where.pattern, prologue, ns, &mut vars, opts)
+            lower_delete_where(&delete_where.pattern, prologue, ns, &mut vars, opts)?
         }
         UpdateOperation::Modify(modify) => {
-            lower_modify(modify, prologue, ns, &mut vars, &mut bnodes, opts)
+            lower_modify(modify, prologue, ns, &mut vars, &mut bnodes, opts)?
         }
-    }
+    };
+    // Hand off the lowering registry's allocations so `stage_transaction_from_txn`
+    // can merge them into its own snapshot-derived registry. Without this, the
+    // first SPARQL `INSERT DATA` on a fresh ledger commits flakes referencing
+    // namespace codes the staging registry never learned about — the commit's
+    // persisted namespace map omits them, and post-commit SELECT can't resolve
+    // the predicate IRI back to the same Sid.
+    txn.namespace_delta = ns.delta().clone();
+    Ok(txn)
 }
 
 /// Lower INSERT DATA operation.
@@ -276,6 +284,7 @@ fn lower_insert_data(
         vars: mem::take(vars),
         txn_meta: Vec::new(),
         graph_delta: FxHashMap::default(),
+        namespace_delta: std::collections::HashMap::new(),
     })
 }
 
@@ -306,6 +315,7 @@ fn lower_delete_data(
         vars: mem::take(vars),
         txn_meta: Vec::new(),
         graph_delta: FxHashMap::default(),
+        namespace_delta: std::collections::HashMap::new(),
     })
 }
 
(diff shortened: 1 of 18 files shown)


fluree-db-transact/src/lower_sparql_update.rs
```diff
@@ -234,20 +234,28 @@ pub fn lower_sparql_update(
...
-    }
+    };
+    // Hand off the lowering registry's allocations so `stage_transaction_from_txn`
+    // can merge them into its own snapshot-derived registry. Without this, the
+    // first SPARQL `INSERT DATA` on a fresh ledger commits flakes referencing
+    // namespace codes the staging registry never learned about — the commit's
+    // persisted namespace map omits them, and post-commit SELECT can't resolve
+    // the predicate IRI back to the same Sid.
+    txn.namespace_delta = ns.delta().clone();
+    Ok(txn)
```
aaj3f, line 258: Not really a recommendation, just thinking it could be easy to regress the work these final two lines do if a future code didn't realize how load-bearing this is and did something like directly return `Option<Txn>` from one of the `lower_*` fns above, where `namespace_delta` would still be empty.

aaj3f (approved):
Looks good to me. One minor note but not really something needing to be fixed. I also thought for a moment about if the retry loop on `NamespaceConflict` could be perf-tuned by hoisting out any expensive work that isn't retry-relevant, but it's such a rare path it doesn't seem worth it.

PR #1219, opened 2026-05-04
Reorganize the query IR module

zonotope:
## Summary

This patch builds on #1215 and restructures `fluree-db-query`'s logical IR around a single `ir` module, collapses two redundant query types into one canonical `Query`, and replaces the free-standing `pattern_contains_function` and dual-purpose `variables()` helpers with cleaner method APIs on `Pattern`. 

## Motivation

`fluree-db-query/src/ir.rs` had grown to ~4,400 lines holding the entire pattern AST, projection specs, expression AST, property paths, adapter scan patterns, and triple-pattern primitives in one file. `triple.rs` lived a directory up. `options.rs` lived at the crate root. There were two structs both called something like a "query": a tiny `Query { select, where_ }` used by a handful of tests, and the much richer `ParsedQuery` that carried output spec, options, projection shape, construct templates, and post-VALUES — the type that actually flows through parse → plan → execute → format. Free-standing helpers (`pattern_contains_function`, `variables`) lived alongside the types they walked, and `variables` returned a single set whose meaning ("referenced" vs "bound") was context-dependent at the call site.

## Key changes

- **Split `ir.rs` into a module.** `fluree-db-query/src/ir/` now contains:
  - `query.rs` — `Query`, `QueryOutput`, `ProjectionShape`, `ConstructTemplate` (the resolved-and-lowered query)
  - `options.rs` — `QueryOptions` (LIMIT/OFFSET/ORDER BY/GROUP BY/aggregates/HAVING/post-binds/DISTINCT) and `ReasoningModes`
  - `pattern.rs` — `Pattern` enum and the `Vec<Pattern>`-wrapping variants (`SubqueryPattern`, `ServicePattern`, `GraphName`, `ServiceEndpoint`)
  - `expression.rs` — filter/bind expression AST, `CompareOp`, `ArithmeticOp`, `Function` catalog
  - `path.rs` — property-path patterns
  - `projection.rs` — graph-crawl selection specs (`GraphSelectSpec`, `NestedSelectSpec`, `SelectionSpec`, `Root`)
  - `adapters.rs` — non-graph scan patterns (BM25, vector, geo, S2, R2RML)
  - `triple.rs` — `TriplePattern`, `Ref`, `Term` (moved in from the crate root)
- **`ir.rs` becomes a thin facade** — module declarations and the public re-export surface.
- **One `Query`, not two.** The simple `Query { select, where_ }` struct (used by ~2 tests) is removed; `ParsedQuery` is renamed to `Query` and lives at `ir::Query`. All consumers (planner, executor, formatter, SPARQL lower, transact stage, view layer) are migrated.
- **`QueryOptions` and `ReasoningModes` move into the IR.** The crate-root `options.rs` is gone; the rewriter's locally-defined `ReasoningModes` is gone. Both now live in `ir/options.rs` and re-export through the `ir` facade.
- **`pattern_contains_function` becomes a method on `Pattern`.** Old: `pattern_contains_function(&pattern, &target)`. New: `pattern.contains_function(&target)`. Mirrors the existing `Expression::contains_function` shape.
- **`variables()` splits into `referenced_vars()` and `produced_vars()`.** Previously a single method whose semantics ("which variables does this pattern *reference*?" vs "which does it *produce* / bind?") had to be inferred per call site. Now explicit on every IR type that has both notions (`Pattern`, `TriplePattern`, `PropertyPathPattern`, `IndexSearchPattern`, `VectorSearchPattern`, `GeoSearchPattern`, `S2SearchPattern`, `R2rmlPattern`, `SubqueryPattern`, `ServicePattern`). Filter / Minus correctly return empty `produced_vars`; Subquery correctly exposes only its SELECT outwards.
- **Dead-code sweep.** Removed several unused IR helpers, a method only referenced by a single test, and the redundant simple `Query` struct.

## Behavior

No semantic changes. Pure structural reorganization: file moves, type renames, helper-shape changes. The `referenced_vars` / `produced_vars` split is a contract refinement — the underlying recursion walks the same set of pattern variants the old `variables()` did. All in-workspace consumers (`fluree-db-sparql/src/lower/*`, `fluree-db-api/src/{format,view,query}/*`, `fluree-db-transact/src/stage.rs`, `fluree-db-reasoner/src/types.rs`, the test suites) are migrated to the new APIs in this PR.


bplatz (approved):
Awesome!

PR #1220, opened 2026-05-05
Restructure query output type hierarchy

zonotope:
## Summary

This branch builds on #1219 and restructures the query output type hierarchy in `fluree-db-query` so that combinations the engine cannot produce are also combinations the type system cannot construct. Eliminates the standalone `graph_select` field that could legally coexist with a `SELECT *`, replaces a parallel `(vars, has_wildcard, hydration_spec)` representation with a single `Projection` enum that mirrors the actual SPARQL semantics, and renames a handful of types and methods so that the names match what they do. Net effect: 53 files changed, ~1,248 insertions / ~1,203 deletions, with no behavior changes.

## Motivation

`QueryOutput::Select { vars: Vec<VarId>, ... }` carried only the bound-variable list. Hydration (the "graph crawl" / nested JSON-LD expansion) lived as a separate `graph_select: Option<HydrationSpec>` field on `Query`. Nothing in the type system prevented `SELECT *` (wildcard) from coexisting with a hydration spec, even though the formatter rejects that combination at runtime — and the parser had to encode wildcard-vs-explicit selection through a `has_wildcard: bool` flag duplicated alongside the actual selection list. `NestedSelectSpec` had the same issue at every level of nested expansion: a `has_wildcard` flag plus a `Vec<SelectionSpec>` that the formatter had to cross-check. `QueryOutput::Boolean` was the only variant whose name didn't match standard SPARQL terminology (everywhere else, `Construct`, `Select`, `Ask`).

## Key changes

- **`QueryOutput::Select` carries `Projection` and `Multiplicity`.** Replaces `Select { vars: Vec<VarId>, ... }` + `graph_select: Option<HydrationSpec>` with `Select { projection: Projection, multiplicity: Multiplicity }`. New types:
  - `Projection { Wildcard, Tuple(Vec<Column>), Scalar(Column) }` — the column structure of the projection. `Wildcard` is now a *variant*, not a flag.
  - `Column { Var(VarId), Hydration(HydrationSpec) }` — what each output column produces. Mixed `Var` + `Hydration` columns in one row are representable by construction.
  - `Multiplicity { All, One }` — replaces the prior `selectOne` discrimination.
- **`NestedSelectSpec` becomes `enum { Wildcard { refinements, reverse }, Explicit { forward, reverse } }`.** Wildcard-vs-explicit is encoded structurally rather than via a `has_wildcard: bool` plus a `Vec<SelectionSpec>` whose contents are interpreted differently depending on the flag. New helpers (`is_wildcard()`, `select_predicate()`, `includes_id()`, `reverse()`) flatten what would otherwise be deep matches in the hydration formatter.
- **`SelectionSpec` deleted, replaced by `ForwardItem { Id, Property { predicate, sub_spec } }`.** The `@id` selection is now its own variant rather than a special-cased `Wildcard` flag.
- **Parse AST mirrors the IR.** `UnresolvedProjection`, `UnresolvedColumn`, `UnresolvedNestedSelectSpec`, `UnresolvedForwardItem` now have the same shape as their IR counterparts, so lowering is a structural translation rather than a multi-source assembly.
- **`SelectMode::Wildcard` dropped.** Wildcard is now reachable via `Projection::Wildcard` directly; the parser-internal `SelectMode` enum no longer has to carry it.
- **`QueryOutput::Boolean` renamed to `Ask`** to match SPARQL terminology and the rest of the variant family.
- **Method renames for clarity.**
  - `QueryOutput::select_vars()` getter → `projected_vars()`
  - `QueryOutput::select_vars_or_empty()` → `projected_vars_or_empty()`
  - `QueryOutput::select(vars)` constructor → `select_vars(vars)` (free of the getter-name collision)
  - `QueryOutput::select_one(vars)` → `select_one_var(vars)`

## Behavior

No semantic changes. Pure type-system tightening: states the engine never produced are no longer expressible. The hydration formatter, planner detect-helpers, and SPARQL/JSON-LD lower paths are migrated to the new shape. All in-workspace consumers (`fluree-db-api/src/{format,view,query,graph_source}/*`, `fluree-db-cli/src/{output,commands}/*`, `fluree-db-sparql/src/lower/*`, the test suites) are updated.

One incidental cleanup: the 13 detect-helper functions in `fluree-db-query/src/execute/operator_tree.rs` previously called `query.output.projected_vars()` twice each (once for an `is_none()` guard, once to bind the variables). The new `projected_vars()` returns an owned `Vec<VarId>` (forced by the `Vec<Column>` storage), so the double call was a double allocation. The early `is_none()` guard was redundant with the existing `?` later in each function — removed.

### DIFF
--- a/fluree-db-query/src/parse/mod.rs
+++ b/fluree-db-query/src/parse/mod.rs
@@ -157,7 +157,7 @@ fn parse_query_ast_internal(
         )?;
         // LIMIT 1 for efficiency — only need to know if any solution exists
         query.options.limit = Some(1);
-        return Ok((query, SelectMode::Boolean));
+        return Ok((query, SelectMode::Ask));
     }
 
     // Determine select mode based on which key is present.
--- a/fluree-db-query/src/parse/mod.rs
+++ b/fluree-db-query/src/parse/mod.rs
@@ -47,9 +47,9 @@ pub use lower::{lower_unresolved_pattern, lower_unresolved_patterns};
 pub use policy::{JsonLdParseCtx, JsonLdParsePolicy};
 pub use where_clause::parse_where_with_counters;
 
-use crate::ir::{Expression, ProjectionShape, Query};
+use crate::ir::{Expression, Query};
 use crate::var_registry::VarRegistry;
-use ast::UnresolvedPathExpr;
+use ast::{SelectShape, UnresolvedPathExpr};
 use fluree_graph_json_ld::{parse_context, ParsedContext};
 use serde_json::Value as JsonValue;
 use std::collections::HashMap;
@@ -552,7 +552,7 @@ fn parse_select(
         JsonValue::Array(arr) => {
             // Record shape: array form always yields tuple rows, regardless of
             // how many items are inside (`["?x"]` → `[[v1],[v2]]`, not `[v1,v2]`).
-            query.select_shape = ProjectionShape::Tuple;
+            query.select_shape = SelectShape::Tuple;
             for item in arr {
                 match item {
                     JsonValue::String(s) => {
@@ -579,7 +579,7 @@ fn parse_select(
 
         // Case 1: Single string form: "?x" — bare variable, scalar shape.
         JsonValue::String(s) => {
-            query.select_shape = ProjectionShape::Scalar;
+            query.select_shape = SelectShape::Scalar;
             parse_select_string(s, query)?;
         }
 
--- a/fluree-db-query/src/parse/mod.rs
+++ b/fluree-db-query/src/parse/mod.rs
@@ -192,7 +192,7 @@ fn parse_query_ast_internal(
         parse_select(select, &ctx, &mut query)?;
     }
 
-    // Parse depth parameter and apply to hydrationion if present
+    // Parse depth parameter and apply to hydration if present
     if let Some(gs) = query.hydration_mut() {
         gs.depth = options::parse_depth(obj)?;
     }
@@ -864,9 +864,7 @@ fn parse_hydration_object(
 }
 
 /// Build an optional boxed nested select spec, returning `None` when empty.
-fn make_nested_spec(
-    level: UnresolvedNestedSelectSpec,
-) -> Option<Box<UnresolvedNestedSelectSpec>> {
+fn make_nested_spec(level: UnresolvedNestedSelectSpec) -> Option<Box<UnresolvedNestedSelectSpec>> {
     if level.is_empty() {
         None
     } else {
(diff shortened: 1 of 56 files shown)


fluree-db-query/src/parse/mod.rs
```diff
@@ -542,28 +535,27 @@ fn parse_construct_template(
...
             for item in arr {
                 match item {
                     JsonValue::String(s) => {
-                        parse_select_string(s, query)?;
+                        columns.push(parse_select_string(s, &mut query.options.aggregates)?);
                     }
                     JsonValue::Object(map) => {
-                        let spec = parse_graph_select_object(map, ctx, query)?;
-                        query.graph_select = Some(spec);
+                        let spec = parse_hydration_object(map, ctx, query)?;
```
bplatz: The "at most one hydration" invariant is broken for the array-form select. `parse_hydration_object` checks `query.hydration()`, but `query.select` isn't assigned until after this loop completes — so on the second `JsonValue::Object` iteration, `query.hydration()` still reads the stale (empty) `query.select` and the duplicate check passes.

Repro: `select: ["?a", {"?b": ["*"]}, {"?c": ["*"]}]` parses without error and lowers to `Projection::Tuple([Var(a), Hydration(b), Hydration(c)])`. `Projection::hydration()` uses `find_map`, so the second hydration is silently dropped at format time.

Pre-refactor this case errored, because the old code set `query.graph_select = Some(spec)` immediately inside the loop. Smallest fix is a local guard against `columns` here before pushing:

```rust
if columns.iter().any(|c| matches!(c, UnresolvedColumn::Hydration(_))) {
    return Err(ParseError::InvalidSelect(
        "only one graph-select object allowed per query".to_string(),
    ));
}
```

Worth a focused parser test for the multi-object array case. The doc on `Projection::hydration()` ("at most one; enforced by the parser") is currently aspirational without this.

bplatz (approved):
Besides one comment, looks good!

fluree-db-query/src/parse/mod.rs
zonotope: @bplatz I decided to allow multiple hydrations (or graph crawls) in the select clause instead. I couldn't find a performance or architectural reason not to, and I think a new user would expect to be able to do it and would be surprised by the error. 

Here is the [implementing commit range](https://github.com/fluree/db/compare/b94e74128...e1bc2ec). I'm happy to roll it back if you disagree that this is something we should support.

bplatz: > @bplatz I decided to allow multiple hydrations (or graph crawls) in the select clause instead. I couldn't find a performance or architectural reason not to, and I think a new user would expect to be able to do it and would be surprised by the error.
> 
> Here is the [implementing commit range](https://github.com/fluree/db/compare/b94e74128...e1bc2ec). I'm happy to roll it back if you disagree that this is something we should support.

Great!

PR #1221, opened 2026-05-06
fix: vector retraction round-trips across rebuild, overlay, and incremental publish paths

bplatz:
`SPARQL DELETE WHERE` (and any `op=false` vector flake) on `f:embeddingVector`
data was not cancelling the indexed assertion — across all three indexing
paths (rebuild, novelty-overlay, incremental publish). And even when it
appeared to cancel, it could cancel the wrong row. The bug had multiple
overlapping causes around how vector arena handles are minted, retracted,
and re-resolved.

This PR makes vector retractions round-trip correctly across **all three**
paths and pins the contract with regression tests for every known
mis-cancellation shape — including the most common production path
(incremental publish, which runs whenever a base index already exists and
the commit gap is small).

## Context

A vector fact has two identities:

- **Logical identity** — `(s, p, FlakeValue::Vector([..]), dt=embeddingVector, m)`
- **Indexed identity** — `(s_id, p_id, o_type=VECTOR_ID, o_key=<arena handle>, o_i)`

Cancellation matches by indexed identity (`o_kind`/`o_key` tuple). The
transaction layer was already producing the right *logical* retraction; the
broken part was translating that into the same indexed identity as the
original assertion. Allocating a fresh arena handle for each retraction
guarantees the `o_key` differs from the assertion's, and merge cancellation
silently misses.

### 1. Resolver fact-identity tracking + dedup (`fluree-db-indexer`)

`SharedResolverState`/`GlobalDicts` gain a new map:

```rust
pub vector_fact_handles:
    FxHashMap<GraphId, FxHashMap<(u16, Arc<str>, u32, u32, Vec<u32>), u32>>
//                                 g_id  → (s_ns_code, s_name, p_id, o_i, f32_bits) → handle
```

The fact-identity key is keyed by **subject-name space** (not s_id) and
includes the **f32 bit pattern** of the quantized vector value:

- Subject-name (not s_id) so the key is stable across the chunk-local
  vs. global s_id boundary inherent to the incremental-resolve
  pipeline. Both rebuild and incremental paths populate / look up the
  same map shape.
- `f32_bits` because `(s, p, o_i)` alone is not unique:
  - Two distinct subjects can hold the same value under the same
    predicate (each gets its own arena handle).
  - One subject can hold MULTIPLE different values under the same
    predicate without a list index (all share `o_i = LIST_INDEX_NONE`).

`resolve_object` (and its chunk variant) now:

- **On assertion**: looks up the fact key; reuses the existing handle if
  present (dedup), otherwise inserts to the arena and records the new
  handle.
- **On retraction**: looks up the fact key; uses the existing handle if
  found, else returns `Ok(None)` — a no-op skip rather than aborting
  the rebuild.

Returning `Option<RunRecord>` lets `resolve_op` skip emitting records for
unmatched retractions (DELETE DATA for never-asserted vectors, historical
no-ops). Other datatypes already encoded unmatched retractions
harmlessly; this brings vectors in line.

### 2. Overlay fact-identity scan (`fluree-db-query`, `fluree-db-binary-index`)

New `BinaryIndexStore::find_vector_handle_by_fact(g_id, s_id, p_id, o_i, value)`:

- SPOT-prefix scan filtered by `(s_id, p_id, o_type=VECTOR, o_i)`.
- For each candidate row, decodes the arena handle's stored f32 vector
  and bit-compares to `value` before returning.

`binary_scan::translate_one_flake_v3_pub` short-circuits `FlakeValue::Vector`
overlay flakes through this lookup so the overlay's
`(s_id, p_id, o_type, o_key, o_i)` tuple matches the indexed assertion
without requiring re-publish. The fallback (no matching base row) goes
through the existing materialized-flake path.

`g_id` is now plumbed into `translate_one_flake_v3_pub` (was implicit at
the for-each-overlay-flake level). Five callers updated:
`binary_scan`, `binary_range`, `fast_group_count_firsts`,
`fast_path_common`, `geo_search`. `store: &Arc<BinaryIndexStore>` ripples
through `translate_overlay_flakes*` and `export.rs::translate_overlay`.

### 3. Incremental publish: pre-load base arenas + seed fact map (`fluree-db-indexer`)

This is the most common production path — once a base index exists and
the commit gap is small, every publish goes through `incremental_resolve`,
not full rebuild. Pre-fix, the chunk-local fact map only knew assertions
in the new chunk, so retractions of base-asserted vectors silently
no-op'd at incremental publish (the user's DELETE appeared to work
against novelty queries until publish, then reverted).

Two new steps in `incremental_resolve.rs`:

- **4b. Pre-load base vector arena shards.** Where the existing code
  recorded only `base_vector_counts`, we now fetch each manifest's
  shard CIDs, decode the VAS1 bytes, and reassemble a `VectorArena` via
  `load_arena_from_shards`. The arena gets seeded into
  `shared.vectors[g][p]` so chunk inserts append to the unified base+chunk
  arena and return globally-correct handles directly.
- **4c. Seed `vector_fact_handles` from base SPOT.** New helper
  `seed_vector_fact_handles` constructs a `BinaryIndexStore::load_from_root_v6`
  and SPOT-scans for `o_type=VECTOR` rows. For each row it decodes
  `(s_ns_code, s_name)` via the new `BinaryIndexStore::resolve_subject_parts`,
  reads the `f32_bits` from the just-loaded arena at the row's handle,
  and inserts `(s_ns_code, s_name, p_id, o_i, f32_bits) → handle` into
  the shared fact map.

The VECTOR_ID offset pass is gone — handles are already global because
chunk inserts append to the pre-loaded base. Replaced with a
`debug_assert!` sanity guard that catches any out-of-range o_key.

This also fixes a related dedup gap: re-asserting the same logical
vector fact across commits (assert → publish → assert same) now reuses
the base arena's handle via the seeded fact map, producing one encoded
row instead of two.

### 4. Arena reverse-lookup primitives (`fluree-db-binary-index`)

`VectorArena::find_handle_by_value` and `LazyVectorArena::find_handle_by_value`
— bit-equality scans that the new `find_vector_handle_by_fact` uses to
compare candidate rows. Public on the read-side too in case future
diagnostics need it. New `BinaryIndexStore::resolve_subject_parts(s_id) → (u16, String)`
returns the canonical (ns_code, suffix) pair without joining into IRI,
used by `seed_vector_fact_handles` for fact-map keying.

## Tests

Twelve tests in `fluree-db-api/tests/it_vector_corruption_repro.rs`, all
green. New / un-ignored in this PR:

- `jsonld_context_vector_bare_array_retracts_after_indexing` (un-ignored)
  — post-rebuild round-trip
- `jsonld_context_vector_bare_array_retracts_via_novelty_overlay` —
  pre-rebuild novelty-overlay round-trip
- `jsonld_vector_retracts_via_incremental_publish` — **incremental
  publish path**: insert → publish (full rebuild) → DELETE → publish
  (incremental) → COUNT == 0
- `jsonld_context_vector_duplicate_values_retract_one_keeps_other` —
  two subjects, same vector value, retract one keeps the other (rebuild)
- `jsonld_context_vector_duplicate_values_overlay_retract_one_keeps_other`
  — same case via overlay
- `jsonld_multi_valued_vectors_retract_one_keeps_other` — one subject
  with two different vector values under the same predicate without a
  list index; retract one keeps the other (rebuild)
- `jsonld_multi_valued_vectors_overlay_retract_one_keeps_other` — same
  case via overlay
- `sparql_delete_data_unmatched_vector_is_no_op` — `DELETE DATA` for a
  never-asserted vector must not abort the transaction
- `jsonld_re_asserting_same_vector_dedups` — assert → publish → assert
  the same logical fact again → COUNT == 1; delete → COUNT == 0
  (verifies dedup AND that one retraction cancels the deduped row)

New test helper `support::build_and_publish_index` uses
`build_index_for_record` (which dispatches to incremental indexing when
a base exists). Distinct from `rebuild_and_publish_index` which always
does a full rebuild — the incremental test specifically uses the new
helper to exercise that path.

## Operational note

Pre-loading base vector arenas at every incremental publish adds
bounded I/O (one `cs.get` per shard CID per (g, p) arena). For ledgers
with no vectors there's no overhead. For ledgers with many vectors,
the cost is proportional to total vector data — typically dwarfed by
SPOT scan and run-record processing. If profiling identifies this as
a hot path, the `seed_vector_fact_handles` helper could grow a slimmer
"scan-only" loader that skips the duplicate dict/arena loading inside
`BinaryIndexStore::load_from_root_v6`.

## Coverage matrix

| Operation                          | Full rebuild | Pre-rebuild (overlay) | Incremental publish |
|------------------------------------|:-:|:-:|:-:|
| Insert vector                      | ✅ | ✅ | ✅ |
| Query vector                       | ✅ | ✅ | ✅ |
| Retract vector asserted same chunk | ✅ | ✅ | ✅ |
| Retract vector asserted in base    | ✅ | ✅ | ✅ |
| Re-assert same logical fact        | ✅ | ✅ | ✅ |


fluree-db-indexer/src/run_index/build/incremental_resolve.rs
aaj3f, line 1: The correctness fixes in this file make sense. My question (specifically re: 4b and 4c) is how much the work/latency increase is on incremental publishes even on commits with no vector data. It seems that even on commits that don't touch any vector data, 4b will reassemble the entire vector arena and 4c will walk SPOT for every vector-type flake. We could always create a follow-up issue w/ an otel bench to see if perf hit is real and, if so, how much it hinges on 4b and 4c. If it is/was real, at the very least we could make 4b/4c conditional on the chunk actually including vectors.

aaj3f (approved):
Seems like a significant fix! One note below that *could* (but not sure) mean a big perf hit w/ incremental publish, but other than that 👍

fluree-db-indexer/src/run_index/build/incremental_resolve.rs
bplatz: Addressed in 46c84632b7f29077b35fc11335b9e71871a0ae29

PR #1222, opened 2026-05-06
add owl imports cookbook to docs

bplatz:
Adds new cookbook on how to use owl:imports to docs

### DIFF
--- a/docs/guides/cookbook-owl-imports.md
+++ b/docs/guides/cookbook-owl-imports.md
@@ -19,9 +19,8 @@ A runnable end-to-end script lives at
 [`cookbook-owl-imports.sh`](cookbook-owl-imports.sh) — it executes every
 command in this tutorial and asserts the expected rows.
 
-> Reasoning is currently **experimental**. See
-> [Reasoning and inference](../concepts/reasoning.md) for background and
-> [Setting groups → reasoningDefaults](../ledger-config/setting-groups.md)
+> See [Reasoning and inference](../concepts/reasoning.md) for background
+> and [Setting groups → reasoningDefaults](../ledger-config/setting-groups.md)
 > for the full config schema.
 
 ---
--- a/docs/guides/cookbook-owl-imports.md
+++ b/docs/guides/cookbook-owl-imports.md
@@ -15,10 +15,6 @@ imports are not supported. This tutorial uses three named graphs in one ledger:
 | `<http://example.org/onto/behaviors>`      | Imported ontology — property characteristics |
 | `<urn:fluree:demo:main#config>`            | Ledger config — wires up reasoning  |
 
-A runnable end-to-end script lives at
-[`cookbook-owl-imports.sh`](cookbook-owl-imports.sh) — it executes every
-command in this tutorial and asserts the expected rows.
-
 > See [Reasoning and inference](../concepts/reasoning.md) for background
 > and [Setting groups → reasoningDefaults](../ledger-config/setting-groups.md)
 > for the full config schema.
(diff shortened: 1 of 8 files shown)


docs/guides/cookbook-owl-imports.md
```diff
@@ -0,0 +1,361 @@
...
+# Connection-scoped (specify ledger via query string)
+curl -X POST 'http://localhost:8090/fluree/upsert?ledger=demo:main' \
+     -H 'Content-Type: application/trig' \
+     --data-binary @02-ontology.trig
+
+# Ledger-scoped path form
+curl -X POST 'http://localhost:8090/fluree/upsert/demo:main' \
+     -H 'Content-Type: application/trig' \
+     --data-binary @02-ontology.trig
+```
```
aaj3f, line 325: I think the route has that `/v1/` namespace, right? For example, `http://localhost:8090/v1/fluree/upsert?ledger=demo:main`. May want to check all the URLs in the PR to ensure they're correct.

docs/guides/cookbook-owl-imports.md
```diff
@@ -0,0 +1,361 @@
...
+| Graph IRI                                  | Role                                |
+|--------------------------------------------|-------------------------------------|
+| *(default graph)*                          | Instance data                       |
+| `<http://example.org/onto/core>`           | Core ontology — class hierarchy + `owl:imports` hub |
+| `<http://example.org/onto/behaviors>`      | Imported ontology — property characteristics |
+| `<urn:fluree:demo:main#config>`            | Ledger config — wires up reasoning  |
+
+A runnable end-to-end script lives at
+[`cookbook-owl-imports.sh`](cookbook-owl-imports.sh) — it executes every
+command in this tutorial and asserts the expected rows.
```
aaj3f: This script file isn't in the PR

docs/guides/cookbook-owl-imports.md
```diff
@@ -0,0 +1,361 @@
...
+
+```bash
+fluree upsert -f 01-data.ttl
+# → Committed t=1, 8 flakes
+```
+
+> Use `upsert` (not `insert`) for any TriG document that contains `GRAPH`
+> blocks. `fluree insert` takes a fast path that only understands plain
+> Turtle. `upsert` routes Turtle/TriG through the parser that recognises
+> `GRAPH` blocks. Plain Turtle works through either command.
```
aaj3f: I'm not sure this is true for the CLI. I think both go through `parse_trig_phase1()`. For HTTP routes, `/insert` blocks `Content-Type: application/trig` outright, so `/upsert` is required. In direct uses of `fluree-db-api`, there's the top-level `Fluree::insert_turtle(...)` which does bypass the TriG phase-1 extraction, but the CLI doesn't use that path and with the CLI both go through `parse_trig_phase1()`

Not sure if we just want to amend the comment or if there's any interest in making the code actually take the fast path that the comment suggests.

aaj3f (approved):
A few notes below but 👍

docs/guides/cookbook-owl-imports.md
bplatz: Good catch — fixed in acea5fee9. Both curl examples now use `/v1/fluree/upsert`. Verified those were the only HTTP URL examples in the doc.

docs/guides/cookbook-owl-imports.md
bplatz: Right — removed the reference in acea5fee9 rather than checking in the script.

docs/guides/cookbook-owl-imports.md
bplatz: Traced this — the CLI `insert` path *does* skip `parse_trig_phase1`. `commands/insert.rs` calls `graph.transact().insert_turtle(content)`, which routes through `commit_with_handle` (`tx_builder.rs:749`), and at line 794 hits the `// Direct flake path for InsertTurtle (bypass JSON-LD / IR)` branch that calls `Fluree::stage_turtle_insert`. That uses `FlakeSink` to parse Turtle straight to flakes — `parse_trig_phase1` is only called from the `to_json_with_trig_meta` branch, which is hit by `UpsertTurtle` but not `InsertTurtle`.

That said, your point that the wording was vague is fair. Tightened the callout in acea5fee9 to be more specific:

> The CLI's `insert` path parses Turtle straight to flakes and does not extract `GRAPH` blocks; over HTTP, `/v1/fluree/insert` rejects `Content-Type: application/trig` outright. `upsert` handles both Turtle and TriG.

PR #1223, opened 2026-05-06
fix: batched join probes ignored time-travel snapshot

bplatz:
SPARQL/JSON-LD time-travel queries silently returned latest-state results
when the WHERE clause combined `?s a <Class>` with a same-subject triple
whose object was a literal (`?s :status "paid"`) or a variable joined to
the type pattern. The `@t:N` snapshot qualifier on the dataset was ignored
in those shapes, so historical UIs showed the same data at every `t`.

This change gates the affected batched-probe paths on
`to_t == store.max_t()` and falls back to the per-row scan path
(`BinaryCursor::set_to_t` + history-sidecar replay) for historical
snapshots.

## Reproduction (from the bug report)

Ledger with two commits where one entity's `ns:status` changes from
`"approved"` (t=1) to `"paid"` (t=2). 18 invoices were paid at t=1, 20 at
t=2.

| # | Pattern | t=1 (expected 18) | t=2 (expected 20) | Verdict |
|---|---------|-------------------|-------------------|---------|
| A | `?inv a ns:Invoice ; ns:status ?status` GROUP BY `?status` | **20** | 20 | ❌ latest |
| B | `... ; ns:status ?s . BIND(?s AS ?aliased)` GROUP BY `?aliased` | 18 | 20 | ✓ |
| C | No `?s a <Class>` triple | 28 | 30 | ✓ |
| D | `... ; ns:status ?s . FILTER(?s = "paid")` | 18 | 20 | ✓ |
| E | `?inv a ns:Invoice ; ns:status "paid"` | **20** | 20 | ❌ latest |

Patterns A and E were broken; B, C, D worked.

## Root cause

The query engine has three batched probe helpers in `fluree-db-query` that
walk base leaflet rows directly via `load_leaflet_columns_cached` and
`decode_leaf_dir_v3_with_base`:

- `batched_subject_probe_binary` — used by
  `flush_batched_exists_accumulator_binary` (pattern E:
  `?s p <literal>`).
- `scan_leaves_into_scatter` — used by
  `flush_batched_accumulator_binary` (pattern A:
  `?s p ?o` joined to a type triple).
- `batched_subject_star_spot` — used by `PropertyJoinOperator`'s
  same-subject star walk.

None of them apply `ctx.to_t` filtering or replay the FIR6 history
sidecar. The base index only stores the *live* state at `max_t`; retracts
go in the sidecar. So a historical query routed through any of these
helpers gets latest-state rows back.

Pattern D worked only because FILTER pushdown turned the literal match
into an inline op evaluated by per-row `BinaryScanOperator` scans, which
*do* honor `ctx.to_t` via `BinaryCursor::set_to_t`.

The `fast_path_*` operators are already gated on `to_t == max_t` via
`fast_path_store(ctx)`. The bug is specifically in the join-side batched
probes that bypassed that gate.

## Fix

Disable the batched probe paths when `to_t < store.max_t()` and let the
per-row scan path handle correctness. Two crates touched:

### `fluree-db-query/src/join.rs`

`NestedLoopJoinOperator::next_batch` — extend the runtime `use_batched`
gate with `at_latest_t` so all three batched modes
(`batched_eligible`, `batched_object_eligible`,
`batched_exists_eligible`) fall through to per-row scans for historical
snapshots.

### `fluree-db-query/src/property_join.rs`

- Add `PropertyJoinOperator::at_latest_t(ctx)` helper.
- Gate `can_batched_probe` (per-predicate batched probe) on it.
- Gate `can_spot_walk_remaining` (driver-side SPOT star walk) on it.

The fallback per-row path was already correct; this just routes
historical queries to it.

## Performance impact

Historical queries (`to_t < max_t`) now fall back to per-row scans for
the affected join shapes. They were silently incorrect before, so this
is a correctness regression on the latency dimension only when applied
to historical snapshots — the expected and now-tested behavior. Latest
queries (`to_t == max_t`, the common case) take the same fast path as
before.

### DIFF
--- a/fluree-db-api/tests/it_query_time_travel_bgp.rs
+++ b/fluree-db-api/tests/it_query_time_travel_bgp.rs
@@ -83,6 +83,47 @@ async fn seed_invoice_ledger(fluree: &fluree_db_api::Fluree) -> fluree_db_api::L
     ledger2
 }
 
+/// Seed a ledger that exercises the empty-after-retract leaflet case.
+///
+/// At t=1 every invoice has a `ns:legacyFlag "true"` triple. At t=2 the
+/// legacy flag is fully retracted from every invoice (no replacement). The
+/// indexer preserves the now-empty leaflet's metadata (`row_count == 0`
+/// but `history_*` populated) so time-travel can still recover the
+/// retracted state. Historical queries at t=1 must see the legacy flag;
+/// queries at t=2 must not.
+async fn seed_fully_retracted_ledger(
+    fluree: &fluree_db_api::Fluree,
+    ledger_id: &str,
+) -> fluree_db_api::LedgerState {
+    let ledger0 = genesis_ledger(fluree, ledger_id);
+
+    // t=1: 5 invoices, all carrying ns:legacyFlag "true" and a status.
+    let mut invoices = Vec::with_capacity(5);
+    for i in 0..5 {
+        invoices.push(json!({
+            "@id": format!("ns:Invoice/inv-{:02}", i),
+            "@type": "ns:Invoice",
+            "ns:status": "paid",
+            "ns:legacyFlag": "true",
+        }));
+    }
+    let tx1 = json!({"@context": ctx(), "@graph": invoices});
+    let _ = fluree.insert(ledger0, &tx1).await.expect("tx1");
+    support::rebuild_and_publish_index(fluree, ledger_id).await;
+    let l1 = fluree.ledger(ledger_id).await.expect("reload after t=1");
+
+    // t=2: retract every ns:legacyFlag triple — no replacement value.
+    let tx2 = json!({
+        "@context": ctx(),
+        "where": {"@id": "?inv", "ns:legacyFlag": "?flag"},
+        "delete": {"@id": "?inv", "ns:legacyFlag": "?flag"}
+    });
+    let l2 = fluree.update(l1, &tx2).await.expect("tx2").ledger;
+    support::rebuild_and_publish_index(fluree, ledger_id).await;
+    fluree.ledger(ledger_id).await.expect("reload after t=2");
+    l2
+}
+
 async fn run_count_sparql(fluree: &fluree_db_api::Fluree, sparql: &str) -> i64 {
     let jsonld = fluree
         .query_from()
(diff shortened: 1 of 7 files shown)


zonotope (approved):
I see a number of code duplication issues that are not blockers but should be addressed later. Otherwise 👍🏾

aaj3f (approved):
Not a big deal, but the PR description seems to describe what this PR did before the last few commits. The performance section mentions that "Historical queries… now fall back to per-row scans" bit, but that's no longer true after your last few commits. Similarly, the PR mentions abandoning the batch probes with the `replay_leaflet_at_t()` sidecar. I only suggest maybe fixing this so that if tools like claude ever try to understand the code vis-a-vis historical PRs landed into `main`, this might be a point of confusion.

In the "Bug Report Table" of the PR description, there's 2 patterns that seem maybe untested and possibly worth adding:

1. A test for `BIND(?s AS ?aliased)` (pattern B from the bug report) to lock in that the existing-correct path still works post-PR.
2. A test that exercises `batched_subject_star_spot` directly via a 3+ predicate same-subject star (the SPOT walk path) at a historical `t`. Pattern A's group-by goes through scatter, not star, I think. Just to make sure the star path doesn't have any subtle gaps.

---

(I would have put the following in as line comments/suggested-edits but GitHub seems to have service failures on those API calls so I'm going to shove them in here)

`fluree-db-query/src/join.rs`
Several places where `let to_t_u32 = u32::try_from(ctx.to_t).unwrap_or(u32::MAX);` gets recomputed when it's invariant across all the various leaflet loop work. Not a huge perf win, but given all the perf tuning for batched-probe hot paths.

`fluree-db-query/src/join.rs`
Several places within the per-leaf loops call `store.fetch_sidecar_bytes_sync(…)`. Those are all leaf-scoped, so it makes sense, but in environments like `solo` where these are network calls, I wonder if we could amortize it with additional caching on the leaflet cache? This wouldn't be a small change, so maybe worth a follow-up issue just to track it.

`fluree-db-query/src/join.rs:2149-2154`
This is more of a question than a recommendation. In this PR, `scan_leaves_into_scatter` switches from `ColumnSet::CORE` to `ColumnProjection::all()` when `needs_replay == true`. Why isn't this switch also needed for `batched_subject_probe_binary` and `batched_subject_star_spot`? Is it just that `scan_leaves_into_scatter` is the only one that narrowed projection so now it needs to widen it back, or...?

`fluree-db-binary-index/src/read/replay.rs:496-512`
Not a recommendation for now, but just wondering if it will be possible to have some bottlenecking w/ certain history queries as `find_base_assert_at_target()` paths to `replay_leaflet()` and does linear scan for first `t` match on each missing fact. We may just want to make sure we have telemetry/profiling for this if there's ever a future-facing scenario where particular queries with lots of asserts over time on the same keys end up being slower than we'd want them. I suppose the only solution for this would be a kind of `key --> most_recent_assert_t` index cache and it would only degrade, again, where we see repeat `FactKeyV3`

bplatz:
@aaj3f @zonotope addressed feedback in 1f34b969b281a2aae0c4939c086aeb9a326635e0 and 85407a5972bae5839e69f7ee33f86b235b94cea0

PR #1225, opened 2026-05-07
fix: restore host ownership in AUR publish docker step

bplatz:
The Arch container in contrib/aur/publish.sh chowns /work to an in-container `builder` uid (1000) so makepkg can run as non-root, but leaves the bind-mounted AUR clone owned by that uid on exit. On the runner (uid 1001), the next host-side `git add` then fails with `Permission denied` writing `.git/index.lock`, and the cleanup trap can't remove the work dir either — observed in the v4.0.3 publish-aur job. Fix: pass the host uid/gid into the container and chown /work back to them before the container exits.


zonotope (approved):
Sorry, I could have sworn that I had approved this when you opened it. LGTM 👍🏾

PR #1237, opened 2026-05-11
feat: close `--remote` coverage gaps + ledger archive + auth-scoping fixes

bplatz:
Based on #1236 

Brings remaining `--remote` commands to feature parity, lands the local + remote variants of `fluree export --format ledger` (the `.flpack` archive workflow), and fixes several auth-scoping / tracked-alias issues uncovered along the way. No protocol or wire-format changes — every endpoint these commands hit either already existed on the server or was added in earlier branches.

### `--remote` on previously-local-only commands

| Command | Endpoint | Auth bracket |
|---|---|---|
| `fluree drop <name> --remote <name>` | `POST /drop` | admin |
| `fluree log <ledger> --remote <name>` | `GET /log/*ledger` (new) | read |
| `fluree context get|set --remote <name>` | `GET/PUT /context/*ledger` | read / write |
| `fluree history --remote <name>` | `POST /query/*ledger` (ledger-scoped) | read |
| `fluree create <ledger> --remote <name>` | `POST /create` | admin |
| `fluree export <ledger> --remote <name>` (RDF) | `POST /export/*ledger` (new) | admin |

All follow the same three-mode dispatch the existing `--remote` commands use: explicit `--remote <name>` → named remote, otherwise auto-route through a locally running `fluree server start` if `server.meta.json` is present, otherwise direct local execution. `--direct` skips auto-routing.

### Ledger-archive export (`.flpack`)

`docs/cli/server-integration.md` claimed `fluree export --format ledger -o mydb.flpack` worked. The export side was never wired up; only `fluree create --from <file>.flpack` (import) actually existed. This branch closes that:

- **Local**: `fluree export mydb --format ledger -o mydb.flpack` calls a new `Fluree::archive_ledger` API that streams pack frames through any `AsyncWrite` and appends a `phase: "nameservice"` manifest frame so `fluree create --from <file>.flpack` can reconstruct head pointers. `--no-indexes` for a smaller archive that reindexes on import. TTY stdout is refused.
- **Remote**: `fluree export mydb --remote origin --format ledger -o mydb.flpack` fetches the remote `NsRecord`, streams `POST /pack/*ledger`, and swaps the terminal End frame for the synthesized nameservice manifest on the fly. Resulting `.flpack` is byte-compatible with a locally-generated archive — `fluree create --from` doesn't care which side produced it.

### `fluree query --remote --at <t>` (time-travel) auth fix

The previous remote time-travel path posted to the connection-level `/query` endpoint, where the server's `can_read` check used `body.from` for the ledger id. With `from: "mydb:main@t:5"`, scoped read tokens (`fluree.ledger.read.mydb:main`) failed because they don't match the time-suffixed form. This branch routes all time-travel cases through the ledger-scoped `POST /query/{ledger}` instead — path drives auth, body's `from` / SPARQL `FROM` drives time-travel resolution.

Applies to JSON-LD `--at`, SPARQL `--at`, JSON-LD `--at --explain`, and SPARQL `--at --explain` (the last two via the server fix from the prerequisite PR).

## Notable engineering choices

- **Tracked-alias resolution for `--remote`.** `fluree-db-cli/src/context.rs::build_remote_mode` now canonicalizes the ledger alias via `to_ledger_id` (so `mydb` → `mydb:main` on the URL path). Without this, scoped tokens 404 because the path differs from the auth identifier. The `--format ledger` remote archive specifically also looks up the tracked-config store: when `<alias>` is tracked at the same remote, it archives the upstream's `tracked.remote_alias` rather than the local alias literally.

- **Archive splicer error handling.** `RemoteLedgerClient::archive_ledger_to_writer` (extracted as `splice_archive_stream` for testability) is careful to distinguish `PackError::Incomplete(_)` (need more bytes) from fatal pack-decoder errors. Previously every decoder error was swallowed as "need more", so a corrupt FPK1 magic or oversize payload would buffer until EOF and surface as "ended before End frame". Five unit tests cover End-frame substitution, chunk-boundary splits inside frames, manifest field selection on `--no-indexes`, server `Error` frame propagation, and prompt rejection of bad magic.

- **`create --remote` doesn't require a project `.fluree/`.** Falls back to global config (`$FLUREE_HOME` or platform default) for remote registration lookups, so the command works from any directory. The local-only `fluree create` still requires a project `.fluree/` so new ledgers land in a discoverable place.

- **`fluree drop --remote` active-ledger handling.** Explicit `--remote <name>` never touches local state. Auto-route (no `--remote`, server detected) operates against the same on-disk storage as `--direct`, so it also clears the local active-ledger pointer when it matched the dropped name.

- **History prefix expansion is client-side.** `fluree history --remote ex:alice ...` expands the compact IRI against the project's stored prefix map before the request leaves the CLI, so the server never has to consult the local prefix table. The body still ships its `@context` (derived from local prefixes) for response display.

## Limitations / known caveats

- **`fluree export --format ledger --remote` requires storage permissions.** Both `/storage/ns/:ledger-id` and `/pack/*ledger` sit in the replication-grade bracket (`fluree.storage.*`), same auth as `fluree clone` / `pull`. Without those permissions the server returns 404 to avoid existence leaks.
- **`fluree export --remote` (RDF) is admin-protected.** RDF export reads from the binary index without per-flake policy filtering, so it lives alongside `/create`, `/drop`, `/reindex` rather than the data-read bracket. Adding policy-filtered streaming export would let it move to read-auth in the future.
- **`fluree export --format ledger` doesn't support `--at` / `--all-graphs` / `--graph` / `--context*`.** Archives capture the current head; the other flags apply only to RDF formats.

## Tests

- `cargo test -p fluree-db-cli`: **41 lib + 73 integration** tests pass. The 5 new lib tests cover the `.flpack` remote archive splicer (End substitution, chunk boundaries, index-field gating, error-frame propagation, magic validation).
- `cargo test -p fluree-db-server --test integration explain`: 4 explain tests pass (from the prerequisite PR).
- `cargo clippy -p fluree-db-cli --all-features --all-targets -- -D warnings`: clean.

## Docs

`docs/cli/server-integration.md` is the canonical reference for implementers building custom servers against the CLI. This PR adds/updates:

- `### fluree create <ledger> --remote <name>` section
- `### fluree context get|set --remote` section
- `### fluree history --remote` section (with note about ledger-scoped routing for token auth)
- `### fluree log --remote` section
- `### fluree export --remote` section (RDF)
- `### fluree export --format ledger` section (local + remote modes)
- "Remote time travel (`--at`)" callout under the data-API section
- "Remote `--at --explain`" callout following the time-travel note
- `## Commit Log Contract` (full schema + required server semantics)
- `## RDF Export Contract` (request body, content-types, error responses)
- Expanded "Data API" intro list, terminology cleanup on the history section, more precise active-ledger pointer rules under `fluree drop --remote`
- Validation script gains `log`, `export` (RDF + ledger), `context get/set`, `history`, `query --at`, `query --at --explain`, `create --remote`, `drop --remote` lines

`docs/operations/pack-archive-restore.md` is updated to drop the old "no dedicated CLI command" stub and document the local + remote archive flows.

`docs/api/endpoints.md` gains `GET /log/*ledger` and `POST /export/*ledger` entries.

### DIFF
--- a/fluree-db-api/src/lib.rs
+++ b/fluree-db-api/src/lib.rs
@@ -3244,6 +3244,43 @@ impl Fluree {
         export_builder::ExportBuilder::new(self, ledger_id.to_string())
     }
 
+    /// Walk the commit chain for a ledger and return per-commit summaries.
+    ///
+    /// `limit` caps the number of returned summaries (newest-first by `t`).
+    /// The returned `total` reflects the full chain length regardless of cap;
+    /// truncation is implied by `summaries.len() < total`.
+    ///
+    /// Uses a branch-aware content store so the walk crosses fork points —
+    /// pre-fork commits live under the source branch's namespace, not the
+    /// current branch's.
+    pub async fn commit_log(
+        &self,
+        ledger_id: &str,
+        limit: Option<usize>,
+    ) -> Result<(Vec<CommitSummary>, usize)> {
+        let record = self
+            .nameservice()
+            .lookup(ledger_id)
+            .await?
+            .ok_or_else(|| ApiError::NotFound(ledger_id.to_string()))?;
+
+        let head = match record.commit_head_id.as_ref() {
+            Some(id) => id.clone(),
+            None => return Ok((Vec::new(), 0)),
+        };
+
+        let store = fluree_db_nameservice::branched_content_store_for_record(
+            self.backend(),
+            self.nameservice(),
+            &record,
+        )
+        .await?;
+
+        let (summaries, total) =
+            fluree_db_core::walk_commit_summaries(&store, &head, 0, limit).await?;
+        Ok((summaries, total))
+    }
+
     /// Get the default JSON-LD context for a ledger.
     ///
     /// Reads the context CID from nameservice config and fetches the blob
--- a/fluree-db-api/src/lib.rs
+++ b/fluree-db-api/src/lib.rs
@@ -3244,6 +3244,109 @@ impl Fluree {
         export_builder::ExportBuilder::new(self, ledger_id.to_string())
     }
 
+    /// Stream a self-contained ledger archive (`.flpack`) for `ledger_id`.
+    ///
+    /// This is the export side of the `fluree create --from <file>.flpack`
+    /// pipeline. Frame bytes (header → commits → optional indexes →
+    /// nameservice manifest → end) are written to `writer` in order, so the
+    /// caller can target a file, stdout, or any `AsyncWrite` sink without
+    /// buffering the full archive in memory.
+    ///
+    /// `include_indexes` controls whether binary index artifacts ride along
+    /// (`true` → instantly queryable on import; `false` → smaller archive,
+    /// import will need to reindex). When the ledger has no index root, the
+    /// flag is silently downgraded to commits-only.
+    pub async fn archive_ledger<W: tokio::io::AsyncWrite + Unpin + Send>(
+        &self,
+        ledger_id: &str,
+        include_indexes: bool,
+        writer: &mut W,
+    ) -> Result<pack::PackStreamResult> {
+        use tokio::io::AsyncWriteExt as _;
+
+        let record = self
+            .nameservice()
+            .lookup(ledger_id)
+            .await?
+            .ok_or_else(|| ApiError::NotFound(ledger_id.to_string()))?;
+
+        let canonical_id = record.ledger_id.clone();
+        let handle = self.ledger_cached(&canonical_id).await?;
+
+        // Source the manifest *and* the pack request from the same view, so
+        // the archive bytes and the manifest pointers always agree. Reading
+        // the heads from the nameservice record while reading the pack
+        // contents from the cached handle could disagree if the cache is
+        // behind a freshly committed head.
+        let view = handle.snapshot().await;
+
+        let head_commit_id = view.head_commit_id.clone().ok_or_else(|| {
+            ApiError::internal(format!("ledger {canonical_id} has no head commit to pack"))
+        })?;
+
+        // `full_ledger_pack_request` silently drops the index when the
+        // ledger has none. Mirror that decision here so we never advertise
+        // an `index_head_id` we did not archive.
+        let archived_index = if include_indexes {
+            view.head_index_id.clone()
+        } else {
+            None
+        };
+        let request = match archived_index.clone() {
+            Some(index_root) => pack::PackRequest::with_indexes(
+                vec![head_commit_id.clone()],
+                vec![],
+                index_root,
+                None,
+            ),
+            None => pack::PackRequest::commits(vec![head_commit_id.clone()], vec![]),
+        };
+
+        let mut manifest = serde_json::json!({
+            "phase": "nameservice",
+            "ledger_id": canonical_id,
+            "name": record.name,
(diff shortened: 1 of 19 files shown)


fluree-db-api/src/lib.rs
```diff
@@ -3244,6 +3244,146 @@ impl Fluree {
...
+                    .map_err(|e| ApiError::internal(format!("archive write: {e}")))?;
+            }
+            writer
+                .flush()
+                .await
+                .map_err(|e| ApiError::internal(format!("archive flush: {e}")))?;
+            Ok::<_, ApiError>(())
+        };
+
+        let (producer_result, consumer_result) = tokio::join!(producer, consumer);
```
zonotope, line 3344: This will hang if the consumer errors because the producer will be blocked until the channel closes, and the channel doesn't close on error, and `join!` waits for all branches regardless of error so it will just keep polling the producer since it technically never closes.

The consumer has only borrowed the receiver here, so it lives on after the consumer errors. If you switch to `async move` on 3325 then the consumer will fully own `rx` and it will be dropped when the consumer fails, the channel will close, and the producer will be unblocked.

fluree-db-cli/src/commands/log.rs
```diff
@@ -8,6 +9,110 @@ pub async fn run(
     oneline: bool,
     count: Option<usize>,
     dirs: &FlureeDir,
+    remote_flag: Option<&str>,
+    direct: bool,
+) -> CliResult<()> {
+    if let Some(remote_name) = remote_flag {
+        let alias = context::resolve_ledger(ledger, dirs)?;
+        let client = context::build_remote_client(remote_name, dirs).await?;
+        let result = run_remote(&alias, oneline, count, &client).await;
```
zonotope: This is using the un-normalized alias, so it will try to read from `.../log/mydb/` instead of `.../log/mydb:main` or `.../log/mydb/main`. is that intended?

fluree-db-server/src/routes/export.rs
```diff
@@ -0,0 +1,167 @@
...
+        }
+        if let Some(at_str) = req.at.as_deref() {
+            builder = builder.as_of(parse_time_spec(at_str));
+        }
+        if let Some(ctx) = req.context.as_ref() {
+            builder = builder.context(ctx);
+        }
+
+        let mut buf: Vec<u8> = Vec::new();
+        let stats = builder.write_to(&mut buf).await.map_err(ServerError::Api)?;
```
zonotope, line 116: This seems like it's writing the entire export to memory before copying it out to the response.  I guess it's fine for now, but I think we should look for a way to stream it if possible for large ledgers.

zonotope (approved):
Lgtm besides what I called out.

fluree-db-api/src/lib.rs
bplatz: Addressed in 36873872b.

fluree-db-cli/src/commands/log.rs
bplatz: Addressed in 11fa01694.

fluree-db-server/src/routes/export.rs
bplatz: @zonotope Noted and agree, leaving out for now.

PR #1238, opened 2026-05-11
fix(query): share EXISTS strategy with OPTIONAL+!bound rewrite

bplatz:
- Fix the `OPTIONAL { ...binds ?v... } FILTER(!bound(?v))` planner, which unconditionally chose per-row `ExistsOperator` and timed out on large outer streams. Both the pattern-level `Pattern::Exists` / `Pattern::NotExists` dispatch and this rewrite now share one strategy chooser, so the rewrite picks `SemijoinOperator` (build-once + hash probe) when the inner shares produced key vars with the outer schema.
- Append an actionable hint to the "Unknown function" parse error covering common query attempts (`!bound`, `nil?`, `not-bound`, …) without misleading users who try bare `!` for plain logical negation.
- Document `not-exists`, `(not (bound ?v))`, and the SPARQL → JSON-LD translation in `docs/query/jsonld-query.md`; reconcile the conditional-updates doc.

## Why

The rewrite at `fluree-db-query/src/execute/where_plan.rs` recognised the shape but unconditionally dispatched it through `ExistsOperator`, citing a "outer is selective, inner may be large" heuristic that doesn't hold when the outer is a full class scan. `ExistsOperator::has_match` rebuilds and re-executes the inner operator tree per outer row, so a n-row outer drives n subquery plan-and-executes.

The pattern-level `Pattern::NotExists` dispatch already had the right logic — pick `SemijoinOperator` when the inner is correlated only via produced vars. Two callers, one decision, one helper.

### Code

- **`fluree-db-query/src/execute/where_plan.rs`** — introduce `ExistsStrategy { Semijoin { key_vars } | Exists { reason } }` and `choose_exists_strategy(outer_schema, inner_patterns) -> ExistsStrategy`. The chooser is pure (no operator construction, no stats, no I/O), unit-testable in isolation. `build_exists_strategy` is the thin wrapper that calls it, emits `tracing::debug!(strategy, …, "exists dispatch")`, and constructs the operator. The pattern-level `Pattern::Exists` / `Pattern::NotExists` dispatch and the OPTIONAL+!bound rewrite both call it.

- **`fluree-db-query/src/parse/lower.rs`** — `unknown_function_hint(name)` appended to the "Unknown function" parse error. Bound-test hint for `!bound`, `nil?`, `not-bound`, `notbound`, `unbound`, `missing?`. Generic "spelled `not`" hint for bare `!` and any other `!`-prefixed token (so `(! (= ?age 18))` doesn't get pushed toward bound-testing).

### Tests

- Five new unit tests in `where_plan.rs` pinning the strategy decision:
  - `choose_exists_strategy_semijoin_for_root_level_absence_shape` — regression guard for the OPTIONAL+!bound rewrite
  - `choose_exists_strategy_exists_when_uncorrelated`
  - `choose_exists_strategy_exists_when_outer_only_consumed`
  - `choose_exists_strategy_exists_when_outer_schema_empty`
  - `choose_exists_strategy_semijoin_key_vars_follow_outer_schema_order`
- Two new integration tests in `it_query_negation.rs` asserting `OPTIONAL + (not (bound ?v))` returns the same rows as the equivalent `["not-exists", ...]`, in both s-expression and `["not", ["bound", "?v"]]` data filter forms.
- Four hint unit tests in `lower.rs`.

### Docs

- **`docs/query/jsonld-query.md`** — new "Negation Patterns" section: `not-exists` (canonical), `OPTIONAL + (not (bound ?v))`, `minus`, supported-vs-unsupported filter idiom table, SPARQL → JSON-LD translation table. The perf note states the precise strategy condition (inner-produced shared vars + no outer-only consumed vars) rather than the broader "shares a variable" wording.
- **`docs/transactions/conditional-updates.md`** — reconciled with the new query docs. Best-practice guidance now describes both supported absence idioms and when to pick each.

## Behaviour

- Correctness is unchanged for any query: `ExistsOperator` and `SemijoinOperator` produce the same solutions for every input the chooser distinguishes. The fix is purely about which one runs.
- The `OPTIONAL + (not (bound ?v))` shape now plans the same as `["not-exists", { ... }]` whenever the inner shares an inner-produced var with the outer and references no outer-only var.
- Failure cases that legitimately need per-row evaluation (uncorrelated inner, or inner referencing outer-only consumed vars via a filter) still get `ExistsOperator` — confirmed by `choose_exists_strategy_exists_when_outer_only_consumed`.

### DIFF
--- a/docs/query/jsonld-query.md
+++ b/docs/query/jsonld-query.md
@@ -401,6 +401,93 @@ Match data from multiple alternative patterns:
 }
 ```
 
+### Negation Patterns
+
+Restrict results to subjects that *do not* match a given pattern. Two forms
+are available; **prefer `not-exists` for the common "subjects without
+property `p`" case.**
+
+#### `not-exists` (canonical)
+
+`["not-exists", { ... }]` keeps rows for which the inner pattern has **no**
+solution under the current bindings:
+
+```json
+{
+  "where": [
+    { "@id": "?file", "@type": "ex:File" },
+    ["not-exists", { "@id": "?file", "ex:parent": "?p" }]
+  ],
+  "select": "?file"
+}
+```
+
+When the inner pattern is correlated to the outer **only via vars the inner
+itself produces** (here `?file`, produced by both the outer triple and the
+inner triple), and the inner does not reference any outer-only variable
+(e.g. via a `["filter", ...]` mentioning an outer var the inner doesn't
+produce), the planner builds the inner solution set **once** and probes it
+per outer row — the performant path on large outer streams. If either
+condition fails, the inner is rebuilt and re-run per outer row (still
+correct, but linear in the outer cardinality).
+
+`["exists", { ... }]` is the affirmative dual and uses the same dispatch.
+
+#### `OPTIONAL` + `(not (bound ?v))`
+
+The SPARQL-muscle-memory form `OPTIONAL { ... } FILTER (!bound(?v))` is
+spelled in JSON-LD as an `optional` followed by a filter using the supported
+`bound` function:
+
+```json
+{
+  "where": [
+    { "@id": "?file", "@type": "ex:File" },
+    ["optional", { "@id": "?file", "ex:parent": "?p" }],
+    ["filter", "(not (bound ?p))"]
+  ],
+  "select": "?file"
+}
+```
+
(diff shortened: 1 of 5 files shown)


zonotope (approved):
🪭

PR #1239, opened 2026-05-13
perf: Improve novelty and property-path query performance

bplatz:
- Resolve novelty-only subjects during binary scans so recent writes can use constrained lookups.
- Enable property joins for exact-threshold star patterns, improving vector similarity query plans.
- Short-circuit bound property-path reachability checks once the target is found.

- Manual benchmark: vector query improved from ~36s to ~720ms warm; bounded inverse `^storage:parent+` folder query runs in ~750-860ms warm versus forward `storage:parent+` timing out through CloudFront.

### DIFF
--- a/fluree-db-query/src/binary_scan.rs
+++ b/fluree-db-query/src/binary_scan.rs
@@ -985,6 +985,16 @@ impl BinaryScanOperator {
         })
     }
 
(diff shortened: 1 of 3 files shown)


zonotope (approved):
🏯

PR #1240, opened 2026-05-13
Fix/import issue

bplatz:
Fixes incorrect datatype labels in per-class stats (`stats.classes[*].properties[*].types`).

The class stats collector stores datatype buckets as `ValueTypeTag` values, but the stats builder was treating those values as `DatatypeDictId` indices. This caused initial imports to report literal property types as `UNKNOWN`, and rebuilds to report deterministic but wrong labels like `@id`, `xsd:boolean`, or `rdf:langString`.

This PR keeps the class stats path consistently in the `ValueTypeTag` domain and removes the now-misleading `dt_tags` lookup. It also adds missing unambiguous `OType` mappings for `xsd:gYearMonth` and `xsd:gMonthDay`, while intentionally leaving ambiguous carriers like `NUM_BIG_OVERFLOW` as `UNKNOWN` until they can be disambiguated with arena access or preserved semantic datatype tags.

## Test Plan

- Added regression coverage for per-class datatype stats across import/reindex paths.
- Verified literal properties report the expected `ValueTypeTag` values, including string, integer, boolean, dateTime, date, gYearMonth, and `@id` references.

### DIFF
--- a/fluree-db-api/src/import.rs
+++ b/fluree-db-api/src/import.rs
@@ -3911,7 +3911,6 @@ where
                 let mut per_graph_classes = fluree_db_indexer::stats::build_class_stat_entries(
                     cs,
                     &predicate_sids_v6,
-                    &[],
                     &uploaded_dicts.language_tags,
                     input.run_dir,
                     input.namespace_codes,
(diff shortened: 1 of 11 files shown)


zonotope (approved):
🧁

PR #1241, opened 2026-05-14
fix: report tracked transaction fuel

bplatz:
## Summary
- Exposes transaction tracking tallies (`time`, `fuel`, `policy`) in HTTP transaction response bodies when tracking is enabled.
- Wires header-driven tracking through SPARQL UPDATE and Turtle/TriG transaction paths.
- Ensures direct Turtle insert staging receives the tracker so transaction fuel is counted consistently.


zonotope (approved):
💾

PR #1243, opened 2026-05-15
fix: Make hard ledger drop fully clean up on S3 + DynamoDB

bplatz:
Hard-dropping a ledger on the AWS stack (S3 storage + DynamoDB nameservice) was silently behaving like a soft drop: the alias remained unreusable in DynamoDB and 0 artifacts were deleted in S3. This PR fixes four cooperating bugs so `drop_ledger(.., DropMode::Hard)` actually clears the ledger end-to-end.

### 1. `DynamoDbNameService::purge` was missing — fell back to `retract`

`Publisher::purge` has a default impl that delegates to `retract()`, which only flips `meta.retracted = true`. `DynamoDbNameService` never overrode it, so `head` / `index` / `config` / `status` rows were left behind and `publish_ledger_init` (which uses `attribute_not_exists(pk)` per row) couldn't recreate the alias.

**Fix:** Override `purge()` to delete every row under `pk`. Factored a `delete_all_rows_for_pk` helper that's also reused by `drop_branch`. The helper always sweeps non-meta rows regardless of the conditional-meta-delete outcome, so a purge that crashed mid-sweep is recovered on the next call.

### 2. `S3Storage::list_prefix*` didn't parse fluree addresses

Reads, writes, and deletes route through `address_to_key` (which strips `fluree:s3://`). The list APIs didn't, so a caller passing `fluree:s3://ledger/main/` ended up querying S3 with `fluree-data/fluree:s3://ledger/main/` — a literal substring that never matches real keys. Result: list returned 0 → fast-path drop deleted nothing.

**Fix:** New `build_list_prefix` helper that runs the same `parse_fluree_address` path. Applied to all four list entry points (`StorageRead::list_prefix`, `StorageList::list_prefix`, `list_prefix_paginated`, `list_prefix_with_metadata`). Six unit tests covering scheme stripping, trailing-slash preservation, empty input, and bucket-prefix trimming.

### 3. `TieredStorage` routing missed entire tiers on a ledger-root list

`TieredStorage::list_prefix` routes by substring: `/commit/` or `/txn/` → commit tier; everything else → index tier. A ledger-root prefix like `fluree:s3://ledger/main/` contains neither, so a single root-level list misses the whole commit bucket in split commit/index deployments. The original `drop_artifacts` also missed `{prefix}/config/` (where `ContentKind::LedgerConfig` blobs — config + default-context — are written).

**Fix:** `drop_artifacts` now enumerates per-subprefix: `commit/`, `txn/`, `index/`, `config/`, and conditionally `{ledger}/@shared/dicts/`. Each list call routes to the correct tier.

### 4. Partial list failures were silently treated as success

The original "if any subprefix listed OK, declare success" branch could purge the nameservice while leaving an entire artifact class behind if one tier returned an error.

**Fix:** Failures are collected as warnings on the `DropReport`. When any subprefix errors, a CID-walk cleanup pass runs after the fast path (`release` is idempotent on already-deleted CIDs). The nameservice is only purged once `drop_artifacts` has had its best shot.

## Branch safety

- Cross-branch `@shared/dicts/` is only wiped when `is_last_live_branch` returns true (uses `NameService::list_branches` to confirm no sibling branches remain). Branch-scoped `drop_branch` / `purge_branch` always pass `false`.
- `drop_ledger`'s doc-comment now calls out that hard-dropping a parent branch while child branches are still alive will delete commit/index artifacts the children may resolve through the branched content store. Drop child branches first.

### DIFF
--- a/docs/api/endpoints.md
+++ b/docs/api/endpoints.md
@@ -1203,7 +1203,7 @@ curl -X POST http://localhost:8090/v1/fluree/create \
 
 ### POST /drop
 
-Drop (delete) a ledger.
+Drop a ledger or graph source.
 
 **URL:**
 ```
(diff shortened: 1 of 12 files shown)


zonotope (approved):
🛅

PR #1252, opened 2026-05-23
fix: history query with non-matching `@id` returns all genesis facts

bplatz:
History queries with a literal `@id` that doesn't match any subject silently
dropped the subject filter and returned every base row at `t:1`. On larger
ledgers this blew past Lambda's 6 MB response limit and surfaced as a generic
upstream 502; on smaller ledgers it returned plausible-looking garbage.

## Root cause

`BinaryHistoryScanOperator::collect_history_flakes` builds a `BinaryFilter` from
the pattern. When the subject literal doesn't exist in the persisted store,
`store.find_subject_id(iri)` returns `None`, so `filter.s_id = None`. The
persisted pass only constrains by `s_id`/`p_id` (object filtering isn't applied
there), so an unconstrained filter walked every leaflet and emitted every base
row whose `t` fell in range.

Novelty was unaffected — it filters via `flake_matches_range_eq` keyed on the
snapshot-space `s_sid`, which correctly short-circuits to empty.

## Fix

Before opening the persisted branch, detect literal-bound pattern components
(subject `Ref::Iri`/`Ref::Sid`, same for predicate) that failed store
resolution and skip the persisted pass entirely. Novelty still runs so subjects
present only in unindexed data continue to match.

### DIFF
--- a/fluree-db-api/tests/it_query_history_range.rs
+++ b/fluree-db-api/tests/it_query_history_range.rs
@@ -680,3 +680,112 @@ async fn history_range_iri_object_novelty_only() {
         "novelty-only history over a ref-valued predicate must bind @t and @op"
     );
 }
+
+// ---------------------------------------------------------------------------
+// Case: literal `@id` that does not resolve to any subject in the data.
+//
+// Bug report: a history query with `{"@id": "<non-existent IRI>", ...}` was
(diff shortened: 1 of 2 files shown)


zonotope (approved):
⛵

PR #1253, opened 2026-05-26
fix: consistent fuel reporting across import, query, server, and CLI

bplatz:
Before this branch, fuel/time/policy reporting was inconsistent across the surfaces a user actually touches:

- **Bulk imports** silently skipped fuel accounting entirely — a 100M-flake import returned `flake_count` but no `TrackingTally`, and `max-fuel` caps had no effect.
- **`stage.rs`** filtered out "schema flakes" from the per-flake charge using a misleading "mirrors query-side fuel counting" comment — query-side actually charges per flake unconditionally. The filter under-charged schema-heavy transactions and added a cross-crate hot-path dependency on `fluree-db-policy` with no documented rationale.
- **Connection-scoped SPARQL** on `POST /v1/fluree/query` ignored `fluree-track-*` / `fluree-max-fuel` headers and returned no tally, while JSON-LD on the same endpoint honored body opts. The AgentJson SPARQL variant on the same route had the same gap.
- **`FromQueryBuilder`** (`Fluree::query_from()`) had no `.tracking()` / `.track_all()` setters, so SDK callers couldn't request tracking from the connection-scoped builder. The four `query_connection_jsonld_tracked*` helpers also hard-coded `None` for tracking overrides.
- **`fluree query`** CLI had no flag to report fuel for either language, against any backend.

Net: fuel only worked correctly on the ledger-scoped SDK paths and the ledger-scoped HTTP route. Everywhere else it was silently absent or inconsistently shaped.

### Engine (`fluree-db-transact`, `fluree-db-api`, `fluree-db-core`)

- **`stage.rs`** now charges 1 micro-fuel per staged flake unconditionally — dropped the `is_schema_flake` filter (and its premise), collapsed the per-flake loop into a single `consume_fuel(flakes.len())`. Comment updated to reflect reality.
- **Bulk import** (`fluree-db-api/src/import.rs`) accepts an optional `Tracker` via the new `ImportBuilder::tracker()` setter. After each `finalize_parsed_chunk`, the pipeline charges `100_000 + flake_count` micro-fuel — same formula as the per-transaction stage path — across all three commit loops (parallel `commit_parsed_chunks_in_order`, remote-serial, local-serial fallback). `ImportResult` gained `tally: Option<TrackingTally>` and `ImportError::FuelExceeded`.
- **`PolicyStats`** now derives `Deserialize` so the CLI can round-trip the policy tally from JSON responses.

### Server (`fluree-db-server`)

- Connection-scoped SPARQL on `POST /v1/fluree/query` now checks `headers.has_tracking()`, builds `TrackingOptions` from headers, calls `query_connection_sparql_tracked` with the override, and returns a `TrackedQueryResponse` body plus `x-fdb-fuel` / `x-fdb-time` / `x-fdb-policy` headers — mirroring the JSON-LD pattern already on that route.
- The agent-json branch of the same route does the same, but preserves the bare agent-json envelope as the response body and surfaces the tally only via the `x-fdb-*` headers so LLM agents see the shape they expect.

### SDK builder (`fluree-db-api`)

- Threaded `tracking_override: Option<TrackingOptions>` through the four `query_connection_sparql_tracked*` AND the four `query_connection_jsonld_tracked*` helpers (forwarded to `query_tracked` / `query_dataset_tracked`, which already honored it).
- Added `.tracking(opts)` and `.track_all()` setters to `FromQueryBuilder`; `execute_tracked` now pulls `core.tracking` and forwards it to both JSON-LD and SPARQL helpers instead of hard-coding `None`.

### CLI (`fluree-db-cli`)

- New `fluree query` flags that map 1:1 to the HTTP header vocabulary:
  - `--track` — fuel + time + policy (shorthand)
  - `--track-fuel`, `--track-time`, `--track-policy` — granular, additive
  - `--max-fuel <N>` — abort at N decimal fuel; implies `--track-fuel`
- Footer prints only the metrics actually requested (no noise when only `--track-fuel`).
- **Remote mode wiring**: the same flags now work against an explicit `--remote`, the auto-routed local server, and any custom HTTP server speaking the same contract. The CLI sends `fluree-track-*` / `fluree-max-fuel` request headers via the new `_with_headers` variants on `RemoteLedgerClient` and parses the `TrackedQueryResponse` envelope back into a (result, tally) pair.
- Mutually exclusive with `--bench`, `--explain`, `--format tsv/csv` (those would bypass the tracked path).

### Docs

- New "Tracking Contract" section in `docs/cli/server-integration.md` documenting the request headers, response envelope, x-fdb-* response headers, and required server behavior — so a third-party HTTP server can support the CLI's tracking flags by implementing one header family + one response shape.
- `docs/query/tracking-and-fuel.md` updated to reflect the new per-flake (no schema filter) and per-import-chunk (100 fuel baseline) accounting.

### DIFF
--- a/fluree-db-api/src/import.rs
+++ b/fluree-db-api/src/import.rs
@@ -31,7 +31,10 @@
 //! even though chunk parsing is parallel.
 
 use crate::error::ApiError;
-use fluree_db_core::{ContentId, ContentKind, ContentStore, RemoteObject, Storage, StorageRead};
+use fluree_db_core::{
+    ContentId, ContentKind, ContentStore, FuelExceededError, RemoteObject, Storage, StorageRead,
+    Tracker, TrackingTally,
+};
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
@@ -144,6 +147,12 @@ pub struct ImportConfig {
     pub leaf_target_rows: usize,
     /// Optional progress callback invoked at key pipeline milestones.
     pub progress: Option<ProgressFn>,
+    /// Optional execution tracker. When enabled, fuel is charged per chunk
+    /// using the same formula as `stage_transaction`: 100 fuel baseline per
+    /// commit plus 1 micro-fuel per encoded flake. If the tracker carries a
+    /// `max_fuel` limit, the import aborts with `ImportError::FuelExceeded`
+    /// as soon as the limit is exceeded. Default: `Tracker::disabled()`.
+    pub tracker: Tracker,
 }
 
 impl std::fmt::Debug for ImportConfig {
@@ -154,6 +163,7 @@ impl std::fmt::Debug for ImportConfig {
             .field("chunk_size_mb", &self.chunk_size_mb)
             .field("chunk_max_flakes", &self.chunk_max_flakes)
             .field("progress", &self.progress.as_ref().map(|_| "..."))
+            .field("tracker_enabled", &self.tracker.is_enabled())
             .finish_non_exhaustive()
     }
 }
@@ -179,6 +189,7 @@ impl Default for ImportConfig {
             leaflets_per_leaf: 10,
             leaf_target_rows: 250_000,
             progress: None,
+            tracker: Tracker::disabled(),
         }
     }
 }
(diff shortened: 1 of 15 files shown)


fluree-db-api/src/import.rs
```diff
@@ -1781,6 +1820,13 @@ where
                     .await
                     .map_err(|e| ImportError::Transact(e.to_string()))?;
 
+                // Fuel: 100 fuel baseline per commit + 1 micro-fuel per flake,
+                // matching the per-chunk-as-transaction model in `stage.rs`.
+                if env.config.tracker.is_enabled() {
+                    env.config.tracker.consume_fuel(100_000)?;
+                    env.config.tracker.consume_fuel(result.flake_count as u64)?;
+                }
```
zonotope: This could be a helper.

zonotope (approved):
🍓

fluree-db-api/src/import.rs
bplatz: Addressed in d36bf85fe931ec99cb713710d00e28adcd37cc28

PR #1254, opened 2026-05-26
Consensus Abstraction

zonotope:
## Summary

Introduces a new `fluree-db-consensus` crate that defines a `Committer` trait for ledger-mutating operations, plus a `MonolithicConsensus` implementation that wraps the existing local Fluree pipeline with idempotency tracking, admission control, and write-lock serialization. All five HTTP entry points that mutate the commit graph (transact, revert, merge, rebase, push) now flow through this abstraction instead of calling `Fluree::*` methods directly. This is the foundation for future replicated consensus implementations (Raft, BFT) to plug in behind the same trait without touching the HTTP layer or each operation's business logic.

### New crate: `fluree-db-consensus`

- Defines the `Committer` trait with one method per operation kind — `transact`, `revert`, `merge`, `rebase`, `push`. Each method takes a typed request struct and returns a typed receipt; implementations multiplex internally onto whatever ordering/replication mechanism they use.
- Defines `SubmissionLookup` for after-the-fact status lookup of submissions identified by a caller-supplied `IdempotencyKey` (typically a ULID).
- Provides `MonolithicConsensus`, a single-process implementation backed by a per-ledger `tokio::sync::Mutex` and the existing local pipeline.

### `MonolithicConsensus` behaviour

- **Per-ledger FIFO ordering** via the existing `LedgerManager` write lock. Concurrent operations on the same `ledger:branch` serialize; different ledgers proceed in parallel.
- **Idempotency cache** (moka TTL) — submissions with the same `IdempotencyKey` collapse onto a single execution. Configurable TTL (default 1 hour) and a hard 100 000-entry capacity so sustained keyed traffic can't grow the cache without bound between evictions.
- **Body-hash-tagged keys.** Each op kind contributes its variant tag to the body hash, so cross-op key reuse always reports as `KeyCollision` from the cache claim rather than silently returning the wrong typed receipt.
- **Pending-operation cap** via `tokio::sync::Semaphore`. Submissions arriving while the cap is reached are refused immediately with `SubmissionError::Overloaded` (mapped to HTTP 503), preventing the in-flight request queue from growing without bound under sustained load. Default 1024 permits, overridable via `with_pending_limit`.
- **Error categories preserved.** `ApiError::status_code()` flows through `SubmissionError::Execution { status, message }` so client errors (404, 422, etc.) don't get collapsed to 500.

### Operation-specific work

- **Revert.** `RevertSource` promoted to `pub` and renamed to `RevertSelection` (cleaner mapping to git terminology — git revert doesn't use "source"). The `selection` field on `RevertRequest` carries either a `Commits(NonEmpty<CommitRef>)` set or a `Range { from, to }`.
- **Merge.** Cache namespacing uses `(ledger:source_branch, idempotency_key)` since the source branch is always known up front; the target branch may be omitted on the wire and resolved from the source's parent.
- **Rebase.** Cache namespacing uses `(ledger:branch, idempotency_key)` — the branch being rebased is the natural client identifier and matches the URL clients would use to check status.
- **Push.** Precomputed-commit ingestion (`POST /v1/fluree/push/*ledger`) now also flows through consensus. `IndexConfig` is owned by `MonolithicConsensus` rather than reconstructed at the route layer.

### Type-system invariants

- **`TransactionBody` flattened.** The old `(TxnType, TransactionBody)` pair encoded invalid combinations at runtime (e.g. Turtle-Update returned a 400 inside the consensus impl; TriG-Insert returned a 400 at the route). Replaced with seven self-describing variants — `JsonLdInsert`, `JsonLdUpsert`, `JsonLdUpdate`, `TurtleInsert`, `TurtleUpsert`, `TrigUpsert`, `Sparql` — so the invalid combinations are unconstructable. The route layer's `execute_turtle_transaction` takes a narrower `TurtleOp { Insert, Upsert }` enum at its boundary so the (Turtle, Update) case can't be passed in either.
- **`TransactionBody::operation_tag()`** centralizes the "insert"/"upsert"/"update"/"sparql-update" string tags used in telemetry on the enum itself, so adding a new body variant is enforced at the call site by the exhaustive match.
- **`ConflictStrategy` is now `Copy`** (pure unit-variant enum). Removes a handful of unnecessary `.clone()` calls across the consensus and CLI crates.
- **`GovernanceOptions::tracking` removed.** It was a duplicate-and-unused tracking field — the live tracking path goes through `TransactionRequest::tracking`. Eliminating the duplicate makes the type honest about where tracking comes from.

### HTTP integration

- `routes/transact.rs`, `routes/ledger.rs` (revert/merge/rebase), and `routes/push.rs` now construct typed consensus requests and call the appropriate `Committer` method. Idempotency keys are extracted from the `Idempotency-Key` header via a shared `extract_idempotency_key` helper.
- New `routes/submissions.rs` exposes `GET /v1/fluree/submissions/:key/*ledger`, returning a polymorphic `OperationStatusResponse` (`Transaction` / `Revert` / `Merge` / `Rebase` / `Push` variants) for clients recovering from a lost submission response.
- Shared error mapping: `submission_error_to_server_error` maps `SubmissionError` → HTTP status (409 for `KeyCollision`/`AlreadyInFlight`, 503 for `Overloaded`, pass-through for `Execution`).

## Design decisions worth flagging

- **Per-method trait, not unified `submit`.** A single-method trait taking an `Operation` enum was prototyped but reverted: it required an `unreachable!` arm in every typed wrapper to extract the matching receipt variant, with no compile-time guarantee that the implementation returned the right one. The per-method trait keeps typing precise at the cost of structural duplication across implementations — implementations multiplex internally either way.
- **Ledger / branch lifecycle operations (`create_ledger`, `drop_ledger`, `create_branch`, `drop_branch`) deliberately do not go through `Committer`.** They are nameservice mutations, and the nameservice already provides atomic CAS. Pushing them through consensus would just duplicate the coordination primitive at a different layer.
- **The CLI's local mode bypasses consensus.** When a `fluree` CLI command runs against an on-disk ledger directory with no server present, it constructs an embedded `Fluree` and calls API methods directly — no admission control, no idempotency cache. Cross-process correctness is preserved via the file-level `flock` + monotonic-`t` CAS in the file-backed nameservice. This is a deployment-policy concern rather than an architectural gap.
- **Admission rejects, doesn't queue.** When the pending cap is reached, new submissions return `Overloaded` immediately instead of waiting for a permit. A wait queue would just move the unbounded memory growth from the in-flight set to the queue — clients should retry with backoff against a 503.

## Testing

- 30 new unit tests in `fluree-db-consensus/src/monolithic.rs` covering each operation's anonymous, keyed, idempotent-retry, status-lookup, key-collision, and admission-rejection paths, plus body-format and tracking interactions.
- Existing `fluree-db-server` and `fluree-db-api` test suites continue to pass.
- `cargo fmt --check` and `cargo clippy --workspace --lib --tests` are clean.

## Migration notes

- **New HTTP wire concept: `Idempotency-Key` header.** Clients that want retry-safe semantics on POST `/transact`, `/insert`, `/upsert`, `/update`, `/revert`, `/merge`, `/rebase`, `/push` should generate a ULID, send it as `Idempotency-Key`, and use `GET /submissions/:key/*ledger` to recover a lost response. Existing clients that omit the header behave exactly as before.
- **New HTTP status: 503 `Overloaded`.** Servers under sustained load now refuse new submissions when the in-flight cap is reached. Clients should retry with exponential backoff.
- **`Committer`, `SubmissionLookup`, `IdempotencyKey`, `OperationReceipt`, `SubmissionState`, and the per-op request/receipt types** are the public API of the new crate. Future replicated consensus implementations (Raft, BFT) plug in behind these.


fluree-db-consensus/src/monolithic.rs
```diff
@@ -0,0 +1,1489 @@
...
+    fn hash_request_body(request: &TransactionRequest) -> [u8; 32] {
+        let mut hasher = Sha256::new();
+        // Each variant tag distinguishes both format and insert/upsert/update
+        // semantics — same bytes under a different variant produce different
+        // hashes, so two retries that disagree on operation kind collide
+        // correctly.
+        match &request.body {
+            TransactionBody::JsonLdInsert(json) => {
+                hasher.update(b"jsonld-insert");
+                hasher.update(json.to_string().as_bytes());
```
bplatz: `hash_request_body` calls `json.to_string()` for each JSON-LD variant, then feeds the resulting bytes to SHA-256. That re-serializes the entire body into a fresh heap-allocated `String` on every keyed submission. Streaming straight into the hasher via `serde_json::to_writer(&mut hasher, json)` (`Sha256` implements `io::Write`) gets the same digest with no intermediate buffer and one less full-body walk. Same trick applies to `hash_push_body` if you ever switch from raw `&[u8]` commits to a structured form.

fluree-db-consensus/src/monolithic.rs
```diff
@@ -0,0 +1,1489 @@
...
+/// Default TTL for idempotency cache entries (1 hour).
+///
+/// After this duration, a previously-recorded submission state is forgotten;
+/// the same idempotency key may then be reused for a new submission with any
+/// body, and status lookups for the expired key return [`SubmissionState::Unknown`].
+pub const DEFAULT_IDEMPOTENCY_TTL: Duration = Duration::from_secs(3600);
+
+/// Upper bound on idempotency cache entries, so sustained keyed traffic
+/// can't grow the cache without limit between TTL evictions.
+const IDEMPOTENCY_CACHE_CAPACITY: u64 = 100_000;
```
bplatz: `IDEMPOTENCY_CACHE_CAPACITY` bounds the cache to 100k entries by count, but `CachedSubmission` size is highly variable — a `Committed(OperationReceipt::Transaction)` with a tracking-enabled `TrackingTally` (especially its `policy: HashMap<String, PolicyStats>`) can be orders of magnitude larger than a bare `RevertReceipt`. Under sustained keyed traffic with tracking on, the cache's resident footprint can drift well above what the entry count suggests. Moka supports `.weigher(|_, v| …)` paired with `.max_capacity(byte_budget)` if you'd rather bound this in bytes.

fluree-db-consensus/src/monolithic.rs
```diff
@@ -0,0 +1,1489 @@
...
+        self
+    }
+
+    /// Try to claim one of the in-flight admission permits, refusing the
+    /// submission outright when the cap is reached. The returned permit
+    /// drops (and releases its slot) when the caller's submission future
+    /// completes.
+    fn try_admit(&self) -> Result<OwnedSemaphorePermit, SubmissionError> {
+        Arc::clone(&self.admission)
+            .try_acquire_owned()
```
bplatz, line 401: `try_admit` clones `self.admission` and calls `try_acquire_owned`, but the returned `OwnedSemaphorePermit` is bound to `_permit` in the same async scope as the rest of the method. `self.admission.try_acquire()` would cover the same lifetime without the Arc bump. Worth keeping the owned form only if execution ever moves into a `tokio::spawn`'d task that needs to carry the permit along.

fluree-db-consensus/src/monolithic.rs
```diff
@@ -0,0 +1,1489 @@
...
+    ) -> Result<TransactionReceipt, SubmissionError> {
+        let _permit = self.try_admit()?;
+
+        // Anonymous submissions (no idempotency key) skip the cache
+        // entirely — no retry-collapse and no later status lookup.
+        let Some(idempotency_key) = request.idempotency_key.clone() else {
+            return self.execute_transaction(request).await;
+        };
+
+        let cache_key = (request.ledger_id.clone(), idempotency_key);
```
bplatz: The `let cache_key = (request.ledger_id.clone(), idempotency_key)` followed by `try_claim_slot(cache_key.clone(), …)` allocates two copies of the ledger_id `String` per keyed submission — one to build the tuple, one because `record_outcome` later needs its own owned copy. Wrapping `SubmissionCacheKey` in `Arc` (or restructuring `try_claim_slot` to take `&SubmissionCacheKey` and only clone when the insert wins) would drop both clones to refcount bumps. Same shape repeats in `revert` (608-612), `merge` (635-640), `rebase` (663-667), and `push` (689-692).

fluree-db-server/src/routes/submissions.rs
```diff
@@ -0,0 +1,221 @@
...
+#[derive(Serialize)]
+pub struct PushStatusResponse {
+    pub idempotency_key: Option<String>,
+    pub ledger: String,
+    pub accepted: usize,
+    pub head_t: i64,
+    pub head_id: String,
+}
+
+pub async fn submission_status(
```
bplatz, line 113: `submission_status` reads the cache and returns commit metadata (commit_id, t, reverted_commits, etc.) with no Bearer / signed-request check and no `can_read(ledger)` scope enforcement. Every other write/read route in this PR honors `DataAuthMode` and the bearer's ledger scope; this one is unauthenticated. A caller who knows or guesses an `Idempotency-Key` can probe submission state on any ledger, and the response shape leaks ledger existence even on a miss. Should match the auth pattern used in `routes/ledger.rs::info`.

fluree-db-consensus/src/monolithic.rs
```diff
@@ -0,0 +1,1489 @@
...
+    pub fn with_ttl(fluree: Arc<Fluree>, index_config: IndexConfig, ttl: Duration) -> Self {
+        let cache = Cache::builder()
+            .time_to_live(ttl)
+            .max_capacity(IDEMPOTENCY_CACHE_CAPACITY)
+            .build();
+        Self {
+            fluree,
+            index_config,
+            cache,
+            admission: Arc::new(Semaphore::new(DEFAULT_PENDING_LIMIT)),
```
bplatz, line 315: Single global `Semaphore` here means a burst of writes against one ledger draws from the same 1024 permits as every other ledger. Either here or as a follow-up: a per-ledger cap (something well below `DEFAULT_PENDING_LIMIT`) layered under the global cap as a backstop would prevent one ledger's flood from blocking everyone else.

fluree-db-consensus/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+/// Caller-provided identifier for a transaction submission.
+///
+/// Used both for idempotent retry (retries with the same key collapse to one
+/// outcome) and for status lookup via [`SubmissionLookup`]. Callers typically
+/// generate a ULID before submission so they can recover after a disconnect.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct IdempotencyKey(String);
+
+impl IdempotencyKey {
+    pub fn new(key: impl Into<String>) -> Self {
```
bplatz: `IdempotencyKey::new` accepts any string, and `extract_idempotency_key` only trims + rejects empty — no upper bound on length. A misbehaving client can send a multi-MB `Idempotency-Key` header that ends up in the moka cache key for the entry's TTL. A constructor-level cap like 128 bytes would make this self-enforcing without relying on axum's header-size config.

fluree-db-consensus/src/monolithic.rs
zonotope: I eliminated both the `ledger_id` clone by wrapping the passed ledger id in an `Arc`, and using mocha's `entry_by_ref` to eliminate the cache key clone in case of a hit.

PR #1255, opened 2026-05-28
refactor: rescale query fuel charges and add per-query floor

bplatz:
Query fuel was dominated by three "1 fuel per I/O touch" charges (index-leaflet
batch reads, persisted dict decodes, history-scan leaflets), so a scan-heavy
query could report fuel thousands of times larger than a transaction's flat
100-fuel baseline. At the same time, a query that touched no persisted data —
or errored during planning — could report `0` fuel. The indexer wasn't metered
at all: a full rebuild on a large ledger reported nothing.

This PR (1) rescales query I/O touches **100×** down, (2) adds a one-time
**query floor**, (3) drops the **transaction baseline** from 100 → 10 fuel so
it sits proportionate to scan-dominated queries, and (4) introduces **indexer
fuel accounting** that bills every CAS write the indexer makes and surfaces
the tally through `/reindex`, `trigger_index`, and the background indexer
logs.

### 1. Query touch rescale (1.000 → 0.010 fuel)
The three "touch" charges drop from `1000` to `10` micro-fuel:
- index-leaflet batch read (`binary_cursor`)
- persisted forward-dict decode (`binary_index_store::charge_dict_touch`)
- history-scan leaflet base charge (`binary_history`)

Per-row/per-flake (1 µf), overlay rows, and the per-call function/expression
micro-charges (1–5 µf) are unchanged.

### 2. Query floor (1.000 fuel, once per query)
A one-time floor is charged at query entry, **before parsing**, so that:
- a successful query always reports ≥ `1.000` fuel,
- a parse/plan error on a tracked query still reports the floor, and
- a sub-floor `max-fuel` is rejected up front (before parse/plan work).

Wired into every query entry path: view + dataset tracked paths (`query_tracked`,
`query_dataset_tracked`, R2RML variants), view + dataset untracked
`max-fuel`-limit paths (placed after the single-ledger fast-path delegate to
avoid double-charging), the BM25 / vector graph-source path
(`query_dataset_with_bm25`), and all eight connection tracked entry points
(JSON-LD/SPARQL × policy/r2rml). The floor is **not** applied to transactions.

### 3. Transaction baseline 100 → 10 fuel
With touches now at 0.010 fuel, a 100-fuel commit baseline overstated the
relative cost of small transactions by ~10×. `TXN_BASELINE_MICRO_FUEL` drops
from `100_000` to `10_000` (10.000 fuel). All commit / bulk-import paths pick
it up via the centralized constant.

### 4. Indexer fuel accounting (new)

**Where the charges land:**
- **1.000 fuel per successful CAS write** the indexer makes (every
  `ContentStore::put` / `put_with_id` — leaves, branch manifests, roots, dict
  packs, reverse-tree nodes, history sidecars, garbage records, stats
  sketches, spatial / fulltext arenas).
- **+1.000 fuel per re-encoded leaflet** inside an FLI3 leaf write.
  Passthrough leaflets (byte-copies carried forward during an incremental
  update) are **not** charged — no zstd work was performed. So a 100-leaflet
  leaf where only two leaflets were touched by novelty bills `1 + 2 = 3` fuel,
  not `1 + 100`.

**Implementation: wrap once at the entry point.** A new
`fluree_db_indexer::fuel::MeteredContentStore` wraps the caller-supplied
`ContentStore` for the duration of a build; every CAS write through it
charges the base per-write fee automatically. Only the two FLI3 leaf upload
sites (`build::upload::upload_indexes_to_cas` for full rebuild,
`build::incremental::upload_leaf_blobs` for incremental) add the per-leaflet
extra charge explicitly (from
`LeafInfo::re_encoded_leaflet_count`, surfaced through the leaf assembly
code). New CAS write sites get billed for free with no signature churn.

**Measurement only — never enforced.** Indexer trackers are no-limit; a
partial index is worse than a slow one, so the indexer never aborts mid-build
on a fuel limit.

### 5. Where fuel surfaces

| Path | Tracker | Fuel returned via |
|---|---|---|
| `/reindex` HTTP | API per request | `ReindexResponse.fuel` (CLI prints `Fuel: N.NNN`) |
| `trigger_index` Rust API | Orchestrator per build | `TriggerIndexResult.fuel`; coalesced trigger callers share the same value |
| Background indexer (no waiter) | Orchestrator per build | `IndexOutcome::Completed.fuel` + completion `info!` log line |
| Combined transactor + post-commit (`maybe_refresh_after_commit`, `require_refresh_before_commit`) | Per-build tracker | Logged on the completion line; intentionally **not** attributed to the transaction response |

`IndexResult.fuel`, `ReindexResult.fuel`, `TriggerIndexResult.fuel`, and
`IndexOutcome::Completed.fuel` are all `Option<f64>`:
- `Some(N)` for a build that did work,
- `Some(0.0)` when no work was needed (already-current early return,
  waiter satisfied by an already-published index, no-commit ledger),
- `None` only when going through the non-tracking public API.

The background worker logs `fuel = ?` on:
- the `info!` "Successfully indexed ledger" line,
- the `warn!` "Failed to publish index, will retry" line (so the
  already-paid-for fuel is visible even when no caller is waiting), and
- the `warn!` "Indexing failed, will retry" line (as `partial_fuel`,
  from a Tracker clone kept across the build call).

### 6. Centralized fuel schedule

All structural charges live in `fluree_db_core::tracking::schedule`:
`QUERY_FLOOR_MICRO_FUEL`, `INDEX_TOUCH_MICRO_FUEL`, `DICT_TOUCH_MICRO_FUEL`,
`HISTORY_LEAF_TOUCH_MICRO_FUEL`, `PER_ROW_MICRO_FUEL`, `TXN_BASELINE_MICRO_FUEL`,
and the new `INDEX_CAS_WRITE_MICRO_FUEL`. Per-call function and R2RML
micro-charges remain inline at their eval sites.

### Public API

Existing entry points (`build_index_for_record`, `build_index_for_ledger`,
`rebuild_index_from_commits`, `rebuild_index_from_commits_with_store`,
`upload_indexes_to_cas`) are **unchanged** and continue to pass a disabled
tracker internally — `fuel` is reported as `None` on those paths. New
`*_with_tracker` variants take a `Tracker` and wrap the store; `/reindex` and
the background indexer call those. No existing caller breaks.

## Behavior changes / breaking notes

- **Breaking re-scale (queries + transactions).** Historical fuel numbers and
  dashboards are not comparable across the cutover. Existing `max-fuel`
  limits now permit ~100× more scan work. Re-derive any externally-configured
  limits.
- **`max-fuel` semantics.** `max-fuel: 1` permits exactly the query floor;
  `max-fuel: 1.01` permits the floor plus one persisted touch; `max-fuel`
  below `1.0` is rejected before parsing.
- **Transaction read-side cost drops.** Upsert lookups / SHACL reads /
  UPDATE-WHERE matching go through the same scan path, so their I/O touches
  get the 100× cut too (transaction baseline + write cost unchanged).
- **`/reindex` response shape (additive).** `ReindexResponse` gains
  `fuel: Option<f64>`; the field is omitted from JSON when `None` so existing
  clients keep working.
- No default `max-fuel` exists in the server/CLI (limits are always
  user-supplied), so there was nothing to re-derive there.

## Docs

- `docs/query/tracking-and-fuel.md`: updated cost ladder (query floor +
  rescaled touches + 10-fuel txn baseline + indexer CAS write + re-encoded
  leaflet); new "Indexing Fuel" section spelling out where fuel surfaces per
  path and the measurement-only semantics.
- `docs/reference/glossary.md`: corrected the inaccurate "1 unit per item"
  Fuel definition.
- `docs/api/endpoints.md` (`POST /reindex`): example response and field table
  now include `fuel`.
- `docs/indexing-and-search/background-indexing.md`: `trigger_index()`
  paragraph describes `TriggerIndexResult.fuel` and coalesced-trigger
  semantics.

## Testing

- **Unit** (`fluree-db-core/src/tracking.rs`): floor = `1.000`, each touch =
  `0.010`, floor + touch = `1.010`, sub-floor `max-fuel` rejected, and the
  add-then-check ordering so an exceeded floor still tallies `1.000`.
- **Unit** (`fluree-db-indexer/src/fuel.rs`, 5 tests): base write = 1 fuel,
  leaf + 3 re-encoded leaflets = 4 fuel, disabled tracker is a no-op,
  `MeteredContentStore` charges 1 fuel per put, leaf + extra leaflets sum
  correctly.
- **Integration — query** (`fluree-db-api/tests/it_fuel_floor.rs`, 10 tests):
  tiny query ≥ floor, parse error reports floor, sub-floor `max-fuel`
  rejected on tracked, untracked-limit, connection JSON-LD (`opts`),
  connection SPARQL (header override), and BM25 paths; connection
  missing-spec / SPARQL parse errors report the floor.
- **Integration — indexing** (`fluree-db-api/tests/it_indexing_fuel.rs`,
  4 tests): rebuild-with-tracker reports `Some(N)` > 0; already-current build
  reports `Some(0.0)`; non-tracking rebuild reports `None`; `trigger_index`
  via the background indexer reports `Some(N)` > 0 end-to-end.
- Exhaustive `IndexOutcome::Completed` pattern matches in 5 existing
  integration tests updated for the new `fuel` field.
- Comments in `it_ledger_lifecycle.rs` updated to reflect the new mechanism;
  `it_policy_tracking.rs` assertion updated to the new 10-fuel baseline
  (`10.003` not `100.003`).

### DIFF
--- a/docs/api/endpoints.md
+++ b/docs/api/endpoints.md
@@ -2348,7 +2348,8 @@ curl -X POST http://localhost:8090/v1/fluree/reindex \
     "leaf_count": 614,
     "branch_count": 23,
     "total_bytes": 47185920
-  }
+  },
+  "fuel": 1734.0
 }
 ```
 
@@ -2361,6 +2362,7 @@ curl -X POST http://localhost:8090/v1/fluree/reindex \
 | `stats.leaf_count` | Number of leaf nodes written |
 | `stats.branch_count` | Number of branch nodes written |
 | `stats.total_bytes` | Bytes written to storage during the reindex |
+| `fuel` | Total decimal fuel charged for the reindex's CAS writes (1.000 per write + 1.000 per re-encoded leaflet in FLI3 leaves). `0.0` if the index was already current. See [Tracking and Fuel](../query/tracking-and-fuel.md#indexing-fuel) for the full schedule. |
 
 **Status Codes:**
 - `200 OK` — reindex complete
--- a/docs/indexing-and-search/background-indexing.md
+++ b/docs/indexing-and-search/background-indexing.md
@@ -198,6 +198,14 @@ is optional and should generally be chosen by the caller. Leave
 explicitly for bounded environments such as Lambda jobs, HTTP gateways, or
 other workers with a fixed maximum runtime.
 
+`TriggerIndexResult` carries `fuel: Option<f64>` — the total fuel charged for
+the build that satisfied the trigger. The background indexer always meters
+each build, so callers see `Some(N)` for a real build, `Some(0.0)` when the
+requested t was already indexed (no work was needed), and the same value for
+all coalesced trigger callers satisfied by a single underlying build. See
+[Tracking and Fuel](../query/tracking-and-fuel.md#indexing-fuel) for the
+cost-ladder and rate semantics.
+
 ### Health Indicators
 
 **Healthy:**
--- a/docs/query/tracking-and-fuel.md
+++ b/docs/query/tracking-and-fuel.md
@@ -54,10 +54,13 @@ Cost ladder (per event):
 
 | Event | Cost (fuel) |
 |---|---|
-| Index leaflet touched (per scan batch, regardless of cache state) | 1.000 |
-| Forward-dict touch (per dict-backed value resolved during result materialization) | 1.000 |
+| Query floor (once per query, charged at entry before parsing) | 1.000 |
+| Index leaflet touched (per scan batch, regardless of cache state) | 0.010 |
+| Forward-dict touch (per dict-backed value resolved during result materialization) | 0.010 |
+| History-scan leaflet base (per leaflet; per-row costs below add on top) | 0.010 |
 | Flake returned from a `db.range` call (e.g. SHACL graph reads, graph crawl) | 0.001 |
 | Overlay/novelty row materialized | 0.001 |
+| History row scanned (base + in-range sidecar rows) | 0.001 |
 | R2RML row emitted (Iceberg/Parquet) | 0.001 |
 | Transaction commit baseline (once per commit, including each bulk-import chunk) | 100.000 |
 | Staged flake (per flake in a transaction or bulk-import chunk) | 0.001 |
@@ -69,6 +72,8 @@ Cost ladder (per event):
 | `Fulltext` (per-row BM25 scoring) | 0.005 |
 
 Cheap operations (comparisons, arithmetic, type checks, simple string ops, datetime extraction, etc.) cost zero — instrumentation overhead would dwarf the actual cost.
+
+The **query floor** guarantees every fuel-tracked query reports at least `1.000` fuel: a query touching no persisted data still costs the floor, and a query that errors during parsing/planning still reports it. I/O "touches" cost `0.010` each, so a scan-dominated query reports roughly `1.000 + 0.010 × (leaflet/dict touches)`. The fuel schedule above is defined in one place — `fluree-db-core/src/tracking.rs` (`tracking::schedule`).
 
 ### Setting Fuel Limits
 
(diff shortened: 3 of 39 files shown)


zonotope (approved):
🥦

PR #1256, opened 2026-05-28
fix: avoid full-predicate scans for bound-object queries

bplatz:
Several common bound-object query shapes were silently doing **full predicate
scans** when they should have been cheap — either provably empty or a tight
index seek. The cost scales with predicate cardinality, so a query that returns
**nothing** could burn hundreds of thousands of fuel. This PR closes that class
across the scan, count, and string-object paths, and adds present-string
narrowing.

Concretely, on a large ledger, `@type {rdfs:Class, rdf:Property}`
returned **0 rows** but consumed **705K fuel** — two full scans of the entire
`rdf:type` predicate to discover that those classes are never used as type
objects. After this PR that query is ~0 fuel.

## Root cause

When `BinaryScanOperator::open` couldn't encode a bound object to a tight
`(o_type, o_key)` range, it fell into an un-narrowed catch-all and post-filtered
the whole predicate. The cases that hit this:

1. **Object absent from the dictionary** (a class nothing is typed as, a ref IRI
   never interned, a string value never stored). This was misclassified as
   "can't narrow" instead of "provably empty" — so the *cheapest* possible query
   became the *most expensive*.
2. **`Term::Sid` objects resolved store-only**, so a snapshot↔store
   namespace-code divergence (e.g. after bulk import) made a *present* ref look
   absent — a latent correctness risk, not just a perf one.
3. **Untyped string objects never narrowed at all** — `datatype_sid_for_untyped_value`
   had no `String` arm, so even a pure `xsd:string` predicate full-scanned.
4. The **`count_bound_object_v6`** fast path errored on an absent object and fell
   back to a generic scan + aggregate (counted zero via a full scan).

## What changed

| Commit | Path | Fix |
|--------|------|-----|
| scan refs absent | `binary_scan.rs` | absent ref object → overlay-only short-circuit (checks novelty, no base scan) |
| count refs absent | `fast_group_count_firsts.rs` | absent count object → `Ok(0)`; snapshot-aware resolution |
| absent strings | `binary_scan.rs` | absent untyped string → overlay-only (shared `str_id`, so a dict miss means no string-typed flake matches under any datatype/lang) |
| scan refs snapshot-aware | `binary_scan.rs` | object-ref resolution routed through `subject_ref_to_s_id` (decode_sid → store IRI lookup), matching the subject and count paths — closes the `Term::Sid` false-negative |
| present strings narrow | `binary_scan.rs`, `value_id.rs` | untyped string narrows to the predicate's lone string datatype when stats prove there's exactly one (non-lang) string-compatible type |
| string o_key prefilter | `binary_scan.rs` | untyped string on a langString / multi-string-compatible predicate uses the interned `str_id` as an `o_key` pre-filter (drops rows before value decode), preserving lenient matching |

The bound-object block in `open()` is restructured into three branches: **ref**
(snapshot-aware tight seek), **untyped non-narrowable string** (`o_key`
pre-filter), and **typed/inferred values** (exact encode).

## Correctness invariants

- **Base-dict miss ≠ logical-DB miss.** A miss in the persisted dictionary only
  becomes "no match" when novelty is also accounted for. The scan path routes
  misses to the overlay-only fallback (which checks novelty); the count fast path
  is gated on no-novelty (`epoch == 0`), so a base-dict miss there *is* a logical
  miss.
- **Novelty-aware stats.** The per-predicate datatype stats used for narrowing
  are novelty-aware (`assemble_fast_stats` merges novelty), so "only `xsd:string`
  present" reflects base **and** novelty — a novelty langString surfaces as a
  second string-compatible tag and declines narrowing. No epoch guard needed for
  string narrowing.
- **Snapshot-aware ref resolution.** Refs (subject, object, and count object) all
  resolve through `subject_ref_to_s_id` (by-parts, then `snapshot.decode_sid` /
  `store.sid_to_iri` → full-IRI lookup), so a raw `Term::Sid` of unknown
  provenance can't be mistaken for absent.
- **Lenient string semantics preserved.** An untyped `"foo"` still matches the
  value across all string datatypes/langs (`xsd:string`, `langString`, etc.). The
  narrowing only fires when a single string-compatible type is present; otherwise
  the `o_key` pre-filter keeps the lenient match while cutting the per-row decode.
  `UNKNOWN` datatype is treated as unsafe (declines narrowing).

## Performance / fuel

- `@type <absent-class>` (0 rows): **~352K → ~0 fuel** per absent class.
- `?s <p> "absent-string"`: full predicate scan → **~0 fuel**.
- `COUNT WHERE { ?s a <absent-class> }`: full scan + aggregate → **`Ok(0)`** with no scan.
- Present untyped string on a pure / mixed-with-non-string predicate: full scan →
  **tight `(XSD_STRING, str_id)` seek**.
- Present untyped string on a langString / multi-string-compatible predicate:
  full value-decode scan → **`o_key` pre-filter** (measured: a 502-flake
  string+langString predicate dropped from a full decode scan to ~3 fuel), still
  returning all lenient matches.

These reduce the actual work (and lambda cost), not just the reported fuel number.

## Testing

- New `it_query_fuel_bound_object.rs`: absent ref/class/string/count all return
  bounded fuel independent of dataset size; present-value lookups still return
  correct rows/counts; present untyped strings narrow (pure + mixed); a
  negative guard proves a `xsd:string + langString` predicate still returns **all**
  lenient matches (declines to narrow) and an absent string there still
  short-circuits.
- Existing guards: `it_query_rdfs_class_repro` (present classes on bulk-imported
  data must return correct counts/rows — covers the `Term::Sid` snapshot path),
  `it_query_datatype` (full datatype/lang semantics suite), and
  `it_import` / `it_query_time_travel` (caught the `Ok(None)` count regression
  during development; now pass and guard `Term::Sid` count objects).
- CI: `cargo fmt --all -- --check`, `cargo clippy --all --all-features
  --all-targets -- -D warnings`, `cargo nextest run --workspace --all-features`
  (clean apart from environment-only Docker/testcontainers and a transient
  swagger-ui asset download).

### DIFF
--- a/fluree-db-api/tests/it_query_fuel_bound_object.rs
+++ b/fluree-db-api/tests/it_query_fuel_bound_object.rs
@@ -1,18 +1,23 @@
-//! Regression test for bound-object scans against the binary (FIR6) index.
+//! Regression tests for bound-object queries whose object is **absent** from
+//! the persisted dictionary, against the binary (FIR6) index.
 //!
-//! Bug (fix #1): a constant object that is **absent** from the persisted
-//! dictionary (a class that no entity is typed as, or a ref IRI that no flake
-//! references) used to drop the object filter and fall into the un-narrowed
-//! catch-all in `BinaryScanOperator::open`, scanning the *entire* predicate and
-//! post-filtering to discover 0 matches. Cost scaled with predicate cardinality
-//! (e.g. ~1 fuel per `rdf:type` flake), so a query returning nothing could cost
-//! hundreds of thousands of fuel.
+//! Bug class: a constant object that no flake references (a class nothing is
+//! typed as, or a ref IRI never interned) used to be mishandled as "can't
+//! narrow" instead of "provably empty", causing a full predicate scan:
+//!   - `BinaryScanOperator::open` dropped the object filter into its
+//!     un-narrowed catch-all and post-filtered the whole predicate;
+//!   - `count_bound_object_v6` errored and fell back to a generic
+//!     scan + aggregate.
 //!
-//! An object value provably absent from the base dictionary cannot be
-//! referenced by any base flake, so the scan must short-circuit to the
-//! overlay-only (novelty) path instead. This test pins that invariant: the
-//! fuel for an absent-object query is bounded by a small constant **independent
-//! of dataset size**, while present-value lookups still return correct rows.
+//! Either way cost scaled with predicate cardinality (~1 fuel per `rdf:type`
+//! flake), so a query returning nothing could burn hundreds of thousands of
+//! fuel.
+//!
+//! An object provably absent from the base dictionary cannot be referenced by
+//! any base flake, so these paths must short-circuit (scan → overlay-only;
+//! count → 0). These tests pin that invariant: fuel for an absent-object query
+//! stays bounded by a small constant **independent of dataset size**, while
+//! present-value lookups still return correct rows/counts.
 
 #![cfg(feature = "native")]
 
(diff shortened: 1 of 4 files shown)


zonotope (approved):
🚠

PR #1257, opened 2026-05-28
fix(memory): drop memory ledger by full name and surface ledger_exists errors

bplatz:
- `drop_and_reinit` now passes `MEMORY_LEDGER` to `drop_ledger` so the whole ledger is dropped, not the bare id
  - `is_initialized` propagates `ledger_exists` errors instead of swallowing them as `false`
  - Adds tests covering both paths


zonotope (approved):
🧢

PR #1264, opened 2026-06-01
rollup: Cumulative work from #1258, #1260, #1261, #1262, #1263 (mis-merged to solo/latest)

bplatz:
## Rollup PR — cumulative `solo/latest` → `main`

The five PRs below were **mistakenly merged into `solo/latest`** instead of `main` on 2026-05-31. Because their head commits became reachable through those merges, GitHub auto-marked all five as `MERGED` and closed them — so they can no longer be reopened or reviewed individually. This PR re-surfaces their **combined diff for review into `main`** as a single rollup.

`main` has no divergent commits; `solo/latest` is strictly ahead, so this diff is exactly the cumulative work and nothing else.

### Consolidated PRs
- #1263 — feature: Expand bulk-import format coverage: N-Triples, N-Quads, TriG named graphs, transparent gzip/zstd
- #1262 — refactor: Fold query fast-paths into the count planner + overlay-aware COUNT + reverse-POST ORDER BY DESC LIMIT
- #1261 — fix: Eliminate S3-contention wedges (indexing, multi-query, reads) + multi-query envelope & indexing-fuel
- #1260 — feat: Gate hydration SPOT crawl by projection predicates
- #1258 — Feature/multi query

### Follow-up commits (not part of any of the 5 PRs above)
Made directly on `solo/latest` later the same evening; all extend the bulk-import work in #1263:
- `d58f04b30` / `3270cc4c9` — fix: guard Turtle lexer/parser against >4GiB input overflow
- `a984205c3` — perf: rechunk directory imports by bytes (split large + coalesce small)
- `9da6c0a00` — Merge branch 'feature/nt-import' into solo/latest

CI must pass: clippy (all features, all targets) + nextest (workspace, all features).


aaj3f (approved):
# PR #1264 — Review Worksheet

Two sections:
1. **Inline comments & suggested edits** — grouped by severity, each with the exact
   file:line to comment on and proposed wording/fix.
2. **Top-level review submission comment** — copy/paste as the review body.

Legend: 🔴 Critical (blocking) · 🟠 High · 🟡 Medium · 🔵 Low · ⚪ Nit/Info

---

## 1. Inline comments and/or suggested edits

### 🔴 CRITICAL-1 — Reverse-POST overlay lane returns wrong DESC order / wrong top-k under novelty
**File:** `fluree-db-query/src/fast_post_order_limit.rs:578` (and the window at `:613-619`, fall-through at `:646`)

> The overlay lane bounds each leaf's assert window with `leaf_lo = leaf.first_key.o_key`
> (line 578), then `take_while(|r| r.o_key >= leaf_lo)`. But `first_key` is the first
> record of the **whole leaf**, and leaves are **not predicate-homogeneous**: leaflets
> flush on `p_id` change (`format/leaf.rs:229-233`) but a *leaf* only flushes on
> `leaf_target_rows` (`:189`). `leaf_entries_for_predicate` (`fast_path_common.rs:292`,
> via `find_leaves_in_range`) returns the boundary leaves, whose `first_key` can belong
> to a **lower predicate** sharing the leaf. So `leaf_lo` is an unrelated predicate's
> `o_key` (possibly a different `o_type`'s encoding space).
>
> Consequence: when a shared leaf's foreign `first_key.o_key` is greater than some of
> this predicate's assert `o_key`s, those asserts are excluded from the window and fall
> to `tail_asserts` (line 646), which is emitted *after* every base row. A freshly
> written value that should rank at/near the top of `ORDER BY DESC(?o) LIMIT k` is then
> ordered last — and if `emit_merged` reaches `need` first (line 635), it is dropped from
> the result entirely. This is wrong order **and** a wrong top-k *set*, data-dependent,
> on the precise "latest feed right after a write" workload this path targets. Most acute
> for small/medium predicates that fit within one shared leaf.
>
> The production cursor does this correctly via `compute_overlay_window`
> (`binary_cursor.rs:691`) using the full `first_key`/`next_first_key` at *leaflet*
> granularity. The base lane here is fine — only the overlay window is wrong.
>
> **Fix options (pick one):**
> - **Correct fix:** bound the window by the actual minimum `o_key` of *this predicate's*
>   rows in the leaf. You already build `leaf_base` (DESC) by filtering
>   `entry.p_const == Some(p_id)`; use `leaf_base.last().map(|r| r.o_key)` as `leaf_lo`
>   (and skip leaves where `leaf_base` is empty), rather than `leaf.first_key.o_key`.
>   Equivalently, reuse `compute_overlay_window` semantics.
> - **Safe interim:** when novelty is present, **return `Ok(None)`** from the overlay lane
>   and fall back to the generic top-k. Loses the fast path under novelty but is provably
>   correct; ship the windowing fix as a follow-up.
>
> **Either way, add e2e regression tests** (none exist for the overlay leaf-walk): a
> predicate sharing a leaf with a lower-`p_id` predicate whose values are larger, plus a
> novelty assert that should top the result; and a novelty-only-subject case.

---

### 🔴 CRITICAL-2 — Multi-query SPARQL aliases are not policy/identity enforced
**Files:** `fluree-db-server/src/routes/query.rs:2703-2714`; `fluree-db-api/src/query/multi_dispatch.rs:692-730`; `fluree-db-api/src/query/multi_run.rs:98-130`

> The per-alias auth loop only runs for JSON-LD:
> ```rust
> // query.rs:2708
> if matches!(sub.language, SubqueryLanguage::JsonLd) {
>     premerge_opts_into_subquery_body(...);
>     apply_envelope_subquery_auth(...).await;   // impersonation gate + identity inject
> }
> ```
> SPARQL aliases skip the impersonation gate **and** identity injection. And the
> dispatcher's SPARQL branch (`multi_dispatch.rs:692`) → `run_sparql_subquery`
> (`multi_run.rs:98`) only threads `TrackingOptions` (time/fuel/policy *tracking* flags +
> `max_fuel`) — there is **no channel** for `identity` / `policy_class` / `policy` /
> `policy_values`. `query_from().sparql()` resolves to `build_dataset_view` (no policy),
> not `_with_policy`.
>
> Contrast the **single-query** SPARQL path (`query.rs:1473-1577`), which builds
> `QueryConnectionOptions { identity, policy_class, policy, policy_values, default_allow }`
> and routes through `db_with_policy` when `has_any_policy_inputs()`, and runs
> `resolve_sparql_identity` (`policy_auth.rs:148`). JSON-LD aliases here are enforced
> correctly; SPARQL aliases are not.
>
> **Effect:** a caller authorized to *reach* a ledger (ledger-scope **is** enforced for
> both languages, `query.rs:2666`) but whose identity should see a restricted view via
> `f:policyClass` will see **all** rows/predicates of that ledger through a SPARQL alias.
> Server-side impersonation decisions and `fluree-policy-*` headers are silently dropped.
>
> **Fix options:**
> - **Correct fix:** build a policy context from the merged opts for SPARQL too, run the
>   impersonation gate for SPARQL aliases (mirror `resolve_sparql_identity`), and add a
>   policy-aware SPARQL execution helper (`query_from().sparql().policy(ctx)`).
> - **Safe interim (fail-closed):** reject (envelope validation error) any SPARQL alias
>   when the request carries identity/policy inputs, until enforcement is wired.
>
> **Add an integration test:** ledger with an `f:policyClass` restricting an identity →
> assert a SPARQL alias under that identity returns the restricted view. (Current
> `multi_query_auth_integration.rs` only covers ledger-scope 404.)

---

### 🟡 MEDIUM-1 — Overlay COUNT cursors project narrow columns; identity-merge is correct only because the cache always loads all columns
**Files:** `fluree-db-query/src/fast_path_common.rs:186-233` (`cursor_projection_*`); `fluree-db-binary-index/.../binary_cursor.rs:450-492`; masked by `column_loader.rs:248-251`

> The count-plan overlay cursors project narrow subsets (SId-only, OType+OKey).
> `merge_overlay_into_batch` compares base rows vs overlay ops on the full V3 identity
> (`p_id`, `o_type`, `o_key`, `o_i`). Unprojected columns return `AbsentDefault`
> (`p_id→0`, `o_i→u32::MAX`), which would break the sort/identity comparison and
> over/undercount. This is **masked in production** because
> `load_columns_cached_via_handle` ignores the caller projection and always loads
> `ColumnProjection::all()`, and api/server always inject a leaflet cache. A
> `BinaryIndexStore` built **without** a cache (`LedgerManagerConfig::default()`,
> `leaflet_cache: None`) would take the non-cached path and silently miscount under
> overlay. PSOT had this shape pre-PR; this PR extends it to POST.
>
> **Recommendation:** make the invariant explicit — either have
> `build_overlay_cursor_for_predicate` augment the projection with the identity columns
> (PId/OType/OI) whenever overlay ops are attached, or add a `debug_assert` in
> `BinaryCursor` that the projection includes identity columns when `set_overlay_ops` is
> non-empty. Add one overlay COUNT test on a no-cache store to lock it.

---

### 🔵 LOW / ⚪ INFO — non-blocking

- 🔵 **`fluree-db-query/src/count_plan_exec.rs:989`** — `assert!(chain.predicates.len() >= 2, …)`
  on the query path (release-mode panic) while the overlay twin (`:1717`) uses
  `debug_assert!`. Make it `debug_assert!` or `return Ok(None)`.
- 🔵 **`fluree-db-query/src/fast_post_order_limit.rs:77,480,639`** — `SCAN_BUDGET` (≈200k)
  is checked per-leaf and is smaller than one default leaf (250k rows), so a single huge
  leaf is fully scanned before the bail. Not a cliff (bounded by one leaf), but worth a
  note or a per-leaflet budget check.
- 🔵 **`fluree-db-api/src/query/multi.rs:403-408`** — no aggregate fuel ceiling for an
  envelope (per-alias `max-fuel` only); 64 aliases × unbounded per-query fuel, bounded
  only by concurrency (16) + timeout (60s). Documented v1 limitation — call it out
  explicitly or add an optional aggregate cap later.
- 🔵 **`fluree-db-api/src/import.rs:1259-1265`** — a highly compressible "small" `.gz`
  (ratio ≫ the `×6` `COMPRESSED_DECODE_RATIO`) can be read whole above the chunk budget
  (e.g. ~128 MB gz → multi-GB `String`). Adversarial only; cap the whole-read or document
  the ratio as a hard assumption.
- 🔵 **`fluree-db-api/src/import.rs:4174,4238`** — named-graph SPOT class stats are
  discarded (`let (ng_result, _) = …`), so `stats.graphs[user].classes` is empty. Query
  correctness unaffected (planner/`info` estimate only; per-property graph stats still
  collected), but a completeness gap for the new named-graph feature. Merge them or
  document as default-graph-only.
- 🔵 **`fluree-db-api/src/ledger_manager.rs:1147`** — reload swap-skip compares total
  `t()` (incl. novelty), not `index_t()`. Conservative/data-safe (never clobbers a newer
  in-memory commit) and self-heals next reload, but can transiently forgo a newer index.
  Worth a one-line comment acknowledging `index_t` is sacrificed to protect novelty.
- ⚪ **`fluree-db-api/src/import.rs:4215`** — comment says user graphs are "g_id >= 2";
  the code correctly starts at `FIRST_USER_GRAPH_ID` (3). Fix the comment.
- ⚪ **`fluree-db-binary-index/.../artifact_cache.rs:498-507`** — coalesced-fetch disk
  re-check swallows `Err(_)` and re-fetches from S3. Correct for transient faults, but a
  *persistent* bad cache dir becomes sustained S3 traffic instead of a surfaced error.
- ⚪ **History:** `d58f04b30` and `3270cc4c9` are byte-identical (`patch-id 5c189ad7…`) —
  the >4 GiB guard committed twice. Harmless (merge dedups), but squash/drop one before
  any future replay.
- ⚪ **`fluree-db-api/src/query/multi_snapshot.rs:290`** — `has_temporal_marker` duplicates
  `string_has_explicit_pin`; collapse to one helper.

---

## 2. Top-level approve/disapprove review submission comment

> **Status: Request changes** — two blocking defects, both isolated to the new surfaces;
> the rest of the PR is strong and, importantly, shows no performance regression.
>
> This is a well-engineered, well-tested rollup. I checked it specifically against the
> project's performance bar and found **no speed or memory regression**: the no-overlay
> COUNT lane reuses the original metadata iterators unchanged, the single-graph `.ttl`
> import hot path is provably untouched (the named-graph build range is empty for
> single-graph imports), the local-fs / cached-read path never enters the new
> `block_in_place` bridge or S3 single-flight, the reverse-POST scan is bounded, and the
> class-stats follow-up *restores* a 3.9× import regression. bench-gate is green. The
> concurrency wedge fixes (#1261) are careful and defensive (RAII slot cleanup,
> generation/ABA tokens, no-lock-across-IO, watchdog'd regression tests). Nicely done.
>
> Two things must change before merge:
>
> 1. **🔴 Correctness — reverse-POST overlay lane** (`fast_post_order_limit.rs:578`,
>    `613-619`). The per-leaf assert window is bounded by `leaf.first_key.o_key`, but
>    leaves are not predicate-homogeneous, so that bound can come from a neighbor
>    predicate. Under novelty this yields wrong `ORDER BY DESC(?o) LIMIT k` ordering and a
>    wrong top-k set — exactly the write-then-read case the path targets. Fix the window to
>    use this predicate's actual min `o_key` in the leaf (`leaf_base.last()`), or
>    reuse `compute_overlay_window`. Interim: fall back to the generic top-k when novelty is
>    present. The base lane is correct. Please add an e2e test with a predicate sharing a
>    leaf with a larger-valued lower predicate + a novelty assert.
>
> 2. **🔴 Security — multi-query SPARQL aliases bypass per-identity policy**
>    (`query.rs:2708`, `multi_dispatch.rs:692`, `multi_run.rs:98`). The per-alias auth gate
>    runs only for JSON-LD, and the SPARQL execution path has no channel for
>    identity/policy. Ledger-scope is enforced for both languages, but within-ledger
>    `f:policyClass` filtering is silently dropped for SPARQL — a caller sees unrestricted
>    data through a SPARQL alias. Thread identity/policy into the SPARQL sub-query path and
>    run the impersonation gate for SPARQL too (interim: reject SPARQL aliases when
>    identity/policy inputs are present). Please add a policy-enforcement integration test
>    for SPARQL aliases.
>
> Also worth addressing (non-blocking): make the overlay-COUNT narrow-projection invariant
> explicit so a cache-less `BinaryIndexStore` can't miscount (MEDIUM-1), downgrade the
> release `assert!` at `count_plan_exec.rs:989`, and the LOW/INFO items in the worksheet
> (no aggregate envelope fuel cap, compressed-small-file memory, named-graph class stats,
> duplicate >4 GiB commit).
>
> Once the two 🔴 items are fixed (or fail-closed with a follow-up tracked), this is an
> approve.

bplatz:
@aaj3f thanks for the detailed review — both blockers addressed:

- **🔴 CRITICAL-1** (reverse-POST overlay window, `fast_post_order_limit.rs:578`): window now bounded by the predicate's own min `o_key`, not `leaf.first_key`; + e2e regression test — `9ec038cb6`
- **🔴 CRITICAL-2** (SPARQL aliases bypass per-identity policy): per-alias impersonation gate for SPARQL + opts→policy threading through the SPARQL connection path; + policy-enforcement integration tests — `7ea295217`

LOW/INFO items noted and tracked separately.

bplatz:
@aaj3f follow-ups addressed:

- **MEDIUM-1** (overlay COUNT narrow projection — cache-less miscount): identity columns (SId/PId/OType/OKey/OI) now forced into the overlay cursor projection in `build_overlay_cursor_for_predicate`, with a `set_overlay_ops` debug-assert backstop; + no-cache overlay COUNT regression test — `4bf24256c`
- **LOW** (`count_plan_exec.rs:989` release-mode `assert!`): downgraded to `debug_assert!` to match the overlay twin `execute_chain_overlay` — `4bf24256c`

Remaining LOW/INFO items still tracked separately.

bplatz:
@aaj3f LOW/INFO cleanups addressed in `1cdb257db`:

- `import.rs:4215` — corrected the stale "g_id >= 2" comment (named-graph builds start at `FIRST_USER_GRAPH_ID`, i.e. 3+).
- `import.rs:4238` — documented that named-graph `SpotClassStats` are intentionally default-graph-only for now (per-property stats still flow through the shared `IdStatsHook`); merging the class breakdown is a tracked follow-up.
- `ledger_manager.rs:1147` — added the note that the reload swap-skip compares total `t()` (novelty included), not `index_t()`: deliberately novelty-protective, transiently forgoing a newer index rather than clobbering a newer in-memory commit.
- `fast_post_order_limit.rs` — noted `SCAN_BUDGET` is checked per-leaf, so one large leaf (up to `leaf_target_rows`, default 250k) may be fully scanned before the bail; the bound is one-leaf-granular, not exact.
- `multi_snapshot.rs` — collapsed `has_temporal_marker` into the identical `string_has_explicit_pin`.

Deliberately deferred (not quick/non-behavioral): the named-graph SPOT-stats **merge** (planner-stats plumbing, wants its own tests), the `.gz` whole-read cap and per-envelope aggregate fuel cap (behavior changes to documented v1 limits), and the `artifact_cache` `Err(_)`-swallow change (S3-traffic implications). The duplicate >4 GiB-guard commit was already merged, so no squash. Happy to take any of these on if you'd like them prioritized.

fluree-db-transact/src/parse/nquads.rs
```diff
@@ -0,0 +1,209 @@
...
+/// Most terms are a single token; a string literal may be followed by a
+/// datatype (`^^ <iri>`) or a language tag (`@lang`), which are folded into the
+/// same term.
+fn group_terms(stmt: &[&Token]) -> Result<Vec<(usize, usize)>> {
+    let mut terms = Vec::with_capacity(4);
+    let mut i = 0;
+    while i < stmt.len() {
+        let start = stmt[i].start as usize;
+        let mut end = stmt[i].end as usize;
+        let is_string = matches!(stmt[i].kind, TokenKind::String | TokenKind::LongString);
```
zonotope: this is missing `TokenKind::StringEscaped`.

fluree-db-api/src/query/multi/mod.rs
```diff
@@ -0,0 +1,1814 @@
+//! Multi-query envelope: bundle multiple independent queries against a single
```
zonotope, line 1: I think we should put the multi-query functionality in a module hierarchy. We could have a `query/multi` that contains `query/multi/dispatch`, `query/multi/response`, `query/multi/run`, and `query/multi/snapshot`, with the relevant bindings re-exported from the sub-modules through `query/multi`. That way we could build a unified multi-query api from a single module that the other components could use and we keep related functionality together.

fluree-db-transact/src/import_sink.rs
```diff
@@ -439,6 +438,43 @@ mod inner {
...
+        }
+
+        /// Spool one flake belonging to an explicit named graph (`g_id`).
+        ///
+        /// The default-graph fast path writes records under the per-chunk `g_id`
+        /// fixed at construction; named `GRAPH`-block flakes go through here so
+        /// they enter the Tier-2 index (and the `named_graphs` routing) exactly
+        /// like default-graph triples. Only invoked when an import contains named
+        /// graphs — the single-graph hot path never touches this.
+        #[allow(clippy::too_many_arguments)]
```
zonotope: I think we should only override clippy as a last resort. In this case, I think the code would be more readable if we either re-used or created a type that bundled all the flake related fields like `s`, `p`, etc. together.

zonotope (approved):

bplatz:
@aaj3f follow-up on the one LOW item I'd deferred — the named-graph SPOT class-stats merge — now implemented in `dafca1172`.

Context: the bulk import builds each graph in a separate pass and only the default-graph pass kept its `SpotClassStats`; named-graph passes discarded them, so `stats.graphs[named].classes` was always `None` (planner had no class-level cardinality for named graphs; `info` under-reported them).

- Confirmed the prerequisite first: ids are **global across all build passes** (subject/string ids assigned once at spool and shared via each `CommitInput`'s remap; `p_id` from the single shared `predicate_alloc`; same `commits` slice feeds every pass, differing only in `g_id`/`run_dir`). So the `(g_id, class_sid)`-keyed `SpotClassStats` merge has no id-space hazard — and, separately, named-graph segments are not at risk of id mismatch.
- `SpotClassStats::merge` unions the four maps; each named pass folds into the default-graph accumulator before `build_class_stat_entries`. Cross-graph ref-targets a pass can't resolve stay counted as `@id` flakes with no target-class entry (by design — no cross-graph class resolution). g1 (txn-meta, system graph) intentionally left unmerged.
- On the cap notes from your worksheet: both are left as-is, deliberately. The 64-*class* limit is already handled (`ClassMembership` promotes Dense→Sparse past 64); `MAX_CLASS_REF_LEAF_ENTRIES` is a 64M-entry anti-OOM guard, not a class cap.
- Added a TriG import regression test (named-graph class stats + same-graph ref-target class), verified RED before the merge.

That closes out the actionable items from the review.

fluree-db-transact/src/parse/nquads.rs
bplatz: f5a3573d00346b87d99637c7986e53e643205ea6

fluree-db-transact/src/import_sink.rs
bplatz: 51be0d4004ec07130cdac6971f3e29a4e91fbc92

fluree-db-api/src/query/multi/mod.rs
bplatz: 11383767dff3acb92df6a42207041bbad8e18e0d

PR #1265, opened 2026-06-01
perf: parallelize secondary index builds + memory-aware CPU model

bplatz:
Cuts bulk-import wall time ~40% (1,759s → 1,047s) on a 1.574-billion-flake corpus by removing two structural bottlenecks the import pipeline hit on multi-core machines: a serial secondary-index build, and a hard 6-thread cap on parsing. This is a follow-up to the byte-rechunk directory-import work already on `main`.

## Why

Phase-timing logs on a real import (16-core / 128 GB box) showed wall time was total-data-bound, not chunk-count-bound — halving the chunk count moved wall time ~0%. Two phases were pinned far below the machine's core count:

- **Index build** ran the secondary orders strictly serially (`PSOT → POST → OPST`), each a single-threaded k-way merge — ~595s at ~6% CPU while ~23 cores sat idle.
- **Parse** was hard-capped at `min(6)` threads regardless of available cores.

### 1. Parallel secondary-order index builds

`build_all_indexes` now builds the buildable orders **concurrently** via a bounded work-stealing `thread::scope` (new `BuildAllConfig.max_concurrency`, sized from the heavy-worker count), instead of a serial `for` loop.

- Orders read and write **disjoint directories** and return independent results — the only shared state is the atomic progress counter, so the change is low-risk.
- Serial fast path preserved when concurrency is 1; the first worker error is propagated.
- Results are sorted into canonical order before returning, so output is **deterministic regardless of thread scheduling**. (The FIR6 encoder already sorts orders by wire-id before serializing, so the root CID never depended on this — the sort makes the return contract robust for any future positional consumer.)
- SPOT continues to build on its own thread and overlaps this phase.

### 2. Memory-aware CPU resource model

Replaces the fixed `min(6)` thread cap with an auto-detect / override / floor model:

- `parse_threads` defaults to `0` = **auto**: use the machine's logical cores, then cap so peak parse memory (`threads × chunk_size × ~2.5`) stays within half the memory budget. Many-core boxes scale up; small-RAM boxes don't oversubscribe.
- **Explicit values** (builder `.threads()`, `--parallelism`, `FLUREE_IMPORT_THREADS`) are honored as-is (uncapped), with a **hard floor of 1** so imports still run on a 1–2 core box (slower, won't blow up).
- `effective_heavy_workers` derives from the same effective thread count, which also bounds secondary-build concurrency.

### 3. Docs / comments / tests

- Corrected stale references (memory auto is 60% of RAM, not 75%; parse threads no longer capped at 6) in `docs/cli/`, CLI flag help, and config doc comments.
- Added env-isolated unit tests for the thread-resolution logic (a polluted environment cannot perturb them).

## Measured impact (same 1.574B-flake dataset, 128 GB box)

| Phase          | Before  | After   | Δ                          |
|----------------|---------|---------|----------------------------|
| Parse + commit | 936s    | 633s    | −32% (parallelism 6 → 16)  |
| Index build    | 712s    | 336s    | −53% (secondaries parallel)|
| **Total wall** | **1,759s** | **1,047s** | **−40%**                |

Index-build timestamps confirm the mechanism: `post` (218.7s), `opst` (219.8s), and `psot` (233.6s) now complete within ~15s of each other (genuinely concurrent), with SPOT (313s) overlapping. The build critical path collapsed from `remap + 595s sequential` to `remap(67s) + max(SPOT 313s, secondaries 233s) ≈ 336s`.

## Risk

- **Low.** Parallel order builds touch disjoint dirs with independent results; the serial path is preserved at concurrency 1. Root-CID determinism is unchanged (verified: the encoder sorts before serializing, and the return value is now also sorted).
- Memory is self-bounding: auto thread count shrinks to fit the budget, and per-worker run budget scales inversely with worker count. Explicit overrides are the operator's call, floored at 1.

## Tuning knobs (all optional; sensible auto defaults)

- `--parallelism` / `FLUREE_IMPORT_THREADS` — parse/worker threads (`0` = auto).
- `--memory-budget-mb` — caps memory; auto thread count shrinks to fit it.
- `FLUREE_IMPORT_HEAVY_WORKERS` — overrides heavy/index-build worker count.

## Testing

- End-to-end verified on the 1.574B-flake corpus (numbers above)

### DIFF
--- a/docs/cli/README.md
+++ b/docs/cli/README.md
@@ -37,8 +37,8 @@ fluree query 'SELECT ?name WHERE { ?s <http://example.org/name> ?name }'
 | `-q, --quiet` | Suppress non-essential output |
 | `--no-color` | Disable colored output (also respects `NO_COLOR` env var) |
 | `--config <PATH>` | Path to config file |
-| `--memory-budget-mb <MB>` | Memory budget in MB for bulk import (0 = auto: 75% of system RAM). Affects chunk size, concurrency, and run budget when creating a ledger with `--from`. |
-| `--parallelism <N>` | Number of parallel parse threads for bulk import (0 = auto: system cores, default cap 6). Used when creating a ledger with `--from`. |
+| `--memory-budget-mb <MB>` | Memory budget in MB for bulk import (0 = auto: 60% of system RAM). Affects chunk size, concurrency, and run budget when creating a ledger with `--from`. Set this to cap memory use; auto-detected thread count shrinks to fit it. |
+| `--parallelism <N>` | Number of parallel parse threads for bulk import (0 = auto: most logical cores, capped to fit the memory budget; explicit values honored as-is, floored at 1). Used when creating a ledger with `--from`. |
 | `-h, --help` | Print help |
 | `-V, --version` | Print version |
 
--- a/docs/cli/create.md
+++ b/docs/cli/create.md
@@ -27,8 +27,8 @@ fluree create <LEDGER> [OPTIONS]
 
 **Global flags** that affect bulk import when using `--from` (see [CLI README](README.md#global-options)):
 
-- `--memory-budget-mb <MB>` — Memory budget in MB (0 = auto: 75% of system RAM). Drives chunk size, concurrency, and indexer run budget.
-- `--parallelism <N>` — Number of parallel parse threads (0 = auto: system cores, cap 6).
+- `--memory-budget-mb <MB>` — Memory budget in MB (0 = auto: 60% of system RAM). Drives chunk size, concurrency, and indexer run budget. Set this to cap how much memory the import uses; auto-detected thread count shrinks to fit it.
+- `--parallelism <N>` — Number of parallel parse threads (0 = auto: most logical cores, capped to fit the memory budget; explicit values honored as-is, floored at 1).
 
 ## Description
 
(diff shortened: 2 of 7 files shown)


aaj3f (approved):
This is really nice and I'm glad for every improvement or upper-bounding we can do on imports.

I'm definitely tracking the new bounds on *parse* memory and *run-buffer* memory. The only part I'm having some trouble tracking is whether those two (and the upper bounds they establish via `effective_parse_threads` and `effective_run_budget_mb`) would also establish upper bounds on the leaf byte accumulation as `LeafWriter` gathers `leaf_bytes` into `completed_leaves` prior to freeing memory in `finish_graph_v2()`. I saw the hard memory clamp via `min(worker_count, buildable.len()`, and it seems maybe that would provide bounds for the memory of leaf-byte accumulation at least vis-a-vis worker count and h ow many concurrent orders of leaf-accumulation happen at once? 

In any event, I think we have solid wins on speed & memory here. Whether or not `max_concurrency` should be given its own memory-aware ceiling for leaf byte accumulation or whether we should eventually have disk flushing inside `LeafWriter::flush_leaf()` rather than accumulating in `completed_leaves`, I think all of that could be deferred until later if we actually encounter upper bound issues on small boxes etc

bplatz:
@aaj3f Great observation, addressed in 3f5fb06cfb0d1b266b35405dd31cd30c5dc177f3.

Leaves are now streamed to disk as they're produced rather than accumulating every compressed FLI3/FHS1 blob in `LeafWriter.completed_leaves` for the whole order. A new indexer-side `PersistingLeafWriter` drains + persists each completed leaf on every `push_record`, so retained memory is the active leaf's working set plus O(#leaves) branch metadata (~100 B/leaf) — not the order's full blob set, regardless of how many orders build concurrently. Both build loops (generic + SPOT) go through it; the binary-index `LeafWriter` stays pure for the incremental paths.

PR #1267, opened 2026-06-02
fix: multi-ledger JSON-LD queries through `/v1/fluree/query` (dispatcher + formatter)

### LINKED ISSUE #1259: Multi-ledger JSON-LD queries fail through `/v1/fluree/query` (dispatcher + formatter gaps)
## Summary

Three distinct bugs prevent multi-ledger JSON-LD queries from working correctly
through the connection-scoped `/v1/fluree/query` HTTP route. Two live in the
HTTP route's body parsing; two live in `FromQueryBuilder::execute_formatted`.
They share a theme: the connection path was built around single-ledger queries
with a string `from`, and multi-ledger support was bolted on without fully
threading the multi-ledger awareness through the HTTP route or the formatter
step.

Reproduced on `release/v4.0.4` (HEAD `dff319988`) and re-confirmed on `main`
after PR #1250 merged. None of the 85 commits between those touch the
affected code paths.

---

### Symptom

```http
POST /v1/fluree/query
Content-Type: application/json

{ "from": ["a:main", "b:main"], "@context": {...}, "select": [...], "where": {...} }
```

Returns `400 { "error": "Missing ledger alias", "@type": "err:api/MissingLedger" }`.

Same for `from: [{"@id": "a"}, {"@id": "b"}]`. The
[datasets documentation](https://labs.flur.ee/docs/db/query/datasets) shows
`from: ["mydb:main", "otherdb:main"]` as a supported multi-default-graph
union form.

### Root cause

[fluree-db-server/src/routes/query.rs:167](fluree-db-server/src/routes/query.rs#L167)

```rust
if let Some(from) = body.get("from").and_then(|v| v.as_str()) {
    return Ok(from.to_string());
}
Err(ServerError::MissingLedger)
```

`.as_str()` returns `None` for arrays and objects → bails before the dataset
builder ever runs. The engine itself accepts array `from` just fine when
called via `query_from().jsonld()` directly.

---

### Symptom

A query with `from: A` + `fromNamed: { alias: B }` and a nested `select` that
hydrates properties through a ref. The hydrated `@id` comes back with the
**wrong prefix** — the prefix from `A`'s namespace dict applied to a SID
that was encoded against `B`'s dict.

Observed in production: `unit.@id` returned as `"categories/ml"` when the
source ledger stores it as `"units/ml"` (the application ledgers happened
to assign matching namespace codes to different prefixes).

### Root cause

`FromQueryBuilder::execute_formatted` at
[fluree-db-api/src/query/builder.rs:897-905](fluree-db-api/src/query/builder.rs#L897-L905):

```rust
let (spec, _) = parse_dataset_spec(json)?;
if let Some(alias) = spec.default_graphs.first() {
    let view = self.fluree.db(alias.identifier.as_str()).await?;
    Ok(result.format_async(view.as_graph_db_ref(), &format_config).await?)
} else {
    Err(ApiError::query("No default graph for formatting"))
}
```

It loads **only the first default-graph view** for formatting. Then
`format_async` builds an `IriCompactor` from that view's `namespaces()` map
([fluree-db-api/src/format/iri.rs:92-101](fluree-db-api/src/format/iri.rs#L92-L101)):

```rust
pub fn decode_sid(&self, sid: &Sid) -> Result<String> {
    ...
    let prefix = self.namespace_codes
        .get(&sid.namespace_code)
        .ok_or(FormatError::UnknownNamespace(sid.namespace_code))?;
    Ok(format!("{}{}", prefix, sid.name))
}
```

Each ledger numbers its namespaces independently at insert time. A SID from
the named ledger gets decoded against the *default* ledger's dict — either
erroring (`UnknownNamespace`) when the code doesn't exist, or **silently
mis-decoding** when the code exists but maps to a different prefix.

Compounding the problem: `query_connection` already built a complete
`DataSetDb` with all views and namespace dicts, but the dataset gets dropped
between query execution and formatting. The formatter then re-parses the
spec and reloads a single view from scratch via `self.fluree.db(...)`.

### Why existing tests don't catch it

The existing multi-ledger tests (e.g.
`query_connection_from_combined_datasets_selecting_subgraphs_depth_3`,
`query_connection_from_named_selecting_subgraphs_depth_3`) all use
`schema.org` + `wikidata.org` vocabularies across every test ledger. Those
get assigned identical namespace codes in every ledger, so the
single-snapshot lookup happens to land on the right prefix. With
application-specific vocabularies whose registration order differs between
ledgers, the bug surfaces.

---

### Symptom

```http
POST /v1/fluree/query
{ "fromNamed": { "alias": { "@id": "ledger:main" } }, ... }
```

Returns `400 "Missing ledger alias"`. If you bypass the HTTP layer and call
the engine directly via `query_from().jsonld().execute_formatted()`, you
get a different error: `"No default graph for formatting"`.

### Root cause

Two stacked bugs:

1. **HTTP layer.** Same `get_ledger_id` as Issue 1 — rejects because there's
   no `from` string. Never reaches the engine.
2. **Formatter layer.** Even bypassing the HTTP layer, the
   `else` branch at [builder.rs:903-905](fluree-db-api/src/query/builder.rs#L903-L905)
   bails out:

   ```rust
   } else {
       Err(ApiError::query("No default graph for formatting"))
   }
   ```

The sibling `DatasetQueryBuilder::execute_formatted` at
[builder.rs:562-573](fluree-db-api/src/query/builder.rs#L562-L573) handles
this correctly via `dataset.primary()` (which falls back to the first named
graph at
[view/dataset.rs:121-127](fluree-db-api/src/view/dataset.rs#L121-L127)).
The connection-scoped path never got that same fallback.

---

## Why these three go together

The connection path was built around single-ledger queries with a string
`from`. Multi-ledger support (`from` array, `fromNamed`, GRAPH alias) was
added in the engine and in `parse_dataset_spec` / the dataset builder,
where it works correctly. But neither the HTTP route's `get_ledger_id` nor
`FromQueryBuilder::execute_formatted` were updated to match.

The dataset builder produces a complete multi-view `DataSetDb` with policy,
aliases, and namespace dicts all in place. The execute step uses it
correctly. The formatter then throws it away and reloads a single view.
That's the architectural drift these three bugs share.

---

## Suggested fix direction

Independent fixes, smallest to largest:

1. **Issue 1 + Issue 3 layer 1 — HTTP dispatcher.** Relax
   [get_ledger_id](fluree-db-server/src/routes/query.rs#L153) to accept
   array/object `from` (pick the first identifier for the ledger-id slot, or
   defer the `MissingLedger` check until `requires_dataset_features` has run).

2. **Issue 3 layer 2 — formatter fallback.** Copy the `dataset.primary()`
   fallback from `DatasetQueryBuilder::execute_formatted` into
   `FromQueryBuilder::execute_formatted`.

3. **Issue 2 — cross-graph IRI decoding.** Restructure
   `FromQueryBuilder::execute_formatted` to keep the `DataSetDb` alive
   through formatting (don't drop it between `query_connection` and the
   format step). Then teach `format_results_async` / `IriCompactor` to
   accept the full dataset and build a multi-snapshot namespace map.

Note: the recent cross-ledger fix in commit
[`3a09e9d79`](https://github.com/fluree/db/commit/3a09e9d79) addresses the
analogous "use the wrong ledger's namespace dict" problem in cross-ledger
*policy* materialization. The query result formatter is the parallel piece
of work.

---

## Versions

- Reproduced on: `release/v4.0.4` (HEAD `dff319988`)
- Re-confirmed on: `main` post-PR #1250
- Toolchain: `cargo 1.93.0`, `rustc 1.93.0`

bplatz:
Closes #1259.

## Summary

Multi-ledger JSON-LD queries did not work correctly through the connection-scoped
`/v1/fluree/query` route. The reported issue covered three bugs; investigating it
surfaced a fourth, deeper one (cross-ledger hydration **expansion**). This PR fixes
all four, end to end, and adds regression coverage that reproduces each against
ledgers with **divergent** namespace vocabularies (the condition the existing tests
never exercised).

The work is split into four reviewable commits, smallest → largest:

| Commit | Fixes | Layer |
|--------|-------|-------|
| `correct multi-ledger JSON-LD query formatting` | Issue 2 (root @id), Issue 3-layer-2 | `fluree-db-api` formatter |
| `accept array/object from and fromNamed-only in query dispatcher` | Issue 1, Issue 3-layer-1 | `fluree-db-server` route |
| `route hydration roots to their home ledger` | cross-ledger expansion, slice 1 | `fluree-db-api` formatter |
| `expand cross-ledger refs against their home ledger` | cross-ledger expansion, slice 2 | `fluree-db-api` formatter |

## Background — why it broke

The connection path was built around single-ledger queries with a string `from`.
Multi-ledger support (`from` array, `fromNamed`, GRAPH alias) was wired into the
engine and `parse_dataset_spec` — where it works — but neither the HTTP route's
`get_ledger_id` nor the result **formatter** were updated to match. The formatter,
in particular, threw away the multi-view `DataSetDb` after execution and reloaded a
**single** ledger view, so every result IRI was decoded against one ledger's
namespace dictionary. Because each ledger numbers its namespaces independently, a
SID from another ledger silently decoded to the wrong prefix — or its properties
didn't resolve at all. The existing multi-ledger tests only passed because they
shared `schema.org`/`wikidata.org` across every ledger, so the colliding codes
happened to line up.

### 1. Formatter — wrong cross-graph `@id` (Issue 2, silent data corruption)

Multi-ledger queries already stamp result bindings as `Binding::IriMatch`, which
carries the canonical IRI decoded against the **source** ledger. The flat SELECT /
CONSTRUCT formatters use it; the **hydration** formatter discarded it and re-decoded
the foreign SID against the primary dict. Hydration now emits the root `@id` from
the canonical IRI (`compact_iri`, which is namespace-dict-independent), and the
expansion cache key gained the active ledger so a view-local SID can't collide
across ledgers.

### 2. Formatter — `fromNamed`-only formatting error (Issue 3-layer-2)

`FromQueryBuilder::execute_formatted` errored `"No default graph for formatting"`
when a query had only named graphs. It now falls back to the first named graph
(`default_graphs.first().or_else(|| named_graphs.first())`), matching the
default-then-named precedence the sibling `DatasetQueryBuilder` already used.

### 3. HTTP dispatcher — array/object `from` and `fromNamed`-only rejected (Issues 1 & 3-layer-1)

`get_ledger_id` only matched a bare string `from`, so array `from` (multi-default-graph
union), object `from` (time travel / graph selectors), and `fromNamed`-only queries
were rejected with `400 MissingLedger` before the dataset path could run. A new
`first_ledger_identifier` extracts a representative source from any supported shape;
it is used only for the conservative bearer scope check and span recording, while
per-ledger policy and routing are resolved downstream by `parse_dataset_spec`. A
request with no path/header/`from`/`fromNamed` is still `400`.

### 4. Formatter — cross-ledger hydration **expansion** (the deeper bug)

Fixing #2 made the cross-graph `@id` correct but did **not** make a foreign
subject's nested properties expand: hydration scanned only the primary snapshot
(plus a composite novelty overlay keyed by raw ledger-local SID), so a ref into
another ledger collapsed to a bare `{"@id": ...}` — and indexed foreign data was
unreachable entirely. The formatter is now **dataset-aware**:

- It keeps the `DataSetDb` alive (`query_connection_jsonld_returning_dataset`) and
  builds one `(GraphDbRef, IriCompactor, policy)` view per ledger.
- A hydration **root** is routed to its home ledger via the `IriMatch.ledger_alias`
  provenance, so its properties, predicate keys, and `@id` decode against the ledger
  that stores them — under that ledger's own policy enforcer, never the primary's.
- A **nested ref** is decoded to its canonical IRI in the current view, then
  re-encoded (`encode_iri_strict`) and expanded in the target ledger(s). A ref inside
  the default-graph union resolves across **all** default views, merging each
  contributing ledger's view of the subject (RDF merge, de-duplicating identical
  triples) under that ledger's own policy; a ref inside a named graph stays in that
  graph. Views whose namespace dict can't encode the IRI contribute nothing and are
  skipped without a scan.

Single-ledger queries are unchanged: one view, every lookup falls back to it. Flat
SELECT / CONSTRUCT / SPARQL already carry `IriMatch.iri` and stay on the primary
view.

## Design decisions

- **Per-ledger policy is preserved.** A foreign subject is read under its own view's
  policy enforcer — reusing the primary's would be both semantically wrong and a
  plausible leak path (policy checks and class caches are SID-local).
- **Default-graph union semantics.** A `from: [A, B]` default graph is a union, so a
  subject visible in both is merged from both (each under its own policy), rather
  than taking the first match (order-dependent, hides data) or guessing ownership
  from vocabulary.
- **Identity is the canonical IRI.** A `Sid` is view-local; cache and routing keys
  incorporate the active ledger / canonical IRI so one ledger's rendering can never
  be served for another's.

## Testing

New `fluree-db-api/tests/it_multi_ledger_formatting.rs` reproduces every issue
against ledgers with **divergent** entity/predicate vocabularies:

- `cross_graph_projection_iri_decode` — foreign-ledger root `@id` (Issue 2)
- `flat_select_cross_graph_iri_decode` — guards the already-correct flat path
- `from_named_only_formats` — `fromNamed`-only formats (Issue 3-layer-2)
- `cross_graph_root_properties_decode_in_home_ledger` — foreign root's divergent
  predicate decodes in its home ledger
- `cross_graph_hydration_expansion_divergent_vocab` — full depth-3
  `movie → book → author` chain across three ledgers that share **no** entity
  vocabulary

New `fluree-db-server/tests/multi_ledger_query_dispatch.rs` drives the Axum router
end to end for `from: [array]`, `fromNamed`-only, and the still-rejected no-ledger
case.

No regressions across `it_query_connection` (including the shared-vocab depth-3
hydration test), `it_query_dataset`, `it_named_graphs`, all `it_query_jsonld*` /
hydration / reverse / construct / typed / sparql suites, the format unit tests, and
server `integration` / `cross_ledger_http_integration` / `data_auth_integration`.

## Compatibility & risk

- Single-ledger and shared-vocab multi-ledger behavior is unchanged.
- Path- and header-scoped query routes are unaffected (path/header take priority
  before `from` is consulted).
- No `QueryResult` schema change, no core/index changes; the dataset-aware work is
  contained in `fluree-db-api`'s formatter + connection builder, plus the
  server dispatcher relaxation.
- Docs: `docs/api/endpoints.md` now documents the dataset `from` / `fromNamed` forms
  on the connection route.

## Out of scope / follow-ups

- An **indexed-data** integration variant (background-indexer harness per ledger):
  the cross-ledger routing path is identical for indexed vs novelty data (both go
  through `encode_iri_strict` + the target view's `db.range`), so the divergent-vocab
  test already exercises the logic; a dedicated indexed test is belt-and-suspenders.
- Cross-ledger cycle detection currently relies on the per-column depth budget; it
  cannot loop infinitely, but a pathological cross-ledger cycle within budget could
  re-expand rather than collapse. Tightening `visited` to canonical-IRI keying is a
  possible hardening follow-up.

### DIFF
--- a/fluree-db-api/src/format/hydration.rs
+++ b/fluree-db-api/src/format/hydration.rs
@@ -323,6 +323,54 @@ impl RootRef {
     }
 }
 
+/// Merge two JSON-LD objects describing the **same subject** seen through
+/// different default-graph ledgers (RDF merge of the union). Predicates present
+/// in only one are kept; predicates in both have their values unioned and
+/// de-duplicated (so identical triples, including `@id` / `@type`, collapse).
+fn merge_subject_objects(a: JsonValue, b: JsonValue) -> JsonValue {
+    let (JsonValue::Object(mut am), JsonValue::Object(bm)) = (a, b) else {
+        // Both are always objects in practice; if not, prefer the first.
+        return JsonValue::Null;
+    };
+    for (k, bv) in bm {
+        match am.remove(&k) {
+            None => {
+                am.insert(k, bv);
+            }
+            Some(av) => {
+                am.insert(k, merge_values(av, bv));
+            }
+        }
+    }
+    JsonValue::Object(am)
+}
+
+/// Union two property values into a de-duplicated value: a single survivor is
+/// returned bare, multiple as an array. Arrays on either side are flattened.
+fn merge_values(a: JsonValue, b: JsonValue) -> JsonValue {
+    let mut items: Vec<JsonValue> = Vec::new();
+    collect_values(a, &mut items);
+    collect_values(b, &mut items);
+    let mut out: Vec<JsonValue> = Vec::with_capacity(items.len());
+    for it in items {
+        if !out.contains(&it) {
+            out.push(it);
+        }
+    }
+    if out.len() == 1 {
+        out.into_iter().next().unwrap()
+    } else {
+        JsonValue::Array(out)
+    }
+}
+
+fn collect_values(v: JsonValue, out: &mut Vec<JsonValue>) {
+    match v {
+        JsonValue::Array(arr) => out.extend(arr),
+        other => out.push(other),
+    }
+}
+
 /// Format one hydration column for one solution row.
 ///
 /// Resolves the column's root (variable or IRI constant) into a `Sid` and
@@ -437,57 +485,111 @@ pub async fn format_async_dataset(
         ));
     }
 
-    // All dataset views — default graphs first, then named. Owned per-view
-    // compactors / policies live in parallel vecs that outlive the
-    // borrowed-state formatters.
-    let graph_views: Vec<&crate::view::GraphDb> = dataset
-        .default
-        .iter()
-        .chain(dataset.named.values())
-        .collect();
-    let compactors: Vec<IriCompactor> = graph_views
-        .iter()
-        .map(|g| IriCompactor::new(g.snapshot.namespaces(), context))
-        .collect();
-    let policies: Vec<Option<PolicyContext>> =
-        graph_views.iter().map(|g| g.policy().cloned()).collect();
-
-    let mut formatters: Vec<HydrationFormatter> = Vec::with_capacity(graph_views.len());
-    // Keyed by `GraphDb::ledger_id`, which is exactly what `IriMatch.ledger_alias`
-    // carries (set from the runtime `GraphRef.ledger_id`) — so root routing
-    // looks up the correct view.
-    let mut by_ledger: HashMap<Arc<str>, usize> = HashMap::new();
-    for (i, g) in graph_views.iter().enumerate() {
-        let mut view_db = g.as_graph_db_ref();
-        if let Some(t) = tracker {
-            view_db = view_db.with_tracker(t);
-        }
-        let formatter = HydrationFormatter::new(
-            view_db,
-            &compactors[i],
-            Arc::clone(&g.ledger_id),
-            config,
-            policies[i].as_ref(),
-            tracker,
-        );
-        // First occurrence wins (a ledger may appear as both default and named).
-        by_ledger.entry(Arc::clone(&g.ledger_id)).or_insert(i);
-        formatters.push(formatter);
-    }
-
-    // Primary = the dataset's primary view (first default, else first named),
-    // used for flat columns and as the fallback for unrouted roots.
-    let primary = dataset
-        .primary()
-        .and_then(|p| by_ledger.get(&p.ledger_id).copied())
-        .unwrap_or(0);
-
+    let ctx = DatasetCtx::build(dataset, context, config, tracker);
+    let formatters: Vec<HydrationFormatter> =
+        (0..ctx.views.len()).map(|i| ctx.formatter_for(i)).collect();
     let set = FormatterSet {
         formatters,
-        by_ledger,
-        primary,
+        by_ledger: ctx.by_ledger.clone(),
+        primary: ctx.primary,
     };
     run_hydration_rows(&set, result).await
+}
+
+/// One ledger's hydration view: its db, namespace-aware compactor, policy, and
+/// ledger id. Owned by [`DatasetCtx`]; formatters borrow from it.
+struct LedgerView<'a> {
+    db: GraphDbRef<'a>,
+    compactor: IriCompactor,
+    policy: Option<PolicyContext>,
+    ledger_id: Arc<str>,
+}
+
+/// Owns every dataset ledger's [`LedgerView`] and the routing metadata that
+/// cross-ledger hydration needs: which views back the default-graph union and
+/// which ledger maps to which view. Formatters borrow from this; it must
+/// outlive them.
+struct DatasetCtx<'a> {
+    views: Vec<LedgerView<'a>>,
+    /// `GraphDb::ledger_id` → index into `views` (matches `IriMatch.ledger_alias`).
+    by_ledger: HashMap<Arc<str>, usize>,
+    /// Indices of the default-graph views — the union scope for default-graph refs.
+    default_indices: Vec<usize>,
+    /// Primary view index (flat columns + unrouted-root fallback).
+    primary: usize,
+    typed: bool,
+    normalize_arrays: bool,
+    tracker: Option<&'a Tracker>,
+}
+
+impl<'a> DatasetCtx<'a> {
+    fn build(
+        dataset: &'a crate::view::DataSetDb,
+        context: &crate::ParsedContext,
+        config: &FormatterConfig,
+        tracker: Option<&'a Tracker>,
+    ) -> Self {
+        // Default graphs first (their indices form the union scope), then named.
+        let graph_views: Vec<&crate::view::GraphDb> = dataset
+            .default
+            .iter()
+            .chain(dataset.named.values())
+            .collect();
+        let default_count = dataset.default.len();
+
+        let mut views: Vec<LedgerView<'a>> = Vec::with_capacity(graph_views.len());
+        let mut by_ledger: HashMap<Arc<str>, usize> = HashMap::new();
+        let mut default_indices: Vec<usize> = Vec::with_capacity(default_count);
+        for (i, g) in graph_views.iter().enumerate() {
+            let mut view_db = g.as_graph_db_ref();
+            if let Some(t) = tracker {
+                view_db = view_db.with_tracker(t);
+            }
+            views.push(LedgerView {
+                db: view_db,
+                compactor: IriCompactor::new(g.snapshot.namespaces(), context),
+                policy: g.policy().cloned(),
+                ledger_id: Arc::clone(&g.ledger_id),
+            });
+            // First occurrence wins (a ledger may appear as both default and named).
+            by_ledger.entry(Arc::clone(&g.ledger_id)).or_insert(i);
+            if i < default_count {
+                default_indices.push(i);
+            }
+        }
+
+        let primary = dataset
+            .primary()
+            .and_then(|p| by_ledger.get(&p.ledger_id).copied())
+            .unwrap_or(0);
+
+        DatasetCtx {
+            views,
+            by_ledger,
+            default_indices,
+            primary,
+            typed: config.format == OutputFormat::TypedJson,
+            normalize_arrays: config.normalize_arrays,
+            tracker,
+        }
+    }
+
+    /// Build a formatter bound to one view, wired back to this context so it can
+    /// resolve cross-ledger refs.
+    fn formatter_for(&'a self, idx: usize) -> HydrationFormatter<'a> {
+        let view = &self.views[idx];
+        HydrationFormatter {
+            db: view.db,
+            compactor: &view.compactor,
+            active_ledger: Arc::clone(&view.ledger_id),
+            typed: self.typed,
+            normalize_arrays: self.normalize_arrays,
+            policy: view.policy.as_ref(),
+            tracker: self.tracker,
+            dataset: Some(self),
+            active_idx: idx,
+        }
+    }
 }
 
 /// A collection of per-ledger hydration formatters with root-routing.
@@ -628,6 +730,15 @@ struct HydrationFormatter<'a> {
     policy: Option<&'a PolicyContext>,
     /// Optional execution tracker for fuel/policy tracking.
     tracker: Option<&'a Tracker>,
+    /// Dataset context for cross-ledger reference resolution.
+    ///
+    /// `None` in single-ledger mode (refs stay within `db`). `Some` in dataset
+    /// mode: when expansion follows a ref into another ledger, the ref's SID is
+    /// decoded to its canonical IRI here, then re-encoded and expanded in the
+    /// target ledger's view (issue #1259).
+    dataset: Option<&'a DatasetCtx<'a>>,
+    /// This formatter's index into [`DatasetCtx::views`] (0 in single-ledger mode).
+    active_idx: usize,
 }
 
 impl<'a> HydrationFormatter<'a> {
@@ -647,6 +758,8 @@ impl<'a> HydrationFormatter<'a> {
             normalize_arrays: config.normalize_arrays,
             policy,
             tracker,
+            dataset: None,
+            active_idx: 0,
         }
     }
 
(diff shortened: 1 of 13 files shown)


Jackamus29:
Awesome! Thanks for picking up this issue, Brian.
I ran my reproduction tests against your fixes and there's still one test that fails.

Here's Claude's description of the remaining work with the failing test included
---

One gap I want to flag before this merges. The fix routes per `IriMatch.ledger_alias`, which works for `from: [array]` multi-default-graph union because `DatasetOperator::stamp_provenance` stamps bindings when multiple ledgers are active in the same scan. But for the `fromNamed: {alias: {"@id": ...}}` + `["graph", "<alias>", ...]` shape — which is the exact pattern the original reporter used — the inner scope has only one active graph, `DatasetOperator::needs_provenance` stays false, and bindings come out as plain `Binding::Sid`. With no `IriMatch` to consult, the formatter falls back to the primary view and the original wrong-IRI symptom returns.

### Repro

```rust
#[tokio::test]
async fn dataset_nested_projection_cross_graph_iri_resolution_via_connection() {

    let catalog0 = genesis_ledger(&fluree, "catalog:main");
    let _catalog = fluree.insert(catalog0, &json!({
        "@context": {"@vocab": "http://example.org/catalog/"},
        "@graph": [
            {"@id": "http://example.org/items/q1", "@type": "Item",
             "name": "Item One", "isbn": "0001"}
        ]
    })).await.unwrap().ledger;

    let lists0 = genesis_ledger(&fluree, "lists:main");
    let _lists = fluree.insert(lists0, &json!({
        "@context": {"@vocab": "http://example.org/lists/"},
        "@graph": [
            {"@id": "http://example.org/lists/summer", "@type": "List",
             "name": "Summer",
             "contains": [{"@id": "http://example.org/items/q1"}]}
        ]
    })).await.unwrap().ledger;

    let query = json!({
        "from": "catalog:main",
        "fromNamed": { "lists_g": { "@id": "lists:main" } },
        "@context": {
            "@vocab": "http://example.org/lists/",
            "catalog": "http://example.org/catalog/"
        },
        "select": {"?list": ["@id", "name",
            {"contains": ["@id", "catalog:name", "catalog:isbn"]}
        ]},
        "where": [["graph", "lists_g", {"@id": "?list", "@type": "List"}]]
    });

    let result = fluree.query_from().jsonld(&query).execute_formatted().await
        .expect("nested cross-graph projection should succeed");

    let s = result.to_string();
    assert!(
        s.contains("http://example.org/lists/summer"),
        "expected @id `http://example.org/lists/summer`; got: {s}"
    );
}
```

Observed on this branch:

```json
[{"@id": "http://example.org/items/summer"}]
```

The `summer` SID encoded against lists's namespace dict gets decoded against catalog's dict at format time — catalog's code-for-the-same-number happens to map to `http://example.org/items/`, hence the silent mis-decode. 

### Why this shape matters

This isn't an edge case — it's the canonical multi-ledger pattern the repo's own docs document at [`docs/query/datasets.md`](docs/query/datasets.md):


- [L87-99](docs/query/datasets.md#L87-L99): `fromNamed: {<alias>: {"@id": "ledger"}}` object form, with `["graph", "?g", ...]` clauses, in the headline example.
- [L257](docs/query/datasets.md#L257): object form is called the "preferred format" over the legacy array.
- [L249-253](docs/query/datasets.md#L249-L253): "Object keys in `fromNamed` serve as dataset-local aliases for use in GRAPH patterns" — the literal documented contract.
- [L289-314](docs/query/datasets.md#L289-L314): a worked cross-ledger example *specifically* shows using aliases to disambiguate two different `@id` ledgers under separate alias keys, queried with `["graph", "?g", ...]`. This is exactly the reporter's pattern.
- [L231-247](docs/query/datasets.md#L231-L247): combining a string `from` with aliased `fromNamed` + `["graph", "<alias>", ...]` — structurally identical to the reporter's query.


### Where the gap sits

`fluree-db-query/src/graph.rs::GraphOperator::execute_in_graph` reads batches from the inner pattern executor against a single active graph and merges them into the parent without stamping. The flake's predicate and subject Sids end up as plain `Binding::Sid` — no `ledger_alias` for the formatter to route by.

### Two fix shapes

I'm not sure which fits your architecture better:

1. **Stamp in `GraphOperator`.** Extend `GraphOperator::execute_in_graph` to run `stamp_provenance` on inner batches when the broader dataset spans multiple ledgers (detect via `ctx.dataset.iter().any(|g| g.ledger_id != first.ledger_id)`). `stamp_provenance` would need `pub(crate)` exposure. Single-ledger queries with GRAPH patterns short-circuit on the multi-ledger check.

2. **Recover provenance at format-time.** When a hydration root resolves to a plain `Binding::Sid` rather than `IriMatch`, look up which view in the `DataSetDb` actually contains the subject (try the active graph from the WHERE-clause GRAPH context if available) and route to that view. Doesn't touch the engine but adds a fallback to the formatter's routing.

(1) is what I prototyped in the work we'd been carrying for the same issue and it's the smaller change — engine produces consistent `IriMatch` provenance regardless of single-graph vs multi-graph scope, and the formatter contract stays uniform. Happy to push that as a delta on top of this PR if helpful.

Either way — the substantive piece is that the originally-reported shape isn't covered by the new test fixtures (which all use `from: [array]` / `fromNamed: [array]`). One additional regression test using `fromNamed: {alias: {"@id": ...}}` + `["graph", "<alias>", ...]` would catch it.

fluree-db-api/src/format/hydration.rs
```diff
@@ -603,6 +946,72 @@ impl<'a> HydrationFormatter<'a> {
...
+            }
+
+            // No view could resolve the subject → bare canonical @id.
+            match merged {
+                Some(v) => Ok(v),
+                None => Ok(json!({ "@id": self.compactor.compact_iri(&iri)? })),
+            }
+        }
+        .boxed()
+    }
```
aaj3f, line 1008: Even though `expand_ref()` tries to short-circuit for the most common single-ledger graph-crawl paths, the fact that all nested ref expansion now always paths through `expand_ref()` AND the fact that this short-circuit returns the `.boxed()` heap alloc means that the short-circuit for single-ledger graph crawls now always allocates 2 boxed futures where it previously only allocated one. I think the best solution is, at the call sites where `self.expand_ref()` currently gets called, we should be checking THERE whether `self.dataset.is_some()`. If `true`, then `expand_ref()` is fine. Where `false`, we should continue straight to `self.format_subject()`

fluree-db-api/src/format/hydration.rs
```diff
@@ -1235,15 +1626,31 @@ mod tests {
...
 
         // Different depths should produce different keys
         assert_ne!(key1, key2);
         // Same Sid + spec + depth should be equal
         assert_eq!(key1, key3);
+        // Different canonical IRI → different key
+        assert_ne!(key1, key4);
+        // Different active ledger → different key
+        assert_ne!(key1, key5);
     }
```
aaj3f, line 1679: The change to `CacheKey` to move from `(Sid,u64,usize)` to `(Arc<str>, Sid, u64, usize, Option<Arc<str>>)` seems right from a correctness perspective, but I'm just thinking about the performance impact of (1) hashing `active_ledger` str and also `Arc` cloning it on **every** `format_subject()` call. Not sure exactly how to solve for this though. If we had a `u32` ledger index to key on instead of `Arc<str>` this would be cheaper to the point of being effectively free

aaj3f (approved):
Agree w/ Jack's assessment above and have a few additional notes.

A connection query with `"from": ["A", "B"]` only seems to bearer-auth check on the first ledger. If the check passes on `A` the response can include `B` without a check on `B` (also seems true for `fromNamed`). The valid / correct pattern does already seem to exist within the path for multi-query e.g. `fluree-db-server/src/routes/query.rs:2847`:

```rust
        // Bearer ledger-scope enforcement — parity with single-query
        // /query and /query/:ledger. Unsigned bearer tokens may carry a
        // scope that limits which ledgers they can read; any envelope
        // referencing a ledger outside that scope is rejected with 404
        // (avoiding existence leak), matching the single-query response.
        if let Some(principal) = bearer.0.as_ref() {
            if !credential.is_signed() {
                for ledger_id in &distinct_ledgers {
                    if !principal.can_read(ledger_id) {
                        set_span_error_code(&span, "error:Forbidden");
                        return Err(ServerError::not_found("Ledger not found"));
                    }
                }
            }
        }
```

Some additional perf-related comments line-by-line.

Approving so you can merge when ready, but maybe worth considering a few of these first.

aaj3f (approved):

bplatz:
Good catch on the GRAPH-scope gap, @Jackamus29 — thanks for the detailed repro. Fixed in c5c45e075: `GraphOperator` now stamps inner GRAPH-pattern results with the named graph's home ledger when the dataset spans multiple ledgers (gated on `DataSet::spans_multiple_ledgers()`, with eager materialization so SIDs resolve before stamping). Added your `fromNamed:{alias:{@id}}` + `["graph","<alias>",..]` case as a regression test.

Separately, your repro surfaced a non-conformant `@vocab`-compacts-`@id` behavior in the formatter (independent of this PR) — filed as #1280.

bplatz:
@aaj3f addressed your perf + auth notes:

- Double-boxed future on single-ledger ref crawls — 6f5089412936e317717208a6c7c9a42dbe01b2e8
- `CacheKey` `Arc<str>` → view index (`active_idx`) — 6f5089412936e317717208a6c7c9a42dbe01b2e8
- Bearer scope now checked over every `from`/`fromNamed` ledger (404 on any out-of-scope), both connection + ledger-scoped handlers — 318d10bbc633fae968a251ed99f4affc3badbf8c

Tests added for all three.

Issue #1272, opened 2026-06-02
fix: incremental class-stat base-store load failure publishes wrong stats (rdf:type-star fold can serve wrong COUNT)


bplatz:
## Summary
Incremental Phase 3b loads the base V6 store for class attribution (`fluree-db-indexer/src/build/incremental.rs`, ~line 1897). On a `load_from_root_v6` **failure** it currently only `warn!`s and continues with `store_opt = None` — it does **not** `IncrementalAbort`. With `store = None`, every base `ClassStatEntry` is dropped (the subject sid64 resolves to 0 in the seed loop, ~line 2604), so **novelty-only (wrong) class stats get published into the root**.

This became a HEAD-correctness risk once the `lex_sorted_string_ids` provenance gate was removed (#1266): `count_plan_exec::try_type_star_pred_fold` now runs on any base index, so it would serve a **wrong `COUNT(*)`** at HEAD from those stale class stats.

## Severity
Narrow — requires a *transient* base-store load failure during an incremental build that includes `rdf:type` — but it is a correctness bug (silently wrong query result), and it precedes adding more class-stat fold paths (e.g. the bound-class fold), which would inherit the same exposure.

## Fix
Make the base-store load failure `return Err(IncrementalAbort(...))` → caller falls back to a full rebuild (which recomputes class + ref stats from scratch), mirroring the oversized-re-type gate (`incremental.rs` ~line 2086).

## Minor sibling (perf-only, not correctness)
An all-quiet increment (no class deltas *and* no subject-property records) fails the Phase 3b gate (~line 1840); `id_hook` finalize sets `classes: None` with no carry-forward else-branch, so class stats are wiped → the fold declines and falls back to a scan (correct, just unaccelerated until the next rebuild). Worth a carry-forward of the prior class stats when there are no deltas.

aaj3f:
Fixed by #1275 `ab2b5cb23` ("abort incremental to rebuild when class-stat store load fails"), on `main` since 2026-06-02. This landed alongside the #1266 work but the PR didn't cross-reference this issue.

Incremental Phase 3b now returns `IndexerError::IncrementalAbort` when the base V6 store fails to load for class attribution, instead of `warn!`-ing and continuing with `store_opt = None` (which dropped every base `ClassStatEntry` and published novelty-only class stats into the root). See the `load_from_root_v6` error arm in `fluree-db-indexer/src/build/incremental.rs`, which now aborts to a full rebuild.

[...]

Issue #1271, opened 2026-06-02
test: it_query_rdfs_class_repro flaky under concurrent nextest; bulk-import staging not cleaned


bplatz:
## Summary
Two related CI-reliability problems in the bulk-import integration tests (not correctness of the product code).

### 1. Flaky under concurrent execution
Under a full-workspace `cargo nextest run`, a **different** test in `it_query_rdfs_class_repro` fails on each run (observed: `sparql_select_class_prefixed`, `sparql_select_class_full_iris`, `sparql_select_property_works`, `jsonld_select_star_class`), while **all pass in isolation** (including `--all-features`). The pattern (random test, only under load) points to resource contention in the threaded `create().import().execute()` setup racing under parallel test execution.

### 2. Bulk-import staging not cleaned up
Tests using `.cleanup(false)` (e.g. `it_repro_bulk_filter_count`) leave import/index staging under the temp dir. Observed accumulation of ~17 GB, which exhausted disk and broke an unrelated full-run mid-way ("no space left on device").

## Suggested fixes
- Tighten test isolation under concurrent nextest (per-test temp dirs are used, but the threaded import + background indexer appear contention-sensitive); consider bounding import threads or serializing these tests.
- Ensure staging cleanup (avoid `cleanup(false)` where not essential; scope/limit temp usage) so repeated runs don't accumulate.

CI reliability only — no product behavior change.

Issue #1270, opened 2026-06-02
perf: metadata fold for class-to-class ref-edge COUNT(*)


bplatz:
## Summary
`SELECT (COUNT(*) AS ?c) WHERE { ?s a :A . ?s P ?o . ?o a :B }` is a single metadata lookup:
`classStat[g][:A][P].ref_classes[:B].count` — the count of `P` reference edges from instances of `:A` to instances of `:B`. Today it runs as a general 3-pattern join.

The per-target-class ref-edge counts (`ref_classes`) are now **current-state-exact on any index** (incremental + bulk) after #1266 and the ref-class re-attribution follow-up, but **no query path reads them yet**.

## Approach
Add a detection for this both-ends-type-constrained ref join (`?s a <ConstA> . ?s P ?o . ?o a <ConstB>` under `COUNT(*)`) plus a single `ref_classes` lookup. Reuse the class-stat reading from `count_plan_exec::try_type_star_pred_fold`. HEAD-only (`!ec.overlay`).

## Notes
- `COUNT(*)` / `COUNT(?o)` only (flake count); `COUNT(DISTINCT)` is a different statistic and out of scope.
- Specific shape, but clean and fully unblocked — good self-contained fast path.

Issue #1269, opened 2026-06-02
perf: extend overlay-aware fast paths to the remaining aggregate shapes


bplatz:
## Summary
The single-predicate and join COUNT fast paths now fold novelty under overlay/time-travel (parallel base scan + per-partition merge via bounded overlay cursors, or a touched-leaf novelty delta). The remaining aggregate fast paths are still **HEAD-only**: under uncommitted novelty they drop to the general pipeline (correct, but unaccelerated). Worth porting for live-write query latency.

## Shapes still HEAD-only
- `GROUP BY ?o (COUNT) ORDER BY DESC LIMIT` top-K (`fast_group_count_firsts::group_count_v6`)
- `COUNT(DISTINCT ?o)` / `COUNT(DISTINCT ?s)` (`fast_predicate_scalar_agg`)
- scalar `SUM` / `AVG` / `MIN` / `MAX`
- the POST numeric global-extent shortcut (`fast_count::predicate_post_global_extent`)
- string-prefix counts (`fast_count` / `binary_scan`)

## Approach
Reuse the established pattern: a bounded overlay cursor (`build_overlay_cursor_for_subject_range` / `build_overlay_cursor_for_predicate`) for the per-partition merged read, or a touched-leaf novelty delta (`count_predicate_overlay_delta`). Gate `to_t >= max_t` (at/above indexed head); time-travel below `max_t` defers (base replay).

Moderate value; correct today, so this is purely a latency improvement for write-heavy workloads.

aaj3f:
Sweep re-verification at `47e96b7b6`: the body's enumeration has gone stale in the good direction. Of the five listed shapes, `COUNT(DISTINCT ?o)` + SUM/AVG were overlay-routed in the #1275 merge itself, MIN/MAX gained an overlay-merged lane in #1449 (merged 7/8), and the string-prefix COUNT fast path has since been deleted entirely.

The real remainder is three shapes: (1) `GROUP BY ?o` COUNT top-K — `group_count_v6` still carries the "overlay delta merge not yet implemented" decline; (2) the directory-only distinct-lead folds, which are BASE-index-only by contract; (3) optionally the POST global-extent shortcut, which is HEAD-only by design since stats exclude novelty — arguably not a gap at all.

I'd suggest we either rewrite the body down to those shapes or close this in favor of a fresh, correctly-scoped issue — as it stands, anyone picking it up would spend the first day rediscovering that 60% of it is done.

PR #1275, opened 2026-06-02
perf: COUNT/aggregate fast paths — close the SPARQL aggregate benchmark gaps + overlay support

bplatz:
Based on #1265 

## Summary

Overhauls `COUNT` / `SUM` / aggregate execution with metadata-driven fast paths, asymmetric-seek joins, and parallel partitioned merges, collapsing the slow-spot aggregate queries on the DBLP "KG with associated data" benchmark (1.574B triples, r7a.4xlarge / 16-core). It then extends the **entire** count/filter family to fold uncommitted **novelty** and honor **time-travel** (parallel-under-overlay + a novelty-delta path), and lands the class-stat correctness fixes that make the metadata folds exact on incremental indexes — not just bulk imports.

## Results (DBLP, as of `c32b5cd8`)

Per-query latency on the targeted slow-spots dropped by up to ~18×:

| query | before (ms) | after (ms) | speedup |
|---|---|---|---|
| filter-many-results | 6,576 | 375 | ~18× |
| minus-join-3-star-2 | 4,725 | 329 | ~14× |
| union-constraint-large-join | 5,129 | 373 | ~14× |
| exists-join-3-star-2 | 4,673 | 330 | ~14× |
| optional-join-large-large | 4,455 | 372 | ~12× |
| filter-few-results | 2,761 | 370 | ~7× |

*(The overlay-parallel lanes and the bound-class fold landed after this run and were not re-benchmarked — they target live-write latency, not the bulk-imported benchmark, which is queried at HEAD.)*

### 1. HEAD COUNT/aggregate fast paths (the benchmark wins)
- **Metadata folds (zero scan):** single-predicate `COUNT(*)` and `COUNT(DISTINCT ?p)` from branch-manifest / leaflet counts; no-constraint `UNION` = Σ per-branch; rdf:type inner-star `COUNT(*)` from per-(class,property) stats; POST numeric-compare global min/max shortcut; **bound-class** property `COUNT(*)` (`?s a :C . ?s P ?o` → `classStat[:C][P]`).
- **Asymmetric seek** for skewed small+large joins: drive from the small predicate and leapfrog-seek the large one (`PsotSubjectSeek`) — join / OPTIONAL / MINUS / EXISTS.
- **Parallel partitioned merge** for both-large cases: partition the subject space at leaf boundaries and run the N-way merge per range across cores — inner-star, OPTIONAL, constrained-UNION, MINUS/EXISTS.
- **Parallel filter-scan:** `COUNT` with encoded FILTERs (`?s != ?o`, `isBlank`, `LANG`) and `SUM(?o cmp K)` / numeric-compare `COUNT`, split across leaf chunks.
- **GROUP BY ?o COUNT ORDER BY DESC LIMIT:** streaming run-length + bounded top-K heap.

### 2. Shared thread pool (multi-tenant)
Parallel count/scan work runs on the shared global thread pool instead of a per-query `thread::scope`, so N concurrent queries no longer each spawn a fresh K-thread fan-out — total worker threads stay bounded to ≈ cores.

### 3. Overlay / time-travel support
The whole count/filter family now folds uncommitted novelty and honors `to_t`:
- **Parallel-under-overlay:** per-partition bounded overlay cursors (parallel base scan + per-partition novelty merge via the proven `merge_overlay_into_batch`) — filter-scans and all four count-plan reducers (star / optional / union / minus-exists).
- **Novelty-delta:** single-predicate and no-constraint `UNION` `COUNT(*)` keep the instant manifest base count and rescan only the leaves novelty touches (`base_total − base(touched) + merged(touched)`).
- Gated to `to_t >= max_t` (at/above the indexed head); time-travel below the indexed head defers to the replay path.

### 4. Class-stat correctness (enables the folds on any index)
- **#1266:** incremental per-(class,property) datatype/language stats made current-state-exact (base-vs-net class attribution + base-index re-scan of re-typed subjects) → the rdf:type-star fold drops its bulk-import-only gate and runs on any base index.
- **Ref-class edge re-attribution on re-type** (outgoing edges from the same base scan + inbound edges via a reverse OPST lookup) plus a **rebuild gate** that falls back to a full rebuild for oversized re-type batches.

### 5. Supporting correctness fixes
- `isIRI`/`isURI` returns `false` for skolemized blank nodes.
- `COUNT(*)` over `p1/p2+` sequence paths counts join multiplicity (not distinct pairs).

## Correctness model

Every fast path **defers** (`Ok(None)` → general pipeline / cursor merge) when it can't apply: overlay/time-travel below the indexed head, absent or novelty-only predicates, non-root policy, multi-ledger, history mode, overlay-flake translation failures, or a 0 result. Correctness never depends on a fast path firing — the general path remains the source of truth.

## Testing
- `it_query_sparql_indexed.rs` (+2,880): parity tests for each HEAD fast path, plus overlay/novelty and parallel variants (>50k-row datasets confirm the parallel paths actually fire; novelty tests cover asserts, retracts, and time-travel).
- `it_indexing_stats.rs` (+801): #1266 and ref-class re-attribution regressions (retract, re-type, oversized-batch rebuild).
- `it_query_seq_path_count_repro.rs`, `it_import_class_stats.rs`.
- `cargo fmt --all --check` + `clippy --all-features` clean; full-workspace `nextest` green except known-environmental suites (external-catalog and container-based integration tests) and a pre-existing flaky test (see follow-ups).

## Follow-ups (filed)
- **#1268** — parallel paths block the calling async worker under concurrent load (`spawn_blocking` + `'static` refactor).
- **#1269** — extend overlay-aware fast paths to the remaining aggregate shapes (GROUP BY top-K, COUNT DISTINCT, scalar SUM/AVG/MIN/MAX).
- **#1270** — class-to-class ref-edge `COUNT(*)` fold (now-exact `ref_classes`).
- **#1271** — `it_query_rdfs_class_repro` flakiness under concurrent nextest + bulk-import staging cleanup.
- **#1272** — incremental class-stat base-store-load-failure correctness gap (`IncrementalAbort` → rebuild).

### DIFF
--- a/fluree-db-indexer/src/build/incremental.rs
+++ b/fluree-db-indexer/src/build/incremental.rs
@@ -1905,8 +1905,17 @@ pub async fn incremental_index(
                     {
                         Ok(s) => Some(Arc::new(s)),
                         Err(e) => {
-                            tracing::warn!(error = %e, "V6 store load for class attribution failed");
-                            None
+                            // We need the base store to seed base class entries and to
+                            // re-attribute re-typed subjects. Without it, base entries
+                            // resolve to sid64==0 and get dropped, publishing novelty-only
+                            // (wrong) class-property stats — which the rdf:type-star COUNT
+                            // fold then serves verbatim at HEAD. Abort to a full rebuild
+                            // (which loads its own data and recomputes class stats exactly),
+                            // mirroring the spatial-snapshot and re-type-workload gates above.
+                            return Err(IndexerError::IncrementalAbort(format!(
+                                "V6 store load for class attribution failed: {e}; \
+                                 falling back to full rebuild for correct class stats"
+                            )));
                         }
                     }
                 } else {
--- a/fluree-db-indexer/src/build/incremental.rs
+++ b/fluree-db-indexer/src/build/incremental.rs
@@ -2019,10 +2019,15 @@ pub async fn incremental_index(
                                 }
                             }
                             Err(e) => {
-                                tracing::warn!(
-                                    g_id = scan_g_id, error = %e,
-                                    "batched PSOT class lookup failed for graph"
-                                );
+                                // Loads the base class membership that all re-attribution
+                                // (datatype/lang AND ref-class) is computed against. A
+                                // failure here makes every downstream class delta unsound,
+                                // so abort to a full rebuild rather than warn-and-continue.
+                                return Err(IndexerError::IncrementalAbort(format!(
+                                    "Phase 3b base PSOT class lookup failed for graph \
+                                     {scan_g_id}: {e} (deferring class-stat recompute to \
+                                     full rebuild)"
+                                )));
                             }
                         }
                     }
@@ -2273,12 +2278,19 @@ pub async fn incremental_index(
                                 ) {
                                     Ok(m) => m,
                                     Err(e) => {
-                                        tracing::warn!(
-                                            g_id, error = %e,
-                                            "Phase 3b: base subject-property scan failed; \
-                                             re-typed subjects may keep stale class stats"
-                                        );
-                                        continue;
+                                        // This scan drives the datatype/language
+                                        // re-attribution that makes the now-ungated
+                                        // class-stat folds current-state-exact. A
+                                        // warn-and-continue here would silently leave
+                                        // re-typed subjects with stale per-(class,
+                                        // property) stats, so abort to a full rebuild
+                                        // (which recomputes class stats from the SPOT
+                                        // pass). Caught in lib.rs and routed to rebuild.
+                                        return Err(IndexerError::IncrementalAbort(format!(
+                                            "Phase 3b base subject-property scan failed \
+                                             for graph {g_id}: {e} (deferring class-stat \
+                                             recompute to full rebuild)"
+                                        )));
                                     }
                                 };
                             for &s_id in sids {
@@ -2462,11 +2474,15 @@ pub async fn incremental_index(
                                     }
                                 }
                                 Err(e) => {
-                                    tracing::warn!(
-                                        g_id, error = %e,
-                                        "Phase 3b: inbound ref scan failed; re-typed objects \
-                                         may keep stale ref-class stats"
-                                    );
+                                    // Case B inbound edges of re-typed objects. A miss here
+                                    // would silently leave stale ref-class edge stats; abort
+                                    // to a full rebuild for consistency with the other base
+                                    // re-attribution scans (the rebuild recomputes ref_classes
+                                    // exactly). Cheap insurance for the #1270 fold to come.
+                                    return Err(IndexerError::IncrementalAbort(format!(
+                                        "Phase 3b inbound ref scan failed for graph {g_id}: \
+                                         {e} (deferring class-stat recompute to full rebuild)"
+                                    )));
                                 }
                             }
                             // Drop edges touched this batch — the forward pass owns them.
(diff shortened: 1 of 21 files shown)


fluree-db-query/src/fast_path_common.rs
```diff
@@ -1702,6 +2125,259 @@ impl CursorSubjectCountStream {
...
+        let lo = leaves[i].first_key.s_id.as_u64();
+        let hi = if i + 1 < n {
+            leaves[i + 1].first_key.s_id.as_u64()
+        } else {
+            u64::MAX
+        };
+        // Ops with s_id in [lo, hi). Ops are sorted by s_id; advance the cursor.
+        while op_idx < ops.len() && ops[op_idx].s_id < lo {
+            op_idx += 1;
+        }
```
aaj3f, line 2327: I may be understanding this wrong, but it seems this would drop count consideration for overlay ops that are below the first index leaf's first subject.

It seems we'd need something like

```rust
// The first leaf's range starts at 0 so novelty ops below the predicate's
// first indexed subject (a low-id subject newly gaining this predicate) are
// not dropped — matching parallel_partition_count's bounds[0] == 0.
let lo = if i == 0 { 0 } else { leaves[i].first_key.s_id.as_u64() };
```

Maybe then also worth test / regression tests where a novelty assert at a subject id is below the indexed predicate's current minimum to ensure the count case isn't undercounting

fluree-db-indexer/src/build/incremental.rs
```diff
@@ -2122,6 +2231,328 @@ pub async fn incremental_index(
...
+                                    Ok(m) => m,
+                                    Err(e) => {
+                                        tracing::warn!(
+                                            g_id, error = %e,
+                                            "Phase 3b: base subject-property scan failed; \
+                                             re-typed subjects may keep stale class stats"
+                                        );
+                                        continue;
+                                    }
+                                };
```
aaj3f, line 2295: I'm trying to grok the degree to which this warn & continue should be acceptable. It seems maybe, given the PR description / commit promises etc that we should not simply warn but return `IndexerError::IncrementalAbort`?

There are other places where we swallow an error `batched_lookup_inbound_refs` and `batched_lookup_predicate_refs` but they seem to serve `ref_classes` which, as far as I can tell, aren't consumed meaningfully in the same way as this path is. But consistency-wise, perhaps they should go through the same "abort & rebuild" path as well?

aaj3f (approved):
It seems there are a few places where we have an overlay-parallel wrapper (`sum_star_join_overlay_parallel`, `sum_optional_join_overlay_parallel`, `try_modifier_intersect_overlay_parallel`, `try_union_constraint_overlay_parallel`) that calls `collect_resolved_overlay_ops` for every predicate and which walks all novelty flakes via `for_each_overlay_flake` before then dispatching `parallel_partition_count`. It's inside of that where the `total_rows < 50k || ncpu < 2` gate lives. For scenarios under that gate, then, the novelty collection is useless overhead, if I'm understanding correctly. It's also then re-done by the serial cursor-merge fallback. This seems like a double-effort perf scenario and a probably regression for the same queries before this PR

A cheap pre-gate check w/ a computed `total_rows` and a check like `available_parallelism() >= 2 && total_rows >= PARALLEL_STAR_MIN_ROWS`, if done before the `collect_resolved_overlay_ops` loop may solve for this

We may want some more test coverage for the claims (to validate them and avoid regressions):
  - The claim re: "parallel path fires" isn't tested. Given `parallel_partition_count` is serial when `ncpu < 2`, the parallel partition-boundary code would never be exercised
  - I don't see any tests that involve policy-hidden subjects such that there's a fold vs defer and where fold would over-count
  - Per the other comment I made re: novelty ids < index leaf ids, we likely want a test where overlay-delta novelty uses a low-id-subject < the leftmost index leaf's idea of "first"
  - We may want a test involving retraction where we retract an entire boundary leaf to zero etc to cover `merged < base`

Some additional in-line comments below but approving for whenever you're ready

fluree-db-query/src/fast_path_common.rs
bplatz: Fixed in dfd48b3b75d7b68718505041349ae2cdfac1845b: leaf 0 now ranges from 0, so a low-id existing subject newly gaining the predicate is counted (+ regression test).

fluree-db-indexer/src/build/incremental.rs
bplatz: Fixed in dfd48b3b75d7b68718505041349ae2cdfac1845b: all four Phase 3b base-scan failures now return IncrementalAbort (rebuild) instead of warn-and-continue.

PR #1277, opened 2026-06-02
fix: serve SPARQL CONSTRUCT/DESCRIBE as JSON-LD via content negotiation

### LINKED ISSUE #1274: SPARQL CONSTRUCT cannot be served as JSON-LD via Accept header
## Versions
- `fluree-server` built from `main` @ commit `e40f5560c` (PR #1264, 2026-06-01)

## Summary
The server tells the caller "CONSTRUCT queries only support JSON-LD output format" — yet there is no `Accept` header that selects JSON-LD for a SPARQL CONSTRUCT. The only Accept that produces a graph-formatted response is `application/rdf+xml`. `application/ld+json` and `application/json` both produce 400.

## Reproducer

Assuming a non-empty ledger:

```bash
# Fails: no Accept header (defaults to SparqlJson which CONSTRUCT rejects)
curl -i -X POST http://localhost:8090/v1/fluree/query/<ledger> \
  -d 'CONSTRUCT { ?s ?p ?o } WHERE { ?s ?p ?o } LIMIT 3'

# Fails: explicit JSON-LD Accept
curl -i -X POST http://localhost:8090/v1/fluree/query/<ledger> \
  -H 'Accept: application/ld+json' \
  -d 'CONSTRUCT { ?s ?p ?o } WHERE { ?s ?p ?o } LIMIT 3'

# Fails: application/json
curl -i -X POST http://localhost:8090/v1/fluree/query/<ledger> \
  -H 'Accept: application/json' \
  -d 'CONSTRUCT { ?s ?p ?o } WHERE { ?s ?p ?o } LIMIT 3'

# Works: rdf+xml
curl -i -X POST http://localhost:8090/v1/fluree/query/<ledger> \
  -H 'Accept: application/rdf+xml' \
  -d 'CONSTRUCT { ?s ?p ?o } WHERE { ?s ?p ?o } LIMIT 3'
```

## Expected
`Accept: application/ld+json` (and probably also no-Accept by default for CONSTRUCT) returns the constructed graph serialized as JSON-LD.

## Actual
All non-`rdf+xml` cases return:

```json
HTTP/1.1 400 Bad Request
{
  "error": "Format error: Invalid binding state: CONSTRUCT queries only support JSON-LD output format",
  "status": 400,
  "@type": "err:system/FormatError"
}
```

The error message is self-contradictory — it says "only JSON-LD is supported" while rejecting requests that ask for JSON-LD.

## Root cause (from a quick code read)
- `fluree-db-api/src/query/builder.rs:161` — default `FormatterConfig` for `QueryInput::Sparql` is `FormatterConfig::sparql_json()`.
- `fluree-db-api/src/format/mod.rs:124-131` — CONSTRUCT requires `config.format == OutputFormat::JsonLd`, else returns the error above.
- `fluree-db-server/src/extract/headers.rs` — has `wants_tsv`, `wants_csv`, `wants_sparql_results_xml`, `wants_agent_json`, `wants_rdf_xml` — **but no `wants_jsonld`**.
- `fluree-db-server/src/routes/query.rs:1846-1858` calls `.execute_formatted()` without overriding the default config for CONSTRUCT.

So the server never selects `OutputFormat::JsonLd` for SPARQL CONSTRUCT, regardless of headers.

## Suggested fix
In the SPARQL execute path, when the parsed AST is a CONSTRUCT (or DESCRIBE), force `FormatterConfig::jsonld()` as the default — overridable only by `Accept: application/rdf+xml`. Concretely, in `fluree-db-server/src/routes/query.rs` after the `parsed.ast` is computed:

```rust
let is_graph = matches!(
    parsed.ast.as_ref().map(|a| &a.body),
    Some(QueryBody::Construct(_) | QueryBody::Describe(_))
);
let config = if wants_rdf_xml {
    FormatterConfig::rdf_xml()
} else if is_graph {
    FormatterConfig::jsonld()
} else {
    FormatterConfig::sparql_json()
};
```

Also add a `wants_jsonld()` helper on `FlureeHeaders` that recognises `application/ld+json` and `application/json` so non-graph SELECT queries can still opt into JSON-LD when asked.

## Workaround used downstream
Until fixed, clients send `Accept: application/rdf+xml` for SPARQL CONSTRUCT and convert to JSON-LD locally with a standard RDF library. Functional but adds a round-trip through XML.

bplatz:
Closes #1274.

A SPARQL `CONSTRUCT`/`DESCRIBE` through `/v1/fluree/query[/{ledger}]` returned a self-contradictory `400 "CONSTRUCT queries only support JSON-LD output format"` for every `Accept` except `application/rdf+xml`. The SPARQL default output is SPARQL-results JSON, which the formatter then rejected for a graph result, so no header actually produced JSON-LD.

## Fix

**Formatter (root cause).** A graph has no binding-table rendering, so `format_results` / `format_results_async` now render `CONSTRUCT`/`DESCRIBE` as JSON-LD for any JSON-producing format instead of erroring. RDF/XML and the delimited formats are bytes formats handled earlier on the string path, so `application/rdf+xml` is unaffected. This one change fixes the server (all paths), the CLI, and library callers.

**Server content negotiation.** Graph JSON responses are now labelled `Content-Type: application/ld+json`, and a new `wants_jsonld()` helper lets `SELECT`/`ASK` opt into JSON-LD via the unambiguous `application/ld+json` type. A bare `application/json` intentionally does **not** flip `SELECT` off SPARQL-results JSON (it is the common default `Accept`).

**406 for mismatched formats.** A shared `is_graph_query()` guards the solution-table branches (SPARQL-results XML, CSV/TSV, AgentJson) so a graph query returns `406` instead of executing into a format `400`/`500` or a JSON-LD body mislabelled as AgentJson.

**Connection-scoped `/query`.** This route returns pre-formatted JSON only; the JSON family works (`CONSTRUCT` → JSON-LD, `SELECT` → results-JSON/JSON-LD, AgentJson), and the byte formats it does not negotiate (RDF/XML, SPARQL-results XML) now return `406` pointing at the ledger-scoped route rather than silently downgrading to JSON.

## Negotiation matrix (ledger-scoped `/query/{ledger}`)

| Query form | default / `application/json` | `application/ld+json` | `application/rdf+xml` | `sparql-results+json` | `csv` / `tsv` | `sparql-results+xml` | `agent+json` |
|---|---|---|---|---|---|---|---|
| `SELECT` / `ASK` | results JSON | JSON-LD | 406 | results JSON | CSV / TSV | results XML | AgentJson |
| `CONSTRUCT` / `DESCRIBE` | **JSON-LD** | JSON-LD | RDF/XML | JSON-LD | 406 | 406 | 406 |

## Tests & docs

New end-to-end server coverage in `sparql_construct_jsonld_accept.rs` (ledger- and connection-scoped: JSON-LD coercion, RDF/XML still works, `SELECT`+json stays results-JSON, mismatched formats → 406) plus an API-level CONSTRUCT default-path test. `docs/api/endpoints.md` documents the matrix and scopes byte-format negotiation to the ledger-scoped route. Also includes a hardware-sizing benchmark doc (CPU vs disk).

Based on the #1267 branch since it touches the same formatter/route files.

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -183,6 +183,7 @@
   - [Running with Docker](operations/docker.md)
   - [Storage modes (memory/file/AWS/IPFS)](operations/storage.md)
   - [Serverless storage choices](operations/serverless-storage.md)
+  - [Hardware sizing: CPU vs disk (benchmark)](operations/hardware-benchmarks.md)
   - [IPFS storage](operations/ipfs-storage.md)
   - [DynamoDB nameservice](operations/dynamodb-guide.md)
   - [Query peers and replication](operations/query-peers.md)
--- a/docs/api/endpoints.md
+++ b/docs/api/endpoints.md
@@ -947,11 +947,43 @@ LIMIT 100
 }
 ```
 
+**SPARQL output negotiation (`Accept` header):**
+
+The full byte-format negotiation below is available on the **ledger-scoped**
+[`POST /query/{ledger}`](#post-queryledger) route. The **connection-scoped**
+`POST /query` route (SPARQL with `FROM <ledger>`) returns pre-formatted JSON only
+— it supports the JSON family (JSON-LD, SPARQL-results JSON, AgentJson) but not the
+byte formats; RDF/XML, SPARQL-results XML, and CSV/TSV require the ledger-scoped
+route (see the [connection-scoped note](#connection-scoped-sparql-output) below).
+
+| Query form | Default (no/`*/*`/`application/json`) | `application/ld+json` | `application/rdf+xml` | `application/sparql-results+json` | `text/csv` / `text/tab-separated-values` | `application/sparql-results+xml` | `application/vnd.fluree.agent+json` |
+|---|---|---|---|---|---|---|---|
+| `SELECT` / `ASK` | SPARQL-results JSON | JSON-LD | **406** | SPARQL-results JSON | CSV / TSV | SPARQL-results XML | AgentJson |
+| `CONSTRUCT` / `DESCRIBE` | **JSON-LD** | JSON-LD | RDF/XML | JSON-LD | **406** | **406** | **406** |
+
+A `CONSTRUCT` / `DESCRIBE` produces an RDF graph, which has no solution/binding-table
+form, so it is always returned as **JSON-LD** (`Content-Type: application/ld+json`)
+unless `application/rdf+xml` is explicitly requested; the solution-table formats
+(SPARQL-results XML, CSV/TSV, AgentJson) are rejected with `406`. A `SELECT` / `ASK`
+defaults to SPARQL-results JSON and only switches to JSON-LD when `application/ld+json`
+is requested explicitly — a bare `application/json` keeps the SPARQL-results-JSON
+shape — and RDF/XML (a graph format) is rejected with `406`.
+
+<a name="connection-scoped-sparql-output"></a>
+> **Connection-scoped `POST /query` (SPARQL):** this route returns pre-formatted
+> JSON only. The JSON-family columns above apply (CONSTRUCT/DESCRIBE → JSON-LD;
+> SELECT/ASK → SPARQL-results JSON, or JSON-LD with `Accept: application/ld+json`;
+> AgentJson via `application/vnd.fluree.agent+json`, rejected `406` for graph
+> queries). CSV/TSV are rejected with `406`, and `application/rdf+xml` /
+> `application/sparql-results+xml` are **not** negotiated here — use
+> `POST /query/{ledger}` for those byte formats.
+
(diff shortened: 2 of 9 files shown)


aaj3f (approved):
Looks good.

It seems on the connection-scoped SPARQL JSON path we may do a double-parsing that we don't do, for example, on the ledger-scoped path (where we re-use `parsed` between format selection and the builder path. I know parsing is cheap compared to execution, but just calling it out.

bplatz:
@aaj3f Good catch on the double-parsing. The clean fix is broader than this handler — it's an API refactor to thread a parsed `SparqlAst` (plus original source text, needed for `SERVICE` body capture) through execution rather than re-parsing at each layer. Rather than expand the surface of this approved PR, I've recorded it as a follow-up: #1283.

PR #1282, opened 2026-06-03
fix: SPARQL 1.1 numeric arithmetic and expression ORDER BY correctness

bplatz:
## Summary

Fixes three SPARQL 1.1 correctness gaps in aggregate/analytic query evaluation. Each is an independent spec-compliance issue in core expression handling.

### 1. Mixed numeric-type arithmetic no longer errors

Arithmetic between two different XSD numeric types — e.g. `xsd:float(?x) / ?int` — failed with a type mismatch. XPath operator mapping requires numeric type promotion: the narrower operand is promoted to the wider type and the operation proceeds. Numeric operands carried as typed literals (string-backed `xsd:float`, the full integer family including values needing `BigInt`, `xsd:decimal`, and the `INF`/`-INF`/`NaN` float lexicals) are now normalized through the canonical core coercion layer before the operation, so promotion applies uniformly instead of falling through to an error.

### 2. Integer division yields `xsd:decimal`

`xsd:integer / xsd:integer` truncated to an integer (`10 / 4` → `2`) — a silent wrong-value deviation. Per XPath `op:numeric-divide`, integer division yields `xsd:decimal` (`10 / 4` → `2.5`). Division now returns a decimal, with precision capped at 34 significant digits (IEEE-754 decimal128, matching `AVG`) so recurring quotients stay bounded. As elsewhere, `xsd:decimal` serializes as a string to preserve exactness, while `xsd:double` remains a JSON number.

### 3. Expression-based `ORDER BY` is supported

`ORDER BY` with an expression rather than a bare variable was rejected at lowering time. SPARQL 1.1 §15.1 permits any expression as an order condition. Expression order keys (e.g. `ORDER BY DESC(?a / ?b)`) are now desugared to a synthetic per-solution bind evaluated after grouping/aggregation and before the sort, so the key may reference `GROUP BY` keys, aggregate outputs, and `SELECT`-expression aliases. Inline aggregates in the order condition (`ORDER BY DESC(COUNT(?x))`, SPARQL 1.1 §18.2.4.1) are hoisted into the grouping phase and deduplicated against any matching `SELECT` aggregate so the count is computed once.

## Scope / limitations

- Expression-based `ORDER BY` inside a subquery is intentionally rejected: the subquery operator sorts its projected output and has no stage to evaluate a synthetic sort key. Top-level queries are unaffected.
- `xsd:float` arithmetic uses the existing double representation (no distinct 32-bit float type), consistent with how `xsd:float`/`xsd:double` already collapse elsewhere in the engine.

### DIFF
--- a/fluree-db-api/tests/it_query_jsonld.rs
+++ b/fluree-db-api/tests/it_query_jsonld.rs
@@ -1882,6 +1882,77 @@ async fn jsonld_bind_arithmetic_in_select() {
     );
 }
 
+/// JSON-LD parity for SPARQL `sparql_integer_division_yields_decimal`.
+/// Per XPath op:numeric-divide, xsd:integer / xsd:integer yields xsd:decimal
+/// (10 / 4 = 2.5), not a truncated integer.
+#[tokio::test]
+async fn jsonld_integer_division_yields_decimal() {
+    let fluree = FlureeBuilder::memory().build_memory();
+    let ledger0 = genesis_ledger(&fluree, "parity/int-div:main");
+    let ctx = context_ex_schema();
+
+    let insert = json!({
+        "@context": ctx,
+        "@graph": [{"@id": "ex:s1", "ex:p": 1}]
+    });
+    let ledger = fluree
+        .insert(ledger0, &insert)
+        .await
+        .expect("insert")
+        .ledger;
+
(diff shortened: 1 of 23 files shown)


zonotope (approved):
🌞

PR #1284, opened 2026-06-03
fix: do not compact JSON-LD @id node identifiers against @vocab

### LINKED ISSUE #1280: JSON-LD formatter compacts @id against @vocab (should only use @base / explicit prefixes)
### Summary

When a query `@context` defines an `@vocab`, the JSON-LD result formatter compacts **`@id` node identifiers** against that `@vocab`, producing a bare term. Per JSON-LD 1.1, `@vocab` governs only properties, `@type` values, and values of `@type: @vocab` terms — it must **not** be used to compact node identifiers. Only `@base` and explicit term/prefix/`@id`-term mappings apply to `@id`.

### Example

Given a subject `http://example.org/lists/summer` and a query context:

```json
{ "@context": { "@vocab": "http://example.org/lists/" }, ... }
```

The formatter currently emits:

```json
{ "@id": "summer" }
```

`summer` round-trips only because it happens to match `@vocab`, but it is non-conformant: a compliant JSON-LD processor expanding `summer` as an `@id` would resolve it against `@base` (or leave it relative), **not** against `@vocab`. Expected output is the full IRI (or a compaction via an explicit prefix / `@base`-relative form):

```json
{ "@id": "http://example.org/lists/summer" }
```

### Root cause

The codebase already distinguishes the two compaction modes:

- `IriCompactor::compact_id` / `IriCompactor::compact_id_iri` — correct for `@id` (no `@vocab`).
- `IriCompactor::compact_vocab` / `compact_vocab_iri` / `compact_iri` / `compact_sid` — for predicates and `@type` (applies `@vocab`).

`fluree-db-api/src/format/construct.rs` already routes `@id` through the vocab=false path correctly. But the hydration / jsonld / typed formatters use the vocab-applying methods for `@id` positions:

- `fluree-db-api/src/format/hydration.rs` — root `@id` (`compact_iri` / `compact_sid`) and bare-`@id` fallback (`compact_iri`).
- `fluree-db-api/src/format/jsonld.rs` — `Binding::Sid` and `Binding::IriMatch` `@id` output (`compact_sid` / `compact_iri`).
- `fluree-db-api/src/format/typed.rs` — `Binding::Sid` and `Binding::IriMatch` `@id` output.

These should route through `compact_id` (the `@id`-position method) instead, matching `construct.rs`.

### Scope / impact

- Affects single-ledger and multi-ledger formatting equally; surfaced while reviewing the multi-ledger work in #1267 but is independent of it.
- Only manifests when the query `@context` sets `@vocab` and a result `@id` falls under that vocab namespace.
- Fix is localized (swap the compaction method at `@id` call sites) but needs care: `compact_sid` both decodes the SID and compacts, so the `@id` paths need a decode-then-`compact_id` variant.

### Suggested regression coverage

Add a formatter test asserting that with `@vocab` set, a result `@id` whose IRI falls under the vocab namespace is **not** reduced to a bare term (expect the full IRI, or an explicit-prefix / `@base`-relative compaction).

bplatz:
## Summary

Per JSON-LD 1.1, `@vocab` governs only property names, `@type` values, and `@type: @vocab` term values — it must not shorten node identifiers. The result formatters were compacting `@id` positions through the `@vocab` path, so with a query `@context` of `{"@vocab": "http://example.org/lists/"}` an `@id` of `http://example.org/lists/summer` was emitted as the bare term `"summer"`, which a conformant processor re-resolves against `@base`, not `@vocab`.

## Changes

- Add `IriCompactor::compact_id_sid` (decode then `@id`-position compaction: `@base` + explicit prefixes only, never `@vocab`).
- Route every `@id` / node-reference output site through the `compact_id_*` path:
  - `hydration.rs`: root `@id`, `expand_ref` bare-`@id` fallback, nested-ref `@id`, reverse-value `@id`, and the defensive Ref-in-literal `@id` fallbacks
  - `jsonld.rs`: flat-select `Binding::Sid` / `IriMatch`
  - `typed.rs`: `Binding::Sid` / `IriMatch`
  - `sparql.rs`: SPARQL JSON `uri` value (a `uri` is a node identifier too)
- Predicate keys, `@type` values, datatype `@type`, and reverse-predicate keys keep the `@vocab` path. `construct.rs` was already correct; `agent_json.rs` inherits the fix via its `jsonld` delegation.

## Behavior preserved

`compact_vocab` intentionally keeps `@base` as a fallback vocabulary so a bare `@type`/predicate in an `@base`-only context round-trips — unchanged. `@vocab: ""` still maps the vocabulary to `@base`.

## Tests & docs

- Unit tests (`iri`/`jsonld`/`typed`) plus an integration suite covering the hydration root and nested `@id`, flat-select references, combined `@base`+`@vocab` separation, and the `@vocab: ""` → `@base` round-trip.
- Documented the `@vocab`/`@base` compaction rules in `docs/concepts/iri-and-context.md`.

Closes: #1280

### DIFF
--- a/docs/concepts/iri-and-context.md
+++ b/docs/concepts/iri-and-context.md
@@ -139,6 +139,26 @@ Fluree includes many standard prefixes by default:
 }
 ```
 
+### How `@vocab` and `@base` compact results
+
+When a query `@context` sets `@vocab` and/or `@base`, Fluree follows JSON-LD 1.1
+about *which* IRIs each keyword may shorten:
+
+- **`@vocab`** governs **property names, `@type` values, and `@type: @vocab`
+  term values** only. A predicate or class IRI that falls under `@vocab` is
+  emitted as a bare term (e.g. `name`, `Person`).
+- **`@base`** governs **node identifiers** — `@id` and reference values. An
+  `@id` that falls under `@base` is emitted as a base-relative reference.
+- **`@id` is never shortened by `@vocab`.** A node identifier whose IRI falls
+  under the `@vocab` namespace keeps its full IRI (or an explicit-prefix /
+  `@base`-relative form) — it is *not* collapsed to a bare term, because a
+  compliant JSON-LD processor would re-resolve a bare `@id` against `@base`,
+  not `@vocab`.
+
+Explicit prefix/term mappings apply in both positions. As a special case,
+setting `@vocab: ""` maps the vocabulary to `@base`, so bare terms and
+base-relative identifiers share the same namespace.
+
 ## IRI Resolution Rules
 
 Fluree follows strict IRI resolution rules:
--- a/fluree-db-api/src/format/hydration.rs
+++ b/fluree-db-api/src/format/hydration.rs
@@ -794,15 +794,20 @@ impl<'a> HydrationFormatter<'a> {
             // The `@id` of THIS subject: prefer the canonical IRI from a
             // multi-ledger `IriMatch` root over decoding the SID against this
             // formatter's single-snapshot dict, which would mis-decode a SID
-            // that originated in a different ledger (issue #1259). `compact_iri`
+            // that originated in a different ledger (issue #1259). The compaction
             // is a pure context operation, independent of the namespace dict.
             // `root_iri` is only ever set for the root subject; nested refs
             // (recursive calls below) pass `None` and decode their SID against
             // the ledger whose flakes produced them.
+            //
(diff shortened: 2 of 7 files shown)


aaj3f (approved):
It is admittedly trivial but your description says

> Route every @id / node-reference output site through the compact_id_* path

Claude noted that this is not still true for CSV/TSV delimited output paths. Otherwise all good!

PR #1285, opened 2026-06-03
Agent JSON Output for `sparql_query` with `t`-Pinned Pagination

### LINKED ISSUE #1281: feat(mcp): emit Agent JSON from the MCP `sparql_query` tool (byte-budgeted, with resume pagination)
## Summary

The MCP `sparql_query` tool (`fluree-db-server/src/mcp/tools.rs`) returns verbose,
pretty-printed SPARQL-results JSON and then truncates it with **ad-hoc logic** (cap at 100
rows, shrink until under 50 000 chars, append a vague "use LIMIT/OFFSET" note). Fluree already
ships an **Agent JSON** output format built precisely for LLM/agent consumers — a schema-once
header, native JSON values, byte-budget truncation, and a ready-to-run `resume` query — and
it's already wired into the HTTP `/query` endpoint. **The MCP tool should use it instead of
re-implementing a worse truncation.**

## Motivation

- **Fewer tokens per result.** Agent JSON declares datatypes once in a `schema` header and
  emits native JSON values (no `{"type","value"}` wrapper per cell, no pretty-printing). In an
  LLM-in-the-loop, every tool result is re-sent to the model on each subsequent step, so result
  verbosity compounds. We hit provider tokens-per-minute limits in a real agent loop precisely
  because verbose results accumulate in context.
- **Real pagination.** On truncation, Agent JSON sets `hasMore: true` and emits a concrete
  `resume` SPARQL query (time-pinned with `@t:`, with `OFFSET`/`LIMIT`) — deterministic
  pagination instead of "use LIMIT/OFFSET" guesswork.
- **Consistency + less code.** The HTTP `/query` endpoint already supports Agent JSON; the MCP
  tool maintaining a parallel, inferior truncation path is an inconsistency and a maintenance
  burden.

## Current behavior (to replace)

`fluree-db-server/src/mcp/tools.rs` — `sparql_query` (~L73–204):

- Both execution branches call `.execute_formatted()` with **no formatter config** (~L116 identity/policy, ~L125 unrestricted) → default verbose JSON.
- Ad-hoc truncation: `const MAX_ROWS = 100; const MAX_CHARS = 50_000;` (L82–84); take first 100 rows, then drop 20% at a time until under 50 000 chars (L140–176); append:
  ```
  Note: Results truncated. Showing N of M total rows. Use SPARQL LIMIT and OFFSET clauses to paginate through results.
  ```

## Existing reference — the HTTP endpoint already does this

`fluree-db-server/src/routes/query.rs` (~L332–403) is the pattern to mirror:

```rust
if headers.wants_agent_json() {                                  // extract/headers.rs:246
    let parsed = fluree_db_sparql::parse_sparql(&sparql);        // count FROM clauses
    let from_count = parsed.ast.as_ref().and_then(/* dataset.default_graphs.len() */).unwrap_or(0);

    let agent_ctx = fluree_db_api::AgentJsonContext {
        sparql_text: Some(sparql.to_string()),
        from_count,
        iso_timestamp: Some(chrono::Utc::now().to_rfc3339_opts(chrono::SecondsFormat::Secs, true)),
        ..Default::default()
    };
    let mut config = fluree_db_api::FormatterConfig::agent_json().with_agent_json_context(agent_ctx);
    if let Some(max_bytes) = headers.max_bytes() {               // extract/headers.rs:259 (Fluree-Max-Bytes)
        config = config.with_max_bytes(max_bytes);
    }
    // ...
    state.fluree.query_from().sparql(&sparql).format(config).execute_formatted().await   // L403
}
```

The relevant API (in `fluree-db-api/src/format/config.rs` and `query/mod.rs`):
`FormatterConfig::agent_json()`, `.with_max_bytes(usize)`, `.with_agent_json_context(AgentJsonContext)`,
and `AgentJsonContext { sparql_text: Option<String>, from_count: usize, iso_timestamp: Option<String>, resume_limit: usize /* default 100 */ }`.
The formatter generates the `resume` query itself (single-`FROM` only) from `sparql_text` + `result.t`.

## Proposed change

In `sparql_query`, build an Agent JSON `FormatterConfig` and chain `.format(config)` before
`.execute_formatted()` on **both** branches; delete the `MAX_ROWS`/`MAX_CHARS` block.

```rust
let from_count = /* parse_sparql(&req.query) -> dataset.default_graphs.len(), as in routes/query.rs */;
let config = fluree_db_api::FormatterConfig::agent_json()
    .with_max_bytes(self.state.config.mcp_agent_json_max_bytes())   // see "byte budget" below
    .with_agent_json_context(fluree_db_api::AgentJsonContext {
        sparql_text: Some(req.query.clone()),
        from_count,
        iso_timestamp: Some(chrono::Utc::now().to_rfc3339_opts(chrono::SecondsFormat::Secs, true)),
        ..Default::default()
    });

// identity / policy branch:
view.query(self.state.fluree.as_ref()).sparql(&req.query).format(config.clone()).execute_formatted().await
// unrestricted branch:
self.state.fluree.graph(&req.ledger).query().sparql(&req.query).format(config).execute_formatted().await
```

Return the resulting envelope as the tool's text content. The formatter handles truncation,
`hasMore`, `resume`, and `t`/`iso` — so the ad-hoc truncation/notice code is removed.

## Byte budget — the one real design decision

MCP tool calls have **no per-request headers**, so there's no `Fluree-Max-Bytes` equivalent.
The budget must come from server config or a constant. Proposed:

- Add `agent_json_max_bytes` to `[server.mcp]` (`McpFileConfig`, `config_file.rs:156`) + CLI/env
  `--mcp-agent-json-max-bytes` / `FLUREE_MCP_AGENT_JSON_MAX_BYTES`, **default 32768 (32 KB)**.

(Alternative: a hardcoded constant with no knob. Lean toward the config knob with a sane default.)

## Tool description update

Update the `#[tool(description = …)]` for `sparql_query` so the model knows the result is an
Agent JSON envelope (`schema` / `rows` / `rowCount` / `hasMore` / `resume`) and that, when
`hasMore` is `true`, it should run the provided `resume` query for the next page.

## Open questions (let's agree before implementing)

1. **Default byte budget** = 32 KB? And configurable via `FLUREE_MCP_AGENT_JSON_MAX_BYTES`?
2. **Secondary row cap?** Rely solely on the byte budget, or keep a hard row guard too? (Lean: byte budget only.)
3. **Compact vs pretty** envelope text — compact = fewer tokens (lean compact).
4. **Output-shape change** is technically breaking for any MCP client parsing the current shape.
   Acceptable (LLM-facing, no known stable consumers)?
5. **Builder parity:** confirm `.format(config)` is available on both the policy `view.query()`
   builder and the `graph().query()` builder (it's used today on `query_from()` in the HTTP route).
6. **`get_data_model`** is also a large token sink — propose **out of scope** here (separate follow-up).

## Acceptance criteria

- [ ] `sparql_query` returns an Agent JSON envelope (`schema`, `rows`, `rowCount`, `t`/`iso`, `hasMore`).
- [ ] Truncation governed by a (config-driven, default 32 KB) byte budget; truncated single-`FROM` results include a `resume` query.
- [ ] Applies to both the identity/policy and unrestricted branches.
- [ ] Ad-hoc `MAX_ROWS`/`MAX_CHARS` logic removed.
- [ ] Tool description documents the envelope + `resume` pagination.
- [ ] New MCP-level test passes.

## Testing

- Add `fluree-db-server/tests/mcp_agent_json.rs` (or extend the MCP integration tests): assert
  envelope shape; byte-budget truncation sets `hasMore`; `resume` present for a single-`FROM`
  query; identity/policy branch formats identically.
- Existing coverage to reference: `fluree-db-api/tests/it_query_agent_json.rs` (formatter),
  `fluree-db-server/tests/integration.rs` (HTTP agent-json with tracking).

## Out of scope

- `get_data_model` Agent-JSON-ification (separate issue).
- MCP tracking/fuel reporting (HTTP-only today).

## References

- MCP tool: `fluree-db-server/src/mcp/tools.rs` (`sparql_query`, L73–204; truncation L82–84, L140–185)
- HTTP reference: `fluree-db-server/src/routes/query.rs` (L332–403)
- Header helpers: `fluree-db-server/src/extract/headers.rs` (`wants_agent_json` L246, `max_bytes` L259)
- Formatter config: `fluree-db-api/src/format/config.rs` (`agent_json` L217, `AgentJsonContext` L104–123)
- Result formatting: `fluree-db-api/src/query/mod.rs` (`to_agent_json[_with_config]` L182–212)
- Resume-query generation: `fluree-db-api/src/format/agent_json.rs` (`generate_resume_query` L271–321)
- MCP config: `fluree-db-server/src/config_file.rs` (`McpFileConfig` L156), `config.rs` (MCP args L603–618)

aaj3f:
## Summary

The MCP `sparql_query` tool (`fluree-db-server/src/mcp/tools.rs`) returned verbose,
pretty-printed SPARQL-results JSON and then truncated it with **ad-hoc logic** (cap at 100 rows,
shrink 20% at a time until under 50 000 chars, append a vague "use LIMIT/OFFSET" note). Fluree
already ships an **Agent JSON** output format built for LLM/agent consumers — a schema-once
header, native JSON values, byte-budget truncation — already wired into the HTTP `/query`
endpoint. This PR makes the MCP tool use it and deletes the parallel truncation path.

Closes #1281.

## Motivation

- **Fewer tokens per result.** Agent JSON declares datatypes once in a `schema` header and emits
  native JSON values (no `{"type","value"}` wrapper per cell, no pretty-printing). In an
  LLM-in-the-loop every tool result is re-sent to the model on each step, so result verbosity
  compounds and pushes against provider tokens-per-minute limits.
- **Deterministic truncation + pagination.** A configurable byte budget sets `hasMore`, and an
  explicit `t` lets agents page over a stable snapshot.
- **Consistency + less code.** One formatting path instead of two; one shared FROM-count helper.

## The pagination design decision (divergence from the issue)

The issue proposed mirroring the HTTP path's FROM-rewritten `resume` query
(`FROM <ledger@t:N> … OFFSET … LIMIT`). While implementing I verified that this does **not** fit
the MCP tool:

- The `@t:` time-travel suffix is parsed **only** on the FROM/dataset path
  (`dataset.rs::parse_ledger_id_time_travel`), used by connection-scoped `query_from()`.
- The MCP tool is **ledger-scoped** via a separate `ledger` param and runs through
  `graph(ledger)`, which uses `TimeSpec::Latest` and passes the raw string to
  `load_graph_db(ledger)` — it does **not** parse `@t:`. So `graph("mydb:main@t:5")` would look
  for a ledger literally named that and fail, and a FROM-rewritten resume query would not
  round-trip through this tool.

Instead, pagination uses an explicit optional **`t` request field**: the first call returns the
snapshot `t`; the agent re-runs with that same `t` (plus an increased `OFFSET`/`LIMIT`) to page
deterministically, routed through the existing `graph_at(ledger, AtT(t))` /
`db_at_t_with_policy(ledger, t, opts)`. `t` is always returned as the snapshot marker. No fragile
query-string rewriting.

## What changed

**`fluree-db-api`**
- Add `pub fn sparql_from_count(sparql) -> usize` next to `sparql_dataset_ledger_ids` and export
  it; use it for the connection-scoped agent-json site in `routes/query.rs` (removes a duplicated
  AST match).
- Add `DEFAULT_MCP_AGENT_JSON_MAX_BYTES` (32 KB) to `server_defaults`.

**`fluree-db-server` config**
- New `mcp_agent_json_max_bytes` knob: `--mcp-agent-json-max-bytes`,
  `FLUREE_MCP_AGENT_JSON_MAX_BYTES`, and `[server.mcp] agent_json_max_bytes` (default 32 KB). MCP
  tool calls carry no per-request headers (unlike the HTTP `Fluree-Max-Bytes`), so the budget is
  server-configured.

**`fluree-db-server/src/mcp/tools.rs`**
- Add optional `t` to `SparqlQueryRequest`.
- Extract a testable `execute_sparql_agent_json` helper (identity/unrestricted × latest/pinned),
  so the core can be tested without an rmcp `RequestContext`.
- Rewrite `sparql_query` to build an Agent JSON `FormatterConfig` (byte budget + context) and
  return the **compact** envelope; delete the `MAX_ROWS`/`MAX_CHARS` block.
- Update the tool description and `get_info` instructions to document the envelope
  (`schema`/`rows`/`rowCount`/`t`/`iso`/`hasMore`) and the `t` pagination protocol.

## Result shape

```json
{
  "schema": { "?s": "uri", "?name": "xsd:string" },
  "rows": [ { "?s": "ex:item0", "?name": "name-0" } ],
  "rowCount": 1,
  "t": 3,
  "iso": "2026-06-03T12:00:00Z",
  "hasMore": true,
  "message": "Response truncated due to size limit of 64 bytes. 1 of 20 total rows included."
}
```

On `hasMore`, the agent re-runs with the same `t` and an increased `OFFSET`/`LIMIT`.

## Testing

New `fluree-db-server/tests/mcp_agent_json.rs`:
- envelope shape (`schema`/`rows`/`rowCount`/`hasMore`, no `resume` key);
- byte-budget truncation sets `hasMore` with a partial page;
- **`t`-pinning determinism** — a pinned snapshot is unaffected by later writes while a latest
  query sees them;
- the identity/policy branch returns the same shape.

Verification (all green): `sparql_from_count` unit tests, the formatter `it_query_agent_json`
suite, the HTTP `integration sparql_query` tests (refactor regression), the server `config` tests
(incl. `arg_ids_match_server_config`), and `cargo clippy -p fluree-db-api -p fluree-db-server
--all-targets`.

## Out of scope / follow-ups

- `get_data_model` Agent-JSON-ification (separate issue).
- MCP tracking/fuel reporting (HTTP-only today).
- A future enhancement could let the tool surface a copy-paste resume request object now that
  pagination is `t`-driven.

### DIFF
--- a/fluree-db-server/src/mcp/tools.rs
+++ b/fluree-db-server/src/mcp/tools.rs
@@ -43,8 +43,8 @@ pub struct SparqlQueryRequest {
     #[schemars(
         description = "Optional transaction time `t` to pin the query to a specific historical \
         snapshot. Pass the `t` value returned in a previous result's envelope to paginate \
-        deterministically across calls: re-run with the same `t`, an ORDER BY clause, and a \
-        larger OFFSET. Omit to query the latest snapshot."
+        deterministically across calls: re-run with the same `t`, an ORDER BY clause, and OFFSET \
+        advanced by the previous result's `rowCount`. Omit to query the latest snapshot."
     )]
     pub t: Option<i64>,
 }
@@ -83,13 +83,14 @@ impl FlureeToolService {
     /// Execute a SPARQL query against a Fluree ledger
     #[tool(
         description = "Execute a SPARQL SELECT query against a Fluree ledger (the ledger is set by \
-        the `ledger` parameter, not a FROM clause). Returns a compact Agent JSON envelope: \
-        `schema` (per-variable datatypes, declared once), `rows` (objects with native JSON \
-        values), `rowCount`, `t` (the snapshot's transaction time), and `hasMore`. When `hasMore` \
-        is true the result was truncated to a byte budget — fetch the next page by re-running this \
-        tool with the SAME `t` (to stay on one snapshot) and your query extended with ORDER BY \
-        (for stable page boundaries) plus a larger OFFSET. Use get_data_model first to understand \
-        the schema before querying."
+        the `ledger` parameter, not a FROM clause; only SELECT is supported). Returns a compact \
+        Agent JSON envelope: `schema` (per-variable datatypes inferred from the rows on this page), \
+        `rows` (objects with native JSON values), `rowCount`, `t` (the snapshot's transaction \
+        time), and `hasMore`. When `hasMore` is true the result was truncated to a byte budget — \
+        fetch the next page by re-running this tool with the SAME `t` (to stay on one snapshot), \
+        keeping your ORDER BY (for stable page boundaries), and advancing OFFSET by the returned \
+        `rowCount` (the byte budget can return fewer rows than `LIMIT`). Use get_data_model first \
+        to understand the schema before querying."
     )]
     async fn sparql_query(
         &self,
@@ -165,6 +166,21 @@ impl FlureeToolService {
         t: Option<i64>,
         max_bytes: usize,
     ) -> Result<serde_json::Value, fluree_db_api::ApiError> {
+        // The Agent JSON envelope (and its byte budget) assume a SELECT solution table. ASK
+        // returns a bare boolean, and CONSTRUCT/DESCRIBE return a JSON-LD graph that bypasses
+        // `max_bytes` entirely — both contradict the advertised contract, so reject non-SELECT up
+        // front. A genuine parse error (`ast` is None) falls through to execution, which surfaces
+        // it as a normal query error.
+        if let Some(ast) = fluree_db_sparql::parse_sparql(query).ast {
+            if !matches!(ast.body, fluree_db_sparql::ast::QueryBody::Select(_)) {
+                return Err(fluree_db_api::ApiError::http(
+                    400,
+                    "sparql_query supports SELECT queries only; ASK, CONSTRUCT, DESCRIBE, and \
+                     UPDATE are not supported by this tool",
+                ));
+            }
+        }
+
         // No AgentJsonContext: the ledger (and optional `t`) fully scope a single-ledger query,
         // so `t` is always included and the formatter's FROM-based resume / multi-ledger advice
         // — which target the connection-scoped path and don't round-trip through this
@@ -223,15 +239,22 @@ impl FlureeToolService {
         if envelope.get("hasMore").and_then(serde_json::Value::as_bool) != Some(true) {
             return;
         }
-        // Read `t` before taking the mutable borrow below.
+        // Read `t` / `rowCount` before taking the mutable borrow below.
         let pin = match envelope.get("t").and_then(serde_json::Value::as_i64) {
             Some(t) => format!("t={t}"),
             None => "the same `t`".to_string(),
         };
+        // Advance by the returned rowCount, not by LIMIT: the byte budget can trip before the
+        // query's LIMIT, leaving rowCount < LIMIT, so advancing by LIMIT would skip dropped rows.
+        let advance = match envelope.get("rowCount").and_then(serde_json::Value::as_u64) {
+            Some(n) => format!("the {n} rows returned here"),
+            None => "the rows returned here".to_string(),
+        };
         let guidance = format!(
             " To page through the remaining rows deterministically, re-run sparql_query with \
-             {pin} and extend your query with ORDER BY (for stable page boundaries) and a larger \
-             OFFSET."
+             {pin} and the same ORDER BY (for stable page boundaries), then add {advance} to your \
+             current OFFSET (advance OFFSET by rowCount, not by LIMIT, since the byte budget can \
+             return fewer rows than requested)."
         );
         let Some(obj) = envelope.as_object_mut() else {
             return;
@@ -305,8 +328,8 @@ impl ServerHandler for FlureeToolService {
                    Use this FIRST to understand what data exists.\n\
                  - sparql_query: Execute a SPARQL SELECT query against a ledger. Returns a \
                    compact Agent JSON envelope (schema, rows, rowCount, t, hasMore). When hasMore \
-                   is true, paginate by re-running with the same `t`, an ORDER BY, and a larger \
-                   OFFSET."
+                   is true, paginate by re-running with the same `t`, an ORDER BY, and OFFSET \
+                   advanced by the returned rowCount."
                     .to_string(),
             ),
         }
--- a/fluree-db-server/src/mcp/tools.rs
+++ b/fluree-db-server/src/mcp/tools.rs
@@ -43,8 +43,8 @@ pub struct SparqlQueryRequest {
     #[schemars(
         description = "Optional transaction time `t` to pin the query to a specific historical \
         snapshot. Pass the `t` value returned in a previous result's envelope to paginate \
-        deterministically across calls (same `t` plus an increased SPARQL OFFSET/LIMIT). Omit to \
-        query the latest snapshot."
+        deterministically across calls: re-run with the same `t`, an ORDER BY clause, and a \
+        larger OFFSET. Omit to query the latest snapshot."
     )]
     pub t: Option<i64>,
 }
@@ -82,13 +82,14 @@ impl FlureeToolService {
 
     /// Execute a SPARQL query against a Fluree ledger
     #[tool(
-        description = "Execute a SPARQL SELECT query against a Fluree ledger. Returns a compact \
-        Agent JSON envelope: `schema` (per-variable datatypes, declared once), `rows` (objects \
-        with native JSON values), `rowCount`, `t` (the snapshot's transaction time), `iso` (the \
-        snapshot timestamp), and `hasMore`. When `hasMore` is true the result was truncated to a \
-        byte budget — fetch the next page by re-running this tool with the SAME `t` value (for a \
-        consistent snapshot) plus an increased SPARQL OFFSET/LIMIT. Use get_data_model first to \
-        understand the schema before querying."
+        description = "Execute a SPARQL SELECT query against a Fluree ledger (the ledger is set by \
+        the `ledger` parameter, not a FROM clause). Returns a compact Agent JSON envelope: \
+        `schema` (per-variable datatypes, declared once), `rows` (objects with native JSON \
+        values), `rowCount`, `t` (the snapshot's transaction time), and `hasMore`. When `hasMore` \
+        is true the result was truncated to a byte budget — fetch the next page by re-running this \
+        tool with the SAME `t` (to stay on one snapshot) and your query extended with ORDER BY \
+        (for stable page boundaries) plus a larger OFFSET. Use get_data_model first to understand \
+        the schema before querying."
     )]
     async fn sparql_query(
         &self,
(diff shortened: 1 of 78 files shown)


aaj3f:
Pushed `03962b91c` addressing the review feedback on the agent-json envelope. The adjustments:

**1. Dropped `iso` from the envelope.** It was `Utc::now()` — wall-clock query time — but the description called it "the snapshot timestamp," which is actively misleading for a `t`-pinned historical query. Since pagination keys on `t`, the field is removed rather than relabeled.

**2. No `AgentJsonContext` — the tool is always single-ledger.** Deriving the formatter context from the query's `FROM` clauses was wrong for this ledger-scoped tool: a query with ≥2 `FROM` clauses would drop `t` (the `from_count <= 1` gate) and, on truncation, emit "re-issue with `@iso:` on each FROM" advice that doesn't round-trip through the `graph(ledger)` path. Passing no context makes `t` always present and that multi-ledger advice unreachable, by construction. (No cross-ledger escape either way — the graph-scoped path runs against a single pre-loaded view, so a `FROM` can't pull in another ledger.) The description now also states the ledger is set by the `ledger` parameter, not a `FROM` clause.

**3. Pagination guidance now lives in the result.** Previously the `t` + `OFFSET` protocol was only in the tool description; on truncation the shared formatter emits a generic "truncated to N bytes" message. The tool now appends "re-run with `t=<t>`, `ORDER BY`, and a larger `OFFSET`" to the envelope's `message`, so the guidance is present exactly where an agent looks when it sees `hasMore`.

**4. Recommend `ORDER BY` for paging.** `t` fixes the snapshot but not row order, and SPARQL guarantees no order without `ORDER BY` — so disjoint, complete pages rest on stable scan order (an engine property, not a SPARQL one). The description, the `t`-field doc, and `get_info` now recommend it, and a new test pages a pinned `t` (`OFFSET 0/2 LIMIT 2`) asserting disjoint pages whose union is the full result.

**5. Docs.** Documented `--mcp-agent-json-max-bytes` / `FLUREE_MCP_AGENT_JSON_MAX_BYTES` / `[server.mcp] agent_json_max_bytes` in `docs/operations/configuration.md` (MCP table + env-var reference).

Note: `sparql_from_count` is no longer used by the MCP tool; it stays as the de-duplicated helper for the connection-scoped agent-json site in `routes/query.rs`.

Out of scope, flagged for a separate look: the shared formatter's `generate_resume_query` advances `OFFSET` by the page's `row_count` rather than cumulatively, which looks wrong for HTTP pages beyond the first.

Full test suite green, clippy clean.

fluree-db-server/src/mcp/tools.rs
```diff
@@ -203,6 +147,102 @@ impl FlureeToolService {
...
+            return;
+        }
+        // Read `t` before taking the mutable borrow below.
+        let pin = match envelope.get("t").and_then(serde_json::Value::as_i64) {
+            Some(t) => format!("t={t}"),
+            None => "the same `t`".to_string(),
+        };
+        let guidance = format!(
+            " To page through the remaining rows deterministically, re-run sparql_query with \
+             {pin} and extend your query with ORDER BY (for stable page boundaries) and a larger \
```
bplatz: This says "a larger OFFSET", but there are two truncation sources — the query's own `LIMIT` and the server byte budget. When the budget trips first, `rowCount < LIMIT` and `hasMore` is true; if the agent then advances OFFSET by its `LIMIT`, it silently skips the rows the budget dropped. Suggest telling it to advance OFFSET by the returned `rowCount` — which is exactly what `generate_resume_query` does on the HTTP path.

fluree-db-server/src/mcp/tools.rs
```diff
@@ -93,97 +106,28 @@ impl FlureeToolService {
...
-                    .graph(&req.ledger)
-                    .query()
-                    .sparql(&req.query)
-                    .execute_formatted()
-                    .await
-            }
-        };
+        let max_bytes = self.state.config.mcp_agent_json_max_bytes;
+        let result = self
+            .execute_sparql_agent_json(&req.ledger, &req.query, identity, req.t, max_bytes)
```
bplatz, line 116: No query-kind check here, so non-SELECT silently breaks the advertised envelope: ASK returns a bare boolean and CONSTRUCT/DESCRIBE return a JSON-LD graph (`format/mod.rs`) — the latter also bypasses `max_bytes` entirely, which defeats the budget this tool is built around. SELECT is already the documented contract (the `query` field says "Must be a valid SPARQL SELECT query"), so suggest gating to SELECT-only here and rejecting the rest with a clear error.

fluree-db-server/src/mcp/tools.rs
bplatz: To be precise: tell it to *add* the returned `rowCount` to the previous OFFSET, not set OFFSET to `rowCount` — otherwise a page that already started at e.g. OFFSET 100 is still ambiguous.

fluree-db-server/src/mcp/tools.rs
```diff
@@ -72,17 +82,20 @@ impl FlureeToolService {
 
     /// Execute a SPARQL query against a Fluree ledger
     #[tool(
-        description = "Execute a SPARQL query against a Fluree ledger. Returns JSON results as tuples matching SELECT variable order. Use get_data_model first to understand the schema before querying."
+        description = "Execute a SPARQL SELECT query against a Fluree ledger (the ledger is set by \
+        the `ledger` parameter, not a FROM clause). Returns a compact Agent JSON envelope: \
+        `schema` (per-variable datatypes, declared once), `rows` (objects with native JSON \
```
bplatz: Minor: `schema` is inferred only from the rows on the current page, so under pagination a datatype that first appears on a later page won't show here. Fine as-is (shared formatter behavior), just worth not implying the schema is complete across all pages.

bplatz (approved):
Great, some feedback inline in code.

fluree-db-server/src/mcp/tools.rs
aaj3f: Addressed in `02dd9f6`

fluree-db-server/src/mcp/tools.rs
aaj3f: Addressed in `02dd9f6`. Gated to `SELECT`-only

fluree-db-server/src/mcp/tools.rs
aaj3f: Addressed in `02dd9f6` to make unambiguous

PR #1286, opened 2026-06-03
feat: unify subquery execution with the top-level query pipeline

bplatz:
Subqueries (`{ SELECT … }`, SPARQL and JSON-LD) re-implemented a subset of the query pipeline and silently lacked several solution modifiers. This routes them through the **same** modifier tail and lowering as top-level queries, so a subquery inherits the full feature set — HAVING, post-aggregation binds, expression/aggregate ORDER BY, correct ORDER-BY-before-PROJECT ordering, and grouped-variable sort validation — from one shared code path.

Per-row correlation is untouched: the seeded WHERE still runs once per parent row and merges back.

### Queries that used to fail (or silently return wrong results) — now correct

- **HAVING in a subquery** — was a parse error, then a dropped clause:
  `{ SELECT ?p (COUNT(?o) AS ?c) WHERE { … } GROUP BY ?p HAVING (COUNT(?o) > 5) }`
- **Expression ORDER BY in a subquery** — was rejected at lowering:
  `{ SELECT ?x WHERE { … } ORDER BY DESC(?score / ?weight) LIMIT 10 }`
- **Aggregate ORDER BY in a subquery** (including when the aggregate is not selected) — was rejected:
  `{ SELECT ?p WHERE { … } GROUP BY ?p ORDER BY DESC(COUNT(?o)) }`
- **Post-aggregation SELECT expression** — was silently dropped:
  `{ SELECT ?p (CEIL(AVG(?o)) AS ?avg) WHERE { … } GROUP BY ?p }`
- **ORDER BY a non-projected variable** — was silently unordered (the sort ran after the projection removed the key):
  `{ SELECT ?x WHERE { ?x :p ?y } ORDER BY ?y }`
- **Aggregate over an expression in joined grouped sub-SELECTs** — was an HTTP 500 at plan time (blocks BSBM BI-5):
  two `{ SELECT ?f (AVG(xsd:float(?n)) AS ?x) … GROUP BY ?f }` joined on `?f`. The shared aggregate-input expression was deduplicated across subquery scopes, leaving the second's input variable unbound; it is now scoped per subquery.
- **GROUP BY / SELECT of a variable the pattern never binds** — was an HTTP 500 at plan time (blocks BSBM BI-4); per SPARQL 1.1 §11.4/§18.5 the variable is simply unbound (one group), now reported as such:
  `SELECT ?country ?product (AVG(?price) AS ?a) WHERE { … } GROUP BY ?country ?product` where `?country` is bound only by an enclosing join. (Not subquery-specific — fixed for top-level queries too.)

Grouped-variable sort attempts now produce a clean error instead of undefined ordering.

### DIFF
--- a/docs/api/endpoints.md
+++ b/docs/api/endpoints.md
@@ -150,8 +150,19 @@ WHERE {
 - `401 Unauthorized` - Authentication required
 - `403 Forbidden` - Not authorized for this ledger
 - `404 Not Found` - Ledger not found
+- `409 Conflict` - Optimistic-concurrency conflict that survived the server's
+  bounded reconcile-and-retry (rare; safe to retry the request)
 - `413 Payload Too Large` - Transaction exceeds size limit
 - `500 Internal Server Error` - Server error
+
+**Concurrency:** Writes to a single ledger are serialized by a per-ledger write
+lock (concurrent writes to *different* ledgers proceed in parallel). When a
+transaction is lowered/sequenced against a snapshot that is no longer the head
+by commit time — two writers racing on a first-time namespace code, or a cached
+writer state that fell behind the durable head — the server reconciles the
+cached state to the current head and re-tries (bounded, up to 16 attempts). A
+`409 Conflict` is returned only if the retry budget is exhausted; clients should
+treat it as retryable (distinct from a `400` bad request).
 
 **Examples:**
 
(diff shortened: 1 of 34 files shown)


aaj3f (approved):
Great find & fix. Maybe it's worth a future investigation (not necessarily in this PR) as to whether we still have a good amount of room to optimize on sub-query performance. Specifically, it seems that this correctness fix now does the following: we rebuild the entire subquery operator tree (`WHERE` planning and modifier tail) for every parent row, but row-to-row the only changes are seed changes. If there are highly-correlated subqueries, this would become perf wasting. If we could do a "build plan once, then re-seed per row" refactor, it would probably be a meaningful improvement. Again, though, nothing to block this PR on.

PR #1289, opened 2026-06-04
fix: persist the namespace split mode chosen by bulk import

bplatz:
## Problem

On a bulk import large/diverse enough to trip the namespace preflight, every concrete-IRI object/subject lookup silently returns zero rows (e.g. `?s a <ClassIRI>`, `?s <p> <obj>`), while predicate scans and `GROUP BY` are unaffected. `reindex` does not recover it.

This is triggered by first-chunk namespace diversity, not commit count or dataset scale: a single large import that crosses the prefix budget flips; a small one does not.

## Root cause

When the import preflight sees more than `NS_PREFLIGHT_BUDGET` (255) distinct namespace prefixes in the streamed input, it coarsens the forward-dict allocator to `NsSplitMode::HostPlusN(1)`, and the whole index is keyed with that mode. But three write-side records the read path consults hard-coded `NsSplitMode::default()` (`MostGranular`):

- the published FIR6 index root,
- the predicate-sid split,
- the genesis commit (recorded from `ns_registry`, which the allocator flip never touched).

At query time, `find_subject_id` splits the query IRI `MostGranular` against the coarse-keyed dictionary, so the namespace prefix is absent and the reverse lookup returns `None`. Scans are unaffected because they read the forward (s_id → IRI) dict and never re-split a query IRI. `reindex` reproduces the bug because it rebuilds from the genesis commit's wrong mode.

## Fix

Thread the allocator's actual split mode through `ChunkImportResult` → `IndexBuildInput` into the root and predicate sids, and mirror it into the genesis commit's `ns_registry` so a later reindex agrees. The mode is captured once (one relaxed atomic load) and copied — ingestion, namespace splitting, encoding, and index building are unchanged, so there is **no import-speed impact** (correctness only).

A regression test forces the streaming/preflight path with >255 namespaces and asserts a bound-IRI lookup resolves after both import and reindex; it returns 0 without this change.

## Migration

Ledgers imported before this fix that tripped the preflight still have the wrong split mode recorded and will keep returning empty results for concrete-IRI lookups. They need a **re-import** to recover — `reindex` alone is not sufficient, since the genesis commit carries the wrong mode.

### DIFF
--- a/fluree-db-api/src/import.rs
+++ b/fluree-db-api/src/import.rs
@@ -2089,6 +2089,8 @@ struct IndexBuildInput<'a> {
     /// User-defined named graph g_ids (>= 2) to build index segments for.
     /// Empty for single-graph imports.
     named_g_ids: Vec<u16>,
+    /// Actual split mode the import used; written into the root and predicate sids.
+    ns_split_mode: fluree_db_core::ns_encoding::NsSplitMode,
 }
 
 /// Run phases 2-6: import chunks, build indexes, upload to CAS, write V4 root, publish.
@@ -2147,6 +2149,7 @@ where
             collect_id_stats: config.collect_id_stats,
             prefix_map: &import_result.prefix_map,
             named_g_ids: import_result.named_g_ids,
+            ns_split_mode: import_result.ns_split_mode,
         };
         let index_result = build_and_upload(
             storage,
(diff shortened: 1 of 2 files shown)


zonotope (approved):
🐤

PR #1290, opened 2026-06-04
perf: share the namespace table via Arc to avoid per-query clones

bplatz:
A CPU flamegraph of the server under concurrent read load showed the single
largest CPU consumer was `IriCompactor::new` — the per-query setup used to
compact result IRIs — and that it was almost entirely a **deep clone of the
snapshot's namespace table, rebuilt on every query** (hashing and allocating
every entry). The actual per-IRI compaction work was negligible by comparison.

The namespace table is stable ledger metadata, so there's no reason to copy it
per request. This change:

- wraps `LedgerSnapshot.namespace_codes` in `Arc<HashMap<…>>` (with a
  `namespaces_arc()` accessor; the few mutation sites use `Arc::make_mut`), and
- has `IriCompactor` hold that `Arc`, so building a compactor per query is a
  refcount bump instead of a full map copy.

Result-formatting call sites now pass `snapshot.namespaces_arc()`.

Behaviour is unchanged — it's the same table, shared rather than copied — and
the format/query test suites pass. Also includes a small `#[ignore]` bench used
to locate where per-query CPU goes.

Stacked on `fix/ns-split-mode-not-persisted`.

### DIFF
--- a/fluree-db-core/src/db.rs
+++ b/fluree-db-core/src/db.rs
@@ -361,7 +361,7 @@ impl LedgerSnapshot {
     /// Returns an `Arc` clone (refcount bump, no copy) so per-query consumers
     /// such as the result-formatting `IriCompactor` can hold the stable table
     /// without deep-cloning it on every request.
-    pub fn namespaces_arc(&self) -> Arc<HashMap<u16, String>> {
+    pub fn shared_namespaces(&self) -> Arc<HashMap<u16, String>> {
         Arc::clone(&self.namespace_codes)
     }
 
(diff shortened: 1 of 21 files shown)


fluree-db-core/src/db.rs
```diff
@@ -351,6 +356,15 @@ impl LedgerSnapshot {
         &self.namespace_codes
     }
 
+    /// Get a cheap, shareable handle to the namespace table.
+    ///
+    /// Returns an `Arc` clone (refcount bump, no copy) so per-query consumers
+    /// such as the result-formatting `IriCompactor` can hold the stable table
+    /// without deep-cloning it on every request.
+    pub fn namespaces_arc(&self) -> Arc<HashMap<u16, String>> {
```
zonotope: I think this method should just be called `namespaces`.

zonotope (approved):
Looks good to me besides a minor suggestion you can take or leave.

fluree-db-core/src/db.rs
bplatz: @zonotope  There is already a `namespaces` but named it `shared_namespaces` instead which I think is better than the _arc

PR #1291, opened 2026-06-04
fix: preserve language-tagged literals that share a value (#1273)

### LINKED ISSUE #1273: Language-tagged literals with identical `@value` get deduplicated on insert, losing language variants
## Versions
- `fluree-server` built from `main` @ commit `e40f5560c` (PR #1264, 2026-06-01)
- Linux x86_64, local single-node, default file storage

## Summary
When a subject has multiple `skos:prefLabel` (or any `langString` predicate) triples whose **values are identical strings** but **language tags differ**, only one survives the insert. Per the docs (`api/endpoints.md:371` mentions a 6th element for the language tag), these should be distinct flakes.

## Reproducer

```bash
# 1. Create ledger
curl -X POST http://localhost:8090/v1/fluree/create \
  -d '{"ledger":"langbug:main"}'

# 2. Insert one concept with 3 prefLabels — en and fr share the value "animal"
curl -X POST http://localhost:8090/v1/fluree/insert/langbug:main \
  -d '{
    "@context": {"ex":"http://ex/","skos":"http://www.w3.org/2004/02/skos/core#"},
    "@id":"ex:animal", "@type":"skos:Concept",
    "skos:prefLabel":[
      {"@language":"en","@value":"animal"},
      {"@language":"es","@value":"animales"},
      {"@language":"fr","@value":"animal"}
    ]
  }'

# 3. Read back
curl -s -X POST http://localhost:8090/v1/fluree/query/langbug:main \
  -H 'Accept: application/sparql-results+json' \
  -d 'PREFIX skos:<http://www.w3.org/2004/02/skos/core#>
SELECT ?l (LANG(?l) AS ?lang) WHERE { ?s skos:prefLabel ?l }'
```

## Expected
Three rows: `("animal","en")`, `("animales","es")`, `("animal","fr")`.

## Actual
Two rows: `("animales","es")`, `("animal","fr")`. The `en` triple is silently dropped.

## Real-world impact
Ingesting the full GEMET thesaurus (`GEMET_en_es_fr.jsonld`, multilingual SKOS, [eionet.europa.eu](https://www.eionet.europa.eu/gemet/)):

| Lang | In source | After insert | Loss |
|---|---:|---:|---:|
| en | 5609 | 5321 | **−288** |
| es | 5388 | 5359 | −29 |
| fr | 5387 | 5387 | 0 |

EN bears the brunt because EN and FR share many literal forms (e.g. "animal", "lichen", many proper nouns and Latin loanwords). FR is fully preserved.

## Observed precedence
When `(s, p, "v")` exists with multiple `@language` tags, the surviving lang appears to follow a deterministic order — `fr > es > en` in our runs. This suggests dedup is performed before the language tag is taken into account, with insertion order or alphabetical order on lang code deciding the winner.

## Suspected area
Flake encoding / dictionary lookup for `langString` literals. The 6th element (language tag) documented at `docs/api/endpoints.md:371` should be part of the flake key, not metadata appended after dedup.

bplatz:
Inserting a subject with several language-tagged values whose strings are
identical but whose `@language` tags differ — e.g. `"animal"@en` and
`"animal"@fr` — silently dropped all but one variant. Ingesting a multilingual
SKOS thesaurus lost hundreds of labels, hitting English hardest because EN and
FR share many literal forms.

The root cause is fact identity. A language-tagged literal's identity includes
its language tag, and a list member's includes its position, so both live in the
flake metadata `m` alongside `(s, p, o, dt)`. The overlay-only read path
(`range::remove_stale_flakes`, used for committed-but-not-yet-indexed data and
genesis ledgers) keyed its set-semantics dedup on `(s, p, o, dt)` only and
dropped `m`. Two variants that shared a value therefore looked like one fact and
collapsed. The binary-scan path already keyed on the full identity; this just
brings the two into agreement with `Flake`'s own equality.

Adding `m` to that key fixes three failure modes from the same omission:

- distinct language variants sharing a value are preserved on insert;
- deleting one language variant no longer over-cancels the others (a retraction
  matches only the variant it names);
- `@list` members that repeat a value at different positions
  (`["a", "b", "a"]`) are no longer collapsed.

A second, related path is the binary overlay encoder. When a language tag has
not yet been added to the persisted language dictionary (a fresh tag in novelty
over an indexed ledger), it was encoded with a fixed fallback id, collapsing
every such tag to one value and mis-labelling the survivor. It now declines that
encoding and routes the flake through the raw-flake overlay path — which carries
the real metadata and dedups on full identity — mirroring how novelty-only
custom datatypes are already handled.

Finally, RDF export attaches the same novelty overlay so it can dump
committed-but-unindexed data. It previously discarded any flake it could not
encode into the columnar overlay (the new-tag language literals above, plus
vector/bigint/decimal and novelty-only custom datatypes), and it scanned with a
projection too narrow for the overlay merge. Export now emits those values
directly in every format (Turtle, N-Triples, N-Quads, JSON-LD), deriving the
language tag and datatype from the flake, and scans with the full identity
projection.

Closes #1273.

### DIFF
--- a/.fluree-memory/repo.ttl
+++ b/.fluree-memory/repo.ttl
@@ -1289,6 +1289,35 @@ mem:fact-01kj8k5nzpe59p5vp7zrehbrr9 a mem:Fact ;
     mem:branch "feature/memory" ;
     mem:createdAt "2026-02-24T19:49:14.358637+00:00"^^xsd:dateTime .
 
+mem:fact-01kta2t986nsjneban0hgbwzq7 a mem:Fact ;
+    mem:content "RDF term/fact identity in Fluree is (s, p, o, dt, m) where m=FlakeMeta{lang, i} — matching Flake::eq/hash. Any dedup/stale-removal key that omits m collapses distinct facts: langString variants sharing a value (\"animal\"@en vs @fr), and @list members repeating a value at different indices. Issue #1273 was range::remove_stale_flakes (overlay-only/genesis read path) keying on (s,p,o,dt) only; the binary-scan stale-removal (binary_scan.rs FactKeyRef) already included m. Including m also correctly scopes retraction cancellation per variant." ;
+    mem:tag "dedup" ;
+    mem:tag "flake-identity" ;
+    mem:tag "issue-1273" ;
+    mem:tag "langstring" ;
+    mem:tag "range" ;
+    mem:scope mem:repo ;
+    mem:artifactRef "fluree-db-core/src/flake.rs" ;
+    mem:artifactRef "fluree-db-core/src/range.rs" ;
+    mem:artifactRef "fluree-db-query/src/binary_scan.rs" ;
+    mem:branch "fix/langstring-novelty-dedup-1273" ;
+    mem:createdAt "2026-06-04T19:47:27.110421+00:00"^^xsd:dateTime ;
+    mem:rationale "Recurring class of bug: parallel dedup implementations diverging on whether the key includes FlakeMeta. Recall before touching any flake-dedup/stale-removal code." .
+
+mem:fact-01kta3264pg54aabvt97spk5qt a mem:Fact ;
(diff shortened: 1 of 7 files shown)


zonotope (approved):
🥼

PR #1292, opened 2026-06-05
fix: end-to-end named-graph support — TriG/SPARQL ingest and GRAPH queries

### LINKED ISSUE #1278: TriG parser rejects the W3C-compliant compact graph block syntax `<iri> { ... }`, only accepts the SPARQL-keyword form `GRAPH <iri> { ... }`
## Versions
- `fluree-server` built from `main` @ commit `e40f5560c` (PR #1264, merged 2026-06-01)
- Linux x86_64, local single-node, default file storage

## Summary
The TriG parser used by `/v1/fluree/upsert/{ledger}` accepts only the SPARQL-style graph block header (`GRAPH <iri> { ... }`). The W3C TriG 1.1 recommendation specifies that the `GRAPH` keyword is **optional** — `<iri> { ... }` alone is a valid named-graph block. Major tooling (rdflib, Apache Jena, RDF4J's TriGWriter) emits the compact form by default, so any client using a stock library to produce TriG hits a 400.

Spec reference: https://www.w3.org/TR/trig/#sec-graph
> Block ::= Labeled\_or\_Subject Wrapped\_or\_Empty\_Graph
> Labeled\_or\_Subject ::= ( 'GRAPH' )? Labeled\_or\_Subject\_Inner

(The `GRAPH` token is wrapped in optional brackets in the grammar.)

## Reproducer

```bash
# 1. Create a ledger
curl -X POST http://localhost:8090/v1/fluree/create \
  -d '{"ledger":"trigbug:main"}'

# 2. POST a compact-form TriG payload (this is what rdflib emits)
curl -i -X POST http://localhost:8090/v1/fluree/upsert/trigbug:main \
  -H 'Content-Type: application/trig' \
  --data-binary @- <<'EOF'

<urn:graph:test> {
    ex:alice ex:knows ex:bob .
}
EOF

# 3. The same payload with the explicit GRAPH keyword — works
curl -i -X POST http://localhost:8090/v1/fluree/upsert/trigbug:main \
  -H 'Content-Type: application/trig' \
  --data-binary @- <<'EOF'

GRAPH <urn:graph:test> {
    ex:alice ex:knows ex:bob .
}
EOF
```

## Expected
Both requests succeed (W3C TriG treats them as equivalent).

## Actual
The first (compact form) returns:

```json
HTTP/1.1 400 Bad Request
{
  "error": "Turtle parse error: Parse error at position <N>: expected predicate, found LBrace",
  "status": 400,
  "@type": "err:parse/TurtleParse"
}
```

The parser took `<urn:graph:test>` as the subject of a Turtle triple, expected a predicate, then found `{`.

The second (keyword form) returns 200.

## Real-world impact
Stock tooling defaults to the compact form:

- **rdflib** (Python) — `ConjunctiveGraph.serialize(format="trig")` / `Dataset.serialize(format="trig")` emit `<iri> { ... }`.
- **Apache Jena** — `RDFDataMgr.write(out, dataset, Lang.TRIG)` emits `<iri> { ... }` in default mode.
- **RDF4J TriGWriter** — emits `<iri> { ... }`.

Any caller that generates TriG via a library has to post-process the output to insert the `GRAPH` keyword (regex on top-level lines matching `^<[^>]+>\s*\{`). Workable but adds an annoying step and isn't discoverable from the error message.

## Suspected area / suggested fix
`fluree-db-transact/src/parse/trig_meta.rs`, `parse_statement()` around line 480:

```rust
TokenKind::KwGraph => {
    self.parse_graph_block()?;
}
TokenKind::Eof => {}
_ => {
    // Default graph triple
    self.parse_default_triple(start_pos)?;
}
```

The parser dispatches on the `GRAPH` keyword. When it sees an IRI or prefixed name followed by `{`, it should look ahead and route to `parse_graph_block()` as well (treating `GRAPH` as optional). Roughly:

```rust
TokenKind::KwGraph => {
    self.advance(); // consume GRAPH
    self.parse_graph_block()?;
}
TokenKind::Iri | TokenKind::IriEscaped(_) | TokenKind::PrefixedName | TokenKind::PrefixedNameNs
    if matches!(self.peek_next_significant().kind, TokenKind::LBrace) =>
{
    // Compact W3C TriG form: <iri> { ... }
    self.parse_graph_block()?;
}
TokenKind::Eof => {}
_ => {
    self.parse_default_triple(start_pos)?;
}
```

Lookahead is needed because a bare `<iri>` could also be the subject of a default-graph triple — the `{` disambiguates.

## Workaround used downstream
Until fixed, clients post-process the TriG body with a regex:

```python
body = re.sub(r'(?m)^(<[^>]+>)\s*\{', r'GRAPH \1 {', body)
```

before sending. Robust because `<…> {` at the start of a line cannot appear as part of a regular triple — only as a graph block header.

### LINKED ISSUE #1279: SPARQL `GRAPH <iri> { ... }` returns no results unless preceded by `FROM NAMED <iri>` — every named graph must be explicitly enumerated in the dataset clause
## Versions
- `fluree-server` built from `main` @ commit `e40f5560c` (PR #1264, merged 2026-06-01)
- Linux x86_64, local single-node, default file storage

## Summary
On the ledger-scoped `POST /v1/fluree/query/{ledger}` endpoint, a SPARQL query that walks named graphs via `GRAPH <iri> { ... }` returns empty bindings unless the same IRI is also listed in `FROM NAMED <iri>`. Most SPARQL endpoints (Virtuoso, Jena ARQ, RDF4J, Stardog, Blazegraph) expose all named graphs of the active database to `GRAPH` patterns by default — Fluree requires every name to be opted in by the caller.

This is technically permitted by [SPARQL 1.1 §13.2](https://www.w3.org/TR/sparql11-query/#namedGraphs) (the "active dataset" is server-defined), and the Fluree docs do hint at it. But every other engine I tested takes the friendlier default, and the failure mode here is silent: queries return `bindings: []` with no warning that the named graph wasn't part of the dataset. Many existing SPARQL clients and dashboards (YASGUI, Wikidata UI patterns, etc.) generate `GRAPH ?g { ?s ?p ?o }` without any `FROM NAMED`, expecting to discover all named graphs from the endpoint.

## Reproducer

```bash
# 1. Create + populate a ledger with one named graph
curl -X POST http://localhost:8090/v1/fluree/create \
  -d '{"ledger":"ngquirk:main"}'

cat > /tmp/probe.trig <<'EOF'
@prefix ex: <http://example.org/probe/> .
GRAPH <urn:probegraph> {
  ex:a ex:p ex:b .
  ex:a ex:q "hello" .
}
EOF
curl -X POST http://localhost:8090/v1/fluree/upsert/ngquirk:main \
  -H 'Content-Type: application/trig' --data-binary @/tmp/probe.trig

# 2. Query the named graph *without* FROM NAMED — returns empty
curl -s -X POST http://localhost:8090/v1/fluree/query/ngquirk:main \
  -H 'Accept: application/sparql-results+json' \
  -d 'SELECT ?s ?p ?o WHERE { GRAPH <urn:probegraph> { ?s ?p ?o } }'

# 3. Same query *with* FROM NAMED — returns the 2 triples
curl -s -X POST http://localhost:8090/v1/fluree/query/ngquirk:main \
  -H 'Accept: application/sparql-results+json' \
  -d 'SELECT ?s ?p ?o FROM NAMED <urn:probegraph>
      WHERE { GRAPH <urn:probegraph> { ?s ?p ?o } }'

# 4. SELECT DISTINCT ?g over GRAPH ?g — also empty (no graph discovery possible
#    without enumerating each name first)
curl -s -X POST http://localhost:8090/v1/fluree/query/ngquirk:main \
  -H 'Accept: application/sparql-results+json' \
  -d 'SELECT DISTINCT ?g WHERE { GRAPH ?g { ?s ?p ?o } }'
```

## Expected
Step 2 returns the same 2 bindings as step 3 — i.e., the named graphs of the ledger are visible to `GRAPH` patterns by default.

Step 4 discovers `<urn:probegraph>` (and any other named graph in the ledger).

## Actual
Step 2: `"bindings": []`
Step 3: 2 bindings
Step 4: `"bindings": []`

Same behavior on the connection-scoped `/v1/fluree/query` endpoint when targeting one ledger via `FROM <ledger>`.

## Real-world impact
- Federated SPARQL dashboards (YASGUI, RDF Browser, etc.) generate exploratory queries with `GRAPH ?g { ... }` and show nothing on Fluree — users see empty results and assume no data was published.
- A common pattern for multi-document ingestion is per-document named graphs (e.g. CAM publishes annotations under `<urn:cam:doc:NNN>`); aggregation queries like *"give me every concept tagged across all documents"* (`SELECT ?c WHERE { GRAPH ?g { ?c a skos:Concept } }`) silently return nothing.
- Forces every consumer to either (a) maintain its own list of named graph IRIs to inject via `FROM NAMED`, or (b) query the ledger's graph registry first (separate round trip) and then build the dataset clause dynamically.

## Suggested behavior
On `/query/{ledger}`, the active dataset should default to **the ledger's default graph + all named graphs registered in that ledger**. This matches the principle of least surprise and aligns with how Virtuoso/Jena/RDF4J behave on a single graph store.

Caller-supplied `FROM` / `FROM NAMED` would still narrow / override this default, preserving the existing escape hatch.

## Discovery / docs gap
The current behavior is referenced in `docs/api/endpoints.md` (the `/query/{ledger}` section) but as a feature, not a constraint. A short note like:

> Without an explicit `FROM NAMED`, the SPARQL named graphs of the ledger are **not** part of the active dataset — `GRAPH` patterns will return empty. To query a named graph, add `FROM NAMED <graph-iri>`.

would help users diagnose this without an issue.

## Workaround used downstream
Two patterns work:

```sparql
# Pattern A — `FROM` makes the named graph the default for triple patterns
SELECT ?s ?p ?o FROM <urn:probegraph> WHERE { ?s ?p ?o }

# Pattern B — `FROM NAMED` makes the graph addressable via GRAPH
SELECT ?s ?p ?o
FROM NAMED <urn:probegraph>
WHERE { GRAPH <urn:probegraph> { ?s ?p ?o } }
```

For dynamic dashboards we hit `GET /info/{ledger}` (or `/ledgers`) first, enumerate the named graphs, then inject all of them as `FROM NAMED` clauses before sending the user's query.

### LINKED ISSUE #1288: INSERT DATA { GRAPH <g> { … } } rejected with "expected subject in quad data"
## Summary

Fluree Core v4's SPARQL UPDATE parser rejects `INSERT DATA` operations that contain a `GRAPH` block. According to SPARQL 1.1 Update §3.1.1, the body of `INSERT DATA` is `QuadData`, defined as `'{' Quads '}'` where each `Quads` may be a plain `TriplesTemplate` *or* a `QuadsNotTriples` of the form `GRAPH VarOrIri '{' TriplesTemplate? '}'`. Core v4 only parses the first form and returns HTTP 400 on the second.

This blocks any RDF4J client (and by extension, anything built on Sesame 2.x / RDF4J's `SPARQLRepository`) configured to write to a named graph — the standard library serialises `connection.add(stmts, contextIRI)` as `INSERT DATA { GRAPH <ctx> { … } }`.

## Environment

- Fluree Core v4, server commit `12d739b` (local, port 8090)
- Endpoint: HTTP POST `/v1/fluree/update/<ledger>`, `Content-Type: application/sparql-update`

## Reproduction

```sparql
INSERT DATA {
  GRAPH <https://example.org/g/1> {
    <https://example.org/s/1> <https://example.org/p> "v" .
  }
}
```

Response:

```
HTTP/1.1 400 Bad Request

{
  "error": "expected subject in quad data",
  "status": 400,
  "@type": "err:parse/SparqlLower"
}
```

The same content, without `GRAPH`, succeeds:

```sparql
INSERT DATA {
  <https://example.org/s/1> <https://example.org/p> "v" .
}
```
→ HTTP 200, triples land in the default graph.

## Expected behaviour

Per SPARQL 1.1 Update §3.1.1:

```
QuadData         ::= '{' Quads '}'
Quads            ::= TriplesTemplate? ( QuadsNotTriples '.'? TriplesTemplate? )*
QuadsNotTriples  ::= 'GRAPH' VarOrIri '{' TriplesTemplate? '}'
```

The parser should accept the `QuadsNotTriples` alternative and insert the inner triples into the named graph `<g>`.

Symmetrical fix should apply to `DELETE DATA`, since both share the `QuadData` production.

## Impact

RDF4J's `SPARQLConnection.add(Iterable<Statement>, Resource... contexts)` unconditionally emits `INSERT DATA { GRAPH <ctx> { … } }` when a context is supplied. That is the path CAM's `SesameStore` walker takes when configured with `store.context=uri` (one named graph per processed document, the standard pattern for our content-augmentation pipelines).

Concrete consequence: we had to fall back to `store.context=default` on the Fluree V4 connector, meaning all documents merge into the ledger's default graph and document-level isolation is lost (no per-document graph queries, no per-document drop). See the `SesameStore-FlureeV4.xml` analysis engine in CAM where this is hardcoded as a workaround with a comment pointing at this bug.

The alternative (POST `/v1/fluree/upsert/<ledger>` with a TriG payload) *does* accept named graphs and works — but it bypasses the RDF4J client abstraction, requiring a custom store implementation rather than the stock `SPARQLConnection`.

## Suggested first-look areas

- `fluree-db-sparql/src/parse/query/update.rs` — the `parse_quad_data` function (or whatever covers `QuadData`): does it branch on the `GRAPH` keyword, or does it eagerly call into a triples-only template parser?
- Any companion code path for `DELETE DATA` should be checked at the same time.

Repro environment, ledger dump, and full HTTP capture available on request.

bplatz:
Completes named-graph support across the write and read paths.

## Write / ingest
- Accept the compact W3C TriG graph-block form `<iri> { ... }`, so named-graph data round-trips through `/upsert` and bulk import. Closes #1278.
- Accept `GRAPH` blocks in SPARQL `INSERT DATA` / `DELETE DATA` by making `QuadData` quad-aware and reusing the existing graph-delta lowering. Closes #1288.
- Scope flake dedup/cancellation by graph in the transaction accumulator, so the same `(s,p,o)` asserted in the default graph and a named graph within one transaction no longer collapses into a single flake.

## Read / query
- Resolve a ledger's registered named graphs in `GRAPH <iri>` and `GRAPH ?g` **without** requiring an explicit `FROM NAMED`. Closes #1279.

In single-db mode every graph of a ledger shares one snapshot/overlay partitioned by `g_id`, so resolution reads the snapshot's graph registry directly — no dataset is constructed, and queries that contain no `GRAPH` block are unaffected. The ledger alias keeps precedence for the default graph (so a registered graph whose IRI equals the ledger id can't shadow it), and reserved system graphs (`txn-meta`, `config`) stay private. Bound `GRAPH ?g` now also accepts IRI-typed terms, so `VALUES ?g { <iri> }` resolves named graphs.

This covers the ledger-scoped query endpoint, MCP `sparql_query`, and the library `query()` path. Connection-scoped `FROM <ledger>` GRAPH resolution is a planned follow-up.


fluree-db-query/src/context.rs
```diff
@@ -738,17 +738,50 @@ impl<'a> ExecutionContext<'a> {
         }
     }
 
+    /// Resolve a `GRAPH <iri>` to a user-registered named graph's `g_id`, or
+    /// `None` when a dataset scopes the named set (`FROM NAMED`), the IRI is
+    /// unregistered, or it names a reserved system graph. The ledger alias is
+    /// reserved for the default graph and never resolves here, even if a
+    /// registered graph shares the ledger's IRI.
+    pub fn single_db_user_graph_id(&self, iri: &str) -> Option<GraphId> {
+        if self.dataset.is_some() || iri == self.active_snapshot.ledger_id.as_str() {
```
zonotope, line 747: probably not for this pr, but explicit checks for dataset mode (or any other "queryable" type) suggests that we could benefit from a trait here.

zonotope (approved):
🧔🏾

fluree-db-query/src/context.rs
bplatz: Created issue #1304 to track it

PR #1293, opened 2026-06-05
perf: query read scaling, object→subject hash join, planner fixes, and EXPLAIN physical-plan tooling

### LINKED ISSUE #1287: Query planner: joining a property-path-produced variable with `skos:prefLabel` triggers a full predicate scan (88 s vs 1 ms)
## Summary

When a variable produced by a SPARQL property path (e.g. `skos:broader+`) is
joined with another triple pattern on a high-cardinality predicate like
`skos:prefLabel`, the query planner appears to evaluate the second triple
pattern independently (full scan of all subjects carrying that predicate)
instead of restricting it to the small set the property path already
produced. Wall-clock cost on a small ledger (~6 k concepts × multilingual
labels) jumps from a few ms to ~90 s.

Manually materialising the set with a `VALUES` clause runs the same join in
~1 ms, so the cost is purely a planner-heuristic problem, not an evaluation
ceiling.

## Environment

- Fluree Core v4, server commit `12d739b` (running locally on port 8090)
- Ledger: `gemet:main` (GEMET thesaurus loaded as SKOS, ~6 000 concepts,
  preferred labels in 30+ languages, ~150 000 prefLabel triples total)
- Query path: HTTP POST `/v1/fluree/query/gemet:main`,
  `Content-Type: application/sparql-query`,
  `Accept: application/sparql-results+json`

## Reproduction

The three queries below were run back-to-back against the same ledger
without warm-up between them. Starting subject is `gemet/concept/912`
(taxon "carnivore"), which has exactly **5 ancestors** under
`skos:broader+`.

### Query A — property path alone: **8 ms** ✅

```sparql
PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
SELECT ?b WHERE {
  <http://www.eionet.europa.eu/gemet/concept/912> skos:broader+ ?b .
}
```

Result: 5 bindings, `time=0.007945s`.

### Query B — same path + `skos:prefLabel` join: **88.9 s** ❌

```sparql
PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
SELECT ?b ?lbl WHERE {
  <http://www.eionet.europa.eu/gemet/concept/912> skos:broader+ ?b .
  ?b skos:prefLabel ?lbl .
}
```

Result: ~150 bindings (5 broaders × ~30 languages), `time=88.946929s`.

### Query C — manually materialised set: **1 ms** ✅

```sparql
PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
SELECT ?b ?lbl WHERE {
  VALUES ?b {
    <http://www.eionet.europa.eu/gemet/concept/8946>
    <http://www.eionet.europa.eu/gemet/concept/1798>
    <http://www.eionet.europa.eu/gemet/concept/450>
    <http://www.eionet.europa.eu/gemet/concept/5914>
    <http://www.eionet.europa.eu/gemet/concept/892>
  }
  ?b skos:prefLabel ?lbl .
}
```

Result: same ~150 bindings, `time=0.000980s`.

Queries B and C are semantically equivalent: B asks the planner to compute
the 5-element set then probe `skos:prefLabel`; C hands it the set
pre-computed. Wall-clock ratio is **>88 000×**.

## Additional data points

These were also tested to narrow down what triggers the explosion. All
runs use the same `<gemet/concept/912>` starting subject.

| WHERE pattern                                          | Time      |
| ------------------------------------------------------ | --------- |
| `?u skos:broader+ ?b`                                  | 0.006 s   |
| `?u skos:broader+ ?b . ?b rdfs:label ?lbl`             | 3.4 s     |
| `?u skos:broader+ ?b . ?b skos:prefLabel ?lbl`         | 88.9 s    |
| ... + `OPTIONAL { ?b rdf:type ?t . FILTER … }`         | 88.8 s    |
| ... + `OPTIONAL { ?b skos:broader ?bb }`               | 88.2 s    |
| ... + both OPTIONALs                                   | 89.3 s    |
| ... + `FILTER(LANG(?lbl)="en")`                        | 30.3 s    |
| `VALUES ?b { … 5 IRIs … } . ?b skos:prefLabel ?lbl`    | 0.001 s   |

Observations:

- The OPTIONAL clauses and the FILTER on `?broaderType` are essentially free.
- Replacing `skos:prefLabel` with `rdfs:label` (which has far fewer triples
  in this ledger — only ~10 on group-level concepts) drops the time from 88 s
  to 3.4 s, consistent with a full predicate scan whose cost is proportional
  to the number of triples bearing that predicate.
- A `FILTER(LANG(?lbl)="en")` after the join only saves the late-stage
  language filtering, not the scan itself — 30 s suggests the planner still
  loads all `skos:prefLabel` triples then filters.

## Workaround that does NOT work

Wrapping the property path in a sub-SELECT to try to force evaluation order:

```sparql
SELECT ?b ?lbl WHERE {
  { SELECT ?b WHERE { <…/912> skos:broader+ ?b . } }
  ?b skos:prefLabel ?lbl .
}
```

— still times out at `--max-time 60`. The planner appears to flatten the
sub-SELECT into the outer BGP rather than evaluating it first.

Bounded paths via `?` (`skos:broader/skos:broader?/…`) are rejected at parse
time:

```
SPARQL lowering error: Invalid property path: Sequence (/) steps must be
simple predicates, inverse simple predicates (^ex:p), transitive predicates
(ex:p+ or ex:p*), or alternatives of simple predicates ((ex:a|ex:b)); got
ZeroOrOne (?)
```

— a separate issue (or part of incomplete SPARQL 1.1 property path support).

## Expected behaviour

The planner should detect that:

1. A property-path BGP step produces a bounded variable, and
2. A subsequent triple pattern joins on that variable

and pick an iteration plan that probes the second triple by subject for each
materialised binding — i.e. an *index nested loop join driven by the
property-path output* — rather than a full predicate scan + hash/merge join.

Equivalently, the sub-SELECT form should be honoured: the planner should
evaluate the inner SELECT first when the outer BGP joins on its
projected variables.

## Impact

This breaks any CAM workflow over a Fluree-hosted SKOS thesaurus that
traverses ancestors/descendants and pulls labels — every "Ancestors"
inferencer call on a single document goes from milliseconds to ~90 s. The
client-side CAM connector currently raises the socket timeout to 180 s as
a band-aid, but the per-document cost remains unworkable for production.

## Suggested first-look areas in Fluree

- `fluree-db-sparql/src/parse/...` — sub-SELECT lowering: is the inner
  SELECT being flattened into the parent group?
- The query planner / optimiser: cardinality estimate for the output of a
  `+` path on a constant subject — is it being treated as "unknown / scan
  the world" instead of "small enumerated set"?

Happy to provide a dump of the GEMET ledger or run additional micro-bench
queries against it.

bplatz:
This PR improves query scaling and planner diagnostics for large read/query workloads.

- Replaces per-query deep snapshot clones with `Arc<LedgerSnapshot>` and switches cached ledger state to `RwLock`, allowing concurrent read snapshots to proceed in parallel while preserving exclusive write access.
- Adds a cost-based object-to-subject `HashJoinOperator` for large path joins, including planner tie-breaks that choose the driving side most likely to unlock a large hashable predicate scan.
- Extends `/fluree/explain` with compound-aware logical plans and planned physical operator trees, including fast-path labels, scan details, hash-join decisions, and expanded subquery bodies.
- Fixes planner estimates for anchored property paths so bounded closures drive joins instead of being treated like full-world scans.

Closes: #1287 

## Performance Notes

- Concurrent read benchmark improved from roughly flat scaling after 2 cores to about `8.05x` at 16 cores, with peak QPS reported around `107k`.
- BSBM-BI object-to-subject joins avoid repeated scattered OPST seeks by scanning the large predicate partition once and probing a hash table.
- Bowtie join ordering now prefers the side that unlocks the larger hash join, with the commit notes reporting a `110s -> 2.38s` improvement for the BI F2 shape.
- Anchored property paths such as `<concept/912> skos:broader+ ?b` are estimated as bounded closures rather than million-row scans.

## Explain Changes

`plan.logical` now preserves compound query structure while reflecting the same planner order used by execution.

`plan.physical` renders the planned operator tree without executing the query, including:

- `HashJoinOperator` vs `NestedLoopJoinOperator` selection
- hash-join cost inputs and rejection reasons
- fast-path operator labels and fallback edges
- dataset scan details such as predicate and planned index hints
- nested subquery bodies under `SubqueryBody`

### DIFF
--- a/docs/query/explain.md
+++ b/docs/query/explain.md
@@ -157,45 +157,74 @@ last among sources, minimizing data sent to the remote endpoint.
 
 ## Reading Explain Output
 
-The explain plan shows two sections: the original pattern order and the
-optimized order. Each pattern is annotated with its category and estimate.
+Explain returns a JSON object `{ "query": <echo>, "plan": { ... } }`. The
+`plan` object contains:
 
-Example output for a multi-pattern query:
+| Field                  | Meaning |
+| ---------------------- | ------- |
+| `optimization`         | `"reordered"`, `"unchanged"`, or `"none"` (no statistics) |
+| `statistics-available` | whether HLL statistics were available for cost estimation |
+| `statistics`           | summary stats (e.g. `total-flakes`) |
+| `logical`              | the **compound-aware** join order the planner produces (see below) |
+| `original`             | the query's triple patterns in original order, with selectivity inputs |
+| `optimized`            | the same triples in the planner's chosen order |
+| `execution-hints`      | specialized execution strategies the executor will use (see [Execution Hints](#execution-hints)) |
 
-```
-=== Query Optimization Explain (Generalized) ===
+The `original` and `optimized` arrays are a flattened, triple-level view. The
+`optimized` order is produced by the **same** `reorder_patterns` routine the
+executor uses, so the order shown is the order that runs — there is no separate
+explain-only ordering algorithm.
 
-Statistics available: yes
-Optimization: patterns reordered
+### The `logical` plan
 
---- Original Pattern Order ---
-  [1] ?s :age ?age | category=Source row_count=5000
-  [2] ?s :name ?name | category=Source row_count=10000
-  [3] FILTER((> ?age 25)) | category=Deferred
-  [4] OPTIONAL { ?s :email ?email } | category=Expander multiplier=1.00
+`logical` is the recommended view. Unlike `original`/`optimized` (which flatten
+all triples into one list), it preserves compound structure — `OPTIONAL`,
+`UNION`, `MINUS`, `EXISTS`, subqueries — and shows each node in the planner's
+chosen execution order. It is present even when statistics are unavailable
+(the planner falls back to heuristic estimates). Each node carries:
 
---- Optimized Pattern Order ---
-  [1] ?s :age ?age | category=Source row_count=5000
-  [2] FILTER((> ?age 25)) | category=Deferred
-  [3] ?s :name ?name | category=Source row_count=10000
-  [4] OPTIONAL { ?s :email ?email } | category=Expander multiplier=1.00
+- `kind`: `triple`, `optional`, `union`, `minus`, `exists`, `not-exists`,
+  `subquery`, `filter`, `bind`, `values`, `property-path`, `graph`, `service`,
+  or a search kind.
+- `category`: `source` (produces rows), `reducer` (shrinks), `expander`
+  (grows — e.g. `OPTIONAL`), or `deferred` (`FILTER`/`BIND`).
+- `estimate`: `{ "row-count": N }` for sources, `{ "multiplier": M }` for
+  reducers/expanders.
+- For triples, a `pattern` object (`subject`/`property`/`object`); for compound
+  nodes, a `patterns` array (or `branches` for `UNION`) of child nodes.
+
+Example (`?person :name ?name . OPTIONAL { ?person :email ?email }`):
+
+```jsonc
+"logical": [
+  { "kind": "triple",   "category": "source",
+    "estimate": { "row-count": 50 },
+    "pattern": { "subject": "?person", "property": "ex:name", "object": "?name" } },
+  { "kind": "optional", "category": "expander",
+    "estimate": { "multiplier": 1.0 },
+    "patterns": [
+      { "kind": "triple", "category": "source",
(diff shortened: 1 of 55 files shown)


zonotope (approved):
🎼

PR #1294, opened 2026-06-05
fix: reasoning derived-fact joins return empty or duplicate results

bplatz:
## Summary

OWL2-RL reasoning queries returned wrong results whenever a query joined a derived (inferred) fact with another pattern. The headline symptom — `?x a ub:Student . ?x ub:takesCourse ?c` returning zero rows when `Student` is only inferred — turned out to be three distinct bugs, all on the join path rather than in materialization itself. With these fixed, the full LUBM-1 query set returns correct answers.

## Root causes

**1. UNION drops variable-free existence rows.** A UNION branch that binds no variables (an all-bound existence check like `<g> a Student`) produces rows with zero columns. `UnionOperator::normalize_batch` rebuilt these via `Batch::new`, which infers the row count from the first column and so reports `len = 0` for a zero-column batch — silently dropping the rows. UNION was the only operator missing the `empty_schema_with_len` guard that the join, optional, and seed operators already use. Fixed by preserving the row count explicitly when the effective schema is empty.

**2. `owl2rl` double-reasons and duplicates.** `effective_with_hierarchy` auto-enables RDFS whenever a class hierarchy exists, so `reasoning: "owl2rl"` also ran the RDFS subclass/subproperty UNION rewrite. `<g> a Student` expanded to `UNION(Student, GraduateStudent, …)`, matching both the materialized `Student` derived fact and the base `GraduateStudent` branch, which double-counts under bag semantics (COUNT, aggregation, non-DISTINCT SELECT). Since materialization is a superset of what the rewrite produces, the query rewrite is now skipped entirely when OWL2-RL materialization is active — which also removes the redundant double-work.

**3. Derived facts dropped from joins on fully-indexed ledgers.** Derived facts live in the reasoning overlay as decoded `Binding::Sid` (they have no binary-store `s_id`). A binary scan late-materializes base rows to `Binding::EncodedSid` whenever it deems the store authoritative for decoding — which it does when the effective overlay reports epoch 0, i.e. a fully-indexed ledger with no novelty. `EncodedSid` and `Sid` never compare equal, so a join between a base row and a derived fact about the same entity silently yielded nothing. This affected 3+ pattern chains and joins on derived property facts (e.g. transitive `subOrganizationOf`). Fixed by forcing eager materialization when reasoning produced a derived-facts overlay (so every scan emits decoded `Sid`), and disabling the batched nested-loop-join path under eager materialization (it works in `EncodedSid` space and reads binary leaves directly, bypassing the overlay merge).

## Testing

New `it_reasoning_join_repro.rs` covers the derived-type join cases (2-pattern, 3+ pattern chains, derived transitive-property joins) on both the in-memory novelty path and the rebuilt/indexed binary-store path, plus the variable-free UNION path with reasoning disabled and the base-plus-derived dedup case. Verified against the standard LUBM-1 reference: all queries return correct answers.

## Known follow-up (performance)

The fix for (3) routes reasoning queries onto the per-row join path with eager decoding, which is correct but slow on multi-join LUBM queries: the batched fast-path reads binary leaves directly and structurally cannot see overlay-resident derived facts. Closing this gap needs an overlay-aware join (a hash join that scans the overlay once) or persisting the derived closure into the index so the fast path sees it natively. Tracked separately; out of scope for this PR.


fluree-db-core/src/ns_encoding.rs
```diff
@@ -138,6 +138,17 @@ pub fn builtin_prefix_trie() -> &'static PrefixTrie {
...
 /// check built-ins because they have no standard IRI delimiters.
 pub fn canonical_split(iri: &str, mode: NsSplitMode) -> (&str, &str) {
+    // Blank nodes always use the well-known `_:` prefix. Split it off first and
+    // unconditionally: the skolemized local part can contain colons (e.g.
+    // `_:fdb-<ledger>:<branch>-...`), which would otherwise make the generic
+    // opaque-IRI splitter cut at an embedded colon and drop the `_:` prefix,
+    // re-encoding the blank node into the EMPTY namespace (code 0). That breaks
+    // round-tripping: a blank node written via the dedicated blank-node code (10)
+    // would read back as a different Sid, so subject-bounded index lookups miss.
+    if let Some(rest) = iri.strip_prefix("_:") {
```
zonotope: super minor, but there should be a constant for blank node prefixes.

fluree-db-query/src/execute/reasoning_prep.rs
```diff
@@ -17,48 +17,72 @@ use std::sync::Arc;
...
-                subclass_of
-                    .entry(flake.s.clone())
-                    .or_default()
-                    .push(parent.clone());
-                return;
+
+    // Scan the full default-graph state (indexed + unindexed commits + overlay).
+    let db = GraphDbRef::new(snapshot, 0, overlay, to_t);
+
+    if let Ok(flakes) = db
```
zonotope: This will not propagate errors from the range call. Is that intentional?

This function returns an empty hierarchy on range error that will then disable reasoning in `effective_reasoning_modes` and the query will return no reasoned results. I _think_  it might be better to return an explicit error so that the user knows that there was an issue with reasoning instead of assuming that there were just no reasoned results.

fluree-db-query/src/join.rs
```diff
@@ -1084,9 +1084,16 @@ impl Operator for NestedLoopJoinOperator {
...
-            (self.batched_eligible || self.batched_object_eligible || self.batched_exists_eligible)
+        let use_batched = (self.batched_eligible || self.batched_object_eligible || self.batched_exists_eligible)
                 && ctx.binary_store.is_some()
+                // The batched path reads binary leaves directly and works in
+                // `EncodedSid` space (it even emits `EncodedSid` for newly-bound
+                // subject vars). Eager materialization means the query needs
+                // decoded `Binding::Sid` everywhere — notably reasoning queries,
+                // whose derived facts live in the overlay as `Sid` and must join
+                // with base rows. Fall back to the per-row path, which merges the
+                // overlay and honours eager decoding.
```
zonotope: The same logic applies for federated queries with no reasoning involved.

zonotope (approved):
🌸

fluree-db-core/src/ns_encoding.rs
bplatz: Addressed in d9f9c05a1a1189d280bde0d97f812ca723be9e2b

fluree-db-query/src/join.rs
bplatz: Addressed in d9f9c05a1a1189d280bde0d97f812ca723be9e2b

fluree-db-query/src/execute/reasoning_prep.rs
bplatz: Addressed in 43562a12902857637ce002222d0af3b7b29bdaeb

PR #1296, opened 2026-06-06
perf: Optimize query hot paths from CPU profiling

bplatz:
This branch improves query throughput by using CPU-profile-guided optimizations across serialization, binary index scanning, batched joins, filter evaluation, and late materialization.

The largest wins come from removing repeated loop-invariant work from hot paths, reducing binary scan work before decoding, and keeping values encoded longer so joins and DISTINCT process cheaper integer-backed bindings instead of materialized values.

## Key Changes

- Improved SPARQL XML serialization by reducing per-cell allocation, bulk-copying escaped XML spans, and lazily building IRI compaction fallback prefixes.
- Added binary scan diagnostics and a scan-stats endpoint to make CPU profiles easier to localize.
- Made binary leaflet loading projection-aware and cache-keyed by decoded column set, reducing unnecessary column decode and cache pressure.
- Added key-range pre-skip in the binary cursor so leaflets outside the bound key range are skipped before decode/filter work.
- Reduced batched join overhead by using faster hash maps, pre-sizing grouping maps, streaming matched rows directly, and emitting batched subject-join results column-wise.
- Added single-ledger encoded-ID fast paths for IRI equality/inequality filters, including constant-IRI comparisons.
- Hoisted repeated per-row filter work by memoizing constant operand resolution to internal subject IDs.
- Removed expensive namespace table deep clones by sharing namespace data in binary graph views.
- Kept inline integer/double values encoded through binary scans and joins, reducing Binding hash/eq/clone/drop cost through DISTINCT and join pipelines.
- Added direct decode for inline numeric encoded literals without rebuilding a graph view per row.
- Added regression coverage for indexed inline numerics through projection, DISTINCT, filter evaluation, SUM, and DATATYPE.
- Streamed JSON result serialization (JSON-LD, SPARQL JSON, Typed JSON, Agent JSON) directly into the output buffer, eliminating the per-cell serde_json::Value DOM and its second serialization pass on the hot query string path; output stays byte-identical (enforced by parity tests), with the object-heavy SPARQL JSON and Typed JSON formats formatting large result sets roughly 7–9× and ~4× faster respectively.
- Centralized the byte-identical JSON-writing primitives (string escaping, integer/float formatting) in one parity-tested module, and dropped Agent JSON's per-row double serialization under a byte budget by measuring row size from output-buffer length deltas instead of re-serializing each row.

## Performance Notes

CPU profiling showed repeated low-level costs were often caused by loop-invariant work happening per row or per call. This branch addresses those cases directly: namespace cloning, constant IRI-to-subject-ID lookup, graph-view reconstruction for inline numeric decode, avoidable binary leaf scans, and materializing numeric values too early.

On the benchmark workload used during development, these changes produced substantial throughput improvements across single-client and high-concurrency runs.

### DIFF
--- a/fluree-db-binary-index/src/read/column_types.rs
+++ b/fluree-db-binary-index/src/read/column_types.rs
@@ -332,30 +332,34 @@ impl BinaryFilter {
         first_key: &[u8; ORDERED_KEY_V2_SIZE],
         last_key: &[u8; ORDERED_KEY_V2_SIZE],
     ) -> bool {
-        // Sort-column sequence per order (`o_type` precedes `o_key`; `t` and the
-        // identity tiebreaks are never bound, so they are not consulted).
+        // Sort-column sequence per order (`o_type` precedes `o_key`; `t` is not
+        // part of the V3 ordered key and is never consulted).
         let fields: &[SortField] = match order {
             RunSortOrder::Spot => &[
                 SortField::S,
                 SortField::P,
                 SortField::OType,
                 SortField::OKey,
+                SortField::OI,
             ],
             RunSortOrder::Psot => &[
                 SortField::P,
                 SortField::S,
                 SortField::OType,
                 SortField::OKey,
+                SortField::OI,
             ],
             RunSortOrder::Post => &[
                 SortField::P,
                 SortField::OType,
                 SortField::OKey,
+                SortField::OI,
                 SortField::S,
             ],
             RunSortOrder::Opst => &[
                 SortField::OType,
                 SortField::OKey,
+                SortField::OI,
                 SortField::P,
                 SortField::S,
             ],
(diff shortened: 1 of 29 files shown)


fluree-db-binary-index/src/read/column_types.rs
```diff
@@ -305,6 +306,112 @@ impl BinaryFilter {
...
+    /// the same guards `skip_leaflet` relies on.
+    pub fn leaflet_out_of_range(
+        &self,
+        order: RunSortOrder,
+        first_key: &[u8; ORDERED_KEY_V2_SIZE],
+        last_key: &[u8; ORDERED_KEY_V2_SIZE],
+    ) -> bool {
+        // Sort-column sequence per order (`o_type` precedes `o_key`; `t` and the
+        // identity tiebreaks are never bound, so they are not consulted).
+        let fields: &[SortField] = match order {
```
zonotope, line 337: How are the extra `o` fields like `lang` and `i` taken into account? That probably doesn't matter for `spot` and `psot`, but they could change the predicted order of the `s` binding for `post` and `opst`.

zonotope (approved):
🏃🏾‍♂️

fluree-db-binary-index/src/read/column_types.rs
bplatz: OType encodes language now so it is not a missing field (for this reason, one less thing to have to look up) -- langString has its own OType range. This is new with our latest index version from several months ago, it wasn't always this way.

`i` was a gap and addressed in 978daa19eef33f39f10866a118a90b12f98acc75 in a way that it should not affect non-list data at all for performance but data containing lists might prune a bit less but now will be accurate.

PR #1297, opened 2026-06-07
perf: bounded-parallel prefetch and parallel rewrite for the incremental fold

bplatz:
## Summary

The incremental "backlog fold" — folding a backlog of accumulated commits into the persisted index in a single indexing run — was dominated by serial, one-commit-and-one-leaf-at-a-time S3 round-trips on a cold remote store, and on a large dataset could stall past the CAS fetch timeout mid-fold and abort to a full rebuild that does not fit in memory on a typical box. This branch turns each such hotspot into a bounded-parallel prefetch or prewarm drawing from a single shared S3 in-flight budget, and parallelizes the per-leaf copy-on-write rewrite, so a large catch-up fold stays on the incremental path and the cold-fetch stalls collapse.

## Performance

Validated on the 561M-triple DBLP dataset against real S3 + DynamoDB, and re-confirmed on a half-size 8 vCPU / 16 GB instance.

- The decisive outcome: on a cold fold the previous behavior reliably timed out on a serial Phase 3b class-stats fetch and fell back to a full rebuild, which needs ~31.7 GB and is OOM-killed on a 32 GB box (and is instantly fatal on a 16 GB box). The fold now stays incremental and completes in ~28–30s at ~0.5 GB resident, running identically on the 8 vCPU / 16 GB box.
- Resolve: replacing serial commit-DAG discovery with a maintained commit-CID index dropped the discovery-and-fetch walk from ~59s of serial S3 round-trips to ~4.5s, roughly halving the full fold (~105s → ~50s) on the first measurement pass.
- Phase 3b on a cold store: the rdf:type PSOT scan went from a 35,000 ms timeout to 129 ms, and the base-class seed lookups from ~35,400 ms to ~0.3 s, by prewarming exactly the leaves those serial lookups read.

## What changed

- A per-ledger commit-CID index, maintained on commit-head advance, lets the fold skip serial commit-DAG discovery and fetch commit bodies in bounded parallel.
- Phase 2 leaf prefetch and upload are streamed and globally bounded: leaves are prefetched and uploaded under one shared S3 in-flight budget and freed per put, so peak resident memory stays flat regardless of fold size.
- The dict reverse-tree leaves the reconcile reads, the Phase 3b rdf:type PSOT leaves, and the base-class reverse-tree leaves are each prewarmed in bounded parallel under the same shared budget, gated to remote stores only.
- The Phase 2 per-leaf copy-on-write rewrite runs across the shared rayon pool in bounded windows, degrading to the serial path on a single core or a small touched-leaf set so a 2 vCPU box is never oversubscribed.

## Correctness and bounds

- Every prefetch and prewarm only warms the on-disk leaf cache, and the windowed parallel rewrite reassembles leaves in manifest order, so the produced index is byte-for-byte identical to the serial path, with identical content IDs.
- All concurrent S3 reads and writes across the entire fold draw from a single shared in-flight semaphore, so the added parallelism cannot flood S3 or oversubscribe a small box.


fluree-db-indexer/src/build/incremental.rs
```diff
@@ -262,12 +318,52 @@ async fn run_update_branch(
...
+            .await?;
+            totals.leaf_bytes += c.leaf_bytes;
+            totals.sidecar_bytes += c.sidecar_bytes;
+            totals.sidecar_count += c.sidecar_count;
+        }
+        Ok::<LeafUploadCounts, IndexerError>(totals)
+    };
+
+    // Drive both to completion. If the uploader fails it drops `rx`, so the
+    // CoW's next `blocking_send` errors out instead of hanging.
```
zonotope, line 355: This isn't true. `rx` is borrowed from `run_update_branch`,  so completing the future doesn't actually drop `rx`. it only returns ownership to `run_update_branch`.  Uploader needs to be converted to `async move` to take ownership of `rx` instead of borrowing.

fluree-db-indexer/src/run_index/build/incremental_resolve.rs
```diff
@@ -861,23 +889,142 @@ async fn seed_vector_fact_handles(
...
+    let Some((_, tip_cid)) = pending.last() else {
+        return false;
+    };
+    if tip_cid != head_id {
+        return false;
+    }
+    // Contiguous from `from_t + 1` (gap-free coverage up to the tip's t).
+    (from_t + 1..)
+        .zip(pending.iter())
+        .all(|(expected, (t, _))| *t == expected)
```
zonotope, line 921: This doesn't take merges into account. There could be multiple different commits at the same `t` in the ancestry because it's a dag as opposed to a chain.

fluree-db-indexer/src/run_index/build/incremental_resolve.rs
```diff
@@ -861,23 +889,142 @@ async fn seed_vector_fact_handles(
...
+}
+
+#[allow(clippy::too_many_arguments)]
```
zonotope: I don't think we should override clippy. We should fix the code that triggers the warning instead.

In this case, all the arguments to this function come from an `IncrementalResolveConfig` except for `cs`, so we could just pass that and drop this function to 2 arguments. We could also create a new `CommitWalkConfig` or something that only contains the required info for this function if we don't want to pass the resolve config directly.

zonotope (approved):
Looks good besides what I flagged.

fluree-db-indexer/src/run_index/build/incremental_resolve.rs
bplatz: Addressed in e910ac5a971b1a795066e8c6271defab6c0e14ce

fluree-db-indexer/src/run_index/build/incremental_resolve.rs
bplatz: Addressed in ba7c57fd5e8fcfa38206d19a4c99c46b84eb5b88. The fix stops the file nameservice from deduplicating by `t` alone — both commits at a shared `t` are now retained, so the existing contiguity check in `pending_covers_range` naturally fails and falls back to the serial DAG walk.

fluree-db-indexer/src/build/incremental.rs
bplatz: Addressed in 246e709bcab2736ec019e37acfc79955b7a17ab6

PR #1298, opened 2026-06-07
feat: RDF 1.2 edge annotations

bplatz:
## Summary

Adds RDF 1.2 / SPARQL 1.2 edge annotations: attach properties to a *relationship* (the edge between two subjects) without hand-modeling an intermediate node. RDF 1.2 reification (`reifier rdf:reifies <<( s p o )>>`) is mapped onto a native, property-graph-style edge-annotation primitive, so the same data reads correctly from both the RDF/SPARQL and the LPG side.

## What it adds

- **Write surfaces.** JSON-LD `@annotation` / `@edge` / `@reifies` on value objects and nodes (full coverage: literal-valued edges incl. language-tagged, named graphs, parallel annotations, named/explicit-IRI reifiers, body cascades); SPARQL 1.2 UPDATE annotation tails `{| ... |}`, `~` reifiers, INSERT/DELETE templates; bulk import of already-lowered bundles as an administrative bootstrap path.
- **Read surfaces.** `@annotation` hydration in JSON-LD results; SPARQL `{| ... |}`, `~`, and `rdf:reifies <<( s p o )>>` query forms; a reserved-predicate firewall that keeps the internal `f:reifies*` system facts out of ordinary queries (with `includeSystemFacts` as the documented escape hatch).
- **Storage / indexing.** An `EdgeKey` `(g, s, p, o, dt, lang, list_i)` encoded as seven durable `f:reifies*` system facts, a forward/reverse annotation arena on the index root, a sticky `has_annotations` bit, incremental and full-rebuild seal paths, and GC. A ledger with no annotations creates no annotation artifacts and the fact indexes are unchanged.
- **Zero-cost when unused.** Non-annotated queries and writes pay nothing: the constant-predicate scan path short-circuits, the WHERE planner skips annotation expansion when no annotation patterns are present, and the cascade write path returns immediately when the ledger has never carried an annotation.

## Conformance scope (v1)

Fluree's annotation is lifecycle-coupled to an asserted edge. Spec shapes that the LPG model does not represent are rejected cleanly (clear validation error, never a silent wrong result) and documented as deferred: unasserted triple terms, triple terms under a non-`rdf:reifies` predicate, nested triple terms as values, list-occurrence annotations, the SPARQL 1.2 triple-term functions (`TRIPLE` / `SUBJECT` / `PREDICATE` / `OBJECT` / `isTRIPLE`) and `BIND(<<( ... )>>)` constructor, the compact `<< s p o ~ r >>` reifier delimiter (it collides with the legacy `f:t` / `f:op` extraction form), native Turtle/TriG annotation parsing, and non-JSON-LD annotation output. The mandated SPARQL 1.2 `VERSION "1.2"` prologue is accepted (lex-and-skipped).

## Recent hardening

A spec-conformance + correctness + performance audit of the feature drove a round of fixes folded into this branch: language-tag fidelity on the SPARQL UPDATE write path, a firewall bypass via VALUES/`@id`-bound predicates, a GC guard so re-sealing an unchanged annotation arena cannot delete still-live CIDs, cascade / novelty tie-break / reindex-bootstrap correctness, and the planner clone gate above. Core read/write/hydration round-trips, the firewall, the RETRACT path, and the GC guard are covered by tests.

## Notes

Draft. Base is `refactor/column-caching` (this branch merges it in). See `docs/concepts/edge-annotations.md` for the user-facing surface and `docs/design/edge-annotations.md` for the storage internals.


bplatz:
Closed in favor of https://github.com/fluree/db/pull/1357

PR #1299, opened 2026-06-07
fix: fluree-memory -- fix concurrent mcp use block

bplatz:
- Move the file-sync rebuild lock wait onto Tokio’s blocking pool so `flock` cannot block async worker threads.
- Serialize file-backed memory mutations across both in-process MCP calls and shared store instances.
- Add a manual concurrent MCP-like add stress example for validating the regression.


zonotope (approved):
🥍

PR #1300, opened 2026-06-08
fix: Bound Wikidata-scale bulk import memory

bplatz:
This PR fixes several memory-consuming bulk import phases that did not respect the intended memory envelope at Wikidata scale. The changes were driven by a ~20B triple Wikidata import, where distinct-term cardinality and class-membership/index-build structures pushed the process into swap on a 256GB machine.

The branch keeps the required class/ref-class stats and indexing behavior, but changes the implementation so large structures are streamed, bounded, disk-backed, or prevented from overlapping at peak memory.

## Changes
- Stream vocab merge remap writes instead of dirtying large mmap files.
  - Replaces mmap remap writers with sequential buffered writers.
  - Preserves the existing on-disk `.rmp` format.
  - Avoids dirty RSS proportional to total chunk-local term cardinality.

- Stream bulk dictionary upload for trusted sorted bulk-import artifacts.
  - Avoids materializing subject/string term-sized vectors during dict upload.
  - Uses sequential readers for subject SID maps and forward indexes.
  - Keeps the non-trusted/materialized path for paths that cannot rely on bulk import ordering invariants.

- Add cgroup-aware memory detection.
  - Import memory sizing now respects container/systemd cgroup limits instead of host RAM only.
  - Cache sizing also uses the effective memory limit.

- Add phase-level memory observability.
  - Logs RSS/projection data for bulk import phases.
  - Adds targeted index-build memory snapshots including anon/file RSS and swap.
  - Logs class-membership representation, counts, and per-order index-build memory.

- Add a Docker hard-cap memory repro script.
  - Provides a Linux/cgroup repro for proving old materialized dict upload behavior OOMs while the streaming path survives.

- Parallelize local directory rechunking with bounded read-ahead.
  - Removes the single `ttl-local-rechunk` producer bottleneck seen during Wikidata parse.
  - Uses automatic rechunk worker sizing, with hidden env override for field testing.
  - Adds sliding-window backpressure so parallel producers cannot decompress hundreds of files ahead of the parser/commit pipeline.

- Add hidden import tuning/diagnostic overrides.
  - `FLUREE_IMPORT_MAX_INFLIGHT`
  - `FLUREE_IMPORT_RECHUNK_THREADS`
  - Existing user-facing import knobs remain unchanged.

- Bound index-build memory overlap.
  - Runs SPOT/class-membership stats before secondary-order run generation/builds.
  - Ensures the large class-membership heap is dropped before PSOT/POST/OPST buffers start.
  - Preserves required class/ref-class stats.

- Replace import-path class membership heap with disk-backed membership.
  - Partitions `(graph, subject, class)` membership entries into disk buckets.
  - Sorts and compacts one bucket at a time.
  - Uses mmap-backed compact indexes for SPOT ref-class lookups.
  - Prevents Wikidata-scale imports from accumulating an unbounded subject→class heap.

## Notes
This PR is focused on making the Wikidata-scale bulk import complete within memory limits. It may trade some index-build speed for bounded memory in the largest class-membership/stat phases, but avoids the swap-crawl/OOM failure modes observed in the 20B triple import.


zonotope (approved):
🪶

PR #1301, opened 2026-06-09
perf: Reduce transaction novelty cost and refresh local index cache

bplatz:
This improves transaction throughput by reducing novelty write costs and avoiding unnecessary cloning during commit application. In local background-indexing runtimes, it also wires successful index publishes back into the cached `LedgerManager`, so hot cached ledgers advance to the new index and trim active novelty without requiring an explicit `refresh()`.

The local cache listener now reacts only to `LedgerIndexPublished` events, avoiding a background nameservice lookup on every commit event. It also drops its extra event-bus sender after subscribing so the task exits cleanly when the owning `Fluree` instance is dropped.

## Changes
- Add `NoveltyFactState` for efficient set-semantics deduplication in novelty.
- Use `Arc::make_mut` in transaction/ledger commit paths to avoid unnecessary deep clones.
- Add a local nameservice event listener that routes `LedgerIndexPublished` to `LedgerManager::notify` and `LedgerRetracted` to `disconnect`.
- Ensure AWS dynamic builder path publishes through `NotifyingNameService`.
- Add regression coverage proving cached ledgers apply locally published indexes and trim novelty without manual refresh.
- Remove the segmented novelty planning doc from checked-in docs.


zonotope (approved):
🪘

PR #1302, opened 2026-06-09
perf: append-only segmented novelty — O(batch) write path

bplatz:
Reworks the in-memory **novelty** write buffer (the uncommitted-changes overlay that sits between index publishes) from a single merged, sorted index vector — rebuilt on every commit — into an **append-only list of immutable per-commit segments**. Each commit appends one `Arc<Segment>` per touched graph instead of re-merging the whole novelty index, dropping **per-commit write cost from O(accumulated novelty) to O(batch)**. Reads merge across segments with a comparator-ordered k-way merge that reproduces the exact pre-change ordering.

## Performance (before → after)

| Metric | Before (single merged vector) | After (append-only segments) | Change |
|---|---|---|---|
| Per-commit write cost | O(accumulated novelty) | O(batch) | algorithmic; O(N²)→O(N) over a burst |
| Write slope (µs / 1k novelty flakes) † | 144 | 0.089 | O(N) → flat |
| Mean commit latency at depth † | ~26 ms | ~0.5 ms | ~53× |
| Snapshot-isolation clone of novelty | O(novelty) deep copy | O(#segments) pointer copy | — |
| Update p50, real-server mixed load ‡ | ~1471 µs | ~620 µs | ~2.4× faster |
| In-memory overlay under sustained load ‡ | ~1.75 MiB | ~0.09 MiB | ~19× smaller (indexer keeps pace) |
| Net throughput, mixed update+query ‡ | baseline | +~8% | modest (read-bound mix) |
| Point / key-lookup latency ‡ | baseline | unchanged | parity |
| Broad scan / join latency ‡ | baseline | unchanged | parity (no read regression) |

† Two identical m7i.xlarge boxes, `/dev/shm`, release — before-vs-after segmentation (original write-slope measurement).
‡ This change set, replicated, on a **BSBM-shaped simulation** (in-repo generator + representative lookup/scan/join + an insert update), driven through the real server path (cached `LedgerHandle` + `stage()` commits + warm `snapshot()` queries + a live background indexer). **Not** the official BSBM harness.

## What changed

- **Append-only segmented novelty** — per-graph `Vec<Arc<Segment>>`; `apply_commit` builds and appends one immutable segment, with no re-merge of existing data (O(batch) commits). The whole-novelty deep clone for snapshot isolation becomes an O(#segments) pointer copy (immutable segments are never mutated under readers).
- **Read path** — comparator-ordered k-way merge across a graph's segments, byte-identical to the old single-vector order (golden equivalence digest unchanged); zone-map segment pruning + a single-segment fast path matching the old single-vector read speed.
- **Segment-aware overlay translation cache** — query-time overlay translation is cached per immutable segment, so a query after a write re-translates only the newly-appended segment rather than the whole overlay (byte-bounded, folded into the existing leaflet-cache budget).
- **LIMIT row-budget** for the eager nested-loop join flush.
- **Two indexed-write-path fixes** — detach the novelty range provider before the per-commit dict `make_mut` (it co-held the dictionary `Arc`s, forcing an O(novelty) deep clone on every commit once an index was published); skip the config-graph resolve when that graph is empty (it full-scanned the `rdf:type` predicate over the growing overlay on every request).
- **Optional structural compaction** to bound segment count, on the read / maintenance path only — preserves every flake (stats-safe, observationally transparent). Kept as a maintenance primitive, **not** on the write path.

## Tradeoffs

- **Segment fan-out on reads.** A read now does one binary-search probe per segment plus a k-way merge, where the old path scanned one contiguous vector. Bounded by zone-map pruning (point reads prune to ~1 segment), the single-segment fast path, and — with the indexer keeping the overlay small — negligible in practice. The cost only grows if the overlay is allowed to grow large (indexer far behind).
- **Broad scans over a large overlay** re-assemble the overlay's merged op-stream per epoch (the first query after each commit), O(overlay); the per-segment translation cache limits that to the *delta*, and it amortizes across queries between commits. Note `main` also degrades on broad reads with a non-empty overlay (it shares the overlay) — this adds only a small fan-out increment on top. The remaining flat per-query broad-scan cost is a **pre-existing** base-index issue (untyped-string predicate scans), not introduced here.
- **Memory:** a byte-bounded per-segment overlay-translation cache (capped, co-resident in the leaflet-cache budget).
- **Complexity:** a new segment data structure, the translation cache, and an optional compaction path; reads go through a merge rather than a flat scan.

## Testing

- Golden `novelty_equivalence_contract` unchanged — byte-identical results across all four index orders, graphs, time-travel, and edge cases (reassert-after-retract, duplicate metadata, named graphs). Compaction observational-transparency tests (compact-all + tiered) across all orders/graphs.
- `fluree-db-novelty` / `fluree-db-ledger` / `fluree-db-transact` unit suites green; API integration suites green (RDF 1.2 edge annotations, indexing workflow, enforce-unique-after-index, namespace-after-index, config-graph).


fluree-db-ledger/src/staged.rs
aaj3f: Given that `slice_for_range()` went from a borrowed `&[FlakeId]` to an owned `Vec<FlakeId>`, and given the introduction of `read_flakes` / `for_each_overlay_flake` on `Novelty`, we should move as many callers off `slice_for_range` as possible to avoid the unnecessary allocations. I'll try to comment on any others I see, but worth a full sweep to see if I've missed any

fluree-db-novelty/src/lib.rs
```diff
@@ -828,19 +1380,70 @@ impl OverlayProvider for Novelty {
...
+        if let Some(Some(segs)) = self.graphs.get(g_id as usize) {
+            if let Some(seg) = segs.iter().find(|s| s.seg_id == seg_id) {
+                let flakes = &seg.flakes;
+                for &local in seg.order(index) {
+                    callback(&flakes[local as usize]);
+                }
             }
         }
     }
 }
```
aaj3f, line 1429: This gets called in `collect_segment_merged_ops` of `binary_scan.rs` so that it loops over all `K` segments of a graph, calling this once per segment. So the first query after a `K`-commit burst seems to have a cost increase in segment-lookup work compared against the legacy single-merge whole-graph path it replaced. With tiered compaction this would be bounded, but wouldn't be if compaction wasn't enabled or if the indexer is falling behind. 

It seems maybe the segments are already enumerated, though, in `overlay_segments(g_id)` order. If we either passed the segment index or handed the caller `&Segment`, the lookup could instead be `segs[i]` instead of a linear `find`

fluree-db-api/src/ledger_manager.rs
```diff
@@ -181,10 +190,59 @@ impl LedgerHandle {
...
+        };
+        if !needs {
+            return;
+        }
+        let mut state = self.inner.state.write().await;
+        // Re-check: another query may have compacted between the locks.
+        if state.novelty.needs_tier_compaction(tier_width) {
+            let pre = state.novelty.max_segment_count();
+            let started = Instant::now();
+            let merges = Arc::make_mut(&mut state.novelty).tier_compact(tier_width);
```
aaj3f, line 234: So I think the call of `tier_compact()` (i.e. when a size class fills for a single graph and triggers compaction) ends up bumping `novelty.epoch`, but `epoch`, is the `overlay_epoch` of the **global** overlay-translation cache key. So it would flush the **per-epoch translation cache for every graph**.

fluree-db-novelty/src/lib.rs
```diff
@@ -680,63 +1100,187 @@ impl Novelty {
...
 
-        // Update size
-        self.size = new_size;
+    /// Graph ids (ascending) that currently hold at least one non-empty segment.
+    fn present_graphs(&self) -> Vec<GraphId> {
+        (0..self.graphs.len())
+            .filter(|&g| self.graphs[g].as_ref().is_some_and(|s| !s.is_empty()))
+            .map(|g| g as GraphId)
+            .collect()
+    }
```
aaj3f, line 1156: This is mainly re: `present_graphs()` and `graph_merge()`. 

Within `present_graphs()`, we allocate a fresh `Vec<GraphId>` on **every** `iter_flakes` / `iter_index` call. Before this PR `iter_index` was a pure borrowing iterator and had zero allocation. 

`graph_merge()`, for each K>1, allocates both a `Vec<MergeStream>` **and** a  `BinaryHeap<MergeHead>` per read (where pre-PR `main` returned a borrowed slice).

I know this isn't exactly the hot path and so is a lower perf cost, but perhaps we should turn `present_graphs` from a collector into just an iterator, and for very small K (e.g. 2-3, maybe) perhaps it's worth a direct linear merge of borrowed slices rather than (or before) reaching for the heap, and then otherwise for larger K, using `Vec::with_capacity(seg_count)` / `BinaryHeap::with_capacity`

fluree-db-novelty/src/lib.rs
```diff
@@ -262,30 +521,303 @@ impl Novelty {
...
+            for seg in std::mem::take(segs) {
+                match Arc::try_unwrap(seg) {
+                    Ok(s) => flakes.extend(s.flakes),
+                    Err(shared) => flakes.extend(shared.flakes.iter().cloned()),
+                }
+            }
+            flakes
+        };
+        self.set_graph_segments(g as GraphId, flakes, true);
+    }
```
aaj3f: I think that `compact_over` / `compact_all` / `compact_graph` / `needs_compaction` (and, though not in this code block, `DEFAULT_COMPACTION_THRESHOLD`) are all unused except for in tests.

fluree-db-novelty/src/lib.rs
aaj3f, line 1348: This comment is now likely stale as this PR removes this arena. Also, now that `clear_up_to` drops/rebuilds whole segments, is there still a significant difference between `size == 0` and `flake_count == 0` == `is_empty()`? At minimum maybe a comment update, but possibly we could also collapse the two predicates?

fluree-db-novelty/src/lib.rs
```diff
@@ -749,49 +1293,57 @@ impl Novelty {
...
 
-    /// Get flake reference by ID
+    /// Resolve a [`FlakeId`] produced by this novelty back to its flake.
     pub fn get_flake(&self, id: FlakeId) -> &Flake {
-        self.store.get(id)
+        let segs = self.graphs[id.graph()]
+            .as_ref()
+            .expect("FlakeId references a live graph");
+        &segs[id.seg()].flakes[id.local()]
     }
```
aaj3f, line 1279: Given that `slice_for_range`/`iter_index` now return an **owned** `FlakeId`, a caller of one of them that returns an id across a mutation that rebuilds segments (e.g. `clear_up_to`, `tier_compact`, `bulk_apply_commits`), could either panic or silently resolve a different flake. Previously, the old `u32` arena index was borrow-checked against this. 

All the current callers resolve ids --> flake immediately, so it's not an introduction of any active bugs anywhere in this PR, but it's a kind of latent footgun looking forwards. I don't know that we need a refactor so much as maybe just doc comments on `slice_for_range`/`iter_index` that any ids will be invalidated by segment-rebuilding mutation or maybe some nudging via `#[must_use]` kinds of annotations.

aaj3f (approved):
This is really exciting, @bplatz! Seems like big wins for index updates!

fluree-db-ledger/src/staged.rs
bplatz: Addressed in 6b76100c0662efe984ea896d7dc169e7cd274618

fluree-db-novelty/src/lib.rs
bplatz: Addressed in 6b76100c0662efe984ea896d7dc169e7cd274618

fluree-db-novelty/src/lib.rs
bplatz: Addressed in 6b76100c0662efe984ea896d7dc169e7cd274618

fluree-db-novelty/src/lib.rs
bplatz: Addressed in 6b76100c0662efe984ea896d7dc169e7cd274618

fluree-db-novelty/src/lib.rs
bplatz: Addressed in 6b76100c0662efe984ea896d7dc169e7cd274618

fluree-db-novelty/src/lib.rs
bplatz: Partially addressed in 2dfc0866e858b0fd48a169e27b4fa0025426fa00 — `graph_merge` now pre-sizes its `MergeStream` Vec and `MergeHead` heap to the graph segment count. Leaving the `present_graphs` → iterator conversion and the small-K (2–3) borrowed-slice linear merge as a follow-up (more invasive in the merge core for a modest constant-factor win on a path the indexer + tiered compaction keep small).

fluree-db-api/src/ledger_manager.rs
bplatz: Going to leave this one as-is. The per-segment translation cache is keyed by `seg_id` (not epoch), so it survives the bump — an untouched graph only pays a re-merge of already-translated segments, not a re-translation. Single-graph ledgers (the common case) never over-invalidate at all; the cost only shows up with multiple active graphs, which is uncommon and usually has centralized writes anyway. Not worth changing the cross-query invalidation semantics for that.

PR #1303, opened 2026-06-09
feature: cooperative query cancellation primitives

bplatz:
# Query timeout and cooperative cancellation

Long-running queries had no way to stop early: a slow property-path traversal, a large star-join, or a multi-query bundle would hold server resources until it finished even if the client had already gone away or a wall-clock limit had been exceeded. This PR adds a complete cooperative cancellation system and wires it end-to-end.

## Cancellation primitive (`fluree-db-core`)

`QueryCancellation` is a runtime-agnostic, atomic cancel handle. Three reasons are tracked — `Cancelled`, `Timeout`, and `ClientDisconnected` — and the first one signalled wins (compare-exchange). A disabled handle is a single `None` pointer, so callers that do not opt in pay only a branch at each checkpoint (~1 ns).

## Query operator checkpoints (`fluree-db-query`)

`check_cancelled()` calls are placed at the hot iteration boundaries in:

- property path traversal
- direct index / binary scans
- count plan execution (all overlay variants)
- join, optional, minus, semijoin, union, subquery, aggregate, groupby, sort, and dataset operators

A cancelled query returns `QueryError::Cancelled { reason }` and unwinds normally — no panics, no leaked state.

## Builder and execution wiring (`fluree-db-api`)

`QueryExecutionOptions` carries the cancellation handle and any lifecycle guards. `ViewQueryBuilder` and `DatasetQueryBuilder` each gain `.execution_options()` and `.cancellation()` methods. All internal `query` / `query_tracked` / `query_view_with_r2rml` call sites are updated to forward these options so the handle flows from server route → builder → execution context → checkpoints.

## Server task isolation and disconnect detection (`fluree-db-server`)

`run_query_task` spawns query work in a separate tokio task. A drop guard on the *awaiting* side signals `ClientDisconnected` if the HTTP or MCP handler future is dropped before the query task completes — catching both graceful client closes and aborted requests without polling. A lightweight timeout task fires `Timeout` after the configured duration and then exits; its abort handle is kept in a second drop guard so it is cleaned up when the query finishes normally.

`current_query_execution_options()` uses a task-local to make already-installed options available to route helpers without threading them through every call site.

## Configuration

| Setting | Env var | Default |
|---|---|---|
| HTTP query timeout | `FLUREE_QUERY_TIMEOUT_MS` | 15 min |
| MCP `sparql_query` timeout | `FLUREE_MCP_QUERY_TIMEOUT_MS` | 5 min |

Set to `0` to disable. Both are also settable via the config file.

## Tests

- Unit tests in `cancellation.rs` verify disabled/cloned/first-wins semantics.
- Unit tests in `query_control.rs` verify disconnect signalling (waiter abort → `ClientDisconnected`), clean-completion no-signal, and timeout-fires-before-query-ends.
- Integration tests in `it_query_sparql_indexed.rs` cover end-to-end cancellation through query execution.


aaj3f (approved):
I'm quite excited about this -- I think very large datasets where LLM behavior degraded because timed-out queries were still running on fluree resources will now improve drastically.

Claude found a few direct-leaf-read operators (i.e. not child operators that would inherit checkpoints) that have zero `check_cancelled()` sites:

- `fluree-db-query/src/fast_predicate_scalar_agg.rs` (e.g. `:357-393`, `:486-492`
  — full leaf/leaflet/row walk of a predicate for SUM/MIN/MAX)
- `fluree-db-query/src/fast_min_max_string.rs`
- `fluree-db-query/src/fast_post_order_limit.rs`
- `fluree-db-query/src/fast_count.rs`
- `fluree-db-query/src/binary_history.rs` (history replay walk)
- `fluree-db-query/src/fast_label_regex_type.rs`, `fast_exists_join_count_distinct_object.rs`, `fast_sum_strlen_group_concat.rs`

I'm not sure if any of them deserve cancellation checks within this PR or if we just want to file an issue to consider them additively in future PRs

bplatz:
> I'm quite excited about this -- I think very large datasets where LLM behavior degraded because timed-out queries were still running on fluree resources will now improve drastically.
> 
> Claude found a few direct-leaf-read operators (i.e. not child operators that would inherit checkpoints) that have zero `check_cancelled()` sites:
> 
> * `fluree-db-query/src/fast_predicate_scalar_agg.rs` (e.g. `:357-393`, `:486-492`
>   — full leaf/leaflet/row walk of a predicate for SUM/MIN/MAX)
> * `fluree-db-query/src/fast_min_max_string.rs`
> * `fluree-db-query/src/fast_post_order_limit.rs`
> * `fluree-db-query/src/fast_count.rs`
> * `fluree-db-query/src/binary_history.rs` (history replay walk)
> * `fluree-db-query/src/fast_label_regex_type.rs`, `fast_exists_join_count_distinct_object.rs`, `fast_sum_strlen_group_concat.rs`
> 
> I'm not sure if any of them deserve cancellation checks within this PR or if we just want to file an issue to consider them additively in future PRs

I'd like to take another pass at fine tuning this as we go forward, I had some super high performing queries in benchmarks degrade up to 15% because there were too many checks so already moved some things 'up' so they would not been the hot path as much.

The goal here is to, I'd say, fully cancel within say 1/4 of a second, not instantly. Some of these hot paths give instant but at a cost.

PR #1305, opened 2026-06-10
fix: same-subject star joins drop novelty bindings on indexed ledgers

bplatz:
## Summary

On an indexed ledger with pending novelty, joins that use the batched base-leaflet probes could silently drop novelty facts that a single-predicate scan sees. In the reported repro, `SUM(?v)` over `?s ex:price ?v . ?s ex:currency ?c` returned 60 instead of 67 because the novelty assertion `ex:a ex:price 7` was lost by the join. Any aggregate over such a join can undercount (`SUM`/`AVG`/`COUNT`, and `MIN`/`MAX` when the extremum lives in novelty), and retracted-in-novelty rows could incorrectly reappear.

## Root cause

Several operators share batched helpers that read base index leaflets directly for amortization (one leaflet load → many subject probes): the nested-loop join's batched flush paths, the property join's subject probe and SPOT star walk, and the batched OPTIONAL builders. These helpers handle historical time travel via sidecar replay, but never merge the novelty overlay — base leaflets were treated as authoritative even when novelty was present.

The live trigger is the nested-loop join: under overlay, scans materialize subjects eagerly as `Sid` bindings, but the join's batched key extraction resolves those back to raw subject ids via `find_subject_id_by_parts` and proceeds with the base-only probe. A ref-hop shape (`?o ex:item ?s . ?s ex:price ?v`) reproduces the exact symptom. The property-join and OPTIONAL probe paths carry the same latent hazard and are gated as well.

## Fix

Adds `ExecutionContext::overlay_free_single_graph()` — true when the active scope resolves to exactly one graph whose overlay epoch is 0, the same convention used by late materialization and the fast paths. Every batched leaflet-probe entry point now checks it and falls back to the per-row scan path, whose binary cursor merges overlay ops correctly:

- `join.rs` `use_batched` (covers the subject, object, and exists flush helpers)
- `property_join.rs` `can_batched_probe` and `can_spot_walk_remaining`
- `optional.rs` both `build_batch` implementations

This is the bail strategy: overlay-carrying queries lose the batched amortization but stay correct. A follow-up could merge overlay ops inside the probes (mirroring `BinaryCursor::set_overlay_ops`) to recover the batched path under novelty.

## Testing

New regression test `indexed_inline_type_star_aggregate_with_overlay_multivalue_and_facet` in `fluree-db-api/tests/it_query_sparql_indexed.rs` seeds and indexes line items, adds a novelty-only price, and asserts `SUM = 67` across: the single-predicate control, the faceted same-subject star in both pattern orders, a `FROM`-clause dataset-mode variant, the ref-hop join shape (verified failing before the fix), an OPTIONAL probe variant, and the typed star. Existing query, time-travel BGP, dataset, aggregate, and OPTIONAL suites all pass.


fluree-db-query/src/context.rs
```diff
@@ -700,6 +700,25 @@ impl<'a> ExecutionContext<'a> {
...
+    /// (whose binary cursor merges the overlay) when it returns `false`.
+    pub fn overlay_free_single_graph(&self) -> bool {
+        match self.active_graphs() {
+            ActiveGraphs::Single => self.overlay().epoch() == 0,
+            ActiveGraphs::Many(graphs) => match graphs.as_slice() {
+                [g] => g.overlay.epoch() == 0,
+                _ => false,
+            },
+        }
+    }
```
aaj3f, line 724: I may be wrong here, but it seems `epoch() == 0` is not necessarily sufficient, as it only describes the empty novelty state obtained at point of `LedgerState` load. I think, for example, when background indexing catches up (i.e. `apply_index() --> clear_up_to()`), you would end up with "empty novelty" but `epoch > 0`, and then the paths made possible by this `overlay_free_single_graph() -> bool` check would be lost even as novelty is empty.

Perhaps we could leverage `Novelty::is_empty()` (which should be available on `overlay()` if downcast to `Novelty`.

aaj3f (approved):
Good fix for this issue. If my comment below is correct w/ its assumptions, we may just have a scenario where your batching improvements COULD get applied but don't because of a mischaracterization of `epoch > 0` empty novelty as being non-empty novelty

fluree-db-query/src/context.rs
bplatz: Addressed in ddedce863fa62a2f840f3018850da342716bcb13

PR #1306, opened 2026-06-10
perf(query): cyclic BGP fast path and star-join retunes for graph-pattern workloads

bplatz:
Targets the query shapes that previously planned as left-deep nested-loop joins and dominated our slow tail on large graph-pattern workloads: cyclic BGPs (triangles and 4-cycles) and deep star/chain joins. On an 850-query graph-pattern suite over a 20+ Billion-triple ledger, this work raises the sub-second share from 78% to 88.9% and the sub-100ms share from 56% to 67.6% (median 42ms, zero timeouts); the former >5s cyclic tail now tops out at 3.3s.

## CyclicBgpOperator

A conservative fast path for fixed-predicate triangle/square BGPs whose join variables form one cycle. Eligible blocks load each edge's relation, semi-join prune to a fixpoint, and enumerate from the cheapest driver; encoded squares can use an exact wedge-pair hash decomposition.

Edge loading cascades: the cheapest edge full-scans, then remaining edges load by bounded per-subject probes (one amortized monotonic PSOT seek per edge) when an already-scanned relation bounds their subject frontier under a cap-and-ratio gate, falling back to full scans otherwise. When statistics prove no probe can engage, edge scans and index builds run concurrently on the shared worker pool.

The data plane joins on Copy values (CSR-style relation indexes; bindings materialize only for emitted rows). Probing is HEAD-gated; history/overlay/multi-ledger execution declines to the previous operator tree, which every instance carries as a fallback child. `FLUREE_CYCLIC_BGP=0` disables the operator entirely; see `docs/query/explain.md` for the full knob set and EXPLAIN fields.

## Star/chain join retunes

- The object hash-join cost model now admits small build sides and modest wide intermediates where per-row object seeks dominated deep star chains, and EXPLAIN reports each decision with its rejection reason.
- `PropertyJoinOperator` is now gated to anchored stars (a bound object or pushed-down bounds); pure variable stars stream through the sequential chain so LIMIT can exit early.
- A VALUES clause preceding a star block now seeds the chain when it constrains it (fixing a cross-product regression the gate change exposed, plus the same latent hole for SPARQL trailing VALUES); non-overlapping parameter VALUES still defer above the block.
- With populated statistics, predicates absent from stats estimate as empty and drive first, so unsatisfiable blocks exit in milliseconds.

## Observability

The cyclic operator exposes per-phase debug spans (edge scan, index build, prune, wedge build, enumerate), ledger view acquisition is now spanned (`ledger_view_load`), and probe-gate decisions log their decline reasons. Span hierarchy docs and tracing acceptance tests updated.

## Also included

The global cache auto-tier on ≥8GB hosts drops from 50% to 35% of RAM (cgroup-clamped), which benchmarked better by leaving headroom for the OS page cache; docs updated accordingly.

## Testing

Correctness is backed by differential tests running each cyclic shape under fallback, probed, and full-scan configurations against identical results, plus the existing W3C SPARQL suite.


fluree-db-query/src/cyclic_bgp.rs
```diff
@@ -0,0 +1,2742 @@
...
+        Ok(v) if v == "0" || v.eq_ignore_ascii_case("false")
+    )
+}
+
+fn max_predicate_rows() -> u64 {
+    std::env::var("FLUREE_CYCLIC_BGP_MAX_ROWS")
+        .ok()
+        .and_then(|v| v.parse().ok())
+        .unwrap_or(DEFAULT_MAX_PREDICATE_ROWS)
+}
```
aaj3f, line 47: Whether pre-merge or later, I wonder if we might want to leverage the other patterns in the codebase for establishing more of a memory upper bound / byte budget rather than an arbitrary one like this 10mil const.

fluree-db-connection/src/cache.rs
aaj3f, line 1: We may just want to prominently call this 50% -> 30% change in release notes. Not that there are a whole lot of folks using this yet, but for those people who use the defaults (auto-detect) rather than an explicit override... even with your bench testing of performance improvement, it may be a significant change re: cache eviction frequency etc

fluree-db-api/tests/it_cyclic_bgp_probe.rs
aaj3f, line 1: These tests helped me understand this a bit better. A few additional tests we could maybe add:

- **novelty present** i.e. assert that the operator's full-scan path merges overlay and that probes disable correctly
- a cycle with two edges sharing one predicate
- a `LIMIT 1` cyclic query (exercises early exist against the eager `open()` build)

aaj3f (approved):
Well I learned a lot about the ways people apparently talk about Basic Graph Patterns and cycles etc. Which probably means I'm not *best* to review this, since I had to look up what 4-cycle squares were, etc -- but it looks good to me!

fluree-db-api/tests/it_cyclic_bgp_probe.rs
bplatz: OK, will add some more testing, sounds good.

bplatz: Actually going to add tests to fix/query-timeout-cancellation #1303 as that sits on top of this code and introduces cancellation handles that would need to get refactored.

PR #1307, opened 2026-06-10
Native `ndjson`/`jsonl` Streaming & Import Support

aaj3f:
Adds newline-delimited JSON-LD (`.jsonl` / `.ndjson`) as a first-class bulk-import
format, streamed end-to-end in bounded memory across all three source shapes:

- a **single local file** (`fluree create my/ledger --from data.jsonl`, incl. `.gz`/`.zst`)
- a **local directory** of `.jsonl`/`.ndjson` files
- **remote objects** (S3 et al.) via `read_byte_range` windows — no whole-object materialization

Each line is one complete JSON-LD node, optionally preceded by a lone
`{"@context": …}` line shared by the rest of the file. Every file/object keeps its
*own* leading `@context` (readers are chained per-source and concatenated only at
the JSON-LD-chunk level, never byte-concatenated).

## Design

**`NdjsonReader`** (new `ndjson_splitter` module in `fluree-graph-json-ld`): a
streaming reader over any `BufRead` that groups lines into ~`chunk_size` chunks and
wraps each as a standalone `{"@context":…,"@graph":[…]}` document — the exact shape
the existing `splitter` emits, so downstream chunk parsing is reused unchanged.
Every line is cheaply validated (`IgnoredAny`, no value tree) so malformed or
pretty-printed input fails loudly with the **true file line number** rather than
producing a silently-broken chunk. A `FirstLineContextPolicy { Auto | Context |
Entity }` option (`ImportBuilder::ndjson_first_line_context`, default `Auto`)
controls first-line interpretation.

**One pipeline for all source shapes**: `ChunkSource::JsonLdStream` holds a single
`RemoteChunkProducer` fed by `spawn_chained_ndjson_producer` — one `NdjsonReader`
per source, lazily opened on a dedicated producer thread, chunks forwarded with a
global index into the same channel + `error_rx` completion protocol the remote
path already uses. The serial parse+commit loop body is reused verbatim.

**`StorageRangeReader`**: a synchronous `BufRead` over a remote object's byte
ranges (8 MiB windows), `block_on`'d on the reader's own non-runtime thread.
Backends without native ranged reads (the trait default fetches the full object
per call — proxy, encrypted) and objects with unknown size (`size_bytes == 0`)
fall back to **one** whole-object fetch, gated by a new
`StorageRead::supports_ranged_reads()` capability (default `false`; `true` for
S3/file/memory; conservative delegation in the tiered/resolver wrappers).

## Hardening (self-review pass, last commit)

A multi-angle review of the feature surfaced and fixed, with regression tests:

- **Runtime deadlock**: the serial consume loop now receives chunks via
  `spawn_blocking` — a sync `mpsc::recv` on a runtime thread deadlocked imports on
  current_thread / single-worker runtimes against real async storage (also fixes
  the pre-existing remote serial path).
- **Silent data loss**: remote `.jsonl.gz`/`.zst` now decode (previously silently
  skipped while the CLI advertised them); skipped remote objects are warned with
  count + examples; `size_bytes: 0` means "unknown → fetch whole" instead of
  "empty → import zero rows successfully".
- **Clear failures**: empty/context-only sources fail up front with
  `NoChunks("source contained no data records…")` instead of "no commit head after
  import"; error line numbers account for the prelude; UTF-8 BOM tolerated;
  non-object lines (stray scalar/array) fail per-line instead of as a
  location-free expansion error downstream.
- **Memory & copies**: inner reader channels at capacity 1 restore the documented
  `max_inflight × chunk_size` peak bound (was ~2×); the splitter reuses one line
  buffer and `mem::take`s a prefix-seeded chunk buffer (no per-line allocs, no
  full-chunk re-copy, no transient 2×chunk spike); `StorageRangeReader` implements
  `BufRead` directly (no redundant copy layer).
- **Single sources of truth**: `is_ndjson_ext` + public `is_bulk_import_file` back
  every classification site including the CLI's import-vs-transact routing
  (replaces a hand-enumerated 21-suffix list); mixed-format-family rules shared by
  local and remote via `validate_format_families`; `FirstLineContextPolicy`
  re-exported from `fluree-db-api`.
- **Progress**: ndjson sources report a real bytes/chunk-size estimate; the CLI
  commit bar handles unknown totals with the same sentinel as the scan bar (was
  pegged at 100% for the whole run).

## Compatibility with recent `main`

`main` is merged in (clean — no textual conflicts). The cgroup-aware memory
budget (#1300) flows into these paths automatically via
`effective_max_inflight()`/`effective_chunk_size_mb()`; the sliding-window rechunk
work is Turtle-path-only and orthogonal. One follow-up noted for a separate PR:
`commit_parsed_chunks_in_order` (parallel path) still uses a blocking sync recv in
async context — same pattern class fixed here for the serial loop, but no deadlock
cycle there (pure-CPU producers), so it's a consistency fix, not a bug fix.

## Testing

- `ndjson_splitter` unit tests (18): context detection & policies, CRLF/blank/
  trailing-newline, multi-chunk boundaries, BOM, prelude-aware line numbers,
  non-object lines, loud failure on malformed input.
- `it_import_ndjson` (7): single file (with/without leading context), `.jsonl.gz`,
  local directory with per-file contexts, remote objects with per-object contexts,
  unknown-size remote objects on a current_thread runtime, empty-source error.
- Full regression: `it_import` 27/27, `it_import_remote` 7/7,
  `it_file_storage_jsonld` 2/2, CLI 93/93, `fluree-graph-json-ld` 125/125;
  `fluree-db-server`/`-crypto`/`-nameservice` compile against the trait addition;
  fmt + clippy clean.

**Review tip**: the branch reads well commit-by-commit — reader + local file →
remote byte-range arm → CLI acceptance → local directories → review hardening →
merge of `main`. Each feature commit message documents its slice in detail.


bplatz:
docs/operations/running-fluree.md needs updating — the data format support table and the `create --from` bulk-import section don't mention `.jsonl`/`.ndjson`. Worth sweeping docs/ for other references to supported import formats.

fluree-graph-json-ld/src/ndjson_splitter.rs
```diff
@@ -0,0 +1,624 @@
...
+        push_node(&mut chunk_buf, &mut element_count, &node);
+    }
+
+    while read_line(&mut reader, &mut line_buf)? {
+        line_no += 1;
+        let trimmed = line_buf.trim_ascii();
+        if trimmed.is_empty() {
+            continue;
+        }
+        validate_json_line(trimmed, line_no)?;
```
bplatz, line 326: A lone `{"@context": …}` line after line 1 is treated as a data node — e.g. `cat a.jsonl b.jsonl` puts b's context mid-file and b's records resolve against a's context. Could fail loudly here with the line number.

bplatz:
Labeled blank nodes should unify across the whole stream — each file/object = one JSON-LD document scope. Since the namespace allocator is already shared across chunks, this is just making the blank-node id prefix per-source (e.g. `{ledger_id}-{source_idx}`) instead of the current per-chunk `{ledger_id}-{t}` in `import_sink.rs`. Anonymous nodes need to keep a per-chunk component so they don't unify. The `.jsonld` splitter path should get the same treatment for consistency.

fluree-db-api/src/import.rs
```diff
@@ -1216,6 +1338,298 @@ fn spawn_remote_producer(
...
+                                // Consumer aborted upstream — exit cleanly.
+                                let _ = error_tx.send(None);
+                                return;
+                            }
+                            global_idx += 1;
+                        }
+                        Ok(None) => break,
+                        Err(e) => {
+                            let _ = error_tx.send(Some(ImportError::Transact(format!(
+                                "ndjson read of {label}: {e}"
```
bplatz, line 1513: Malformed-line and IO errors from the producer surface as `ImportError::Transact`, reader-init as `NoChunks` — neither name matches the failure; a parse/source variant would diagnose better.

fluree-db-api/src/import.rs
```diff
@@ -1763,6 +2185,32 @@ pub enum DirectoryFormat {
...
+/// `.jsonl`/`.ndjson` (they take different parse paths).
+fn validate_format_families(
+    has_turtle: bool,
+    has_jsonld: bool,
+    has_ndjson: bool,
+    source_desc: &str,
+) -> std::result::Result<(), ImportError> {
+    if has_turtle && (has_jsonld || has_ndjson) {
+        return Err(ImportError::MixedFormats(format!(
+            "{source_desc} contains both Turtle (.ttl/.trig) and JSON-LD \
```
bplatz: Family lists here are incomplete: Turtle says "(.ttl/.trig)" (omits .nt/.nq), JSON-LD says "(.jsonld/.jsonl)" (omits .ndjson).

bplatz (approved):
Comments in line, but otherwise looks good!

fluree-graph-json-ld/src/ndjson_splitter.rs
aaj3f: Addressed -- I actually went with your comment in-office yesterday that any single-line context line now simply replaces whatever top-level context preceded it in that file

fluree-db-api/src/import.rs
aaj3f: Added `ImportError::Source(String)`

fluree-db-api/src/import.rs
aaj3f: Fixed — the messages now name the full families: `(.ttl/.nt/.nq/.trig)` and `(.jsonld/.jsonl/.ndjson)`. Since the mixing rules were consolidated into a single `validate_format_families` shared by `scan_directory_format` and `resolve_remote_objects`, one edit covers both the local and remote error paths. While in there, the two `NoChunks` extension lists were also missing `.ndjson`, so those are completed too.

PR #1308, opened 2026-06-11
feat: SPARQL per-query reasoning, OWL2-RL correctness at scale, and overlay query performance

bplatz:
## Summary

Makes OWL2-RL reasoning correct and usable at scale, and gives SPARQL per-query reasoning control.

### SPARQL per-query reasoning

Queries can now request reasoning via a comment directive:

```sparql
# PRAGMA reasoning: owl2rl
SELECT ?x WHERE { ?x a ub:Student }
```

(also `rdfs`, `owl2ql`, `datalog`, `owl-datalog`, `none`, or a comma-separated list — implementing the documented-but-unimplemented `PRAGMA reasoning`). Directives are extracted from lexer-identified comments, so a `#` inside string literals or IRIs is never misread and the query stays valid SPARQL for standard tooling. Query-level modes participate in ledger-config precedence exactly like the JSON-LD `"reasoning"` field; unknown modes error rather than silently no-op.

### Materialization correctness and visibility

- A materialization that hits its budget (default 1M facts / 30s) is an **incomplete closure**, but was logged as "completed" with no signal — and the budget was hardcoded with no way to raise it. Confirmed on LUBM-25 (3.4M triples): the true closure is 1.58M facts; the cap truncated it at 1.31M, silently dropping restriction-class entailments while subclass entailments survived (q08 7790→5916, q10 4→0, q12 15→0). Capped materializations now WARN on every query that uses them, and `FLUREE_REASONING_MAX_FACTS` / `FLUREE_REASONING_MAX_SECONDS` provide an operator override until the budget is wired into ledger config.
- `enabled_rules` now actually filters rule families (W3C OWL2-RL identifiers; `cax-eqc1/2` and `cls-maxqc3/4` accepted as aliases). Previously every rule always ran and the option only perturbed the cache key.
- Seeded base facts are excluded from the derived overlay (previously the entire rdf:type extent was duplicated into it), and budgets/diagnostics count genuine derivations only.

### Reasoner efficiency

prp-fp/prp-ifp keep persistent grouping state across fixpoint iterations instead of rescanning delta ∪ derived each pass; sameAs canonical lookups become single-hop after each equivalence change; the cached overlay is handed to execution as built instead of being re-collected and re-sorted four ways on every query.

### Query-side overlay performance

Reasoning queries attach the whole materialization as an overlay, and three per-probe costs made nested-loop joins pathological (LUBM-1 q08/q09 exceeded 300s):

- every scan open re-translated the entire overlay through the dictionaries — now memoized per execution and shared across queries via a small keyed LRU;
- range cursors carried all overlay ops (losing every leaflet pre-skip and paying an O(overlay) merge walk per probe) — ops are now windowed to the cursor's key range with an O(1) per-leaflet overlay gate;
- leaves were walked linearly from the first directory entry — range cursors now binary-search the starting leaflet and early-break once past the range.

Results — LUBM-1: q08/q09 >300s → 1.5s/1.9s, q02 36s → 131ms, q12 60s → 154ms, all 14 queries correct. LUBM-25 (3.4M triples, 1.58M derived facts): q01 72ms, q07 8ms, q11 10ms, q12 88ms, q06 (260k rows) 0.8s.

Building the cross-query cache exposed a latent aliasing bug: `ReasoningOverlay`'s combined epoch collided with the plain-novelty epoch on freshly-indexed ledgers (both 0) and aliased materializations built under different rule configs. `DerivedFactsOverlay` now carries a process-unique instance id folded into a high-bit-tagged combined epoch, disjoint from plain novelty counters.

### Notes for review

- The three `it_query_values` / `it_query_jsonld` VALUES failures pre-exist on the base branch (verified at the merge base) and are unrelated to this change.
- Remaining LUBM outliers (q02, q09) are per-row-probe bound — the overlay disables the batched join paths; overlay-aware batched probes are follow-up work.


aaj3f (approved):
@bplatz -- I apologize, I neither am familiar enough with the reasoning code, nor do I have an abundance of time before being OOO tomorrow but wanted to get you a review, so I delegated to claude and am sharing the output below

---

````
### 🔴 CRITICAL (correctness regression — must fix before merge)

#### 1. `fluree-db-reasoner/src/execute/property_rules.rs:602` (and `:722`) + `fluree-db-reasoner/src/fixpoint.rs:227`
**Semi-naive prp-fp/prp-ifp state misses entailments after a B.6 sameAs merge.**

The fold branch (`} else {` at 602/722) assumes: *"Canonical forms are stable: only this iteration's
delta can introduce conflicts."* That invariant does not hold. `same_as_changed` is recomputed fresh
each iteration at Phase A (`fixpoint.rs:163`), but sameAs facts derived by the identity rules are
unioned at **B.6** (`fixpoint.rs:220-227`) — *after* Phase B already ran. The next iteration's
Phase A re-unions those same pairs, which are no-ops, so it reports **no change** — and the fold
branch runs against a `rep_map` whose keys are canonicals from *before* the merge. A new delta fact
canonicalizing to the merged root misses the stale entry entirely: vacant insert, no conflict, no
`sameAs` derivation.

**Repro** (written during this review, left in the tree):
`fluree-db-reasoner/tests/prp_fp_stale_state_repro.rs`

```
cargo test -p fluree-db-reasoner --test prp_fp_stale_state_repro
```

**FAILS at HEAD (`a4319fe`), PASSES at merge base (`405dfb4`)** → regression from commit `6cfc5f4`.

*Test conditions.* The harness is the same in-memory `SortedOverlay` + `reason_owl2rl` setup as the
existing `lubm_defined_class_repro.rs` (empty genesis snapshot, all facts supplied via overlay,
`ReasoningOptions::default()`, fresh `ReasoningCache`). The input is seven triples — three ontology,
four data (all `Ref`-valued, t=1, namespace `ex:` = code 100):

```turtle
# ontology
ex:p  rdf:type            owl:FunctionalProperty .
ex:q  rdf:type            owl:InverseFunctionalProperty .
ex:r  rdfs:subPropertyOf  ex:p .

# data
ex:B  ex:p  ex:y1 .      # pre-existing functional-property fact
ex:A  ex:q  ex:v .       # \  prp-ifp: same object v under inverse-functional q
ex:B  ex:q  ex:v .       # /  → sameAs(A, B)
ex:A  ex:r  ex:y2 .      # prp-spo1: r ⊑ p  →  A p y2 (arrives one iteration later)
```

Required OWL2-RL derivation chain:

1. **prp-ifp** (`A q v`, `B q v`) → `sameAs(A, B)` — derived in iteration 1's Phase B,
   **unioned at B.6**, i.e. *after* prp-fp already folded `{canonical(B) → y1}` into its
   persistent `rep_map` that same iteration.
2. **prp-spo1** (`A r y2`, `r ⊑ p`) → `A p y2` — lands in iteration 2's delta.
3. **prp-fp** (iteration 2): the merged subject class `{A, B}` now has p-objects `{y1, y2}` →
   **must derive `sameAs(y1, y2)`**. This is where HEAD fails: Phase A of iteration 2 re-unions
   the already-merged `(A, B)` pair as a no-op, reports `same_as_changed == false`, so the fold
   branch runs; `canonical(A)` misses the stale `rep_map` key (recorded under pre-merge
   `canonical(B)`), vacant-inserts `y2`, and detects no conflict.

*Assertions* (the test runs the scenario twice with `A`/`B` swapped — `(B p y1, A r y2)` and
`(A p y1, B r y2)` — so it cannot pass by luck of union-find root selection; a correct prp-fp must
derive the conflict in **both** mirrors):

```rust
// sanity: prp-ifp must have merged A and B (passes on both branches)
assert_eq!(same_as.canonical(ex("A")), same_as.canonical(ex("B")));

// the regression: y1 and y2 must be in the same equivalence class
assert!(same_as.canonical(ex("y1")) == same_as.canonical(ex("y2")),
        "prp-fp must derive sameAs(y1, y2) after the A≡B merge");
```

*Observed at HEAD* (mirror 1): the final derived overlay contains
`A sameAs B`, `B sameAs A`, **`A p y1`, `A p y2`** — i.e. the materialization itself proves the
merged subject has two distinct objects under a functional property — yet no `sameAs` fact links
`y1` and `y2`, and `FrozenSameAs` keeps them in separate classes. At the merge base the identical
input yields `y1 ≡ y2` as required.

Note LUBM would never catch this: Univ-Bench has no inverse-functional properties or
maxCardinality-1 restrictions, so the identity-rule cascade never fires there.

**Suggested fix** — carry B.6's union outcome into the next iteration's flag so the rebuild branch
(which recomputes canonicals from scratch and refreshes the state) runs:

```rust
// fixpoint.rs — before the loop
let mut same_as_changed_carry = false;

// top of each iteration (Phase A)
let mut same_as_changed = std::mem::take(&mut same_as_changed_carry)
    | apply_same_as_rule(&delta, &mut same_as_tracker, &owl_same_as_sid, &mut diagnostics);

// at B.6
let identity_same_as_changed = apply_same_as_rule(&new_delta, ...);
same_as_changed = same_as_changed || identity_same_as_changed;
same_as_changed_carry = identity_same_as_changed;
```

Also consider extending the loop condition to `while !delta.is_empty() || same_as_changed_carry` —
see finding #2, which the carry-flag alone does not cover when the merge happens on the final
iteration.

After fixing, please also keep (a possibly tidied version of) the repro test, and consider a
mirrored ifp-direction case in `property_rule_tests.rs` exercising fold→merge→fold across three
iterations.

### 🟠 MAJOR (correctness, pre-existing — fix here or file an issue)

#### 2. `fluree-db-reasoner/src/fixpoint.rs:220-227` — B.6 merges never re-trigger identity rules at all when no new delta facts arrive
Distinct from #1 (and **pre-existing** — the old code has the same gap): if the *last* fixpoint
iteration derives `sameAs(A,B)` via prp-ifp and no other new facts survive dedup, the loop exits
(`delta` empty) without ever re-running prp-fp over the merged class — e.g. base facts
`A p y1 ; B p y2` (`p` functional) never yield `sameAs(y1,y2)`. The old always-rebuild code only
caught the cascade when an unrelated fact happened to keep the loop alive and `delta_has_p` happened
to be true. Since this PR's stated goal is OWL2-RL *correctness* and it already touches this exact
loop, extending the loop condition as suggested in #1 closes both holes at once. If deferred, file
it — silent incompleteness is precisely what this PR set out to eliminate.

### 🟡 MINOR (worth doing in this PR; none block on their own)

#### 3. `fluree-db-reasoner/src/lib.rs:119` — unknown rule names in `enabled_rules` silently disable everything they were meant to enable
`rule_enabled_any` matches by string equality against a documented vocabulary, but nothing validates
the caller's strings. `enabled_rules: vec!["cax-sco1"]` (typo) silently runs *no* cax-sco — the
exact "option silently no-ops" failure mode this PR fixes elsewhere (unknown pragma modes error;
`enabled_rules` used to be inert). Suggest validating against the known-identifier set at
`reason_owl2rl` entry (or in `ReasoningOptions`) and returning an error listing the unknown names.

#### 4. `fluree-db-query/src/binary_scan.rs:2037` — clone-then-retain of `untranslated` per scan open
`translated.untranslated.clone()` copies the whole `Vec<Flake>` on every scan open before `retain`
filters it per-pattern. Reasoning overlays normally keep this list tiny (`@vector` etc.), but the
known degenerate case (uninitialized `DictNovelty` → *every* overlay flake untranslated) would pay a
full clone per probe row — the same shape of cost this PR just eliminated for `ops`. Build the
filtered vec directly instead:

```rust
let untranslated: Vec<Flake> = translated.untranslated.iter()
    .filter(|f| /* s_sid / p_sid / bound_o / object_bounds checks */)
    .cloned()
    .collect();
```

#### 5. `fluree-db-query/src/binary_scan.rs:2294` — `TranslationCache::new(4)` is count-bounded, not byte-bounded
Each entry at LUBM-25 scale is ~1.58M ops (≈60–100 MB) and the key includes the index, so **one**
hot reasoning ledger queried across SPOT/POST/OPST/PSOT can occupy all 4 slots; a second concurrent
reasoning ledger thrashes the LRU, and worst case ~400 MB sits pinned in a process-global static.
The codebase precedent is byte-budgeted caches (`FLUREE_LEAFLET_CACHE_BYTES`, moka-based
`LeafletCache`). Suggest at minimum an env override for capacity (matching the
`FLUREE_REASONING_*` pattern introduced here), ideally a byte-weighted bound; also worth a `debug!`
on eviction so thrash is diagnosable. Not blocking: the entries are `Arc`-shared so eviction never
invalidates in-flight queries, and 4 was a deliberate "one or two hot states" choice per the doc
comment.

#### 6. `fluree-db-query/src/execute/reasoning_prep.rs:171` — env vars re-read on every reasoning query
`reasoning_budget_from_env()` runs per `compute_derived_facts` call. Cost is trivial (two
`env::var`s ahead of a materialization-cache lookup), and re-reading does make live tuning possible
— but if that's not intended, a `once_cell::Lazy` would make the budget immutable-per-process and
self-documenting. Either way, a one-line comment stating the choice would stop future readers from
"fixing" it in either direction. Also: invalid values (`FLUREE_REASONING_MAX_FACTS=abc`) silently
fall back to defaults — a `warn!` would match this PR's no-silent-failure ethos.

#### 7. `fluree-db-reasoner/src/fixpoint.rs:382` — pre-existing self-assignment no-op loses `rules_fired`
`diagnostics.rules_fired = std::mem::take(&mut diagnostics.rules_fired);` — `diagnostics` was just
*replaced* by `ReasoningDiagnostics::completed(...)` on the line above, so the rule-fire counts
accumulated during the loop are dropped and this line takes-and-reassigns from the already-fresh
struct (a no-op). Pre-existing, but this PR now documents `rules_fired` as the `enabled_rules`
vocabulary reference, and the diff touches the adjacent line. Intended shape is presumably:

```rust
let rules_fired = std::mem::take(&mut diagnostics.rules_fired);
diagnostics = ReasoningDiagnostics::completed(iterations, derived.derived_len(), start.elapsed());
diagnostics.rules_fired = rules_fired;
```

(The capped paths have the same loss; `capped()` likewise constructs fresh.)

#### 8. `fluree-db-binary-index/src/read/binary_cursor.rs:61` — cursor `epoch` is write-only
`set_epoch` is still called from `binary_scan.rs:2029`, but nothing reads `self.epoch` anywhere in
the crate ("cache key differentiation" no longer exists at this layer — leaflet caching keys on
`LeafletDecodeSpec` pre-merge). Dead state; consider removing field + setter in a follow-up rather
than carrying the suggestion that batches are epoch-keyed here.

### 💬 Notes / questions (no action required)

- **`fluree-db-sparql/src/lex/lexer.rs:34`** — `tokenize` now allocates a `String` per comment and
  discards them. Comments are rare in queries and parsing isn't per-row hot, so fine; if it ever
  shows up, an `Option<&mut Vec<String>>` sink keeps the common path allocation-free.
- **`fluree-db-query/src/reasoning.rs:26`** — the masked combine
  `base*1000003 + instance_id` can in principle collide across (base_epoch, instance_id) pairs, but
  a collision additionally requires identical `(ledger, snapshot_t, store_max_t, to_t)` in every
  consuming cache key — practically unreachable. The high-bit tag also incidentally fixes
  `overlay_free_single_graph()` (context.rs:763) treating a fresh-ledger ReasoningOverlay
  (old combined epoch 0) as overlay-free, which would have let batched join paths drop derived
  facts — worth a sentence in the commit history/changelog since it's a user-visible correctness
  improvement.
- **`fluree-db-query/src/binary_scan.rs:1948-2005`** — the per-execution mutex is held across the
  global-cache miss + full translation. Intra-query that's the right dedupe; across *queries* two
  concurrent misses translate redundantly (global cache get/insert are separate). Acceptable
  first-storm cost; an entry-future pattern could dedupe later if it shows up.
- `count_plan_exec.rs` boxing of `SoRows` variants: straightforwardly correct (clippy
  `large_enum_variant`).
- Pragma design choice — `# PRAGMA reasoning: rdfs` *anywhere* in the query (including trailing)
  is honored, and last-wins on repeats: both documented and tested; fine, just flagging the
  semantics since "pragma must precede the query" is the convention some users may assume.
````

bplatz:
Thanks for the review — fixes landed on this branch:

- **#1 (critical, stale prp-fp/prp-ifp state) + #2 (final-iteration merge)**: 4dbe87a268bc9c365f1b3e1d1b8cf9e372fe1ed1 (carry flag + extended loop condition; repro tests recreated incl. both mirrors and the final-iteration case)
- **#7 (rules_fired dropped)**: da2c179f06fb3f9ec807a7fe5fdff28a1c538cd6 (in-place finalizers; capped paths preserved too)
- **#3 (enabled_rules validation), #4 (clone-then-retain), #6 (env re-read docs + warn)**: a9a8df864fcbab876fdaf5522bb0c60e68d0be29
- **#8 (write-only cursor epoch)**: 05a04936f705db1e795fcc7ef8332305deb4db7b (field, setter, all call sites, and GeoSearch's feeder field removed)

PR #1312, opened 2026-06-11
fix(api): resolve novelty-allocated encoded IDs in result formatting

### LINKED ISSUE #1310: SPARQL VALUES ?c {…} ?s a ?c returns deleted/historical subjects (logical DELETE not honored in this query shape)
## Summary
On a ledger that was emptied with a SPARQL `DELETE` and then re-populated, a class-membership count returns **53** when the class is supplied via `VALUES` (variable in the `rdf:type` position), but **5** when the class is a constant IRI — on the **same ledger, same state**. The extra 48 subjects are entities that were **deleted** from the ledger (they have **0 triples** in the current state, and 0 in every other ledger on the server). They appear to be resurfaced from the ledger's pre-DELETE history.

## Environment
- Fluree Core `fluree/server:4.0.5`
- Endpoint: `POST /v1/fluree/query/{ledger}`, `Content-Type: application/sparql-query`
- Ledger `cam-output-flat`: created → inserted ~16,965 triples → `DELETE { ?s ?p ?o } WHERE { ?s ?p ?o }` → re-inserted ~2,337 triples → inserted an RDFS / `f:DataModel` schema

## Steps to reproduce
```sparql
-- (A) constant class  -> 5  (correct: current state)
SELECT (COUNT(DISTINCT ?s) AS ?n) WHERE { ?s a <http://Magna/EngOntC#Part> }

-- (B) class via VALUES -> 53 (wrong: includes deleted/historical subjects)
SELECT (COUNT(DISTINCT ?s) AS ?n)
WHERE { VALUES ?c { <http://Magna/EngOntC#Part> } ?s a ?c }

-- (C) list them
SELECT DISTINCT ?s WHERE { ?s a ?c . VALUES ?c { <http://Magna/EngOntC#Part> } }
```

`(C)` returns 53 subjects. Spot-check any of the 48 extra ones — e.g. `http://flur.ee/cam/individual/Beikappantriebes`:

```sparql
SELECT (COUNT(*) AS ?n) WHERE { <http://flur.ee/cam/individual/Beikappantriebes> ?p ?o }
SELECT (COUNT(*) AS ?n) WHERE { ?s ?p <http://flur.ee/cam/individual/Beikappantriebes> }
```

Both return **0**. The same entity returns 0 triples on **every** ledger on the server, yet query (B)/(C) returns it on `cam-output-flat`.

## Expected behavior
(A) and (B) return the same result (**5**). A logically `DELETE`d assertion `<s> a <Class>` must not be matched by any current-state query, whether the class IRI is a constant or bound via `VALUES`.

## Actual behavior
- (A) = **5** (current state, correct)
- (B) / (C) = **53** (current 5 + 48 deleted subjects from the pre-DELETE version of the ledger)

## Scope — only the DELETE-then-reuse ledger is affected
| Ledger | `?s a engc:Part` (constant) | `VALUES ?c {…} ?s a ?c` |
| --- | --- | --- |
| `magna-organization-108` (never deleted) | 2121 | 2121 ✅ |
| `magna-organization-107` (never deleted) | 2116 | 2116 ✅ |
| `cam-output-0610` (same data, never deleted) | 5 | 5 ✅ |
| **`cam-output-flat` (DELETE then re-insert)** | **5** | **53 ❌** |

A twin ledger with identical current data but never emptied (`cam-output-0610`) is consistent (5 = 5). Only the ledger that went through `DELETE` shows the discrepancy.

## Impact
- Tooling that counts class membership via `VALUES` (e.g. to union several class IRIs / casings) silently returns historical/deleted data → wrong "how many X" answers (here 53 vs 5).
- Logical `DELETE` appears insufficient to remove data from query results in this code path; combined with the admin-only `/drop-graph`, a non-admin user has no reliable way to make deleted entities disappear from all query shapes.

## Suspected cause
The `VALUES`-bound `rdf:type` pattern seems to resolve against an index / flake set that still includes retracted (historical) assertions, whereas a constant-IRI `rdf:type` pattern resolves against current state only. Possibly amplified by RDFS reasoning being engaged when the class term is a variable (the affected ledger had a transacted schema with `rdfs:domain` / `rdfs:range` / `rdfs:Class`).

## Suggested checks
- Confirm both query shapes resolve at the same (current) `t`.
- Verify retracted assertions are excluded when the class term is bound via `VALUES` / a variable.
- Document whether `DELETE` is expected to purge from query results or only from the latest commit, and how to truly purge without admin `/drop-graph`.

bplatz:
Closes #1310

## Problem

On a ledger that was emptied with a wildcard DELETE and re-populated, the `VALUES ?c { <iri> } ?s a ?c` query shape misbehaved while the constant-IRI shape `?s a <iri>` was correct.

Two layers:

1. **v4.0.5 (as reported):** the batched-object join lane didn't merge the novelty overlay — post-delete it returned every deleted subject still in the index (the reported 53-vs-5 divergence) and post-reinsert it dropped novelty-only subjects. Already fixed on main by the `ObjectProbeOps` overlay-merge work (unreleased).
2. **Still broken at HEAD:** that lane emits novelty subjects as `EncodedSid` with `DictNovelty`-allocated IDs (above the persisted watermark), but formatters resolved encoded bindings against the on-disk forward packs only — so the issue's `SELECT DISTINCT ?s` shape failed with `InvalidBinding: subject local_id N not found in ns M`.

## Fix

- `GraphDb::binary_graph()` attaches the ledger's `DictNovelty`, so the view handed to formatters supports watermark routing
- `format/materialize.rs`, `format/delimited.rs`, `format/sparql_xml.rs` resolve subjects (and CSV string/ref literals) through the novelty-aware `BinaryGraphView` methods instead of raw store lookups

No transaction or query-execution changes. Formatting adds one Arc clone per query and a per-namespace watermark check before pack lookups; the CSV string fast path stays zero-copy, falling back to novelty decode only when the persisted lookup misses (novelty IDs sit above the watermark, so a miss can't silently alias).

## Tests

`it_values_type_deleted_repro.rs` pins constant vs `VALUES` agreement (counts and listed subjects) across four ledger states: novelty-only, indexed-before-delete, indexed-after-reinsert, and the production-like incremental-index timeline with an RDFS schema. The indexed-before-delete and incremental scenarios fail on v4.0.5 with the exact reported symptom and fail at HEAD with the materialization error before this fix.

### DIFF
--- a/fluree-db-api/src/format/delimited.rs
+++ b/fluree-db-api/src/format/delimited.rs
@@ -386,8 +386,7 @@ fn write_binding_cell(
         }
         Binding::EncodedSid { s_id, .. } => {
             let gv = require_graph_view(gv)?;
-            let store = gv.store();
-            let iri = store.resolve_subject_iri(*s_id).map_err(|e| {
+            let iri = gv.resolve_subject_iri(*s_id).map_err(|e| {
                 FormatError::InvalidBinding(format!(
                     "Failed to resolve subject IRI for s_id {s_id}: {e}"
                 ))
@@ -417,15 +416,21 @@ fn write_binding_cell(
             let gv = require_graph_view(gv)?;
             let store = gv.store();
             if *o_kind == ObjKind::LEX_ID.as_u8() || *o_kind == ObjKind::JSON_ID.as_u8() {
-                store
-                    .write_string_value_bytes(*o_key as u32, cell)
-                    .map_err(|e| {
-                        FormatError::InvalidBinding(format!(
-                            "Failed to resolve string (kind={o_kind}, key={o_key}): {e}"
-                        ))
-                    })?;
+                // Zero-copy persisted lookup first; novelty string IDs sit above
+                // the pack watermark so the store lookup fails cleanly (without
+                // writing) and the novelty-aware decode takes over.
+                if store.write_string_value_bytes(*o_key as u32, cell).is_err() {
+                    let val = gv
+                        .decode_value_from_kind(*o_kind, *o_key, *p_id, *dt_id, *lang_id)
+                        .map_err(|e| {
+                            FormatError::InvalidBinding(format!(
+                                "Failed to resolve string (kind={o_kind}, key={o_key}): {e}"
+                            ))
+                        })?;
+                    write_flake_value(cell, &val, compactor);
+                }
             } else if *o_kind == ObjKind::REF_ID.as_u8() {
-                let iri = store.resolve_subject_iri(*o_key).map_err(|e| {
+                let iri = gv.resolve_subject_iri(*o_key).map_err(|e| {
                     FormatError::InvalidBinding(format!(
                         "Failed to resolve ref IRI for s_id {o_key}: {e}"
                     ))
--- a/fluree-db-api/src/format/materialize.rs
+++ b/fluree-db-api/src/format/materialize.rs
@@ -41,9 +41,10 @@ fn materialize_encoded_binding(
 ) -> std::io::Result<Binding> {
     let store = gv.store();
     match binding {
+        // Novelty-aware: subjects first seen after the index snapshot resolve
+        // from DictNovelty (watermark routing), not the persisted packs.
         Binding::EncodedSid { s_id, .. } => {
-            let iri = store.resolve_subject_iri(*s_id)?;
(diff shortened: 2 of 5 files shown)


zonotope (approved):
👕

PR #1311, opened 2026-06-11
Phase 0.0 perf guardrails: baseline capture/compare, allocation tracking, overlay-matrix bench

aaj3f:
## Summary

First PR of the architecture-audit roadmap (`docs/audit/2026-06-architecture-audit.md`, included here). Implements **Phase 0.0 — performance guardrails**: the measurement substrate every later refactor phase gates on, so "no regression, ideally improvement" is a mechanical check (time **and** memory, across scales and ledger conditions) instead of a hope.

## What's included

1. **Architecture audit + roadmap** (`docs/audit/2026-06-architecture-audit.md`) — full-workspace audit; core finding is that the three chronic-bug contracts (overlay translation/merge, LedgerState coherence, fast-path gating) are convention-enforced rather than type-enforced. Phases 0–4 convert them to compile-time / single-chokepoint guarantees without touching hot kernels. Phase 0.0 defines the per-phase baseline protocol this PR implements.
2. **`bench-baseline`** (bin in `fluree-bench-support`) — `capture` walks criterion estimates + memory sidecars into a labeled, git-stamped `bench-baselines/<label>.json`; `compare` checks a later run against it using `regression-budget.json` budgets, prints regressions *and* improvements, exits 1 on breach. `--only` scopes PR-level subset runs. See BENCHMARKING.md "Baselines: capture & compare".
3. **`fluree-bench-alloc`** — tracking `GlobalAlloc` wrapper (current/peak/total; two relaxed atomics per alloc) in its own crate so `fluree-bench-support` keeps `#![forbid(unsafe_code)]`. New `mem` module records per-scenario sidecars; `compare` gates `peak_mem` with the same budgets as time.
4. **`query_overlay_matrix` bench** — count / star-join / group-by shapes at three ledger conditions: `base` (indexed, epoch 0), `overlay` (indexed + ~10% trailing novelty), `novelty` (no index). Closes the coverage gap where every existing hot-path bench ran at epoch 0 — i.e. the overlay-merge lane the roadmap's Phases 1–2 refactor had **zero** benchmark coverage.

## Why the matrix matters — first measurements

Tiny scale (100 products, quick profile), base → overlay:

| shape | base | overlay | penalty |
|---|---|---|---|
| count | 31 µs | 273 µs | ~9× |
| star | 63 µs | 911 µs | ~14× |
| groupby | 154 µs | 19.8 ms | ~128× |

[...]

PR #1313, opened 2026-06-12
Phase 0.7 differential fast-path harness (+ kill switch) — finds 3 divergence classes

aaj3f:
## Summary

Stacked on #1311. Implements **Phase 0.7** of the audit roadmap: the differential correctness harness that gates every later refactor phase — and it earned its keep immediately by finding **three reproducible fast-vs-generic divergence classes on its first run**, two of which are user-visible wrong answers.

## What's included

1. **Planner fast-path kill switch** (`fluree-db-query`): `set_fast_paths_disabled(bool)` / `FLUREE_DISABLE_QUERY_FAST_PATHS` skips every `detect_*` shape recognizer (the fused chain *and* the three non-fused history-gated paths), forcing the generic pipeline. Re-exported through `fluree-db-api`. Runtime operator-internal optimizations are unaffected — this isolates exactly the planner fast-path layer.
2. **Differential harness** (`fluree-db-api/tests/it_differential_fastpath.rs`), two axes over a dataset with deliberate overlay churn (upserts that retract/replace indexed values and dominate `ORDER BY DESC` top-k; novelty-only subjects/strings; duplicate re-asserts):
   - **Axis 1 — fast ≡ generic**: 16 catalog queries (each annotated with the `detect_*` path it targets) run with fast paths on/off per condition; results must match.
   - **Axis 2 — representation equivalence**: generic results must agree across `base` (epoch 0), `overlay` (trailing novelty), and `novelty` (no index) conditions.

## Findings (first run, all reproducible)

**Axis 2 passed on all 16 cases** — the historical overlay/dict-novelty bug classes stay fixed across all three physical representations. Axis 1 found:

| ID | Case | What diverges | Severity |
|---|---|---|---|
| FD-1 | `avg_numeric` | fast emits f64 JSON number; generic emits arbitrary-precision `xsd:decimal` string | datatype contract drift |
| FD-2 | `max_label` / `min_label` | **wrong answers** on reindex-built ledgers: fast returns the last-inserted label instead of the lexicographic max — consistent with treating string-ID order as lex order (`lex_sorted_string_ids` only holds after bulk import) | **high — wrong results** |
| FD-3 | `group_by_predicate_count` | stats-answered path counts duplicate re-asserts (base), applies novelty deltas inconsistently per predicate (overlay), and drops the `rdf:type` row entirely (pure novelty) | **high — wrong results** |

Each is pinned via a `known_divergence` marker: the mismatch is reported to stderr but not enforced, so the harness is green while pinning the contract for everything else. Full analysis in the test's module docs. **When a fix lands, remove the marker so the case becomes enforced.**

These land squarely on the audit's thesis: FD-2 is a convention-gate miss (an invariant that only sometimes holds, checked nowhere), FD-3 is stats/novelty accounting divergence — both in the catalog of risks the roadmap's Phases 1–3 are designed to make structurally impossible.

## Validation

- Harness passes: 16 cases × 3 conditions × 2 modes (96 executions) + cross-condition assertions, ~16s
- All 1,045 `fluree-db-query` unit tests pass with the kill switch in place
- Clippy + fmt clean on touched crates

## Suggested follow-ups (not in this PR)

1. Fix FD-2 (`fast_min_max_string` lex-order gate) — wrong results, likely small fix
2. Fix FD-3 (`StatsCountByPredicateOperator` accounting) — wrong results
3. Fix FD-1 (AVG datatype fidelity)
4. Confirm the four lifecycle bug candidates flagged in the audit (`apply_loaded_db` graph-registry replay, `apply_index_v2` store sync, reload race, export translation drops)

### DIFF
--- a/fluree-db-api/src/lib.rs
+++ b/fluree-db-api/src/lib.rs
@@ -228,6 +228,9 @@ pub use fluree_db_query::{
     execute, execute_pattern, Batch, ContextConfig, ExecutableQuery, NoOpR2rmlProvider, Pattern,
     ReasoningConfig, VarRegistry,
 };
+// Planner fast-path kill switch — differential correctness harness
+// (tests/it_differential_fastpath.rs) and operational triage.
+pub use fluree_db_query::{fast_paths_disabled, set_fast_paths_disabled};
 // Re-export for lower-level pattern-based queries (internal/advanced use)
 pub use fluree_db_query::{Term, TriplePattern};
 // Re-export parse types for query results
--- a/fluree-db-api/tests/it_differential_fastpath.rs
+++ b/fluree-db-api/tests/it_differential_fastpath.rs
@@ -1,0 +1,506 @@
+//! Differential fast-path correctness harness (audit roadmap Phase 0.7,
+//! `docs/audit/2026-06-architecture-audit.md`).
+//!
+//! The planner has 30+ `detect_*` fast paths whose results must be
+//! indistinguishable from the generic operator pipeline, and the engine has
+//! three physical representations of the same logical data (binary index,
+//! index + novelty overlay, pure novelty) whose query results must be
+//! indistinguishable from each other. Both contracts are exactly where the
+//! historical correctness bugs lived. This harness pins them:
+//!
+//! **Axis 1 — fast vs generic.** Every catalog query runs twice per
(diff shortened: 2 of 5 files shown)


aaj3f:

PR #1314, opened 2026-06-12
Fix the three fast-path divergences (FD-1/2/3) + scenario-attributable memory gating

aaj3f:
## Summary

Stacked on #1313. Fixes all three divergence classes the differential harness found on its first run, removes their `known_divergence` markers so the harness now **hard-enforces** them across the base/overlay/novelty condition matrix, and improves the Phase 0.0 memory gate based on what dogfooding it on this branch revealed.

## The fixes

**FD-2 — MIN/MAX-string wrong answers** (`fast_min_max_string.rs`): the path took per-leaflet first/last POST directory keys as candidates — extremes by StringId, not by lexicographic value. Sound only under the bulk-import `lex_sorted_string_ids` invariant; on reindex-built ledgers it returned the first/last *inserted* label. Fix: the same lex-order gate `fast_string_prefix_count_all` already used; otherwise defer to generic. One missed convention gate — exactly the audit's predicted failure mode.

**FD-3 — stats-answered per-predicate counts wrong three ways** (`stats_query.rs` rewrite): `StatsView` numbers are planner estimates (count duplicate re-asserts, track `rdf:type` asymmetrically, apply novelty deltas inconsistently) — fine for selectivity, wrong as query answers. Rewritten to count **exactly** from POST leaf-directory metadata (`row_count` per homogeneous leaflet; PId-column decode only for mixed leaflets; V3 leaflets are current-state rows so directory counts are exact), gated by `fast_path_store` at epoch-0/at-head with the generic pipeline as planned fallback. The dispatch joins the fused chain (prevents fallback-recursion re-entry) and the detector declines OFFSET (the one non-idempotent modifier under fallback re-wrapping).

**FD-1 — AVG datatype drift** (`fast_predicate_scalar_agg.rs`): fast path emitted Kahan-f64 `xsd:double`; generic emits exact `xsd:decimal` for integer inputs per SPARQL 1.1. The accumulator now mirrors `finalize_avg` per number kind: exact i64 + BigDecimal division at the shared `AVG_DECIMAL_PRECISION` for integers (overflow → fallback), plain non-Kahan f64 for doubles (bit-parity with generic), fallback for other numeric encodings.

## Guardrail dogfooding → metric improvement

Running this branch through the Phase 0.0 bench gate flagged 9 `peak_mem` breaches: a **uniform** ~+1.6 MiB across every scenario, including pure-novelty lanes that execute none of the changed code. Classification: same-binary rerun variance is ±0.04% (metric is stable); the shift is ambient process heap moving with the recompiled binary, which absolute peak inherits since it measures from process zero. **Time was within noise on all 9 scenarios (max +1.7%).**

Fix: `fluree-bench-alloc` now records live-bytes-at-reset and exposes `scenario_peak_bytes` (peak − scenario start); `bench-baseline compare` gates on `scenario_mem` when both sides carry it (legacy absolute fallback otherwise). Verified ±0.03% across same-code runs.

## Validation

- Differential harness fully enforced (zero markers): 16 cases × 3 conditions × 2 modes, green
- All 1,045 `fluree-db-query` unit tests green
- Bench gate: time within noise on all scenarios; `scenario_mem` stable ±0.03%
- Clippy + fmt clean on all touched crates

## Remaining from the audit's near-term queue

- Underlying `StatsView` accounting inaccuracies (dup re-asserts, rdf:type asymmetry) still exist — now harmless for query answers, but worth fixing for planner estimate quality (filed in audit Phase 1 context)
- The four lifecycle bug candidates (`apply_loaded_db` graph-registry replay, `apply_index_v2` store sync, reload race, export translation drops) — next up

### DIFF
--- a/.fluree-memory/repo.ttl
+++ b/.fluree-memory/repo.ttl
@@ -1307,6 +1307,22 @@ mem:fact-01kj8k5nzpe59p5vp7zrehbrr9 a mem:Fact ;
     mem:branch "feature/memory" ;
     mem:createdAt "2026-02-24T19:49:14.358637+00:00"^^xsd:dateTime .
 
+mem:constraint-01ktwn58n7jgsxdax2rsxv7rg2 a mem:Constraint ;
+    mem:content "Bench memory gating must use scenario_peak_bytes (peak minus live-bytes-at-reset), not absolute peak_bytes: absolute peak measures from process zero and inherits ambient heap shifts that a recompile moves uniformly (~+1.6MiB observed across ALL scenarios including untouched code paths), while same-binary variance is ±0.04%. bench-baseline compare prefers scenario_mem when both sides carry it. Also: StatsView/IndexStats property counts are planner estimates (count dup re-asserts, rdf:type tracked asymmetrically between indexer and novelty accumulation) — never answer user queries from them." ;
+    mem:tag "benchmarking" ;
+    mem:tag "gotcha" ;
+    mem:tag "guardrails" ;
+    mem:tag "memory-metrics" ;
+    mem:tag "stats" ;
+    mem:scope mem:repo ;
+    mem:severity "must" ;
+    mem:artifactRef "fluree-bench-alloc/src/lib.rs" ;
+    mem:artifactRef "fluree-bench-support/src/baseline.rs" ;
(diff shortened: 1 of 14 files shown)


aaj3f:

PR #1315, opened 2026-06-12
Lifecycle bug-candidate triage: one real fix, one regression pin, two false positives

aaj3f:
## Summary

Stacked on #1314. Closes out the four lifecycle bug candidates flagged by the architecture audit (`docs/audit/2026-06-architecture-audit.md` §3.3 / Phase 0.3): each was either fixed, pinned, or affirmatively cleared with the analysis recorded here. Net: **one real data-loss fix, one regression test pinning behavior that already worked, two false positives** whose clearing is itself valuable — it corrects the audit record.

## Candidate 4 — REAL, fixed: silent data loss on failed overlay translation

`translate_overlay_flakes_with_untranslated` (`binary_scan.rs`) routed `Unsupported` failures into the untranslated raw-flake stream but **WARN-dropped every other failure kind** (`NotFound` from a stale/detached DictNovelty, `InvalidData`). A dropped flake lands in neither the overlay ops nor the raw stream — the fact silently vanishes from export and scan output under exactly the degraded state the warning describes. Same bug class as the previously-fixed graph-crawl disappearing-properties issue, surviving in this one path.

Fix: all failures keep the raw flake; both callers already lifecycle-resolve, filter, and stream untranslated flakes after the cursor. Widening is safe — failures only occur for identities the persisted dicts can't resolve, so matching base rows (which always translate) are unaffected, and assert/retract pairs fail together and resolve within the raw set. Not reachable via the public API without corrupted dictionary state, hence no repro test; Phase 1's translation chokepoint gives this a typed home.

## Candidate 1 — already works, now pinned: graph-registry survival through `apply_loaded_db`

The audit flagged `apply_loaded_db` as possibly losing named-graph registrations from commits still in novelty. Current code preserves them (old IRIs merged via `graph_registry.apply_delta` alongside namespace codes) — the audit agent's read was stale. New regression test `it_graph_registry_apply_loaded_db.rs` pins it, including the property that actually matters: surviving graphs keep their **original `g_id`s**, since remaining novelty flakes are stored under those numeric ids.

## Candidates 2 & 3 — false positives, cleared with analysis

- **`apply_index_v2` snapshot/binary-store TOCTOU**: the out-of-band `binary_store` write (`ledger_manager.rs:351`) happens *inside* the block holding the `state` write lock; readers take `state.read()` first (documented lock ordering), so new-state/old-store is unobservable. The audit misread the scoping.
- **Reload `t()` swap race**: the `new_state.t() >= current.t()` comparison executes *after* `lock_for_write()`, so no commit can move the in-memory state mid-comparison; when in-memory is newer the swap is skipped (novelty-protective, explicitly documented at the site). No lost commits.

Related: the audit's Arc-identity concern (`BinaryRangeProvider` capturing `state.dict_novelty`) has no live bug — the provider is rebuilt at every mutation point (commit path `commit.rs:707`, index apply, reload). The *structural* exposure (correctness depends on remembering to rebuild) stands as Phase 2 motivation, unchanged.

## Validation

- New regression test green; all 1,045 `fluree-db-query` unit tests green
- Differential harness (fully enforced) green — the translation fix doesn't perturb healthy-state results
- Export (`it_flpack_round_trip`) and `it_named_graph_isolation` green
- Clippy + fmt clean

### DIFF
--- a/.fluree-memory/repo.ttl
+++ b/.fluree-memory/repo.ttl
@@ -1513,13 +1513,13 @@ mem:fact-01ktwep3pg2dxz9r44083zdrxy a mem:Fact ;
     mem:rationale "Anchor document for any future refactor work; recall before structural changes to query/ledger/binary-index." .
 
 mem:fact-01ktwepk8ernqdycdrns8e8kyq a mem:Fact ;
-    mem:content "Audit flagged unconfirmed lifecycle/correctness candidates — verify before fixing: (1) apply_loaded_db (~ledger lib.rs:606-622) merges old namespace codes but may not replay graph_registry deltas from remaining novelty; (2) apply_index_v2 (~ledger_manager.rs:351) doesn't call sync_binary_store_from_state before releasing state lock (snapshot/store TOCTOU); (3) reload swap condition new_state.t() >= current.t() (~ledger_manager.rs:1157) races concurrent commits; (4) export translate path WARN-drops non-Unsupported translation errors (binary_scan.rs:2238-2242) — same silent-drop class fixed earlier in graph crawl. Also: DecodedRow.dt u32 vs OverlayOp/RunRecord.dt u16; GraphId is bare u16 alias." ;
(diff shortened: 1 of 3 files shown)


aaj3f:

PR #1316, opened 2026-06-12
Phase 0.1/0.2: GraphId newtype + dead dt-mismatch type deletion

aaj3f:
## Summary

Stacked on #1315. The audit roadmap's Phase 0.1/0.2 newtypes: `GraphId` goes from a bare `u16` alias to a `#[repr(transparent)]` newtype across the workspace (~400 sites, 120 files), and the audit-flagged dt-width mismatch resolves by **deleting dead code**.

## GraphId newtype (Phase 0.1)

`pub type GraphId = u16` made graph ids freely cross-assignable with the system's other pervasive `u16` spaces — namespace codes above all, the confusion class behind the historical namespace-after-index-attach bug. Now a newtype in the established `ids.rs` style (`DEFAULT`/`TXN_META` consts, `DictId` impl, `#[serde(transparent)]` so wire formats are unchanged; manual codecs convert via `as_u16`/`from_u16` at the boundary).

**The migration surfaced two real id-space seams the alias had been hiding:**

1. **transact:** `txn.graph_delta` keys are *transaction-local* graph ids — a different space from ledger `GraphId`s that flowed through the same alias name. Txn-local maps now stay raw `u16`, and the point where staging adopts txn-local numbering as ledger numbering is one visible `GraphId()` wrap in `build_reverse_graph_lookup`, commented. (A `TxnGraphId` newtype is the natural follow-up if we want that seam compiler-enforced too.)
2. **commit_transfer:** replay routing resolves *ledger* `GraphId`s from the binary store but feeds staging APIs that take envelope-keyed `u16` maps. That adoption is now an explicit conversion instead of an invisible identity — and the previously dead `GraphRoutingResult.graph_iris` field (the lone `dead_code` warning on main) is now genuinely read, so the warning is gone.

## Dead-type deletion (Phase 0.2)

`fluree-db-binary-index/src/types.rs` carried pre-V3 `RowColumnSlice` / `DecodedRow` / `OverlayOp` / `sort_overlay_ops` with **zero users in the workspace** — every live caller uses `read::types::{DecodedRowV3, OverlayOp}`. The dead set included the audit's FD-flagged width mismatch (`DecodedRow.dt: u32` vs `OverlayOp.dt`/`RunRecord.dt: u16`); the live V3 types are width-coherent (`o_type: u16` throughout). Deleted, with a tombstone note in the module docs.

## Validation

- Workspace compiles with **zero errors and zero warnings**; clippy clean (`-D warnings` posture)
- Full workspace test suite green; full `fluree-db-api` integration suite green (includes the fully-enforced differential harness)
- **Bench gate vs pre-change baseline: 18 comparisons, 0 breached; `scenario_mem` flat to ±0.06%** — the newtype is empirically zero-cost, per the Phase 0.0 protocol
- Wire-format safety: serde-transparent + explicit `as_u16()` at every manual codec write (`index_root`, `stats_wire`); decode sides wrap symmetrically

## Not in this PR (follow-ups per roadmap)

- `NsCode(u16)` newtype with snapshot-space/store-space tagging (Phase 0.1's second half — bigger design surface)
- `TxnGraphId` for the transaction-local space surfaced above


aaj3f:

PR #1331, opened 2026-06-12
fix(memory): make developer-memory store safe across processes

bplatz:
## Problem

Several processes can share one project's `.fluree-memory` at once — two editors, or multiple coding-agent sessions, each spawning its own `fluree mcp serve`. Concurrent writes **silently lost memories**.

Root cause: `needs_rebuild` compared the live `.ttl` hash against the shared on-disk `build-hash`, which only records who wrote *last*. A process whose own ledger never ingested another process's write believed it was in sync, then `add`/`update`/`forget` rewrote the whole `.ttl` file from its stale ledger (`all_memories_for_scope` reads the ledger, not the file) — dropping the other process's memories.

Investigation also surfaced two pre-existing hazards that any real fix exposes: concurrent rebuilds on a *shared* on-disk `__memory` ledger delete commits another process is reading (`Failed to read commit … Not found`), and the in-lock rebuild composes a deep transact/commit chain that overflows the default 2 MB tokio worker stack.

## Fix

- **Per-process watermark** — each store tracks the file content its own ledger reflects and decides rebuilds against that, not the shared on-disk hash.
- **In-lock sync** — `add`/`update`/`forget` rebuild-from-file inside the mutation's own cross-process file lock, before reading the set, so sync and the file rewrite are one critical section.
- **Process-private in-memory ledger for `mcp serve`** — rebuilt from the `.ttl` files on startup, so one process's cache refresh can't corrupt another's. CLI `fluree memory` keeps a file-backed ledger (for `import` and legacy migration); watermark seeding is gated per backend.
- **8 MB CLI worker stack** — covers the bounded depth of the in-lock rebuild chain.

The `.ttl` file stays the single source of truth, guarded by the existing file lock. This also right-sizes memory: the in-memory ledger never holds the server's RAM-tiered leaflet cache.

## Validation

- New `it_concurrent_cross_process` regression test: **8/24 memories survive on the old logic, 24/24 with the fix**.
- Stress harness (`concurrent_add_repro`, 3 stores / high concurrency): no lost updates, zero errors; the old shared-on-disk ledger still throws `commit … Not found` under the same load.
- Manual end-to-end: CLI memory roundtrip and `mcp serve` (rebuilds from file, recall returns seeded memories).

## Docs

Adds a "Concurrent access" section to `docs/memory/reference/ttl-format.md`.

### DIFF
--- a/docs/memory/reference/ttl-format.md
+++ b/docs/memory/reference/ttl-format.md
@@ -52,6 +52,13 @@ When you `memory add`, the CLI / MCP server:
 
 If the ledger write fails, the hash is left stale and the next `ensure_synced` call rebuilds the ledger from the files. When git pulls in a new version of `repo.ttl`, the hash mismatch triggers the same rebuild. In practice this is invisible.
 
+## Concurrent access
+
+It's normal for several processes to touch one project's memory at once — Cursor and Claude Code together, or several Claude sessions, each running its own `fluree mcp serve`. Two guarantees keep that safe:
+
+- **The TTL file is the only shared mutable state, and writes are serialized.** Every mutation takes a cross-process file lock (`.fluree-memory/.local/rebuild.lock`) before rewriting the file, and re-reads the file under that lock, so two writers can't lose each other's memories.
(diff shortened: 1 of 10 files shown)


zonotope (approved):
🥘

Issue #1328, opened 2026-06-12
Bound-object prefilters skip NumBig arena resolution for decimal/bigint constants


bplatz:
## Context

`value_to_otype_okey` (fluree-db-query/src/binary_scan.rs) resolves big numerics against the per-(graph, predicate) NumBig arena when given a `numbig_ctx: Option<(GraphId, u32)>`. The overlay-translation caller passes the context; the two bound-object prefilter callers pass `None`:

- the bound-object scan prefilter (`encode_bound_object_prefilter`)
- the batched-join bound-object encode path

So a query constant like `?s ex:budget 19.99` (typed `xsd:decimal`) cannot prefilter-encode to its `(NUM_BIG_OVERFLOW, handle)` key and falls back to post-decode equality checks, even though the predicate and graph are known at those call sites.

## Proposal

Thread `(g_id, p_id)` through the prefilter call sites so bound decimal/overflow-integer constants resolve via `find_numbig_handle`. An arena miss means the constant matches no base row — which is itself a useful prefilter outcome (empty base scan, overlay-only check).

Low priority: post-decode equality is correct today; this is a scan-narrowing optimization for equality lookups on arena-keyed numerics.

Issue #1325, opened 2026-06-12
Inline encoding for small xsd:decimal values (index format evolution)


bplatz:
## Context

Every typed `xsd:decimal` routes to the per-(graph, predicate) NumBig arena — even `\"1\"^^xsd:decimal` (`fluree-db-indexer/src/run_index/resolve/resolver.rs`, `DecimalStr` arm). That makes decimals handle-keyed and equality-only, costs an arena round-trip on decode, and is the root reason novelty decimals are hard to translate into V3 overlay ops (see #1324).

Integers already have the right shape: i64-fitting values encode inline (`NUM_INT`), only overflow goes to the arena. Decimals deserve the same split.

## Proposal

Canonical `xsd:decimal` (trailing zeros stripped) gives every value a unique `(unscaled mantissa, scale)` pair, so a packed inline o_key is well-defined, e.g.:

```
1 bit sign | 6 bits scale (0–63) | 57 bits |mantissa|    fits-check: |m| < 2^56, scale <= 63
```

~17 significant digits with up to 63 decimal places — fixed-point money (scale 2) inlines to ±~$7e14. Decode rebuilds the exact `BigDecimal`, so precision math is unaffected. Values that don't fit keep the arena path, exactly like overflow integers.

Two design decisions to settle in spec:

1. **Equality-keyed vs order-preserving.** Equality-keyed (packed canonical form) matches current decimal semantics — decimals already compare by decode, never by o_key order. An order-preserving variant (normalized scientific form: sign, biased exponent, mantissa) would additionally enable range/FILTER pushdown on decimals, at meaningfully higher design risk.
2. **Migration.** `(o_type, o_key)` is persisted fact identity: the same value must never exist under two encodings within one index root, or retracts re-encoding inline would fail to match arena-stored base rows. This requires an index-format version gate with reindex as the migration path, and a sweep through resolver spool kinds, the o_type registry, decode paths, stats hooks, overlay translation, and filter-pushdown gating (which must not assume numeric order unless the order-preserving variant is chosen).

## Out of scope

This is independent of #1324 (which needs no format change and covers existing ledgers); inline decimals are the long-term fix that removes the arena dependency for the common case.

PR #1332, opened 2026-06-13
fix: accounting-grade exactness for xsd:decimal and unbounded xsd:integer

bplatz:
## Summary

A review of the engine's numeric handling identified several areas where `xsd:decimal` and large `xsd:integer` values needed correction. The storage layer (BigDecimal → NumBig arena) represents these values exactly, but a number of paths into and through it converted via `f64` or narrowed to `i64` before that representation was reached, which could lose precision or cause values to mismatch. This PR closes those gaps so decimal and big-integer handling is exact end to end: ingestion, query evaluation, aggregation, comparison/dedup, validation, reasoning, and graph-source materialization.

Behavior is unchanged for `string`/`boolean`/`double` data and ordinary `i64`-range integers; the fixes add type-specific arms that the common paths skip.

## Ingestion — exact values from every entry point

Previously only two ingest lanes preserved decimals exactly (JSON-LD string `@value`, default-graph Turtle). Now exact across the board:

- **SPARQL query constants** — `19.99` and `"19.99"^^xsd:decimal` lower to exact `BigDecimal` in triple patterns, `FILTER`/`BIND`, and `VALUES` instead of `f64`.
- **SPARQL UPDATE** — `INSERT`/`DELETE DATA` and templates persist exact decimals; a delete written with a different scale (`1.5` vs `1.50`) still matches the stored fact.
- **JSON-number `@value`** — a JSON number typed `xsd:decimal` is recovered exactly via its shortest round-trip form rather than stored as `f64`.
- **Bulk-import / graph adapter** — JSON-number decimals keep their declared `xsd:decimal` datatype (they were landing as `xsd:double`).
- **TriG GRAPH blocks** — bare decimals parse exactly, matching the default-graph path; the named-graph transact path now parses all typed literals instead of storing mislabeled strings.
- **Unbounded `xsd:integer`** — integers beyond `i64` now promote to `BigInt` in both the Turtle and SPARQL lexers and in typed-literal lowering. Previously a large bare integer silently became `0`, and a typed one errored or degraded to a string.

## Query engine — exact values produce correct answers

- **SUM / AVG** include arena-backed (indexed) decimals; previously those rows contributed nothing, so a sum over an indexed decimal column was wrong and shifted after indexing. `GROUP_CONCAT` no longer drops decimals/bigints.
- **GROUP BY / DISTINCT / MINUS / joins / `VALUES`** key decimals by canonical numeric value, so the same value served encoded from the index and decoded from novelty unifies into one group/row, and scale variants (`1.5` / `1.50`) no longer split.
- **GRAPH scope** — NumBig handles are per-(graph, predicate); values escaping a `GRAPH { … }` block are now decoded against their own graph's arena instead of silently resolving to a different value in the outer graph.
- **Numeric-filter COUNT fast path** defers to the exact pipeline on decimal rows instead of counting them as non-matches (an undercount), and bails from a manifest read rather than launching a doomed leaf/partition scan first.

## Storage identity

- The NumBig arena keys decimals on their **normalized** value, so `1.5` and `1.50` share one handle and one persisted encoding (`(o_type, o_key)` is fact identity — two encodings broke retract matching). Legacy pre-normalization arenas load with handles preserved and ambiguous lookups decline cleanly.
- `FlakeValue`'s `Hash` is now consistent with its cross-type numeric `Eq`: equal binary-fraction values like `Double(2.5)` and `Decimal(2.5)` hash to the same bucket.

## Validation, reasoning, graph sources

- **SHACL** range facets (`sh:minInclusive` etc.) and `sh:lessThan`/`lessThanOrEquals` compare decimals and bigints via exact cross-type ordering. Previously every decimal value unconditionally failed every range facet.
- **Datalog reasoner** gains exact `Decimal`/`BigInt` binding variants: decimal objects now match in rule bodies (they matched nothing before), overflow integers no longer zero out, derived facts round-trip losslessly, and equality is exact rather than epsilon-based.
- **R2RML** decimal formatting keeps the sign for values between −1 and 0 (`-0.5` was materializing as `"0.5"`, corrupting literals and template-generated IRIs).

## Performance

Non-decimal paths are unaffected (new arms sit after the existing fast arms). Decimal paths trade a few previously-instant-but-wrong answers for correct ones (e.g. the COUNT fast path defers). One follow-up worth noting: exact decimal constants now resolve to arena handles, so equality lookups currently scan-and-post-filter rather than prefilter-narrow — addressed by #1328 (prefilter arena threading) and #1325 (inline small-decimal encoding).

## Testing

New `it_decimal_exactness.rs` integration suite (12 tests) covers the full loop — exact round-trips, high-precision survival, cross-path retract identity, mixed index/novelty grouping, per-graph arena decoding, and mixed-type COUNT — plus unit tests at each layer. Full workspace suite passes (excluding the known LocalStack testcontainer flakes); clippy clean with `--all-features --all-targets`; W3C SPARQL suite stable at 250/327 with no regressions.


zonotope (approved):
🔦

PR #1333, opened 2026-06-13
fix(query): GROUP BY (expr) collapses to one row per group and honors LIMIT

bplatz:
## Summary

`GROUP BY` over an expression that is *also* projected in the SELECT (e.g. `SELECT (LCASE(?cur) AS ?k) ... GROUP BY (LCASE(?cur))`) row-exploded and appeared to ignore `LIMIT`. Reported from the field against 4.0.5 and 4.0.6.

## Root cause

The SPARQL lowering produced **two disconnected variables for the same expression**:

- `SELECT (LCASE(?cur) AS ?k)` → an independent per-row bind `?k = LCASE(?cur)` in the WHERE patterns
- `GROUP BY (LCASE(?cur))` → a *separate* synthetic group key (`?__group_expr_N`)

The group key was the synthetic var, so the projected `?k` was neither a group key nor an aggregate. Grouping and `COUNT` were computed correctly, but `?k` survived grouping as a **multi-valued list** which the results formatter then **exploded into one row per input binding**. `LIMIT` had already been applied to the collapsed grouped rows upstream of the explosion, so it had no visible effect.

## Fix

When an unaliased `GROUP BY (expr)` matches a SELECT expression projected as `(expr AS ?k)` (matched on the structural expression key), group on `?k` directly. The SELECT pre-bind already computes `?k = expr`, so grouping happens on the projected variable and each row equals the group value, per SPARQL 1.1 §11.2. This is the BIND-then-`GROUP BY ?k` form that already worked, applied automatically.

Explicit `GROUP BY (expr AS ?alias)` and the synthetic-var fallback (when the expression is not projected) are unchanged.

## Behavior

For the reporter's query, the SPARQL-results JSON now returns one row per distinct group (collapsed, `LIMIT` honored) instead of one row per input line item.

## Tests

- `fluree-db-api/tests/it_query_sparql.rs`: collapse + `LIMIT` integration test, plus a BIND-workaround control.
- `fluree-db-sparql/src/lower/mod.rs`: lowering unit test asserting the group key reuses the SELECT alias var (no synthetic var, single bind).
- Verified no regression against the W3C SPARQL aggregates category (identical failure set before/after).

### DIFF
--- a/fluree-db-api/tests/it_query_sparql.rs
+++ b/fluree-db-api/tests/it_query_sparql.rs
@@ -4723,3 +4723,111 @@ async fn sparql_service_remote_no_executor_errors() {
         "Error should mention missing executor, got: {err}"
     );
 }
+
+// =========================================================================
+// Field report repro: GROUP BY over an expression must collapse + honor LIMIT
+// =========================================================================
+
+/// Seed line items across three currencies (mixed case) so that
+/// `GROUP BY (LCASE(?cur))` collapses to exactly three groups.
+async fn seed_currency_line_items(fluree: &MemoryFluree, ledger_id: &str) -> MemoryLedger {
+    let ledger0 = genesis_ledger(fluree, ledger_id);
+
+    // 5 USD, 3 CAD, 2 EUR (mixed case to exercise LCASE collapsing).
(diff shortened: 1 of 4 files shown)


zonotope (approved):
🇰🇪

PR #1335, opened 2026-06-14
feat: inline order-preserving xsd:decimal encoding + numeric range pushdown

### LINKED ISSUE #1325: Inline encoding for small xsd:decimal values (index format evolution)

bplatz:
Closes #1325.

Builds on the exactness work in #1332 (base branch `fix/decimal-exactness`); merge after that lands.

## What this does

Small `xsd:decimal` values now store **inline** in the index, the way integers already do — a value that fits encodes directly into the 64-bit object key, and only genuinely large/high-precision decimals fall back to the per-`(graph, predicate)` NumBig arena. The inline key is an **order-preserving, canonical base-10 float** code: equal values produce identical bits (equality, dedup, joins) *and* raw `u64` order equals numeric order (range / `ORDER BY` pushdown). Because the format change is the one-time expensive cost, the order-preserving layout was locked in directly rather than shipping an equality-only form first and needing a second migration.

The inline range is generous: up to 17 significant digits with the decimal point floating in `[-32, 31]` — e.g. money to the cent up to ~$1 quadrillion, six decimal places up to ~$100 billion. Beyond that, values spill to the arena and remain fully correct, just without the inline benefits.

### Backward compatibility / migration
- The encoding is a new index-root format version. **Existing ledgers keep working unchanged — no reindex required.**
- A full reindex (or fresh bulk import) adopts the new format automatically and unlocks inlining for that ledger going forward. Ledgers migrate independently; old binaries refuse the new root version rather than misreading it.
- Decode is always dual-capable, so new code reads old and new roots transparently.

### Query pushdown harvested
With order-preserving keys, the fast paths now exploit decimals (and, as a no-format-change bonus, integers/doubles — their keys were already order-preserving on disk, so this works on existing indexes too):
- `ORDER BY [DESC] … LIMIT k` — sorted-tail scan, sort elimination.
- `COUNT(… FILTER(?o cmp k))` — encoded-key comparison with manifest-extent shortcuts.
- `MIN` / `MAX` — boundary-key read, no row scan.
- `SELECT … FILTER(?o cmp k)` — seeks the o_key range instead of post-filtering the whole predicate, **when the predicate is uniformly one order-preserving numeric type and there is no overlay**.

`SUM`/`AVG` deliberately stay on the exact `BigDecimal` path (the f64 fast-agg lane would be lossy for decimals).

### Correctness guardrails
- Range narrowing requires a uniform predicate (manifest extent `min_o_type == max_o_type`) **and** no overlay — novelty can introduce a cross-type value whose op sorts outside the narrowed window; with overlay present we fall back to the full scan + merge + post-filter. Regression tests cover both a decimal base with an integer-novelty match and an integer base with a decimal-novelty match.
- A mixed-type predicate (e.g. integers + decimals under one property) never narrows, so no cross-type matches are dropped.

## Notable changes
- Order-preserving inline decimal codec with monotonicity property tests (sorted-set walk + exhaustive pairwise).
- New `OType::XSD_DECIMAL_INLINE` / `ObjKind::NUM_DEC`, a root-derived `DecimalEncoding` policy (sticky per root, preserved across incremental writes), and a root version bump that old binaries refuse.
- Resolver, bulk-import sink, query-constant encoding, decode, and class/property datatype stats all wired for the inline lane.
- Overlay-translation cache keyed on the decimal policy so a same-`t` reindex can't serve a stale arena-keyed translation against an inline root.
- End-to-end and differential tests: write→read round-trips across the inline/arena boundary, equality-constant matching, and identical results between an indexed-inline read and the canonical novelty representation.

## Testing
`cargo fmt --check`, `cargo clippy --all --all-features --all-targets`, and `cargo nextest run --workspace --all-features` all pass (the lone failure is the known LocalStack `PortNotExposed` testcontainer flake, unrelated to this change).


fluree-db-query/src/binary_scan.rs
```diff
@@ -1829,60 +1829,130 @@ impl Operator for BinaryScanOperator {
...
+                        crate::fast_count::encode_numeric_threshold_for_otype(pred_ot, v)
+                            .ok()
+                            .flatten()
+                    };
+                    if let Some((v, _inclusive)) = bounds.lower.as_ref() {
+                        range_min_okey = enc(v);
+                    }
+                    if let Some((v, _inclusive)) = bounds.upper.as_ref() {
+                        range_max_okey = enc(v);
+                    }
```
aaj3f, line 2006: I was a little confused here, because it seemed like `bounds.lower`/`bounds.upper` would be inclusive and would have ignored the `_inclusive` flag later when computing `range_min_okey`/`range_max_okey`. I see now the post-filter is where enforcement takes place. So for an exclusive filter it just seems to over-read by 1 at each boundary and then enforces exclusivity in post-filter rejection.

This is all fine, but maybe just a comment around line 1896 that makes that clear to avoid the validation exercise for future devs

fluree-db-query/src/fast_star_const_order_topk.rs
```diff
@@ -360,6 +388,23 @@ fn filter_subjects_by_numeric_gt(
...
+                        saw_uncomparable_numeric = true;
+                        false
+                    }
+                },
+                // Numeric but not o_key-comparable (arena big numerics, other
+                // integer widths/floats): can't decide here.
+                _ if ot.is_numeric() || ot == OType::NUM_BIG_OVERFLOW => {
+                    saw_uncomparable_numeric = true;
+                    false
+                }
```
aaj3f, line 405: This all seems to guarantee correctness, but I thought perhaps that the fast-path would still support `xsd:long`, `xsd:int`, `xsd:short`,  or `xsd:float` as those would be o_key-comparable as well?

aaj3f (approved):
The order preserving fix and the numeric range optimization are really nice. Some CI failures you may want to look into before merging. Also, I made a note in #1355 but that PR fixes a condition that is still latent in this PR in `fluree-db-query/src/binary_scan.rs:2668` that retains the integer-valued-double --> `i64` shortcut that was corrupting indexed doubles -- just be sure to get those fixes into this PR before it's merged (possibly by merging #1355 first then updating this PR's code to match it). Here is Claude's detailed description of this, fwiw:

````
This branch keeps the **double's** OType (`ot == XSD_DOUBLE`) but encodes the
key with `ObjKey::encode_i64` instead of `encode_f64`. A reader decoding an
`XSD_DOUBLE` row interprets `o_key` via `decode_f64`, so an i64-encoded key
(e.g. `100.0` → `encode_i64(100)` = `0x8000000000000064`) is read back as a
subnormal/garbage f64 — silent data corruption over an indexed double predicate.
It is also internally inconsistent with the *overlay* path in
`dict_overlay.rs:487`, which for the same integral-double case emits
`ObjKind::NUM_INT` (a different OType entirely). The two paths disagree, so a
constant encoded here will not match an overlay row encoded there, nor a
resolver-written `encode_f64` base row.

This is exactly the shortcut PR **#1355** (fix/integer-float-index-corruption)
removes — in `binary_scan.rs` *and* the mirror at `dict_overlay.rs:487` (which
this PR does not touch at all). As written, #1335 ships with the corruption
still present.

Recommendation: **land #1355 first, rebase #1335 on top.** This PR's own
`FlakeValue::Decimal` arm (`:2756`) is correct and unrelated; the problem is the
pre-existing Double lane it inherits. On rebase there will be a textual conflict
in `binary_scan.rs` (this region) and in
`fluree-db-api/tests/it_decimal_exactness.rs` (both PRs append to it). Resolve by
taking #1355's removal of the shortcut and keeping #1335's new Decimal arm + new
tests.

(Strictly, this is not a defect *introduced* by #1335 — it is a cross-PR ordering
constraint. Flagging CRITICAL because merging #1335 without #1355 lands known
corruption, and the generalized double range pushdown this PR adds runs *over the
same corrupt keys*.)
````

aaj3f:
Backlog-sweep nudge: this is approved, but it's now 1,616 commits behind main and conflicting, and the index-format ground it touches has shifted meaningfully since (#1552's graph-ir line, #1624's rebuild chain, the incremental-stats rework). I don't think the design is in question — the order-preserving inline code + format-version gate were reviewed and approved, and they settle both open decisions #1325's `needs-specification` label was asking about (I'm removing that label on the issue accordingly).

What it needs is an explicit owner decision rather than more quiet aging: either rebase-and-revalidate the existing 29 commits, or re-cut the format change fresh against current main carrying the approved design. Given how much the substrate moved, I suspect re-cut may honestly be cheaper than the rebase, but I'd defer to whoever takes it. Happy to talk through sequencing if it helps — mostly I don't want an approved format evolution silently rotting while #1325 sits open pointing at it.

PR #1336, opened 2026-06-14
fix(transact): zero WHERE solutions no longer fire all-literal INSERTs

bplatz:
## Summary

INSERT/DELETE templates must instantiate **once per WHERE solution** (SPARQL 1.1 Update §3.1.3), so a *present* WHERE that matches zero rows is a no-op. The shared staging layer previously had a fallback that fired **all-literal** INSERT templates (constants only, no WHERE-bound vars) against a synthetic empty solution when WHERE returned zero rows — so e.g. `INSERT { <s> <p> <o> } WHERE { FILTER(false) }` wrongly inserted.

This removes the fallback in the shared staging path, so **SPARQL UPDATE and JSON-LD update now behave identically and per-spec**. A required pattern (or `FILTER`) that collapses to zero solutions inserts nothing.

## Why remove it for both surfaces (no opt-in flag)

The fallback lives in the single shared `stage.rs` path consumed by both SPARQL and JSON-LD. The one idiom that looked like it depended on it — "delete-if-exists, always insert" — actually rides on `OPTIONAL` producing **one** row (probe var unbound), not zero, so it is unaffected. No scenario was found where firing on an empty guard is the desirable outcome, and keeping behavior consistent between the two surfaces is the goal.

## Unaffected paths

- **No WHERE clause at all** — the `SingleEmpty` cursor still emits one empty solution, firing all-literal templates exactly once (plain insert).
- **INSERT DATA / DELETE DATA** — separate path.
- **OPTIONAL-based insert-if-not-exists / always-insert** — yields one row, still fires correctly.

## Changes

- `fluree-db-transact/src/stage.rs` — remove the zero-row all-literal fallback and the now-unused `has_where_clause`; update surrounding comments.
- `docs/transactions/conditional-updates.md` — rewrite the "Two INSERT behaviors" section that documented the old behavior.
- `fluree-db-api/tests/it_transact_conditional.rs` — fix two stale comments; add regression tests for required-pattern no-op, FILTER-excludes-all no-op, no-WHERE-still-fires, and SPARQL `FILTER(false)` parity.

## Testing

- `fluree-db-transact`: 197/197 pass.
- API transact + sparql + named-graph sweep: 265/265 pass.
- Clippy clean on the changed crate; `cargo fmt --all` applied.

> Base is `fix/docker-cli-aws-support` (the AWS CLI/server branch, PR #1334) since this stacks on it.

### DIFF
--- a/docs/transactions/conditional-updates.md
+++ b/docs/transactions/conditional-updates.md
@@ -21,10 +21,15 @@ This guide covers common patterns for state-dependent updates with both **JSON-L
 
 The WHERE clause runs against the **current** database state. If it matches, the bound variables flow into DELETE (to retract old values) and INSERT (to assert new ones). If WHERE returns zero rows — because a FILTER eliminated them or a pattern didn't match — DELETE is skipped entirely (nothing to retract) and INSERT templates with unbound variables produce zero flakes.
 
-### Two INSERT behaviors
+### How INSERT interacts with WHERE solutions
 
-- **INSERT with variables from WHERE** (e.g., `"@id": "?s"`) — conditional. When WHERE returns zero rows, the variable is unbound and the INSERT produces nothing. Use this for CAS, state machines, and guards.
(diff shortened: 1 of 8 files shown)


zonotope (approved):
🍠

PR #1337, opened 2026-06-14
feat(server): query freshness for disconnected query servers (TTL refresh + read-after-write min-t)

bplatz:
## Summary

Adds two complementary freshness mechanisms for query servers that run **disconnected** from any transaction server — no direct link to the writer and no subscription to the writer's SSE commit stream — yet read from a shared nameservice (e.g. a DynamoDB table) the writer keeps current. Without a refresh signal such a server serves whatever its caches were warmed with and never observes commits made elsewhere.

### 1. TTL-gated query-time nameservice refresh (opt-in, default off)
Before a current-head read, the server calls `Fluree::refresh()` to pull the latest nameservice record and apply it to the shared cache. The check is gated by a per-process, per-canonical-ledger TTL so high-QPS traffic does not hit the nameservice on every request; the gate records *attempts* (not successes) so bursty failures cannot stampede it. `ttl_ms = 0` checks every request. Demand-driven, not a background poller.

- Config: `--query-refresh-enabled` / `--query-refresh-ttl-ms`, `FLUREE_QUERY_REFRESH_*`, or `[server.query_refresh]`.
- Proxy-storage peers (which track head via their SSE subscription) skip the check.

### 2. Read-after-write `min-t` (hard, per-request guarantee)
A client that just wrote and knows the resulting `t` can require that the server observe at least that `t` before executing. Requested via:
- `Fluree-Min-T` header
- JSON-LD `opts.min-t` / `min_t` / `minT`
- a numeric snapshot pin (`from: "ledger:main@t:N"` or `{"@id":"ledger:main","t":N}`)
- multi-query envelope `opts` / `asOf`

The server refreshes each referenced ledger until it reaches the requested `t`, then runs the query (a numeric pin reads that snapshot; a plain header waits on current HEAD). Per-ledger requirements are merged by maximum across header, opts, and pins.

- Bounded by a dedicated `query_min_t_timeout_ms` (default 5s) so the wait never inherits a disabled or 15-minute query timeout; capped exponential backoff with jitter avoids stampeding the nameservice.
- Absent ledger fails fast as not-found; exhausting the budget returns `ReadAfterWriteTimeout` (HTTP 408).
- On success the refresh gate is stamped so the downstream current-head path does not re-check; other referenced ledgers without a `min-t` still get the ordinary TTL refresh.
- Proxy-storage peers satisfy the requirement from their observed head and reject only when that head is behind.

The TTL bounds ordinary staleness; `min-t` proves a specific transaction is visible. The two work together.

## Docs
`docs/operations/configuration.md`, `docs/concepts/time-travel.md`, `docs/api/headers.md`, `docs/api/endpoints.md`, `docs/api/multi-query.md`.

## Testing
Config round-trip, header parsing/validation, and `min-t` requirement-collection unit tests added. `cargo check -p fluree-db-server` clean. Full test suite not run locally.

### DIFF
--- a/fluree-db-server/src/routes/query.rs
+++ b/fluree-db-server/src/routes/query.rs
@@ -115,9 +115,14 @@ async fn await_query_min_t_requirements(
                 .map_err(ServerError::Api)?;
             let current = ledger.t();
             if current < min_t {
-                return Err(ServerError::not_implemented(format!(
-                    "fluree-min-t requires t={min_t}, but proxy storage mode can only observe current t={current} and cannot perform local read-after-write refresh"
-                )));
+                // A proxy peer tracks head passively via its SSE subscription and
+                // cannot refresh on demand, so report a behind head as a
+                // read-after-write timeout (HTTP 408) — consistent with the
+                // non-proxy wait path — rather than a misleading "not implemented".
+                return Err(ServerError::Api(ApiError::AwaitTNotReached {
+                    requested: min_t,
+                    current,
+                }));
             }
         }
         return Ok(());
--- a/fluree-db-server/src/routes/query.rs
+++ b/fluree-db-server/src/routes/query.rs
@@ -21,14 +21,16 @@ use axum::http::HeaderMap;
 use axum::response::{IntoResponse, Response};
 use fluree_db_api::dataset::GraphSelector;
 use fluree_db_api::{
-    DatasetSpec, FreshnessCheck, FreshnessSource, GraphDb, GraphSource, LedgerState,
+    ApiError, DatasetSpec, FreshnessCheck, FreshnessSource, GraphDb, GraphSource, LedgerState,
     QueryExecutionOptions, RefreshOpts, TimeSpec, TrackingTally,
 };
+use rand::Rng;
 use serde::Deserialize;
 use serde_json::Value as JsonValue;
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
 use std::sync::Arc;
 use std::sync::atomic::Ordering;
+use std::time::{Duration, Instant};
 use tracing::Instrument;
 
 fn query_execution_options(state: &AppState) -> QueryExecutionOptions {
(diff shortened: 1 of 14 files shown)


fluree-db-server/src/routes/query.rs
```diff
@@ -1275,6 +1478,199 @@ fn base_ledger_id(s: &str) -> Result<String> {
...
+
+fn collect_sparql_min_t_requirements(
+    headers: &FlureeHeaders,
+    sparql: &str,
+    default_ledger: Option<&str>,
+) -> BTreeMap<String, i64> {
+    let mut requirements = BTreeMap::new();
+    let ledger_ids = fluree_db_api::sparql_dataset_ledger_ids(sparql).unwrap_or_default();
+
+    let parsed = fluree_db_sparql::parse_sparql(sparql);
```
aaj3f, line 1643: This `collect_sparql_min_t_requirements()` seems to get called regardless of whether `query_refresh_enabled` is set or whether a `Fluree-Min-T`/`@t` pin is present, and internally the `sparql_dataset_ledger_ids()` and `parse_sparql()` fns at 1627 and 1629 each do full SPARQL parses unconditionally. Those parses then happen again downstream as part of formatting and within the engine itself. 

Minimally this could be fixed if we correct to abide by a default-off promise for ths fn and short-circuit this fn w/ something like

```rust
    if headers.min_t.is_none() && !sparql.contains("@t:") {
        return BTreeMap::new();
    }
```

Or we could parse it once and thread the `&ParsedSparql` into this min-t collector and other downstream callsites

fluree-db-server/src/routes/query.rs
```diff
@@ -8,32 +8,197 @@
...
+            let ledger = state
+                .fluree
+                .ledger(&ledger_id)
+                .await
+                .map_err(ServerError::Api)?;
+            let current = ledger.t();
+            if current < min_t {
+                return Err(ServerError::not_implemented(format!(
+                    "fluree-min-t requires t={min_t}, but proxy storage mode can only observe current t={current} and cannot perform local read-after-write refresh"
+                )));
```
aaj3f: Seems a bit of a confusing error here. Elsewhere it seems we'd be returning a `408 ReadAfterWriteTimeout` or `AwaitTNotReached`

fluree-db-server/src/routes/query.rs
```diff
@@ -474,6 +639,22 @@ pub async fn query(
...
+        } else {
+            let ledger_ids = fluree_db_api::sparql_dataset_ledger_ids(&sparql).unwrap_or_default();
+            await_query_min_t_requirements(state.as_ref(), min_t_requirements.clone()).await?;
+            refresh_ledgers_without_min_t(
+                state.as_ref(),
+                ledger_ids,
+                &min_t_requirements,
+            )
+            .await;
+        }
```
aaj3f: I'm trying to reason through what would happen if we had a `Fluree-Min-T` header but no `FROM`. I think `ledger_ids` would be empty and `default_ledger` would be `None`. So there'd be no requirements produced here but no error either. Maybe we *should* error on something like `Fluree-Min-T requires a target ledger`?

aaj3f (approved):
Looks good -- left some notes

zonotope (approved):
♒

bplatz:
Addressed the review notes:

- \`f503a7a7e2715e4c18e4ab20224b042398ac8a7a\` — guard the min-t SPARQL parses behind a header/\`@t:\` check, and reject a \`Fluree-Min-T\` with no target ledger (400)
- \`e4fb81bd4003583c8c4e6a7ee79b80587be0ed75\` — proxy-mode behind head now returns read-after-write timeout (408) instead of not_implemented
- \`b49e9a84d728a704e45cb81b27bf81652cd71f77\` — cargo fmt

PR #1338, opened 2026-06-15
feat: wholesale .flpack ledger restore (streaming API + server import endpoint + remote CLI)

bplatz:
## Summary

Adds **wholesale `.flpack` restore** — the inbound counterpart of `.flpack` export — so a full ledger (commits, txn blobs, and prebuilt index) can be materialized from a trusted archive at the API, over HTTP, and from the CLI, under any name. `.flpack` becomes the universal on-ramp for getting data onto a server: build a ledger locally in *any* supported format, export it, then import it wholesale.

```bash
fluree create staging --from data.ttl                 # any format
fluree export staging --format ledger -o snap.flpack
fluree create prod --remote origin --from snap.flpack  # restore onto the server, one step
```

This is distinct from `publish`/`POST /push` (which replays + re-validates + reindexes a local commit chain): import trusts the archive byte-for-byte (every frame SHA-256 verified), does not replay, and the prebuilt index rides along, so the restored ledger is immediately queryable.

## What's included

- **API** — `Fluree::restore_ledger`, generic over `AsyncRead` (file / S3 `GetObject` / HTTP body). Decodes frame-by-frame (never buffers the whole archive), verifies each object, finalizes commit/index heads from the embedded nameservice manifest, and rolls back the half-created ledger on any failure. Verifies the manifest's head CIDs were actually ingested before setting heads (no dangling head from a truncated/mismatched archive). Normalizes a bare name to `name:main`.
- **Server** — `POST /v1/fluree/import/*ledger`, admin-gated, peer-forwarding, streaming body → `restore_ledger`. `201` + JSON summary; `409` on name collision; `400` on malformed archive.
- **CLI** — `create --remote --from x.flpack` streams the archive to the import endpoint (no local staging instance needed); local `--from x.flpack` now routes through the same `restore_ledger` path (deletes the duplicated mmap loop, gains rollback, drops the now-unused `memmap2` dep).
- **Docs** — HTTP endpoints, operations guide, CLI reference, and the external-implementer **Ledger Import Contract** in `server-integration.md`.

## Testing

- 8 flpack tests: API round-trips (with/without index), rollback on truncation, bare-name normalization, server HTTP round-trip, and `409` collision — all pass.
- 136-test CLI + server regression smoke — all pass.
- Clippy clean across api/server/cli (all-features, all-targets).

## Notes / out of scope

- Built on top of `feature/query-refresh-ttl` (stacked PR).
- CLI `--from s3://…`/`https://…` is **not** included — the API already accepts any `AsyncRead` source, but the CLI `--from` remains a local path (download-then-import works today).
- Name collision is a hard `409`; no `?overwrite` yet.


fluree-db-server/src/import_jobs.rs
aaj3f, line 1: It seems we don't actually enforce/implement the TTL/GC for the registry or staged files. The minted jobs set up `created_at` and the response advertises `expires_at_unix` but the memory `DashMap` entry and disk `.flpack` / part files don't have cleanup. Perhaps either a periodic sweep or a sweep-on-mint pattern to drop jobs / staging older than advertised expiry.

aaj3f (approved):
One note re TTL/GC but otherwise 👍

zonotope (approved):

bplatz:
Addressed the TTL/GC note: minted import jobs and their staged files are now reclaimed (sweep-on-mint, 1h TTL, running restores spared) and expiry is enforced on upload/complete — 8bf44c829482d424492222ebcbb997d7fc97ff52

PR #1339, opened 2026-06-15
fix: DATATYPE() returns the datatype IRI, not a compact string

bplatz:
## Problem

`DATATYPE(?v)` returned a compact **string** (e.g. `"xsd:decimal"`) rather than the datatype **IRI** required by SPARQL 1.1 §17.4.2.3. As a result, the idiomatic filter

```sparql
FILTER(DATATYPE(?v) = xsd:decimal)
```

compared a string against an IRI term and matched **0 rows** (the `xsd:decimal` on the RHS lowers to an IRI/Sid). Users had to fall back to `FILTER(STR(DATATYPE(?v)) = "xsd:decimal")`. Reported against 4.0.6 / 4.0.6-hotfix.1.

## Fix

`eval_datatype` now returns the datatype as an IRI term (`ComparableValue::Sid`):

- `Lit` → the literal's datatype Sid (lang-tagged → `rdf:langString`).
- `EncodedLit` → reserved-datatype fast path (store-free) or `store.dt_sids()`.
- IRI/ref bindings → the `@id` ref type (existing Fluree extension), now as an IRI term.

Because `xsd:decimal` on the filter side lowers via `IRI()` → `encode_iri_strict` to the same canonical Sid, the comparison is now IRI-vs-IRI and matches correctly. The compact-string formatters (`format_datatype_sid` / `format_reserved_datatype`) were only used by `DATATYPE` and are removed.

## Behavior change (unify on IRI term)

`DATATYPE` is the same function for SPARQL and JSON-LD (shared engine), so this also changes the JSON-LD surface, which previously treated `datatype()` as a compact string:

- **SPARQL results**: render the full datatype IRI (W3C results format has no prefix table).
- **JSON-LD results**: compact the IRI against the active `@context` (e.g. `xsd:decimal` when an `xsd` prefix is present), so display is largely unchanged.
- **JSON-LD filters** that compared `datatype()` against a compact string now compare against the datatype IRI — use `(iri "<full-iri>")`. `STR(DATATYPE(?v))` now yields the full IRI.

## Tests

- New SPARQL regression: `FILTER(DATATYPE(?v) = xsd:decimal)` matches the decimal-typed subjects.
- Updated the JSON-LD datatype filter/insert tests to the IRI semantics.
- Docs: clarified the `DATATYPE` entry in `docs/query/sparql.md`.

Full `fluree-db-query` suite + the affected `fluree-db-api` integration tests pass; clippy clean.


aaj3f (approved):
Only really one main question. The old `format_reserved_datatype` could answer `DatatypeDictId::LANG_STRING` and `DatatypeDictId::FULL_TEXT` without falling back to touching the store (i.e. had a store-free fast-path). The new `reserved_datatype_sid` doesn't seem to have arms for these (e.g. no field for either in `WellKnownDatatypes`), so both now fall back to `store.dt_sids()` lookups. 

It doesn't seem any of this PR's needed changes also did require those arms to be dropped, but perhaps I'm being shortsighted on that

zonotope (approved):
🖍️

bplatz:
Good catch — restored the store-free arms for both in 9c4e892970d6674c24b48ca4de6d93e7f89ff5fc (added `rdf_lang_string` and `fluree_full_text` to `WellKnownDatatypes`).

PR #1341, opened 2026-06-16
feat: embedded version-pinned docs lookup (fluree docs + fluree-docs MCP server)

Jackamus29:
## What

Makes Fluree's documentation reachable to coding agents as a first-class, **version-pinned** tool — so an agent working in any Fluree project stops guessing syntax and burning iterations. The motivating miss: `<prop>+` vs `<prop>*` for SPARQL property paths, a pure lookup that's now one call.

A new `fluree-db-docs` crate embeds the `docs/` mdBook into the binary and exposes a section-level lookup over it. Because the docs ship inside the binary, every result is **version-exact for that build** (and, in the standard local-dev workflow, the exact Fluree DB being developed against). Offline, no network, no database required.

## Surfaces (one core, two surfaces)

- **`fluree docs`** CLI — `search` / `get` / `examples` / `tree` (human + `--json`)
- **Standalone `fluree-docs` MCP server** (`fluree docs serve`) exposing `docs_search` / `docs_get` / `docs_examples` / `docs_tree`. Read-only over static content → safe to auto-allow.

It is a **separate** MCP server from developer-memory — `fluree-db-memory` is untouched; the two share only the `docs/` source. The memory namespace (`fluree memory init` / `fluree memory mcp-install`) installs the memory server **only**; the docs server is opt-in via the unified installer.

## Index & retrieval

- BM25 over field-weighted sections (title ×3 / heading ×2 / body ×1) with Snowball stemming (so `property path` matches the *Property Paths* heading).
- Navigation/boilerplate sections ("Related Documentation", link-list footers) are **excluded from the search index** (conservatively classified, at section granularity) but remain retrievable via `get`/`tree`.
- `README.md` files are indexed and fetchable — in mdBook they are the section-index pages, and `SUMMARY.md` links to them by path (e.g. `cli/README.md`). Only `SUMMARY.md` (the TOC, surfaced via `tree`) is skipped wholesale.
- `docs_get` on a heading returns its **full subtree**, not just the intro prose.
- `docs_tree` returns the curated `SUMMARY.md` table of contents for cheap browse/orientation.
- Parsing uses `pulldown-cmark` (the parser mdBook itself uses); the corpus is embedded with `rust-embed`, **uncompressed** — the `compression` feature is incompatible with the workspace (Cargo unifies it onto other `rust-embed` users like utoipa-swagger-ui, which it then breaks; see the note in `fluree-db-docs/Cargo.toml`). Binary grows ~3 MB.

## Installer

IDE registration is hoisted to **`fluree mcp install [--ide] [--server memory|docs|all]`**, which registers both MCP servers in one step and needs **no `.fluree/` directory** (so `--server docs` works in a non-Fluree project). `fluree memory mcp-install` is kept as a back-compat alias that registers the memory server only (equivalent to `--server memory`).

## Validation

- `fluree-db-docs`: 15 unit tests (parsing, anchors, BM25 ranking incl. the property-path case, subtree get, nav exclusion, tree, plus README section-index coverage and a "every tree path is fetchable" guard).
- `fluree-db-cli`: 96 lib tests pass (35 installer tests, incl. `--server` filtering and dual-server registration).
- Clippy clean; MCP server verified over stdio (`tools/list` + `tools/call`).

## Docs

New `cli/docs.md`; `cli/mcp.md` gains the `install` subcommand. The `docs/memory/**` pages are intentionally left on `main` (memory onboarding stays memory-only, pointing at `fluree memory mcp-install`) so the docs feature is self-contained.

## Notes for review

- Built from real-usage feedback (subtree `get`, nav exclusion, and `docs_tree` browse were all added in response to a session kicking the tires).
- A complementary `llms.txt` (docs-website) effort is tracked separately and shares only the `docs/` source.


bplatz:
## Design feedback: converge MCP on one configurable surface (toolsets), not one server per feature

Great crate and a genuinely useful feature. Before this lands, I want to lock the **MCP command/DX surface** so we don't ship a shape we have to unwind once database-over-stdio arrives.

**Direction:** treat MCP as a *single configurable surface* exposing a selectable set of **toolsets** — `memory`, `database`, `docs` — over two transports (**stdio** via the CLI, **HTTP** via the server's `/mcp`). Not N separate servers each with a static tool set.

### Target command surface (eventual)

```
fluree mcp init   [--ide auto|claude-code|cursor|vscode|windsurf|zed] [--toolsets memory,database,docs | all]
fluree mcp serve  [--transport stdio|http] [--toolsets … | all]   # spawned BY the IDE; run by hand only to debug
fluree mcp status                                                  # what's enabled / where it's installed
```

Server parity — the same surface, same vocabulary, over HTTP:

```toml
[server.mcp]
enabled  = true
toolsets = ["database", "docs"]   # /mcp exposes exactly these
```

Human-facing docs CLI stays as-is (it's transport/model-agnostic and good):

```
fluree docs search | get | examples | tree
```

**The three toolsets:**
- `memory` — `memory_*` (local store; lazy-created)
- `database` — `sparql_query`, `get_data_model` (operate on a ledger). **NOT available over stdio today** — these tools currently exist only on the running server's HTTP `/mcp`. Exposing them via `fluree mcp serve --toolsets database` (no server required) is **future work**; listed here so the command surface is designed for it now. The server should keep exposing them over HTTP regardless.
- `docs` — `docs_*` (embedded corpus; stateless)

### DX invariants (the anti-footgun rules)

- **`serve` is self-sufficient.** It lazy-inits everything (store dirs, etc.) — it never requires `init` to have run first. Confirmed today: the memory service calls `ensure_initialized()` in every tool handler and the serve path falls back to global dirs.
- **`init` only writes IDE config** (and pins per-toolset context). It does **not** run `serve`; for stdio the IDE spawns `serve` on demand each session. `init` should finish with a "reload your editor to activate" hint so we don't get "I installed it but the tools aren't there."
- **Each toolset lazy-resolves its context or returns a clear, actionable error** — never a silent empty result. Especially `database` with no ledger configured.

### Current → target mapping

| Today (incl. this PR) | Target | Notes |
|---|---|---|
| `fluree mcp serve` (memory, stdio) | `fluree mcp serve --toolsets memory` | unified `serve`; bare `serve` keeps memory as the default for back-compat |
| `fluree mcp install --server memory\|docs\|all` *(this PR)* | `fluree mcp init --toolsets memory\|docs\|all` | rename `install`→`init`, `--server`→`--toolsets`; default `all` |
| `fluree docs serve` *(this PR)* | `fluree mcp serve --toolsets docs` | no separate `fluree-docs` server and no second IDE entry |
| `fluree docs search\|get\|examples\|tree` *(this PR)* | unchanged — **keep** | model-agnostic human CLI |
| `fluree memory mcp-install [--ide]` | `fluree mcp init --toolsets memory` | deprecate the alias |
| `fluree memory init` (store init **+** Phase-2 IDE install) | store → lazy/optional; IDE install → `fluree mcp init` | split the two concerns it currently fuses |
| server `[server.mcp]` `/mcp` (database tools) | same + `toolsets` config | one surface, two transports |

### Deprecate / stop using

- `fluree mcp install` → **`fluree mcp init`** (keep `install` as a hidden alias).
- `--server` flag → **`--toolsets`**.
- **`fluree docs serve` as a standalone MCP server** → fold `docs_*` into the unified `serve` surface as the `docs` toolset (drop the separate server + second IDE registration).
- `fluree memory mcp-install` → `fluree mcp init --toolsets memory` (back-compat alias only).
- **`fluree memory init`'s IDE-install behavior** (Phase 2) → remove; since `serve` lazy-inits the store, the command can likely be retired entirely (to git-commit the store, just commit `.fluree-memory/` after first use). Keep a thin alias that points at `fluree mcp init`.

### What this means for THIS PR

- **Keep:** the `fluree-db-docs` crate (embedding, parser, BM25, examples) and the `fluree docs` human CLI — all model-agnostic and solid.
- **Change:** don't ship the standalone `fluree-docs` MCP server or the second `fluree-docs` IDE entry. Instead register `docs_*` as the **`docs` toolset inside the unified MCP service**, installed via `fluree mcp init` (default `all`).
- Net effect: one IDE entry (`fluree`), one `serve`, one `init`, toolsets select what's on.

### Open decisions (not blocking the crate, but settle before the MCP packaging)

1. **`database` in a non-database project:** install it dormant with a clear "no ledger configured" error, **or** make the default context-aware (enable `database` only when a ledger is detected). Leaning context-aware, with `--toolsets all` as the override.
2. Verb is **`init`** (preferred over `install`) — namespaced under `mcp`, reads as "set up MCP integration."
3. Confirm retiring `fluree memory init` rather than keeping two `init` verbs.

Jackamus29:
**Implemented the unified toolsets surface (pushed: `56a0a3fa2`, `4893a20e0`)**

Thanks for the design steer, Brian — converged the MCP surface onto a single configurable server with selectable toolsets as proposed, before this lands. Summary of what changed and a couple of things to confirm.

**Architecture.** New `fluree-db-mcp` crate hosts one `FlureeMcpService`. Each toolset is a named rmcp `#[tool_router]` block on that single type (`memory`, `docs`), and `new()` composes only the enabled toolsets' routers into the router that `#[tool_handler]` dispatches against — so `tools/list` reflects exactly the selection. The old `MemoryToolService` / `DocsToolService` wrappers are gone; the domain crates keep only their logic and dropped their `mcp` features. The server's HTTP `database` tools (`FlureeToolService`) are untouched. The new crate is reusable by the server later for the HTTP `/mcp` toolsets parity.

**Command surface.**
- `fluree mcp init [--ide] [--toolsets memory,docs|all]` — writes a single IDE entry named `fluree`, and migrates away any legacy `fluree-memory` / `fluree-docs` entries from a prior install. Ends with a "reload your editor" hint.
- `fluree mcp serve [--toolsets …]` — bare `serve` defaults to `memory` for back-compat; `init` always writes an explicit `--toolsets`. Docs-only is stateless (no `.fluree/` required or created).
- `fluree mcp status` — installed toolsets per detected IDE.
- `fluree docs serve` removed; the human `fluree docs search|get|examples|tree` kept.
- `fluree mcp install`, `fluree memory mcp-install`, `fluree memory init` kept as hidden back-compat aliases. Per your note, `memory init` no longer creates the store up front — it's lazy-created on first use.

**DX invariants honored.** `serve` lazy-inits everything (never needs `init` first); `init` only writes IDE config; the single `fluree` entry replaces the per-feature ones.

**Verification.** Unit tests cover toolset→tool-list composition, single-entry install, and legacy-entry migration (mcp 8 / cli-lib 84 / docs 15 / memory 60). Also exercised the live server from an editor session: `--toolsets all` surfaces all ten tools under one `fluree` server; `docs_search` / `docs_get` return version-pinned results; the memory toolset lazy-inits cleanly.

**Deferred (intentionally not in this PR):**
1. **`database` toolset over stdio** + server `[server.mcp] toolsets` HTTP parity. `Toolset::Database` is reserved so the surface is ready for it.
2. **Tool-naming convention** — `memory_*` / `docs_*` are namespaced, but the server's database tools are `sparql_query` / `get_data_model` (unprefixed). Worth deciding before `database` lands, since it's a breaking rename later.

(Filing the memory `.ttl` migration question as a separate comment below.)

Jackamus29:
**Question for @bplatz: do we preserve the legacy memory ledger→`.ttl` migration?**

Retiring `fluree memory init` (now a thin alias) also removed its one-time export of legacy *ledger-only* memories into `repo.ttl` / `user.ttl`. Flagging the data-safety implication so we decide deliberately.

**Context.** The memory store keeps memories in two places: the `.ttl` files (authoritative, git-tracked) and the `__memory` ledger (a derived cache rebuilt from the `.ttl` files on hash mismatch). Early on, memories lived *only* in the ledger; the old `memory init` bridged those stores by reading the ledger and writing the `.ttl` files.

**Why it's not an immediate regression.** On a legacy ledger-only store the lazy path is safe today: `initialize()` is skipped (ledger already exists, so no empty `.ttl` files get created), and `ensure_synced()` no-ops when no `.ttl` files exist — so reads keep working off the ledger.

**The latent risk.** Such a store is never exported to `.ttl`, so the memories aren't git-shareable, and they'd be **wiped on the next forced ledger rebuild** (a `SCHEMA_VERSION_SALT` bump, an external `.ttl` edit, or any hash mismatch), since the rebuild repopulates the ledger from the empty `.ttl` files.

**Where we actually are.** v4.1 (released yesterday) is our first public release, so no public users are affected. But internal devs have been on memory for a few months and may have ledger-only stores on their machines. So this is a judgment call, not a hard compat requirement — your call:
- **Accept it** — affected internal devs re-add the handful of memories; the migration stays dropped (current state).
- **Preserve it** — fold the export into the store's lazy init (on first use: ledger has memories + `.ttl` empty/absent → export ledger→`.ttl` and stamp the hash), so months-old internal memories don't evaporate on the next rebuild.

Happy to implement the lazy-init version if you'd prefer to keep it.

bplatz (approved):
Looks good, I don't see the data loss issue you mention, the memory .ttl is the system of record here (not the other way around), but maybe I'm missing something.

Jackamus29:
> Looks good, I don't see the data loss issue you mention, the memory .ttl is the system of record here (not the other way around), but maybe I'm missing something.

Yeah - false alarm. 
Looks like Claude got confused by a process in the old `mcp init`.
Thanks Brian!

PR #1345, opened 2026-06-17
feat: streaming query endpoint (NDJSON) with keep-alive heartbeats

bplatz:
## Summary

Adds an opt-in **streaming query endpoint** that emits SELECT results
incrementally as newline-delimited JSON (`application/x-ndjson`) instead of
buffering the entire result set into one response body.

Streaming has the obvious wins — incremental delivery, lower peak memory, and
pipelined client consumption — but a concrete driver here was **infrastructure
timeouts**: AWS CloudFront (and ALB) drop a connection after ~60s of idle time.
A long-running query that takes more than 60s to produce its first bytes (or
stalls mid-stream on a blocking operator) gets its connection killed before the
client ever sees a result. This endpoint keeps **bytes flowing across the wire**
during execution via a periodic **heartbeat**, so long queries survive the
proxy idle timeout.

The standard `/query` path is untouched — this is a separate route family with
its own handler, so the benchmark-critical buffered path never pays for any of
the streaming machinery.

## Endpoints

- `POST /v1/fluree/stream/query/<ledger...>` — ledger-scoped (ledger in path)
- `POST /v1/fluree/stream/query` — connection-scoped (ledgers from the request:
  JSON-LD `from`/`fromNamed` or SPARQL `FROM`)

Both negotiate JSON-LD (`application/json`) vs SPARQL (`application/sparql-query`)
exactly like `/query`.

## The NDJSON record protocol

One self-describing JSON object per line:

- `{"type":"head","vars":[...]}` — emitted first (before the first batch pull),
  so the client learns the schema immediately and the idle clock starts fresh.
- `{"type":"row","row":{...}}` — one per result row (SPARQL-Results-JSON binding
  object, byte-identical to what `/query` returns).
- `{"type":"heartbeat","t_ms":N,"fuel":F?}` — keep-alive during stalls; carries
  the live fuel total as a progress signal.
- `{"type":"end","rows":N,...}` / `{"type":"error","error":{"code":...},"rows":N}`
  — exactly one terminal record. Absence of a terminal = truncated/dropped
  stream (the client must treat it as failure, since the `200 OK` is committed at
  the first byte).

## The heartbeat (the CloudFront fix)

A wall-clock timer in the transport layer races the producer channel
(`select!` over `rx.recv()` vs `interval.tick()`). Because it's independent of
query execution, it fires even while a blocking `ORDER BY`/`GROUP BY` is
draining and producing no rows — exactly the window that would otherwise trip
the proxy idle timeout. Interval is configurable via `FLUREE_STREAM_HEARTBEAT_MS`
/ `--stream-heartbeat-ms` (default **15s**, well under ~60s; `0` disables). The
heartbeat reads fuel via a lock-free atomic, so it never contends the query.

## Design

- **Share the engine, fork the driver.** Planner + operators are unchanged. A
  new `BatchSink` trait backs both the buffered runner (collects to `Vec`) and a
  streaming runner (formats + flushes each batch); `execute_prepared` and
  `execute_prepared_streaming` share one context-build so they can't drift.
- **Policy parity with `/query`.** Policy / `from`/`fromNamed` / multi-ledger
  queries route to the same connection/dataset path and are enforced
  identically (JSON-LD everywhere; SPARQL on the ledger-scoped form). Bearer
  scope, `Fluree-Min-T` freshness, and stored default-context are all applied
  before planning, matching `/query`.
- Client disconnect cancels the producer (disconnect guard on the response
  stream); the query timeout still applies.

## Not supported (rejected `4xx`, use `/query`)

ASK, CONSTRUCT/DESCRIBE, `selectOne`, JSON-LD hydration, and history
(JSON-LD `to` / SPARQL `FROM … TO …`). Connection-scoped SPARQL refuses policy
signals (no single ledger to resolve identity against).

## Testing

- 5 API-producer tests + 17 HTTP integration tests covering: JSON-LD/SPARQL
  SELECT framing, multi-ledger `from` union (both forms), **policy enforcement
  filtering streamed rows**, **bearer-scope blocking cross-ledger reads**, error
  codes (e.g. `fuel_exhausted`), unsupported-shape rejections, and
  heartbeat-disabled behavior.
- Full CI green: `fmt --check`, `clippy --all --all-features --all-targets`,
  `nextest --workspace --all-features` (the one failure is the known LocalStack
  testcontainer flake, which passes in isolation).

## Docs

`docs/api/streaming-query.md` (full protocol, terminal/truncation contract,
policy behavior, heartbeat/CloudFront note, client examples), plus
`endpoints.md`, `configuration.md`, and `SUMMARY.md`.

## Known gap

The heartbeat *firing* is wired and unit-tested but not asserted end-to-end on a
genuinely slow query (needs a deterministically-stalled query); follow-up.


fluree-db-api/src/view/stream_query.rs
```diff
@@ -0,0 +1,555 @@
...
+            result: &meta,
+            compactor: &compactor,
+            head_vars: &head_vars,
+            var_names: &var_names,
+            tx: tx.clone(),
+            rows: 0,
+            buf: String::new(),
+        };
+
+        let noop = crate::NoOpR2rmlProvider::new();
```
aaj3f: In the buffered dataset path, there are 2 entrypoints: `query_dataset_with_options` (no-op) and  `query_dataset_with_r2rml_options` (real providers, `dataset_query.rs:148-245`). If we ever introduce R2RML providers for a mapped graph source, a streaming dataset query on that source would not apply the R2RML mapping.

fluree-db-server/tests/stream_query_integration.rs
aaj3f, line 1: These are solid tests. The only additions we may want might be a test exercising an emitted heartbeat and/or a client-disconnect cancellation test

aaj3f (approved):
This is neat! I like it

The dataset-query path nicely gets the abstraction from `execute_dataset_into_with_r2rml()` regardless of whether buffered or streaming, but the single-ledger path doesn't. `execute_view_streaming` rebuilds `ContextConfig` independently and duplicates the pattern used with/for the buffered builder pattern.

fluree-db-api/src/view/stream_query.rs
bplatz: Addressed in 7e90e457e27f48d23489f5c01c6600cbc0ccf9fe

fluree-db-server/tests/stream_query_integration.rs
bplatz: Addressed in 9f783f030acdaf5e659dcbadb1e8e8bff4c243d7

bplatz:
Addressed in 9f783f030acdaf5e659dcbadb1e8e8bff4c243d7

PR #1348, opened 2026-06-17
Feature/raft

zonotope:
## Summary

Adds an end-to-end Raft consensus mode to Fluree DB. Every node holds a replicated nameservice state machine, transactions flow through a per-branch command queue with one worker on the current leader, and read paths observe committed state without going through Raft RPC. The branch lands on top of `feature/consensus` (which introduced the `Committer` / `LocalCommitter` / `CachingCommitter` split) and supplies the cluster-mode peer of that split. Gated behind the new `raft` feature flag on both `fluree-db-consensus` and `fluree-db-server`; monolithic builds remain unaffected.

### Replicated state machine

- `fluree-db-consensus/src/raft/state_machine.rs` — pure `apply(state, command, log_index) -> Response` over a `NameServiceState` that carries ledger records, ref entries, retraction marks, idempotency cache, per-branch queues, and admin-clear markers. Commands span ledger lifecycle (`CreateLedger`, `CreateBranch`, `RetractLedger`, `PurgeLedger`, `DropBranch`, `ResetHead`), head advancement (`AdvanceIndexHead`, `ApplyHead`), and queue-mediated submission (`EnqueueCommand`, `PoisonQueueEntry`, `EvictIdempotency`, `ReleaseContent`).
- `state_machine_adapter.rs` — openraft `RaftStateMachine` adapter. Runs `apply` under a write lock, then dispatches side effects after the lock drops: event-bus notifications, waiter-map resolutions, CAS-release fan-out. Snapshots persist through `RaftStorage::snapshots()`.
- `storage.rs` + `storage/fs.rs` + `storage/memory.rs` — log/vote/snapshot persistence with an on-disk store for production and an in-memory store for tests.
- `log_adapter.rs` — adapts openraft `RaftLogStorage` over the persistence layer.

### Networking and admin

- `raft/network.rs` — three POST endpoints (`append-entries`, `vote`, `install-snapshot`) with postcard wire format. Listens on a private VPC-internal port; auth is delegated to the embedding network.
- `raft/admin.rs` — operator endpoints for cluster bootstrap (`initialize`), membership change (`add-learner`, `change-membership`), and status (`metrics`, `leader`).
- `raft/forward.rs` — follower-side middleware that proxies leader-only client routes to the current leader using the membership snapshot. Body-limited to 64 MiB to bound follower memory under hostile traffic.

### Read paths and event distribution

- `raft/nameservice.rs::RaftNameService` — `NameServiceLookup` + `LifecycleNameService` + `IndexingNameService` + `CommitPublisher` over the shared state handle. Lookups read directly from the replicated state; writes propose through Raft. `publish_commit` looks the queue front up under the read lock, builds an `ApplyHead`, and surfaces `QueueDesync` on stepped-down leaders so the worker can stash-clean instead of falsely succeeding.
- `LedgerEventBus` wired through `state_machine_adapter` so subscribers observe commit/index/lifecycle events after each successful apply. Server-side maintenance tasks subscribe via the existing `NameServiceEvent` channel.

### Queue-mediated transactor

- `raft/queued_transactor.rs::QueuedTransactor` — `Committer` impl that routes `transact` / `push` / `revert` / `merge` / `rebase` through a single submission path: serialize a `QueuedRequest` envelope into shared CAS, propose `EnqueueCommand` with the resulting `request_cid`, register a `WaiterMap` waiter on the assigned `queue_id`, await with a bounded timeout. Idempotency-keyed submissions retry through the same path on stale-waiter / leader-transition cases — the state machine's cache short-circuits the second propose into an `IdempotencyHit`.
- `raft/waiter.rs::WaiterMap` — per-process oneshot map keyed by `queue_id`, with `abort_all_for_branch(&ref_key)` for the admin-clear sweep.
- `raft/staged_receipt.rs::StagedReceiptMap` — per-process side channel the worker stashes typed `AppliedReceipt`s into before proposing `ApplyHead`; the adapter takes them during waiter resolution. Lets transactors return rich per-op detail (conflict counts, tally, indexing status) without replicating any of it through Raft.

### Commit worker

- `raft/commit_worker.rs::CommitWorker` — leader-only loop that drains every per-branch queue front concurrently. For each entry: read the envelope from CAS, dispatch by `body_kind`, stage and persist through the existing `Fluree` pipeline (`prepare_transact` / `prepare_push` / `prepare_revert` / `prepare_merge` / `prepare_rebase`), then stash the receipt, propose `publish_commit`, and finalize the local state under the held write guard on success. On any failure between stage and publish the install is dropped, leaving the local cache aligned with the unchanged replicated head.
- `raft/eviction_scheduler.rs::EvictionScheduler` — leader-only periodic `EvictIdempotency` proposer. Carries cutoffs for both the idempotency cache (1h default) and the `recently_cleared` admin-clear marker map (5 min default).
- Both tasks are spawned by `RaftIntegration::spawn_leader_watcher`, which tracks leadership via `raft.metrics()` and tears the worker stack down cleanly on leadership loss.

### Server integration

- `fluree-db-server/src/raft.rs::RaftIntegration` — single-call bootstrap that opens storage, wires adapters, builds a pooled HTTP client, validates the raft config, and returns a clone-cheap handle. Carries the shared state, event bus, waiter map, staged-receipt map, leader forwarder, and the release-channel receiver.
- `FlureeServer::start` spawns the leader watcher, the per-node CAS-release task, and the listener for the private RPC port. The router layers in the forwarding middleware so leader-only client routes transparently proxy from a follower.
- New CLI commands: cluster `initialize` / `join` / `leave` / `members` / `metrics` plus config-file fields for node identity and storage path.

### Other crate-level changes

- `fluree-db-consensus` — `Committer` trait is now async-trait-shaped with `transact` / `push` / `revert` / `merge` / `rebase` methods that all return typed receipts. `CachingCommitter` wraps any inner `Committer` for in-process idempotency dedup ahead of the propose. `LocalCommitter` extracted from the legacy monolithic path so it composes with both the in-process and Raft routes.
- `fluree-db-nameservice` — `NameService` trait split into `NameServiceLookup` (read) and lifecycle/publisher writes, with a `LifecycleNameService` supertrait the API consumes. `IndexingNameService` carves off the indexer's contract from the broader publisher surface.
- `fluree-db-api` — `prepare_revert` / `prepare_merge` / `prepare_rebase` factored to build the resulting commit without writing the head, so the worker can stage atomically and commit through `publish_commit` instead of the old direct-write path.


fluree-db-consensus/src/raft/state_machine_adapter.rs
```diff
@@ -0,0 +1,1128 @@
...
+        Self::with_state(storage, Arc::new(RwLock::new(NameServiceState::default())))
+    }
+
+    /// Construct an adapter sharing the provided state handle. Use
+    /// this when the same state must be visible to a `RaftNameService`
+    /// constructed alongside the adapter.
+    pub fn with_state(storage: Arc<S>, state: SharedState) -> Self {
+        Self {
+            state,
+            last_applied: None,
```
bplatz, line 133: Boots empty and never restores the current snapshot, so `applied_state()` reports `last_applied: None`. After a snapshot + log purge (default `SnapshotPolicy`), restart replays only the purged post-`last_applied` tail and committed nameservice state is silently lost.

Restore `state`/`last_applied`/`last_membership` from `get_current_snapshot()` here (needs an async ctor), and gate `apply()` side effects above the restored index so recovery replay doesn't re-fire CAS releases.

fluree-db-consensus/src/raft/queued_transactor.rs
```diff
@@ -0,0 +1,1008 @@
...
+                    self.release_envelope(&full_ledger_id, &request_cid).await;
+                    return Err(SubmissionError::Execution {
+                        status: 500,
+                        message: format!("raft client_write failed: {e}"),
+                    });
+                }
+            };
+            match response.data {
+                SmResponse::Enqueued { queue_id, .. } | SmResponse::InFlight { queue_id, .. } => {
+                    let rx = self.waiter_map.register(queue_id, ref_key.clone());
```
bplatz, line 197: `register` runs only after `client_write` returns, so on a quiet branch the worker can apply `ApplyHead` and call `resolve_*` before the slot exists — `resolve_applied` is then a no-op and the outcome is dropped. An anonymous (no idempotency key) write returns a stranded 504 after committing, and the retry re-proposes under a fresh `request_cid` the state machine can't dedup → duplicate commit.

Register/claim the waiter before the worker can resolve (or base submission status on replicated state), and require an idempotency key for Raft writes.

fluree-db-consensus/src/raft/commit_worker.rs
```diff
@@ -0,0 +1,1025 @@
...
+        self.staged_receipts
+            .stash(entry.queue_id, ref_key.clone(), receipt);
+        match self.publish_head_advance(ref_key, commit_id, commit_t).await {
+            Ok(()) => {
+                if let Some(install) = install {
+                    self.finalize_local_state(install).await?;
+                }
+                Ok(())
+            }
+            Err(err) => {
```
bplatz, line 253: This drops the local install on *any* publish error, but `publish_head_advance` can return `Err` (notably `Fatal`, or a lost response) for an `ApplyHead` that actually committed. The replicated head then advances while this node's Fluree cache stays at its pre-stage head → silent stale reads.

On an ambiguous error, reconcile against the replicated head / idempotency before dropping the install, or refresh the local cache from the replicated head.

fluree-db-server/src/raft.rs
```diff
@@ -0,0 +1,534 @@
...
+        }
+
+        // Watcher shutdown: tear down any in-flight leader tasks.
+        abort_all(&mut current_tasks);
+    })
+}
+
+fn abort_all(handles: &mut Vec<JoinHandle<()>>) {
+    for handle in handles.drain(..) {
+        handle.abort();
```
bplatz, line 495: Handles are `abort()`ed but never awaited, so a leader flap can respawn a second commit-worker before the aborted one reaches a yield point; and aborting the watcher at its `.await` skips this cleanup entirely, leaking the leader tasks on shutdown. Use a cancellation token + awaited joins.

fluree-db-server/src/routes/submissions.rs
```diff
@@ -111,7 +111,7 @@ pub async fn submission_status(
     Path(params): Path<SubmissionStatusParams>,
 ) -> Json<SubmissionStateResponse> {
     let key = IdempotencyKey::new(params.key);
-    let lookup_state = state.consensus.status(&params.ledger, &key).await;
+    let lookup_state = state.committer.status(&params.ledger, &key).await;
```
bplatz, line 190: This route isn't leader-forwarded, so a follower reads its per-process committer state and returns not-found for a write the leader accepted. Forward status to the leader, or serve it from replicated idempotency state.

fluree-db-consensus/src/raft/network.rs
```diff
@@ -0,0 +1,505 @@
...
+///
+/// Example:
+/// ```ignore
+/// let raft = Arc::clone(&raft_handle);
+/// let app = axum::Router::new()
+///     .nest("/raft", fluree_db_consensus::raft::network::router(raft));
+/// ```
+pub fn router(raft: Arc<Raft<TypeConfig>>) -> Router {
+    Router::new()
+        .route(PATH_APPEND_ENTRIES, post(handle_append_entries))
```
bplatz: Only `install-snapshot` disables axum's 2 MiB default body limit; `append-entries` and `vote` keep it, so a large catch-up batch hits 413 and stalls replication to a lagging follower. Disable/raise the limit on these routes too.

fluree-db-consensus/src/raft/state_machine.rs
```diff
@@ -0,0 +1,3477 @@
...
+            reason: DesyncReason::WrongFront {
+                actual_queue_id: actual_front,
+            },
+        }));
+    }
+
+    Ok(queue.pop_front().expect("non-empty checked above"))
+}
+
+fn apply_head(
```
bplatz: `apply_head` (and `reset_head`) advance the head without reconciling `state.retracted`, while `get_ref` ignores retraction and `lookup` exposes it. Needs a product decision (reject writes to retracted branches vs head-advance un-retracts), then make the two lookup surfaces consistent.

fluree-db-consensus/src/raft/queued_transactor.rs
```diff
@@ -0,0 +1,1008 @@
...
+            SubmissionError::Execution {
+                status: 400,
+                message: format!("invalid ledger_id: {e}"),
+            }
+        })?;
+        let ref_key = RefKey::new(&ledger_name, &branch);
+
+        let idempotency_cache_key = idempotency_key
+            .as_ref()
+            .map(|k| IdempotencyCacheKey::new(ledger_id.clone(), k.clone()));
```
bplatz, line 397: Keyed on the raw `ledger_id` while `split_ledger_id` already produced the normalized `ledger_name`/`branch` above — so `transact("test/db")` and `transact("test/db:main")` land in different idempotency keys and different CAS namespaces (line 317) for the same branch. Normalize once (e.g. `format!("{ledger_name}:{branch}")`) and use that for both the key and `content_store(...)`, matching the other submission paths.

fluree-db-consensus/src/lib.rs
```diff
@@ -108,13 +156,213 @@ impl TransactionBody {
...
+        Ok(serde_json::from_slice(bytes)?)
+    }
+
+    /// Canonical bytes used to derive the idempotency body CID. Skips
+    /// per-request transient context (commit-opts timestamp, tracking
+    /// flags, governance metadata) so a retry of the same logical
+    /// request produces the same body CID even when the client recomputes
+    /// those fields between attempts. The full envelope still carries
+    /// them — only the idempotency comparison is normalized.
+    pub fn canonical_body_bytes(&self) -> Result<Vec<u8>, QueuedRequestCodecError> {
```
bplatz, line 362: Hashes only the body and skips `txn_opts`/governance, so the same idempotency key + same body under a different policy/governing identity returns the first receipt instead of re-executing under the new context. Fold the governing context into the comparison (or bind the key to it).

fluree-db-consensus/src/raft/commit_worker.rs
```diff
@@ -0,0 +1,1025 @@
...
+    /// the entry has reached a terminal state in the queue (advanced
+    /// or poisoned).
+    pub async fn process_entry(
+        &self,
+        ref_key: &RefKey,
+        entry: QueueEntry,
+    ) -> Result<(), WorkerError> {
+        match self.try_advance_head(ref_key, &entry).await {
+            Ok(()) => Ok(()),
+            Err(WorkerError::Stage(reason)) => {
```
bplatz: Any staging failure poisons the entry with no retry budget, so a transient CAS/storage/load blip permanently rejects the submission (and a same-key retry returns the cached failure). Classify transient vs permanent errors and bounded-retry the transient ones before poisoning.

fluree-db-consensus/src/raft/forward.rs
```diff
@@ -0,0 +1,333 @@
...
+        }
+        let metrics = self.raft.metrics().borrow().clone();
+        let leader_node = metrics
+            .membership_config
+            .nodes()
+            .find(|(id, _)| **id == leader_id)
+            .map(|(_, node)| node.clone());
+        match leader_node {
+            Some(node) if !node.client_addr.is_empty() => {
+                ForwardDecision::Forward(node.client_addr)
```
bplatz, line 153: Leader `client_addr` comes from the membership snapshot, which can be stale right after a membership change → forwards to a decommissioned address. There's also no hop/loop guard, so two nodes with disagreeing leader views can ping-pong (each re-buffering the body). Add a hop header and stricter leader-URL validation.

fluree-db-server/src/lib.rs
```diff
@@ -242,6 +267,22 @@ impl FlureeServer {
...
+        #[cfg(feature = "raft")]
+        if let Some(task) = self.raft_leader_watcher {
+            // Aborting the watcher's outer task drops its metrics
+            // receiver; on its way down it also aborts every
+            // currently-running leader task — see
+            // `raft::spawn_leader_watcher` for the teardown path.
+            task.abort();
+        }
+        #[cfg(feature = "raft")]
+        if let Some(task) = self.raft_release_task {
```
bplatz: Aborting the release task on shutdown discards `(ledger, cid)` pairs still buffered in the channel → orphaned CAS blobs; drain before abort. (Also `take_release_receiver` is single-shot, so a second `build()` of the same integration spawns no release task at all.)

fluree-db-server/src/routes/mod.rs
```diff
@@ -56,13 +85,48 @@ pub fn build_router(state: Arc<AppState>) -> Router {
     let v1_admin_protected_routes =
         v1_admin_protected_routes.route("/iceberg/map", post(iceberg::iceberg_map));
 
+    // Forward to the Raft leader (when running in Raft mode) before
+    // admin-token auth: an out-of-date follower with stale credentials
+    // shouldn't reject a request the leader would accept.
+    let v1_admin_protected_routes = apply_leader_forward(v1_admin_protected_routes, &state);
```
bplatz: The leader-forward layer wraps the whole group, including the read-only `/export/*ledger` route added above, so followers proxy exports to the leader and lose read-scaling. Keep read-only routes off the leader-forward layer.

fluree-db-server/src/config.rs
```diff
@@ -872,6 +915,29 @@ impl ServerConfig {
...
+        // listen_addr must all be set, and proxy storage is
+        // incompatible with raft (raft replicates writes via the
+        // log; proxy mode forwards to a remote tx server).
+        #[cfg(feature = "raft")]
+        if self.raft_enabled {
+            if self.raft_node_id.is_none() {
+                return Err("raft.enabled=true requires --raft-node-id".to_string());
+            }
+            if self.raft_storage_path.is_none() {
+                return Err("raft.enabled=true requires --raft-storage-path".to_string());
```
bplatz, line 1027: Also assert `raft_storage_path != storage_path` here — otherwise pointing the raft log/snapshot tree at the data dir (or two co-located nodes sharing a path) passes validation and corrupts both.

fluree-db-consensus/src/raft/commit_worker.rs
```diff
@@ -0,0 +1,1025 @@
...
+                    Err(WorkerError::Raft(error)) => {
+                        warn!(
+                            ledger_id = %ref_key.ledger_id,
+                            branch = %ref_key.branch,
+                            queue_id,
+                            error = %error,
+                            "raft publish failed; backing off and re-polling"
+                        );
+                        raft_blocked = true;
+                        break;
```
bplatz: `break` on one branch's raft error abandons every other branch's front for this tick, so a single stuck branch head-of-line-blocks all others. Skip the failed branch and continue the drain instead of breaking the whole tick.

bplatz (approved):
LGTM

PR #1351, opened 2026-06-17
fix(policy): close view/modify enforcement gaps across operators, reasoning, and turtle writes

bplatz:
## Summary

Consolidates view-policy enforcement behind a single bypass gate and closes the data-emitting paths that previously read raw flakes without per-flake filtering, hardens the policy combining algorithm, restricts query-time reasoning rule injection under policy, and extends `f:modify` enforcement to Turtle/N-Triples/TriG writes.

## View-policy enforcement

- **Canonical bypass gate.** Consolidated the duplicated "no view filtering needed" check into `ExecutionContext::allow_unfiltered()`, and closed a batched-join/probe-lane early-return that skipped the gate before the overlay-free fast path.
- **Plan-time predicate-coverage skip.** Single-predicate cardinality / scalar-aggregate / string fast paths now run (instead of bailing to a filtered scan) when the active policy provably cannot restrict the scanned predicate, while still forcing the filtered path when it could.
- **Closed per-operator leaks** — paths that emitted subjects/edges/hits from raw reads with no per-flake filter now route through the policy filter:
  - property-path edge traversal (hidden edges are no longer walked)
  - inline BM25, vector, geo, and S2 search hits (a hit is returned only if the identity can view the indexed/searched flake)
  - the stats count-by-predicate wildcard `GROUP BY ?p` fast path (delegates to a policy-enforced fallback under a non-root policy)
  - transaction `WHERE`-clause reads (a writer can no longer launder hidden values via a conditional match)
  - hydration (direct-id selects / nested-ref expansion) and commit/block detail, which now populate the subject class cache and run the async evaluator so `f:onClass` and `f:query` are honored

## Combining algorithm

- **Required policies are AND gates.** Within the required subset every policy must grant; a required `f:query` returning no rows denies and can no longer be bypassed by a co-located required allow that happened to be evaluated first (an order-dependent fail-open).
- **Per-flake filter errors propagate.** Execution failures during filtering (malformed `f:query`, storage IO, cancellation, fuel exhaustion) now fail the query instead of being swallowed into a silent under-count.

## Reasoning under policy

Reasoning-derived facts already pass through the same per-flake view filter as base data, but the engine does not trace a derived fact's provenance. Query-time datalog rule injection is now admin-only: under a non-root view policy, caller-supplied `rules` are stripped before execution, since a rule with a viewable head could otherwise surface hidden data the policy author never anticipated. Database-stored rules and OWL/RDFS reasoning are unaffected. The docs spell out the posture that, with reasoning enabled under a policy, the policy must cover the derived properties/classes (or run `default-allow: false`).

## Write policy

`f:modify` is now enforced on Turtle/N-Triples/TriG writes — both the server write route and the direct-flake staging path carry the request's policy into staging — closing a path that bypassed fine-grained modify rules while coarse ledger write-access still applied.

## Docs

Updated `docs/security/policy-in-queries.md`, `docs/security/policy-in-transactions.md`, and `docs/query/reasoning.md` to document per-operation enforcement, the required-AND combining rule, the reasoning-under-policy posture, and wire-format-independent modify enforcement.

## New test coverage

- `where_read_respects_view_policy` — transaction `WHERE` is filtered by view policy (a salary-probe matches nothing when the property is hidden; a viewable property still matches)
- `policy_onclass_denies_hydration_only_subject` — `f:onClass` is honored on a subject reached only via hydration
- `turtle_insert_enforces_modify_policy` — a Turtle write touching a modify-denied property is rejected
- `required_policies_are_and_gated` — two required policies (allow + failing `f:query`) deny in both insertion orders
- `datalog_query_time_rules_stripped_under_non_root_policy` — a query-time rule fires without a policy but is stripped under a non-root policy
- `policy_count_respects_predicate_coverage` and `policy_stats_count_by_predicate_uses_filtered_fallback` — the fast-path predicate-coverage skip and the policy-enforced count fallback
- `policy_property_path_filters_hidden_edges` — property paths do not traverse hidden edges
- `bm25_search_enforces_view_policy_on_indexed_flake`, `vector_search_enforces_view_policy_on_embedding_flake`, `geo_search_enforces_view_policy_on_location_flake` — inline search hits respect view policy


fluree-db-query/src/bm25/operator.rs
```diff
@@ -225,6 +225,37 @@ impl Bm25SearchOperator {
...
+    /// is dropped.
+    ///
+    /// Scope: enforced only on the inline (local index) path; the dedicated
+    /// search-provider path is out of scope (the engine has no local flakes to
+    /// check against). The indexed-flake read runs in the single active graph;
+    /// under a non-root policy in multi-graph mode the hit is conservatively
+    /// hidden (no leak). No-op for root / no policy.
+    /// Indexed-flake readability for a BM25 hit: visible iff the identity can
+    /// view at least one of the subject's flakes on a BM25-indexed property (the
+    /// text that could have produced the match). Only the inline (local index)
```
zonotope: Minor, but this just repeats lines 231-234. One of these chunks should be deleted.

fluree-db-query/src/bm25/operator.rs
```diff
@@ -225,6 +225,37 @@ impl Bm25SearchOperator {
...
+    /// path is checked; the dedicated search-provider path loads no index here.
+    async fn hit_readable(&self, ctx: &ExecutionContext<'_>, subject_iri: &str) -> Result<bool> {
+        let Some(index) = self.index.as_ref() else {
+            return Ok(true);
```
zonotope, line 257: First, my read here is that if the index is `Some` then we're in local provider mode and have access to the subject flakes that produce the search hit and can do the policy check below, and if the index is `None`, then we're in search provider mode and have no flakes to check against. If this is the case, then I think it would be much more readable if `index` was renamed to `local_index`, or, at the very least, there was a comment on this line describing the situation.

Next, if we don't have flakes to check policy against, then would it be safer to default to false and assume the user doesn't have access in a non-root policy or to throw an error?

fluree-db-query/src/context.rs
```diff
@@ -617,6 +617,23 @@ impl<'a> ExecutionContext<'a> {
...
+    /// enforcement. The only operator that actually removes policy-hidden rows
+    /// is the range-fallback scan ([`BinaryScanOperator::filter_flakes_by_policy`]);
+    /// every fast path and raw-leaflet reader that does *not* route emitted
+    /// flakes through that filter MUST gate on `allow_unfiltered()` and decline
+    /// (fall back to the filtered scan) when it returns false. Adding a new
+    /// data-emitting operator that reads the index directly without consulting
+    /// this — or applying its own filter — is a policy leak.
+    #[inline]
+    pub fn allow_unfiltered(&self) -> bool {
+        !self.has_policy()
```
zonotope, line 643: `has_policy` doesn't take datasets into account, so this function could allow an unfiltered dataset query when one of the constituents is filtered.

fluree-db-query/src/fast_string_fold.rs
```diff
@@ -208,7 +209,19 @@ pub fn predicate_string_fold_operator(
...
+                if !matches!(
+                    cursor_fast_path_for_predicate(ctx, &pred_sid),
+                    PredicateFastPath::Allow
+                ) {
+                    return Ok(None);
+                }
+            }
+            let Some(store) = fast_path_store_policy_cleared(ctx) else {
                 return Ok(None);
             };
```
zonotope, line 216: This same structure is repeated almost verbatim in fast_string_prefix_count_all, fast_sum_strlen_group_concat, and `fast_predicate_scalar_agg. There should be a single helper to extract the store, normalize the predicate sid, and make the fast path decision that all of these can use.

zonotope (approved):
lgtm besides what I flagged.

fluree-db-query/src/bm25/operator.rs
bplatz: Addressed in d2d1e22afcfab39809f7f7c2c09cd29924d98316 — deleted the duplicated doc block on `hit_readable`.

fluree-db-query/src/bm25/operator.rs
bplatz: Addressed in d2d1e22afcfab39809f7f7c2c09cd29924d98316 — renamed `index` to `local_index` (with a doc note that `Some` = local-flake mode and `None` = search-provider mode), and changed the fail-open to fail **closed**: under an active non-root policy in search-provider mode (no local flakes to check) the hit is now hidden rather than returned unfiltered.

fluree-db-query/src/context.rs
bplatz: Addressed in d2d1e22afcfab39809f7f7c2c09cd29924d98316 — `has_policy()` is now dataset-aware via the new `DataSet::has_any_policy()`, so it reports a policy whenever any constituent graph is filtered (the enforcer lives on the `GraphRef` in dataset mode). `allow_unfiltered()` is now sound even for callers that don't fold to a per-graph context first.

fluree-db-query/src/fast_string_fold.rs
bplatz: Addressed in d2d1e22afcfab39809f7f7c2c09cd29924d98316 — extracted `predicate_fast_path_allowed` (the shared normalize + fast-path decision) and `fast_path_store_for_predicate` (adds the policy-cleared store), and reused them across the single-predicate string fast paths. Pure repackaging — identical control flow, no perf change.

PR #1353, opened 2026-06-18
fix: Reuse shared runtime for sync CAS fetches

bplatz:
## Summary

Fixes a bug where remote CAS leaf fetches could stall when querying an S3-backed ledger. Discovered while running a query against a ~0.6B-flake database — a query whose result set had to span many index leaves, fanning out enough remote leaf range scans to surface the failure.

The sync→async CAS bridge (`run_sync_on_runtime`) used to build and **drop a fresh `current_thread` Tokio runtime on every fetch** in the no-ambient-runtime and current-thread paths. Pooled S3/hyper connections bind their dispatch tasks to the runtime that created them, so dropping that runtime after each fetch killed the dispatch task. The next fetch that reused a pooled connection had nothing driving it and stalled.

## Fix

Replace the per-fetch runtime with a **process-wide, long-lived multi-thread helper runtime** (`OnceLock`), so pooled connections keep their dispatch tasks alive across fetches. This also makes cross-path connection reuse safe (a connection first dispatched on the helper runtime, later reused from the ambient runtime, still has a live task).

Dispatch by runtime context:
- **No ambient runtime** (Rayon fast-path leaf scans — the failure mode here): drive the future directly on the shared runtime.
- **Current-thread / unknown flavor**: move the blocking wait to a plain thread (can't call `block_on` from inside a runtime), but run the async work on the same shared runtime.
- **Multi-thread** (server hot path): unchanged — `block_in_place(handle.block_on)` on the ambient runtime.

## Testing

- Adds a no-runtime fan-out regression guard (`run_sync_on_runtime_no_runtime_fanout_uses_shared_runtime`) that drives 32 concurrent bridged fetches with no ambient runtime and asserts all complete.
- `cargo check` / `cargo test` pass for `fluree-db-binary-index` and `fluree-db-storage-aws`.

### DIFF
--- a/fluree-db-binary-index/src/arena/vector.rs
+++ b/fluree-db-binary-index/src/arena/vector.rs
@@ -63,14 +63,13 @@ fn write_atomic(path: &Path, bytes: &[u8]) -> io::Result<()> {
         Some(dir) => dir.join(format!(".{file_name}.tmp.{pid}.{seq}")),
         None => PathBuf::from(format!(".{file_name}.tmp.{pid}.{seq}")),
     };
-    std::fs::write(&tmp, bytes)?;
-    match std::fs::rename(&tmp, path) {
-        Ok(()) => Ok(()),
-        Err(e) => {
-            let _ = std::fs::remove_file(&tmp);
-            Err(e)
-        }
+    // Clean up the staging file on any failure (write or rename) so a mid-write
+    // error doesn't leave an orphaned temp file behind.
+    let staged = std::fs::write(&tmp, bytes).and_then(|()| std::fs::rename(&tmp, path));
+    if staged.is_err() {
+        let _ = std::fs::remove_file(&tmp);
     }
+    staged
 }
 
 /// Maximum vectors per shard. At 768-dim f32 each shard ≈ 9 MB.
--- a/fluree-db-binary-index/src/arena/vector.rs
+++ b/fluree-db-binary-index/src/arena/vector.rs
@@ -752,10 +752,26 @@ impl LazyVectorArena {
         let cs = Arc::clone(cas);
         let cid = cid.clone();
         let path = source.path.clone();
+        // Bound the fetch with the same `cas_sync_timeout()` the index-leaf and
+        // dict-pack bridges use, so a stalled remote (S3) GET fails fast instead
+        // of blocking the waiting thread indefinitely.
+        let timeout = crate::read::binary_index_store::cas_sync_timeout();
         let bytes = crate::read::binary_index_store::run_sync_on_runtime(async move {
-            cs.get(&cid)
-                .await
-                .map_err(|e| io::Error::other(e.to_string()))
+            let fut = cs.get(&cid);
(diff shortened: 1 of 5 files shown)


fluree-db-binary-index/src/read/binary_index_store.rs
```diff
@@ -79,15 +80,49 @@ pub(crate) fn cas_sync_timeout() -> Option<Duration> {
...
+    std::thread::Builder::new()
+        .name("fluree-cas-helper-bridge".into())
+        .spawn(move || {
+            let result = shared_cas_helper_runtime().and_then(|rt| rt.block_on(fut));
+            let _ = tx.send(result);
+        })
+        .map_err(|e| io::Error::other(format!("failed to spawn CAS helper bridge: {e}")))?;
+    rx.recv()
+        .map_err(|_| io::Error::other("CAS helper bridge thread panicked"))?
+}
```
aaj3f, line 113: It seems most callers of this wrap the call in `tokio::time::timeout` but the vector `cs.get` future in `arena/vector.rs` doesn't have a timeout wrapper, so a stalled S3 GET call would block the thread/waiter indefinitely. It isn't as if that's a regression in this PR, because the old code did that too, but perhaps worth applying `cas_sync_timeout()` uniformly to prevent this

fluree-db-binary-index/src/arena/vector.rs
```diff
@@ -46,9 +46,33 @@ use serde::{Deserialize, Serialize};
...
+    };
+    std::fs::write(&tmp, bytes)?;
+    match std::fs::rename(&tmp, path) {
+        Ok(()) => Ok(()),
+        Err(e) => {
+            let _ = std::fs::remove_file(&tmp);
+            Err(e)
+        }
+    }
+}
```
aaj3f, line 73: The `rename()`'s `Err(e)` arm performs a `remove_file()`, but `std::fs::write(&tmp, bytes)?` could error mid-write and the `?` would only return the error and leave the orphaned corrupt file. It is a temp file so maybe this is trivial though.

aaj3f (approved):
Seems like a good improvement!

fluree-db-binary-index/src/arena/vector.rs
bplatz: Good catch — fixed in 92121589ac74c5ed423114721e5ed358b834fbc3. `write_atomic` now removes the staging temp file on a mid-write error too, not just on rename failure.

fluree-db-binary-index/src/read/binary_index_store.rs
bplatz: Agreed — applied uniformly in 06392cb09b9a333fdc7b2f7f3d852849c8ce031d. The vector shard fetch in `ensure_on_disk` now wraps `cs.get` in `cas_sync_timeout()` like the index-leaf and dict-pack bridges, so a stalled S3 GET fails fast instead of blocking indefinitely.

PR #1354, opened 2026-06-18
fix: harden CLI auth token storage and robustness

bplatz:
Addresses a set of auth findings in `fluree-db-cli` reported against the CLI auth/replication paths.

### Token file permissions (security)
The `.fluree` config dir and token-bearing config files were written with the default umask (typically `0644` → group/world-readable). Now, on Unix, the CLI restricts the config dir to `0700` and the config file to `0600` whenever it writes token material (and at init). No-op on non-Unix platforms, where ACLs are managed differently. Matches the convention used by `gh`, `aws`, `gcloud`, etc. Token storage remains plaintext (consistent with peer CLIs) — this closes the actual exposure (backups, synced dotfiles, multi-user boxes).

### Replication fail-fast on mid-transfer token expiry
The bulk pack transfer uses `HttpRemoteClient`, which holds a static token and does not auto-refresh (intentional: replication is for long-lived operator/service tokens). Previously a token expiry partway through a transfer printed a misleading "falling back to paginated export" and then 401'd again. The CLI now detects the auth failure mid-transfer and fails fast with the `fluree auth login` guidance the docs promise. Applied to both bearer-token flows (`run_pull`, `run_clone`).

### Login robustness
- Warn when the IdP grants fewer scopes than requested, so a silent downgrade doesn't surface later as an unrelated authorization failure.
- Note at login when no refresh token is issued (manual re-login will be needed) — no network probe.
- Warn when a `refresh_token` is configured under a non-`oidc_device` auth type, where it is silently ignored as dead config.

### Docs
- Clarify the two-client refresh layering in the auth contract (`RemoteLedgerClient` auto-refreshes and persists rotated tokens; `HttpRemoteClient` transfer does not).
- Add an end-to-end OIDC login walkthrough and a headless/automation section (steers CI/containers to a service-account JWS token; explains the `FLUREE_AUTH_PORT` / `redirect_port` overrides and why the callback port cannot be OS-assigned).
- Document the new file-permission behavior.

## Notes
- The reported `:0` (OS-assigned) callback-port fallback for PKCE is **not** implemented: it's architecturally infeasible because OIDC requires the redirect URI to be pre-allowlisted at the IdP. Handled via the new automation docs instead.

## Validation
`cargo check`, `cargo clippy -p fluree-db-cli --all-features --all-targets -D warnings`, config unit tests (incl. a new permission-mode test), and `cargo fmt --all` all pass.

### DIFF
--- a/fluree-db-cli/src/commands/sync.rs
+++ b/fluree-db-cli/src/commands/sync.rs
@@ -92,9 +92,15 @@ fn confirm_large_transfer(estimated_bytes: u64) -> bool {
 /// Whether a remote error string indicates an auth/permission failure
 /// (expired or insufficient token). `fluree-db-nameservice-sync` reports
 /// remote failures as strings, so we match on the common server messages.
+///
+/// The HTTP status checks match the canonical reason phrase (e.g.
+/// `401 Unauthorized`) rather than the bare code: the sync client always
+/// renders statuses via reqwest's `StatusCode` Display, and the error
+/// strings interpolate the request URL and response body, so a bare `401`
+/// could false-match a port or ledger name that merely contains those digits.
 fn is_replication_auth_error(err: &str) -> bool {
-    err.contains("401")
-        || err.contains("403")
+    err.contains("401 Unauthorized")
+        || err.contains("403 Forbidden")
         || err.contains("Bearer token required")
         || err.contains("Untrusted issuer")
         || err.contains("Token lacks storage proxy permissions")
--- a/fluree-db-cli/src/commands/sync.rs
+++ b/fluree-db-cli/src/commands/sync.rs
@@ -89,16 +89,20 @@ fn confirm_large_transfer(estimated_bytes: u64) -> bool {
     trimmed.is_empty() || trimmed == "y" || trimmed == "yes"
 }
 
-fn map_sync_auth_error(remote: &str, err: &str) -> Option<CliError> {
-    // `fluree-db-nameservice-sync` reports remote failures as strings; match the common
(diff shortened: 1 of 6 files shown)


fluree-db-cli/src/commands/sync.rs
```diff
@@ -89,16 +89,20 @@ fn confirm_large_transfer(estimated_bytes: u64) -> bool {
...
-fn map_sync_auth_error(remote: &str, err: &str) -> Option<CliError> {
-    // `fluree-db-nameservice-sync` reports remote failures as strings; match the common
-    // permission-related server errors and provide a clearer CLI message.
-    if err.contains("401")
+/// Whether a remote error string indicates an auth/permission failure
+/// (expired or insufficient token). `fluree-db-nameservice-sync` reports
+/// remote failures as strings, so we match on the common server messages.
+fn is_replication_auth_error(err: &str) -> bool {
+    err.contains("401")
         || err.contains("403")
```
aaj3f: I wonder if we want a tighter check here than `err.contains("401") || err.contains("403")`. An error that was not **STATUS** 401 or 403 but had 401 or 403 just in the error message would flag as a permission error and then prompt the "re-login" error message

aaj3f (approved):
These seem like good improvements 👍

fluree-db-cli/src/commands/sync.rs
bplatz: Addressed in a0537f370b843883757f8fcc7b307cf0792780b1

PR #1355, opened 2026-06-18
fix(query): stop corrupting integer-valued doubles read through the overlay

bplatz:
## Summary

An integer-valued double (e.g. `55000.0`) inserted into a `xsd:double`/`xsd:float` predicate that **already has indexed data** was silently read back as a tiny subnormal (`2.71736e-319`). No error, deterministic, and it surfaces in queries and aggregates (`SUM` over the column reads ≈ 0).

## Root cause

The novelty→overlay encoder `value_to_otype_okey` (used when reading novelty over an indexed predicate) optimized integer-valued doubles to `encode_i64`, but paired the key with the **datatype-derived** `OType` (`XSD_DOUBLE`/`XSD_FLOAT`), whose decode kind is `F64`. The reader then ran `decode_f64` over the `encode_i64` bits:

```
decode_f64(encode_i64(55000)) == f64::from_bits(55000) == 2.71736e-319   (ratio 2^-1074)
```

This is the same anti-pattern already guarded at three **encode-side** sites (`resolver.rs`, `import_sink.rs`); the read-side overlay was the one path that still had it.

## Scope / impact

- Triggers only when the predicate has a **persisted index**; novelty-only ledgers encode correctly.
- The persisted commits were always correct — corruption was purely on the read path. **Affected data is recoverable with the fixed reader; no re-ingestion required.**
- Non-integer doubles (`55000.5`) and integer types were never affected.

## Changes

- `binary_scan.rs` — `value_to_otype_okey` always encodes finite doubles with `encode_f64`.
- `dict_overlay.rs` — same guard applied to the unused twin `value_to_obj_pair` (defensive).
- `it_decimal_exactness.rs` — regression test: index a double, insert an integer-valued double, assert exact round-trip. Verified it fails without the fix with the exact `2.71736e-319` signature.

## Testing

- New regression test passes; fails when the fix is reverted.
- `fluree-db-query` unit (1078) + decimal exactness (14), batched-overlay (4), aggregates (16), SPARQL (136) all green. `fmt` + `clippy` clean.

### DIFF
--- a/fluree-db-api/tests/it_decimal_exactness.rs
+++ b/fluree-db-api/tests/it_decimal_exactness.rs
@@ -56,6 +56,20 @@ fn binding_values(sparql_json: &JsonValue, var: &str) -> Vec<String> {
         .collect()
 }
 
+fn binding_datatypes(sparql_json: &JsonValue, var: &str) -> Vec<String> {
+    sparql_json["results"]["bindings"]
+        .as_array()
+        .expect("bindings array")
+        .iter()
+        .map(|b| {
+            b[var]["datatype"]
+                .as_str()
+                .expect("binding datatype string")
+                .to_string()
+        })
+        .collect()
+}
+
 fn memory_fluree() -> MemoryFluree {
     assert_index_defaults();
     FlureeBuilder::memory().build_memory()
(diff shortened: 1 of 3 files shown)


fluree-db-api/tests/it_decimal_exactness.rs
```diff
@@ -889,3 +889,77 @@ async fn sparql_delete_data_decimal_retracts_exactly() {
...
+            assert_eq!(values.len(), 1, "expected exactly one row, got {values:?}");
+            let got: f64 = values[0].parse().expect("double result");
+            assert_eq!(
+                got, 55000.0,
+                "integer-valued double over an indexed predicate must round-trip \
+                 exactly (was corrupted to a subnormal), got {got:e}"
+            );
+        })
+        .await;
+}
```
aaj3f, line 1008: This test is good for the bug that prompted this PR. Just small additional suggestions:
1. Also assert the **datatype** of the returned binding is `xsd:double` (not silently downgraded to `xsd:integer` or `xsd:long`), to lock in that uniform-f64 encoding doesn't change the reported type.
2. Add a negative/boundary companion value (e.g. `-55000.0` and a large integral double like `9.007199254740992e15`) in the same indexed predicate to cover the sign-flip branch of `encode_f64`/`decode_f64` and the i64-range edge. Not required for the fix, but cheap insurance.

fluree-db-query/src/dict_overlay.rs
aaj3f: Given that `d.is_finite()` will guard against conditions like `NaN`/`Inf`, the `Err(_)` arm is effectively unreachable. It may just be worth a comment if nothing else to make it clear that `Ok((ObjKind::NULL, ObjKey::from_u64(0)))` is never reachable/returnable for a truly finite `d`

aaj3f (approved):
Good catch on this 👍 

It seems after this PR's work, `DictOverlay::value_to_obj_pair` is unused anywhere but is not tracking as dead code because it's `pub`. If we want to keep it just in case, perhaps we at least downgrade to `pub(crate)` with `#[cfg(test)]` coverage or with some comment like `// retained for parity with value_to_otype_okey`

Also, I was also just reviewing #1335 but that PR sustains the bugs that you've fixed in this PR. It's probably worth making sure that this PR lands first and (if you make the test additions I suggested in the review comments below), then in #1335 you can (a) have test-guards that will keep the issue/bug from persisting in that PR and also (b) you can have some test-driven requirements for the rework you'll want to do in that branch's `it_decimal_exactness.rs` code.

bplatz:
@aaj3f addressed in 15ea8d945e807f0965f312c9cbdec11ad86ff7c2

PR #1356, opened 2026-06-19
perf(query): query-planner improvements (hash-joins, source seeding, EXISTS)

bplatz:
These query-planner and execution improvements were identified while building out RDF 1.2 edge-annotation support, but they are entirely independent of RDF 1.2: they apply to ordinary SPARQL and JSON-LD queries and carry no annotation- or edge-specific code. Pulling them out so they can land and be reviewed on their own, ahead of the larger RDF 1.2 work.

## What's included

- **object→subject hash join after a DISTINCT producer** — the object→subject hash join now fires when its left input is an uncorrelated DISTINCT sub-SELECT producer, instead of falling back to a scattered object-driven scan.
- **connectedness-aware source seeding** — when a pipeline already has a connected alternative, a disconnected `rdf:type <const>` class anchor is demoted from the seed pool so it can't flip the join order into a scattered reverse object-drive.
- **subquery producer placement** — an uncorrelated sub-SELECT producer is placed before the triples that consume its output, so a cheap consumer scan can't be hoisted ahead of the producer and silently collapse its bindings.
- **subject-driving batched hash-left-join for correlated OPTIONAL** — a correlated multi-pattern OPTIONAL is evaluated as a single batched hash left-join keyed on the correlation vars, replacing the per-driving-row OPTIONAL subplan rebuild.
- **evaluate EXISTS as a projected/BIND expression** — `EXISTS { ... }` can be evaluated in projection / `BIND` position, not only as a pattern-level filter.

## Testing

- New `it_query_optional_hashjoin.rs` (JSON-LD) covers the correlated multi-pattern OPTIONAL left-join, including the no-match `count = 0` group.
- Existing planner-exercising suites pass: `it_query_negation`, `it_query_aggregates`, `it_query_jsonld`.

### DIFF
--- a/fluree-db-api/tests/it_policy_optional_hashjoin.rs
+++ b/fluree-db-api/tests/it_policy_optional_hashjoin.rs
@@ -1,0 +1,144 @@
+//! View-policy enforcement on the batched correlated-OPTIONAL hash-join.
+//!
+//! `PlanTreeOptionalBuilder::build_batch` seeds and executes the OPTIONAL inner
+//! subplan once per required batch via `build_where_operators_seeded`, reusing
+//! the normal WHERE operators and the SAME `ExecutionContext`. So the inner's
+//! `BinaryScanOperator`s must see the active policy enforcer and filter the
+//! seeded scan exactly as the per-row path would.
+//!
+//! This guards against the batched fast path becoming a view-policy blind spot:
+//! a forbidden optional-side row must not leak into the count, and a row whose
+//! only matches are all policy-hidden must collapse to the left-join zero.
+
+mod support;
+
+use fluree_db_api::FlureeBuilder;
+use serde_json::{json, Value as JsonValue};
+use support::{assert_index_defaults, genesis_ledger, normalize_rows};
+
+fn ctx() -> JsonValue {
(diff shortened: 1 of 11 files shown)


aaj3f (approved):
This is nice and the improvements make sense / are sound.

If you wouldn't mind, this would be an excellent candidate PR for additions to `regression-budget.json` / added `[[bench]]` config (see `docs/contributing/benches.md`). I'd suggest adding an IC5-shaped OPTIONAL bench and a BSBM-BI F2 bench so this PR's claims are guarded against future drift, and so the deliberate trade-offs documented at `optional.rs:1150` can't silently regress a different shape.

bplatz:
Added the BSBM-BI F2 bowtie bench (`query_hot_bsbm_bi`) on `test/query-planner-bi-bench` — guards the seed tie-break drive direction. One caveat: it builds/smoke-runs in `bench-gate` per PR, but the actual budget comparison only kicks in once `bench-nightly` exists. Deferring the IC5 OPTIONAL bench (needs a new social-graph generator; the path is already correctness-covered by `it_query_optional_hashjoin`).

PR #1357, opened 2026-06-19
feat: RDF 1.2 / SPARQL 1.2 edge annotations + GQL query operators in JSON-LD

bplatz:
RDF 1.2 / SPARQL 1.2 edge annotations, plus the shared GQL query operators they ship alongside, exposed in JSON-LD query. Stacked on the base query-planner PR; the Cypher front-end (parser, lowering, cypher-json) lands in a follow-up PR.

## Edge annotations (RDF 1.2 / SPARQL 1.2)

Attach properties to a relationship — the `(subject, predicate, object)` edge — without hand-modeling an intermediate node. Reads correctly as property-graph "edge properties" or as SPARQL "annotations on a quoted triple".

- Annotation-arena storage, indexer build/seal + GC, attachment novelty
- Core edge / annotation-index model + per-graph annotation stats
- SPARQL 1.2 quoted-triple annotation surface (`{| ... |}`, `~`, `<<( ... )>>`, `VERSION "1.2"`) on query and UPDATE
- JSON-LD `@annotation` / `@edge` write surface (incl. literal-valued and named-graph edges)
- Planner edge-annotation expansion + read-side firewall for system predicates
- CSV (neo4j-admin header convention) import to RDF 1.2 annotations

## GQL query operators (exposed in JSON-LD)

Generic query constructs that came with the property-graph work, now reachable from JSON-LD (not just the Cypher front-end):

- `unwind` + `range` / `list` — expand a computed list into rows (e.g. dense / gap-filled series)
- `collect` / `collect-distinct` — gather group values into a list; inverse of `unwind`
- List functions — `size`, `head`, `last`, `tail`, `reverse`, `nth`
- `shortestPath` / `allShortestPaths` + `nodes` / `pathPairs`
- Alternation-transitive property paths `(a|b)*` via `@path`

## Docs

- Edge-annotation concept, cookbook, vocabulary, and storage-internals design docs
- A new Query Patterns cookbook with realistic recipes (dense series, collect/unwind round-trip, shortest path)
- JSON-LD query reference sections for every new operator

Each operator and edge-annotation surface has integration test coverage.


fluree-db-transact/src/parse/jsonld.rs
```diff
@@ -95,13 +95,67 @@ impl GraphIdAssigner {
...
+    let top_ctx = super::edge_annotations::top_level_context(&lowered)?;
+    super::edge_annotations::run_user_authored_reifies_firewall(&lowered, &top_ctx)?;
+    if matches!(txn_type, TxnType::Update) {
+        super::edge_annotations::lower_delete_annotation_blocks(&mut lowered)?;
+    }
+    super::edge_annotations::lower_edge_annotations_after_firewall(
+        &mut lowered,
+        &top_ctx,
+        lpg_mode,
+    )?;
```
aaj3f: In your sibling changes to import.rs (i.e. line 962), the same work is gated by `jsonld.contains("@annotation") || …("@edge") || …("@reifies")`. But here, non-annotated transact ops (possibly large ones) pay for the large payload clone and the two tree walks.

fluree-db-transact/src/import.rs
```diff
@@ -949,9 +949,25 @@ mod inner {
...
+        // (lpg_mode=false): a non-empty `@annotation` lowers to a reifier
+        // bundle; the CSV/LPG loader never emits an empty `@annotation: {}`.
+        if jsonld.contains("@annotation") || jsonld.contains("@edge") || jsonld.contains("@reifies")
+        {
+            let top_ctx = crate::parse::edge_annotations::top_level_context(&doc)?;
+            crate::parse::edge_annotations::run_user_authored_reifies_firewall(&doc, &top_ctx)?;
+            crate::parse::edge_annotations::lower_edge_annotations_after_firewall(
+                &mut doc, &top_ctx, false,
+            )?;
+        }
```
aaj3f: Do we need to worry about the scenario where, for example, `https://ns.flur.ee/db#reifiesObject`, is used instead of the keyword? Or should the firewall always run on the expanded form?

fluree-db-api/src/csv_import.rs
```diff
@@ -0,0 +1,702 @@
...
+                    .map(Value::Number)
+                    .ok_or_else(|| bad("double"))?
+            }
+            CsvType::Boolean => Value::Bool(match raw.trim().to_ascii_lowercase().as_str() {
+                "true" | "1" | "yes" => true,
+                "false" | "0" | "no" => false,
+                _ => return Err(bad("boolean")),
+            }),
+            CsvType::Date => typed_literal(raw.trim(), "date"),
+            CsvType::DateTime => typed_literal(raw.trim(), "dateTime"),
```
aaj3f: `:int`, `:double`, and `:boolean` return `BadValue` on malformed inputs. Do we need parse-or-`BadValue` handling on these two as well?

fluree-db-api/src/csv_import.rs
aaj3f, line 1: There are a few places in here (lines 319, `mint_iri()` at 282) where there's a direct xform into `format!("{base_iri}{name}")`. A `:TYPE` cell containing `http://evil/p`, spaces,
`#`, or `/` produces malformed or namespace-escaping IRIs (e.g. predicate
`http://example.org/http://evil/p`; an id value `../../x` escapes the id space, etc.)

aaj3f (approved):
This is really exciting, although admittedly it's a lot of code to review. Almost everywhere I looked it was well-gated to not affect normal paths, though, so I think it's a lot lower risk than I initially worried it would be. A few remarks below, but otherwise things look correct & good

fluree-db-transact/src/parse/jsonld.rs
bplatz: Fixed in `73845c4e555b41838906edd8e7ef0b1fc43328db`. The firewall now runs read-only on the original document unconditionally (it is the security guard), and only the payload clone + lowering walks are gated — on a new `document_has_annotation_keys` scan — so a non-annotated transaction skips them.

fluree-db-transact/src/import.rs
bplatz: Fixed in `73845c4e555b41838906edd8e7ef0b1fc43328db`. The firewall gate is now deliberately broader than the lowering gate — it also fires on the bare `reifies` token, so a fully-expanded `https://ns.flur.ee/db#reifies*` IRI (which carries no `@`-keyword) can no longer bypass it on the bulk-import path. Regression test added.

fluree-db-api/src/csv_import.rs
bplatz: Fixed in `045301625b9fd9c8feab4a3b1aad262bcc4e55f1`. `:date`/`:datetime` cells now validate against the same `temporal` parsers the downstream pipeline uses, returning `BadValue` on malformed input — matching `:int`/`:double`/`:boolean`.

fluree-db-api/src/csv_import.rs
bplatz: Fixed in `da902f131d2fb3b6e9fce04204a2ee2b749dfee3`. Cell-derived IRI segments (id / `:TYPE` / `:LABEL` values) are now percent-encoded — whitespace/controls, the RFC 3987-forbidden ASCII, and the structural delimiters (`/ # ? [ ] @`) — so a value cannot break IRI syntax or escape `base_iri` (e.g. `http://evil/p`, `../../x`). Schema-declared column names are unchanged.

PR #1359, opened 2026-06-20
Fix complex aggregates in SELECT

zonotope:
A SPARQL query that wraps multiple aggregates inside an arithmetic expression — for example, `SELECT ?part ((MAX(?u) - MIN(?u)) AS ?spread) ... GROUP BY ?part LIMIT 5` — failed with `SparqlLower(NotImplemented { feature: "Aggregate function Max", ... })`. The equivalent JSON-LD form `(as (- (max ?u) (min ?u)) ?spread)` failed with `Parse(InvalidFilter("Unknown function: max"))`. Both languages already supported bare `(MAX(?u) AS ?hi)` columns; only the *compound-aggregate* shape was broken. This branch fixes both, end to end.

## Reported query

```sparql
PREFIX sup: <http://Magna/SupplyChain#>
SELECT ?part ((MAX(?u) - MIN(?u)) AS ?spread)
WHERE { ?r a sup:ReceiptLineItem ; sup:forEngcPart ?part ; sup:receiptUnitPrice ?u }
GROUP BY ?part
LIMIT 5
```

## Root cause — SPARQL side

`fluree-db-sparql/src/lower/` had three SELECT helpers that only recognised `SelectVariable::Expr { expr: Expression::Aggregate, .. }` — i.e. *bare* aggregates. A compound `(MAX(?u) - MIN(?u)) AS ?spread` fell into `lower_select_expression_binds`, which called `lower_expression` without an `aggregate_aliases` map in scope; the recursive lowerer then rejected the nested `MAX` with `NotImplemented`. HAVING and ORDER BY already had the correct pattern (`collect_inline_aggregates` + a temporarily-set `self.aggregate_aliases`); SELECT did not.

## Fix — SPARQL side

Extend SELECT lowering to mirror the HAVING / ORDER BY path that already worked:

- `collect_aggregate_alias_names` now uses `expr_contains_aggregate` instead of `matches!(_, Aggregate)` so compound-aggregate SELECT aliases also count as post-aggregation aliases.
- `lower_select_expression_binds` skips any SELECT expression that contains an aggregate (compound or bare) — `lower_solution_modifiers` handles them later.
- `lower_solution_modifiers` gains a new pass `collect_compound_aggregate_select_items` that hoists each compound expression's inner aggregates via `collect_inline_aggregates`, then lowers the outer expression to a `(VarId, Expression)` post-aggregation bind. The hoist shares the same `aggregate_aliases` map as HAVING and ORDER BY, so duplicate aggregates dedupe across all three.
- `LoweredModifiers` gains a `select_post_binds: Vec<(VarId, Expression)>` field; the two callers (`lower::lower` and `lower_subselect`) append it onto `select_binds.post` before `Grouping::assemble`.

## Root cause — JSON-LD side

`fluree-db-query/src/parse/mod.rs::parse_select_string` only recognised an S-expression as an aggregate when the aggregate name was the *immediate* head. For `(as (- (max ?u) (min ?u)) ?spread)` the head is `-`, so the call fell into the scalar-expression branch (`filter_sexpr::expr_from_sexpr_token`). That path lowered `max` and `min` as generic `UnresolvedExpression::Call`s, and the function-call lowerer subsequently rejected them with "Unknown function: max".

## Fix — JSON-LD side

A new `hoist_inline_aggregates` helper walks the SELECT-clause S-expression token tree *before* it is converted to an `UnresolvedExpression`. Each nested aggregate call (`(max ?u)`, `(min ?u)`, …) is replaced in place with a synthetic atom `?__inline_agg_N`, and a corresponding `UnresolvedAggregateSpec` is pushed onto the running `aggregates` vector. Identical `(fn, input)` pairs within a single column dedupe to one output var via a small `seen: HashMap<String, Arc<str>>`. The existing `lower_select_expr_bind` then sees the synthetic vars in `aggregate_output_vars` and routes the column as a post-aggregation bind — the same channel `subquery_post_aggregation_bind_jsonld` already exercises.

## Tests

- `it_query_sparql::sparql_arithmetic_over_min_max_in_select_repro` — the user's exact query verbatim; was the failing repro, now passes.
- `it_query_sparql::sparql_bare_min_max_in_select_works` — control test for the bare-aggregate shape, guards against the SPARQL-side fix accidentally breaking the simpler path.
- `it_query_aggregates::aggregates_arithmetic_over_min_max_jsonld` — the JSON-LD equivalent `(as (- (max ?u) (min ?u)) ?spread)`, computing the compound aggregate directly (no `?hi` / `?lo` helper aliases).

### DIFF
--- a/fluree-db-api/tests/it_query_aggregates.rs
+++ b/fluree-db-api/tests/it_query_aggregates.rs
@@ -590,3 +590,97 @@ async fn select_expr_if_branches_on_age() {
         ]))
     );
 }
+
+async fn seed_receipt_line_items_jsonld(
+    fluree: &MemoryFluree,
+    ledger_id: &str,
+) -> MemoryLedger {
+    let ledger0 = genesis_ledger(fluree, ledger_id);
+
+    let insert = json!({
+        "@context": {
+            "sup": "http://Magna/SupplyChain#",
+            "xsd": "http://www.w3.org/2001/XMLSchema#"
+        },
+        "@graph": [
+            {
+                "@id": "sup:r1",
+                "@type": "sup:ReceiptLineItem",
+                "sup:forEngcPart": {"@id": "sup:partA"},
+                "sup:receiptUnitPrice": 10
+            },
+            {
+                "@id": "sup:r2",
+                "@type": "sup:ReceiptLineItem",
+                "sup:forEngcPart": {"@id": "sup:partA"},
+                "sup:receiptUnitPrice": 14
+            },
+            {
+                "@id": "sup:r3",
+                "@type": "sup:ReceiptLineItem",
+                "sup:forEngcPart": {"@id": "sup:partA"},
+                "sup:receiptUnitPrice": 12
+            },
+            {
+                "@id": "sup:r4",
+                "@type": "sup:ReceiptLineItem",
+                "sup:forEngcPart": {"@id": "sup:partB"},
+                "sup:receiptUnitPrice": 5
+            },
+            {
+                "@id": "sup:r5",
+                "@type": "sup:ReceiptLineItem",
+                "sup:forEngcPart": {"@id": "sup:partB"},
+                "sup:receiptUnitPrice": 9
+            }
+        ]
+    });
+
+    fluree
+        .insert(ledger0, &insert)
+        .await
+        .expect("seed receipt line items")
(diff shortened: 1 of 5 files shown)


bplatz (approved):
Looks Great!

PR #1361, opened 2026-06-23
feat: openCypher 9 front-end (parser, read/write, paths, relationships)

bplatz:
## Summary

Adds an **openCypher 9 front-end** on top of the RDF 1.2 edge-annotation work in
the base branch. Cypher parses and lowers to the **same query IR and transaction
pipeline** as JSON-LD and SPARQL — the planner, executor, and result formatter
are shared — so this is a new surface, not a new engine. Cypher relationships
map onto the base branch's edge-annotation primitive, so property-graph edges
and RDF quoted-triple annotations are the same data read from two angles.

## What's included

**Read path**
- Clauses: `MATCH` / `OPTIONAL MATCH` / `WITH` / `UNWIND` / `RETURN`, `UNION` /
  `UNION ALL`.
- `CALL { … }` subqueries: explicit imports `(a, b)`, `(*)`, uncorrelated
  broadcast, inner `UNION`, nesting, and strict scope/shadowing validation with
  correlated-aggregate soundness.
- Paths: `shortestPath` / `allShortestPaths`; bounded variable-length paths with
  trail enumeration and relationship-uniqueness; binding a path (`MATCH p = …`)
  or a relationship list (`-[r:T*1..n]->`).
- Relationship/path values: `relationships(p)`, `nodes(p)`, `type(r)`,
  `startNode(r)` / `endNode(r)`, `properties(r)`, edge property access.
- Expressions: arithmetic (incl. `^`), comparison/boolean, string predicates,
  `IN`, `CASE`, list/map literals and indexing, list comprehensions / `reduce` /
  quantifiers, pattern comprehensions, `EXISTS`, parameters.
- A scalar-function pass (string, math, casts, `id`/`elementId`), aggregates
  with implicit grouping + `WITH`-as-HAVING, and `collect()` carried through a
  `WITH` boundary.
- Map values and map projection (`n{.name, .*}`).

**Write path**
- `CREATE` / `SET` / `REMOVE` / `DELETE` / `DETACH DELETE`; node and relationship
  `MERGE` (with `ON CREATE SET`); `WITH` before a write.

**Surfaces**
- HTTP `application/cypher` query route and CLI `--format cypher-json`
  (Neo4j-compatible output, with `jsonld` opt-in), both policy-enforced like the
  JSON-LD / SPARQL paths.

**Docs**
- New concept doc, cookbook, and a tracked openCypher support matrix
  (`docs/reference/cypher-support-matrix.md`) marking each feature
  supported / divergent-by-design / deferred.

## Model divergences (intentional)

Fluree enables a unified layer accessible by both SPARQL and Cypher: nodes are
IRIs (labels are `rdf:type`), relationships are edge annotations,
`id()`/`elementId()` returns the IRI string, and unbounded variable-length
traversal is reachability (bounded enumerates trails).

## Testing

Covered by the Cypher parser/lowering unit suites and the
`it_query_cypher` / `cypher_http_integration` integration suites (read, write,
paths, relationships, `CALL`, null/type semantics), plus the cross-surface
edge-annotation tests shared with the base branch.


aaj3f (approved):
Leaving an approve review, but I found some non-cypher-path (i.e. jsonld / sparql) performance and policy-security regressions, amongst other things, that prompted me to then ask Claude to do a full review. I mentioned this in-person, but I've passed that full audit to you in Slack (https://fluree-internal.slack.com/files/UKWLAAHBQ/F0BCYDH697C/pr-1361-review.md) for you to investigate and address as you see fit

(Oh, should add--super excited for this!)

PR #1363, opened 2026-06-24
docs(readme): add "For AI agents" section linking llms.txt

Jackamus29:
Point agents at the published llms.txt / llms-full.txt artifacts from the README — the entry point for the "user tells agent to use Fluree → agent finds the repo" path.


bplatz (approved):
👍

PR #1364, opened 2026-06-24
perf: Iceberg/R2RML graph-source query optimizations

bplatz:
## Summary

A set of query-execution optimizations for Iceberg/R2RML graph sources that let common analytical query patterns scale efficiently to much larger Iceberg tables. The work removes per-row and per-pattern overhead across the scan → join → materialize path and parallelizes file reads and materialization. Validated on TPC-H up to SF1; results verified exact.

## Query classes optimized

- **Same-subject star patterns** (many predicates on one subject) — now scale linearly with row count.
- **Cross-source joins** between two graph sources — keyed hash join, O(N + M).
- **Filtered scans** — `FILTER` predicates pushed to the Iceberg scan to skip non-matching data files.
- **Group-by / aggregations over large scans** — parallel RDF materialization.
- **Wide multi-file scans** — bounded parallel Parquet reads.

## Representative results

TPC-H, 6M-row `lineitem` on Direct-mode Iceberg/S3, SF1:

| Query class | Result |
|---|---|
| same-subject star | linear scaling; 2-predicate star over 60k rows in ~0.4 s |
| cross-source join | 6M ⋈ 1.5M in ~7.6 s |
| filtered scan (one-year date filter) | ~5× fewer files read (Q6 ~7 s, pruned) |
| group-by aggregation | TPC-H Q1 (8 aggregates) in ~26 s |
| join + filter + aggregate | TPC-H Q3-style in ~21 s |

## What's in this PR

- **Same-subject star collapse.** Triple patterns sharing a subject are grouped into a single scan that materializes all predicates and emits the row directly, instead of one scan per predicate joined row-by-row.
- **Cross-source hash join.** An R2RML graph block now runs once over the whole parent batch (new multi-row seed), and the scan operator hash-joins on the shared variables, instead of re-scanning the second source per parent row.
- **FILTER pushdown / file pruning.** Date/integer/boolean `?col <op> const` comparisons are pushed to the Iceberg scan for min/max file skipping (bypassing the metadata-keyed file-selection cache, since the pruned set depends on the filter). Conservative — the in-engine `FILTER` still runs, so pruning only skips files, never drops rows; includes guards for int32 overflow and multi-column predicates.
- **Parallel materialization + datatype hoist.** Datatype SIDs are resolved once per scan instead of per literal cell; RDF materialization runs across the rayon pool.
- **Bounded parallel Parquet reads.** Files read as bounded "file → batches" workers (concurrency = `min(cpus, files, 8)`, overridable via `FLUREE_ICEBERG_SCAN_CONCURRENCY`), structured as the first slice of a streaming scan. Largest win on remote S3 and multi-table joins.
- **Two correctness fixes** surfaced by the TPC-H suite: arithmetic over materialized decimals, and an Iceberg date materialization off-by-one.
- **Opt-in `mimalloc` global allocator** for server/cli — see "Open decision".

## Correctness & testing

- Queries cross-checked against reference TPC-H results at matching scale (counts, decimal sums, date filters, join cardinality) — verified exact at SF0.001 and SF1.
- Existing R2RML engine/e2e + RefObjectMap tests, query-crate units, and r2rml-crate tests are green; clippy clean on touched crates.
- Adds a `TPCH_BENCH`-gated, `#[ignore]`d benchmark harness (not run in CI) and a `smoke_minio` example.

## Blast radius

Scoped to the Iceberg/R2RML path — native ledger queries take the same code paths and produce the same results. The only shared executable file is the graph operator, where the new fast path is gated (dataset-mode and non-R2RML graphs skip it). The one process-wide change, `mimalloc`, is opt-in and off by default.

## Open decision: mimalloc

The parallel materialization is allocation-heavy; with the system allocator the realized speedup is roughly half (Q1 ~62 s vs ~31 s with mimalloc). This PR adds an opt-in `mimalloc` Cargo feature to server and cli (off by default; documented in `operations/configuration.md`) but does not enable it. Decision for reviewers: whether to enable it in release packaging after a soak, and whether to pair it with bounded query parallelism. Tradeoffs (process-wide allocator change, possibly higher retained RSS, native-dep / cross-build check) are documented.

- Streaming scan (consume batches as workers complete) to bound memory for very large scans.
- A shared persisted disk cache so repeated server/Lambda queries don't re-pull from S3.
- Row-group-level skipping (pushdown below file pruning).


fluree-db-query/src/r2rml/operator.rs
```diff
@@ -497,235 +786,101 @@ impl Operator for R2rmlScanOperator {
...
+                        let mut ok = true;
+                        for &jv in &join_vars {
+                            let pos = child_schema.iter().position(|&v| v == jv).unwrap();
+                            let b = &child_batch.column_by_idx(pos).unwrap()[row_idx];
+                            if !b.is_bound() || b.is_poisoned() {
+                                ok = false;
+                                break;
+                            }
+                            key.push(binding_to_group_key_normalized(b, store, gv));
+                        }
```
aaj3f, line 893: I think the `if !b.is_bound() || b.is_poisoned() { ok = false; break }` will treat an **unbound** child join-var the same as a non-match and skip the whole child row. So for an upstream `OPTIONAL` or `UNION` w/ partial bindings for a variable, those rows would be dropped instead of joined

Unless I'm wrong about this path and those characteristics would never path to this code

fluree-db-query/src/r2rml/operator.rs
```diff
@@ -113,61 +121,315 @@ impl R2rmlScanOperator {
...
+        out
+    }
+
+    /// Materialize the object term for one POM at a table row, resolving a
+    /// RefObjectMap through the pre-built parent lookup. Returns None when the
+    /// value is null or the foreign key is orphaned.
+    /// Materialize all produced rows of a TriplesMap (independent of the child
+    /// stream) so the result can be hash-joined against the child. Datatype Sids
+    /// are resolved once into a `LiteralEncoder` (not per cell) and the data
+    /// files are materialized in parallel on the rayon pool.
```
aaj3f: I think maybe these doc lines got duplicated?

fluree-db-query/src/r2rml/operator.rs
```diff
@@ -113,61 +121,315 @@ impl R2rmlScanOperator {
...
+        parent_lookups: &HashMap<(String, Vec<String>), ParentLookup>,
+        ctx: &ExecutionContext<'_>,
+    ) -> Result<Vec<Vec<(VarId, Binding)>>> {
+        use rayon::prelude::*;
+        let encoder = LiteralEncoder::build(triples_map, ctx.active_snapshot);
+        let pattern = &self.pattern;
+        let per_batch: Vec<Vec<Vec<(VarId, Binding)>>> = batches
+            .par_iter()
+            .map(|batch| materialize_batch(pattern, triples_map, batch, parent_lookups, &encoder))
+            .collect::<Result<Vec<_>>>()?;
```
aaj3f, line 201: Not an issue for single-query scenarios but under concurrent load, the fact that this is a synchronous/thread-blocking wait on the materialization may be a reason to refactor this (e.g. to avoid the expense of removing a worker from the pool during the wait). Either may just want to flag this w/ a comment that this blocks the worker thread and can be improved in follow-ups for better bounded-query-parallelism, or we could wrap it now in `tokio::task::spawn_blocking` or use rayon scope-to-oneshot bridge if `materialize_batch` needs cross-worker coordination

aaj3f (approved):
Looks great! I see your point about wanting follow-ups for disk caching as some of the materialization & batch draining is likely to hold a LOT in memory

fluree-db-query/src/r2rml/operator.rs
bplatz: Addressed in 40d27e325aa9b3ea086bbd3fa48ac275a454f484

fluree-db-query/src/r2rml/operator.rs
bplatz: Addressed in 40d27e325aa9b3ea086bbd3fa48ac275a454f484

fluree-db-query/src/r2rml/operator.rs
bplatz: Good call — leaving this as-is for now; it blocks the worker thread but is correct and only matters under concurrent load. Will address in a follow-up PR (alongside bounded query parallelism), coming soon.

PR #1365, opened 2026-06-24
test(api): consolidate integration test binaries into domain groups

bplatz:
## Why

In `fluree-db-api`, every `tests/*.rs` is its own test **binary** that links the entire crate dependency graph. With ~160 integration files that meant ~160 full link steps per `cargo test`/`nextest` build — the dominant cost of building the suite — plus the shared `support` harness recompiled ~160× and ~27 GB of debug artifacts.

## What

Group the integration files into **12 domain binaries** (`query`, `query_sparql`, `query_reason`, `query_history`, `policy`, `import`, `index`, `transact`, `ledger`, `graphsource`, `reasoning`, `misc`) via `autotests = false` + explicit `[[test]]` targets. Each case file is included as a `#[path]` module and refers to the shared harness as `crate::support` (compiled **once per binary**). Files become modules, so item names never collide across files.

**Kept standalone** (their own `[[test]]` binary):
- Feature-gated suites — `credential`, `iceberg`, `aws-testcontainers`, `vector` (with `required-features`).
- Tests that mutate **process-global `FLUREE_*` env vars** — so they stay process-isolated under bare `cargo test`, exactly as before. (Under `nextest` every test is already its own process.)

**168 → 22 test binaries.**

## Results (local; CI gains more)

| Metric | Before (168) | After (22) |
|---|---|---|
| Clean test build (`--all-features --tests --no-run`, deps cached) | 68 s | **27 s (~2.5×)** |
| Debug build artifacts | ~27 GB | a few GB (~7–10× smaller) |

Local numbers **understate CI**: CI runners are 2-core and use the `bfd` linker, where per-binary *link* cost dominates — eliminating 146 link steps saves proportionally more, and the smaller artifacts shrink the rust-cache.

## Validation

- Full suite under `nextest`: **2368 passed, 0 failed** (12 skipped = Docker/LocalStack not present locally).
- Biggest group `grp_misc` under bare `cargo test` (one process): **233 passed, 0 failed** — no intra-process races.
- `cargo fmt --all --check` clean; **0 compiler warnings**.

## Adding a test

Create `tests/it_<name>.rs`, then add `#[path = "it_<name>.rs"] mod it_<name>;` to the appropriate `grp_*.rs`. No new `[[test]]` entry unless you start a new group. See `docs/contributing/tests.md`.

- Optionally split `grp_misc` (32 files) / `grp_query` (21) further for finer edit-locality vs. link-count tradeoff.
- The same pattern could be applied to `fluree-db-server` (16 binaries).

### DIFF
--- a/docs/contributing/sparql-compliance.md
+++ b/docs/contributing/sparql-compliance.md
@@ -438,9 +438,11 @@ SPARQL and JSON-LD queries in Fluree compile to the **same intermediate represen
 # SPARQL W3C tests (from testsuite-sparql/)
 make test-eval-cat CAT=<category>
 
-# JSON-LD query tests (from workspace root)
-cargo test -p fluree-db-api --test it_query
-cargo test -p fluree-db-api --test it_query_analytical
+# JSON-LD query tests (from workspace root) — integration tests are grouped
+# into grp_* binaries; run a whole group or filter to one file's module.
+cargo test -p fluree-db-api --test grp_query
+cargo test -p fluree-db-api --test grp_query_sparql
+# e.g. just one file's tests: cargo test -p fluree-db-api --test grp_query it_query_jsonld
 ```
 
 ## Architecture Overview
--- a/docs/contributing/tests.md
+++ b/docs/contributing/tests.md
@@ -50,6 +50,47 @@ fn test_query_workflow() {
 }
 ```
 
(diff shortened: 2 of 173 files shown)


zonotope (approved):
🚗

PR #1366, opened 2026-06-24
ci: mold linker + move bench-gate to nightly

bplatz:
Stacked on #1365 (test-binary consolidation) — base is `test/integration-consolidation`, so this diff is just the CI changes.

## Changes
1. **Linker: bfd → mold.** The v4 baseline forced `-fuse-ld=bfd` (the slowest GNU linker) with no recorded rationale. Switch to mold, installed per compiling job via `rui314/setup-mold`. Validated green across clippy/test/testsuite-sparql — the unexplained bfd was not hiding a mold incompatibility.
2. **bench-gate → nightly.** bench-gate was wired into `ci.yml` (commit cb4916968) and ran on every PR, where its release-mode build of the whole workspace's benches is a ~27–30 min long pole. It was meant to be nightly-only. Moved to a dedicated `.github/workflows/bench.yml` (daily cron + `workflow_dispatch`). Per-PR coverage is preserved: `clippy --all --all-features --all-targets` still compiles every bench on every PR (all bench crates are workspace members), so compile/API breakage is still caught; the nightly adds the runtime smoke + the `[[bench]]` ↔ `regression-budget.json` reconcile.

## Also included
- The import-deadlock fix (9068039dc) is cherry-picked here because base/`main` lack it; without it the per-PR `test` job intermittently times out on the load-dependent `remote_import_matches_local_flake_count` (slow/cold CI runs trigger it). De-dupes on merge once the fix reaches main.

## Measured impact (per-PR CI wall-clock)
~27–30 min → **~7.5 min** combined with #1365. bench→nightly removes the long pole; mold trims link time (modest on warm, larger on cold builds).

## Notes
- `bench.yml`'s schedule/`workflow_dispatch` only activate once it lands on `main` (GitHub limitation). YAML validated; the job is a verbatim move of the working bench-gate.
- `main` has no required status checks, so dropping bench-gate from the per-PR path blocks nothing.
- CI triggers on PRs targeting `main`; as a stacked PR onto `test/integration-consolidation` it won't auto-run here — CI re-runs once rebased onto main after #1365 merges (see the green runs already on this branch).


zonotope (approved):
💜

PR #1371, opened 2026-06-24
feat: shared on-disk cache for Iceberg data files

bplatz:
## Summary

Adds a read-through on-disk cache so repeated Iceberg queries serve data files from local disk instead of re-pulling them from S3 on every query. It shares **one global disk budget** with Fluree's own object-storage cache — there is no separate Iceberg budget.

## What it does

The whole-file-vs-range decision is made **in the Parquet reader**, after the footer metadata reveals how much of the file a query actually fetches. A 3-rule policy, no per-query knobs:

1. Serve a whole file already cached on disk (size-validated).
2. Otherwise cache + read the whole file when the read is **cheap** (file ≤ 32 MiB) or **broad** (projected column ranges ≥ 50% of the file by bytes).
3. Otherwise range-read the projected chunks and do not cache.

Narrow projections keep doing cheap range reads (no whole-file pull); broad analytical scans become whole-file cached automatically. The policy applies to both the small (sparse-buffer) and large (> 64 MiB) read paths:

- Large cache hits and fills are served as **lazy byte ranges from the local file** (via a small local-file storage shim) — a large file is never loaded whole into memory.
- Large broad-read admission is capped by a fill ceiling (256 MiB) to bound the one-time fill.
- Fills go through the cache's **single-flight coalescing**, so concurrent queries touching the same file share one S3 GET and one write.

The local cache key embeds the expected file size and reads are length-checked, so a truncated or otherwise wrong cached file is never served.

## Configuration

A first-class global disk-budget knob mirroring `cache_max_mb`:

- server `--disk-cache-max-mb` / `FLUREE_DISK_CACHE_MAX_MB` / config-file `disk_cache_max_mb`
- `FlureeBuilder::disk_cache_max_mb`

Resolved as: `FLUREE_DISK_CACHE_BUDGET_BYTES` (bytes) env override > configured `disk_cache_max_mb` > auto-detect from free disk. `0` disables disk caching (the cache becomes a near-free passthrough). Documented in `operations/configuration.md`.

## Design notes

- `DiskArtifactCache` (LRU byte budget, eviction, in-flight fetch coalescing) is hoisted from `fluree-db-binary-index` to `fluree-db-core` (behind the existing `native` feature). binary-index re-exports it and behaves identically.
- Iceberg and Fluree object storage share one cache instance and budget automatically — the cache is a process-wide singleton per directory, so using the same cache directory gives one global budget with no explicit handoff.
- Putting the admission decision in the read path (not the provider) is what keeps general queries fast: a narrow projection still range-reads cheaply, while a broad scan caches the whole file without the caller having to classify the workload.

## Correctness & testing

- Reader-level unit tests: the admission decision (cheap / broad / narrow, inclusive share boundary), the size-validation guards (key includes size; `valid_local` / `read_whole_local` reject a wrong-length file), and local-file positioned range reads.
- The core disk-cache tests move with the hoist and pass under the `native` feature.
- Validated end-to-end on TPC-H over MinIO: narrow scans cache nothing (range-read), broad/small scans cache whole, warm scans serve size-validated hits, and the large-file path fills + serves local ranges — all results exact. (Large-path validated by temporarily lowering the sparse threshold so real data files route through it.)

## Blast radius

Scoped to the native disk cache + the Iceberg read path. binary-index uses the hoisted cache unchanged; with the disk budget at `0` the whole thing is a passthrough. Only **data** files are disk-cached — metadata/manifest files are not (they sit in the in-process cache and `version-hint.text` is mutable).

- Cache hit/miss/evict counters, to tune the 32 MiB / 50% thresholds in production.
- Sparse-buffer over-allocation for narrow reads of 32–64 MiB files.
- Iceberg-vs-Fluree eviction interaction under the shared budget (no per-class reservation yet — by design for now).

Reviewers: @aaj3f @zonotope


fluree-db-iceberg/src/io/send_parquet.rs
```diff
@@ -262,37 +414,107 @@ impl<'a, S: SendIcebergStorage> SendParquetReader<'a, S> {
...
+                    runtime,
+                )
+                .await;
+            }
+            // Rule 2: admit a broad read of a large file within the fill ceiling.
+            // Fill once (coalesced) then serve from local; narrow large reads
+            // skip this and range-read from source.
+            if file_size <= LARGE_FILE_ADMIT_MAX_BYTES {
+                let metadata = self.read_metadata(&path).await?;
+                let projected_bytes = projected_chunk_bytes(task, &metadata)?;
```
aaj3f, line 480: I think for the "narrow" "large-file" path we're doing `read_metadata` then discarding it. Maybe `decode_large_file` should get that metadata threaded so it doesn't 2x perform the operation on each file

fluree-db-iceberg/src/io/send_parquet.rs
```diff
@@ -262,37 +414,107 @@ impl<'a, S: SendIcebergStorage> SendParquetReader<'a, S> {
...
+                return Self::decode_large_file(
+                    storage,
+                    path,
+                    file_size,
+                    projected_field_ids,
+                    iceberg_schema,
+                    runtime,
+                )
+                .await;
+            }
```
aaj3f, line 474: This improvement seems nice, but does it become possible for a lazily-loaded set of ranges--between the ranges being served at `valid_local()` and then the deferred range reads actually being performed--does it become possible for the cache to evict that same file so that we end up getting a `IcebergError::Storage` and a top-level query failure? This wasn't a possibility prior to these changes, I think. At the very least we could solve with a fallback to read from source.

aaj3f (approved):
Some notes below but 🚀

fluree-db-iceberg/src/io/send_parquet.rs
bplatz: Good catch — addressed in 71535c684cad3a0b468880534562f86b2dc3c3e5: a failed local read now falls back to the source store instead of erroring the query.

PR #1372, opened 2026-06-25
perf: fused analytical-aggregate execution for R2RML/Iceberg graph sources

bplatz:
## Summary

Adds a **fused execution path for analytical aggregate queries over R2RML/Iceberg graph sources**. When a query is a single graph-source scan feeding a `COUNT` / `SUM` / `AVG` aggregation (with or without `GROUP BY`, a `FILTER`, or simple arithmetic in the aggregate), it is now answered by folding the aggregate **directly from the decoded column batches** — skipping the per-row RDF term materialization that these analytical shapes don't need.

This targets the aggregation- and group-by-heavy queries (the TPC-H Q1 / Q6 family) where most of the work was materializing intermediate RDF bindings that the aggregate immediately folded away. Point lookups, joins, and non-aggregate queries are unaffected.

## What it does

- **Fused aggregate operator** (`r2rml::fused_aggregate`): detects the shape at plan time and folds `COUNT(*)`, `COUNT(?col)`, `COUNT(?subject)`, `SUM`/`AVG` of a column or of an exact decimal arithmetic expression (e.g. `SUM(?price * (1 - ?discount))`), grouped by string/integer key columns, with an optional `FILTER`. `ORDER BY` / `LIMIT` / `OFFSET` wrap the operator with the engine's existing sort/limit.
  - Numeric folds accumulate natively (`i128` for decimal/integer, `f64` for double) and finalize through the same `NumericAcc` the normal aggregate uses, so results are **identical** to the standard pipeline.
  - A row-validity check mirrors the R2RML star's row-drop semantics (a row participates only when the subject and every predicate column are present), so results stay correct in the presence of nulls.
- **Streaming table scan**: `R2rmlTableProvider::scan_table` now yields column batches as data files are read (`ColumnBatchStream`) instead of collecting the whole table first, and the scan operator consumes them in bounded windows.
- **Conservative by design**: detection is gated, and the operator resolves its columns at open against the live mapping. Anything it can't fold soundly — a join (RefObjectMap), a non-numeric/non-column aggregate, an unsupported expression, a non-R2RML graph — **falls back to the exact normal pipeline**, so general graph-source semantics are unchanged. A `FILTER` is only fused alongside a `GROUP BY`, where the savings outweigh the per-row filter evaluation; otherwise the standard path (with Iceberg file pruning) handles it.
- **Kill switch**: `FLUREE_FUSED_R2RML_AGG=0` forces the normal pipeline.

## Results

Measured on TPC-H SF1 (6M-row `lineitem`) over a Direct Iceberg catalog, each **bit-identical to the standard pipeline**:

| Query | Shape | Before | After |
|---|---|---|---|
| Single-predicate `COUNT` | scan + count | ~3.9 s / 5.3 GB | **~0.19 s / 0.3 GB** |
| `SUM` / `AVG` of a column | scan + reduce | ~4.0 s / 5.1 GB | **~0.18 s / 0.3 GB** |
| `GROUP BY` + `COUNT`/`SUM`/`AVG` | grouped aggregate | ~10.2 s / 13.9 GB | **~0.6 s / 0.5 GB** |
| Q1 (`GROUP BY` + `FILTER` + 8 aggregates incl. 2 expressions) | full analytical report | ~23.6 s / 19.3 GB | **~10.2 s / 0.8 GB** |
| `COUNT` over a same-subject 2-predicate star | grouped count | ~5.5 s / 7.2 GB | **~0.3 s / 0.5 GB** |

The memory reductions are the headline: these grouped analytical queries previously built and held the full table's worth of intermediate bindings, which is also what limited their scaling to larger datasets. Folding from columns removes that intermediate, so the same queries scale much further.

## Testing

- Unit tests for the shape detector and decimal arithmetic; the existing R2RML integration suite passes unchanged.
- Each shape above was validated against the standard pipeline (kill switch on/off) for exact value equality, including the expression aggregates and `ORDER BY`.

## Notes

- Base branch is `feature/shared-disk-cache` (this builds on the shared on-disk cache work).
- Decimal expression folding uses fixed-width `i128` (≈38 digits) for the exact intermediate, which is ample for analytical decimal aggregates; the kill switch is available if a workload ever needs the unbounded path.

Reviewers: @aaj3f @zonotope


fluree-db-query/src/execute/operator_tree.rs
```diff
@@ -2430,6 +2430,32 @@ fn build_operator_tree_inner(
...
+            // semantics on the small grouped output).
+            if !query.ordering.is_empty() {
+                op = Box::new(crate::sort::SortOperator::new(op, query.ordering.clone()));
+            }
+            if let Some(offset) = query.offset {
+                op = Box::new(OffsetOperator::new(op, offset));
+            }
+            if let Some(limit) = query.limit {
+                op = Box::new(LimitOperator::new(op, limit));
+            }
```
aaj3f, line 2464: The fallback was built from the full unmodified query so `let fallback` still carries `LIMIT` / `OFFSET` / `ORDER BY` but then they're doubly-applied in lines 2446-2454. I think `LIMIT` and `ORDER BY` may be harmless but `OFFSET` wouldn't be. 

It seems at `operator_tree.rs:~2797-2802` we clear those on fallback and likely should here as well

aaj3f: Maybe a regression test here as well

fluree-db-query/src/r2rml/fused_aggregate.rs
aaj3f, line 1: Was trying to find a good place to make this a line-specific set of comments, but it's kind of all across this file. In accumulator paths, the file seems to handle `i128` overflow well (i.e. `sum.checked_add` -> `None` and then `overflowed` re-runs with `BigDec`), but `eval_dec` and one of the expression arms of `next_batch` doesn't. It seems maybe there's two reasons for a `None` there: (1) a null operand, which is a legitimate cause for row drop, and (2) an i128 overflow. But both seem to collapse into `true` / "contributes-nothing" arm, so an overflow silently omits that row's term and doesn't inc avg counts etc

fluree-db-query/src/r2rml/fused_aggregate.rs
```diff
@@ -0,0 +1,1377 @@
...
+        }
+        Column::Float32(values) => {
+            if let Some(Some(v)) = values.get(row) {
+                *sum += *v as f64;
+                *count += 1;
+            }
+        }
+        _ => {}
+    }
+}
```
aaj3f: In paths like `fluree-db-query/src/aggregate.rs:627-651`, there's a guard and explicit drop for `NaN`, but here there isn't. So I think here an `xsd:double` column like `[1.0, NaN, 3.0]` would be `SUM = NaN`, `count = 3`, `AVG = NaN` where elsewhere it would be `SUM = 4.0`, `AVG = 4.0/2 = 2.0`

aaj3f: I can't find it now but I swear there was some other fn like `accumulate_double` that had the same issue

aaj3f (approved):
Big wins! Just a few notes

PR #1373, opened 2026-06-25
fix: stabilize raft leader failover (election livelock, follower cache staleness, test race)

bplatz:
## Background

CI/CD was intermittently failing on `fluree-db-server::raft_multi_node leader_failover_resumes_writes` — a transient red that didn't reproduce on a re-run and didn't obviously point at any one change. Reproducing it locally showed a ~60% failure rate with *three distinct* panics, which turned out to be three independent issues. Two are real product bugs in the consensus layer; the third is a race in the test itself. The flakiness was self-masking: the test usually failed at an earlier stage before it could exercise the most important bug.

### 1. Election livelock — no new leader after failover (product)
The default election timeout (openraft's stock 150–300 ms) sat *below* `NetworkConfig::rpc_timeout` (500 ms) and near `connect_timeout` (250 ms to a dead-but-not-yet-evicted voter). A candidate re-elected — bumping its term — before its own vote RPCs from the previous round resolved, so every vote landed on a now-stale term and no quorum ever formed. Survivors livelocked: all four climbed to the same term with `leader = None` and never converged.

Fix: derive the election window from the transport timeouts so the invariant `election_timeout_min > rpc_timeout` holds by construction (`2× rpc_timeout`, 750 ms floor, 2× spread → 1000–2000 ms at the defaults). `RaftIntegration::bootstrap` also warns at startup if `election_timeout_min <= rpc_timeout`, so an embedder who tunes either config independently and reintroduces the risk is told about it.

### 2. Committed write invisible on followers (product — the actual CI failure)
After a write replicated and applied on **every** node (all at the same raft applied index), only the staging node's query reflected it; followers kept serving the prior head. The local cache event listener reconciled the ledger view only on `LedgerIndexPublished`, never `LedgerCommitPublished` — so a committed-but-unindexed write left follower caches stale until some later index publish happened to land. The leader stayed correct because its own commit path updates its cache directly.

Fix: reconcile on commit publishes as well as index publishes (matching the listener's own documented contract). While here, the listener now also recovers from broadcast `Lagged`: rather than only logging, it sweeps the loaded ledgers and reconciles each against the nameservice head, so a dropped event can't strand a ledger's cache.

### 3. Test race on a transient leadership 503 (test)
The test read the new leader from one node, then immediately wrote through a *different* surviving follower that hadn't yet learned the new leader — getting a retryable `503 "no leader currently elected"`. It now waits until every surviving node agrees on the new leader before resuming writes.

## Verification

`leader_failover_resumes_writes` was run in a tight loop (90+ iterations across the iterations of this change) and is now consistently green; the full `raft_multi_node` and `raft_integration` suites pass, as do the `fluree-db-api` unit tests. fmt + clippy clean.

### DIFF
--- a/fluree-db-api/src/lib.rs
+++ b/fluree-db-api/src/lib.rs
@@ -1222,28 +1222,7 @@ pub fn spawn_local_cache_event_listener(
                         ..
                     },
                 ) => {
-                    match ledger_manager
-                        .notify(NsNotify {
-                            ledger_id: ledger_id.clone(),
-                            record: None,
-                        })
-                        .await
-                    {
-                        Ok(result) => {
-                            tracing::debug!(
-                                alias = %ledger_id,
-                                ?result,
-                                "local cache event listener reconciled ledger"
-                            );
-                        }
-                        Err(error) => {
-                            tracing::warn!(
-                                alias = %ledger_id,
-                                error = %error,
-                                "local cache event listener failed to reconcile ledger"
-                            );
-                        }
-                    }
+                    reconcile_cached_ledger(&ledger_manager, &ledger_id).await;
                 }
                 Ok(fluree_db_nameservice::NameServiceEvent::LedgerRetracted { ledger_id }) => {
                     ledger_manager.disconnect(&ledger_id).await;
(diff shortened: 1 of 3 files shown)


fluree-db-api/src/lib.rs
```diff
@@ -1240,10 +1233,22 @@ pub fn spawn_local_cache_event_listener(
...
                     tracing::warn!(
                         skipped,
-                        "local cache event listener lagged behind nameservice events"
+                        ledgers = aliases.len(),
+                        "local cache event listener lagged; sweeping loaded ledgers"
                     );
+                    for alias in aliases {
+                        reconcile_cached_ledger(&ledger_manager, &alias).await;
+                    }
                 }
```
aaj3f, line 1260: I don't know if we've observed this `Lagged` state demonstrably, but my only concern is that, because the reconciliation sweep happens within an inner loop and so isn't helping drain the receiver, if the listener is lagging **because** reconciles are slow, then it could possibly contribute to a re-trigger of `Lagged`. Perhaps this could be hardened w/ a future bound or a scope to sweep only ledgers that are provably behind etc? Or we can just make a comment for now and revisit only if we prove this issue can/does arise

aaj3f (approved):
Seems like a good & correct fix. Only one detailed note below. At some point it may be valuable to have a kind of firehose bench to assess the conditions under which we start to see `Lagged` get triggered and to validate the fallback handling. There may be improvements we could make to take pressure off the broadcast channel or to better improve fallback reconciliation to recover from it, but not likely worth doing now

fluree-db-api/src/lib.rs
bplatz: Good call — added a comment for now noting the re-trigger risk and that we'll revisit with a bound or provably-behind scope only if it proves to arise (29b46cfe7).

PR #1374, opened 2026-06-25
fix(sparql): aggregate and sub-select edge cases

### LINKED ISSUE #1362: `GROUP BY DATATYPE(?v)` returns one row per binding (group key not collapsed; `LIMIT` does not cap output)
## Summary

First off — thanks for the expression-based `GROUP BY` work in **#1333**; grouping on expressions like `LCASE(?x)` collapses correctly now. I think there may be a small residual case that fix doesn't cover: grouping **directly** on a `DATATYPE(?v)` expression. In that case the group key doesn't appear to collapse — the query returns **one row per input binding** rather than one row per distinct datatype — and `LIMIT` does not cap the emitted result set.

## Version

- `fluree/server:4.1` (`:latest`, v4.1.0)
- Single binary, local filesystem storage
- `POST /v1/fluree/query/<ledger>` with `Content-Type: application/sparql-query`

## Steps to reproduce

Create a ledger and insert three `xsd:decimal` values:

```bash
BASE=http://localhost:8090/v1/fluree

curl -sS -X POST "$BASE/create" -H "Content-Type: application/json" \
  --data '{"ledger":"groupby-datatype-demo"}'

curl -sS -X POST "$BASE/insert?ledger=groupby-datatype-demo" \
  -H "Content-Type: application/json" --data '{
  "@context": {"ex":"http://example.org/","xsd":"http://www.w3.org/2001/XMLSchema#"},
  "@graph": [
    {"@id":"ex:a","ex:amount":{"@value":"1.5","@type":"xsd:decimal"}},
    {"@id":"ex:b","ex:amount":{"@value":"2.5","@type":"xsd:decimal"}},
    {"@id":"ex:c","ex:amount":{"@value":"3.0","@type":"xsd:decimal"}}
  ]
}'
```

Then group by the datatype:

```sparql
SELECT (DATATYPE(?v) AS ?dt) (COUNT(?v) AS ?n)
WHERE { ?s ex:amount ?v }
GROUP BY DATATYPE(?v)
LIMIT 10
```

## Expected

A single group row:

| `?dt` | `?n` |
|-------|------|
| `http://www.w3.org/2001/XMLSchema#decimal` | `3` |

## Actual

**Three** rows — one per input binding, each with `?n = 1` — i.e. the grouping key isn't collapsed. As the number of matching triples grows, the emitted row count tracks the number of bindings, and adding `LIMIT` does not cap the result set.

## What collapses vs. what doesn't (same build)

| Query shape | Result |
|-------------|--------|
| `GROUP BY (LCASE(?strVar))` | ✅ collapses, honors `LIMIT` |
| `… BIND(DATATYPE(?v) AS ?dt) … GROUP BY ?dt` | ✅ collapses to one row |
| `GROUP BY DATATYPE(?v)` (expression directly) | ⚠️ one row per binding; `LIMIT` doesn't cap |

So it looks like the grouping key derived from `DATATYPE()` isn't being matched/normalized the same way other expression group keys are.

## A query-level workaround

Binding the expression to a variable first behaves as expected:

```sparql
SELECT ?dt (COUNT(?v) AS ?n)
WHERE { ?s ex:amount ?v . BIND(DATATYPE(?v) AS ?dt) }
GROUP BY ?dt
```

## Possibly related

This looks adjacent to **#1333** (which fixed the general expression-`GROUP BY` collapse + `LIMIT` behavior) — it may just be a `DATATYPE()`-specific gap in that same path rather than a separate area. Happy to test a patch or provide more detail.

## Why it'd be nice to fix

`DATATYPE()` returns the datatype IRI, so `GROUP BY DATATYPE(?v)` is a natural way to summarize which value types appear on a predicate. Having it collapse (and honor `LIMIT`) would make that a one-liner. Thanks again for all the recent query-engine work!

bplatz:
Addresses some SPARQL aggregate and sub-select edge cases surfaced while exercising analytical queries. These are small cleanups that bring a few corner cases in line with how the rest of the engine already behaves — no large feature work.

### What's here

- **Sub-SELECT `ORDER BY` / `LIMIT` scoping** — an inner `ORDER BY … LIMIT n` on a sub-SELECT is now evaluated to its complete sliced result before it joins the enclosing pattern (instead of the slice effectively applying after the join). Also fixes a related case where a parent batch that matched nothing could end the join early.
- **Empty-group `SUM` / `AVG`** — over an empty match these now return `0`, matching what `COUNT` already did, rather than an unbound cell. Covers both the generic and the indexed scalar-aggregate fast paths.
- **`GROUP BY` on a bare built-in** — `GROUP BY DATATYPE(?v)` (a built-in / function call without surrounding parens) now groups correctly; the unparenthesized form was previously dropped, collapsing the query to one implicit group.
- **Inline blank-node property lists** — `[ :p ?o ]` in subject/object position (and in `CONSTRUCT` / `INSERT` / `DELETE` templates) now binds its nested variables.
- **Projection expression errors** — a value error in a `SELECT` / `BIND` / `ORDER BY` expression (e.g. arithmetic over a heterogeneous column) leaves that one cell unbound for the row instead of failing the whole query; structural errors (wrong arity, etc.) still surface, and transactions keep strict evaluation.

### Testing

- Added integration, parser, and unit regression tests for each case.
- A handful of previously-failing W3C SPARQL eval cases now pass, with no regressions in the suite.
- Clippy + nextest green on the touched crates.

Closes: #1362


fluree-db-query/src/subquery.rs
```diff
@@ -154,28 +154,36 @@ impl SubqueryOperator {
...
+            // complete, ordered, sliced solution sequence) and then joined — it
+            // is never seeded per parent row, which would correlate it (and let
+            // an inner LIMIT/OFFSET apply per row, leaking rows past the slice;
+            // W3C subquery/sq11). Materialize once + hash-join on `join_keys`.
+            // A non-produced correlation var is unbound in every inner row, so
+            // omitting it from the key is the correct natural join (unbound is
+            // compatible with any parent value), and the merge fills it from the
+            // parent. So this is correct regardless of `pass_through_ok`.
+            true
+        } else {
```
aaj3f, line 187: It seems the SPARQL lowering sites in the code(`lower/select.rs:683`, `lower/pattern.rs:63`, `lower/describe.rs:87`) set `uncorrelated = true` on **every** SPARQL sub-SELECT, which then forces `join_mode = true` (materialize-once + hash-join) **unconditionally**. That would then bypass both the `limit/offset` eligibility check and the `SUBQUERY_MATERIALIZE_MIN_PARENT_ROWS` cardinality guard for all of them, including plain sub-SELECTs with no ORDER BY/LIMIT/OFFSET and no aggregate

It seems per §18.2, materialize-once is "correctness-mandated" only when the subquery is sliced or aggregated. For the plain, equivalent case this then is, I think, an over-broad de-optimization. There aren't any benchmarks in this PR for this case, but in a case where the shape is a small, selective outer that shares a produced join-key var with a large, selectively-keyable inner, we'll no longer benefit from the legacy cardinality-guarded per-row pruning path that computed only the inner rows matching the outer's few keys

fluree-db-query/src/aggregate.rs
```diff
@@ -529,15 +530,16 @@ impl NumericAcc {
...
+    /// `"0"^^xsd:integer` (SPARQL 1.1 §18.5.1.3; W3C agg-empty-group-sum).
     ///
     /// Result datatype follows the W3C arithmetic promotion lattice:
     ///   all integer  → xsd:integer (Long if it fits, else BigInt)
     ///   any decimal  → xsd:decimal
     ///   any double   → xsd:double
     pub(crate) fn finalize_sum(self) -> Binding {
         if self.count == 0 {
-            return Binding::Unbound;
+            return Binding::lit(FlakeValue::Long(0), xsd_integer());
```
aaj3f, line 542: More of a question than a suggestion. The general fix here makes sense to me, but it seems this would also now affect scenarios where `count == 0` is also reached by a non-empty group where all values are non-numeric/unbound (i.e. reached by both truly-empty groups and some non-empty groups). I don't know what's right for this, though. If this is not the desired behavior, we'd also need a kind of "saw any row" flag

fluree-db-query/src/subquery.rs
```diff
@@ -154,28 +154,36 @@ impl SubqueryOperator {
...
+        let join_mode = if subquery.uncorrelated {
+            // SPARQL 1.1 §18.2: a sub-SELECT is evaluated INDEPENDENTLY (its
+            // complete, ordered, sliced solution sequence) and then joined — it
+            // is never seeded per parent row, which would correlate it (and let
+            // an inner LIMIT/OFFSET apply per row, leaking rows past the slice;
+            // W3C subquery/sq11). Materialize once + hash-join on `join_keys`.
+            // A non-produced correlation var is unbound in every inner row, so
+            // omitting it from the key is the correct natural join (unbound is
+            // compatible with any parent value), and the merge fills it from the
+            // parent. So this is correct regardless of `pass_through_ok`.
```
aaj3f: I'm not entirely positive I'm right about this, but it seems maybe that the claim, "correct regardless of `pass_through_ok`", may **not** be true for a correlation var that's bound **inside** the sub-SELECT by an aggregate/BIND/OPTIONAL rather than a top-level triple. Those would be excluded from `join_keys` and yet would be bound in the materialized rows. And then the non-clobbering merge in `subquery.rs:417-423` would keep the parent's value with no equality check, so a row that should fail the natural join might survive with the parent's binding. 

I'm thinking of something like

```sparql
SELECT ?s 
WHERE { 
   ?s :p ?c . 
   { 
      SELECT ?s (COUNT(?x) AS ?c) 
      WHERE { ?s :r ?x } 
      GROUP BY ?s 
   } 
}
```

I'm not sure in this query we'd see enforcement of `:p == count`, for example.

With that said, if the above is right, I don't think this PR introduced this (I think it's pre-existing if true), and it's probably beyond the scope of this PR, but perhaps it's worth investigation and a follow-up issue.

aaj3f (approved):
Good SPARQL conformance fixes @bplatz 👍 A few notes below, one of them possibly more significant from a performance regression perspective.

Also, it seems all new regression tests landed only in `it_query_sparql.rs` /  `it_query_sparql_indexed.rs`. That makes sense, because the focus of the PR was SPARQL, but there were additional consequences on the JSON-LD surface as well, and so perhaps it's worth some valuable additions to `it_query.rs` / `it_query_analytical.rs` / `it_query_grouping.rs`. e.g. a JSON-LD analytical test asserting SUM/AVG over an empty match = 0, and a JSON-LD projection/BIND expression-error test asserting an unbound cell rather than a 400. Those tests might guarantee against regressions for both (i.e. guarantee we don't ever diverge in correctness between SPARQL & JSON-LD).

bplatz:
Addressed in 245e67d97b712c5b82f9987c090dfc9f7a0af114

- **Sub-SELECT materialize-once over-broad (perf):** fixed — narrowed the unconditional `join_mode = true` so materialize-once is forced only when §18.2 actually requires it (LIMIT/OFFSET, DISTINCT, aggregate, or a correlation var referenced-but-not-produced). Plain projection sub-SELECTs fall through to the cardinality-guarded path and can prune via per-row seeding again. sq11 (inner LIMIT) still materializes once; full SPARQL + query suites green.
- **JSON-LD parity tests:** added — SUM/AVG over an empty match = 0, and a SELECT/BIND projection value error → unbound cell (not 400), both on the JSON-LD surface.
- **Empty-group SUM/AVG = 0 for non-empty all-unbound groups:** no change — that's spec-aligned (`Sum(())`/`Avg(())` = `0^^xsd:integer`, and unbound values aren't in the multiset), and it's now covered by the JSON-LD parity test. No "saw any row" flag needed.
- **Aggregate-produced correlation var not join-enforced:** confirmed real, but **pre-existing** — I reproduced the identical wrong output (`[ex:a, ex:b]`) on `origin/main`, so it's out of scope here. Tracked in #1388 with the repro and a suggested fix direction.

PR #1376, opened 2026-06-25
`docs`: fix landing page links

Haleshot:
I first noticed the [AI docs link](https://fluree.github.io/db/ai/README.html) was sending readers to a 404. After building the `mdBook` locally and running [`lychee`](https://github.com/lycheeverse/lychee), similar issues showed up on the rest of the landing page links, so this keeps that intro page from sending readers to incorrect paths.

This might be worth turning into a lightweight docs check with [`lychee-action`](https://github.com/lycheeverse/lychee-action) (the workflow could start by checking `.md` files first, since that maps failures directly back to the files authors edit. Once that is clean, checking the generated `mdBook` outputs would catch publishing-specific issues like `README.md` links becoming `README.html`.).

The repo does have enough links for it to help: the local pass checked 1,406 links and the remaining failures might be useful for future _follow-up_ docs cleanup PRs.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -19,7 +19,7 @@ Fluree DB is a single binary that stores your data as an RDF knowledge graph, qu
 
 ## Start here
 
-- **New to Fluree?** → [Getting started](getting-started/README.md)
+- **New to Fluree?** → [Getting started](getting-started/)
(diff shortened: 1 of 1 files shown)


bplatz (approved):
Thanks!!

PR #1377, opened 2026-06-25
fix(cli): accept -l/--ledger on query/insert/upsert/update

### LINKED ISSUE #1368: Documentation Error: `fluree query` examples use unsupported `--ledger` option
## Summary

The [documentation](https://fluree.github.io/db/getting-started/tutorial-end-to-end.html#step-4-branch-to-experiment-safely) shows examples using:

```bash
fluree query --ledger <ledger-name>:<branch> '<SPARQL>'
```

However, the current CLI rejects the `--ledger` option and requires the ledger to be supplied as a positional argument.

As a result, users following the documentation encounter an error even though the query itself is valid.

## Environment

* Fluree CLI: [please insert version from `fluree --version`]
* OS: macOS
* Server: Local Fluree server

## Documentation Example

The documentation shows examples similar to:

```bash
fluree query --ledger knowledge-base:main '
PREFIX schema: <http://schema.org/>

SELECT ?title ?visibility
WHERE {
  ?doc a ex:Article ;
       schema:name ?title ;
       ex:visibility ?visibility .
}'
```

## Actual Behavior

Running the documented command produces:

```text
error: unexpected argument '--ledger' found

  tip: to pass '--ledger' as a value, use '-- --ledger'

Usage: fluree query [OPTIONS] [ARGS]...
```

## Reproduction

```bash
fluree query --ledger knowledge-base:main '
PREFIX schema: <http://schema.org/>

SELECT ?title ?visibility
WHERE {
  ?doc a ex:Article ;
       schema:name ?title ;
       ex:visibility ?visibility .
}'
```

Result:

```text
error: unexpected argument '--ledger' found
```

## Working Command

The query succeeds when the ledger is supplied as the first positional argument:

```bash
fluree query knowledge-base:main '
PREFIX schema: <http://schema.org/>

SELECT ?title ?visibility
WHERE {
  ?doc a ex:Article ;
       schema:name ?title ;
       ex:visibility ?visibility .
}
ORDER BY ?title'
```

Output:

```text
+----------------------------+--------------+
| title                      | visibility   |
+===========================================+
| Deployment Runbook         | internal     |
| Incident Response Playbook | confidential |
| Onboarding Guide           | public       |
+----------------------------+--------------+
```

## Expected Behavior

Either:

1. The documentation should be updated to use the positional ledger argument syntax:

```bash
fluree query knowledge-base:main '<SPARQL>'
```

or

2. The CLI should support the documented `--ledger` option.

## Additional Observation

The CLI appears inconsistent across commands. For example:

```bash
fluree history http://example.org/doc1 --ledger knowledge-base
```

accepts a `--ledger` option, while:

```bash
fluree query --ledger knowledge-base:main ...
```

does not.

Clarifying the intended command-line conventions would improve usability and reduce confusion for new users.

bplatz:
## Summary

The data commands (`query`, `insert`, `upsert`, `update`) only accepted the ledger as a **positional** argument, so the documented form

```bash
fluree query --ledger knowledge-base:main '<SPARQL>'
```

errored with `unexpected argument '--ledger' found`. Meanwhile the metadata commands (`history`, `branch`, `graph`, `show`, `drop`) already used a `--ledger` flag — an inconsistency reported in #1368.

## Changes

- Add `-l`/`--ledger` to `query`, `insert`, `upsert`, and `update` as an explicit alternative to the positional ledger argument.
- New shared `resolve_inputs(ledger_flag, args)` wraps the existing `resolve_positional_args`: when `--ledger` is given it takes precedence and **bypasses the `looks_like_query` heuristic** (positionals carry only inline data/query or a file path; a stray second positional is rejected with a clear usage error). When the flag is absent, positional behavior is unchanged.
- Both forms now work:
  ```bash
  fluree query knowledge-base:main 'SELECT …'            # positional (still works)
  fluree query --ledger knowledge-base:main 'SELECT …'   # flag (now works)
  ```
- Document the flag in the `docs/cli/{query,insert,upsert,update}.md` option tables.

## Tests

- `query_and_insert_accept_ledger_flag` and `query_ledger_flag_rejects_extra_positional` in `fluree-db-cli/tests/integration.rs`.
- Full CLI integration suite passes (95); `cargo fmt` + `clippy --all-features --all-targets -D warnings` clean.

Closes #1368

### DIFF
--- a/fluree-db-cli/src/cli.rs
+++ b/fluree-db-cli/src/cli.rs
@@ -356,6 +356,12 @@ pub enum Commands {
         #[arg(num_args = 0..=2)]
         args: Vec<String>,
 
+        /// Ledger name (defaults to active ledger). Explicit alternative to
+        /// the positional ledger argument; when given, positional args carry
+        /// only the inline data or a file path.
+        #[arg(short = 'l', long)]
+        ledger: Option<String>,
+
         /// Inline data expression (Turtle or JSON-LD).
         #[arg(short = 'e', long = "expr")]
         expr: Option<String>,
@@ -394,6 +400,12 @@ pub enum Commands {
         #[arg(num_args = 0..=2)]
         args: Vec<String>,
 
+        /// Ledger name (defaults to active ledger). Explicit alternative to
+        /// the positional ledger argument; when given, positional args carry
+        /// only the inline data or a file path.
+        #[arg(short = 'l', long)]
+        ledger: Option<String>,
+
         /// Inline data expression (JSON-LD or SPARQL UPDATE).
         #[arg(short = 'e', long = "expr")]
         expr: Option<String>,
@@ -430,6 +442,12 @@ pub enum Commands {
         /// With 2 args: first is ledger name, second is inline data.
         #[arg(num_args = 0..=2)]
         args: Vec<String>,
+
+        /// Ledger name (defaults to active ledger). Explicit alternative to
+        /// the positional ledger argument; when given, positional args carry
+        /// only the inline data or a file path.
+        #[arg(short = 'l', long)]
+        ledger: Option<String>,
 
         /// Inline data expression (Turtle or JSON-LD).
         #[arg(short = 'e', long = "expr")]
@@ -468,6 +486,12 @@ pub enum Commands {
         /// With 2 args: first is ledger name, second is inline query.
         #[arg(num_args = 0..=2)]
         args: Vec<String>,
+
+        /// Ledger name (defaults to active ledger). Explicit alternative to
+        /// the positional ledger argument; when given, positional args carry
+        /// only the inline query or a file path.
+        #[arg(short = 'l', long)]
+        ledger: Option<String>,
 
         /// Inline query expression (SPARQL or JSON-LD query).
         #[arg(short = 'e', long = "expr")]
--- a/fluree-db-cli/src/cli.rs
+++ b/fluree-db-cli/src/cli.rs
@@ -630,7 +630,7 @@ pub enum Commands {
         entity: String,
 
         /// Ledger name (defaults to active ledger)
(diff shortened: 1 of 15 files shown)


fluree-db-cli/src/commands/insert.rs
```diff
@@ -41,6 +41,43 @@ pub fn resolve_positional_args(
...
+        0 => Ok((Some(ledger), None, None)),
+        1 => {
+            let p = Path::new(&args[0]);
+            if !looks_like_query(&args[0]) && p.is_file() {
+                Ok((Some(ledger), None, Some(p.to_path_buf())))
+            } else {
+                // Inline data/query (or a path the input reader will reject).
+                Ok((Some(ledger), Some(&args[0]), None))
+            }
+        }
```
aaj3f, line 81: In an end-user mistake scenario where `--ledger` is set and a single positional arg is supplied that's neither query-shaped nor an existing file (e.g. a typo'd file path or a stray fat-fingered token), it would silently get treated as inline data and fail later with a possibly-confusing parse error response. Compare that, though, to the 2-positional case in lines 73-77. Perhaps a hint that positionals after `--ledger` must be in-line data/query or an existing file?

fluree-db-cli/src/cli.rs
aaj3f, line 634: Super small, but the metadata commands that *do* use a `--ledger` flag (`history` at 633-634, and the `graph …` subcommands) still expose only the long form `#[arg(long)]` with no `short = 'l'`, so `-l` is now available on the four data commands but not on these

aaj3f (approved):
@bplatz looks like a good resolution to #1368. Two very small nits below to take-or-leave:

bplatz:
Addressed in 050aa28629b2f3cbecc334037235dcfa2e88f05b

PR #1378, opened 2026-06-25
fix(cli): render history records in table/csv output

### LINKED ISSUE #1367: Bug: `fluree history` table output does not render history records returned by server
## Summary

Following along the end-to-end [tutorial](https://fluree.github.io/db/getting-started/tutorial-end-to-end.html) a problem seems to arise with the history section. The example shows an SPARQL query - that does not seem to work. Using `fluree history http://example.org/doc1 --ledger knowledge-base ` seem to work, but the  `fluree history` command returns correct history data only when using `--format json`, the default table output fails to render the returned records.

Instead of displaying transaction number, operation, predicate, and value, the table contains blank cells and `?` placeholders.

## Environment

* Fluree CLI: [please fill in version]
* Fluree Server: local server
* Ledger: `knowledge-base`
* OS: macOS

### Query entity

```bash
fluree query '

SELECT ?p ?o
WHERE {
  ex:doc1 ?p ?o .
}'
```

Result confirms that the entity exists and contains data.

### Request history as JSON

```bash
fluree history http://example.org/doc1 \
  --ledger knowledge-base \
  --format json
```

Returns:

```json
[
  [
    "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "ex:Article",
    1,
    true
  ],
  [
    "http://schema.org/author",
    "ex:alice",
    1,
    true
  ],
  [
    "http://schema.org/name",
    "Deployment Runbook",
    1,
    true
  ],
  [
    "ex:content",
    {
      "@value": "Step 1: Check the monitoring dashboard...",
      "@type": "https://ns.flur.ee/db#fullText"
    },
    1,
    true
  ],
  [
    "ex:content",
    {
      "@value": "Step 1: Check the monitoring dashboard and verify all health checks pass...",
      "@type": "https://ns.flur.ee/db#fullText"
    },
    4,
    true
  ],
  [
    "ex:content",
    {
      "@value": "Step 1: Check the monitoring dashboard...",
      "@type": "https://ns.flur.ee/db#fullText"
    },
    4,
    false
  ]
]
```

The history data appears correct.

### Request history using default output

```bash
fluree history http://example.org/doc1 \
  --ledger knowledge-base
```

Produces:

```text
+---+----+-------+
| t | op | value |
+================+
|   | ?  |       |
|---+----+-------|
|   | ?  |       |
|---+----+-------|
|   | ?  |       |
|---+----+-------|
|   | ?  |       |
+---+----+-------+
```

## Expected Behavior

The table renderer should display the history records, for example:

```text
+---+---------+------------------+----------------+
| t | op      | predicate        | value          |
+---+---------+------------------+----------------+
| 1 | assert  | rdf:type         | ex:Article     |
| 1 | assert  | schema:author    | ex:alice       |
| 1 | assert  | schema:name      | Deployment...  |
| 1 | assert  | ex:content       | Step 1...      |
| 4 | assert  | ex:content       | Updated text   |
| 4 | retract | ex:content       | Original text  |
+---+---------+------------------+----------------+
```

## Suspected Cause

The JSON output suggests that history rows are returned as positional arrays:

```json
[predicate, value, t, op]
```

The table formatter may be expecting a different structure (e.g. named fields such as `t`, `op`, and `value`) and therefore cannot populate the columns correctly.

The issue appears to be limited to the CLI table rendering layer because the underlying history data returned by `--format json` is correct.

bplatz:
> Stacked on #1377 (`fix/cli-ledger-flag-1368`) — base is that branch, not `main`. Review/merge #1377 first.

## Summary

`fluree history <entity>` with the default (table) output rendered blank cells and `?` placeholders instead of the history records, even though `--format json` returned correct data (#1367).

## Root cause

History queries project a select list, so the engine returns each row as a **positional JSON array**:

- no predicate filter: `[?p, ?v, ?t, ?op]`
- `--predicate` filter: `[?v, ?t, ?op]`

The table and CSV formatters instead read each row as an **object** (`row.get("?t")`, `row.get("?op")`, …). On an array every lookup returns `None`, producing blank `t`/`value` cells, a `?` op placeholder (the `unwrap_or("?")`), and a missing predicate column. `--format json` was unaffected because it prints the raw array verbatim. The gap slipped through because existing history tests only exercised `--format json`.

## Fix

- Add a `HistoryRow` accessor that reads rows positionally by array length (4 → `[p,v,t,op]`, 3 → `[v,t,op]`), still tolerating object-keyed rows defensively.
- Route both the table and CSV formatters through it.

## Output after the fix

```
+---+----+-------------------------------------------------+-------------------------------------------+
| t | op | predicate                                       | value                                     |
+===========================================================================================...=======+
| 1 | +  | http://www.w3.org/1999/02/22-rdf-syntax-ns#type | ex:Article                                |
| 1 | +  | ex:author                                       | ex:alice                                  |
| 1 | +  | ex:content                                      | Step 1: Check the monitoring dashboard... |
| 1 | +  | ex:name                                         | Deployment Runbook                        |
| 2 | +  | ex:content                                      | ...verify all health checks pass...       |
| 2 | -  | ex:content                                      | Step 1: Check the monitoring dashboard... |
+---+----+-------------------------------------------------+-------------------------------------------+
```

## Tests

- Unit tests on both formatters covering both array shapes, CSV output, the object-row fallback, and the empty case.
- Extended the `history_shows_changes` integration test to assert the default table renders populated cells (no `?`), exercising the real local query path.
- `cargo fmt` + `clippy --all-features --all-targets -D warnings` clean.

Closes #1367


fluree-db-cli/tests/integration.rs
```diff
@@ -1444,6 +1444,18 @@ fn history_shows_changes() {
...
+    fluree_cmd(&tmp)
+        .args(["history", "ex:alice"])
+        .assert()
+        .success()
+        // Header + populated value cells.
+        .stdout(predicate::str::contains("predicate"))
+        .stdout(predicate::str::contains("Alice Smith"))
+        .stdout(predicate::str::contains("ex:name"))
+        // The pre-fix bug produced a `?` op column and blank t/value cells.
+        .stdout(predicate::str::contains("| ? |").not());
```
aaj3f: Claude tested this line in reality and indicated that, given how `comfy-table` handles ASCII values within `|` borders, the actual single-char `?` scenario would render as `| ?  |` not `| ? |`. If that's true, then this line of the test will never actually catch a valid regression. Perhaps we should better assert the condition on a whitespace-normalized form.

fluree-db-cli/src/commands/history.rs
```diff
@@ -194,6 +194,66 @@ fn format_history_result(json: &serde_json::Value, format: OutputFormatKind) ->
...
+                value: cols.get(1),
+                t: cols.get(2),
+                op: cols.get(3),
+            },
+            serde_json::Value::Array(cols) if cols.len() == 3 => HistoryRow {
+                predicate: None,
+                value: cols.first(),
+                t: cols.get(1),
+                op: cols.get(2),
+            },
```
aaj3f, line 231: We may just want a comment near the `select` code site making it clear that, because the arity handling here is pinned to the contract of the `select`, if the `select` site changes, then this would need changing or risk regression

aaj3f (approved):
Looks good! 👍 two nit comments below

bplatz:
Addressed in 4c39d62d23e2b517098fbd47e114d63b289af958

PR #1379, opened 2026-06-25
feat: Complete SPARQL property-path support (ZeroOrOne, negated sets, composite & inverse transitive)

bplatz:
Rounds out the final SPARQL 1.1 property-path forms that were not yet covered, so the full W3C property-path grammar now evaluates in the engine instead of returning HTTP 400. Wired in both the SPARQL and JSON-LD (`@path`) lowering paths, which share the same IR and traversal operator.

## What's now supported

- **`p?` (ZeroOrOne)** — at top level and as a sequence step (`p?/q`, `^p?/q`). New `PathModifier::ZeroOrOne`; the traversal operator short-circuits to the node itself plus a single hop.
- **Negated property sets** — `!p`, `!(a|b)`, `!(^p)`, and mixed. Lowered to a fresh predicate-variable triple plus a `FILTER` excluding the listed predicates, unioning forward and inverse directions as needed.
- **Nested modifiers** — `((p)*)*`, `(p+)?`, `(p?)?`, … collapse algebraically to a single modifier.
- **Transitive over a composite sub-path** — `(a/b)+`, `(a/b)*`, `(a/b)?`. Each hop follows the whole sub-path. Steps may be alternations of simple predicates and may be inverse (`(^a/b)+`, `(a/^b)+`, `(^(a|b)/c)+`).
- **Parenthesized inverse-transitive** — `(^p)+`, `(^p)*`, `(^p)?` (and the JSON-LD array form), via the identity `(^X)+ ≡ ^(X+)`.

The one remaining grammar corner — a *nested transitive* step inside a composite unit (`(a+/b)+`) — is intentionally rejected and guarded by a negative test. It is not exercised by any W3C conformance test.

## Implementation notes

- `PropertyPathPattern` gained direction-aware composite steps (`PathStep { predicates, inverse }` + `first_inverse`); the traversal operator's `read_step` maps step direction to a SPOT or POST read (and the opposite index when retreating for backward traversal). Single-predicate count fast paths bail on composite, so hot paths are untouched.
- Negated-set filtering reuses the predicate-variable `FILTER(?p != IRI(...))` comparison, matching how a hand-written filter lowers.
- The both-unbound composite closure is correct but unoptimized (per-start BFS, no shared adjacency map) — noted in code for a future pass.

## Tests & docs

- New SPARQL and JSON-LD parity tests for every form above (forward, backward, both-unbound, and reachability modes), mirroring the relevant W3C `property-path` / `nps_*` cases.
- Updated `docs/query/sparql.md`, `docs/query/jsonld-query.md`, `docs/query/graph-crawl.md`, and `docs/reference/compatibility.md`; corrected stale Rules text that predated this work.

CI: clippy (all features, all targets) + nextest (workspace, all features).


fluree-db-query/src/property_path.rs
```diff
@@ -329,6 +414,41 @@ impl PropertyPathOperator {
...
+                PathModifier::ZeroOrMore | PathModifier::ZeroOrOne
+            ) {
+                for node in self.composite_domain_nodes(ctx).await? {
+                    if !start_set.contains(&node) {
+                        out.push((node.clone(), node));
+                    }
+                }
+            }
+            return Ok(out);
+        }
```
aaj3f, line 609: In the "both-unbound composite closure" scenario of `?s (a/b)+ ?o`, we would run a separate full `traverse_forward` BFS from every start candidate, and every hop of every BFS issues fresh DB index reads via `read_step`. In the code for simple closures (lines 452-543), by contrast, we read each predicate once via PSOT and store in an in-memory adjacency map, then the BFSs run with zero further db calls. The composite branch code builds no such map, so total index reads scale in rough proportion to `start_candidates × closure_size × steps`. 

Also as there's no shared visited/cost budget across the per-start loop, the only cap is each `traverse_forward`'s `max_visited` (10k), so the aggregate work is only capped by `starts × 10k`.

It's a bit of a scalability cliff for the new capability. Whether now or later, perhaps something like a step-wise adjacency map (read each step's edges once via PSOT, compose in memory) and then BFS purely in-memory like the simple path. Or, at minimum, maybe thread a shared visited/budget counter across the per-start loop

fluree-db-query/src/property_path.rs
```diff
@@ -139,57 +139,116 @@ impl PropertyPathOperator {
...
+                    };
+                    if let Some(n) = next {
+                        if seen.insert(n.clone()) {
+                            out.push(n);
+                        }
+                    }
+                }
             }
         }
         Ok(out)
```
aaj3f, line 330: It seems here that `read_step` unconditionally allocates a `HashSet<Sid>` (line 208) and inserts every result to dedup, but does so on every hop, even on ordinary single-predicate `p+`/`p*` traversal, which previously pushed straight to the `Vec` with no per-hop set. For the simple BFS case the dedup is redundant with the operator's outer `visited` set, so it is pure added work (one alloc + N inserts per hop) on the correlated/bound-endpoint traversal path.

I don't think the impact of this is that huge... it's reduced by the per-hop `range_with_overlay` index read. But the PR does mention "hot paths untouched" (which is true for the `COUNT` fast-paths which bail via `single_predicate()==None`), but this is possibly not true for the operator's own per-hop work, which did change. 

In any event, I'd suggest maybe--for the simple single-step single-node case--we skip the `HashSet` (let the BFS `visited` dedup), or gate the dedup on something like `preds.len() > 1 || nodes.len() > 1`

aaj3f (approved):
This is nice work, @bplatz , and I'm glad to have it. Two performance-related notes below. They're not regressions of pre-existing queries, as they relate only to net-new capabilities made possible by this PR, but still likely worth addressing here or in follow-up

PR #1380, opened 2026-06-26
fix(sparql): CONSTRUCT WHERE and STRBEFORE/STRAFTER edge cases

bplatz:
A couple of narrow SPARQL correctness fixes surfaced by the W3C 1.1 query test suite. These are edge-case bugs in already-supported features — not gaps in functionality.

## CONSTRUCT

- **Compact-IRI comparison in the test harness.** Our engine already emits valid *compact* JSON-LD for CONSTRUCT (e.g. `@id: ":s1"` with `@context: {"": "http://example.org/"}`). The W3C harness's graph comparator was reading `@id`/predicate keys literally and ignoring `@context`, so equivalent graphs compared unequal. Added a minimal context expander (prefix terms incl. the empty prefix, `@vocab`, `@base`). Engine output unchanged.
- **`CONSTRUCT WHERE` shorthand was too permissive.** Per the SPARQL 1.1 grammar the shorthand block is triples-only; we were routing it through the full group-graph-pattern parser and accepting `FILTER`/`GRAPH`/etc. Now rejected as a syntax error, matching the spec's negative tests.

## STRBEFORE / STRAFTER

- **Datatyping edge cases** (SPARQL 1.1 §17.4.3.7/8): a not-found search now returns a plain empty literal (previously it inherited arg1's language tag), a found match — including the empty search string — preserves arg1's tag, and an arg2 with an incompatible language tag now raises a type error per the argument-compatibility rules.
- **Language tags on literal constants in expression position** were dropped at lowering (`"x"@lang` became a bare `"x"`). They now lower to `STRLANG("x", "lang")` so the tag survives — needed for the compatibility checks above, and correct for language-aware expressions generally.

## Validation

- W3C SPARQL 1.1 eval suite: **+9 tests, zero regressions**.
- Full `fluree-db-sparql`, `fluree-db-query`, and `fluree-db-api` test suites pass; added unit tests covering the STRBEFORE/STRAFTER cases and the CONSTRUCT context expansion.
- `cargo fmt` + `clippy` clean on the touched crates.

### DIFF
--- a/fluree-db-query/src/eval/string.rs
+++ b/fluree-db-query/src/eval/string.rs
@@ -325,22 +325,71 @@ pub fn eval_concat<R: RowAccess>(
     Ok(Some(string_with_lang(&result, lang)))
 }
 
+/// Extract the `(string content, language tag)` of a STRBEFORE/STRAFTER argument.
+///
+/// Returns `None` when the argument is unbound or not a string-typed literal
+/// (e.g. a number or IRI) — both cases make the function raise a type error,
+/// which demotes to an unbound result. The language tag comes from a
+/// language-tagged `TypedLiteral` value (constants) or, for variable bindings
+/// that materialize as a plain string, from [`extract_lang_tag`].
+fn str_arg_and_lang<R: RowAccess>(
+    expr: &Expression,
+    value: Option<ComparableValue>,
+    row: &R,
+    ctx: Option<&ExecutionContext<'_>>,
+) -> Option<(String, Option<Arc<str>>)> {
+    match value? {
+        ComparableValue::String(s) => Some((s.to_string(), extract_lang_tag(expr, row, ctx))),
+        ComparableValue::TypedLiteral {
+            val: FlakeValue::String(s),
+            dtc,
+        } => {
+            let lang = match dtc {
+                Some(UnresolvedDatatypeConstraint::LangTag(tag)) => Some(tag),
+                // xsd:string (or any non-lang datatype) carries no language tag.
+                _ => extract_lang_tag(expr, row, ctx),
+            };
+            Some((s, lang))
+        }
+        _ => None,
+    }
+}
+
+/// SPARQL 1.1 §17.4.3.5 argument compatibility for STRBEFORE/STRAFTER.
+///
+/// The search string (`arg2`) is compatible when it is a simple literal /
+/// `xsd:string` (no language tag), or when it shares `arg1`'s language tag.
+/// An incompatible pair raises a type error (→ unbound).
+fn args_compatible(lang1: &Option<Arc<str>>, lang2: &Option<Arc<str>>) -> bool {
+    match lang2 {
+        None => true,
+        Some(l2) => lang1.as_deref() == Some(l2.as_ref()),
+    }
+}
+
 pub fn eval_str_before<R: RowAccess>(
     args: &[Expression],
     row: &R,
     ctx: Option<&ExecutionContext<'_>>,
 ) -> Result<Option<ComparableValue>> {
     check_arity(args, 2, "STRBEFORE")?;
-    let lang = extract_lang_tag(&args[0], row, ctx);
     let arg1 = args[0].eval_to_comparable(row, ctx)?;
     let arg2 = args[1].eval_to_comparable(row, ctx)?;
-    match (arg1, arg2) {
-        (Some(ComparableValue::String(s)), Some(ComparableValue::String(d))) => {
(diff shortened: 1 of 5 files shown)


fluree-db-query/src/eval/string.rs
```diff
@@ -325,22 +325,71 @@ pub fn eval_concat<R: RowAccess>(
...
+/// language-tagged `TypedLiteral` value (constants) or, for variable bindings
+/// that materialize as a plain string, from [`extract_lang_tag`].
+fn str_arg_and_lang<R: RowAccess>(
+    expr: &Expression,
+    value: Option<ComparableValue>,
+    row: &R,
+    ctx: Option<&ExecutionContext<'_>>,
+) -> Option<(String, Option<Arc<str>>)> {
+    match value? {
+        ComparableValue::String(s) => Some((s.to_string(), extract_lang_tag(expr, row, ctx))),
```
aaj3f: Really really minor perf note, but `STRBEFORE/STRAFTER` now will allocate two extra string copies per row where the old code had zero-copy-allocation patterns with `Arc<str>/&str`. Again, I think this is pretty trivial but if we wanted we could return `Arc<str>` and `Arc::clone(&s)` in the String arm and `Arc::from(s)` in the `TypedLiteral` arm

fluree-db-query/src/eval/string.rs
```diff
@@ -325,22 +325,71 @@ pub fn eval_concat<R: RowAccess>(
...
+        ComparableValue::TypedLiteral {
+            val: FlakeValue::String(s),
+            dtc,
+        } => {
+            let lang = match dtc {
+                Some(UnresolvedDatatypeConstraint::LangTag(tag)) => Some(tag),
+                // xsd:string (or any non-lang datatype) carries no language tag.
+                _ => extract_lang_tag(expr, row, ctx),
+            };
+            Some((s, lang))
```
aaj3f: Pasting as-is this claude-code-observed very-very-edge-case correctness note:
````
  arm accepts **any** string-valued typed literal as a valid argument, including a custom datatype (e.g. `STRDT("x","ex:dt")`, which materializes as `TypedLiteral{ val: String, dtc: Explicit }`): the `_ => extract_lang_tag(...)` fallback at line 350 treats it as a no-lang string instead of raising a type error. SPARQL 1.1 admits only simple literals, `xsd:string`, and lang-tagged strings; the old code (`_ => Ok(None)`) rejected non-string `TypedLiteral` args. Obscure (only reachable via STRDT/STRLANG-built non-`xsd:string` literals; ordinary variable-bound custom-typed values arrive as `ComparableValue::String` in both old and new code), but worth gating: in the `_` arm, accept only when `dtc` is absent or `xsd:string`, else return `None`.
````

File: fluree-db-query/src/eval/string.rs [hunk 3]
```diff
@@ -325,22 +325,71 @@ pub fn eval_concat<R: RowAccess>(
...
+///
+/// The search string (`arg2`) is compatible when it is a simple literal /
+/// `xsd:string` (no language tag), or when it shares `arg1`'s language tag.
+/// An incompatible pair raises a type error (→ unbound).
+fn args_compatible(lang1: &Option<Arc<str>>, lang2: &Option<Arc<str>>) -> bool {
+    match lang2 {
+        None => true,
+        Some(l2) => lang1.as_deref() == Some(l2.as_ref()),
+    }
+}
```
aaj3f, line 378: Pasting as-is this claude-code-observed very-very-edge-case correctness note:
````
`args_compatible` compares language tags case-sensitively (`lang1.as_deref() == Some(l2.as_ref())`). RDF 1.1 language tags are case-insensitive, so `@en` vs `@EN` would be wrongly deemed incompatible. Tags are normally lowercased on ingest and the new STRLANG-constant path passes them through verbatim, so this is an edge case; use `eq_ignore_ascii_case` for strict correctness.
````

aaj3f (approved):
@bplatz looks good. Very minor nit comments below to take-or-leave

fluree-db-query/src/eval/string.rs
bplatz: Addressed in 86aedda2b85027233995dcee4fd58245c5add244

fluree-db-query/src/eval/string.rs
bplatz: Addressed in 86aedda2b85027233995dcee4fd58245c5add244

fluree-db-query/src/eval/string.rs
bplatz: Addressed in 86aedda2b85027233995dcee4fd58245c5add244

PR #1383, opened 2026-06-26
fix: Body-ledger @annotation reification + update-clause annotation minting

bplatz:
## Summary

Fixes silent `@annotation` data-model corruption and unblocks minting annotations through update transactions. Three related issues, surfaced from a body-ledger bug report and a follow-on audit:

1. **Body-ledger annotation drop (reported bug).** Supplying the target ledger in the request **body** instead of the URL path silently dropped `@annotation` reification — the annotation expanded into a dangling literal predicate with no reification bundle and no error. The annotation-lowering pass had its own envelope detector (`is_envelope`) that only tolerated `@context`/`@graph`/`opts`, so a stray top-level `ledger` key made it misclassify the envelope as a data node and skip `@graph`.

2. **Reserved-key drift.** Four independent stages each hard-coded their own notion of "reserved top-level transactor key" and had diverged. Consolidated into a single canonical `RESERVED_TXN_KEYS` / `CLAUSE_KEYS` in `parse/mod.rs`, sourced by all four sites, with unit sync-guards that fail if any detector drifts. This also fixed a latent single-object-form leak (a body-form `ledger` leaking as a stray predicate).

3. **Update-clause annotations no-opped.** Inline `@annotation` in an UPDATE `insert` clause silently dropped the *entire* clause: synthesized `f:reifies*` sibling nodes were flushed onto the top-level wrapper, rewrapping `{where, insert, …}` into `{@graph: […]}` so `parse_update` then found no clauses. Fixed by scoping sibling attachment per clause.

## Changes

- `parse/mod.rs` — canonical `RESERVED_TXN_KEYS` / `CLAUSE_KEYS`.
- `parse/jsonld.rs` — `strip_opts_for_expansion` strips the full reserved set (fixes single-object leak).
- `parse/edge_annotations.rs` — `is_envelope` / `is_transaction_wrapper` source the shared set; `attach_siblings` scopes synthesized siblings per clause and lifts reserved keys out of the rewrapped node.
- Docs: worked update-mint example (concepts) + per-clause sibling invariant (design).

## Coverage

An audit of the annotation surface (operation × surface × value-type × identity × cardinality) found **no functional holes** in the JSON-LD add/update/delete lifecycle. New tests pin the high-value, previously-unpinned capabilities:

- body-ledger annotation: envelope + single-object forms (no stray routing-key predicate)
- update insert clause: constant, WHERE-bound, anonymous, explicit-`@id`, parallel, literal/typed
- editing an **anonymous** annotation's metadata via inline `@annotation` WHERE binding
- selector delete of an anonymous annotation
- unit sync-guards for the reserved-key set

## Scope boundaries (pre-existing, unchanged)

- Inline named-graph annotation **queries** remain outside the read-side correctness envelope (the write works).
- SPARQL UPDATE annotations are default-graph only; `@reifies` minting is deferred; `upsert` is not a `parse_update` clause.

## Validation

`fluree-db-transact` (263) · `grp_misc` edge annotations (78) · `grp_transact` (159) · `grp_query_sparql` (233) all green. Clippy clean on changed targets; `cargo fmt --all` applied.


fluree-db-transact/src/parse/jsonld.rs
aaj3f, line 920: This is more of a question than a suggestion. Since this strip runs on the raw JSON before expansion, if a single-object insert has `@context` and maps one of those bare to-strip names to a real predicate IRI, I think maybe that predicate is now stripped w/ no warning/error. I think this would be a very rare occurrence window, and it's consistent with keeping them as "reserved keys", but it may still produce end-user surprising results when it occurs. Not sure if the solution is to avoid stripping keys present in such a `@context` or to more forcefully error earlier to prevent users from using "reserved keys" as `@context` terms in such a way that this could even be possible

aaj3f (approved):
This looks great @bplatz (and reminds me of `@annotation` generally which I'd kind of forgotten about but would like to use more). One very small but non-blocking nit/question below

fluree-db-transact/src/parse/jsonld.rs
bplatz: Addressed in 1fd1accf5a493bddcb8c01ade75e92fff49a05b1

PR #1384, opened 2026-06-26
fix: speed up fluree-memory adds and stop them hanging under concurrency

bplatz:
## Problem

`memory_add` (CLI and MCP) re-read every memory from the `__memory` ledger via a SPARQL query (`all_memories_for_scope`) and rewrote the whole `repo.ttl` on each call. Under concurrent MCP requests — e.g. an agent saving 2–3 memories at once, which the client dispatches in parallel — that O(N) work serialized behind the per-process mutation lock and made adds take seconds (~13.5s at 100 memories), which users experienced as hangs. A separately-observed failure mode: a wedged operation could hold a lock forever (a CPU-spun MCP server froze every later call in that process).

## Changes

- **Fast add via sorted file splice.** `add` now inserts the new block at its sorted `(branch, id)` position by reading only the `.ttl` text (`turtle_io::insert_memory_into_file`), never the ledger. Untouched blocks stay byte-identical (asserted equal to a full rewrite), so the on-disk result and merge behavior are unchanged. The add itself drops to ~2ms; the per-add ledger round-trip is gone.
- **Merge behavior preserved and safe.** The file keeps its `(branch, id)` sort under a default git merge: distinct-branch adds auto-merge, while two branches changing the *same* memory correctly conflict — no silent corruption.
- **Bounded locks.** The cross-process flock now polls `try_lock` to a deadline, and the in-process mutation lock uses a timeout, so a wedged operation surfaces a recoverable error instead of freezing the session. Shared `file_sync::lock_timeout` (30s default, override with `FLUREE_MEMORY_LOCK_TIMEOUT_SECS`).
- Memory docs updated to match.

## Tests

- `it_merge`: distinct-region adds merge cleanly; same-memory edits conflict.
- `turtle_io`: incremental splice equals a full sorted rewrite byte-for-byte.
- `file_sync`: the rebuild lock times out when already held.

### DIFF
--- a/fluree-db-memory/src/file_sync.rs
+++ b/fluree-db-memory/src/file_sync.rs
@@ -440,6 +440,11 @@ mod tests {
         let dir = tempfile::tempdir().unwrap();
         std::fs::create_dir_all(dir.path().join(".local")).unwrap();
         // Keep the timeout short so the test is fast.
+        // NOTE: this mutates the process-global env that `lock_timeout()` reads.
+        // It is safe only because no sibling test in this binary reads
+        // `lock_timeout()`; if one is added, this test must serialize against it
+        // (e.g. a shared mutex) or scope the var, since nextest runs siblings on
+        // parallel threads in the same process.
         std::env::set_var("FLUREE_MEMORY_LOCK_TIMEOUT_SECS", "1");
 
         // First acquisition succeeds and holds the lock.
--- a/fluree-db-memory/src/file_sync.rs
+++ b/fluree-db-memory/src/file_sync.rs
@@ -190,6 +190,7 @@ pub(crate) async fn acquire_memory_lock(memory_dir: &Path) -> Result<fs::File> {
 }
 
 async fn acquire_lock_file(lock_path: PathBuf) -> Result<fs::File> {
+    let timeout = lock_timeout();
     tokio::task::spawn_blocking(move || {
         if let Some(parent) = lock_path.parent() {
             fs::create_dir_all(parent)?;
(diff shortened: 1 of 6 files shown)


fluree-db-memory/src/turtle_io.rs
```diff
@@ -491,6 +582,46 @@ mod tests {
...
+        c.content = "memory C".to_string();
+        c.tags = vec![];
+        c.artifact_refs = vec![];
+        c.branch = Some("zzz".to_string()); // sorts after "main"
+
+        // Insert in an order that exercises front/middle/end placement.
+        insert_memory_into_file(&path, &a, REPO_HEADER).unwrap();
+        insert_memory_into_file(&path, &c, REPO_HEADER).unwrap();
+        insert_memory_into_file(&path, &b, REPO_HEADER).unwrap();
+        let spliced = fs::read_to_string(&path).unwrap();
```
aaj3f: Super minor but I don't know that the `front/middle/end placement` comment is 100% true. The three inserts never splice a new block strictly between two pre-existing blocks. Maybe worth a test case that seeds `[X, Z]` then inserting a `Y` that would sort between them.

fluree-db-memory/src/file_sync.rs
```diff
@@ -397,6 +435,35 @@ mod hex {
...
+            "second acquisition should time out while the lock is held"
+        );
+        assert!(
+            elapsed < std::time::Duration::from_secs(10),
+            "timeout should fire promptly (took {elapsed:?})"
+        );
+
+        drop(held);
+        std::env::remove_var("FLUREE_MEMORY_LOCK_TIMEOUT_SECS");
+    }
```
aaj3f, line 470: This is harmless today, but this test mutates the process-global env for `FLUREE_MEMORY_LOCK_TIMEOUT_SECS` which `lock_timeout()` reads. No other sibling test reads `lock_timeout()`, but if any other test ended up reading it, we'd have possible issues. Probably just leave it for now (unless maybe you want to add a comment to guard against that future scenario)

aaj3f (approved):
Nice fix, @bplatz, for a minor but annoying issue I'd observed myself multiple times.

fluree-db-memory/src/turtle_io.rs
bplatz: Addressed in b9c60e7aebb291d4388f5a7b9cb94a269a4d961d — added insert_between_existing_blocks (seed [X, Z], splice Y between, assert == full rewrite) and corrected the comment. You were right: the old insert order only hit front/end, never the middle branch.

fluree-db-memory/src/file_sync.rs
bplatz: Addressed in b9c60e7aebb291d4388f5a7b9cb94a269a4d961d — kept it as-is (no sibling reads lock_timeout()) and added a comment documenting the constraint and what a future reader would need to do (serialize/scope), since nextest runs siblings on parallel threads in the same process.

bplatz:
Addressed in b9c60e7aebb291d4388f5a7b9cb94a269a4d961d

PR #1385, opened 2026-06-27
fix: db() reflects owned-state commits (read-your-writes)

### LINKED ISSUE #1330: fluree.db() can serve a stale cached view after subsequent commits (read-your-writes violation)
## Problem

Calling `fluree.db(ledger_id)` after committing additional transactions can return a previously cached `GraphDb` view that predates those commits. Observed twice while writing integration tests, in both ledger-manager configurations:

1. `FlureeBuilder::memory().with_ledger_cache_config(LedgerManagerConfig::default())` + background indexer: after `trigger_index_and_wait_outcome`, two further commits (verified `receipt.ledger.t() == 3`), then `fluree.db(ledger_id)` returned a view with `snapshot.t == 1` — neither commit visible (`it_cyclic_bgp_probe.rs`, phase 2).
2. Plain `FlureeBuilder::memory().build_memory()` + synchronous `fluree.reindex(...)`: the *second* `fluree.db()` call in a test (after an earlier call had populated the cache, followed by more commits) served the pre-commit view; even a post-reindex `fluree.db()` reported `t = 1` while the data was at `t = 5` (`it_join_batched_overlay.rs::batched_join_decimal_novelty`).

Both tests work around it by pinning explicitly: `fluree.db_at_t(ledger_id, receipt.ledger.t())` — the workaround sites are marked with comments ("a cached pre-commit view can otherwise serve").

## Why it matters beyond tests

If the same lookup path backs server-side view resolution, a client that commits and immediately queries can read stale state — a read-your-writes violation that would be intermittent and hard to diagnose in production. If the embedded-API path is the only one affected, the fix may just be cache invalidation (or a freshness check against the nameservice head) in the ledger manager's view lookup.

## Repro

See the two test files above: drop the `db_at_t` pinning back to `fluree.db(ledger_id)` and both tests fail with pre-commit data.

bplatz:
## Problem

Closes #1330

Calling `fluree.db(ledger_id)` after committing through the owned-state API could return a previously cached `GraphDb` view that predates those commits — a read-your-writes violation.

The owned-state API (`insert` / `update` / `transact(ledger, ...)` and the `transact(...).execute()` builder) consumes a caller-supplied `LedgerState`, commits, and publishes to the nameservice, then returns a *new* `LedgerState` — but it never wrote that state back into the `LedgerManager` cache. A subsequent `db()` / `ledger_cached()` on a warm cache therefore served the pre-commit view.

The guarded builder path (`fluree.stage(&handle).execute()`) was never affected: it holds the ledger write lock across stage + commit and replaces the cached state via `finalize_commit`. The server transaction path (`LocalCommitter`) uses that guarded path, so server-side read-your-writes was already correct — this only affected the embedded owned-state API.

## Fix

- Add `Fluree::finalize_owned_commit`, a shared post-commit tail for the owned-state methods that writes the committed state back into the cache (via `sync_owned_commit_to_cache`) before computing the indexing status and triggering background indexing. This also consolidates six duplicated post-commit tails into one.
- `sync_owned_commit_to_cache` is monotonic (skips when the cache has advanced past the new state) and only touches an already-cached handle — an uncached ledger loads fresh from the nameservice head on next access. On a `refresh_index` failure it evicts the entry so the next access reloads rather than serving stale state.
- Add `LedgerManager::get_loaded_handle` to fetch a cached handle without forcing a load.

## Tests

- Drop the `db_at_t` workarounds in `it_cyclic_bgp_probe` and `it_join_batched_overlay` (the sites the issue called out).
- Add `it_read_your_writes` as a direct regression test.
- `it_notify_incremental`: in-process commits no longer leave the writer's own cache stale, so these now create a genuine cross-writer gap via a peer `LedgerManager` (the realistic scenario `notify` catch-up exists for).
- `it_query_sparql_indexed`: head-with-novelty views now take the production overlay fast path, which renders a decimal `AVG` as a precision-preserving string.

## Verification

- Full `fluree-db-api` suite (2253 tests) + `fluree-db-server` / `fluree-db-consensus` tests pass.
- `cargo fmt --all --check` and `cargo clippy --all --all-features --all-targets -- -D warnings` clean.

## Docs

Updated `docs/getting-started/rust-api.md` and `docs/concepts/time-travel.md` to state that a writer reads its own writes in-process, with eventual consistency remaining a cross-process concern.

### DIFF
--- a/fluree-db-api/tests/it_cyclic_bgp_probe.rs
+++ b/fluree-db-api/tests/it_cyclic_bgp_probe.rs
@@ -18,7 +18,7 @@ use fluree_db_api::{FlureeBuilder, IndexConfig, LedgerManagerConfig, QueryInput}
 use fluree_db_transact::{CommitOpts, TxnOpts};
 use serde_json::json;
 use support::{
-    genesis_ledger_for_fluree, normalize_rows, start_background_indexer_local,
+    genesis_ledger_for_fluree, normalize_rows, span_capture, start_background_indexer_local,
     trigger_index_and_wait_outcome,
 };
 
@@ -131,6 +131,13 @@ async fn cyclic_bgp_bounded_probe_matches_fallback() {
                 ),
             ];
 
+            // Track whether the bounded probe actually engaged on at least one
+            // query. At the indexed HEAD with no novelty the overlay lane is
+            // Clean and probing is allowed, so a probe-eligible query must emit
+            // the per-subject probe event. Without this, every assertion below
+            // is result-equivalence against the fallback — which a silently
+            // always-declining probe would still satisfy.
+            let mut probe_engaged = false;
             for (name, query) in queries {
                 // 1. Fallback operator tree = ground truth.
                 clear_cyclic_env();
@@ -141,8 +148,11 @@ async fn cyclic_bgp_bounded_probe_matches_fallback() {
                 // 2. Cascade with probing forced on (every gate-passing edge probes).
                 clear_cyclic_env();
                 std::env::set_var("FLUREE_CYCLIC_BGP_PROBE_SCAN_RATIO", "1");
+                let (spans, guard) = span_capture::init_test_tracing();
                 let probed = run_query(&fluree, &view, query).await;
+                drop(guard);
                 assert_eq!(probed, expected, "{name}: probed cascade != fallback");
+                probe_engaged |= spans.has_event("cyclic cascade: probing edge per-subject");
 
                 // 3. Cascade with probing forced off (full-scan semi-join path).
                 clear_cyclic_env();
@@ -150,6 +160,11 @@ async fn cyclic_bgp_bounded_probe_matches_fallback() {
                 let scanned = run_query(&fluree, &view, query).await;
                 assert_eq!(scanned, expected, "{name}: full-scan cascade != fallback");
             }
+            assert!(
+                probe_engaged,
+                "bounded probe never engaged on the indexed HEAD view; \
+                 probe-activation coverage is gone"
+            );
 
             // Phase 2: novelty tail — retract triangle 2's closing edge and
             // assert a triangle that exists only in novelty. Probes must merge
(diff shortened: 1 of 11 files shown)


fluree-db-api/tests/it_cyclic_bgp_probe.rs
```diff
@@ -195,18 +189,16 @@ async fn cyclic_bgp_bounded_probe_matches_fallback() {
...
                 );
                 if name == "directed-triangle" {
-                    assert!(
-                        spans.has_event("cyclic cascade: probing edge per-subject"),
-                        "{name}: bounded probes should engage under novelty"
-                    );
+                    // At head the view is the binary index plus a separate
+                    // novelty overlay; the bounded probe declines on that fast
+                    // path (the overlay-merging fallback serves these rows), so
+                    // we assert the merged result rather than probe engagement.
```
aaj3f: This is a claude-code-discovered item re: test coverage and I'm just going to paste-in as-is:
````
  — The switch from `db_at_t(novelty_t)` to `db()` is correct for the
  read-your-writes story, but it changes which lane serves the query: at head the
  view is the binary index plus a separate novelty overlay, so the bounded probe
  declines and the overlay-merging fallback serves the rows. To accommodate this
  the test **dropped its only assertion that the bounded cyclic probe actually
  engages** (the former `spans.has_event("cyclic cascade: probing edge
  per-subject")` check). Verified repo-wide there is no other coverage of probe
  activation: the event is emitted only at `fluree-db-query/src/cyclic_bgp.rs:1340`
  and `:1442`, and `span_capture`/`has_event` no longer appear anywhere in this
  test file. Bounded-probe activation in `cyclic_bgp.rs` is now uncovered;
  only result-equivalence remains. Suggest retaining the probe-engagement
  assertion on a separate case that still engages the probe (e.g. a time-pinned /
  merged view) while keeping the new `db()` result comparison.
````

aaj3f (approved):
@bplatz seems to be a very good / correct fix for #1330. Only one note below that's only test-coverage related

bplatz:
Addressed in 21ce914d53a55009f78119dc69b75022081c48ff

Note on approach: I restored the probe-engagement assertion in Phase 1 (indexed HEAD, no novelty) rather than on a time-pinned/merged view as suggested — post-fix that view no longer engages the probe. `compute_probe_overlay_state` disables bounded probes off-HEAD (`ctx.to_t < store.max_t()`), and now that `db()` correctly reflects the commit, `db_at_t(novelty_t)` returns the same head-with-separate-overlay view where the probe declines. At the clean indexed HEAD the overlay lane is `Clean` and probing is allowed, so a probe-eligible query emits the per-subject event; the test now captures spans across the probed lane and asserts the bounded probe engaged at least once (fails if it silently stops).

PR #1386, opened 2026-06-27
fix: consistent xsd:decimal/xsd:integer datatype across index and novelty (#1329)

### LINKED ISSUE #1329: Inconsistent JSON-LD rendering of xsd:decimal values: empty @type from arena decode vs plain string from flake path
## Problem

The same `xsd:decimal` value renders differently in query results depending on which internal path served it:

- Arena-decoded (binary index / NumBig): `{"@value": "19.99", "@type": ""}` — note the **empty `@type`**
- Raw-flake-served (novelty raw-merge path): plain `"19.99"` string

Observed while writing the overlay differential test `it_join_batched_overlay.rs::batched_join_decimal_novelty`, where the same query returns mixed forms when part of the data is in the index and part in novelty. The test works around it by comparing extracted `@value`s instead of exact JSON (see the `decimal_values` helper and the comment marking this as a pre-existing formatter discrepancy).

Repro shape: index a typed decimal, commit another via novelty, `SELECT` both — the indexed one comes back as an object with an empty `@type`, the novelty one as a bare string.

## Expected

One canonical rendering for `xsd:decimal` regardless of serving path — and in no case an empty `@type` string, which is malformed JSON-LD.

Likely area: the formatter's handling of `NUM_BIG_OVERFLOW`-decoded values fails to map the o_type back to the `xsd:decimal` datatype IRI (hence `@type: ""`), while the flake path drops the type annotation entirely.

bplatz:
## Summary

Fixes #1329: an `xsd:decimal` served from the binary index rendered with an empty `@type` (`{"@value":"19.99","@type":""}`) while the same value in novelty rendered as a bare string. Also fixes a sibling mislabel where indexed `> i64` integers came back as `xsd:decimal`.

## Root cause

`OType::NUM_BIG_OVERFLOW` names no single datatype from `o_type` alone — its arena holds **both** overflow `xsd:integer` (BigInt) **and** `xsd:decimal` (BigDecimal). So `resolve_datatype_sid(o_type)` returns `None`, and every flake/binding builder fell back to `Sid::new(0, "")` → empty `@type`. Novelty copies still carry `o_type = XSD_DECIMAL`, so they resolved correctly — hence the index-vs-novelty asymmetry.

A second, related issue: the late-materialized `EncodedLit` for the NumBig arena hardcodes `dt_id = DECIMAL` (it has no decoded value at creation time), so an indexed overflow integer materialized as `xsd:decimal`.

## Fix

- New `BinaryIndexStore::resolve_datatype_sid_for_value(o_type, &val)` derives the datatype from the decoded `FlakeValue` (`Decimal`→`xsd:decimal`, `BigInt`→`xsd:integer`) when `o_type` can't name it. Wired into every flake/binding builder that has the decoded value in scope: `binary_range` (x2), `binary_history`, `object_binding`, `fast_group_count_firsts` (x2), `block_fetch`.
- `binary_scan::matches_datatype_constraint` is a **pre-decode** filter — it dropped indexed big numerics from datatype-constrained patterns (e.g. `{"@value":"?p","@type":"xsd:decimal"}`) while novelty copies passed. It now decodes to disambiguate **only** in the rare ambiguous case, keeping the common path decode-free.
- New `FlakeValue::overflow_numeric_datatype_sid()` recovers the datatype at materialization time (the EncodedLit can't be fixed at creation). Consulted before the `dt_id` lookup in both the query materializer and the API formatter; `resolve_datatype_sid_for_value` delegates to it too.

Because `xsd:decimal` is an inferable datatype, the consistent output is the bare-string form across crawl and tabular lanes (matching the existing convention). `export.rs` already had per-variant fallbacks, so it needed no change.

## Tests

- `jsonld_decimal_renders_consistently_across_index_and_novelty` — crawl + tabular SPARQL + datatype-constrained pattern, asserting the indexed and novelty decimals render identically and both match a `xsd:decimal`-constrained query.
- `indexed_overflow_integer_reports_xsd_integer_not_decimal` — pins the late-materialization datatype recovery.

Both confirmed to fail before the fix and pass after. Full `fluree-db-api` suite, `fluree-db-query`/`fluree-db-core` unit tests, and clippy (all-features) are green.

### DIFF
--- a/fluree-db-api/src/block_fetch.rs
+++ b/fluree-db-api/src/block_fetch.rs
@@ -299,7 +299,7 @@ pub fn decode_leaf_block(
                 );
             }
 
-            let dt = match store.resolve_datatype_sid(o_type) {
+            let dt = match store.resolve_datatype_sid_for_value(o_type, &o) {
                 None => fluree_db_core::Sid::new(0, ""),
                 Some(sid) => {
                     let iri = store.sid_to_iri(&sid).ok_or_else(|| {
--- a/fluree-db-api/src/format/materialize.rs
+++ b/fluree-db-api/src/format/materialize.rs
@@ -78,11 +78,16 @@ fn materialize_encoded_lit(binding: &Binding, gv: &BinaryGraphView) -> std::io::
     match val {
         FlakeValue::Ref(sid) => Ok(Binding::sid(sid)),
         other => {
-            let dt_sid = store
-                .dt_sids()
-                .get(*dt_id as usize)
-                .cloned()
-                .unwrap_or_else(|| Sid::new(0, ""));
+            // NUM_BIG arena values share one EncodedLit whose dt_id is hardcoded
+            // to decimal — recover xsd:integer vs xsd:decimal from the decoded
+            // value, not dt_id (issue #1329).
+            let dt_sid = other.overflow_numeric_datatype_sid().unwrap_or_else(|| {
+                store
+                    .dt_sids()
+                    .get(*dt_id as usize)
+                    .cloned()
+                    .unwrap_or_else(|| Sid::new(0, ""))
+            });
             let meta = store.decode_meta(*lang_id, *i_val);
             let dtc = match meta.and_then(|m| m.lang.map(std::sync::Arc::from)) {
                 Some(lang) => DatatypeConstraint::LangTag(lang),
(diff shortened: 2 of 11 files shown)


aaj3f (approved):
LGTM! 🚀

PR #1387, opened 2026-06-29
Enable direct-local CLI for SPARQL UPDATE and NDJSON time-travel / policy

zonotope:
Three feature gaps between the embedded (direct/local) CLI and the server-routed CLI are closed. The underlying `fluree-db-api` already supported each capability in process; the CLI was bailing out and telling users to start a server. The branch wires the existing capabilities through, and adds one shared piece of plumbing on the API side so SPARQL UPDATE works from both terminal builder paths.

### 1. `fluree update` with SPARQL UPDATE in direct local mode

Previously rejected with *"SPARQL UPDATE is not supported in direct local mode."* `RefTransactBuilder` already implemented SPARQL UPDATE end-to-end; `GraphTransactBuilder` (which the CLI's `Local` arm drives) did not expose it. Added `GraphTransactBuilder::sparql_update(...)` so `fluree.graph(id).transact().sparql_update(sparql).commit()` works. The CLI's `update` command now calls into it directly — same write-lock-held parse-and-lower pipeline the HTTP server uses internally.

### 2. `GraphTransactBuilder::stage()` with SPARQL UPDATE

`stage()` previously executed `let op = self.core.operation.unwrap();` unconditionally, which would panic on a `pending_sparql` core. Reworked to check `pending_sparql` first and use the same parse → namespace-registry → `lower_sparql_update_ast` → `stage_transaction_from_txn` pipeline that `commit()` uses, so `fluree.graph(id).transact().sparql_update(sparql).stage()` now returns a queryable `StagedGraph` for preview. Policy and tracking flow through.

### 3. `fluree query --format ndjson` with `--at` and `--policy*` in direct local mode

Previously rejected with *"time travel (--at) is not supported with --format ndjson on local ledgers"* and the analogous policy error. Routed the local NDJSON streaming path through `run_stream_query_dataset` (the dataset producer the server uses for the same options), mirroring `/v1/fluree/stream/query/<ledger>`'s ledger-scoped routing: plain head queries with no policy keep the lean single-ledger `run_stream_query` producer; `--at` and/or `--policy*` route through the dataset producer. Same NDJSON wire protocol either way (`head` → `row`* → terminal `end`/`error`).

## Implementation notes

- **SPARQL UPDATE parse + lower helper.** Both `RefTransactBuilder::execute()` and the new `GraphTransactBuilder::stage()` SPARQL branch share `tx_builder::parse_and_lower_sparql_update(sparql, snapshot, txn_opts) -> Result<Txn>`. Identical error mapping (HTTP 400 with structured messages) lives in one place.
- **CLI policy injection consolidation.** Both `multi_query` and the new NDJSON dataset path need to fold `--policy*` flags into a JSON-LD body's `opts` map with a default-vs-override rule (explicit body opts always win over CLI flags). Hoisted into `PolicyArgs::inject_into_opts(opts: &mut Map<String, Value>) -> Result<(), String>`; the two call sites are now thin adapters that extract / restore the opts map and delegate.
- **`update` command Local branch dedupe.** The `SparqlUpdate` and `JsonLd` arms shared ~20 lines of identical policy attachment, builder construction, commit, post-commit reporting, and novelty warning. Collapsed into one path that picks the builder method by format and shares everything else.
- **SPARQL `--at` + inline FROM.** Rejected with the same *"encode time travel in the FROM IRI (e.g., FROM <ledger@t:1>)"* guidance the remote path already gave — the dataset spec we'd build would conflict with in-query clauses.

## Tests

`fluree-db-cli/tests/integration.rs`:

- Flipped `update_sparql_update_in_direct_mode_fails` → `update_sparql_update_in_direct_mode_commits` — `INSERT DATA` commits at t=1 in direct local mode.
- Flipped `query_ndjson_local_rejects_time_travel` → `query_ndjson_local_time_travel_streams_historical_rows` — SPARQL `--at 1` streams Alice/Bob.
- Added `query_ndjson_local_time_travel_jsonld_streams_historical_rows` — JSON-LD parity for `--at`.
- Added `query_ndjson_local_sparql_at_rejects_inline_from` — preserves the FROM + `--at` error.
- Added `query_ndjson_local_with_default_allow_policy_streams` — exercises the policy-driven dataset routing path with the cheapest `--policy*` flag.


bplatz:
**Default `@context` is dropped on the new `--at` / `--policy` ndjson path.**

The dataset path builds its view via `load_view_from_source` → `self.db()` / `self.db_at()` (`dataset_builder.rs:204/217`), neither of which attaches the ledger's stored default context. So `primary.default_context` is `None` and prefix injection is skipped — whereas the lean ndjson path uses `with_default_context(...)` and the buffered path uses `db_with_default_context`.

Result: a query relying on the ledger's stored context resolves prefixes everywhere except under `--format ndjson --at` / `--policy`, where it silently returns empty/wrong rows. The existing tests miss it because they use explicit `@context` / full IRIs.

Suggest routing the streaming dataset's single source through the `*_with_default_context` loaders (or attaching the ledger default context to the built view) so all three CLI query paths behave identically.

bplatz (approved):
Just the one comment, looks good!

PR #1390, opened 2026-06-29
Fix flaky raft muliti-node test

zonotope:
Fixes the `liveness_monitor_demotes_killed_follower` test in `fluree-db-server/tests/raft_multi_node.rs`, which flaked at ~25–37% under `cargo nextest` parallelism on CI. The underlying bug is in `LivenessMonitor::tick`, not the test: under the bootstrap sequence (`initialize` → `add-learner` → `change-membership`), the monitor was seeding peer trackers for learners before they were promoted to voters, leaving each seeded tracker in a state where the hysteresis check permanently blocked any future Demote propose for that peer.

## Root cause

`LivenessMonitor::tick` iterated openraft's `metrics.replication` map to build per-peer trackers. That map includes **learners** as well as voters — a node added via `add-learner` shows up there from the first append-entries it receives. At that moment the learner is not in `worker_eligible_voters` (only voters are), so `or_insert_with` seeded the tracker via `PeerTracker::for_ineligible_peer()`, which sets `last_proposed = Some(Demote)` to support the cross-leader-transition promote path.

When `change-membership` later promoted the learner to a voter, the state-machine apply atomically added it to both `configured_voters` and `worker_eligible_voters` — but the monitor's existing tracker was untouched. The stale `last_proposed = Some(Demote)` then short-circuited the Demote arm of `next_eligibility_proposal` forever: when the voter was later killed, the monitor correctly observed lag (`unreachable_since` set, `unreachable_after` elapsed) but the `tracker.last_proposed != Some(Demote)` guard returned `None`, and `worker_eligible_voters` never lost the dead voter.

Local repro under parallel load showed the leader stable at `last_log_index = 12`, peer 2 stuck at match index 7 — five entries of unambiguous "leader ahead of peer" lag — with the monitor proposing nothing. Sequential runs passed reliably because the bootstrap window closed before the test killed the follower, and the monitor's first tick caught a steady-state membership; only the contention-induced delay between `add-learner` and `change-membership` made the bug observable.

## Fix

`fluree-db-consensus/src/raft/liveness_monitor.rs`:

- Snapshot `configured_voters` from the replicated state machine alongside `worker_eligible_voters` in the same `shared_state.read().await`. Both sets are updated atomically by the membership-apply in `state_machine_adapter.rs`, so the snapshot can never see a voter present in one but not the other.
- Skip any peer not in `configured_voters` before creating its tracker. Learners are not workers (`apply_set_worker_eligibility` refuses any voter outside `configured_voters` anyway), so there's no reason to track them — and skipping them is what prevents the pre-promotion `for_ineligible_peer` seeding.
- Drop the `last_membership_log_id` caching that previously gated the tracker-pruning `retain`. The optimization predates pulling `configured_voters` from state every tick; with the state read already happening for `eligible_voters`, the additional `BTreeSet<NodeId>` clone is free at realistic cluster sizes. The `retain` now runs every tick directly against the per-tick voter snapshot.

`fluree-db-server/tests/raft_multi_node.rs`:

- Add `leader_metrics_snapshot` and include its output (`current_leader`, `last_log_index`, per-peer match indices) in `wait_for_voter_demoted`'s timeout panic. The pre-fix flake message was `last={1, 2, 3, 4, 5}` — which says the demote didn't land but doesn't say why. The new message attaches the openraft metrics the monitor reads, so a future regression points straight at whether the leader was advancing, which peers were stuck, and at what index.

## Verification

- `raft_multi_node` test binary, 20 consecutive parallel runs under `cargo test` default parallelism: **20/20 pass**. Pre-fix baseline on the same machine flaked 25–37% on the same workload.
- 5 consecutive sequential runs (`--test-threads=1`): 5/5 pass.
- `fluree-db-consensus` lib tests: 345 passed, 0 failed.
- `liveness_monitor` unit tests (the tracker-level logic): 12/12 pass — no behavior change at that layer.
- `raft_integration` tests: 5/5 pass.
- `cargo fmt --check` and `cargo clippy --all-targets` clean on both touched crates.

### DIFF
--- a/fluree-db-consensus/src/raft/liveness_monitor.rs
+++ b/fluree-db-consensus/src/raft/liveness_monitor.rs
@@ -201,24 +201,13 @@ impl LivenessMonitor {
     /// them on abort.
     pub async fn run(self) {
         let mut trackers: HashMap<NodeId, PeerTracker> = HashMap::new();
-        // Cache the log id of the most recently observed membership
-        // change. When it doesn't advance between ticks, the voter
-        // set is unchanged and the tracker-pruning collect+retain
-        // pair can be skipped.
-        let mut last_membership_log_id: Option<LogId<NodeId>> = None;
         loop {
             tokio::time::sleep(self.config.sample_interval).await;
-            self.tick(&mut trackers, &mut last_membership_log_id, Instant::now())
-                .await;
+            self.tick(&mut trackers, Instant::now()).await;
         }
     }
 
-    async fn tick(
-        &self,
-        trackers: &mut HashMap<NodeId, PeerTracker>,
-        last_membership_log_id: &mut Option<LogId<NodeId>>,
-        now: Instant,
-    ) {
+    async fn tick(&self, trackers: &mut HashMap<NodeId, PeerTracker>, now: Instant) {
         // Read what we need from the metrics ref and drop it
         // before any await. Cloning the whole `RaftMetrics` (the
         // earlier shape) deep-copied the per-peer replication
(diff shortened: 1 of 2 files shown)


bplatz (approved):
👍

PR #1393, opened 2026-06-30
docs: Surface Cypher in overviews + minor repo cleanup

bplatz:
Small docs + housekeeping pass.

**Docs:** Cypher (openCypher) is a supported read/write surface alongside JSON-LD Query and SPARQL, but several overview/reference docs still enumerated only "JSON-LD and SPARQL". Updated the query, transaction, getting-started, and CLI/API docs to name all three languages and document the previously-missing CLI (`--cypher` / `--format cypher`) and HTTP (`Content-Type: application/cypher`) entry points.

**Cleanup:**
- Removed `scripts/import-memory-repro.sh` (one-off debug repro) and its dangling doc-comment reference
- Untracked `.cursor/` editor config and added it to `.gitignore` (kept locally)


docs/getting-started/quickstart-query.md
```diff
@@ -220,6 +221,23 @@ curl -X POST http://localhost:8090/v1/fluree/query \
   '
 ```
 
+## Cypher
+
+Cypher targets a ledger directly (no `FROM`/dataset clause) — send the statement
+with `Content-Type: application/cypher` to the ledger-scoped query endpoint:
+
+```bash
+curl -X POST http://localhost:8090/v1/fluree/query/mydb/main \
```
aaj3f: Should this be `mydb:main`?

aaj3f (approved):
👍

docs/getting-started/quickstart-query.md
bplatz: Addressed in 463c548623b5bdc38d4f8257de1d7ea479de3a68

PR #1399, opened 2026-06-30
fix(turtle): resolve RFC 3986 §5.3 fragment refs against @base

### LINKED ISSUE #1395: Turtle parser drops the fragment resolving `<#name>` against `@base`, collapsing multi-TriplesMap R2RML mappings to one table
## Summary

A multi-`TriplesMap` R2RML mapping written in the idiomatic `@base` + `<#Name>`
style **collapses into a single TriplesMap / single logical table**. Registering a
16-TriplesMap star-schema mapping reports `TriplesMaps: 1`, and queries scan only
the first table (with a union projection of *every* column across all maps);
classes from non-first TriplesMaps return 0 rows.

**The root cause is a Turtle-parser RFC 3986 §5.3 violation, not an R2RML
limitation.** Multi-table R2RML — per-table scans and cross-table `rr:refObjectMap`
joins — is already fully implemented and covered by passing offline tests. This is
a small, surgical fix.

## Root cause

`fluree-graph-turtle/src/parser.rs::resolve_iri` **drops the fragment** when
resolving a relative reference against an `@base`:
- the `#`-only branch (~`parser.rs:919`) discards the reference fragment (returns `None`);
- the result assembly (~`parser.rs:945-957`) never appends `#<fragment>`;
- helpers `parse_iri_components` (~`:1010`) return a 4-tuple with no fragment, and `split_path_query` (~`:1043`) strips everything after `#`.

So `<#DimDate>`, `<#DimProduct>`, … **all resolve to the same IRI** (the base with
its fragment stripped), violating RFC 3986 §5.3 (a same-document fragment ref must
resolve to base-minus-its-fragment **+ `#` + the reference's fragment**).

## How the collapse produces the exact symptoms

Once all 16 `a rr:TriplesMap` triples share one subject IRI:
1. `loader/extractor.rs:42-58` extracts the same IRI 16×; `CompiledR2rmlMapping::add_triples_map` keys by IRI → **1 entry** → `triples_map_count = 1`.
2. The merged subject carries all 16 `rr:logicalTable` / `rr:subjectMap` / `rr:predicateObjectMap` edges.
3. `extract_logical_table` / `extract_subject_map` use **first-wins** (`extractor.rs:84,126`) → first table `DW.DIM_DATE`, first class `edw:Date` (so `?p a edw:Product` matches nothing → 0 rows).
4. POM extraction uses **all** (`extractor.rs:179`) → **union of ~130 columns** → the all-columns projection scan of `DW.DIM_DATE` seen in the logs.

## Why existing tests miss it

Every R2RML fixture uses **absolute** TriplesMap IRIs (e.g.
`<http://example.org/mapping#AirlineMapping>`), which short-circuit IRI resolution
entirely. The base-resolution unit tests cover relative *paths* and the empty ref
`<>`, but **none test a `<#fragment>` ref** — the exact idiomatic R2RML style
(`@base` + `<#Name>`, as the W3C R2RML spec itself writes its examples).

Multi-table support is proven present: `fluree-db-query/src/r2rml/operator.rs:338-565`
fans out one independent per-table scan per matched TriplesMap (not a union) and
executes `rr:refObjectMap` cross-table joins; green tests exist at
`fluree-db-api/tests/it_graph_source_r2rml.rs:990` (2-map compile), `:1889`
(cross-table join execution, 4 rows), `:1985` (composite key).

## Design / fix

- **Phase 1 (the fix):** in `parser.rs::resolve_iri`, implement RFC 3986 §5.3
  fragment handling for **all** branches — thread the fragment through
  `parse_iri_components` / `parse_hier_part` / `split_path_query` and append
  `#<fragment>` in the result. `<#X>`-only → keep base scheme/authority/path/query +
  ref fragment; `path#frag` → carry that fragment; ref with no fragment → no
  fragment (don't inherit the base's); `<>` → base minus its fragment.
- **Phase 2 (hardening):** make a silent collapse impossible — `extractor.rs` /
  `CompiledR2rmlMapping::add_triples_map` should **error** (or at least warn) when
  two `rr:TriplesMap` subjects resolve to the same IRI, instead of merging
  (first-wins for table/subject, union for POMs) into plausible-but-wrong data.
- **Phase 3 (UX):** report `Tables: N (names…)` alongside `TriplesMaps: N` at
  registration (`graph_source/result.rs`, CLI `commands/iceberg.rs`) so a user
  mapping 16 tables immediately sees `1` if anything is wrong.
- **Phase 4 (optional perf):** wire the operator's prebuilt
  `find_maps_for_class/predicate/pattern` indexes (`mapping/compiled.rs:99-171`,
  currently used only in tests) for O(1) map selection at star-schema scale.

## Tests

- Parser unit tests for `@base <http://ex/edw>` + `<#A>`/`<#B>` → two **distinct**
  IRIs (the regression guard); `<rel#frag>`, fragmented base, `<>` per RFC 3986
  §5.3. Consider wiring the vendored W3C Turtle eval suite (ignored by default).
- An idiomatic `@base` + `<#X>` multi-table fixture compiled offline (assert
  `triples_maps.len() == N`, distinct tables, correct per-table projection, joins);
  drive via the existing `MultiTableMockProvider`.
- Hardening test: colliding-IRI doc → error, not a merged map.

## Risk / compat

`resolve_iri` is the resolution path for all relative IRIs in Turtle ingestion
(used by the R2RML loader and `fluree-db-memory/turtle_io.rs`). Today fragments are
**always** dropped, so any `@base`+fragment Turtle is already parsed wrong — the fix
is strictly more correct. Grep Turtle/golden fixtures for fragment-less IRIs before
landing. Absolute-IRI and prefixed-name inputs are unaffected (existing tests stay
green). Phase-2 hardening may surface a previously "working" (silently corrupt)
mapping as an error — correct outcome; can ship as a warning first.

## Critical files

- `fluree-graph-turtle/src/parser.rs` — `resolve_iri` (~858-958), `parse_iri_components` (~1010), `split_path_query` (~1043) — **the fix**
- `fluree-db-r2rml/src/loader/extractor.rs` (collapse propagation; dup-IRI hardening)
- `fluree-db-r2rml/src/mapping/compiled.rs` (`add_triples_map` dedup; unused selection indexes)
- `fluree-db-query/src/r2rml/operator.rs` (proves multi-table works; Phase-4)
- `fluree-db-api/tests/it_graph_source_r2rml.rs` (add `@base`/`<#X>` fixtures)

## Stacking

**Phase 1 is the foundation — bottom of the stack (PR-1), independent, size S.**
Phase 2/3 share files with the R2RML media-type fix and should stack after it.
Overall size **S–M**.

_Context: surfaced while reading a live 16-table Snowflake-managed Iceberg star
schema (8 dims + 8 facts) through fluree/db via Horizon REST + R2RML._

aaj3f:
# fix(turtle): resolve RFC 3986 §5.3 fragment refs against `@base`

## Summary

The Turtle parser's `resolve_iri` dropped the **fragment** when resolving a
relative IRI reference such as `<#Name>` against an `@base`. As a result, every
`<#DimDate>`, `<#DimProduct>`, … collapsed to the *same* IRI (the base with its
fragment stripped). This is a direct RFC 3986 §5.2.2 / §5.3 violation and is the
root cause of issue #1395: an idiomatic `@base` + `<#Name>` multi-`TriplesMap`
R2RML mapping silently collapsed into a single TriplesMap / single logical table.

This PR is **Phase 1** — the foundational parser fix at the bottom of the
fluree/db ↔ Snowflake-managed-Iceberg stack. It is fully self-contained in the
`fluree-graph-turtle` crate.

## Root cause

Per RFC 3986 §5.2.2, the *fragment* of a resolved reference is **always** the
reference's own fragment (`T.fragment = R.fragment`); it is never inherited from
the base. The previous `resolve_iri` (a) returned `None` for the fragment on the
`#`-only branch, (b) never appended `#<fragment>` during component recomposition,
and (c) used helpers (`split_path_query`) that strip everything after `#`. So the
reference fragment was discarded on every relative resolution, and the empty `<>`
reference incorrectly returned the base *with* its fragment intact. Existing
R2RML fixtures all used **absolute** TriplesMap IRIs, which short-circuit
resolution entirely, so the bug was never exercised.

## What changed

`fluree-graph-turtle/src/parser.rs` — `resolve_iri` only:

- Split the reference into its fragment (everything after the first `#`) and the
  fragment-less remainder up front (RFC 3986 §5.2.1). Scheme/authority/path/query
  are resolved against the remainder; the reference fragment is re-attached during
  recomposition (§5.3).
- Same-document references (`<>` and `<#frag>`) now inherit the base path **and
  query** but **not** the base fragment (the base fragment is naturally dropped by
  `parse_iri_components`).
- Append `#<fragment>` in the result assembly when (and only when) the reference
  itself carries a fragment.
- Removed the special-case early return for the empty `<>` reference, which had
  returned the full base *including* its fragment; `<>` now correctly resolves to
  base-minus-fragment.

No helper signatures changed and no other call sites were touched. Absolute-IRI
and prefixed-name inputs are completely unaffected.

## Tests added

Six new parser unit tests in `fluree-graph-turtle/src/parser.rs` (all driven
through the real `parse` path via `parse_to_graph`):

- `test_base_fragment_refs_resolve_to_distinct_iris` — **the regression guard for
  #1395**: with `@base <http://example.org/edw>`, `<#A>` and `<#B>` resolve to two
  DISTINCT IRIs (`…/edw#A`, `…/edw#B`).
- `test_base_relative_path_with_fragment` — `<DimDate#col>` → `…/DimDate#col`.
- `test_fragment_ref_against_base_with_fragment` — a base that itself has a
  fragment: `<#new>` replaces it; `<other#f>` carries its own fragment.
- `test_empty_ref_against_fragmented_base_drops_fragment` — `<>` against a
  fragmented base → base minus fragment.
- `test_ref_without_fragment_does_not_inherit_base_fragment` — a ref with no
  fragment does NOT inherit the base's.
- `test_query_and_fragment_resolution_rfc3986` — RFC 3986 §5.4.1 query+fragment
  examples against `@base <http://a/b/c/d;p?q>` (`<#s>`, `<?y#s>`, `<g#s>`,
  `<g?y#s>`, `<>`).

## Results

- `cargo test -p fluree-graph-turtle` — **77 passed, 0 failed** (incl. the 6 new + 1 doctest).
- `cargo clippy -p fluree-graph-turtle --all-targets` — clean, no warnings.
- `cargo fmt -p fluree-graph-turtle` — applied.
- Downstream: `cargo test -p fluree-db-r2rml` — **46 passed, 0 failed**;
  `cargo check -p fluree-db-api` — clean.

## Risk / compatibility

`resolve_iri` is the resolution path for all relative IRIs in Turtle ingestion
(R2RML loader, `fluree-db-memory/turtle_io.rs`). Fragments were *always* dropped
before, so any `@base` + fragment Turtle was already parsed incorrectly — the fix
is strictly more correct. The change is conservative:

- No fixtures or golden snapshots encode the old fragment-dropped behavior. The
  only `.ttl` carrying `@base` is the MCP server's own `.fluree-memory/repo.ttl`
  (not a test fixture). The W3C SPARQL testsuite uses its **own** separate base
  resolver (`testsuite-sparql/src/result_format.rs`), not this `resolve_iri`.
- The one existing empty-ref test (`test_empty_iri_resolves_to_base`) uses a base
  with no fragment and stays green.
- Absolute-IRI and prefixed-name inputs are unchanged.

Phases 2–4 from issue #1395 (R2RML dup-IRI hardening, `Tables: N` UX, O(1) map
selection) are intentionally **out of scope** for this PR and stack on top.

Fixes #1395

### DIFF
--- a/fluree-graph-turtle/src/parser.rs
+++ b/fluree-graph-turtle/src/parser.rs
@@ -855,17 +855,10 @@ impl<'a, 'input, S: GraphSink> Parser<'a, 'input, S> {
         Ok(first_node)
     }
 
-    /// Resolve a potentially relative IRI against the base (RFC3986).
+    /// Resolve a potentially relative IRI against the base (RFC 3986 §5).
     fn resolve_iri(&self, reference: &str) -> Result<String> {
-        if reference.is_empty() {
-            return match &self.base {
-                Some(base) => Ok(base.clone()),
-                None => Err(TurtleError::IriResolution(
-                    "empty IRI reference without base".to_string(),
-                )),
-            };
-        }
-
+        // Absolute IRI reference (carries a valid scheme) — returned verbatim,
+        // fragment included; resolution does not apply.
         if let Some(colon_pos) = reference.find(':') {
             let potential_scheme = &reference[..colon_pos];
             if !potential_scheme.is_empty()
@@ -891,9 +884,31 @@ impl<'a, 'input, S: GraphSink> Parser<'a, 'input, S> {
             }
         };
 
-        let (base_scheme, base_authority, base_path, _base_query) = parse_iri_components(base);
+        // RFC 3986 §5.2.1: split the reference into its fragment and everything
+        // before it. The fragment is the portion after the FIRST `#`. Per
+        // §5.2.2 the resolved fragment is ALWAYS the reference's fragment and is
+        // never inherited from the base, so scheme/authority/path/query are
+        // resolved against the fragment-less portion and the reference fragment
+        // is re-attached during recomposition (§5.3).
+        let (ref_no_fragment, ref_fragment) = match reference.find('#') {
+            Some(pos) => (&reference[..pos], Some(&reference[pos + 1..])),
+            None => (reference, None),
+        };
 
-        let (scheme, authority, path, query) = if let Some(rest) = reference.strip_prefix("//") {
+        let (base_scheme, base_authority, base_path, base_query) = parse_iri_components(base);
+
+        let (scheme, authority, path, query) = if ref_no_fragment.is_empty() {
+            // Same-document reference (`<>` or `<#frag>`): the reference has an
+            // empty path and no query, so the target inherits the base path and
+            // query (RFC 3986 §5.2.2). The base's own fragment is dropped because
+            // `parse_iri_components` never returns it.
+            (
+                base_scheme.to_string(),
+                base_authority.map(std::string::ToString::to_string),
+                base_path.to_string(),
+                base_query.map(std::string::ToString::to_string),
+            )
+        } else if let Some(rest) = ref_no_fragment.strip_prefix("//") {
             let (ref_authority, ref_path, ref_query) = parse_hier_part(rest);
             (
                 base_scheme.to_string(),
@@ -901,30 +916,23 @@ impl<'a, 'input, S: GraphSink> Parser<'a, 'input, S> {
                 remove_dot_segments(&ref_path),
                 ref_query,
             )
-        } else if reference.starts_with('/') {
-            let (ref_path, ref_query) = split_path_query(reference);
+        } else if ref_no_fragment.starts_with('/') {
+            let (ref_path, ref_query) = split_path_query(ref_no_fragment);
             (
                 base_scheme.to_string(),
                 base_authority.map(std::string::ToString::to_string),
                 remove_dot_segments(ref_path),
                 ref_query.map(std::string::ToString::to_string),
             )
-        } else if let Some(query_rest) = reference.strip_prefix('?') {
+        } else if let Some(query_rest) = ref_no_fragment.strip_prefix('?') {
             (
                 base_scheme.to_string(),
                 base_authority.map(std::string::ToString::to_string),
                 base_path.to_string(),
                 Some(query_rest.to_string()),
             )
-        } else if reference.starts_with('#') {
-            (
-                base_scheme.to_string(),
-                base_authority.map(std::string::ToString::to_string),
-                base_path.to_string(),
-                None,
-            )
         } else {
-            let (ref_path, ref_query) = split_path_query(reference);
+            let (ref_path, ref_query) = split_path_query(ref_no_fragment);
             let merged = if base_authority.is_some() && base_path.is_empty() {
                 format!("/{ref_path}")
             } else {
@@ -952,6 +960,10 @@ impl<'a, 'input, S: GraphSink> Parser<'a, 'input, S> {
         if let Some(q) = query {
             result.push('?');
             result.push_str(&q);
+        }
+        if let Some(fragment) = ref_fragment {
+            result.push('#');
+            result.push_str(fragment);
         }
 
         Ok(result)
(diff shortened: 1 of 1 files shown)


bplatz (approved):
Great!

PR #1400, opened 2026-06-30
(Iceberg) Apply Vended-Credential `client.region` to the S3 Client

### LINKED ISSUE #1396: Iceberg vended-credentials S3 client ignores region (defaults to us-east-1); reads fail for non-us-east-1 buckets
## Summary

Reading a Snowflake-managed Iceberg table through the Horizon (Apache Polaris)
REST catalog with **vended credentials** fails with
`Failed to read table metadata: Storage error: S3 GetObject failed: service error`,
because fluree's vended-credentials S3 client **signs for `us-east-1` regardless of
the bucket's actual region**. Buckets outside us-east-1 (here, us-east-2) are
unreachable. Setting `AWS_REGION=us-east-2` on the server is the only workaround.

## Root cause (region dropped at three points)

The catalog `loadTable` response carries the region as the Iceberg-REST spec key
`client.region` (NOT `s3.region`), alongside `s3.access-key-id` /
`s3.secret-access-key` / `s3.session-token`.

1. **Parser ignores it.** `fluree-db-iceberg/src/credential/vended.rs:66-69` parses
   only `s3.region`; `client.region` is never read, so `creds.region == None`. (The
   raw value *is* present in `LoadTableResponse.config` from `catalog/rest.rs:339-349`
   — nothing consumes it.)
2. **Override not threaded to the vended client.** `fluree-db-api/src/graph_source/r2rml.rs:665-687`
   passes `io.s3_region` / `io.s3_endpoint` / `io.s3_path_style` **only** on the
   ambient (`from_default_chain`) branch; the vended branch
   (`S3IcebergStorage::from_vended_credentials(credentials)`) receives only the creds.
   So `--s3-region us-east-2` (stored as `io.s3_region`, confirmed via `iceberg info`)
   never reaches the vended client.
3. **Client falls to SDK default.** `fluree-db-iceberg/src/io/storage.rs:87-133`
   applies a region only when `creds.region` is `Some`; with `None` and no
   `AWS_REGION`, aws-sdk-s3 falls back to its `us-east-1` global default.

Debug log confirms: `region: Some("us-east-1")`, endpoint
`…s3.us-east-1.amazonaws.com`, `signingRegion: "us-east-1"` — against a us-east-2
bucket → wrong-region request surfaced as "service error". (The creds themselves are
valid; pyiceberg reads the same vended creds fine. `AWS_REGION=us-east-2` fixes it
precisely because no explicit `.region(...)` is set, so the SDK env provider fills
the gap.)

## Design / fix

Target precedence: **vended `client.region` > vended `s3.region` > `io.s3_region`
override > SDK default chain**; apply the same fallback to endpoint and path-style.

- **`credential/vended.rs`** (`from_config_map`, ~66): parse `client.region` first,
  keep `s3.region` as a defensive fallback; update the doc comment listing expected
  keys.
- **`io/storage.rs`** (`from_vended_credentials`, ~87): accept
  `region_override` / `endpoint_override` / `path_style_override`, mirroring its
  sibling `from_default_chain`; resolve `creds.region.or(region_override)` (and same
  for endpoint/path-style). Leaving region `None` when both are absent preserves
  today's SDK-default behavior (no regression for the working us-east-1 case).
- **`graph_source/r2rml.rs:665-687`**: thread `io.s3_region` / `io.s3_endpoint` /
  `io.s3_path_style` into the vended call, matching the ambient branch.

`from_vended_credentials` is `pub`; either change the signature directly (single
in-repo caller) or add a `_with_overrides` variant + a thin shim for API stability.

## Tests (all offline, no live bucket)

- Parser: `client.region` present → `creds.region == Some(...)`; both
  `client.region` and `s3.region` present → `client.region` wins; only `s3.region`
  → still parsed.
- Client region: assert the built client signs for the resolved region via
  `client.config().region()` — vended region applied; precedence (vended >
  override); override used when vended absent. No network needed.

## Risk / compat

Intended behavior change: when the catalog vends `client.region`, it now overrides
ambient `AWS_REGION` (correct — the catalog is authoritative; removes the
workaround). No regression for us-east-1 (region stays `None` → SDK default). Newly
honoring `io.s3_endpoint`/`io.s3_path_style` on the vended path improves
MinIO-via-vended setups. All changes are behind the `aws` feature; `r2rml.rs` is the
sole storage-construction site, so this covers the whole read path.

## Critical files

- `fluree-db-iceberg/src/credential/vended.rs` (parse `client.region`)
- `fluree-db-iceberg/src/io/storage.rs` (`from_vended_credentials` region/endpoint/path-style precedence)
- `fluree-db-api/src/graph_source/r2rml.rs` (thread io overrides into the vended call, ~665-687)

## Stacking

**Self-contained and the actual data-read blocker → land early (PR-2).** Size
**S–M** (~30-50 LOC across 3 files + tests). Any sibling PR that also edits the
`r2rml.rs:665-687` storage-construction block should rebase on top of this.

_Context: surfaced reading a live Snowflake-managed Iceberg table (us-east-2,
`EXTERNAL_VOLUME = SNOWFLAKE_MANAGED`) through Horizon's vended credentials._

aaj3f:
## Summary

Reading a Snowflake-managed Iceberg table through the Horizon (Apache Polaris)
REST catalog with **vended credentials** failed for any bucket outside
`us-east-1` (e.g. `us-east-2`) with:

```
Failed to read table metadata: Storage error: S3 GetObject failed: service error
```

The catalog vends the bucket's region as the Iceberg-REST spec key
`client.region`, but fluree's vended-credentials S3 client never read it, so the
AWS SDK fell back to its global `us-east-1` default and signed requests for the
wrong region. The only workaround was setting `AWS_REGION=us-east-2` on the
server. This PR threads the region (and endpoint / path-style) through to the
vended client so the catalog's region is honored.

## Root cause — region dropped at three points

1. **Parser ignored it.** `fluree-db-iceberg/src/credential/vended.rs`
   (`from_config_map`) parsed only `s3.region`; the actual `client.region` value
   present in the `loadTable` response was never read, so `creds.region == None`.
2. **Override not threaded to the vended client.**
   `fluree-db-api/src/graph_source/r2rml.rs` passed `io.s3_region` /
   `io.s3_endpoint` / `io.s3_path_style` only on the ambient
   (`from_default_chain`) branch; the vended branch received only the creds, so a
   configured `--s3-region` never reached the vended client.
3. **Client fell to the SDK default.** `fluree-db-iceberg/src/io/storage.rs`
   (`from_vended_credentials`) only applied a region when `creds.region` was
   `Some`; with `None` and no `AWS_REGION`, aws-sdk-s3 defaulted to `us-east-1`.

## What changed

- **`fluree-db-iceberg/src/credential/vended.rs`** — `from_config_map` now parses
  `client.region` first (the Iceberg-REST spec key Snowflake Horizon / Polaris
  send), keeping `s3.region` as a defensive fallback. Doc comment updated to list
  the expected keys.
- **`fluree-db-iceberg/src/io/storage.rs`** — `from_vended_credentials` now takes
  `region_override` / `endpoint_override` / `path_style_override` params,
  mirroring its sibling `from_default_chain`. It resolves
  `creds.region.or(region_override)` (same for endpoint; path-style is enabled if
  either source requests it). Leaving region `None` (vended absent *and* no
  override) preserves today's SDK-default behavior, so the working `us-east-1`
  path does not regress. The single in-repo caller made changing the `pub`
  signature directly cleaner than adding a `_with_overrides` shim. The precedence
  itself is factored into a small pure `resolve_io` helper so it is unit-testable
  without touching SDK internals.
- **`fluree-db-api/src/graph_source/r2rml.rs`** — the vended branch now threads
  `io.s3_region` / `io.s3_endpoint` / `io.s3_path_style` into the
  `from_vended_credentials` call, matching the ambient branch.

**Target precedence:** vended `client.region` > vended `s3.region` >
`io.s3_region` override > SDK default chain (same fallback for endpoint and
path-style).

## Tests (all offline — no live bucket / no network)

Parser (`credential/vended.rs`):
- `test_parse_client_region_preferred` — `client.region` present (no
  `s3.region`) → parsed.
- `test_parse_client_region_beats_s3_region` — both present → `client.region`
  wins.
- `test_parse_s3_region_fallback` — only `s3.region` → still parsed.

Client region (`io/storage.rs`, build via `from_vended_credentials`, assert
`client.config().region()`):
- `test_vended_region_applied_to_client` — vended region reaches the client.
- `test_vended_region_beats_override` — precedence: vended beats the override.
- `test_override_region_used_when_vended_absent` — override applied when vended
  absent.

Endpoint + path-style precedence (`io/storage.rs`): this aws-sdk-s3 version
exposes only `region()` on the built `Config` (endpoint/path-style live in a
crate-private config bag with no public getter), so the override precedence is
factored into a small pure `resolve_io(creds, overrides) -> (region, endpoint,
path_style)` helper — exactly the values `from_vended_credentials` feeds the S3
config builder — and asserted directly:
- `test_resolve_io_endpoint_override_used_when_vended_absent` — `io.s3_endpoint`
  override used when the catalog vends no endpoint.
- `test_resolve_io_path_style_override_forces_true` — `path_style_override=true`
  forces path-style.
- `test_resolve_io_vended_endpoint_and_path_style_beat_override` — vended endpoint
  wins over the override, and a vended `path_style=true` holds even when the
  override is false.

### Results

- `cargo test -p fluree-db-iceberg --features aws` — pass (all 9 new tests green:
  3 parser + 3 client-region + 3 `resolve_io`; full crate suite green).
- `cargo clippy -p fluree-db-iceberg --all-targets --features aws` — clean (no
  warnings).
- `cargo fmt` — clean.
- `cargo check -p fluree-db-api` and `cargo check -p fluree-db-api --features iceberg`
  — both pass (the latter type-checks the changed vended call site, which is
  gated behind the `iceberg` feature).

## Risk / compatibility

- Intended behavior change: when the catalog vends `client.region`, it now
  overrides any ambient `AWS_REGION` — correct, since the catalog is
  authoritative — and **removes the `AWS_REGION=us-east-2` workaround**.
- **No regression for `us-east-1`:** with no vended region and no override, region
  stays `None` → SDK default resolution, exactly as before.
- Newly honoring `io.s3_endpoint` / `io.s3_path_style` on the vended path also
  improves MinIO-via-vended setups.
- All changes are behind the `aws` feature; `r2rml.rs` is the sole
  storage-construction site, so this covers the whole Iceberg read path.

Fixes #1396

### DIFF
--- a/fluree-db-iceberg/src/io/storage.rs
+++ b/fluree-db-iceberg/src/io/storage.rs
@@ -83,9 +83,50 @@ impl Debug for S3IcebergStorage {
 
 #[cfg(feature = "aws")]
 impl S3IcebergStorage {
+    /// Resolve effective region/endpoint/path-style from vended credentials plus
+    /// caller-supplied overrides.
+    ///
+    /// Precedence: the value vended by the catalog wins, then the override, then
+    /// none. Region/endpoint stay `None` when neither source supplies one (so the
+    /// AWS SDK default chain applies); path-style is enabled if either the vended
+    /// creds or the override request it.
+    ///
+    /// Returned as plain owned values so the precedence logic is unit-testable
+    /// without touching SDK internals (the built `aws_sdk_s3::Config` exposes only
+    /// `region()`, not endpoint/path-style).
+    fn resolve_io(
+        creds: &crate::credential::VendedCredentials,
+        region_override: Option<&str>,
+        endpoint_override: Option<&str>,
+        path_style_override: bool,
+    ) -> (Option<String>, Option<String>, bool) {
+        let region = creds
+            .region
+            .clone()
+            .or_else(|| region_override.map(std::string::ToString::to_string));
+        let endpoint = creds
+            .endpoint
+            .clone()
+            .or_else(|| endpoint_override.map(std::string::ToString::to_string));
+        let path_style = creds.path_style || path_style_override;
+        (region, endpoint, path_style)
+    }
+
     /// Create a new S3 storage from vended credentials.
+    ///
+    /// Region/endpoint/path-style precedence (mirrors [`from_default_chain`] for the
+    /// override sources): the value vended by the catalog wins, then the caller-supplied
+    /// override (typically `io.s3_region` / `io.s3_endpoint` / `io.s3_path_style`), then
+    /// the AWS SDK default chain. Leaving region `None` (vended absent *and* no override)
+    /// preserves the SDK-default behavior (e.g. `us-east-1` / `AWS_REGION`) so the
+    /// previously-working us-east-1 path does not regress.
+    ///
+    /// [`from_default_chain`]: Self::from_default_chain
     pub async fn from_vended_credentials(
         creds: &crate::credential::VendedCredentials,
+        region_override: Option<&str>,
+        endpoint_override: Option<&str>,
+        path_style_override: bool,
     ) -> Result<Self> {
         use aws_credential_types::Credentials;
 
@@ -101,12 +142,20 @@ impl S3IcebergStorage {
             "vended-credentials",
         );
 
+        // Resolve region/endpoint/path-style precedence (vended > override > none).
+        let (region, endpoint, path_style) = Self::resolve_io(
+            creds,
+            region_override,
+            endpoint_override,
+            path_style_override,
+        );
+
         // Build config with vended credentials
         let mut config_loader = aws_config::defaults(aws_config::BehaviorVersion::latest())
             .credentials_provider(aws_creds);
 
-        // Set region if provided
-        if let Some(region) = &creds.region {
+        // Set region only when resolved; leaving it None preserves SDK default resolution.
+        if let Some(region) = &region {
             config_loader = config_loader.region(aws_config::Region::new(region.clone()));
         }
 
@@ -115,11 +164,11 @@ impl S3IcebergStorage {
         // Build S3 client, optionally with endpoint override
         let mut s3_config = aws_sdk_s3::config::Builder::from(&sdk_config);
 
-        if let Some(endpoint) = &creds.endpoint {
+        if let Some(endpoint) = &endpoint {
             s3_config = s3_config.endpoint_url(endpoint);
         }
 
-        if creds.path_style {
+        if path_style {
             s3_config = s3_config.force_path_style(true);
         }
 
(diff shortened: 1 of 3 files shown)


fluree-db-iceberg/src/io/storage.rs
```diff
@@ -83,9 +83,50 @@ impl Debug for S3IcebergStorage {
...
+    ) -> (Option<String>, Option<String>, bool) {
+        let region = creds
+            .region
+            .clone()
+            .or_else(|| region_override.map(std::string::ToString::to_string));
+        let endpoint = creds
+            .endpoint
+            .clone()
+            .or_else(|| endpoint_override.map(std::string::ToString::to_string));
+        let path_style = creds.path_style || path_style_override;
```
bplatz, line 114: Worth noting in the doc/PR summary: path-style is an OR of vended and override, not "vended wins" like region/endpoint — an override cannot force it back to false. Correct for a bool, just a slight mismatch with the stated precedence.

bplatz (approved):
Just the one comment

fluree-db-iceberg/src/io/storage.rs
aaj3f: Addressed in https://github.com/fluree/db/pull/1400/commits/904dd4b861dabfd787b8b6d6bbf553a9bdbd4ce3

PR #1401, opened 2026-06-30
(R2RML) Error on Colliding TriplesMap IRIs & Report Table Count

### LINKED ISSUE #1395: Turtle parser drops the fragment resolving `<#name>` against `@base`, collapsing multi-TriplesMap R2RML mappings to one table

aaj3f:
> **Stacked PR.** Base branch: `fix/turtle-fragment-resolution` (#1399)

## Summary

This is **Phases 2 & 3** of issue #1395 — the R2RML half of the multi-table
collapse fix, layered on top of PR-1's Turtle parser fragment fix.

- **Phase 2 (hardening):** make a silent multi-`TriplesMap` collapse *impossible*.
  When two or more `rr:TriplesMap` subjects resolve to the same IRI, the extractor
  now returns a hard error instead of merging them (first-wins logical table /
  subject map, union of every predicate-object map) into plausible-but-wrong data.
- **Phase 3 (reporting):** surface table coverage at registration. The R2RML
  create result and `fluree iceberg map` now report `Tables: N (names…)` next to
  `TriplesMaps: N`, so a 16-table star schema that has silently collapsed shows
  `Tables: 1` immediately rather than failing mysteriously at query time.

Phase 1 (the parser fix) is the base branch. **Phase 4** (wiring the operator's
prebuilt `find_maps_for_*` selection indexes for O(1) map selection) is
intentionally **deferred** and out of scope here.

### Phase 2 — collapse is now a hard error

`fluree-db-r2rml/src/loader/extractor.rs` — `MappingExtractor::extract_all`:

- Each `rr:TriplesMap` IRI is extracted exactly once (a repeated
  `a rr:TriplesMap` triple for an already-seen subject is harmless redundancy and
  is skipped).
- Before extracting, a new `ensure_no_collision` guard counts the `rr:logicalTable`
  and `rr:subjectMap` edges on the subject. A well-formed TriplesMap has exactly
  one of each; **more than one is the signature of two definitions merged onto one
  IRI**, and is now rejected with `R2rmlError::DuplicateTriplesMap`. This signal is
  independent of graph bag/set semantics (the merged subject always carries the
  distinct logical-table blank nodes) and produces **no false positive** for a
  redundant type triple on a single well-formed map.

`fluree-db-r2rml/src/error.rs` — new `R2rmlError::DuplicateTriplesMap(String)`
variant whose message names the colliding IRI and points at the usual cause
(relative `<#fragment>` subjects collapsing against `@base`).

### Phase 3 — `Tables: N (…)` at registration

- `fluree-db-api/src/graph_source/result.rs` — `R2rmlCreateResult` gains
  `table_count: usize` and `table_names: Vec<String>` (sorted, distinct).
- `fluree-db-api/src/graph_source/r2rml.rs` — both the CAS-stored
  (`R2rmlMappingInput::Content`) and address-validated
  (`R2rmlMappingInput::Address`) paths now derive the table list from the compiled
  mapping via a small `sorted_table_names` helper; `validate_r2rml_mapping_from_address`
  returns `(count, table_names)`.
- `fluree-db-cli/src/commands/iceberg.rs` — prints
  `Tables:      N (DW.X, DW.Y, …)` after `TriplesMaps:` on **both** the local and
  remote (`fluree iceberg map`) output paths, via a shared `format_table_summary`
  helper (falls back to bare `N` when no names are available, e.g. an unvalidated
  address mapping).
- `fluree-db-server/src/routes/iceberg.rs` — the `/iceberg/map` JSON response
  (`IcebergMapResponse`) carries optional `table_count` / `table_names` so the
  remote CLI path renders the same line. No behavior change for the raw (non-R2RML)
  Iceberg branch.

## Tests added

In `fluree-db-api/tests/it_graph_source_r2rml.rs`:

- `test_base_fragment_multi_table_compiles_to_distinct_tables` — the offline
  regression guard. An idiomatic `@base <http://ex/edw>` + `<#DimDate>` /
  `<#DimProduct>` / `<#FactSales>` fixture compiles to **3 distinct TriplesMaps**
  (`…edw#DimDate`, `…edw#DimProduct`, `…edw#FactSales`) over **3 distinct tables**
  (`DW.DIM_DATE`, `DW.DIM_PRODUCT`, `DW.FACT_SALES`) — not one merged map.
- `engine_e2e_base_fragment_scans_non_first_table` — drives the engine with a
  recording provider (the `MultiTableMockProvider` pattern). Querying a predicate
  owned only by the **non-first** map scans `DW.DIM_PRODUCT` and **never**
  `DW.DIM_DATE` (the first map's table — the exact collapse symptom), returning the
  correct rows.
- `test_colliding_triples_map_iris_error` — the Phase-2 hardening test: a document
  with two `rr:TriplesMap` subjects resolving to one IRI fails to compile with a
  `Duplicate TriplesMap IRI` error naming the colliding subject.

## Results

- `cargo test -p fluree-db-r2rml` — **46 passed, 0 failed** (+1 doctest);
  `--features turtle` — **55 passed, 0 failed** (+1 doctest).
- `cargo test -p fluree-db-api --features iceberg` — full suite green; the touched
  binary `it_graph_source_r2rml` is **32 passed, 0 failed** (2 ignored — external
  Polaris/MinIO), including the 3 new tests. No regressions across the rest of the
  api suite (640 unit + all `grp_*`/`it_*` integration binaries, 0 failures).
- `cargo clippy -- -D warnings` (all `--all-targets`) — clean on
  `fluree-db-r2rml --features turtle`, `fluree-db-api --features iceberg`,
  `fluree-db-cli --features iceberg`, and `fluree-db-server`.
- `cargo fmt` — applied. `cargo check -p fluree-db-cli --features iceberg` — clean.

## Risk / compatibility

- **Phase 2 may surface a previously "working" mapping as an error.** That mapping
  was already silently corrupt (first-wins table + union projection); a loud error
  is the correct, strictly-safer outcome. Only mappings that actually collide are
  affected — distinct-IRI multi-table mappings (the norm) are untouched.
- **Phase 3 is purely additive**: two new result fields and an extra output line;
  the server response fields are `skip_serializing_if = "Option::is_none"`, and the
  only `R2rmlCreateResult` constructor is updated. The raw-Iceberg path is
  unchanged.
- All existing R2RML fixtures use absolute or now-correctly-resolved IRIs and stay
  green.

Fixes #1395

### DIFF
--- a/fluree-db-api/src/graph_source/r2rml.rs
+++ b/fluree-db-api/src/graph_source/r2rml.rs
@@ -185,10 +185,13 @@ impl crate::Fluree {
         config.validate()?;
 
         // Resolve mapping: validate and store to CAS if inline content
-        let (mapping_address, triples_map_count, mapping_validated) = match &config.mapping {
+        let (mapping_address, triples_map_count, table_names, mapping_validated) = match &config
+            .mapping
+        {
             R2rmlMappingInput::Content(content) => {
                 let compiled = Self::compile_r2rml_content(content, &config)?;
                 let count = compiled.len();
+                let tables = Self::sorted_table_names(&compiled);
                 let gs_id = config.graph_source_id();
                 let cs = self.content_store(&gs_id);
                 let cid = cs
@@ -202,20 +205,21 @@ impl crate::Fluree {
                     })?;
                 let addr = cid.to_string();
                 info!(graph_source_id = %graph_source_id, mapping_cid = %addr, "R2RML mapping stored to CAS");
-                (addr, count, true)
+                (addr, count, tables, true)
             }
             R2rmlMappingInput::Address(address) => {
-                let (count, validated) = self
-                    .validate_r2rml_mapping_from_address(address, &config)
-                    .await
-                    .map(|c| (c, true))
-                    .unwrap_or_else(|e| {
-                        warn!(graph_source_id = %graph_source_id, error = %e, "Could not validate R2RML mapping from address");
-                        (0, false)
-                    });
-                (address.clone(), count, validated)
+                let (count, tables, validated) = self
+                        .validate_r2rml_mapping_from_address(address, &config)
+                        .await
+                        .map(|(c, t)| (c, t, true))
+                        .unwrap_or_else(|e| {
+                            warn!(graph_source_id = %graph_source_id, error = %e, "Could not validate R2RML mapping from address");
+                            (0, Vec::new(), false)
+                        });
+                (address.clone(), count, tables, validated)
             }
         };
+        let table_count = table_names.len();
 
         // Test catalog connection (REST mode only)
         let connection_tested = if config.iceberg.is_rest() {
@@ -248,6 +252,8 @@ impl crate::Fluree {
             catalog_uri: config.iceberg.catalog_uri_or_location().to_string(),
             mapping_source: mapping_address,
             triples_map_count,
+            table_count,
+            table_names,
             connection_tested,
             mapping_validated,
         })
@@ -324,11 +330,14 @@ impl crate::Fluree {
     }
 
     /// Validate an R2RML mapping from a pre-existing storage address.
+    ///
+    /// Returns the number of TriplesMap definitions and the sorted list of
+    /// distinct logical table names referenced by the mapping.
     async fn validate_r2rml_mapping_from_address(
         &self,
         address: &str,
         config: &R2rmlCreateConfig,
-    ) -> Result<usize> {
+    ) -> Result<(usize, Vec<String>)> {
         let storage = self.admin_storage().ok_or_else(|| {
             crate::ApiError::Config(format!(
                 "Cannot load R2RML mapping from address '{address}': address-based reads are not supported on this backend"
@@ -342,7 +351,20 @@ impl crate::Fluree {
         let content = String::from_utf8(bytes).map_err(|e| {
             crate::ApiError::Config(format!("R2RML mapping is not valid UTF-8: {e}"))
         })?;
-        Ok(Self::compile_r2rml_content(&content, config)?.len())
+        let compiled = Self::compile_r2rml_content(&content, config)?;
+        Ok((compiled.len(), Self::sorted_table_names(&compiled)))
+    }
+
+    /// Collect the distinct logical table names referenced by a compiled
+    /// mapping, sorted for deterministic reporting.
+    fn sorted_table_names(compiled: &CompiledR2rmlMapping) -> Vec<String> {
+        let mut names: Vec<String> = compiled
+            .table_names()
+            .into_iter()
+            .map(str::to_string)
+            .collect();
+        names.sort();
+        names
     }
 }
 
--- a/fluree-db-api/src/graph_source/result.rs
+++ b/fluree-db-api/src/graph_source/result.rs
@@ -230,6 +230,12 @@ pub struct R2rmlCreateResult {
     /// Number of TriplesMap definitions in the mapping
     pub triples_map_count: usize,
 
+    /// Number of distinct logical tables referenced by the mapping
+    pub table_count: usize,
+
+    /// Names of the distinct logical tables referenced by the mapping (sorted)
+    pub table_names: Vec<String>,
+
     /// Whether the catalog connection was tested successfully
     pub connection_tested: bool,
 
--- a/fluree-db-api/tests/it_graph_source_r2rml.rs
+++ b/fluree-db-api/tests/it_graph_source_r2rml.rs
@@ -16,7 +16,8 @@ use fluree_db_query::error::{QueryError, Result as QueryResult};
 use fluree_db_query::r2rml::{ColumnBatchStream, R2rmlProvider, R2rmlTableProvider, ScanFilter};
 use fluree_db_r2rml::loader::R2rmlLoader;
 use fluree_db_r2rml::mapping::CompiledR2rmlMapping;
-use std::sync::Arc;
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
 
 // Additional imports for engine-level E2E tests
 use fluree_db_api::{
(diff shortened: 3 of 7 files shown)


bplatz (approved):
👏

PR #1402, opened 2026-06-30
(Iceberg) Expose OAuth2 `scope`/`audience` for REST Catalog Auth

aaj3f:
## Summary

Iceberg REST-catalog OAuth2 *client-credentials* auth already form-encodes a
`scope` at the engine level (`OAuth2Config.scope`), but `scope` and `audience`
were not exposed through **any** public config surface — the builders, the CLI,
and the HTTP API all hardcoded them to `None`. As a result, connecting to a
scope-gated Iceberg REST catalog was impossible via supported config. The
concrete blocked case is **Snowflake Horizon** (Snowflake's built-in
Apache-Polaris Iceberg REST endpoint), which **requires**
`scope=session:role:<ROLE>` on the `/v1/oauth/tokens` client-credentials
exchange — and **rejects a non-empty `client_id`** on that exchange with
`400 invalid_scope`.

This PR threads `scope`/`audience` through every layer and makes the empty
`client_id` case work.

## What changed (the middle-layer builder gap)

The engine already knew how to send `scope`; the gap was purely the user-facing
plumbing between the public config types and the engine `AuthConfig`. Nothing in
the token-exchange or downstream read path changed.

- **`fluree-db-api` (builders):** added fluent `with_oauth2_scope` /
  `with_oauth2_audience` setters on `IcebergCreateConfig` that mutate the
  existing `AuthConfig::OAuth2ClientCredentials` in place (warn-and-no-op if
  OAuth2 client-credentials auth has not been set first, or in Direct catalog
  mode). Added delegating setters on `R2rmlCreateConfig`. The existing 3-arg
  `with_auth_oauth2(token_url, client_id, client_secret)` is **unchanged**.
- **`fluree-db-iceberg` (engine):** `fetch_token` now **omits `client_id` from
  the `client_credentials` form when it is empty** (Snowflake Horizon's
  `session:role:` exchange requires an absent/empty `client_id`; a non-empty one
  is interpreted as the principal flow and rejected).
- **`fluree-db-cli`:** added `--oauth2-scope` / `--oauth2-audience` to
  `fluree iceberg map`, wired into both the local `with_oauth2_*` chaining and
  the `args_to_json` remote body. **Relaxed activation:** OAuth2 now engages on
  `token_url + client_secret`, with `client_id` defaulting to `""`.
- **`fluree-db-server`:** added `oauth2_scope` / `oauth2_audience` to the
  `IcebergMapRequest` body with matching activation + build wiring.

## Files

| File | +/- |
|---|---|
[...]

Issue #1394, opened 2026-06-30
Iceberg REST catalog: expose OAuth2 `scope` (+ allow empty `client_id`) so Snowflake Horizon / Polaris `session:role:` auth works


aaj3f:
## Summary

The Iceberg REST-catalog OAuth2 *client-credentials* auth supports a `scope` at the
engine level (`OAuth2Config.scope`, form-encoded into the token request), but
`scope` is **not exposed** through the public config surface — the builders, the
CLI, or the HTTP API all force it to `None`. As a result, connecting to any
Iceberg REST catalog that requires a token-exchange scope is impossible via the
supported config. The concrete blocked case is **Snowflake Horizon Catalog**
(Snowflake's built-in Apache-Polaris Iceberg REST endpoint), which **requires**
`scope=session:role:<ROLE>` on the `/v1/oauth/tokens` client-credentials exchange.

This blocks `fluree/db`-managed Iceberg connections to Snowflake-managed Iceberg
tables (a pattern we want to support / smoke-test).

## Why it matters

Snowflake Horizon (and Polaris generally) gate the catalog session role through the
OAuth scope. Without passing `scope`, the token exchange either fails or yields a
token without the intended role, so the catalog returns no tables. The engine
already knows how to send `scope`; only the user-facing plumbing is missing.

## Evidence (verified against a live Snowflake Horizon account)

Token exchange at `https://<org>-<account>.snowflakecomputing.com/polaris/api/catalog/v1/oauth/tokens`,
`grant_type=client_credentials`, `client_secret=<PAT>`:

| Request | Result |
|---|---|
| `client_secret=<PAT>` + `scope=session:role:ICEBERG_READER`, **no** `client_id` | **200**, `access_token` |
| `client_id=<non-empty>` + `client_secret=<PAT>` + same scope | **400 `invalid_scope`** |
| `client_id=""` (empty) + `client_secret=<PAT>` + same scope | **200**, `access_token` |

So: (a) `scope=session:role:<ROLE>` is **required**; (b) `client_id` must be
**empty** — a non-empty `client_id` makes Snowflake interpret the request as the
Open-Catalog *principal* flow (which expects `PRINCIPAL_ROLE:` scope) and rejects
the `session:role:` scope.

Separately verified: with a correctly-scoped token, the REST catalog lists all
tables and an external engine (pyiceberg) reads data via vended credentials
(`X-Iceberg-Access-Delegation: vended-credentials`) — i.e. everything downstream of
[...]

PR #1403, opened 2026-06-30
(R2RML) Default Mapping `media_type` to Turtle at Both Compile Sites

### LINKED ISSUE #1397: R2RML media-type not persisted/defaulted: querying a `.ttl`-mapped graph source fails with "JSON-LD not supported"
## Summary

Mapping a Turtle R2RML file **without** `--r2rml-type` registers fine ("Mapping:
validated") but every query against the graph source then fails with:

```
Invalid query: R2RML mapping for '<gs>:main' uses JSON-LD format, which is not yet
supported. Please use Turtle format (.ttl).
```

even though the mapping IS Turtle. Passing `--r2rml-type text/turtle` is the only
workaround.

## Root cause — an asymmetric default

Two independent format resolvers disagree on the "no media_type" case:

| Path | Site | `None` resolves to |
|---|---|---|
| Registration | `compile_r2rml_content` `fluree-db-api/src/graph_source/r2rml.rs:308-311` (`is_none_or(\|mt\| mt.contains("turtle"))`) | **Turtle** → validates |
| Query | `compiled_mapping` `r2rml.rs:511-536` (sniffs `mapping_source.ends_with(".ttl")`) | **JSON-LD** → error |

The stored `mapping_source` is a **content-addressed CID** (e.g. `bagiibq…`) with no
extension, so the query-time sniff always fails and falls to the JSON-LD error
branch. The "inferred from extension" the CLI advertises is **never implemented**:
`fluree-db-cli/src/commands/iceberg.rs:571,582` reads the file *content* and forwards
`r2rml_type` unchanged (None stays None) — the API/server never receive a filename.
And the resolved Turtle format from registration is thrown away:
`graph_source/config.rs:837-844` persists `media_type: None` (→ `"media_type": null`,
visible in `fluree iceberg info`). Every existing integration test masks the bug by
always passing `.with_mapping_media_type("text/turtle")`.

## Design / fix

One shared format resolver, used by **both** registration and query, with
precedence **explicit media_type > inferred-from-extension > default (Turtle)**.
Turtle is the only supported format today, so the default must be Turtle (already
the de-facto registration default).

- **Shared resolver** in `fluree-db-r2rml/src/loader/mod.rs` (next to
  `from_turtle`/`from_jsonld`): `MappingFormat::resolve(media_type: Option<&str>,
  source: &str) -> MappingFormat` (None + non-`.jsonld` source → Turtle).
- **Query path** (`r2rml.rs:511-536`) and **registration path**
  (`compile_r2rml_content`, `r2rml.rs:304-324`) both call the resolver — making them
  provably identical (the only behavior change: `None` → Turtle instead of erroring).
- **Persist the resolved type** (`r2rml.rs` create + `config.rs:837-844`) so
  `iceberg info` shows `"media_type":"text/turtle"` and stored configs are
  self-describing.
- **CLI extension inference** (`commands/iceberg.rs`): compute
  `r2rml_type.or_else(|| infer_from_path(&r2rml))` (`.ttl`/`.turtle` → `text/turtle`;
  `.jsonld`/`.json` → `application/ld+json`) so the documented behavior is real.

Minimum viable fix = shared resolver + both call sites; that alone fixes the
reported flow and every already-stored `null`-media_type source.

## Tests

- Resolver precedence unit tests (`None` + CID → Turtle is the regression guard;
  explicit beats extension).
- Integration: `create_r2rml_graph_source` with a `.ttl` mapping and
  `mapping_media_type: None`, then `compiled_mapping(...)` → must be `Ok` (today it
  errors). Mirrors "map a .ttl without `--r2rml-type`, then query."
- Persistence: created config shows `media_type == Some("text/turtle")` (not null).
- Negative: a JSON-LD-resolved source still returns the "not yet supported" error.

## Risk / compat

Small, localized. No schema/nameservice change. **No migration needed** — existing
`null`-media_type sources are fixed in place by the query-side default. The only
semantic change is `None`/unknown → Turtle instead of an error (Turtle is the sole
supported format and registration already assumed it), so it can only turn a
guaranteed failure into a success; a genuine JSON-LD mapping still errors clearly.
(`R2rmlCache::mapping_cache_key` hashes `media_type`; persisting a concrete value is
a one-time cache-key change — pure perf, no correctness impact.)

## Critical files

- `fluree-db-api/src/graph_source/r2rml.rs` (`compile_r2rml_content` ~304-324; `compiled_mapping` ~511-536)
- `fluree-db-r2rml/src/loader/mod.rs` (shared `MappingFormat` resolver)
- `fluree-db-api/src/graph_source/config.rs` (`to_iceberg_gs_config` ~837-844 — persist resolved type)
- `fluree-db-cli/src/commands/iceberg.rs` (extension inference)
- `fluree-db-api/tests/it_graph_source_r2rml.rs` (regression test)

## Stacking

Size **S**, no upstream dependency. Shares `compiled_mapping` / the CLI map command
with the Turtle/multi-TriplesMap work (#different lines) — landing this first gives
that work a clean rebase. Suggested position PR-4 (after the read-unblockers).

_Context: surfaced querying a live Snowflake-managed Iceberg star schema through
fluree/db via Horizon REST + R2RML._

aaj3f:
> **Stacks on #1402** (`fix/iceberg-oauth2-scope`, issue #1394). 

## Summary

R2RML mapping content was compiled with an **asymmetric default media-type**:
the registration path defaulted to **Turtle**, but the query-time path defaulted
to **JSON-LD**. A mapping registered as Turtle *without* an explicit `media_type`
therefore registered fine but failed to recompile at query time with a
"JSON-LD not supported" / parse error. The only workaround was to always pass an
explicit `--r2rml-type text/turtle`.

This PR introduces a single shared format resolver (defaulting to **Turtle**),
routes **both** compile sites through it, and **persists** the resolved
`media_type` so the registration-time and query-time decisions are provably
identical.

## Root cause — two independent format decisions

- **Registration** (`compile_r2rml_content` in
  `fluree-db-api/src/graph_source/r2rml.rs`) treated content as Turtle by default.
- **Query** (`compiled_mapping` in the same module) sniffed `source.ends_with(".ttl")`
  and otherwise fell through to a JSON-LD default — a *different* rule from the
  registration path.
- The persisted `MappingSource.media_type` was left `None` when the user didn't
  pass one, so the query path had nothing authoritative to read and re-derived the
  format with its divergent rule.

## What changed

- **`fluree-db-r2rml/src/loader/mod.rs`** — new shared, dependency-free resolver
  (not behind the `turtle` feature; pure logic):
  - `pub enum MappingFormat { Turtle, JsonLd }`
  - `pub fn resolve(media_type: Option<&str>, source: &str) -> MappingFormat` —
    precedence: **explicit recognized `media_type` > source extension
    (`.jsonld`/`.json` → JSON-LD) > default Turtle**. An unrecognized explicit
    media_type falls through to the extension, then to Turtle. Case-insensitive.
  - `pub fn media_type(self) -> &'static str` → `"text/turtle"` /
    `"application/ld+json"`.
- **`fluree-db-api/src/graph_source/r2rml.rs`** — both compile sites now route
  through `MappingFormat::resolve`. Registration `compile_r2rml_content` is
  re-signatured to `(content, media_type, source)`; the query `compiled_mapping`
  drops the ad-hoc `.ends_with(".ttl")` sniff. The two paths are now identical by
  construction.
- **`fluree-db-api/src/graph_source/config.rs`** — `to_iceberg_gs_config` persists
  the resolved canonical `media_type` into `MappingSource.media_type`: an explicit
  user value is kept verbatim; `None` resolves to `"text/turtle"`.
- **`fluree-db-cli/src/commands/iceberg.rs`** — `infer_mapping_media_type` wired
  into **both** the remote (`args_to_json`) and local (`run_iceberg_map_local`)
  paths, so `.ttl`/`.jsonld` files get the right type without `--r2rml-type`; an
  explicit `--r2rml-type` still overrides inference.

## Persistence / no migration

`MappingSource.media_type` is already `Option<String>` with `#[serde(default)]`,
so pre-existing records with `null`/missing `media_type` still deserialize — the
query-side default (now Turtle, via the resolver) handles them. The only semantic
change is **"None/unknown → Turtle"** at the query site (previously JSON-LD).
Persisting a concrete value is a one-time change to the mapping cache key (pure
perf; no behavioral effect).

## Tests

Resolver (`fluree-db-r2rml`, no features required):
- `resolve_defaults_to_turtle_for_no_media_type_and_no_extension` — the regression
  guard + default-Turtle.
- `resolve_infers_format_from_source_extension`
- `resolve_explicit_media_type_overrides_extension`
- `resolve_unrecognized_media_type_falls_back_to_extension_then_turtle`
- `media_type_returns_canonical_strings`

API (`fluree-db-api --features iceberg`):
- `integration_r2rml_turtle_without_media_type_compiles_at_query_time` — registers
  a `.ttl` mapping with **no** media_type, asserts it persists `text/turtle` **and**
  that the real `FlureeR2rmlProvider::compiled_mapping` returns `Ok` (exercises the
  actual query path, not a mock). This is the bug, captured.
- `test_r2rml_persists_resolved_media_type_no_migration` — serialize→deserialize
  round-trip stays `text/turtle`; an explicit `application/ld+json` is preserved
  verbatim.

CLI (`fluree-db-cli --features iceberg`):
- `infer_mapping_media_type_maps_known_extensions`
- `args_to_json_infers_r2rml_type_from_ttl_extension`
- `args_to_json_explicit_r2rml_type_overrides_extension`

### Results

- `cargo test -p fluree-db-r2rml` (and `--features turtle`) — pass.
- `cargo test -p fluree-db-api --features iceberg` — pass (incl. the two new
  integration tests; `it_graph_source_r2rml` green).
- `cargo test -p fluree-db-cli --features iceberg` — pass.
- `cargo clippy -p fluree-db-r2rml -p fluree-db-api -p fluree-db-cli --all-targets
  --features iceberg` — clean.
- `cargo fmt --all -- --check` — clean.

## Risk / compatibility

- Intended behavior change: a mapping whose format can't be determined from an
  explicit media_type or a source extension now defaults to **Turtle** at the query
  site (it already did at registration) — this is the fix, and it removes the need
  to always pass `--r2rml-type text/turtle`.
- No migration: old `null`/missing `media_type` records still load.
- An explicit user-supplied `media_type` is always preserved verbatim.

Fixes #1397

### DIFF
--- a/fluree-db-cli/src/commands/iceberg.rs
+++ b/fluree-db-cli/src/commands/iceberg.rs
@@ -316,6 +316,21 @@ async fn run_iceberg_map_remote(
 }
 
 /// Convert CLI args to a JSON body for the server endpoint.
+/// Infer an R2RML mapping media type from a file extension.
+///
+/// `.ttl`/`.turtle` → `text/turtle`, `.jsonld`/`.json` → `application/ld+json`
+/// (case-insensitive). Returns `None` for unrecognized extensions so the caller
+/// falls back to the server-side default (Turtle, per issue #1397). This makes
+/// the `--r2rml-type` flag's documented "inferred from extension" behavior real.
+fn infer_mapping_media_type(path: &std::path::Path) -> Option<String> {
+    let ext = path.extension()?.to_str()?.to_ascii_lowercase();
+    match ext.as_str() {
+        "ttl" | "turtle" => Some("text/turtle".to_string()),
+        "jsonld" | "json" => Some("application/ld+json".to_string()),
+        _ => None,
+    }
+}
+
 fn args_to_json(args: &IcebergMapArgs) -> CliResult<serde_json::Value> {
     let mut body = serde_json::json!({
         "name": args.name,
@@ -339,8 +354,15 @@ fn args_to_json(args: &IcebergMapArgs) -> CliResult<serde_json::Value> {
         })?;
         obj.insert("r2rml".into(), content.into());
     }
-    if let Some(ref v) = args.r2rml_type {
-        obj.insert("r2rml_type".into(), v.clone().into());
+    // Explicit `--r2rml-type` wins; otherwise infer from the mapping file
+    // extension so the server records a concrete, self-describing media type
+    // (issue #1397).
+    let r2rml_type = args
+        .r2rml_type
+        .clone()
+        .or_else(|| args.r2rml.as_deref().and_then(infer_mapping_media_type));
+    if let Some(v) = r2rml_type {
+        obj.insert("r2rml_type".into(), v.into());
     }
     if let Some(ref v) = args.branch {
         obj.insert("branch".into(), v.clone().into());
@@ -585,7 +607,12 @@ async fn run_iceberg_map_local(args: IcebergMapArgs, dirs: &FlureeDir) -> CliRes
         let config = fluree_db_api::R2rmlCreateConfig {
             iceberg: iceberg_config,
             mapping: fluree_db_api::R2rmlMappingInput::Content(mapping_content),
-            mapping_media_type: args.r2rml_type.clone(),
+            // Explicit `--r2rml-type` wins; otherwise infer from the file
+            // extension (issue #1397).
+            mapping_media_type: args
+                .r2rml_type
+                .clone()
+                .or_else(|| infer_mapping_media_type(r2rml_path)),
         };
 
         let result = fluree.create_r2rml_graph_source(config).await?;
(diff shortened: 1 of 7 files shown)


fluree-db-cli/src/commands/iceberg.rs
```diff
@@ -316,6 +316,21 @@ async fn run_iceberg_map_remote(
 }
 
 /// Convert CLI args to a JSON body for the server endpoint.
+/// Infer an R2RML mapping media type from a file extension.
+///
+/// `.ttl`/`.turtle` → `text/turtle`, `.jsonld`/`.json` → `application/ld+json`
+/// (case-insensitive). Returns `None` for unrecognized extensions so the caller
+/// falls back to the server-side default (Turtle, per issue #1397). This makes
+/// the `--r2rml-type` flag's documented "inferred from extension" behavior real.
+fn infer_mapping_media_type(path: &std::path::Path) -> Option<String> {
```
bplatz, line 344: This duplicates the ext→media-type table already in `MappingFormat::resolve`. They agree today but could drift; consider routing through `MappingFormat::resolve(None, path_str).media_type()` so there is one source of truth. Minor — different crate and this takes a `Path`, so a local helper is defensible.

bplatz (approved):
Just the one comment, not necessary but if easy to address please do.

fluree-db-cli/src/commands/iceberg.rs
aaj3f: Addressed in https://github.com/fluree/db/pull/1403/commits/61a6b74662b79e43abaf8c6b72067a1df4bbed03

PR #1404, opened 2026-06-30
(CLI) Route Queries Targeting Iceberg/R2RML Graph Sources

### LINKED ISSUE #1398: CLI cannot query a registered Iceberg/R2RML graph source (`fluree query <graph-source>` → "ledger not found")
## Summary

A registered Iceberg/R2RML graph source can be created (`fluree iceberg map`) and
listed (`fluree iceberg list`) but **cannot be queried via the implicit `fluree
query` path**. `fluree query <graph-source>[:main]` returns "ledger not found", and
`GRAPH`/`FROM` references to the source from inside a native-ledger query don't
federate to it. Only the connection-scoped HTTP endpoint (`POST /v1/fluree/query`
with an inline `FROM`) or the explicit `--remote` path reach it.

## Root cause

`fluree query <target>` resolves the target via `resolve_ledger_mode`
(`fluree-db-cli/src/context.rs:56-120`), which uses `ledger_exists`
(`fluree-db-api/src/lib.rs:3769-3771` = `nameservice().lookup(...)`) — **native
ledgers only**. Graph sources live under a separate key
(`lookup_graph_source` / `resolve_graph_source`). So a graph-source name fails the
`ledger_exists` check and returns `NotFound` **before any server routing**.

Federation from the ledger path also can't reach a graph source:
- direct-local + `FROM` → `validate_sparql_for_view` rejects FROM on a single-ledger GraphDb (`view/query.rs:546-566`);
- server-routed + `FROM` → ledger-scoped `/query/<alias>` requires every FROM base == endpoint base → "Ledger mismatch" (`routes/query.rs:2277-2378`), compounded by the CLI sending the un-normalized alias and `base_ledger_id` not stripping `:branch`;
- server-routed + `GRAPH <gs:main>` (no FROM) → treated as a named graph *within* the endpoint ledger → resolves empty (0 rows).

## What already works (narrows the fix)

- The connection-scoped endpoint **works**: `POST /v1/fluree/query` with
  `Content-Type: application/sparql-query` + inline `FROM <gs:main>` executes via
  `query_from()` / `query_connection_sparql`, which auto-attach the R2RML provider.
- The server's ledger-scoped endpoint **already resolves a graph-source-named
  target itself**: `POST /query/<graph-source>` with a no-FROM body works
  (`routes/query.rs:2037-2059, 2853-2897` → `state.fluree.graph(id).query()`).
- Therefore **`fluree query --remote <origin> <graph-source> -e '<no-FROM sparql>'`
  already works today** — the `--remote` branch bypasses `resolve_ledger_mode`. Only
  the *implicit* (non-`--remote`) path is broken.
- The CLI remote client already has `query_connection_sparql` /
  `query_connection_jsonld` (`remote_client.rs:925-952`) — currently **dead code**.
- The documented federated example at `docs/cli/iceberg.md:160`
  (`fluree query mydb 'SELECT … FROM <mydb:main> WHERE { GRAPH <gs:main> {…} }'`) is
  **broken as written** and should be corrected.

## Design / fix (almost entirely CLI; ~zero server change)

- **(a) Resolve graph-source targets → single-target path.** In
  `resolve_ledger_mode`, after `ledger_exists` fails, probe
  `nameservice().lookup_graph_source(&id)`; on hit, mark a graph-source target and
  execute locally via `resolve_graph_source(&alias)` + the R2RML-aware
  `GraphSnapshotQueryBuilder…with_r2rml()` (server already supports the routed form).
  Normalize the alias to `<name>:main` for routing.
- **(b) Auto-route federation.** When the query body carries `FROM`/`from` (or FROM
  NAMED) referencing a source other than the endpoint, route to the connection-scoped
  path — wiring the already-present `query_connection_sparql/jsonld` (Tracked) or
  `query_from()` (Local). Gate on "references a target other than the endpoint" so
  existing same-ledger `FROM <self:main>` is unaffected.
- **(c) `--connection` flag** to force the connection path explicitly (positional
  optional).
- **Docs:** fix `docs/cli/iceberg.md:160` to the single-target form; keep
  `docs/graph-sources/iceberg.md` (`fluree.graph(...).query()` / `query_from()`) as
  the API source of truth.
- Guardrails: reject `--at` / `--explain` / `--format ndjson|csv|tsv` for
  graph-source targets up front (mirrors the server's own restrictions), with clear
  messages.

## Tests

CLI integration (`fluree-db-cli/tests/integration.rs`, build `--features iceberg`):
`fluree query <gs>` resolves (no NotFound); local single-target SPARQL + JSON-LD
return rows; federation auto-route no longer errors with "single-ledger"/"Ledger
mismatch"; `--connection` forces the path; guardrails error clearly; native-ledger
queries / NDJSON / default-context tests stay green. Add a server route test that
`POST /query/<graph-source>` (no FROM) returns rows (codifies the dependency).

## Risk / compat

Backward compat preserved: the new branch is reached only when `ledger_exists` is
false *and* `lookup_graph_source` succeeds — a path that previously always errored.
Local graph-source query requires the CLI built `--features iceberg` (give a clear
"rebuild with --features iceberg" message otherwise); `--remote`/server-routed paths
need no CLI feature. Split-storage server-only sources won't be seen by the local
probe — documented limitation, `--remote` is the workaround.

## Critical files

- `fluree-db-cli/src/context.rs` (graph-source probe in `resolve_ledger_mode`; alias normalization)
- `fluree-db-cli/src/commands/query.rs` (single-target via `with_r2rml`; federation auto-route; guardrails)
- `fluree-db-cli/src/cli.rs` + `lib.rs` (`--connection`, optional positional)
- `fluree-db-cli/src/remote_client.rs` (wire `query_connection_sparql/jsonld`)
- `docs/cli/iceberg.md` (fix the `:160` example)

## Stacking

No hard code dependency on the others, but its E2E test wants a *creatable +
scannable* source → **land after the create/scan fixes (PR-6)**. Size **M** (CLI
routing + `LedgerMode` plumbing; server changes none, or one optional hardening to
make `base_ledger_id` branch-insensitive).

_Context: surfaced trying to query a live Snowflake-managed Iceberg star schema
registered as a fluree/db R2RML graph source._

aaj3f:
## Summary

`fluree query <name>` could only resolve **native ledgers**:
`context.rs::resolve_ledger_mode` had no notion of a registered Iceberg/R2RML
**graph source**, so querying one from the CLI failed even though the source was
registered and the server could serve it. This PR teaches the CLI to resolve a
graph-source name and route the query to the graph-source-aware execution path
(local single-target, or the connection path for `FROM`-federation).

## What changed

- **`fluree-db-cli/src/context.rs`** — the body of `resolve_ledger_mode` is
  refactored into `resolve_query_target` returning a `QueryTarget { Ledger | GraphSource }`.
  A `lookup_graph_source` probe runs **after** the native-ledger check
  (ledger-wins), normalizing the bare name to `<name>:main`, preferring a local
  registration, and skipping retracted sources. `resolve_ledger_mode` becomes a
  thin wrapper that maps a graph-source name to a clear `NotFound` for the ~10
  other commands that still only handle ledgers (no behavior change for real
  ledger names). The graph-source probe is error-tolerant: a graph-source-store
  read error is debug-logged and treated as "not a graph source" (fall through),
  so it cannot break native-ledger resolution for the commands that call
  `resolve_ledger_mode`.
- **`fluree-db-cli/src/commands/query.rs`** — graph-source execution:
  - Implicit `fluree query <graph-source>` runs **locally** via
    `fluree.graph(alias).query()…execute_formatted()` (gated on the `iceberg`
    feature; a clear "rebuild with `--features iceberg`" message otherwise).
  - A query whose `FROM` / `from` references a **foreign** graph source
    auto-routes to the connection path — `query_from` (local) or
    `query_connection_sparql` / `query_connection_jsonld` (remote), wiring methods
    that were previously unreachable.
  - A bare `--connection` federation query does **not** require an active native
    ledger: when no ledger is active/provided, the local connection path falls back
    to a bare `Fluree` handle (symmetric with the `--connection=REMOTE` arm), since
    the query's `FROM` already names its sources. A normal `fluree query` with no
    ledger still errors `NoActiveLedger` (the relaxation is gated on the connection
    path), and a provided ledger still resolves normally.
  - Guardrails reject flags unsupported on the graph-source single-target path with
    a clear message.
- **`fluree-db-cli/src/cli.rs`** — adds `--connection[=REMOTE]` to the **Query**
  command only, using `require_equals` so a bare `--connection` forces the
  connection path while `--connection=REMOTE` targets a named remote (and neither
  swallows the following positional query string).
- **`fluree-db-cli/src/lib.rs`** — threads `connection` into `query::run`.
- **`docs/cli/iceberg.md`** — fixes the broken example at line 160 (single-target
  form + a `--connection` federation example; and `fluree info <gs>` →
  `fluree iceberg info <gs>`, since bare `info` on a graph-source name now returns
  the clear graph-source error).

## Design decisions (intentional deviations from #1398's literal wording)

1. **`QueryTarget` enum instead of a new `LedgerMode` variant.** ~10 command files
   exhaustively match `LedgerMode::{Local,Tracked}`; adding a variant/field there
   would break all of them. The probe still lives in `context.rs` (per the issue's
   critical-files list) and query behavior is exactly as specified; other commands
   simply get a clearer error for a graph-source name.
2. **`--connection[=REMOTE]` with `require_equals`** resolves the issue's ambiguous
   "positional optional" and prevents the flag from eating the query string.
3. **Implicit `fluree query <gs>` executes locally** (`graph().query()`), matching
   the issue's primary "execute locally" design; server-routed forms remain
   reachable via `--remote` / `--connection`.
4. **`--policy` / `--track` are not plumbed** into the graph-source single-target
   path — consistent with the server's graph-source fallback, which doesn't apply
   them either. Noted as a minor limitation, not a silent surprise.
5. **No live-rows E2E test** — a queryable Iceberg/R2RML source needs a live
   catalog (the issue itself defers that). Covered instead by a real
   graph-source *resolution* test plus deterministic unit tests of every routing
   decision.

## Tests

`context::tests` (register a real graph source via the nameservice publisher — no
live catalog):
- `resolves_registered_graph_source_to_graph_source_target`
- `ledger_only_resolution_reports_graph_source_as_clear_error`
- `missing_target_still_reports_not_found`

`commands::query::tests`:
- `base_ledger_id_strips_branch_time_and_fragment`
- `jsonld_from_targets_handles_string_and_array`
- `foreign_source_detection_sparql`, `foreign_source_detection_jsonld`
- `graph_source_guardrails_reject_unsupported_flags`
- `json_display_format_falls_back_for_jsonld`

`tests/integration.rs`:
- `query_connection_remote_routes_to_named_remote`
- `query_bare_connection_keeps_positional_query`
- `query_bare_connection_without_active_ledger_routes_to_connection_path` — bare
  `--connection` with no active ledger routes onward instead of erroring
  `NoActiveLedger`.

### Results

- `cargo build -p fluree-db-cli --features iceberg` — success.
- `cargo test -p fluree-db-cli --features iceberg` — 110 lib + 102 integration
  passed, 0 failed.
- `cargo clippy -p fluree-db-cli --all-targets --features iceberg` — zero warnings.
- `cargo fmt -p fluree-db-cli -- --check` — clean.
- `cargo build -p fluree-db-cli --no-default-features --features "server,shacl,aws"`
  (non-iceberg path) — success (exercises the `cfg(not(feature="iceberg"))`
  rebuild-message branch).

Fixes #1398

### DIFF
--- a/docs/cli/iceberg.md
+++ b/docs/cli/iceberg.md
@@ -154,10 +154,16 @@ Once mapped, the graph source appears in standard commands:
 fluree list
 
 # Inspect configuration
-fluree info warehouse-orders
+fluree iceberg info warehouse-orders
 
-# Query via SPARQL GRAPH pattern
-fluree query mydb 'SELECT ?id ?total FROM <mydb:main> WHERE { GRAPH <warehouse-orders:main> { ?o ex:id ?id ; ex:total ?total } }'
+# Query the graph source directly (single-target): name it as the target, with
+# no FROM. The CLI resolves it as a graph source and runs it through the
+# R2RML/Iceberg engine. Requires the `iceberg` feature.
+fluree query warehouse-orders 'SELECT ?id ?total WHERE { ?o ex:id ?id ; ex:total ?total }'
+
+# Federate into it: a query whose FROM names the source routes to the
+# connection-scoped path automatically (or force it with --connection).
+fluree query --connection 'SELECT ?id ?total FROM <warehouse-orders:main> WHERE { ?o ex:id ?id ; ex:total ?total }'
 
 # Remove the mapping
 fluree drop warehouse-orders --force
--- a/fluree-db-cli/src/cli.rs
+++ b/fluree-db-cli/src/cli.rs
@@ -609,6 +609,24 @@ pub enum Commands {
         #[arg(long)]
         remote: Option<String>,
 
+        /// Force the connection-scoped query path (server root `/query`, with the
+        /// dataset carried by the query's own `FROM` clause) instead of the
+        /// ledger-scoped path. Use this to query a registered Iceberg/R2RML graph
+        /// source by federating into it, or to run a cross-source `FROM` query.
+        ///
+        /// Optionally name a remote that hosts the connection
+        /// (`--connection=origin`); given with no value, the local connection (or
+        /// an auto-routed local server) is used. Queries that already reference a
+        /// non-endpoint source via `FROM` route here automatically.
+        #[arg(
+            long,
+            value_name = "REMOTE",
+            num_args = 0..=1,
+            default_missing_value = "",
+            require_equals = true
+        )]
+        connection: Option<String>,
+
         /// Report tracking tally (fuel + time + policy) to stderr after results.
         /// Shorthand for `--track-fuel --track-time --track-policy`.
         #[arg(long)]
--- a/fluree-db-cli/src/commands/query.rs
+++ b/fluree-db-cli/src/commands/query.rs
@@ -227,6 +227,7 @@ pub async fn run(
     at: Option<&str>,
     dirs: &FlureeDir,
     remote_flag: Option<&str>,
+    connection: Option<&str>,
     direct: bool,
     tracking: TrackingFlags,
     policy: &PolicyArgs,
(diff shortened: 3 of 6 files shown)


bplatz (approved):
🖱️

PR #1406, opened 2026-07-01
perf: Iceberg/R2RML graph-source performance — planning, catalog caching, LIMIT + row-group pruning

bplatz:
## Summary

Performance work on the Iceberg/R2RML graph-source read path, from a live diagnosis against Snowflake Horizon (`ENTERPRISE_DEMO.DW`). Correct-but-slow queries — a 10-row join took ~4 minutes with 36 Snowflake round-trips — traced to R2RML query planning amplifying REST and Parquet costs, **not** to IO (every data read in every trace was a disk-cache hit). Each change is independently shippable and kill-switchable.

## What's here

**Query planning**
- Fuse `rdf:type` into the same-subject star + prune subject-only scans: a subject-only/`rdf:type` pattern no longer projects every POM column, scans RefObjectMap parent tables it never reads, or fans out to predicate-sharing TriplesMaps. Simple Store query 6 scans → 1; Store ⋈ Geography 36 → 5.
- Reuse the inner scan across child batches: a correlated join no longer re-scans the inner dimension table per child batch. Store ⋈ Geography 5 → 3 scans.

**Catalog access**
- Query-scoped Iceberg catalog session: one REST client + one `loadTable` response reused across a query's scans, with the Iceberg snapshot pinned. OAuth 36 → 1, `loadTable` 36 → 2 within a query.
- Pin `metadata_location` across a mid-query credential refresh so a creds refresh cannot shift the query onto a newer snapshot.
- Process-wide OAuth token reuse + cross-query `loadTable` cache (60s TTL, vended-credential-expiry gated): on a warm server the 2nd query against a table skips OAuth + `loadTable` entirely.

**Scan execution**
- LIMIT early-termination: a `LIMIT n` now reaches the scan (`GraphOperator` forwards the row budget through the R2RML subplan), so it stops after ~one materialize window instead of draining the table.
- Row-group pruning via Parquet statistics: the reader skips row groups whose min/max rule out a pushed date/int/bool filter. Conservative — an unsupported or decimal-typed column keeps the group; results are unchanged.

## Live results (Snowflake Horizon; results byte-identical to baseline)

| Query | Before | After |
|---|---|---|
| `Store; storeId; name` LIMIT 5 | 6 scans / 19.6s | 1 scan |
| `Store ⋈ Geography` LIMIT 10 | 36 scans / 229s | 3 scans / ~9s · OAuth 36→1 · loadTable 36→2 |
| `FACT_ORDER_LINE` LIMIT 5 | >90s (was timing out) | ~4.4s |
| warm server, 2nd identical query | 3.0s | 0.2s |

## Config (all default-on, kill-switchable)

Documented in `docs/operations/configuration.md`: `FLUREE_ICEBERG_LOADTABLE_CACHE`, `FLUREE_ICEBERG_LOADTABLE_TTL_SECS` (default 60), `FLUREE_R2RML_SCAN_CACHE`, `FLUREE_R2RML_LIMIT_PUSHDOWN`, `FLUREE_ICEBERG_PREDICATE_PUSHDOWN`, plus the existing `FLUREE_R2RML_MATERIALIZE_WINDOW_ROWS` and `FLUREE_ICEBERG_SCAN_CONCURRENCY`.

## Testing

- `cargo fmt --all -- --check` clean; `cargo clippy --all --all-features --all-targets` clean.
- Guardrails: scan-count / projection / dangling-FK / wildcard invariants; catalog-cache hit + credential-expiry + key isolation; snapshot-pin-across-refresh; LIMIT early-termination (a `LIMIT 5` pulls ~1 of 40 streamed chunks); row-group pruning against a real two-row-group Parquet fixture.
- Live A/B on the `enterprise` graph source for every change, with byte-identical results pushdown on vs off.

## Not included

Row-level residual filtering + planner FILTER-consumption (so a `FILTER + LIMIT` fact query early-terminates) is deferred. Live testing surfaced that `xsd:integer` R2RML columns are physically Iceberg `Decimal`, and a residual filter needs conservative "keep on unevaluable" semantics distinct from the existing row matcher — a follow-up, scoped in the branch notes.

### DIFF
--- a/docs/operations/configuration.md
+++ b/docs/operations/configuration.md
@@ -1003,6 +1003,7 @@ reasonable behavior out of the box.
 |----------|---------|---------|
 | `FLUREE_ICEBERG_LOADTABLE_CACHE` | on | Master switch for all REST catalog caching. Set to `0`/`false`/`off` to build a fresh catalog client and reload the table on **every** scan (restores a per-scan OAuth exchange + `loadTable` round-trip). Disables the client/OAuth reuse, the cross-query `loadTable` cache, and the per-query snapshot pin. |
 | `FLUREE_ICEBERG_LOADTABLE_TTL_SECS` | `60` | TTL (seconds) for the **cross-query** `loadTable`-response cache. A REST `loadTable` GET against a catalog such as Snowflake Horizon costs ~1.3–3 s, so caching it lets a burst of queries against the same table skip the round-trip. The TTL bounds how stale a snapshot a *new* query may observe; `0` disables the cross-query layer (leaving only the per-query pin). Every cache read is additionally gated on vended-credential expiry (30 s buffer), so a long TTL never hands out about-to-expire credentials. |
+| `FLUREE_ICEBERG_REST_CLIENT_TTL_SECS` | `900` | TTL (seconds) for the process-wide REST **catalog-client** cache (the reused OAuth token + HTTPS pool). The cache is keyed by a fingerprint of the raw config JSON, which does **not** change when a secret referenced by env var / secret store is rotated; this TTL bounds how long a rotated secret stays stale before the client is rebuilt and re-authenticated. `0` rebuilds the client every query (restoring a per-query OAuth exchange). |
 | `FLUREE_R2RML_SCAN_CACHE` | on | Toggles the correlated-join inner-scan cache, which reuses a materialized inner (dimension) table across a join's child batches instead of re-scanning it per batch. Set to `0`/`false`/`off` to restore per-child-batch re-scans. |
 | `FLUREE_R2RML_MATERIALIZE_WINDOW_ROWS` | `524288` | Target table rows materialized into RDF-term bindings per parallel window. Bounds resident memory during a scan (materialization explodes the compact columnar form into fat binding rows). Also caps the size of an inner scan eligible for the inner-scan cache above. |
 | `FLUREE_ICEBERG_SCAN_CONCURRENCY` | `min(cores, files, 8)` | Number of data files read concurrently within one scan. Raise it for high-latency remote object stores (it is not capped, but is bounded by the number of files in the scan). |
@@ -1017,7 +1018,12 @@ second and later queries against a table typically skip both the OAuth exchange
 and the `loadTable` GET entirely. The cross-query cache always records the
 catalog's current state, never a query's pinned snapshot, so pin preservation
 cannot leak a stale location to other queries. The client cache is keyed by a
-config fingerprint, so rotating the source's access token rebuilds the client.
+fingerprint of the raw config JSON: editing the config (including an inline
+secret) rebuilds the client immediately, but a secret referenced by env var or
+secret store is invisible to the fingerprint, so a rotation of that secret is
+picked up only when the client cache entry expires
+(`FLUREE_ICEBERG_REST_CLIENT_TTL_SECS`, default 15 min) — until then the reused
+client keeps presenting the old credential.
 
 **Freshness vs. latency.** The only knob with a data-freshness tradeoff is
 `FLUREE_ICEBERG_LOADTABLE_TTL_SECS`: a new query may read a snapshot up to that
--- a/docs/operations/configuration.md
+++ b/docs/operations/configuration.md
@@ -993,6 +993,38 @@ Iceberg table: ~62s with the system allocator vs ~31s with mimalloc).
 Pair it with bounded query parallelism so a single large scan cannot monopolize
 cores and allocator pressure under concurrent load.
 
+## Iceberg / R2RML Graph-Source Tuning
+
+Queries against an Iceberg-backed R2RML graph source are tuned by a set of
+environment-only knobs. All are optional; the defaults are chosen for correct,
+reasonable behavior out of the box.
+
+| Variable | Default | Purpose |
+|----------|---------|---------|
+| `FLUREE_ICEBERG_LOADTABLE_CACHE` | on | Master switch for all REST catalog caching. Set to `0`/`false`/`off` to build a fresh catalog client and reload the table on **every** scan (restores a per-scan OAuth exchange + `loadTable` round-trip). Disables the client/OAuth reuse, the cross-query `loadTable` cache, and the per-query snapshot pin. |
+| `FLUREE_ICEBERG_LOADTABLE_TTL_SECS` | `60` | TTL (seconds) for the **cross-query** `loadTable`-response cache. A REST `loadTable` GET against a catalog such as Snowflake Horizon costs ~1.3–3 s, so caching it lets a burst of queries against the same table skip the round-trip. The TTL bounds how stale a snapshot a *new* query may observe; `0` disables the cross-query layer (leaving only the per-query pin). Every cache read is additionally gated on vended-credential expiry (30 s buffer), so a long TTL never hands out about-to-expire credentials. |
+| `FLUREE_R2RML_SCAN_CACHE` | on | Toggles the correlated-join inner-scan cache, which reuses a materialized inner (dimension) table across a join's child batches instead of re-scanning it per batch. Set to `0`/`false`/`off` to restore per-child-batch re-scans. |
+| `FLUREE_R2RML_MATERIALIZE_WINDOW_ROWS` | `524288` | Target table rows materialized into RDF-term bindings per parallel window. Bounds resident memory during a scan (materialization explodes the compact columnar form into fat binding rows). Also caps the size of an inner scan eligible for the inner-scan cache above. |
+| `FLUREE_ICEBERG_SCAN_CONCURRENCY` | `min(cores, files, 8)` | Number of data files read concurrently within one scan. Raise it for high-latency remote object stores (it is not capped, but is bounded by the number of files in the scan). |
+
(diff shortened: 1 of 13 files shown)


docs/operations/configuration.md
```diff
@@ -993,6 +993,38 @@ Iceberg table: ~62s with the system allocator vs ~31s with mimalloc).
...
+the first scan of a table pins its `metadata_location`, so every scan in that
+query reads one consistent Iceberg snapshot even if the table commits mid-query.
+Across queries, a process-wide cache reuses the REST client (so its OAuth token —
+valid ~1 h — and HTTPS connection pool survive) and the `loadTable` response
+(bounded by `FLUREE_ICEBERG_LOADTABLE_TTL_SECS`). On a warm server this means the
+second and later queries against a table typically skip both the OAuth exchange
+and the `loadTable` GET entirely. The cross-query cache always records the
+catalog's current state, never a query's pinned snapshot, so pin preservation
+cannot leak a stale location to other queries. The client cache is keyed by a
+config fingerprint, so rotating the source's access token rebuilds the client.
```
aaj3f: This is less a doc comment and more of a claim in the docs that I think sounds good to me, but I'm not sure it's right vis-a-vis the code in `fluree-db-api/src/graph_source/r2rml.rs` + `cache.rs`.

It seems `config_fingerprint` hashes only the raw config JSON, while an OAuth2/Bearer secret is a `ConfigValue`, which in practice is `Dynamic { env_var: "…" }`. The JSON stores the env var name, not the secret. An operator might rotate the secret, but `record.config` is unchanged --> fingerprint unchanged --> `cache.rest_client(&fp)` returns the **stale** `Arc<RestCatalogClient>` holding the old secret and its cached OAuth token. `rest_clients` is `moka::sync::Cache::new(64)` with **no TTL** (`cache.rs:145`), so the stale client survives until LRU eviction or process restart. When the cached OAuth token expires (~1h) the client refreshes with the stale secret --> 401s process-wide. Only literal-in-config secrets get the behavior the comment advertises.

We could fold the resolved secret into the fingerprint, or put a TTL on `rest_clients` so it self-heals

aaj3f (approved):
Thanks, @bplatz! Excited for these wins!

I have one in-line comment below, and then I'm going to provide some feedback direct-from-Claude here, but otherwise things look good!

````
### 1. `fluree-db-query/src/r2rml/rewrite.rs:159-166` (and `:181-184`) — CORRECTNESS: class fusion silently empties a legal cross-TriplesMap query

Fusing a `?s a ex:C` pattern into a same-subject star sets `base.class_filter =
Some(class)` on a base that already carries `predicate_filter`. In
`operator.rs::build_progress` (~lines 408-428) TriplesMap resolution then requires
**one** TriplesMap to satisfy *both* the class **and** the predicate:

```rust
if let Some(ref class_filter) = self.pattern.class_filter {
    if !tm.classes().contains(class_filter) { return false; }
}
if let Some(ref pred_filter) = self.pattern.predicate_filter {
    if !tm.predicate_object_maps.iter().any(|pom|
        pom.predicate_map.as_constant() == Some(pred_filter.as_str())) { return false; }
}
```

Same subject **variable** does not imply same TriplesMap. A legal R2RML mapping can
vertically partition a subject: `TM_A { subject person/{id}, rr:class ex:Person }`
and `TM_B { subject person/{id}, predicate ex:name }`. For
`?s a ex:Person ; ex:name ?name`:

- **Before this PR:** the type pattern was a standalone subject-only R2RML scan
  (resolves `TM_A`) and the name pattern a standalone scan (resolves `TM_B`), joined
  on `?s` — correct.
- **After this PR:** the name star fuses `class_filter=Person`, so `build_progress`
  needs one TM with class Person **and** predicate `ex:name`. `TM_A` fails (no name),
  `TM_B` fails (no class) → `triples_maps.is_empty()` → **the query returns zero
  rows.**

This is a silent wrong-result regression, not a crash. It's gated on the
class and the queried predicate living in different TriplesMaps that share a
subject template — uncommon in the tested star-schema dataset (dimensions joined via
RefObjectMap, not shared-subject partitioning), but valid R2RML.

**Suggested fix:** only fuse the class when exactly one TriplesMap resolves the
star predicate **and** declares the class; otherwise leave the class as a joined
subject-only scan (the pre-fusion path, already correct). A guard test covering the
split-TriplesMap shape would lock this down.

---

### 2. `fluree-db-query/src/r2rml/operator.rs:673-674` — PERF: a small LIMIT over a selective *internal* R2RML join scans the whole table in tiny windows

`window_rows = materialize_window_rows().min(b.saturating_sub(self.emitted).max(1))`
is computed **once** per `ScanProgress` and never grows. For the common R2RML case
(single seed child → one `ScanProgress`), a `LIMIT 5` fixes the materialize window at
5 rows for the entire scan.

When the scan's produced rows feed an internal join (RefObjectMap / correlated child)
that filters most of them out, the output budget is never met, so the operator keeps
pulling **5-row windows** across the whole table — many small rayon materialize passes
instead of the full 512K-row windows the un-budgeted path uses. On `main` this same
query drains the table in big windows; post-PR it drains it in tiny ones, so `LIMIT`
can make this shape **slower**. Correctness is unaffected.

**Suggested fix:** floor the window at a reasonable minimum (e.g.
`materialize_window_rows().min(max(budget_remaining, MIN_WINDOW))`) or grow the window
geometrically across iterations when the budget isn't being met. Alternatively, note
the pathology next to the `FLUREE_R2RML_LIMIT_PUSHDOWN` kill-switch.

---

### 3. `fluree-db-iceberg/src/scan/pruning.rs:279-282` — LATENT CORRECTNESS: row-group pruning has no Decimal-logical-type guard

`stat_bounds` keys purely on the Parquet **physical** variant
(`Statistics::Int64` etc.), with no check of the column's logical/converted type. If
a column is stored as INT64 with a *decimal* logical type (unscaled integers) and the
R2RML `xsd:integer → Decimal` coercion emits an `Int64` filter literal, pruning
compares the scaled query value against unscaled bounds → can drop a matching row
group (e.g. Decimal(10,2) storing `5.00` as `500`, filter `= 5`, bounds `[500,504]` →
`Eq` prunes, matching row vanishes). Currently safe only because the Iceberg spec
mandates `fixed_len_byte_array` for decimals (→ `(None,None)` → conservative keep) —
the PR's own "Not included" note flags this xsd:integer↔Decimal mismatch as live.

**Suggested fix:** skip pruning (return `(None,None)`) when the Parquet column's
`LogicalType`/`ConvertedType` is Decimal, so the safety no longer rests on an
implicit encoding assumption.

---

### 4. `fluree-db-api/src/graph_source/r2rml.rs:725-781` — MINOR/UNLIKELY: snapshot pin can slip under concurrent first-load of the same table

The pin is deduped in `store_load_table` (first insert wins), but the caller scans
with its own local `resp.metadata_location`, not the pinned value. If two scans of the
same table both miss the per-query pin and the cross-query cache and both issue a real
`load_table` concurrently, scan A pins `A_loc` while scan B proceeds with `B_loc`; a
mid-query commit then yields two snapshots in one query. Almost certainly unreachable
given the pull-based, sequential operator execution, but worth a defensive re-read of
the pin after `store_load_table` if you want the invariant to hold unconditionally.

---

### 5. `fluree-db-query/src/r2rml/rewrite.rs:162-164, 182-184` — CLEANUP: duplicated fusion block

```rust
if let Some(class) = take_single_class(&mut class_groups, subject) {
    base.class_filter = Some(class);
}
```

appears verbatim in both the single-member and multi-member star arms. Fold into a
small helper (`fuse_class(&mut base, &mut class_groups, subject)`) once `base` is
chosen. Minor.

docs/operations/configuration.md
bplatz: Fixed in e738077464bc5b85cc9a3afbf33ec8f3219186cf: the client cache now has a TTL (`FLUREE_ICEBERG_REST_CLIENT_TTL_SECS`, default 15m) so an env-var/secret-store secret rotation self-heals rather than serving 401s until eviction. Also corrected this doc line and the `config_fingerprint` / `rest_clients` docstrings to stop claiming rotation rebuilds the client — the fingerprint only sees inline config, so the TTL is what bounds staleness.

bplatz:
All review points addressed in e738077464bc5b85cc9a3afbf33ec8f3219186cf:

1. **Class fusion silent empty result** — fuse `?s a ex:C` into a star only when every TriplesMap resolving the base predicate also declares the class (`class_fusion_is_safe`); the split-TriplesMap shape now stays as a separate subject-only scan joined on the subject. Adds a unit test and an end-to-end `engine_e2e_split_triples_map_class_and_predicate_not_fused` guard (verified red pre-fix, green post-fix).
2. **Small LIMIT → tiny materialize windows** — the budgeted window now grows geometrically each pass, ramping to the full size after a few low-yield passes while a selective LIMIT still stops after its cheap first window.
3. **Decimal row-group pruning** — pruning is skipped when the Parquet column has a Decimal logical/converted type, so correctness no longer rests on the fixed_len_byte_array encoding assumption.
4. **Snapshot pin under concurrent first-load** — the pinned `metadata_location` is re-read and adopted after `store_load_table`, so two scans in one query cannot diverge onto different snapshots.
5. **Duplicated fusion block** — folded into a single `fuse_class_if_safe` helper.

(The inline config-fingerprint note is answered in its thread — REST client cache now has a TTL.)

PR #1407, opened 2026-07-01
(Iceberg/R2RML) Deterministic Catalog --> R2RML Engine

aaj3f:
## Summary

This PR gives the fluree/db engine everything needed to turn a live Iceberg catalog (Snowflake Horizon / Apache Polaris REST, or direct S3) into a **deterministically generated, queryable R2RML mapping** — with no hand-written R2RML and no LLM in the happy path. It is the engine-side foundation of the [Iceberg → R2RML → Dataset epic](https://github.com/fluree/solo/issues/728): Fluree Solo's backend (next PR) will call this API so a user can connect a catalog, preview its tables, auto-generate a mapping, review it, and save it as a virtual (query-in-place) Dataset.

Everything here is **additive and metadata-only** (no Parquet/data scans), gated behind the existing `iceberg` / `aws` / `turtle` feature flags, and it does not touch the query scan path or any performance-critical files.

Closes fluree/solo#724.

## What's included

Four new, independently testable capabilities plus one supporting refactor:

- **Catalog browse** — wires the (previously present but unused) `list_namespaces` / `list_tables` catalog primitives into a read-only API.
- **Table preview (Tier-A schema + Tier-B stats)** — a metadata-only preview returning the table schema (columns, types, `identifier_field_ids`, partition/sort specs, snapshot + history, properties, row/file/byte counts) and per-column statistics (null/value counts, NaN counts, typed min/max, on-disk bytes) aggregated from Avro manifests. No Parquet is read.
- **Deterministic R2RML generation** — a greenfield metadata→R2RML emitter and structured→Turtle renderer (the `fluree-db-r2rml` crate was parse/compile-only before this). It builds Solo's `StructuredR2rmlMapping` IR from the preview, renders Turtle, and round-trips that Turtle back through the existing `R2rmlLoader` as a correctness check.
- **Validate / dry-run** — compiles an R2RML mapping through the loader and cross-checks every table/column/join reference and subject-key nullability against the live schema + stats, returning structured diagnostics, without creating a graph source.
- **`IcebergConnectionConfig` refactor** — factors the reusable connection block (catalog URI / warehouse / auth / IO) out of `IcebergCreateConfig` so browse/preview/generate/validate can all run against an unsaved connection during onboarding. Behavior-identical for the existing create path.

## New public API (the surface the Solo backend PR will bind)

Rust methods on `Fluree` (feature `iceberg`), mirrored by thin HTTP wrappers in `fluree-db-server`:

```
Fluree::browse_iceberg_catalog(conn, depth)       -> CatalogBrowse
Fluree::preview_iceberg_table(conn, table, tier)  -> TablePreview          // tier = Schema | Stats
Fluree::generate_r2rml(request)                   -> GenerateR2rmlResponse
Fluree::validate_r2rml(conn, turtle, snapshot)    -> ValidateR2rmlResponse
```

```
POST /v1/fluree/iceberg/catalog/browse
POST /v1/fluree/iceberg/catalog/preview
POST /v1/fluree/iceberg/r2rml/generate
POST /v1/fluree/iceberg/r2rml/validate
```

All four accept an inline `IcebergConnectionConfig` (so onboarding can browse/preview before anything is saved) and are read-semantic — no graph source is created and nothing is published. `generate_r2rml` returns `{ turtle, structured, diagnostics, snapshot_id }`, where `structured` is Solo's camelCase `StructuredR2rmlMapping` — the authoritative wire IR that the Solo backend persists, the UI reviews, and Phase-2 target-matching later rewrites. There are no pre-save sample triples (Phase-1 is metadata-only; real triples come from a post-save `LIMIT` / Explore query).

## How generation works (deterministic, no LLM)

Per table, the emitter:

- **Picks the subject key** from Iceberg's declared `identifier_field_ids` (or an explicit per-table `primary_key` override). A key must be non-null (`required` or `null_fraction == 0`); a name-based fallback earns a `SubjectKeyUnverified` diagnostic; if none is safe it emits `NoSafeSubjectKey` and no subject — it never invents a surrogate row id.
- **Emits one `rr:TriplesMap` per table**, keeping filterable scalar columns (keys, dates, partition/sort columns) as literal predicate-object maps for pushdown.
- **Infers foreign keys** from Iceberg metadata alone: a candidate column must match a target table's primary-key name (exact, or an unambiguous `_<PK>` suffix) **and** type **and** pass typed min/max range-containment. An exactly-one-survivor becomes an `rr:parentTriplesMap` + `rr:joinCondition`; ambiguous or unresolved key-like columns are surfaced as diagnostics and kept as literals — a join is never fabricated.
- **Renders deterministic Turtle** from the structured IR and round-trips it through `R2rmlLoader::from_turtle().compile()` as an internal correctness check. Two runs over the same snapshot produce byte-identical output.

## Correctness properties (enforced and tested)

- **Every non-string column carries an `rr:datatype`.** Runtime materialization does no datatype inference, so omitting it would silently produce plain-string literals for numerics/temporals. Strings are intentionally left untyped.
- **`bytes → xsd:hexBinary`** (not base64). The materializer's byte formatter emits lowercase hex despite its misleading name; a regression test couples the emitter's datatype choice to the materializer's actual output so that a future "fix" fails loudly.
- **Foreign keys are only ever emitted as `rr:parentTriplesMap` + `rr:joinCondition`**, never as a bare templated IRI built from the child key — so a dangling FK yields no triple (correct R2RML RefObjectMap semantics).
- **Byte-for-byte column casing** on `rr:tableName` / `rr:column` / template placeholders (Snowflake folds identifiers to uppercase; a mismatch would silently scan nothing).
- **Metadata-only.** Browse/preview/generate/validate read REST `loadTable` plus Avro manifest-lists and manifests only; a test asserts that no `.parquet` file is ever fetched.

## Foreign-key coverage (honest)

Against the 16-table `ENTERPRISE_DEMO.DW` reference mapping (33 FK joins), the emitter deterministically reproduces **29 of 33**:

- The **4 role-renamed** FKs (`MANAGER_KEY`, `REGION_MANAGER_KEY`, `SALES_REP_KEY`, `AGENT_KEY` → `EMPLOYEE_KEY`) are not resolvable from metadata (the column name doesn't match the target PK, and overlapping surrogate-key ranges can't disambiguate). They are surfaced as `UnresolvedFkCandidate` and kept as literal columns — **not fabricated** — and get resolved later by Phase-2 target-model matching.
- The **3 child-fact → `FACT_ORDER` hub joins** (`ORDER_KEY → ORDER_KEY`) are emitted, each flagged with a `FactHubJoinAdvisory` performance advisory.

## Known limitations / deferred

- **Subject-key uniqueness is unverifiable from metadata** (no NDV — `distinct_counts` / Puffin sketches are not parsed). Generation relies on `identifier_field_ids` or a required key and flags `SubjectKeyUnverified` when falling back; true uniqueness verification (via NDV or a `COUNT(DISTINCT)` pushdown) is a fast-follow.
- **Snapshot pinning is per-table.** `preview_iceberg_table` reads each table's current snapshot, and `generate_r2rml` returns the first requested table's snapshot as a representative `snapshot_id` (documented as such), not a true catalog-wide pin. A follow-up can add an explicit snapshot selection to preview.
- **Secret-at-rest is unchanged here.** Inbound connection tokens are still wrapped as literals in the persisted graph-source config; moving them to a reference/vault is the Solo backend PR's concern (and a later hardening pass).

## Testing

All green across the relevant feature gates:

| Suite | Result |
|---|---|
| `cargo test -p fluree-db-r2rml --features turtle` | 96 passed |
| `cargo test -p fluree-db-iceberg --features aws --lib` | 159 passed |
| `cargo test -p fluree-db-api --features iceberg,aws --lib` | 671 passed |
| `cargo test -p fluree-db-server --features iceberg --lib routes::iceberg` | 6 passed |

Coverage highlights: the emitter's structural match to the `ENTERPRISE_DEMO.DW` reference mapping (29/33 joins, including the role-renamed and hub cases), the Turtle round-trip through `R2rmlLoader`, byte-determinism, the hex-not-base64 guard, the Tier-B "manifests only, never Parquet" guard, and the validate cross-check across every reference class (table / column / casing / join-type / nullable subject key). `cargo check -p fluree-db-api` with default (non-iceberg) features still compiles, and both `cargo fmt --all -- --check` and clippy are clean.

## Scope and safety

- **Additive.** New modules under `fluree-db-iceberg` (Tier-B stats), `fluree-db-api` (browse / preview / generate / validate), and `fluree-db-r2rml/src/emit` (the emitter). The only edits to existing code are the `IcebergConnectionConfig` extraction (behavior-identical), retaining the previously-discarded inline `loadTable` metadata, exposing `emit::naming`, and route registration.
- **No scan-path or performance-critical files are touched** — this is deliberately kept out of the Iceberg query/scan hot path and the ongoing performance work.

## Not in this PR (tracked separately under the epic)

- **Solo backend (fluree/solo#725):** promote the contracts to `@fluree/contracts`, call this API, persist the mapping, wire catalog-connect + OAuth2 scope + secret handling, and the Dataset↔mapping binding.
- **Solo UI (fluree/solo#726):** the onboarding wizard, structured-form review, and save-as-Dataset.
- **Phase-2 (fluree/solo#727):** target-model matching, which resolves the role-renamed FKs.
- **Fast-follow:** NDV / Puffin reader (subject-key uniqueness) and secret-vault hardening.

## How to review

The branch is a linear-ish build-up you can read part by part:

- `refactor(api): factor IcebergConnectionConfig …` — the connection-type extraction (start here).
- `feat(api): catalog browse …` / `feat(api): Tier-A metadata preview …` / `feat(iceberg,api): Tier-B per-column stats …` — the read/preview API.
- `feat(r2rml): deterministic Iceberg metadata → R2RML emitter …` + `feat(r2rml): per-table primary_key/class_name overrides …` — the emitter (`fluree-db-r2rml/src/emit`).
- `feat(api,server): wire end-to-end generate_r2rml …` + `feat(iceberg,api): R2RML validate/dry-run …` — the generate/validate surface + routes.
- `fix(iceberg): dedicated DiagCode::CompileError …` and `style(iceberg): rustfmt …` — a diagnostics-code fix and formatting.

### DIFF
--- a/fluree-db-api/src/graph_source/config.rs
+++ b/fluree-db-api/src/graph_source/config.rs
@@ -1208,6 +1208,41 @@ mod tests {
 
     #[cfg(feature = "iceberg")]
     #[test]
+    fn debug_of_connection_config_redacts_secrets() {
+        // A future `{:?}` on the connection (in a log or error) must not leak the
+        // bearer token or the OAuth2 client secret it transitively holds via the
+        // `AuthConfig` -> `ConfigValue` chain.
+        let bearer = IcebergConnectionConfig::rest("https://catalog.example.com")
+            .with_auth_bearer("super-secret-bearer-token");
+        let dbg = format!("{bearer:?}");
+        assert!(
+            !dbg.contains("super-secret-bearer-token"),
+            "bearer token leaked in Debug: {dbg}"
+        );
+
+        let oauth = IcebergConnectionConfig::rest("https://catalog.example.com").with_auth_oauth2(
+            "https://catalog.example.com/v1/oauth/tokens",
+            "client-id-ok-to-show",
+            "super-secret-oauth-secret",
+        );
+        let dbg = format!("{oauth:?}");
+        assert!(
+            !dbg.contains("super-secret-oauth-secret"),
+            "oauth client_secret leaked in Debug: {dbg}"
+        );
+
+        // The same guarantee must hold one level up, on IcebergCreateConfig,
+        // whose derived Debug prints the connection.
+        let create = IcebergCreateConfig::new("gs", "https://catalog.example.com", "ns.tbl")
+            .with_auth_bearer("super-secret-bearer-token");
+        assert!(
+            !format!("{create:?}").contains("super-secret-bearer-token"),
+            "bearer token leaked in IcebergCreateConfig Debug"
+        );
+    }
+
+    #[cfg(feature = "iceberg")]
+    #[test]
     fn test_oauth2_scope_setter_warns_without_oauth2() {
         // Bearer auth set, scope setter should be a no-op (and not panic).
         let config = IcebergCreateConfig::new("gs", "https://catalog.example.com", "ns.tbl")
--- a/fluree-db-api/src/graph_source/config.rs
+++ b/fluree-db-api/src/graph_source/config.rs
@@ -400,17 +400,29 @@ pub struct IcebergCreateConfig {
     /// Branch name (defaults to "main")
     pub branch: Option<String>,
 
+    /// The reusable connection block (catalog access + IO).
+    pub connection: IcebergConnectionConfig,
+
+    /// Table identifier (e.g., "openflights.airlines"). Empty for Direct mode
+    /// (derived from the table location).
+    pub table_identifier: String,
+}
+
+/// The reusable Iceberg connection block — catalog access + IO, with **no**
+/// table or mapping.
+///
+/// Factored out of [`IcebergCreateConfig`] so catalog browse / metadata preview
+/// can run against an **unsaved** connection during onboarding (before a graph
+/// source record is created). The relationship is:
+/// `IcebergCreateConfig` = `IcebergConnectionConfig` + `table_identifier`.
+#[cfg(feature = "iceberg")]
+#[derive(Debug, Clone)]
+pub struct IcebergConnectionConfig {
     /// Catalog mode: REST or Direct S3 access.
     pub catalog_mode: CatalogMode,
 
-    /// S3 region override
-    pub s3_region: Option<String>,
-
-    /// S3 endpoint override (for MinIO, LocalStack)
-    pub s3_endpoint: Option<String>,
-
-    /// Use path-style S3 URLs
-    pub s3_path_style: bool,
+    /// Storage / IO configuration (vended credentials + S3 region/endpoint/path-style).
+    pub io: fluree_db_iceberg::config::IoConfig,
 }
 
 /// How the Iceberg catalog is accessed.
@@ -428,63 +440,48 @@ pub enum CatalogMode {
 }
 
 /// REST catalog mode configuration.
+///
+/// This carries only the catalog-connection fields (`catalog_uri` / `warehouse`
+/// / `auth`); the table identifier lives on [`IcebergCreateConfig`] and
+/// vended-credential / S3 IO settings live on
+/// [`IcebergConnectionConfig::io`].
 #[cfg(feature = "iceberg")]
 #[derive(Debug, Clone)]
 pub struct RestCatalogMode {
     /// REST catalog URI
     pub catalog_uri: String,
-    /// Table identifier (e.g., "openflights.airlines")
-    pub table_identifier: String,
     /// Optional warehouse identifier
     pub warehouse: Option<String>,
     /// Authentication configuration
     pub auth: fluree_db_iceberg::auth::AuthConfig,
-    /// Whether to use vended credentials (default: true)
-    pub vended_credentials: bool,
 }
 
 #[cfg(feature = "iceberg")]
-impl IcebergCreateConfig {
-    /// Create a new Iceberg graph source config for REST catalog mode.
-    pub fn new(
-        name: impl Into<String>,
-        catalog_uri: impl Into<String>,
-        table_identifier: impl Into<String>,
-    ) -> Self {
+impl IcebergConnectionConfig {
+    /// Create a REST-catalog connection with default IO (vended credentials on).
+    pub fn rest(catalog_uri: impl Into<String>) -> Self {
         Self {
-            name: name.into(),
-            branch: None,
             catalog_mode: CatalogMode::Rest(Box::new(RestCatalogMode {
                 catalog_uri: catalog_uri.into(),
-                table_identifier: table_identifier.into(),
                 warehouse: None,
                 auth: fluree_db_iceberg::auth::AuthConfig::None,
-                vended_credentials: true,
             })),
-            s3_region: None,
-            s3_endpoint: None,
-            s3_path_style: false,
+            io: fluree_db_iceberg::config::IoConfig::default(),
         }
     }
 
-    /// Create a new Iceberg graph source config for direct S3 access (no REST catalog).
-    pub fn new_direct(name: impl Into<String>, table_location: impl Into<String>) -> Self {
+    /// Create a Direct S3 connection (no REST catalog). Vended credentials are
+    /// forced off — Direct mode uses ambient/IAM credentials.
+    pub fn direct(table_location: impl Into<String>) -> Self {
         Self {
-            name: name.into(),
-            branch: None,
             catalog_mode: CatalogMode::Direct {
                 table_location: table_location.into(),
             },
-            s3_region: None,
-            s3_endpoint: None,
-            s3_path_style: false,
+            io: fluree_db_iceberg::config::IoConfig {
+                vended_credentials: false,
+                ..Default::default()
+            },
         }
-    }
-
-    /// Set the branch name.
-    pub fn with_branch(mut self, branch: impl Into<String>) -> Self {
-        self.branch = Some(branch.into());
-        self
     }
 
     /// Set bearer token authentication (REST mode only).
@@ -589,8 +586,8 @@ impl IcebergCreateConfig {
 
     /// Enable or disable vended credentials (REST mode only).
     pub fn with_vended_credentials(mut self, enabled: bool) -> Self {
-        if let CatalogMode::Rest(ref mut rest) = self.catalog_mode {
-            rest.vended_credentials = enabled;
+        if self.is_rest() {
+            self.io.vended_credentials = enabled;
         } else {
             tracing::warn!("with_vended_credentials has no effect in Direct catalog mode");
         }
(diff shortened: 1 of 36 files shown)


fluree-db-r2rml/src/emit/render.rs
```diff
@@ -0,0 +1,108 @@
...
+    format!("{}\n{}\n", header, blocks.join("\n\n"))
+}
+
+/// Render a single TriplesMap block.
+fn render_table(tm: &TableMapping, base: &str, vocab_prefix: &str) -> String {
+    let node = tm_node_for_table(&tm.table_name);
+
+    let mut stmts: Vec<String> = Vec::with_capacity(tm.columns.len() + 2);
+    stmts.push(format!(
+        "rr:logicalTable [ rr:tableName \"{}\" ]",
```
bplatz, line 51: Turtle literals here (and `rr:template` :63, `rr:child`/`rr:parent` :80, `rr:column` :89) are emitted with no escaping. A catalog identifier containing `"`, `\`, or a newline produces invalid Turtle (round-trip parse fails on a legal schema); a crafted name can also inject extra predicate-object maps that still parse, silently changing the mapping. Escape `\ " \n \r \t`/control chars on every literal surface, and validate/escape IRIs + PN_LOCAL in `curie()`.

fluree-db-server/src/routes/iceberg.rs
```diff
@@ -252,6 +252,450 @@ fn build_iceberg_config(req: &IcebergMapRequest) -> Result<fluree_db_api::Iceber
...
+    /// S3 region override
+    pub s3_region: Option<String>,
+    /// S3 endpoint override
+    pub s3_endpoint: Option<String>,
+    /// Use path-style S3 URLs
+    #[serde(default)]
+    pub s3_path_style: bool,
+}
+
+fn build_iceberg_connection(
```
bplatz, line 307: `catalog_uri` / `oauth2_token_url` / `s3_endpoint` flow from the request into an outbound HTTP client with no scheme/host allowlist or private-IP block. These routes are pass-through-unauthenticated when `admin_auth.mode == None` (the default), so this is an unauthenticated SSRF vector (e.g. `http://169.254.169.254/...`). Block loopback/link-local/private ranges here regardless of auth mode.

fluree-db-r2rml/src/emit/heuristic.rs
```diff
@@ -0,0 +1,523 @@
...
+                 metadata-only (NDV deferred)",
+                col.name
+            ),
+        ));
+        return SubjectKey {
+            columns: vec![col.name.clone()],
+        };
+    }
+
+    if !table.identifier_field_ids.is_empty() {
```
bplatz, line 302: Unlike the override path and the `<STEM>_KEY/_ID` fallback, this `identifier_field_ids` branch returns the subject key with no `is_non_null()` gate and no diagnostic — contradicting the stated non-null property. A nullable identifier field yields an unsafe subject template (silent row loss at query time) and, in the single-column case, is indexed as a valid FK parent (:180). Apply the same non-null gate or verify against stats.

fluree-db-r2rml/src/emit/heuristic.rs
```diff
@@ -0,0 +1,523 @@
...
+/// Derive the join predicate IRI for a resolved FK on `child_column`.
+///
+/// Uses `camelCase(strip_key_suffix(child))` (readable: `geography`,
+/// `destGeography`), appending `Ref` only when that local would collide with an
+/// existing literal predicate in the same table (e.g. `orderDate` literal vs.
+/// `ORDER_DATE_KEY` join → `orderDateRef`). Predicate IRIs are not compared by
+/// any structural test; this only keeps the emitted document unambiguous.
+fn join_predicate(child_column: &str, draft: &TableDraft, opts: &EmitOptions) -> String {
+    let mut local = naming::camel_case(naming::strip_key_suffix(child_column));
+    if draft.literal_locals.contains(&local) {
```
bplatz: `Ref` is appended only on collision with literal locals, not with other joins. Two FKs that strip+camel to the same local (e.g. `GEOGRAPHY_KEY` and `GEOGRAPHY_ID`) emit the same predicate IRI for different parents, merging two relationships. Track emitted join locals too.

fluree-db-iceberg/src/stats.rs
```diff
@@ -0,0 +1,749 @@
...
+        row_count += df.record_count;
+        total_bytes += df.file_size_in_bytes;
+
+        for field in &scalar_fields {
+            let fid = field.id;
+            let Some(acc) = accs.get_mut(&fid) else {
+                continue;
+            };
+
+            if let Some(n) = lookup(df.null_value_counts.as_ref(), fid) {
```
bplatz, line 142: `null_count` sums only over files that report `null_value_counts`, with no all-files-reported tracking. A column with nulls in a non-reporting file can aggregate to `Some(0)` → `null_fraction == 0` → treated as a safe non-null subject key. Treat partial coverage as unknown, not zero.

fluree-db-iceberg/src/stats.rs
```diff
@@ -0,0 +1,749 @@
...
+                add_opt(&mut acc.nan_count, n);
+            }
+            if let Some(n) = lookup(df.column_sizes.as_ref(), fid) {
+                add_opt(&mut acc.on_disk_bytes, n);
+            }
+
+            if let Some(bytes) = df.lower_bound(fid) {
+                if let Ok(v) = decode_bound(bytes, field) {
+                    had_column_bounds = true;
+                    acc.min = Some(keep_min(acc.min.take(), v));
```
bplatz, line 160: `keep_min`/`keep_max` rely on `TypedValue::partial_cmp`, which has no `Decimal`/`Uuid` arms (→ `None` → keeps current). A multi-file decimal/uuid column reports the first file's bound, not the true min/max. (FK inference is integer-only, so impact is limited to preview stats.) Add the missing arms.

fluree-db-iceberg/src/stats.rs
```diff
@@ -0,0 +1,749 @@
...
+        .iter()
+        .map(|f| (f.id, Acc::default()))
+        .collect();
+
+    let mut row_count = 0i64;
+    let mut total_bytes = 0i64;
+    let mut had_column_bounds = false;
+
+    for df in data_files {
+        row_count += df.record_count;
```
bplatz, line 133: `row_count` (and value/null counts) sum live data files but don't subtract position/equality delete files, so they overcount on merge-on-read tables while being presented as authoritative.

fluree-db-iceberg/src/stats.rs
```diff
@@ -0,0 +1,749 @@
...
+            if let Some(bytes) = df.lower_bound(fid) {
+                if let Ok(v) = decode_bound(bytes, field) {
+                    had_column_bounds = true;
+                    acc.min = Some(keep_min(acc.min.take(), v));
+                }
+            }
+            if let Some(bytes) = df.upper_bound(fid) {
+                if let Ok(v) = decode_bound(bytes, field) {
+                    had_column_bounds = true;
+                    acc.max = Some(keep_max(acc.max.take(), v));
```
bplatz, line 166: Iceberg truncates string/binary min/max bounds; the decoded prefix is recorded as an exact `min`/`max` with no truncation flag, so the preview presents truncated bounds as exact.

fluree-db-api/src/graph_source/config.rs
```diff
@@ -400,17 +400,29 @@ pub struct IcebergCreateConfig {
...
+/// The reusable Iceberg connection block — catalog access + IO, with **no**
+/// table or mapping.
+///
+/// Factored out of [`IcebergCreateConfig`] so catalog browse / metadata preview
+/// can run against an **unsaved** connection during onboarding (before a graph
+/// source record is created). The relationship is:
+/// `IcebergCreateConfig` = `IcebergConnectionConfig` + `table_identifier`.
+#[cfg(feature = "iceberg")]
+#[derive(Debug, Clone)]
+pub struct IcebergConnectionConfig {
```
bplatz, line 420: `IcebergConnectionConfig` derives `Debug` and transitively holds plaintext bearer/OAuth secrets. Nothing logs it today, but a future `?conn` / `{:?}` in a log or error would leak credentials. Add a redacting `Debug`.

bplatz (approved):
Several comments!

aaj3f:
> Several comments!

Thank you @bplatz -- I didn't want to bug you but I kind of wanted/needed an approval before tomorrow morning

fluree-db-r2rml/src/emit/render.rs
aaj3f: Addressed in fc9e3c42f

fluree-db-server/src/routes/iceberg.rs
aaj3f: Addressed in e1b3c52af → 680704906 → f95dd026c

fluree-db-r2rml/src/emit/heuristic.rs
aaj3f: Addressed in fc9e3c42f + 95be45b68

fluree-db-r2rml/src/emit/heuristic.rs
aaj3f: Addressed in fc9e3c42f

fluree-db-iceberg/src/stats.rs
aaj3f: Addressed in 1c57a59d7

fluree-db-iceberg/src/stats.rs
aaj3f: Addressed in 1c57a59d7

fluree-db-iceberg/src/stats.rs
aaj3f: Addressed in 1c57a59d7

fluree-db-iceberg/src/stats.rs
aaj3f: Addressed in 1c57a59d7

fluree-db-api/src/graph_source/config.rs
aaj3f: Addressed in e2a0e05f9 + f95dd026c

PR #1409, opened 2026-07-01
Fix `select *` Stack Overflow on the `query_from` / `build_client` Path

### LINKED ISSUE #1408: `SELECT *` crawl overflows the default ~2MB stack
**Severity:** blocks fluree/solo bumping its `fluree-db-*` pin to current db main; **potential production overflow** on deep `SELECT *` crawls.
**Found:** bumping fluree/solo's fluree-db pin from `171c8e94be61e1c10ecc154073511f1d2f6b5d2e` → `6daf5e3dce29143dd423c728c75c0871e5ecb54e` (the #1407 `feat/iceberg-r2rml-engine` head, which is based on current db `main`; that PR's own changes are additive Iceberg and do not touch this path).

## Symptom
The solo router integration test `select_star_crawl::file_build_client_select_star_via_query_from` **SIGABRTs (stack overflow)** at the default ~2MB thread stack on the new pin. It **passes with `RUST_MIN_STACK=536870912` (512 MB)**. The same test passes at the *default* stack on the *old* pin — it is part of solo `main`'s green CI, which runs `cargo test --workspace --all-features` with **no `RUST_MIN_STACK`** set. Since the only change between the two is the fluree-db rev, the extra stack depth was introduced somewhere in the db range `171c8e94..6daf5e3`.

## Where
The overflow is in the `SELECT *` crawl on the **type-erased `query_from()` nameservice path** — `build_client()` yielding `AnyStorage` / `AnyNameService`. The recursion is **deep-but-finite** (512 MB clears it): a per-crawl-level stack-frame increase, not a cycle.

## Hypothesis (a bisect would confirm the exact commit)
The db range includes a **nameservice read/write trait split** (separate read vs. write trait variants) plus graph-crawl changes. Adding an indirection or a large future/local per crawl level would raise peak stack roughly proportionally to crawl depth. A `git bisect` over `171c8e94..6daf5e3` against the repro below (or a db-side `SELECT *` crawl on a deep, type-erased client at the default stack) should pinpoint it.

## Impact
- **Blocks fluree/solo#729 CI** — solo CI runs `cargo test --workspace --all-features` at the default stack, so this turns red on the new pin.
- **Potential production overflow** — AWS Lambda tokio worker threads also default to ~2MB stacks, so a real `SELECT *` crawl on the type-erased path at sufficient depth could SIGABRT in prod, not only in tests.

## Repro (from the solo repo, once pinned to current db main)
```bash
gh pr checkout 729
cd fluree/solo/src/rust/lambda
# fails (SIGABRT) at default stack:
cargo test -p fluree-lambda-router --test integration \
  select_star_crawl::file_build_client_select_star_via_query_from
# passes with a large stack:
RUST_MIN_STACK=536870912 cargo test -p fluree-lambda-router --test integration \
  select_star_crawl::file_build_client_select_star_via_query_from
```
(A pure-db repro would be a `SELECT *` over a suitably deep/nested graph through the type-erased `query_from()` / `build_client()` path at the default stack.)

## Options
1. **Preferred (db-side): make peak stack depth-independent.** Convert the recursive crawl to an explicit work-stack (iterative), and/or `Box` the large per-level locals/futures so each crawl level costs O(1) stack.
2. **Stopgap: raise the query runtime stack.** Set the tokio worker `thread_stack_size` (and `RUST_MIN_STACK` in CI). Unblocks CI but only masks the production risk — not a final fix.

---
*Context: surfaced during fluree/solo#729 (Iceberg → R2RML → Dataset backend). It does not block that PR's feature development (that work doesn't touch the crawl path), but it does gate the PR's CI/merge until the pin points at a db rev where this is fixed.*

aaj3f:
Closes #1408

## Summary

A plain `{"select": {iri: ["*"]}}` connection query, run through the type-erased `build_client()` / `query_from()` path, overflowed the default ~2 MB worker-thread stack **in debug builds** and aborted the process (`SIGABRT`). That broke solo CI (`cargo test --workspace --all-features`, which does not set `RUST_MIN_STACK`). This PR boxes a handful of futures along the query path so the stack stays bounded. Debug peak for the repro drops from **>2 MB to ~1.4 MB**; release builds are unchanged (they already peaked well under 0.5 MB). No change to query behavior or results.

## What was actually happening (the issue's hypothesis was wrong)

The issue guessed a *depth-recursive* `SELECT *` graph crawl and suggested rewriting the crawl iteratively. That is **not** the cause. The failing query has no `depth` parameter, so `depth` defaults to `0` and the hydration crawl provably does **not** recurse through references. I confirmed this under `lldb`: at the moment of overflow the stack is only ~57 frames deep with **no repeating cycle** — it is not runaway recursion.

The real cause is that query execution is a **deep chain of ~16 `async fn`s** — query dispatch → ledger load → novelty rebuild → query execution → hydration — and in **debug builds** each of those async state machines is large. Measuring the individual stack frames (via `lldb`, at the deepest point) showed the query dispatch/load chain alone consuming ~1.5 MB, dominated by:

| frame | debug frame size (before) |
|---|---|
| `FromQueryBuilder::execute_tracked` | **~470 KB** |
| `query_connection_jsonld_tracked_with_options` | ~166 KB |
| `prepare_single_view_for_connection_tracked` | ~110 KB |
| `load_view_from_source` | ~104 KB |
| ~12 more frames (view load, ledger load, novelty rebuild) | ~25–60 KB each |

`execute_tracked` is just a dispatcher, yet it reserved ~470 KB. That is the tell.

## Why an `async fn` can have a 470 KB stack frame (background for reviewers)

An `async fn` does not run like a normal function. The compiler turns it into a **state-machine struct** that has to store *every local variable that is still alive across an `.await`*. Crucially, when function `A` does `B().await` and `B` is itself an `async fn`, `A`'s state machine embeds **all of `B`'s state inline** (and `B` embeds `C`'s, and so on). Down a 16-level-deep call chain, the top-level future becomes enormous because it transitively contains every level below it.

When that future is *polled*, the poll executes on the stack, and the stack frame is as large as the state it has to touch. Release builds pack and reuse these slots aggressively, so the frames stay small. **Debug builds do almost none of that packing**, so the frames balloon — which is exactly why this bites CI (debug) but not production (release).

`execute_tracked` is a `match` that dispatches to one of ~8 `query_connection_*` variant futures. Even though only one arm ever runs, the compiler statically reserves stack for constructing the *largest* arm's future — hence ~470 KB.

## The fix: box the fat futures at chokepoints

`Box::pin(some_future)` moves a future's state to the **heap** and leaves behind an 8-byte pointer. So if `A` does `Box::pin(B()).await` instead of `B().await`, `A`'s state machine stores only a pointer to `B` (on the heap) rather than all of `B` inline. `A`'s frame shrinks accordingly. Applied at the right chokepoints, the whole chain's stack footprint drops from "sum of everything inline" to "a few small frames plus some heap".

Concretely:

- **`FromQueryBuilder::execute_tracked`** — the 8-way dispatch is moved into a small, `#[inline(never)]` helper (`dispatch_connection_query`) that *returns an already-boxed future*. Boxing at the call site alone would not help, because the compiler would still materialize the chosen arm's future on `execute_tracked`'s frame before boxing it. By constructing-and-boxing inside a separate non-inlined function, that transient lives in a short-lived frame that **returns (and is popped) before** we await the boxed future and descend into the deep work. Result: **470 KB → 26 KB**.
- **`query_connection_jsonld_tracked_with_options`**, **`prepare_single_view_for_connection_tracked`**, **`load_view_from_source`**, **`query_tracked_with_options`** — box the load-path and execute-path awaits (view prep, ledger load, plan build, execution, hydration formatting). Each `Box::pin` shrinks its caller's frame and cascades: a boxed callee future is smaller, so its caller's transient is smaller too.

Files touched: `fluree-db-api/src/query/builder.rs`, `fluree-db-api/src/query/connection.rs`, `fluree-db-api/src/view/dataset_builder.rs`, `fluree-db-api/src/view/query.rs`.

## Performance impact — full candor (speed **and** memory, release **and** debug)

I want to be explicit here because "box futures" can sound like it trades runtime cost for a stack win. I measured it; it does not rob release performance in any meaningful way.

**Release stack was never at risk.** I built the repro in `--release` and swept the thread stack size:

| build | peak stack for the repro query |
|---|---|
| debug, before fix | **> 2 MB** (SIGABRT) |
| debug, after fix | ~1.4 MB (passes 2 MB with ~600 KB headroom) |
| release, before fix | **< 0.5 MB** (already fit 2 MB with ~1.5 MB headroom) |
| release, after fix | ~0.45 MB (essentially unchanged) |

So the fix rescues **debug/CI only**. In release the frames were already small; the fix neither needs to nor does meaningfully change the release stack. (This also means the issue's "potential production overflow" framing does not apply to this query: it is a fixed-depth query, not a depth-scaling one, and release fits comfortably.)

**Speed (release):** the change adds roughly **6–10 `Box::pin` heap allocations per query**, all on the one-time query-setup / ledger-load path. They are **not** in any per-row, per-flake, or per-result hot loop, and they **do not scale** with data size, result size, or crawl depth. There is also a pointer indirection (one vtable hop) when polling each boxed future — again a handful of polls, not a tight loop. Next to what a real query already does (storage I/O, index scans, novelty merge, parsing, hydration — thousands of allocations), this is unmeasurable. There is **no algorithmic change**; query planning, execution, and results are byte-for-byte identical.

**Memory (release):** boxing relocates a few futures' state from the stack to the heap. This is **transient** (freed the moment the query future completes), on the order of a few KB per in-flight query, and does **not** scale with data. Net process memory is ~neutral — the same state, just on the heap instead of the stack, plus a small allocator-overhead constant. Stack usage strictly goes **down**.

**The honest "peter/paul" answer:** nothing meaningful is being robbed from release. The cost is a small, fixed, per-query allocation/indirection overhead on a non-hot path; the benefit is that the query can no longer SIGABRT a 2 MB worker. If someone micro-benchmarked a tight loop of trivial queries they might see a sub-percent difference from the extra allocations; for any real workload it is noise. Boxing recursive/deep async futures is the standard, idiomatic Rust remedy for exactly this class of problem.

## Testing

- New regression guard `fluree-db-api/tests/it_select_star_deep_crawl_stack.rs`: builds a type-erased `build_client()` connection, inserts (unindexed) data so `query_from` rebuilds novelty on load, and runs `select *` on a thread with a fixed 2 MB stack. It is **red before this change** (SIGABRT) and **green after** — I verified both directions by stashing the fix. It uses a self-contained synthetic ontology (no external fixtures) sized so the pre-fix code overflows at 2 MB with margin.
- Full `fluree-db-api --features native` suite: **~2,700 tests, 0 failures**.
- `--all-features` build clean; `clippy` clean on the changed crate.

## Out of scope (noted for completeness)

There is a *separate*, genuinely depth-driven consideration: a `SELECT *` with a large explicit `depth:` on a deep graph will recurse `format_subject` in the hydration formatter. That path is already boxed per level (`BoxFuture`), so it is O(depth) with a small constant, and it is **not** what #1408 hit (depth 0). Hardening it further is a distinct follow-up and is not addressed here.

### DIFF
--- a/fluree-db-api/Cargo.toml
+++ b/fluree-db-api/Cargo.toml
@@ -227,6 +227,10 @@ name = "it_vector_flatrank"
 path = "tests/it_vector_flatrank.rs"
 required-features = ["vector"]
 
+[[test]]
+name = "it_select_star_deep_crawl_stack"
+path = "tests/it_select_star_deep_crawl_stack.rs"
+
 [[bench]]
 name = "vector_query"
 harness = false
--- a/fluree-db-api/src/query/builder.rs
+++ b/fluree-db-api/src/query/builder.rs
@@ -1332,21 +1332,83 @@ impl<'a> FromQueryBuilder<'a> {
         // `connection_opts`. Resolves & applies policy from the merged opts.
         if let (Some(qc_opts), QueryInput::Sparql(sparql)) = (self.connection_opts.as_ref(), input)
         {
-            return self
-                .fluree
-                .query_connection_sparql_tracked_with_opts_options(
-                    sparql,
-                    qc_opts,
-                    format_config,
-                    tracking,
-                    execution.clone(),
-                )
-                .await;
+            // Box the dispatched future: this async fn `.await`s one of ~9
+            // large `query_connection_*` futures, and (especially in debug
+            // builds) the compiler stores each awaited sub-future inline in
+            // this fn's state machine rather than overlapping the mutually
+            // exclusive arms. That inflated `execute_tracked`'s stack frame to
+            // ~470 KB, which — stacked with the deep ledger-load / novelty-
+            // rebuild chain — overflowed the default ~2 MB worker stack on a
+            // plain `select *` connection query (fluree/db#1408). Boxing moves
+            // the sub-future's state to the heap so the frame holds only a
+            // pointer.
+            return Box::pin(
+                self.fluree
+                    .query_connection_sparql_tracked_with_opts_options(
+                        sparql,
+                        qc_opts,
+                        format_config,
+                        tracking,
+                        execution.clone(),
+                    ),
+            )
+            .await;
         }
-        match input {
+        // Dispatch to the chosen `query_connection_*` variant via a separate,
+        // non-inlined fn that RETURNS the boxed future. The large per-arm future
+        // is materialized transiently inside that helper's frame during
+        // `Box::pin`; the helper returns (and its frame is popped) before we
+        // await here, so that transient never coexists on the stack with the
+        // deep query / load / novelty-rebuild chain the future then drives.
+        // Constructed inline here instead, the dispatcher frame would reserve
+        // stack for the largest arm's future for the whole descent — one of the
+        // frames that overflowed the default ~2 MB worker stack on a plain
+        // `select *` connection query (fluree/db#1408).
+        self.dispatch_connection_query(input, format_config, tracking, execution, r2rml.as_ref())
+            .await
+    }
+
+    /// Build the boxed connection-query future for the chosen input / policy /
+    /// r2rml combination (helper for [`execute_tracked`](Self::execute_tracked)).
+    ///
+    /// Returns an already-heap-allocated future so the caller's frame holds only
+    /// a pointer while driving it. `#[inline(never)]` is load-bearing: it keeps
+    /// the transient materialization of the selected (large) arm future in this
+    /// short-lived frame instead of the caller's long-lived one — see the call
+    /// site and fluree/db#1408.
+    #[inline(never)]
+    #[allow(clippy::type_complexity)]
+    fn dispatch_connection_query<'s>(
+        &'s self,
+        input: QueryInput<'s>,
+        format_config: Option<FormatterConfig>,
+        tracking: Option<TrackingOptions>,
+        execution: QueryExecutionOptions,
+        r2rml: Option<&'s (
+            Arc<dyn R2rmlProvider + 'a>,
+            Arc<dyn R2rmlTableProvider + 'a>,
+        )>,
+    ) -> std::pin::Pin<
+        Box<
+            dyn std::future::Future<
+                    Output = std::result::Result<TrackedQueryResponse, TrackedErrorResponse>,
+                > + Send
+                + 's,
+        >,
+    > {
+        // `let`-bind with the trait-object type so each arm's distinct
+        // `impl Future` unsize-coerces to the shared `dyn` future type.
+        let dispatched: std::pin::Pin<
+            Box<
+                dyn std::future::Future<
+                        Output = std::result::Result<TrackedQueryResponse, TrackedErrorResponse>,
+                    > + Send
+                    + 's,
+            >,
+        > = match input {
             QueryInput::JsonLd(json) => match &self.policy {
-                Some(policy) => match r2rml.as_ref() {
-                    Some((provider, table_provider)) => {
+                Some(policy) => match r2rml {
+                    Some((provider, table_provider)) => Box::pin(
                         self.fluree
                             .query_connection_jsonld_tracked_with_policy_and_r2rml_options(
                                 json,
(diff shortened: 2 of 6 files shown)


bplatz (approved):
▶️

PR #1410, opened 2026-07-01
fix(query): honor @type in JSON-LD select-crawl projections

bplatz:
## Problem

A JSON-LD select-crawl selection spec such as `{"select": {"m:California": ["@id", "@type"]}}` **silently dropped** the subject's types, returning only `{"@id": "m:California"}`. Meanwhile `rdf:type` and `*` both returned the types correctly:

| Query | Result |
|-------|--------|
| `["@id", "@type"]` | `{"@id": "m:California"}` — types dropped ❌ |
| `["@id", "rdf:type"]` | `@type: [skos:Concept, m:State]` ✅ |
| `["*"]` | `@type: [Concept, State]` ✅ |
| WHERE `@type: ?type` | binds Concept, State ✅ |

So `@type` was honored in WHERE but silently dropped in a select-crawl projection.

## Root cause

The bare `@type` key in a selection array encodes to the JSON-LD **sentinel** Sid (namespace `JSON_LD`, name `"type"`), not the real rdf:type Sid. The hydration formatter fetched the subject's properties filtered by that sentinel predicate, so the subject's actual rdf:type flakes never matched and were never emitted. `rdf:type` worked because it encodes to the real Sid; `*` worked because it fetches every predicate and already special-cases rdf:type → `@type`.

The sentinel is genuinely dual-purpose: on a **literal** value (`{"schema:name": ["@type"]}`) it means "emit the literal's datatype", which is resolved only at format time.

## Fix

- **Parser** (`parse/mod.rs`): lower `@type` / `type` / the context type key — both the bare-string form and the nested `{"@type": [...]}` object form — to a `ForwardItem::Property` on the rdf:type predicate, so a node crawl flows through the existing rdf:type rendering identically to `rdf:type` and `*`.
- **Formatter** (`format/hydration.rs`): `format_literal_virtual` now also treats an rdf:type forward item as a datatype request, preserving `{"schema:name": ["@type"]}` → `{"@type": "xsd:string"}`. `@value` / `@language` keep the sentinel.
- **Docs**: note `@type` (≡ `rdf:type`) as a valid selection key in `docs/query/jsonld-query.md`.

## Tests

- `jsonld_basic_single_subject_query_explicit_type` — node case regression.
- `jsonld_basic_type_matches_rdf_type_in_crawl` — asserts `@type` ≡ `rdf:type` at the top level and at a nested-ref level, and that the nested ref carries its type.
- `grp_query` (300) and `grp_query_sparql` (258) pass; `fluree-db-query` unit/doctests pass; fmt + clippy (`-D warnings`) clean on both crates.

### DIFF
--- a/fluree-db-api/src/format/hydration.rs
+++ b/fluree-db-api/src/format/hydration.rs
@@ -2000,7 +2000,9 @@ impl<'a> HydrationFormatter<'a> {
                                 "language" => want_language = true,
                                 _ => {}
                             }
-                        } else if self.compactor.decode_sid(predicate)? == RDF_TYPE_IRI {
+                        } else if self.compactor.decode_sid(predicate).ok().as_deref()
+                            == Some(RDF_TYPE_IRI)
+                        {
                             // `@type` lowers to the rdf:type predicate; on a
                             // literal value that means "emit the datatype".
                             want_type = true;
--- a/fluree-db-api/src/format/hydration.rs
+++ b/fluree-db-api/src/format/hydration.rs
@@ -2000,6 +2000,10 @@ impl<'a> HydrationFormatter<'a> {
                                 "language" => want_language = true,
                                 _ => {}
                             }
+                        } else if self.compactor.decode_sid(predicate)? == RDF_TYPE_IRI {
+                            // `@type` lowers to the rdf:type predicate; on a
+                            // literal value that means "emit the datatype".
+                            want_type = true;
                         }
                     }
                 }
--- a/docs/query/jsonld-query.md
+++ b/docs/query/jsonld-query.md
@@ -83,7 +83,7 @@ Specifies what to return in results. The shape of `select` determines the shape 
 }
 ```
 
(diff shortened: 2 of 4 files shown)


fluree-db-api/src/format/hydration.rs
```diff
@@ -2000,6 +2000,10 @@ impl<'a> HydrationFormatter<'a> {
                                 "language" => want_language = true,
                                 _ => {}
                             }
+                        } else if self.compactor.decode_sid(predicate)? == RDF_TYPE_IRI {
```
aaj3f: Given the `?` here, the evaluation of these `if` / `else` branches could return an `UnknownNamespace` error, which would abort the whole format, technically. In practice I think the violation would be unlikely. If we instead wanted the prior code's behavior of "silently skip unknowns" for a non-error-capable compare we could do `self.compactor.decode_sid(predicate).ok() == Some(RDF_TYPE_IRI)`

aaj3f (approved):
👍

fluree-db-api/src/format/hydration.rs
bplatz: addressed in 7bffcb8c395c68f57c24ef2fea4ce8ae87929cbc

PR #1411, opened 2026-07-01
perf(iceberg/r2rml): Arrow reader, LIMIT/predicate pushdown, and bound-object query support

bplatz:
## Summary

Iceberg/R2RML graph-source performance and query-coverage work, validated live against Snowflake Horizon. Turns the reader into an Arrow columnar pipeline with native pushdown, makes a `FILTER + LIMIT` fact query early-terminate, and closes a set of common query patterns that previously errored on graph sources.

**Stacked on #1406** — base is `feature/r2rml-query-scoped-catalog-session`, so this diff is only the work on top of that PR. Merge #1406 first (or retarget to `main` once it lands).

## What's here

**Arrow reader (single decode path)**
- Replace the row-based `RowIter` decode with `ParquetRecordBatchReaderBuilder`: native projection, row-group pruning (`with_row_groups`), and exact row filtering. Byte-identical output (funnelled through the same `ColumnValue` assembly).
- Row filtering is applied post-decode via `filter_record_batch`, **not** `with_row_filter`: the latter's `skip_records` panics on Snowflake's `DELTA_BINARY_PACKED` integer columns in parquet-rs 54.
- Arrow is now the only Iceberg reader (dropped the feature gate + the untested `iceberg-lean`/RowIter fallback); `arrow` is pulled with `default-features = false`.

**Pushdown**
- Row-group pruning + exact row filtering for date/int/bool **and string** comparisons (Parquet string min/max are valid bounds even when truncated).
- **LIMIT flows to the scan**: when a `FILTER`'s conjuncts are fully consumed by a single scan-local pattern, the planner drops the standalone `FilterOperator` and the operator applies the filter itself, so a `FILTER + LIMIT` stops after enough matching rows.

**Query patterns that used to error on graph sources — now supported**
- Constant objects: plain string (`?s :channel "Online"`), integer/boolean (`?s :year 2024`), and bound ref/IRI (`?s :geography <geo/1>`). Operator-enforced (correct with pushdown off), datatype-aware (untyped/XSD loose-match, language-tagged/custom stay strict), integer comparison is exact `i64` (no `f64` false positives, stays consistent with the scan filter).

## Live results (Snowflake Horizon; byte-identical to the in-engine path)

| Query | Before | After |
|---|---|---|
| `Store ⋈ Geography` join | ~229 s / 36 scans / 1.6 GB RSS | ~9 s / 3 scans / 129 MB (warm server **~1 s**) |
| `FACT_ORDER_LINE` `FILTER + LIMIT 10` | >2 min (timeout) | ~7 s early-terminate |
| `FILTER(?channel = "Online")` string pushdown | n/a | 908 rows, byte-identical on/off |
| `?s :geography <geo/1>` (previously errored) | error | 16 rows, byte-identical to FILTER form |

Warm-server, whole benchmark: the join workload went from **>2 min to ~1 s**.

## Config (all default-on, kill-switchable)

`FLUREE_ICEBERG_PREDICATE_PUSHDOWN`, `FLUREE_R2RML_LIMIT_PUSHDOWN`, `FLUREE_R2RML_FILTER_CONSUMPTION`, `FLUREE_ICEBERG_LOADTABLE_CACHE`, `FLUREE_R2RML_SCAN_CACHE`, `FLUREE_R2RML_MATERIALIZE_WINDOW_ROWS` — documented in `docs/operations/configuration.md`.

## Testing

- `cargo fmt --all -- --check`, `cargo clippy --all --all-features --all-targets -- -D warnings`: clean.
- Unit + guardrail tests: Arrow decode byte-identical round-trip; row-group pruning (int + string) against real 2-row-group Parquet fixtures; exact row filtering incl. the Decimal-backed-integer case; LIMIT early-termination pull-count; scan-local FILTER consumption + leak cases; constant-object matching (string/int/bool/IRI, large-integer boundary, datatype gate).
- Live A/B on the `enterprise` graph source for every change, byte-identical pushdown on vs off.

## Not included (follow-ons)

- Bound **subject** (`<store/5> :name ?n`) — structural (needs an optional subject in the pattern).
- Float/decimal/date literal constant objects; fusing a constant-object into a same-subject star scan; FK-key pushdown for IRI objects (needs subject-template reversal).


fluree-db-iceberg/src/io/arrow_reader.rs
```diff
@@ -0,0 +1,383 @@
...
+    let batch_schema = Arc::new(batch_schema);
+
+    // Real (non-null-sentinel) parquet leaf indices this projection needs.
+    let real: Vec<usize> = column_indices
+        .iter()
+        .copied()
+        .filter(|&idx| idx != NULL_COLUMN_SENTINEL)
+        .collect();
+
+    let mask = ProjectionMask::leaves(builder.parquet_schema(), real.iter().copied());
```
aaj3f: `column_indices` seem to come from `parquet_schema.get_fields()` and are **root** field indices. But their use in `ProjectionMask::leaves(...)` and then below in `leaf_to_pos`/`field_to_pos` treat them as **leaf** column indices. For a flat schema root idx will `==` leaf idx, but if/when the file contains a nested column (struct/list/map/VARIANT) before a projected primitive, those index spaces will diverge and the mask will select the wrong leaf. 

Previously, the RowIter path used `build_projected_schema()` which indexed w/ consistent root-based selection handling nested schemas correctly. 

We should be able to use `SchemaDescriptor`'s leaf-root map before building the mask and `leaf_to_pos` or use `ProjectionMask::roots(...)` w/ root indices and establish the position map on root order.

fluree-db-query/src/r2rml/operator.rs
```diff
@@ -297,6 +329,30 @@ impl R2rmlScanOperator {
...
+                    if let [col] = pom.object_map.referenced_columns().as_slice() {
+                        out.push(crate::r2rml::ScanFilter {
+                            column: (*col).to_string(),
+                            op: crate::r2rml::ScanCmpOp::Eq,
+                            value: value.clone(),
+                        });
+                    }
+                }
+            }
+        }
```
aaj3f, line 349: Leaving as-is this claude-observed invariant violation:
````
scalar constant-object pushdown drops operator-kept rows for templated/transformed object maps (violates the stated pushdown invariant)

The scalar-constant `Eq` `ScanFilter` (lines 336–351) is pushed whenever the matched
`PredicateObjectMap` is *not* a `RefObjectMap` and references *exactly one* column. An
`ObjectMap::Template { template: "PREFIX-{code}", columns: ["code"] }` passes both guards
(`referenced_columns()` = `["code"]`), but its **materialized** value is `"PREFIX-<code>"`, not
the raw column. The pushdown then emits `Eq(code, "PREFIX-A")`, comparing the raw column against
the fully-templated constant.

**Failure scenario:** mapping `:label` → `rr:template "PREFIX-{code}"`, column `code = "A"`;
query `?s :label "PREFIX-A"`. The operator would materialize `"PREFIX-A"` and keep the row, but
the pushed `Eq(code,"PREFIX-A")` row filter (and potentially row-group pruning) drops every row
before the operator sees it → **zero rows.** This is exactly the invariant the PR claims to
uphold ("the pushed ScanFilter is only an optimization … pushdown never removes an operator-kept
row", `provider.rs` / `operator.rs:1176`).

**Suggested fix:** only push the `Eq` filter when `pom.object_map` is a plain
`ObjectMap::Column { .. }` with no template and no lexical-transforming datatype. (The
pre-existing range-predicate loop at `operator.rs:319–325` shares the same single-column guard
and latent flaw — worth tightening in the same spot.)
````

File: fluree-db-query/src/r2rml/operator.rs [hunk 3]
```diff
@@ -1104,6 +1166,40 @@ fn materialize_pom_object(
...
+                    "true" | "1" => *b,
+                    "false" | "0" => !*b,
+                    _ => false,
+                },
+                // Date constant objects are not produced by convert yet.
+                ScanValue::Date(_) => false,
+            }
+        }
+    }
+}
```
aaj3f, line 1221: This may be hard to resolve in this PR (and possibly I may be wrong about the issue / path handling), but looking at the fn comment and the code, I'm trying to assess what would happen w/ say: 

`:amount` → `rr:column "amount"` where `amount` is `DECIMAL(10,2) = 100.00`
query: `?s :amount 100`

I think we'd see `"2024.00".parse::<i64>()`, which would fail and then the row would be dropped

It seems intentional given the tests at line 1594. But it seems like a common column type, possibly. Resolving it would need column-type awareness (parse as decimal when column is decimal).

fluree-db-query/src/r2rml/operator.rs
```diff
@@ -1104,6 +1166,40 @@ fn materialize_pom_object(
...
+                return false;
+            };
+            match value {
+                ScanValue::Str(s) => v == s,
+                ScanValue::Int(n) => v.parse::<i64>().is_ok_and(|x| x == *n),
+                ScanValue::Bool(b) => match v.as_str() {
+                    "true" | "1" => *b,
+                    "false" | "0" => !*b,
+                    _ => false,
+                },
```
aaj3f, line 1215: Narrower than the above scenario, but for text-typed boolean columns stored as `"t"/"f"`, `"TRUE"`/`"FALSE"` etc, they'd fall into `_ => false`

Is it worth broadening the truthy set and/or making case-insensitive etc?

fluree-db-query/src/r2rml/rewrite.rs
```diff
@@ -234,13 +235,114 @@ pub fn rewrite_patterns_for_r2rml(
...
+    match expr {
+        Expression::Var(v) => {
+            out.insert(*v);
+            true
+        }
+        Expression::Const(_) => true,
+        Expression::Call { args, .. } => args.iter().all(|a| collect_expr_vars(a, out)),
+        _ => false,
+    }
+}
```
aaj3f, line 353: Not sure of how reachable this consideration observed by Claude is, so I'll leave verbatim for you to take or leave:

````
metadata-read FILTERs are consumed and applied via the fail-closed sync path, diverging from `FilterOperator` under a view policy

`collect_expr_vars` returns `true` for any `Expression::Call` tree over `Var`/`Const`:

```rust
Expression::Call { args, .. } => args.iter().all(|a| collect_expr_vars(a, out)),
```

A Cypher metadata read (`type`/`labels`/`keys`/`properties`/`startNode`/`endNode`) is such a
`Call`, so `FILTER(type(?o) = "...")` over a single R2RML scan var gets consumed into
`consumed_filter` and applied by `finalize_batch` via plain synchronous `filter_batch`. The
standalone `FilterOperator` instead routes metadata reads through the policy-aware
`filter_batch_with_exists(needs_metadata=true)` path when `!ctx.allow_unfiltered()`. The sync
readers are documented fail-closed under a non-root policy, so the consumed path can silently
drop rows the correct path would keep — violating the "results unchanged" invariant
(`operator.rs:180`, `rewrite.rs:243`).

**Suggested fix:** in `consume_scan_local_filters`, skip any expression where
 `contains_metadata_read(expr)` is true — mirroring the fail-closed treatment `collect_expr_vars` already 
gives `Member`/`Exists`.
````

File: fluree-db-iceberg/src/io/arrow_reader.rs [hunk 6]
```diff
@@ -0,0 +1,383 @@
...
+}
+
+/// Convert a single Arrow array cell to the intermediate [`ColumnValue`],
+/// producing the intermediate `ColumnValue` for the shared column assembly.
+/// Returns `None` for nulls and unsupported types.
+fn arrow_cell_to_column_value(
+    array: &dyn Array,
+    row: usize,
+    field_type: &FieldType,
+) -> Option<ColumnValue> {
```
aaj3f: In lines 165-171, this fn gets invoked once per `(row, column)` and every invocation does a `.downcast_ref::<T>()` & `match array.data_type()` on every cell. 

It seems we only need to perform a downcast once per every column, and instead then iterate the rows over the concrete typed array. And then if we did that we could simply use `.values()` isntead of `.value(row)`

TLDR: if we hoist `downcast_ref` out of the row loop and into a per-column dispatch and reorganize around that, it may be a pretty significant perf win

fluree-db-iceberg/Cargo.toml
aaj3f, line 1: At least at the moment, the only user of `arrow` is `arrow_reader.rs`, and that is, itself, `#[cfg(feature = "aws")]`. We may want to at least make `arrow` optional here and then have `aws` gate it with `aws = [..., "dep:arrow", "parquet/arrow"]`. Otherwise a non-`aws` build will still compile `arrow` and `parquet-arrow` but leave them unused

fluree-db-iceberg/src/io/arrow_reader.rs
```diff
@@ -0,0 +1,383 @@
...
+    // keeping native type handling (Int→Decimal casts). Row-group pruning
+    // (below) still runs, and the in-engine FILTER stays the authority — a kept
+    // row is re-checked; a dropped row (predicate false, or a null cell that
+    // yields no R2RML triple) would be excluded downstream anyway.
+    let predicate_plan = residual_filter
+        .filter(|_| predicate_pushdown_enabled())
+        .and_then(|r| build_predicate_plan(r, md, iceberg_schema, &leaf_to_pos));
+
+    let surviving =
+        crate::io::send_parquet::surviving_row_groups(md, residual_filter, iceberg_schema);
```
aaj3f: `build_predicate_plan()` & `surviving_row_groups` each call `build_field_id_to_column_mapping`. Both run for every file w/ a filter, so the mapping is built 2x per file.

aaj3f (approved):
The perf wins here look great! Some remarks below:

fluree-db-iceberg/src/io/arrow_reader.rs
bplatz: Addressed in a072815d19d6ab5fd48dc79391b07f01b1509b83

fluree-db-query/src/r2rml/operator.rs
bplatz: Addressed in 9086ec5d28daab7c351cc8b92e99f48fa1081e0e

fluree-db-query/src/r2rml/operator.rs
bplatz: Addressed in 67dee94aec68d1f261505ba66fc32c1ed22e8a95

fluree-db-query/src/r2rml/rewrite.rs
bplatz: Addressed in 9086ec5d28daab7c351cc8b92e99f48fa1081e0e

fluree-db-iceberg/src/io/arrow_reader.rs
bplatz: Addressed in a072815d19d6ab5fd48dc79391b07f01b1509b83

fluree-db-iceberg/Cargo.toml
bplatz: Addressed in 67dee94aec68d1f261505ba66fc32c1ed22e8a95

fluree-db-iceberg/src/io/arrow_reader.rs
bplatz: Addressed in a072815d19d6ab5fd48dc79391b07f01b1509b83

PR #1413, opened 2026-07-01
feat(r2rml): bound subjects, decimal/double/date & ref object constants, star fusion, subject-key pushdown

bplatz:
## Summary

Expands R2RML/Iceberg graph-source query support along two axes: **eliminating query shapes that used to hard-error** (correctness-first) and **adding an opt-in bound-subject pushdown**. Stacked on `feature/iceberg-arrow-reader`.

Previously, any triple pattern the R2RML rewriter couldn't convert was left unconverted, which the graph rewriter treats as a hard error. Several common shapes hit that path. Each fix below is **operator-enforced** (the operator is authority), so results are correct with scan pushdown off and are verifiable with mock providers.

## Changes

- **Bound (constant) subjects** (`<store/5> ex:name ?n`): `R2rmlPattern.subject_var` is now `Option` plus a `subject_constant`; the operator materializes each row's subject and keeps only matches, binding just the object var(s).
- **Decimal / double / date constant objects** (`?s ex:price 9.99`, `?s ex:when "2024-01-15"^^xsd:date`): decimals match numerically (scale-insensitive, `9.99` == `9.990`), doubles by f64 value, dates by parsing the materialized ISO-8601 lexical back to days-since-epoch. Decimal/double are operator-only; date also emits a scan filter.
- **Ref objects arriving as `FlakeValue::Ref`** are now decoded to an IRI constant instead of being rejected.
- **Constant-object star fusion**: a constant-object triple sharing a subject with a var-object star (`?s ex:name ?n ; ex:storeId "X"`) folds into the star as an existence constraint, eliminating a separate scan + self-join.
- **Bound-subject key pushdown (opt-in, off by default)**: `reverse_subject_template` — the exact inverse of the R2RML template expansion + percent-encoding — recovers the key column value from a constant subject IRI and pushes it to the Iceberg scan as an equality predicate for pruning. Gated behind `FLUREE_R2RML_SUBJECT_KEY_PUSHDOWN` pending live validation on production backends. The operator still enforces the subject equality, so enabling it only changes which rows the scan returns, never the result.

## Correctness notes

- The pushdown relies on a **percent-encoding contract**: query subject IRIs must use the same encoding the mapping generates (`west/5` → `.../store/west%2F5`). A mis-encoded IRI matches no subject *identically* whether pushdown is on or off, so it can never cause a divergence — only a query-authoring footgun. Documented in `docs/graph-sources/r2rml.md`.
- Reversal engages only for unambiguously-reversible template shapes (single trailing placeholder, or hard-separator multi-placeholder) and integer/string key columns; everything else falls back to a full scan.
- Integer keys on decimal columns reuse the already-validated Int64→Decimal Arrow-cast path. Decimal(scale>0)/date key columns and IRI-object FK pushdown are deliberately left as follow-ups (each wants one live check).

## Tests

- Template reversal round-trips (incl. `/`, space, unicode, literal `%`), ambiguous-shape bails, non-matching IRIs.
- Physical-type coercion (`build_iceberg_filter` TemplateKey → int/long/decimal/string, skip otherwise).
- End-to-end: the flag emits the expected `store_key` filter and the result stays correct via operator authority.
- Guards for bound subject, decimal object equality, and single-scan star fusion.

## Docs

- `docs/graph-sources/r2rml.md`: bound-subject key pushdown section (encoding contract + query-authoring rule).
- `docs/operations/configuration.md`: `FLUREE_R2RML_SUBJECT_KEY_PUSHDOWN` flag.


fluree-db-query/src/r2rml/operator.rs
aaj3f: `pattern_predicates()` (the source for both the star projection and RefObjectMap parent-lookup construction) includes `predicate_filter` and `star_bindings` but **not** `star_constraints`:

```rust
fn pattern_predicates(&self) -> Vec<&str> {
    let mut preds = Vec::new();
    if let Some(p) = self.pattern.predicate_filter.as_deref() { preds.push(p); }
    for (pred, _) in &self.pattern.star_bindings { preds.push(pred.as_str()); }
    preds   // star_constraints predicates are never included
}
```

For the headline case `?s ex:name ?n ; ex:storeId "X"`, the base has a single var-member (`ex:name`), so `star_bindings` is **empty** and `star_constraints = [(ex:storeId, "X")]`. The projection then takes the `star_bindings.is_empty()` branch (`operator.rs:548`) -> `columns_for_predicate(Some("ex:name"))`, which projects the subject columns + `store_name` but **not `store_id`**. The real provider prunes to exactly that projection (`fluree-db-api/src/graph_source/r2rml.rs:~1010`, `.with_projection(projected_field_ids)`), so `store_id` is absent from the batch. In `materialize_batch` the constraint loop (`operator.rs:1335-1358`) calls `materialize_pom_object` for `ex:storeId`, whose column read returns `None` → `matched=false` → `row_ok=false` → **every subject row is dropped.**

**Failure scenario:** mapping `dw.store(store_key, store_name, store_id)`, `ex:name→store_name`, `ex:storeId→store_id`; data STORE-2 = ("Bravo","STORE-2"); query `?s ex:name ?n ; ex:storeId "STORE-2"`. Expected 1 row; a real scan returns 0. The *unfused* standalone form (`?s ex:storeId "STORE-2"` alone) works, so fusion is a strict correctness regression versus not fusing. Same root cause hits the RefObjectMap variant (parent lookups are built only for `star_preds`, `operator.rs:~650`).

**Why tests miss it:** `guard_constant_object_fuses_into_star_single_scan` asserts `rows==1`, but `CountingProvider` (and the other star mocks) *ignore* the projection and return the full pre-built batch — so `store_id` is present in the mock. Only the real reader prunes columns.

**Fix:** include `star_constraints` predicates in `pattern_predicates()` (fixes both the projection and `filtered_poms` parent-lookup construction), and add a projection-honoring mock (or a real-reader integration test) that would have caught it.

fluree-db-query/src/r2rml/operator.rs
```diff
@@ -1220,9 +1276,27 @@ fn materialize_batch(
...
 
-        if !pattern.star_bindings.is_empty() {
+        // Seed a fresh output row with the subject binding, or an empty row when
+        // the subject is a constant (which binds no variable).
+        let seed_row = || -> Vec<(VarId, Binding)> {
+            match pattern.subject_var {
+                Some(sv) => vec![(sv, subject_binding.clone())],
+                None => Vec::new(),
+            }
+        };
```
aaj3f: `seed_row` adds an alloc+realloc (and a clone) per emitted row in the two commonest scan paths

```rust
let seed_row = || -> Vec<(VarId, Binding)> {
    match pattern.subject_var {
        Some(sv) => vec![(sv, subject_binding.clone())],  // capacity 1
        None => Vec::new(),
    }
};
// single-object path:
let mut row = seed_row();            // alloc cap-1
row.push((obj_var, object_binding)); // full → realloc to cap-2 (+ memcpy + free)
```

The old single-object code was `produced.push(vec![(sv, subject_binding.clone()), (obj_var, ob)])` — one allocation at capacity exactly 2, no realloc. The new code allocs cap-1 then reallocs to cap-2 per row → ~2× allocator traffic per output row in the hottest R2RML path (a scan can emit millions of rows). Separately, the **subject-only** path (`operator.rs:~1414`) old code *moved* `subject_binding` into the row; because `seed_row` borrows it, it now `clone()`s every row (an extra `Arc<str>` refcount bump / `Sid` copy). Both regressions fire even when **none** of the new features is used.

**Fix:** don't route the common paths through the closure. 

Single-object: `let mut row = Vec::with_capacity(2); if let Some(sv) = pattern.subject_var { row.push((sv, subject_binding.clone())); } row.push((obj_var, object_binding));`

Subject-only: `produced.push(match pattern.subject_var { Some(sv) => vec![(sv, subject_binding)], None => Vec::new() });` (move, no clone). Keep the closure only for the star branches that need multiple copies.

fluree-db-api/src/graph_source/r2rml.rs
```diff
@@ -97,6 +97,25 @@ fn build_iceberg_filter(
...
+                    Ok(v) => LiteralValue::Int32(v),
+                    Err(_) => continue,
+                },
+                Some(t) if t == "long" || t.starts_with("decimal") => match s.parse::<i64>() {
+                    Ok(v) => LiteralValue::Int64(v),
+                    Err(_) => continue,
+                },
+                Some("string") => LiteralValue::String(s.clone()),
+                _ => continue,
+            },
```
aaj3f: The `Date` arm emits a `Date` literal unconditionally, unlike `Int` and `TemplateKey` which gate on `field.type_string()`:

```rust
ScanValue::Date(d) => LiteralValue::Date(*d),   // no field.type_string() check
ScanValue::Int(n)  => match field.type_string() { Some("int") => ..., _ => continue },  // guarded
ScanValue::TemplateKey(s) => match field.type_string() { ... },                         // guarded
```

`build_scan_filters` pushes the `Scalar(Date)` constant against the object column with no physical-type check, and the Arrow reader applies it as an **exact row filter** (`eval_comparison` casts `Date32 → Utf8` = `"2024-01-15"`). But the operator enforces with lenient `Date::parse` (`operator.rs:~1242`), which accepts `"2024-01-15Z"` / `"2024-01-15+05:00"`. So for an `xsd:date` object mapped onto a **physically string** column holding `"2024-01-15Z"`, the operator keeps the row but the pushed row filter drops it (`"2024-01-15Z" != "2024-01-15"`) → missing rows, violating the "pushdown never removes an operator-kept row" invariant the code states at `operator.rs:~1204`. Default-on.

**Fix:** gate the `Date` arm on `field.type_string() == Some("date")`, matching the other arms. (Related to the templated/transformed-object-map pushdown gap noted in the #1411 review; this is a distinct, cleanly-fixable instance.)

fluree-db-r2rml/src/materialize/term.rs
```diff
@@ -198,6 +198,120 @@ fn iri_escape(value: &str) -> String {
...
+            let v = rest;
+            rest = "";
+            v
+        } else {
+            // The separator must begin with a hard (always-escaped) char so the
+            // value's right boundary is unambiguous.
+            if !sep.chars().next().is_some_and(is_always_escaped) {
+                return None;
+            }
+            let idx = rest.find(sep)?;
```
aaj3f, line 308: a `%`-led separator passes the `is_always_escaped` guard but collides with `%XX` inside encoded values → wrong recovered key → silent missing results

The separator-boundary guard only checks the separator's first char is always-escaped:

```rust
if !sep.chars().next().is_some_and(is_always_escaped) { return None; }
let idx = rest.find(sep)?;
```

The reasoning ("a hard delimiter can never appear literally inside an encoded value") has one hole: `%` **is** always-escaped (a literal `%` in a source value encodes to `%25`), so `is_always_escaped('%')` is `true` — but `%` is also the escape-introducer that begins every `%XX` byte *inside* an encoded value. So a separator starting with `%` passes the guard yet `rest.find(sep)` can match a false
boundary inside a value.

**Failure scenario:** template `http://ex/person/{first}%20{last}` (a `%20`/space composite-key separator), row `first="Mary Ann"`, `last="Smith"` → generated/queried IRI `http://ex/person/Mary%20Ann%20Smith`. Reversal finds `%20` at index 4 (inside `first`'s encoded space), recovering `first="Mary"`, `last="Ann Smith"` — **wrong keys**. The pushdown then filters `first Eq "Mary" AND last Eq "Ann Smith"`, pruning the actual row before the operator's subject check runs → the exact subject the user queried returns nothing. Default-on; no test uses a `%XX` separator.

**Fix (precise):** `%` is provably the only always-escaped char in `iri_escape`'s output alphabet, so reject a `%`-led separator: `if !sep.chars().next().is_some_and(|c| is_always_escaped(c) && c != '%') { return None; }` — such templates then fall back to a full scan (correct, just unpruned).

*(Note: `is_always_escaped` is otherwise the exact complement of `iri_escape`'s literal set, and `percent_decode` / `strip_prefix` / the ambiguous-shape bails are all correct — verified char-by-char. This one lead-char case is the sole hole.)*

fluree-db-query/src/r2rml/rewrite.rs
aaj3f, line 641: With a bound subject and a variable predicate, `predicate_filter = None` (`Ref::Var(_) => None`) and `object_var = Some(?o)`. Pre-PR, bound subjects returned `None` and the triple was preserved for normal evaluation; now it converts to a bound-subject pattern with no predicate-variable field, so `?p` is dropped. At materialize time `predicate_filter=None` matches *every* POM, binding `?o` to every object of `store/5` across all predicates with `?p` left unbound. `SELECT ?p ?o WHERE { <store/5> ?p ?o }` returns objects with `?p = NULL` instead of `(predicate, object)` pairs — a wrong-results regression for that shape (the var-subject `?s ?p ?o` wildcard has the same latent limitation, but this PR newly routes bound-subject wildcards into it). 

**Fix:** in `convert_triple_to_r2rml`, return `None` when `predicate_filter` is `None` and the object is a variable, leaving the triple for normal evaluation.

fluree-db-query/src/r2rml/operator.rs
```diff
@@ -1180,6 +1209,21 @@ fn rdf_term_eq_object_constant(term: &RdfTerm, constant: &crate::r2rml::ObjectCo
...
         ObjectConstant::Iri(iri) => matches!(term, RdfTerm::Iri(v) if v == iri),
+        // Decimal / big-integer object: numeric (scale-insensitive) match, so a
+        // query `9.99` matches a column materialized as `9.990`.
+        ObjectConstant::Decimal(d) => {
+            let RdfTerm::Literal { value: v, .. } = term else {
+                return false;
+            };
+            v.parse::<bigdecimal::BigDecimal>().is_ok_and(|x| &x == d)
+        }
+        // Double object: exact f64 value match.
```
aaj3f, line 1260: decimal object match parses a fresh `BigDecimal` per row across the whole scan

`ObjectConstant::Decimal` is **not** pushed as a scan filter (`build_scan_filters` pushes only `Scalar`), so a query `?s ex:price 9.99` reads every row and `v.parse::<BigDecimal>()` heap-allocates a digit vector per row. Correct, but avoidable: add a lexical fast-path (`v == d.to_string() || v.parse…`) or push a numeric scan filter so the reader prunes first. (The `Double`/f64 arm is fine — `parse::<f64>` doesn't allocate.)

fluree-db-api/src/graph_source/r2rml.rs
```diff
@@ -97,6 +97,25 @@ fn build_iceberg_filter(
...
+            // pushdown — the operator still enforces the subject equality.
+            ScanValue::TemplateKey(s) => match field.type_string() {
+                Some("int") => match s.parse::<i32>() {
+                    Ok(v) => LiteralValue::Int32(v),
+                    Err(_) => continue,
+                },
+                Some(t) if t == "long" || t.starts_with("decimal") => match s.parse::<i64>() {
+                    Ok(v) => LiteralValue::Int64(v),
+                    Err(_) => continue,
+                },
```
aaj3f, line 126: scale>0 decimal key columns ARE pushed, contrary to the docstring

`Some(t) if t == "long" || t.starts_with("decimal")` matches `decimal(10,2)`, but the doc comment and `docs/graph-sources/r2rml.md` say only integer-valued decimals are pushed. It's *sound* (an integer key only parses to `i64` when the column materializes without a point, so the operator's subject-IRI equality keeps a superset of ∅; a real `"5.00"` key fails `i64` parse and is skipped), so no missing rows — but the comment overstates what prunes. Correct the wording to avoid a future reader relying on it.

aaj3f (approved):
Looks great and like the "operator stays authority; pushdown is an optimization" -- just relaying some correctness or perf gaps below:

fluree-db-query/src/r2rml/operator.rs
bplatz: Addressed in 04a471d3ab4e31554d1952e8fbe7198572bf5984

fluree-db-query/src/r2rml/operator.rs
bplatz: Addressed in 371c4c91c3f596b9550dc6cfb1f9cd81ad1584a9

fluree-db-api/src/graph_source/r2rml.rs
bplatz: Addressed in 04a471d3ab4e31554d1952e8fbe7198572bf5984

fluree-db-r2rml/src/materialize/term.rs
bplatz: Addressed in 371c4c91c3f596b9550dc6cfb1f9cd81ad1584a9

fluree-db-query/src/r2rml/rewrite.rs
bplatz: Addressed in 04a471d3ab4e31554d1952e8fbe7198572bf5984

fluree-db-query/src/r2rml/operator.rs
bplatz: Addressed in 0bf7e0b6bfa7d1f81ef39e5c5fe6541bf73a477c

fluree-db-api/src/graph_source/r2rml.rs
bplatz: Addressed in 5663d5b09ca0de2ec08b5ee5c8886a48135c47d1

PR #1417, opened 2026-07-01
fix(query): cross-ledger explicit projection drops foreign-namespace predicates

### LINKED ISSUE #1295: Cross-ledger explicit projection silently drops foreign-namespace predicates
## Summary

PR #1267 made cross-ledger hydration *expansion* work — a ref into another ledger now hydrates instead of collapsing to a bare `{"@id": ...}`. But it only fixed the **wildcard** (`["*"]`) path. With an **explicit** nested projection, every predicate on the foreign subject whose namespace code differs between the primary ledger and the subject's home ledger is silently dropped. Only `@id` and predicates in **reserved** namespaces (`rdfs`, `rdf`, `xsd`, `owl`, …) reliably survive, because those codes are fixed across all ledgers.

Confirmed on `v4.0.5`. Follow-up to #1259 / #1267.

## Reproduction

Two ledgers sharing `schema.org`: `catalog` (primary) holds a `schema:Product` whose `schema:author` ref points into a separate `people` ledger. The person carries `p:fullName` (`http://people.example/`) — a predicate whose namespace is allocated a **different code in each ledger** (see the seed in the test below). The projection crosses `catalog → people`.

```jsonc
{
  "@context": {"schema": "http://schema.org/",
               "cat": "http://catalog.example/",
               "p": "http://people.example/",
               "id": "@id", "type": "@type"},
  "from": ["test/catalog:main", "test/people:main"],
  "select": { "?item": ["@id", { "schema:author": ["@id", "schema:name", "p:fullName"] }] },
  "where": { "@id": "?item", "type": "schema:Product" }
}
```

**Actual** — `p:fullName` is gone:
```json
{ "schema:author": { "@id": "p:ada", "schema:name": "Ada Lovelace" } }
```

- **Wildcard works** — `"schema:author": ["*"]` returns `p:fullName` too.
- **Single-ledger explicit works** — the same `["@id","schema:name","p:fullName"]` projection against `test/people:main` alone returns all three.

So the fault is specific to the **explicit projection × cross-ledger** combination. (`schema:name` survives only because `schema.org` happens to share a code across the two ledgers in this seed; `@id` and any reserved-namespace predicate like `rdfs:label` always survive.)

## Reproduction test

Drops into `fluree-db-api/tests/it_multi_ledger_formatting.rs` and uses the existing suite helpers (`genesis_ledger`, `MemoryFluree`, `FlureeBuilder`). It is **red on `v4.0.5`** — the final assertion fails with `left: None, right: Some("Augusta Ada King")`.

```rust
/// Seed a primary `catalog` ledger whose product REFERENCES a person in a
/// separate `people` ledger, where the person carries a predicate (`p:fullName`)
/// in a namespace that is allocated a DIFFERENT code in each ledger:
///   - catalog: people.example registered last (via the `schema:author` ref) → high code
///   - people:  people.example registered first (the subject `p:ada`)        → low code
/// This is the cross-ledger explicit-projection drop condition.
async fn seed_divergent_predicate_ref_ledgers(fluree: &MemoryFluree) {
    fluree
        .insert(
            genesis_ledger(fluree, "test/catalog:main"),
            &json!({
                "@context": {
                    "schema": "http://schema.org/",
                    "cat": "http://catalog.example/",
                    "p": "http://people.example/",
                    "id": "@id", "type": "@type",
                },
                "@graph": [
                    {"@id": "cat:item1", "@type": "schema:Product",
                     "schema:author": {"@id": "p:ada"}}
                ]
            }),
        )
        .await
        .expect("insert catalog");

    fluree
        .insert(
            genesis_ledger(fluree, "test/people:main"),
            &json!({
                "@context": {
                    "schema": "http://schema.org/",
                    "p": "http://people.example/",
                    "id": "@id", "type": "@type",
                },
                "@graph": [
                    {"@id": "p:ada", "@type": "schema:Person",
                     "schema:name": "Ada Lovelace", "p:fullName": "Augusta Ada King"}
                ]
            }),
        )
        .await
        .expect("insert people");
}

/// Cross-ledger NESTED EXPLICIT projection must return ALL projected predicates
/// of the foreign subject — not just `@id` and reserved-namespace ones.
#[tokio::test]
async fn cross_graph_nested_explicit_projection_divergent_predicate() {
    seed_divergent_predicate_ref_ledgers(&fluree).await;

    let q = json!({
        "@context": {
            "schema": "http://schema.org/",
            "cat": "http://catalog.example/",
            "p": "http://people.example/",
            "id": "@id", "type": "@type",
        },
        "from": ["test/catalog:main", "test/people:main"],
        // EXPLICIT projection (not ["*"]) crossing catalog -> people:
        "select": { "?item": ["@id",
            { "schema:author": ["@id", "schema:name", "p:fullName"] }
        ]},
        "where": { "@id": "?item", "type": "schema:Product" }
    });

    let value = fluree
        .query_from()
        .jsonld(&q)
        .execute_formatted()
        .await
        .expect("execute_formatted should not error");

    let item = value.as_array().and_then(|a| a.first()).expect("one item");
    let author = item.get("schema:author").expect("author present");

    assert_eq!(author.get("@id").and_then(|v| v.as_str()), Some("p:ada"));
    // Reserved/shared-code predicate survives even pre-fix:
    assert_eq!(
        author.get("schema:name").and_then(|v| v.as_str()),
        Some("Ada Lovelace"),
    );
    // THE REGRESSION: divergent-namespace predicate must be present, not dropped.
    assert_eq!(
        author.get("p:fullName").and_then(|v| v.as_str()),
        Some("Augusta Ada King"),
        "cross-ledger explicit projection dropped a non-primary-namespace predicate: {value:#}"
    );
}
```

## Root cause

A dataset query is parsed/lowered **once, against the primary ledger's namespace dictionary** (`fluree-db-api/src/view/dataset_query.rs:95` — *"using primary db for namespace resolution"*). So every predicate in the lowered `NestedSelectSpec` is a `Sid { namespace_code, name }` carrying the **primary** ledger's code.

Cross-ledger expansion in `expand_ref` (`fluree-db-api/src/format/hydration.rs:960`) re-encodes the **subject IRI** for the target view (`encode_iri_strict`, `:993`) but passes the projection `level` — with its primary-dict predicate Sids — through **unchanged** (`:998`). On the target view those stale Sids hit two gates, both keyed on the full `Sid`:

1. the per-predicate fetch filter (`predicate_filter_for_level`, `:150`; applied at `:849`), and
2. the `select_predicate` match in `format_subject` (`:868`).

User-namespace codes are allocated **lazily, per ledger, in first-use order** (`fluree-db-core/src/namespaces.rs:9`), so `Sid(<primary code>, "fullName") ≠ Sid(<home code>, "fullName")` → no match → silently dropped. `rdfs:label` is `Sid(4, "label")` in *every* ledger (`RDFS = 4` is reserved, `fluree-vocab/src/lib.rs:1424`), so it survives — which is why a reserved-namespace predicate can *look* special-cased. It isn't; it's just namespace-code-stable.

The WHERE/executor path already handles this — it stamps `Binding::IriMatch` with per-ledger provenance and re-encodes across ledgers (`fluree-db-query/src/dataset.rs:57-61`). The projection-hydration path never got the same treatment.

## Why #1267's tests didn't catch it

Every cross-ledger expansion test added in #1267 uses `["*"]` + `depth` (`it_multi_ledger_formatting.rs:133,181,319,372`), where the predicate-Sid mismatch is invisible (wildcard sets no filter and matches all predicates). The one explicit nested test (`:460-495`) is a `fromNamed`+GRAPH query whose ref stays inside its named graph and whose assertions only check the root `@id`, not foreign predicate completeness. There's even a `NOTE (staged)` at `format/hydration.rs:474-476` acknowledging *"correct only when the two ledgers allocated matching namespace codes."*

## Direction (gesture, not a prescription)

The wildcard path already works, so the explicit-projection path is where to look. The general shape: the projection hydrator should reconcile predicate identity against the ledger that actually stores the subject, rather than relying on the Sids resolved once against the primary dict — the executor already does an analogous per-ledger reconciliation for WHERE bindings (`fluree-db-query/src/dataset.rs:57-61`), so there's an existing pattern to follow. The same divergence likely affects a non-primary hydration **root** under an explicit projection, not just nested refs. This looks contained to `fluree-db-api`'s formatter with no `QueryResult` change.

Jackamus29:
Closes #1295.

## Problem
A cross-ledger JSON-LD **explicit** projection over a multi-ledger `from` silently drops any predicate whose namespace code differs between the primary ledger and the ledger that actually stores the subject. Reserved-namespace predicates (`rdfs`, `rdf`) survive because their codes are stable across ledgers; user-namespace predicates (e.g. `abbrevName`) are silently elided. This happens both for **nested** refs and for a top-level **root** whose subject lives in a non-primary ledger.

## Root cause
A dataset query is parsed/lowered **once** against the primary ledger's namespace dictionary, so every predicate `Sid` in the projection spec carries the *primary* ledger's namespace code. When hydration renders a subject in another ledger, it re-encodes the *subject* IRI for the target view but passes the projection `level` through unchanged. Those primary-dict predicate `Sid`s then miss the target ledger's index in the predicate filter and fail `select_predicate`'s `Sid`-equality match — both silently.

## Fix
`rebind_level_to_view` re-encodes the immediate projection level's predicate `Sid`s into the target ledger's namespace dictionary, so both the predicate filter and `select_predicate` match the `Sid`s as the target ledger actually stores them. It is applied at the two points a subject can be rendered against a non-primary view:

- **Nested refs** (`expand_ref`) — rebind at each cross-ledger crossing.
- **Root subjects** (`format_hydration_column`) — when `FormatterSet::pick` routes a root to its home (non-primary) ledger, rebind the level into that view's dict.

Single-ledger and same-ledger (primary) subjects are untouched and byte-identical (no allocation). The re-encoding is memoized per `(view, level)` for the response (`HydrationCaches`).

## Tests
- `cross_graph_nested_explicit_projection_divergent_predicate` — the nested repro, now passing.
- `cross_graph_root_projection_divergent_predicate` — the root repro (red→green): a home-graph-bound root hydrates `schema:name` (shared code) but dropped `p:fullName` pre-fix.
- `cross_graph_wildcard_refinement_divergent_ns` — covers the wildcard-refinement re-encode (previously uncovered).
- Existing cross-ledger guards stay green: cross-graph IRI decode, depth-3 wildcard expansion, home-ledger property decode, and named-graph scoped projection.

## Out of scope (follow-up)
The **non-primary hydration root routing** case — a top-level result variable bound as the *object* of a primary-ledger triple, whose subject lives in another ledger — comes back `@id`-only because routing itself fails to reach the home view (so even shared-code predicates are dropped). That's a distinct routing defect (`FormatterSet::pick` has no home-ledger provenance for a bare object SID), left as an `#[ignore]`d repro (`cross_graph_root_bound_as_object_hydrates_in_home_ledger`) pending a separate fix. See the note on #1295.

### DIFF
--- a/fluree-db-api/src/format/hydration.rs
+++ b/fluree-db-api/src/format/hydration.rs
@@ -492,12 +492,46 @@ async fn format_hydration_column(
     };
 
     let formatter = set.pick(root.ledger_alias.as_ref());
+
+    // Re-encode the projection level's predicate Sids into the routed view's
+    // namespace dict when the root subject lives in a non-primary ledger
+    // (issue #1295, root path). The level is lowered once against the primary
+    // dict, so predicate Sids carry the primary ledger's namespace codes; a
+    // root routed to a non-primary view would otherwise miss that ledger's
+    // index for any divergent-code predicate (reserved-namespace predicates
+    // survive only because their codes are stable). A root that stays on the
+    // primary view needs no rebind — keep the original level (no allocation,
+    // byte-identical output). This mirrors the nested `expand_ref` rebind and
+    // shares its memo (per target view + level address). The lowering/decode
+    // source is the primary view's compactor, never the routed view's.
+    let rebound_arc;
+    let level = if formatter.active_idx == set.primary {
+        &spec.level
+    } else {
+        let key = (
+            formatter.active_idx,
+            &spec.level as *const NestedSelectSpec as usize,
+        );
+        rebound_arc = cache
+            .rebinds
+            .entry(key)
+            .or_insert_with(|| {
+                Arc::new(rebind_level_to_view(
+                    &spec.level,
+                    set.primary().compactor,
+                    formatter.db.snapshot,
+                ))
+            })
+            .clone();
+        &*rebound_arc
+    };
+
     let mut visited = HashSet::new();
     formatter
         .format_subject(
             &root.sid,
             root.iri,
-            &spec.level,
+            level,
             DepthBudget::root(spec.depth),
             &mut visited,
             cache,
--- a/fluree-db-api/src/format/hydration.rs
+++ b/fluree-db-api/src/format/hydration.rs
@@ -49,7 +49,7 @@ use fluree_db_core::comparator::IndexType;
 use fluree_db_core::query_bounds::RangeOptions;
 use fluree_db_core::range::{RangeMatch, RangeTest};
 use fluree_db_core::value::FlakeValue;
-use fluree_db_core::{Flake, GraphDbRef, Sid, Tracker};
+use fluree_db_core::{Flake, GraphDbRef, LedgerSnapshot, Sid, Tracker};
 use fluree_db_policy::{is_schema_flake, PolicyContext};
 use fluree_db_query::binding::Binding;
 use fluree_db_query::ir::{Column, ForwardItem, HydrationSpec, NestedSelectSpec, Root};
@@ -81,6 +81,24 @@ use std::sync::Arc;
 /// query can never serve one ledger's rendering for another ledger's subject
 /// (issue #1259).
 type CacheKey = (usize, Sid, u64, usize, Option<Arc<str>>);
+
+/// Per-response hydration caches, threaded through the recursive formatter.
+///
+/// - `results`: memoizes a subject's finished JSON so a subject reached many
+///   times in one response (shared ancestors/units, multiply-referenced
+///   entities) is hydrated once.
+/// - `rebinds`: memoizes a projection level re-encoded into a target ledger's
+///   namespace dict (issue #1295), keyed by `(target view index, level
+///   address)`, so each `(level, target)` is rebound at most once per response
+///   and reused via a shared `Arc` instead of re-cloning the sub-tree on every
+///   cross-ledger crossing. The address key is stable for the response
+///   (originals live in the `HydrationSpec`, rebounds are held here) and is only
+///   ever compared, never dereferenced.
+#[derive(Default)]
+struct HydrationCaches {
+    results: HashMap<CacheKey, JsonValue>,
+    rebinds: HashMap<(usize, usize), Arc<NestedSelectSpec>>,
+}
 
 /// Depth bookkeeping for one hydration call.
 ///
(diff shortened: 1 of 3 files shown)


bplatz (approved):
👍

PR #1418, opened 2026-07-02
fix(query): resolve hydration roots across the default-graph union

Jackamus29:
Stacked on #1417 (base is that PR's branch; will retarget to `main` once it merges).

## Problem
A multi-ledger hydration **root** was routed to a single home ledger (`FormatterSet::pick`) and formatted there, while nested refs (`expand_ref`) resolve across the whole default-graph union and RDF-merge. Two consequences under a `from: [A, B]` default-graph union:

- A subject described in **several** default-graph ledgers — same multi-cardinality predicate, e.g. `skos:altLabel` with different values in each — rendered only its provenance ledger's values at the root. No union.
- A root with **no home-ledger provenance** (bound as the object of a primary-ledger triple) fell back to the primary view and came back `@id`-only (previously tracked as "Finding B").

Both are correct for named-graph queries (a named graph is addressed individually — no merge) and never affected single-ledger queries.

## Fix
Extract `expand_ref`'s union+merge body into a shared `resolve_iri_across_union`, and route the multi-ledger root through it:

- Recover the root's canonical IRI — from the `IriMatch` binding, or by decoding the bare primary-dict `Sid` against the primary view.
- Resolve it across the default-graph union, re-encoding the projection level per contributing view (#1295) and merging (`merge_subject_objects`).

Scope correctness falls out of `pick` + `active_idx`: an IriMatch root on a default ledger unions across defaults; a named-graph root stays single; a provenance-less root starts at primary (a default) and finds its subject in the union. Single-ledger (`dataset == None`) formats directly — byte-identical. `expand_ref` folds in the extracted `async fn` with no extra boxing, so nested-ref behavior/allocation is unchanged.

## Tests
- `cross_graph_root_bound_as_object_hydrates_in_home_ledger` (Finding B) — un-ignored, now passes.
- `cross_graph_root_unions_split_subject_altlabels` — un-ignored, now passes; asserts the union of `skos:altLabel` across both ledgers **and** that the merged `@id` stays a single scalar (merge dedup).
- `cross_graph_nested_ref_unions_split_subject_altlabels` — the passing nested control on the same split subject.
- Full multi-ledger module green (0 ignored); `grp_query` / `grp_query_sparql` unchanged.

### DIFF
--- a/fluree-db-api/src/format/hydration.rs
+++ b/fluree-db-api/src/format/hydration.rs
@@ -484,45 +484,44 @@ async fn format_hydration_column(
     };
 
     let formatter = set.pick(root.ledger_alias.as_ref());
-
-    // Root path for #1295: when `pick` routes a root to a non-primary ledger,
-    // re-encode the level into that view's dict (as the nested `expand_ref` path
-    // does, sharing the memo). A primary-view root needs none — keep the original
-    // level (no allocation, byte-identical). Decode source is the primary
-    // compactor, never the routed view's.
-    let reencoded_arc;
-    let level = if formatter.active_idx == set.primary {
-        &spec.level
-    } else {
-        let key = (
-            formatter.active_idx,
-            &spec.level as *const NestedSelectSpec as usize,
-        );
-        reencoded_arc = cache
-            .reencoded_levels
-            .entry(key)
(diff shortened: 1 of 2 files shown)


bplatz (approved):
👍

PR #1419, opened 2026-07-02
fix(policy): enforce f:policySource and config policy defaults on writes

### LINKED ISSUE #1416: security: f:policySource enforced on reads but ignored on writes (modify paths hardcode default graph)
## Summary

`f:policySource` (the config-graph setting that points policy rules at a specific graph) is honored on the **read/query** path but **ignored on every write/modify path**, which hardcodes the default graph (`g_id = 0`). A policy authored into a named graph (or the config graph) via `f:policySource` is therefore enforced when querying but **silently not enforced when transacting** — an operator who believes writes are constrained by that policy is mistaken.

## Impact

Security-relevant enforcement asymmetry. If an operator relocates access policies into a dedicated policy graph via `f:policySource` (a supported, documented configuration), modify-time policy checks continue to load restrictions only from `g_id = 0`. Writes can bypass a policy the operator reasonably believes is active. Reads behave correctly, which makes the gap easy to miss.

## Evidence

**Read side — correctly wired (same-ledger and cross-ledger):**
- `fluree-db-api/src/view/fluree_ext.rs:790` — `wrap_policy` resolves the source via `resolve_policy_source_g_ids(source, &view.snapshot)` and threads the resulting graph list into `build_policy_context_from_opts`.
- `fluree-db-api/src/policy_builder.rs:50` — `resolve_policy_source_g_ids` maps the `f:policySource` graph selector to a real `g_id`; the resolved slice is looped over in `load_policies_by_identity` / `load_policies_of_classes`, so policy rules are genuinely read from the configured graph.

**Write side — hardcodes `&[0]`, never consults `f:policySource`:**
- `fluree-db-api/src/tx.rs:2554` (`credential_transact`) — `build_policy_context_from_opts(..., &[0])`
- `fluree-db-api/src/graph_commit_builder.rs:293` — `&[0]`
- `fluree-db-api/src/commit_transfer.rs:899` — `&[0]`
- `fluree-db-api/src/block_fetch.rs:385` — `&[0]`

None of these call a config-graph resolver equivalent to `resolve_policy_source_g_ids`.

For contrast, SHACL already resolves its source graph at write/validation time via `resolve_shapes_source_g_ids` (`fluree-db-api/src/tx.rs:361`), so the write path already has a precedent for reading a `GraphSourceRef` from config during staging.

## Suggested fix

Mirror `resolve_shapes_source_g_ids` on the transaction write path: resolve `f:policySource` to a graph-id list from the (pre-transaction) `LedgerConfig` and pass it into `build_policy_context_from_opts` in place of the hardcoded `&[0]` at the four call sites above. The `modify_set` is already built from the same restrictions inside `build_policy_context_from_opts_inner` (`policy_builder.rs:301`), so only the source-graph resolution needs plumbing.

Notes:
- Same-ledger only is sufficient for a first cut (matches the read path's same-ledger `resolve_policy_source_g_ids`; cross-ledger write policy is a larger, separate effort and can stay rejected/fail-closed for now).
- Fail-closed on an unresolved/missing source graph, consistent with the read-side behavior.

## Acceptance criteria

- A policy stored in a named graph via `f:policySource` is enforced on transactions (a write that violates it is rejected), not just on queries.
- Regression test: config a `f:policySource` → named graph, stage a write that the policy should deny, assert rejection; confirm the same write is allowed when the policy is absent. Use the temp-revert trick to prove the test exercises the fix.
- No change to the default behavior when `f:policySource` is unset (still `g_id = 0`).

bplatz:
Fixes #1416.

> Stacked on `feature/shacl-class-value-set` — retarget to `main` once that branch merges; the two SHACL commits will drop out on rebase, leaving this as a single-commit diff.

## Problem

`f:policySource` (the config-graph setting that points policy rules at a specific graph) was honored on the **read/query** path but ignored on the write/modify paths, which hardcoded the default graph (`g_id = 0`). Two consequences:

1. A policy relocated into a named graph — or sourced **cross-ledger** from a model ledger via `f:ledger` — was enforced on queries but silently **not enforced on transactions**.
2. Config-declared policy defaults (`f:policyClass`, `f:defaultAllow`) never merged into write-time governance at all: a request with no policy inputs ran as **root** even on a fully configured ledger, while the same request's reads were policy-constrained.

## Fix

New single choke point: **`build_transact_policy_context`** (`policy_view.rs`) — the write-side counterpart of `wrap_policy`:

- Resolves the ledger config at `to_t` and merges config policy defaults via `merge_policy_opts`, so config-declared policy governs writes even when the request carries no policy inputs.
- Same-ledger `f:policySource` resolves to graph IDs via `resolve_policy_source_g_ids` (fail-closed on unknown selectors).
- Cross-ledger `f:policySource` resolves live against the model ledger (`ArtifactKind::PolicyRules`, latest committed M) and interns the restrictions into the data ledger's term space. A cross-ledger source **always** builds a context, mirroring the read path.
- Returns `None` (root) only when neither the request nor the config supplies any policy input — unchanged behavior for unconfigured ledgers.

The cross-ledger block was extracted out of `wrap_policy` into a shared helper (`resolve_cross_ledger_policy_restrictions`), so the identity contract and the `f:policyClass` filter chain are one implementation on both read and write paths.

### Call sites rewired

| Path | Before | After |
|------|--------|-------|
| Consensus transact (LocalCommitter + Raft commit worker) | `has_any_policy_inputs()` gate → same-ledger only | new builder via `local::build_policy_context` (now takes `&Fluree`) |
| `credential_transact` | `build_policy_context_from_opts(..., &[0])` | new builder (verified DID is bind-only under cross-ledger; works when a `f:policyClass` is configured) |
| Push replication (`build_policy_ctx_for_push`) | `&[0]` | new builder; `stage_commit_flakes` takes `Option<&PolicyContext>` |
| Commit-detail fetch (`graph_commit_builder`) | `&[0]` | new builder (per-request identity/policy_class opt-in gate preserved) |
| CLI local insert/upsert/update | same-ledger `build_policy_context` | new builder (flags-only gate preserved) |
| `block_fetch` | `&[0]` | no `Fluree` handle available: same-ledger config resolution, **fails closed** on cross-ledger configs |

## Tests

New `it_policy_write_path.rs` (registered in `grp_policy`):

- config `f:policyClass`/`f:defaultAllow` enforced on writes with **no request inputs**
- same-ledger named-graph `f:policySource` modify-deny rejects a violating write (root control passes)
- cross-ledger: M holds an `f:AccessPolicy` modify-deny; violating write to D rejected, untargeted write allowed — empty request opts
- identity + cross-ledger fails closed on the write builder **only when no policy class is available anywhere** (see second commit)
- no config + no inputs still runs root

All four enforcement tests were verified to **fail** against a temp-revert to the previous hardcoded-`[0]` behavior.

Suites: grp_policy 53/53, grp_graphsource 101/101, grp_transact 159/159, consensus 48/48, server policy_integration 25/25, credential 2/2. `cargo check --workspace --all-features` and clippy (all features, all targets on touched crates) clean.

## Docs

- `docs/security/policy-in-transactions.md` — new "Config-driven write enforcement" section
- `docs/security/cross-ledger-policy.md` — transactions engage automatically; queries keep the header gate
- `docs/ledger-config/setting-groups.md`, `docs/security/programmatic-policy.md`, `docs/design/cross-ledger-model-enforcement.md`

## Second commit: identity binds `?$identity` under cross-ledger (bug fix on the first commit)

The initial Phase 1a guard rejected **any** identity-carrying request against a cross-ledger `f:policySource`, which made cross-ledger governance unusable for authenticated deployments (and made governed ledgers un-writable for identity-carrying writes). The follow-up commit replaces that with the intended contract:

- **Class-filter chain** for rule selection: request `policy_class` → config `f:policyClass` → `{f:AccessPolicy}` (anonymous only). The config's class is threaded separately because `merge_policy_opts` returns request opts unchanged when the request carries any policy input and override is permitted.
- **Identity is bind-only**: it resolves against D (new `resolve_identity_binding_sid`, mirroring identity-mode's three-state binding contract) and populates `?$identity` for M's `f:query` rules — it never selects rules. D-local `f:policyClass` triples on the identity are intentionally not consulted: a cross-ledger `f:policySource` declares M the policy authority.
- Fail-closed is retained only for the genuinely ambiguous case: identity present with **no policy class anywhere**.

Importantly, relaxing the guard alone would have been a silent fail-open: `build_policy_context_from_opts_inner` gave the identity branch unconditional priority and that branch ignored the cross-ledger restrictions. The inner builder now gives the cross-ledger branch priority, with the identity handled as a binding.

New tests cover: the reported scenario end-to-end on writes (deny enforced, untargeted writes pass), an owner-only `f:query` rule in M driven by `?$identity` on both read and write, and both temp-revert failure modes (blanket guard → fail-closed regression; identity-branch-first → silent bypass).


fluree-db-consensus/src/local.rs
```diff
@@ -362,33 +363,38 @@ pub(crate) fn execution_failure(err: ApiError) -> SubmissionError {
...
         &snap.snapshot,
         snap.novelty.as_ref(),
         Some(snap.novelty.as_ref()),
         snap.t,
         governance,
     )
     .await
-    .map(Some)
     .map_err(execution_failure)
 }
```
aaj3f, line 399: `build_policy_context` dropped its `if !governance.has_any_policy_inputs() { return Ok(None); }` fast-path and now unconditionally calls `build_transact_policy_context`, which resolves the ledger `#config` (`resolve_ledger_config`) on **every consensus commit**, inside the `for attempt in 1..=MAX_TXN_RETRIES` retry loop in `LocalCommitter::commit` (line 116), and again per Raft commit worker. For **unconfigured** ledgers this is cheap — `resolve_ledger_config` has an O(1) guard (novelty `segment_count` + per-graph stats) that returns `Ok(None)` before any scan. But for **configured** ledgers it now adds one `rdf:type f:LedgerConfig` instance scan plus ~7 setting-group range reads to *every* write, even writes that carry no policy inputs and where config declares no policy — work the write path previously never did. This is the write hot path and it is feature-necessary (you must read config to learn `f:policySource`), so it is not a scaling regression (cost is bounded by the tiny single-subject config graph, indexed by SID, not O(novelty)) — but please confirm the config graph reads are genuinely constant and consider memoizing the resolved config per `(ledger_id, to_t)` so the retry loop and back-to-back writes don't re-resolve identical config. At minimum, worth a benchmark on a configured high-write-throughput ledger before merge.

  ```rust
  // Consider: cache resolve_ledger_config by (ledger_id, to_t) — the config graph
  // is immutable at a given t, and both wrap_policy (per query) and now
  // build_transact_policy_context (per write + per retry) re-resolve it.
  ```

fluree-db-api/src/policy_view.rs
```diff
@@ -232,6 +237,188 @@ pub async fn build_policy_context(
...
+        )
+        .await?;
+        return Ok(Some(policy_ctx));
+    }
+
+    if !effective_opts.has_any_policy_inputs() {
+        return Ok(None);
+    }
+
+    let policy_graphs = policy_builder::resolve_policy_source_g_ids(source, snapshot)?;
```
aaj3f: The same-ledger branch early-returns `Ok(None)` on `!has_any_policy_inputs()` **before** calling `resolve_policy_source_g_ids` (line 409), whereas the read path (`fluree_ext.rs:744`) resolves the source graphs **unconditionally** and only then builds. Net effect: a config that declares an *invalid* same-ledger `f:policySource` (unknown graph selector) is a fail-closed error on reads but silently runs as **root** on writes when the request carries no policy inputs and config sets no `f:defaultAllow`/`f:policyClass`. The window is narrow (a meaningless config), but it is a read/write fail-closed divergence — the thing this PR exists to eliminate. Suggest validating the source before the gate:

  ```rust
  // resolve (and validate) the same-ledger selector first, matching the read path's
  // fail-closed-on-unknown-selector contract, then apply the no-inputs shortcut:
  let policy_graphs = policy_builder::resolve_policy_source_g_ids(source, snapshot)?;
  if !effective_opts.has_any_policy_inputs() {
      return Ok(None);
  }
  ```

aaj3f (approved):
This makes sense and good fix

fluree-db-consensus/src/local.rs
bplatz: Addressed in 4f8f8c83a8c3947a86b7255161a4043b2a55f441. The resolved config is now cached per-ledger on `LedgerHandle`, keyed by a new `Novelty::config_write_t` marker that advances only when a commit touches `CONFIG_GRAPH_ID` (set in both `apply_commit` and `bulk_apply_commits`). So a configured-but-static ledger resolves `#config` once per config change rather than once per write — and, since the key is invariant to non-config writes, retries in the stage/commit loop hit the cache (keying on `(ledger_id, to_t)` would not, since `to_t` advances every write). Fail-safe: the cache is consulted only at head with a loaded handle and a readable marker; time-travel, a non-`Novelty` overlay, no handle, or a marker mismatch (including the downward reset after reindex) all resolve fresh, so a miss can never serve stale (fail-open) policy. Unconfigured ledgers still short-circuit before any scan.

PR #1423, opened 2026-07-02
feat: SHACL property paths and full constraint-coverage sweep

bplatz:
## Overview

Builds on the class-value-set branch with two thrusts: full `sh:path` property-path support, and a sweep that closes every silently-broken or unenforced SHACL constraint found in a full audit of the engine. Before this branch, several constructs loaded without error but never constrained data — the worst failure mode for a validation system.

## Property paths (`sh:path`)

- Path expressions compile into a `PropertyPath` AST and evaluate against the focus node: inverse (over **any** path — `^(p1/p2)` rewrites to `^p2/^p1`), sequences (Turtle RDF lists and JSON-LD `@list`), alternatives, and the three closures, all nestable.
- Complex paths work inside nested/logical members too, not just top-level property shapes.
- Malformed paths (literal steps, multiple un-listed operator values) compile to an `Unresolvable` marker that surfaces as a violation **scoped to the shape's targets** — one broken shape cannot wedge unrelated transactions on the ledger.

## Constraints that now enforce (previously silent no-ops)

- `sh:node` — per-value on property shapes, focus-node on node shapes. Recursive shape references over cyclic data (e.g. `FriendShape → knows → sh:node FriendShape`) terminate via an active-`(focus, shape)` guard that assumes conformance on re-entry.
- Value constraints directly on a node shape (no `sh:path`), including value-only anonymous members of logical constraints.
- `sh:qualifiedValueShape` + min/max counts, including inside logical members, plus `sh:qualifiedValueShapesDisjoint` (sibling set per the spec definition: all property shapes of the parent node shape, excluding the own qualified shape by value).
- `sh:uniqueLang` / `sh:languageIn` — language tags threaded from flake metadata through validation (including path evaluation); `langMatches` basic-range semantics; case-insensitive uniqueness per BCP 47.
- `sh:deactivated`, implicit class targets (shape that is also `rdfs:Class`/`owl:Class`), Turtle RDF-list `sh:ignoredProperties`.

## Semantics fixes in already-enforced constraints

- `sh:message` now surfaces (top-level shapes and anonymous nested members).
- `sh:severity` honored on node-level structural constraints and on node shapes with direct value constraints — warn-severity shapes no longer reject.
- `sh:pattern`/`sh:minLength`/`sh:maxLength` match the lexical form per SPARQL `STR()`: numeric/date literals participate, IRIs match their full decoded IRI, blank nodes fail per spec. (An IRI whose namespace is allocated in the same transaction can't be decoded against the base snapshot and fails closed.)
- `sh:in`/`sh:hasValue` compare numerics by value across representations (`1` matches `1.0`).
- `sh:class` and qualified shapes evaluate inside nested members with the full membership context (`f:shapesSource` vocabulary graphs, per-txn memo, cross-ledger model) instead of dropping it at the named-ref boundary.

## Policy

Bulk import deliberately bypasses SHACL (documented in the cookbook): validate source data against the shapes before importing; transaction-time validation keeps the ledger clean from there.

## Testing

62 SHACL integration tests (each constraint has pass + violate cases; regressions pinned via temp-revert where practical, e.g. the disjoint test carries its own control) and 57 unit tests. Full workspace `clippy --all-features --all-targets` and `nextest --workspace --all-features` pass; the one unrelated flake (`raft_multi_node::liveness_monitor_demotes_killed_follower`) passes in isolation.

## Known limitations (documented)

- `sh:targetNode` with literal values is not compiled (focus nodes are subject ids throughout the engine).
- `sh:sparql` constraints are not implemented.
- Shapes recompile per transaction; `ShaclCacheKey.schema_epoch` is ready for a cross-transaction cache if profiling warrants it.


fluree-db-shacl/src/path.rs
```diff
@@ -0,0 +1,574 @@
...
+
+/// Resolve a single `sh:path` object node (a predicate IRI or a path-expression
+/// blank node) into a [`PropertyPath`].
+fn resolve_path_node<'a>(db: GraphDbRef<'a>, node: &'a Sid) -> PathFuture<'a, PropertyPath> {
+    Box::pin(async move {
+        // sh:inversePath — inverse of any path, rewritten into the AST
+        // (inverse of a sequence = reversed sequence of inverses, etc.).
+        if let Some(inner) = operand_path(db, node, &shacl(predicates::INVERSE_PATH)).await? {
+            return Ok(invert(inner));
+        }
```
aaj3f, line 210: `resolve_path_node` probes six operator predicates (`sh:inversePath`, `sh:alternativePath`, the three closures, and `rdf:first`) via `operand_path`/`has_object` — each a `db.range` SPOT scan — before falling through to `PropertyPath::Predicate`. For a plain-predicate `sh:path` (the overwhelmingly common case) that is ~6 index probes that all return empty. This runs inside `resolve_paths`, which `compile_from_dbs` invokes **per transaction** on any ledger that has shapes (`fluree-db-api/src/tx.rs:578` → `from_dbs_with_overlay`; shapes are recompiled every transaction, not cached by `schema_epoch`). So a schema with N property shapes adds ~6·N empty index scans to every write. Per the SHACL spec an IRI-valued `sh:path` is *always* a predicate, and path-expression structures are always blank nodes — so a `namespace_code != BLANK_NODE` short-circuit is both a correctness-preserving and a large constant-factor win. The base branch stored the path as a bare `Sid` with zero scans, so this is a genuine regression introduced here.

  ```rust
  fn resolve_path_node<'a>(db: GraphDbRef<'a>, node: &'a Sid) -> PathFuture<'a, PropertyPath> {
      Box::pin(async move {
          // Fast path: an IRI in sh:path is always a plain predicate (SHACL spec);
          // only blank nodes carry path-expression structure. Skip the six operator
          // probes for the common case — this runs per property shape per compile,
          // and compiles happen per transaction.
          if node.namespace_code != BLANK_NODE {
              return Ok(PropertyPath::Predicate(node.clone()));
          }

          // sh:inversePath — inverse of any path, rewritten into the AST ...
          if let Some(inner) = operand_path(db, node, &shacl(predicates::INVERSE_PATH)).await? {
              return Ok(invert(inner));
          }
          // ... rest unchanged
  ```
  (`BLANK_NODE` is already imported at line 22.)

fluree-db-shacl/src/path.rs
```diff
@@ -0,0 +1,574 @@
...
+    }
+    Ok(dedup(out))
+}
+
+/// Deduplicate value nodes (SHACL value nodes are a set).
+fn dedup(mut values: Vec<PathValue>) -> Vec<PathValue> {
+    let mut seen: HashSet<String> = HashSet::new();
+    values.retain(|(v, dt, lang)| seen.insert(format!("{v:?}|{dt:?}|{lang:?}")));
+    values
+}
```
aaj3f, line 488: `dedup` builds a fresh heap `String` per value node via `format!("{v:?}|{dt:?}|{lang:?}")` as its hash-set key. On a `sh:zeroOrMorePath`/`oneOrMorePath` closure over a large graph (or a wide alternative/sequence) this is an O(n) allocation + formatting pass on every step's output. Prefer a structural key that hashes the components directly — `FlakeValue`, `Sid`, and `Option<String>` should be `Hash`/`Eq` already (they key the class-membership memo and `visited` sets elsewhere), so:

  ```rust
  fn dedup(mut values: Vec<PathValue>) -> Vec<PathValue> {
      let mut seen: HashSet<(FlakeValue, Sid, Option<String>)> = HashSet::new();
      values.retain(|(v, dt, lang)| seen.insert((v.clone(), dt.clone(), lang.clone())));
      values
  }
  ```
  Still clones, but avoids the format-string allocation and the Debug-representation fragility. Scope is complex-path shapes only (opt-in), hence minor. *(Verification: the suggested tuple key compiles — `FlakeValue` has total manual `Hash`/`Eq` impls at value.rs:747/795/865, and `Sid`/`Option<String>` are `Hash`+`Eq`. Note it also makes dedup collapse cross-type-numeric-equal values that share a datatype, which is arguably more correct.)*

fluree-db-shacl/src/compile.rs
```diff
@@ -252,17 +281,104 @@ impl ShapeCompiler {
...
+                        IndexType::Opst,
+                        RangeTest::Eq,
+                        RangeMatch::predicate_object(
+                            rdf_type.clone(),
+                            FlakeValue::Ref(class_class),
+                        ),
+                    )
+                    .await?;
+                class_typed.extend(flakes.iter().map(|f| f.s.clone()));
+            }
```
aaj3f: The implicit-class-target discovery issues two `Opst` bound-object scans (`rdf:type → rdfs:Class`, `rdf:type → owl:Class`) **per input graph on every compile** (i.e. every transaction with shapes), regardless of whether any shape id is actually a class. The result set is bounded by the number of declared classes, so this is cheap in absolute terms, but it is fixed per-transaction overhead that didn't exist before. Acceptable as-is; noting it so it is a conscious cost. If it ever shows up in profiles, gate it on `!self.shapes.is_empty()` (already implied) or restrict to shape ids.

aaj3f (approved):
Nice fix and nice path support addition!

fluree-db-shacl/src/path.rs
bplatz: Addressed in 2063e087a9b54aa059a4d27711d2f089209512f1

fluree-db-shacl/src/path.rs
bplatz: Addressed in 2063e087a9b54aa059a4d27711d2f089209512f1

fluree-db-shacl/src/compile.rs
bplatz: Addressed in 2063e087a9b54aa059a4d27711d2f089209512f1

PR #1424, opened 2026-07-03
feat: SHACL validation reports — fluree validate CLI, Rust API, and HTTP endpoint

bplatz:
Adds **static SHACL validation** as a first-class feature: instead of rejecting a transaction the way staging-time enforcement does, you can now ask "is my existing data clean?" and get a full W3C-shaped validation report — from the CLI, the Rust API, or over HTTP. Built on top of the property-paths work in the base branch.

## `fluree validate` (CLI)

```bash
fluree validate mydb                          # ledger vs its attached shapes
fluree validate mydb --shacl proposed.ttl     # trial ad-hoc shapes (replace attached)
fluree validate mydb --shacl-graph <iri>      # shapes in a named graph
fluree validate data.ttl --shacl shapes.ttl   # standalone file, ephemeral in-memory ledger
```

- **File mode** needs no `.fluree` at all — data loads into an ephemeral memory ledger (staging-time enforcement disabled via the config graph, so files that embed their own shapes report violations instead of failing to load). This is the pre-flight companion to bulk import, which deliberately never runs SHACL.
- Ad-hoc shapes **replace** attached shapes by default ("does this data conform to *these* rules?"); `--include-attached` unions them.
- Output: `table` (human), `--format jsonld|turtle` for the W3C `sh:ValidationReport`.
- CI-friendly exit codes: `0` conforms, `1` findings at/above `--fail-on violation|warning|info`, `2` usage.

## Rust API — `fluree_db_api::validate`

The shared core behind both surfaces: `Fluree::validate_ledger(alias, &ValidateOptions)` / `validate_view(...)` with `ShapesSource::Attached | Graph(iri) | InlineJsonLd | InlineTurtle`, returning a `ValidateReport` with IRI-resolved results and `to_jsonld()` / `to_turtle()` serializers. Validation always runs over the query-visible view (snapshot + novelty); graph IRIs resolve through `graph_registry`, matching `f:shapesSource`. Inline shapes ride the non-persisting inline-shapes bundle, compiled against a detached genesis snapshot so replace semantics never leak the ledger's own shapes — while their value-set facts (e.g. `ex:CA a ex:State` shipped with a `sh:class ex:State` constraint) still reach membership checks via a same-term-space membership source.

## HTTP — `GET|POST /validate/{ledger…}`

Follows the read-endpoint conventions (peer-mode forwarding, bearer `can_read` gate). POST body selects data graph + shapes source; `Accept` negotiates a JSON summary envelope (default), `application/ld+json`, or `text/turtle`. Non-conformance is a `200` — the report is the answer.

## Report core

Validation results now carry everything a standard report needs: `sh:sourceConstraintComponent` IRIs on every result, full RDF term fidelity for `sh:value` (language tags, non-native datatypes as `{"@value", "@type"}`), and focus nodes that can be literals (`FocusNode` enum) for literal `sh:targetNode` targets.

## W3C compliance harness

New `testsuite-shacl/` crate mirroring `testsuite-sparql/` conventions: workspace-excluded, the official `data-shapes` suite as a git submodule, `make count / summary / test-cat / test-one / failures` targets. Each test runs through the same `fluree_db_api::validate` core as the CLI. Comparison semantics, deliberate leniency rules, and the remaining-failure taxonomy are documented in `docs/contributing/shacl-compliance.md`; current status and follow-ups will be tracked in a separate issue.

## Engine conformance fixes (driven by the suite)

- Standalone targeted property shapes (`ex:S a sh:PropertyShape ; sh:path … ; sh:targetNode …` with no wrapping node shape) now compile and validate — previously a silent no-op.
- Literal `sh:targetNode` / literal `sh:targetObjectsOf` objects validate as focus nodes.
- Class-target focus discovery unions a live `rdfs:subClassOf` walk with the indexed hierarchy (novelty-only subclass relations were invisible to targets).
- Node-shape constraint results carry `sh:value` = focus node; `sh:and` produces one result per value node.
- Shapes typed `sh:NodeShape` with only value constraints (e.g. implicit-class-target shapes carrying just `sh:in`) now register during compile.

## ⚠️ Behavior changes to note

1. **`sh:closed` no longer implicitly ignores `rdf:type`** — the spec never grants that; shapes targeting typed instances must declare `sh:ignoredProperties (rdf:type)` (as the spec's own examples do). Existing shapes relying on the auto-ignore will start flagging `rdf:type`.
2. **`sh:conforms` is spec-semantics**: false when the report has *any* results, warnings/infos included. Transaction enforcement is unchanged — it gates on violations explicitly.
3. **JSON-LD embedded node objects with `@id` + properties now assert their triples** (previously silently dropped as bare references). This fixes data loss in JSON-LD bulk import and inline shape/ontology documents.
4. Turtle parser accepts bare `[ … ] .` blank-node-property-list statements (valid Turtle, previously rejected).

## Docs

`docs/cli/validate.md`, `/validate` section in `docs/api/endpoints.md`, `docs/contributing/shacl-compliance.md`, and cookbook updates (validation reports section, strict `sh:closed` guidance).


fluree-db-api/src/validate.rs
```diff
@@ -0,0 +1,695 @@
...
+    }
+    if let Some(dt) = datatype {
+        let dt_iri = resolve(dt);
+        // A String carrying the `@id` datatype is a stringified IRI
+        // (STR() semantics in the string facets) — report the IRI node.
+        if dt_iri == "@id" || &*dt.name == "id" {
+            if let FlakeValue::String(s) = value {
+                return json!({"@id": s});
+            }
+        }
```
aaj3f, line 617: Don't know if we'd ever see this, but the `@id`-datatype heuristic `dt_iri == "@id" || &*dt.name == "id"` will misfire for any *real* datatype whose local name happens to be `id` (e.g. a literal `"foo"^^<http://example.org/id>`): `dt_iri` won't equal `"@id"`, but `dt.name == "id"` is true, so the string literal is rendered as `{"@id":"foo"}` — a plain literal reported as an IRI node in `sh:value`. Report-display only (no effect on the conformance decision), but wrong. Prefer keying off the full IRI (or the datatype's namespace code) rather than the bare local name:
  ```rust
  // drop the `|| &*dt.name == "id"` fallback, or gate it on the id namespace:
  if dt_iri == "@id" {
      if let FlakeValue::String(s) = value {
          return json!({"@id": s});
      }
  }
  ```

fluree-db-transact/src/stage.rs
aaj3f: `validate_staged_nodes` still computes `conforms = all_results.iter().all(|r| r.severity != Violation)` (violation-based), while `fluree_db_shacl::ShaclEngine::validate_staged` was changed in this PR to `conforms = all_results.is_empty()` (spec semantics). Two functions with near-identical names now stamp `ValidationReport.conforms` with two different meanings. It's benign today (every enforcement gate here reads `violation_count()`, not `.conforms`), but it's a latent trap for the next caller who trusts `validate_staged_nodes(...).conforms` to mean what the crate-level report means. Recommend aligning to `all_results.is_empty()` for one consistent definition of `conforms` across crates, or a comment explaining the divergence.

fluree-db-shacl/src/validate.rs
```diff
@@ -532,11 +595,36 @@ async fn get_focus_nodes(
...
+                            ),
+                        )
+                        .await?;
+                    for flake in &sub_flakes {
+                        if visited.insert(flake.s.clone()) {
+                            classes_to_query.push(flake.s.clone());
+                            queue.push_back(flake.s.clone());
+                        }
+                    }
+                }
```
aaj3f, line 639: The live `rdfs:subClassOf` descendant BFS added to `get_focus_nodes` issues one `Opst` range query per discovered class and re-runs in full for *every shape* that targets the same class, on top of the already-consulted indexed hierarchy. Additionally, for a `sh:targetObjectsOf` shape the target predicate is range-scanned twice — once here (line 665, Ref objects) and once in the literal-target loop (line 388, literal objects). This path is only reached by the explicit `validate` endpoint/CLI (`validate_all_with_membership`), **not** the transaction-commit hot path, so it is not a hot-path regression and I'll 100% defer to you if it's even worth fixing — but for large ontologies it is avoidable repeated I/O. Consider memoizing the subclass closure once per validate call (keyed by class Sid) and folding the ObjectsOf literal/Ref discovery into a single predicate scan.

fluree-db-api/src/validate.rs
```diff
@@ -0,0 +1,695 @@
...
+                    } else {
+                        '-'
+                    }
+                })
+                .collect();
+            format!("_:{clean}")
+        }
+        None => format!("<{iri_or_bnode}>"),
+    }
+}
```
aaj3f, line 238: Turtle IRI serialization is not escaped. `turtle_term` sanitizes blank-node labels (the `_:` branch) but renders IRIs as the raw `format!("<{iri_or_bnode}>")`. An IRI carrying a Turtle-IRIREF-illegal character (space, `<`, `>`, `"`, `{`, `}`, `|`, backtick, `\`) would emit non-parseable Turtle for the `text/turtle` response / `to_turtle()`. Low likelihood (SIDs decode to IRIs from normally well-formed parsed input) and the JSON-LD path is unaffected, but it is the same class of bug the author already guarded against for blank nodes — worth a percent-encode or guard on the IRI branch. Not merge-blocking.

aaj3f (approved):
Excited for this PR's work towards a SHACL validation mechanism

fluree-db-api/src/validate.rs
bplatz: Addressed in b7e0b296c62a491fde13700b87d784a3bb30d66e

fluree-db-transact/src/stage.rs
bplatz: Addressed in b7e0b296c62a491fde13700b87d784a3bb30d66e

fluree-db-api/src/validate.rs
bplatz: Addressed in b7e0b296c62a491fde13700b87d784a3bb30d66e

fluree-db-shacl/src/validate.rs
bplatz: Addressed the repeated subclass BFS + instance scans in 40e7e6fa1981937e4bd5ad3786f8fb977565a786: `get_focus_nodes` now memoizes the per-class focus-node set for the duration of a validate call (keyed by the target class Sid — `db` and hierarchy are constant across the shape loop), so multiple shapes targeting the same class compute it once. Added `it_validate_report.rs` (first integration coverage for the validate endpoint) asserting subclass expansion + both shapes firing on the shared memoized focus set.

I deliberately left the `sh:targetObjectsOf` double-scan alone: the Ref-object and literal-object discovery live in separate scopes (`get_focus_nodes` vs the literal-target loop), so folding them into one scan means restructuring that boundary for a win that's just one duplicate scan per ObjectsOf shape — not amplified by hierarchy depth like the BFS was. Poor risk-to-reward on this cold path, so I'm consciously deferring it.

Issue #1405, opened 2026-06-30
Cross-snapshot property-path BFS: lift the multi-graph union-path guard (follow-up to #1425)


Jackamus29:
Property-path / graph-crawl traversal and multi-ledger datasets are each documented, first-class
features, but they **cannot be composed**: a transitive path (`p+` / `p*`) cannot run over a
multi-ledger dataset. The natural single-query forms all fail — one is rejected by a guard, and
the suggested `GRAPH` workaround crashes on indexed data and silently drops rows on cross-graph
joins. Net effect: *"traverse a hierarchy in one ledger and join instances in another"* — graph
crawl across ledgers — can't be expressed in a single query.

Re-verified on **v4.1.1** (running server + a runnable, scrubbed regression test). **Distinct
from #1295** (that's cross-ledger *explicit projection* dropping foreign-namespace predicates — a
formatter bug; this is *traversal / joins* in the execution layer).

## Why this should compose (it's documented)

- **Property paths / graph crawl** are documented first-class (`query/graph-crawl.md`,
  `query/sparql.md`). The docs' own **"Not Yet Supported"** list for property paths contains only
  `p?` and `!p` — the multi-graph restriction below is **not listed anywhere**.
- **Multi-ledger datasets + cross-ledger joins** are documented first-class (`query/datasets.md`:
  *"queries across multiple graphs and ledgers simultaneously … powerful data integration"*; the
  "Multi-Ledger Datasets" and "Cross-Ledger Joins" sections).

So a developer reading the docs reasonably expects transitive traversal across ledgers to work.

## Motivating use case

A node in ledger **A** references a node in ledger **B**, and **A also carries the edge that
bridges into B's hierarchy** — e.g. a user-defined term `broader`→ a canonical term in a shared
taxonomy ledger. Answering "what is reachable via `broader+`, joined to instances?" needs a path
whose hops **span A and B**, plus a join back to A's instances. There is no single-query way to
express this today.

## The three failures (verified v4.1.1)

| | Shape | Result | Class |
|---|---|---|---|
| **1** | property path over a multi-graph `from` (union) | rejected — `Property paths over multi-graph datasets are not supported; use GRAPH to select a single graph` | guard (undocumented limitation) |
| **2** | `GRAPH`-scoped path over an **indexed** multi-ledger dataset | internal error — `EncodedSid/EncodedPid reached stamp_provenance` | **bug** |
[...]

PR #1425, opened 2026-07-03
fix(query): cross-ledger GRAPH queries over indexed data (#1405 bugs 2+3)

Jackamus29:
Partially addresses #1405 — makes the documented `GRAPH`-scoped workaround actually work over **indexed, multi-ledger, divergent-namespace** data. Fixes failures **2** and **3** from the issue. Failure **1** (lifting the union-path guard via cross-snapshot BFS) is a committed follow-up.

## Failures fixed

**Bug 2 — indexed GRAPH path → internal error.** A `GRAPH`-scoped path over an indexed multi-ledger dataset hit `EncodedSid/EncodedPid reached stamp_provenance`. `DatasetOperator` computed `multi_ledger` from the *active* graphs only, so a single default graph (alongside named graphs from other ledgers) was treated as single-ledger — the binary store stayed on, its scans emitted late `EncodedSid`, and those seeded a GRAPH block / crossed a boundary into provenance stamping, which can't decode them. **Fix:** treat the scan as multi-ledger when the whole dataset spans ledgers, forcing full `Binding::Sid` materialization (which stamps to `IriMatch`).

**Bug 3 — divergent-namespace cross-graph join/path → silent `[]`.** The materialization fix resolves the join/seed cases (bound keys now cross as `IriMatch` and re-encode). Property paths additionally matched pattern **constants** (endpoints) and **predicates** against the primary snapshot's codes rather than the per-`GRAPH` graph's. **Fix:** re-encode path predicate + constant-endpoint SIDs into the active graph's dict (`reencode_pred` + the `Ref::Sid` arm of `resolve_sid`), reusing `binary_scan`'s decode-primary/encode-target idiom.

## Scope / safety
- Single-ledger and single-graph queries unchanged (re-encode round-trips to the same SID; materialization only engages for multi-ledger datasets).
- The union-path guard (failure 1) stays in place — its `q2` characterization test asserts it still fires.
- Tradeoff: multi-ledger datasets now eagerly materialize even a single default-graph scan (a scoped cost), which is the price of correct cross-boundary provenance — consistent with the existing multi-ledger disable.

## Tests — `tests/it_multi_graph_property_path.rs`
The `q1`–`q3d` repro (adopted from the issue) plus a usage-pattern matrix, all **indexed** with **divergent** namespace codes and exact-value assertions:

| test | pins |
|---|---|
| q3a / q3c | bug 2 (indexed GRAPH path) / bug 3 (divergent join) |
| A1 | object-position cross-graph join |
| A2 / A3 | multi-value completeness / precision (include match, exclude non-match) |
| A4 | `p+` strict path + join (bug 2 + 3 combined) |
| A5 | three-ledger chained join (re-encode composes) |
| A6 | single-ledger GRAPH path unchanged (regression) |
| A7 | `FILTER EXISTS` across a GRAPH boundary (covered by the root-cause fix) |
| A8 | unbounded closure with a divergent-code predicate |

Verification: 14/14 in the new suite; `fluree-db-query` 1170+; `grp_query` 299, `grp_query_sparql` 258, `grp_misc` 237; fmt + clippy clean.

## Follow-up (separate PR)
Failure 1 — lift the guard with a cross-snapshot BFS so a property path runs directly over a multi-graph union (`p+` across ledgers), honoring the "query your datasets as if they were one materialized dataset" model.


fluree-db-query/src/property_path.rs
```diff
@@ -57,6 +57,20 @@ fn is_reserved_edge_predicate(p: &Sid) -> bool {
...
+/// time, but a path runs against a per-`GRAPH` snapshot that may code the same
+/// IRI differently; without this the traversal reads the wrong SID and finds no
+/// edges. Single-graph round-trips to the same SID.
+#[inline]
+fn reencode_pred(ctx: &ExecutionContext<'_>, db: &fluree_db_core::LedgerSnapshot, p: &Sid) -> Sid {
+    ctx.original_snapshot
+        .decode_sid(p)
+        .and_then(|iri| db.encode_iri(&iri))
+        .unwrap_or_else(|| p.clone())
+}
```
aaj3f: Not blocking but this fn seems identical to the match arm in the same file, lines 1037-1041. Also, the commit that introduced this said it reuses `binary_scan.rs:1616`'s "decode-primary/encode-target", but it actually just copies them. Ideally we might hoist one of the fns as `pub(crate) fn reencode_sid(ctx, target, sid)` into a common module and then calling from both operators for the sake of hygiene and as a guard against regression/fragility

fluree-db-api/tests/it_multi_graph_property_path.rs
aaj3f, line 1: The property-path portions are SPARQL-exempt from JSON-LD parity, but the `dataset_operator.rs` materialization fix is shared-IR and would benefit any dataset query, including a JSON-LD cross-ledger cross-graph join. A single parallel JSON-LD analog for the non-property-path cross-graph join would strengthen the shared-IR parity story

aaj3f (approved):
Looks good and makes sense -- glad to have those two bugs fixed.

PR #1428, opened 2026-07-03
feat: Remote mounts, serving tiers, and peer-mode CLI

bplatz:
Serve ledgers to downstream consumers in two tiers and let a consumer mount a remote Fluree's ledgers as read-only, locally-queryable sources. Design doc: `docs/design/remote-mounts.md`; user guide: `docs/guides/sharing-data.md`.

## The model

- **Query serving** — the origin executes queries (its compute, row-level policy per identity). The only tier with fine-grained permissioning.
- **Block serving** — the origin serves canonical, CID-verified index content and the consumer executes queries locally (their compute, Iceberg-style). Strictly all-or-nothing per (token, ledger); fine-grained access stays on the query tier.

## What's included

**Raw block serving.** `ProxyStorage` gains explicit read modes: `Raw` fetches canonical CAS bytes via `GET /storage/objects/{cid}` with client-side CID verification; `Filtered` keeps the FLKB negotiation. Peer proxy mode now uses `Raw` — which is what makes binary-indexed ledgers actually readable over the proxy (the FLKB tier has no leaf decoder on the read path).

**Per-ledger serving posture.** New `f:servingDefaults` setting group (`f:serveQuery` / `f:serveBlocks` / `f:publicVisibility`) in the ledger config graph, enforced on transaction-role servers only (query gate 403; blocks gate 404 on `/storage/block`, `/storage/objects`, `/commits`, `/pack`) and advertised per-caller on NS record responses plus a coarse block in `/.well-known/fluree.json`.

**Remote mounts.** `FlureeBuilder::with_remote_mount` composes a `CompositeNameService` (prefix-routed reads with record localization — remote `inventory:main` appears as `acme/inventory:main`; writes to mounts rejected) with a new `StorageBackend::Routed` variant (namespace-prefix store selection at the single `content_store` seam). Mounted ledgers keep full native semantics including mixed local+mounted dataset queries. `ProxyStorage`/`ProxyNameService` moved to `fluree-db-nameservice-sync` (server re-exports preserve the old paths).

**CLI peer mode.** `fluree track add --mode peer` runs queries locally over blocks fetched on demand, CID-verified and cached in a persistent per-remote disk cache; writes still forward over HTTP. Plus `fluree remote ledgers` (the token's auth-filtered catalog with serving tiers) and `fluree cache status|clear`.

**HTTP Range.** Single-range requests on `/storage/objects` (206 + `Content-Range`, full-object CID verification before slicing) and native ranged reads in `ProxyStorage`.

**Vended S3 credentials.** For S3-backed origins, `GET /storage/credentials?ledger=` mints STS `AssumeRole` grants narrowed by session policy to the ledger's name-level prefix (covers all branches + the `@shared` dict namespace). Consumers auto-refresh grants inside an expiry margin; the CLI probes and prefers S3-direct automatically, falling back to proxied reads on 404. Single-bucket S3 only in this iteration; short TTLs since grants outlive revocation until expiry.

**Fix along the way:** `@shared` dict-blob addresses carry only the ledger *name*, so proxy clients derive a default-branch alias — the block/object endpoints now branch-resolve dict-blob requests (previously peers tracking non-main branches 404'd on dictionary fetches). The legacy per-branch dict layout also parses now.

## Docs

New: `design/remote-mounts.md`, `guides/sharing-data.md`, `cli/cache.md`. Updated: `operations/query-peers.md` (two read tiers; corrected `/storage/block` leaf semantics), `ledger-config/setting-groups.md`, `design/auth-contract.md`, `cli/track.md`, `cli/remote.md`, `api/endpoints.md`, and `cli/server-integration.md` (the contract an embedding server must expose for peer-mode consumers, including the dict-blob branch-resolution and 404-fallback requirements).

## Testing

- `proxy_integration.rs`: 28 tests including end-to-end peer query over HTTP against an indexed ledger, remote-mount mixed-dataset query + write rejection, serving-posture gate/advertisement flips, raw-mode byte identity + ranged reads, vended-credentials 404 gate.
- LocalStack round trip (`it_vended_credentials_testcontainers`): mint against real STS → build reader from the grant → read a CAS object through the fluree address layer.
- Wiremock coverage for grant fetch 404-fallback and refresh-on-expiry; unit tests for the session-policy shape, mount routing/localization, and CLI target resolution (peer → local queries, downgrade to HTTP for writes).
- CI parity: clippy (all features/targets, `-D warnings`) and `cargo check --workspace --all-features --all-targets` clean.

## Notes for review

- Base is `feature/rdfs-enforcement-entailment` (this branch was cut from its tip).
- Serving posture binds only the origin's serving surface by design — a consumer holding the blocks always queries its own copy; rationale in the design doc.
- LocalStack community doesn't enforce IAM session policies, so prefix-scoping enforcement rests on the policy-shape unit tests + AWS semantics; worth one manual verification against real AWS before production reliance.
- Follow-ups deliberately out of scope: server-level mount config flags, `--mode auto` negotiation, the `f:publicVisibility` anonymous tier, split-bucket vend grants.


fluree-db-server/src/routes/query.rs
```diff
@@ -3400,9 +3400,21 @@ pub(crate) async fn load_ledger_for_query(
...
 
-    // In transaction mode, just return the cached state
+    // In transaction mode, return the cached state after the serving gate:
+    // the origin honors the ledger's f:serveQuery posture. Peer-role servers
+    // skip the gate — they serve queries from their own replicated copy by
+    // design (see routes/serving.rs).
     if state.config.server_role != ServerRole::Peer {
-        return Ok(handle.snapshot().await.to_ledger_state());
+        let ledger_state = handle.snapshot().await.to_ledger_state();
+        let serving = crate::routes::serving::effective_serving_from_state(&ledger_state).await?;
```
aaj3f, line 3409: The query serving gate calls `effective_serving_from_state` → `config_resolver::resolve_ledger_config`, which resolves the **entire** ledger config graph (policy, shacl, reasoning, datalog, transact, full_text, serving, graph_overrides — 8 group reads plus a `find_instances_of_type` scan) on **every query** to a transaction-role server, purely to read `f:serveQuery`. It *duplicates* the config resolution the db-view build already performs per query (`fluree-db-api/src/view/fluree_ext.rs:120`). **Bounding factors** (why this is minor, not major): `resolve_ledger_config` early-returns cheaply when the config graph is empty (guard at `config_resolver.rs:67-84`), so unconfigured ledgers pay almost nothing; the gate runs only on transaction/**origin** servers (peers skip it); and the in-memory config graph is dwarfed by the query it precedes. Still worth fixing: resolve only the serving group (see the serving.rs suggestion), thread the already-resolved config into the gate, or memoize the posture per `(ledger_id, t)`.
  ```rust
  // Prefer a targeted resolver (see serving.rs suggestion) so the gate reads
  // only f:servingDefaults instead of the full config graph, or reuse the
  // ResolvedConfig the view build already computes for this same snapshot/t.
  let serving = crate::routes::serving::effective_serving_from_state(&ledger_state).await?;
  ```

fluree-db-server/src/routes/storage_proxy.rs
```diff
@@ -528,8 +703,22 @@ pub async fn get_object_by_cid(
...
+
+    // 3b. Authorize: principal must have access to the resolved ledger.
+    if !principal.is_authorized_for_ledger(&effective_ledger) {
+        return Err(ServerError::not_found("Object not found"));
+    }
+
+    // 3c. Serving gate: the ledger's f:serveBlocks posture must allow raw
+    //     content serving.
+    let serving =
+        crate::routes::serving::effective_serving(&state.fluree, &effective_ledger).await?;
```
aaj3f: `get_object_by_cid` is the peer-mode raw-read hot path: a cold peer sync fetches *every* index leaf, branch, root, dict blob, and commit through this endpoint. This PR adds, per object, both a nameservice `lookup` (via `resolve_block_ledger`, line 708) **and** a full `resolve_ledger_config` (via `effective_serving`, line 720). Previously this handler only checked token scope and read bytes — zero ledger loads, zero config resolution. **Bounding factors** (why this is minor, not major): the ledger handle is *cached* (no per-object reload — `effective_serving` reuses `ledger_cached`); `resolve_ledger_config` early-returns for unconfigured ledgers; and per-request cost is dominated by JWT verification + storage IO + full-object SHA-256, so the aggregate config-resolution overhead on a cold sync is negligible next to the byte transfer. The `ns.lookup` namespace guard is a **legitimate correctness addition** (namespace guard + dict-blob branch resolution), not waste — only the repeated full-config resolution is avoidable. Memoize `EffectiveServing` per `(effective_ledger, t)` on `AppState` and/or resolve only the serving group.
  ```rust
  // e.g. a DashMap<(String, i64), EffectiveServing> on AppState keyed by
  // (effective_ledger, snapshot.t), populated on miss — a config change bumps t.
  let serving = state.effective_serving_cached(&effective_ledger).await?;
  ```

fluree-db-server/src/routes/serving.rs
```diff
@@ -0,0 +1,85 @@
...
+    state: &LedgerState,
+) -> Result<EffectiveServing, ServerError> {
+    let overlay: &dyn OverlayProvider = &*state.novelty;
+    let config = config_resolver::resolve_ledger_config(&state.snapshot, overlay, state.t())
+        .await
+        .map_err(|e| ServerError::internal(format!("Serving config resolution failed: {e}")))?;
+    Ok(EffectiveServing::from_config(
+        config.as_ref().and_then(|c| c.serving.as_ref()),
+    ))
+}
```
aaj3f, line 67: `effective_serving_from_state` resolves the whole `LedgerConfig` (all 8 setting groups) but uses only `config.serving`. Every gate check (query, object, block, commits, pack, credentials, NS advertisement) pays for 7 unused group reads. Add a targeted resolver that reads only `f:servingDefaults`.
  ```rust
  // In config_resolver: expose a resolve_serving_only(snapshot, overlay, to_t)
  // that runs find_instances_of_type + read_serving_defaults and nothing else.
  let serving = config_resolver::resolve_serving_only(&state.snapshot, overlay, state.t()).await?;
  Ok(EffectiveServing::from_config(serving.as_ref()))
  ```

aaj3f (approved):
This is a nice feature addition -- glad to have it and eager to use it

fluree-db-server/src/routes/query.rs
bplatz: Addressed in d1034aa0391a4747162e429b3942c4fc2203ed4f

fluree-db-server/src/routes/storage_proxy.rs
bplatz: Addressed in d1034aa0391a4747162e429b3942c4fc2203ed4f

fluree-db-server/src/routes/serving.rs
bplatz: Addressed in d1034aa0391a4747162e429b3942c4fc2203ed4f

PR #1432, opened 2026-07-04
feat: stable _:fdb blank-node ids addressable in queries and updates

bplatz:
## What

Labels with the reserved `_:fdb-` prefix (the skolem ids Fluree mints for every blank node and returns from queries) now denote the **existing stored node** in queries and transactions, instead of silently minting a fresh node. This makes blank-node-rooted structures — OWL restrictions, address objects, RDF lists — editable in place: query the node's id, then use it as an ordinary `@id` to insert/delete individual triples, without retracting and re-asserting the whole subtree.

The read side already resolved `_:fdb-...` to the stored Sid (`canonical_split` routes any `_:` string to the blank-node namespace); this closes the write side and SPARQL:

- **JSON-LD transactions** — `@id: "_:fdb-..."` in insert/update templates and VALUES addresses the stored node
- **SPARQL SELECT** — `_:fdb-...` in WHERE patterns lowers to a constant; ordinary labels stay non-distinguished variables per spec
- **SPARQL UPDATE** — INSERT/DELETE templates, `DELETE DATA`/`INSERT DATA`, and `DELETE WHERE` all resolve stable ids (a bnode in a DELETE template was previously skolemized fresh, i.e. a silent no-op)
- **Turtle/TriG** — `FlakeSink::term_blank` and the upsert named-graph path pass stable labels through
- **Edge annotations** — delete-by-@id accepts stable ids (they are now addressable); client-authored blank labels still rejected

Ordinary labels (`_:b0`) keep standard RDF semantics: fresh node per transaction on write, existential variable in SPARQL WHERE. Clients cannot collide with the reserved space because skolemization wraps client labels with a txn id before prefixing `fdb-`. All checks run at parse/lowering time inside existing `_:`-prefixed branches — zero cost for transactions not using stable ids.

Spec framing: this is RDF 1.1 §3.5 skolemization kept in blank-node syntax — the same extension Virtuoso's `nodeID://` refs and Jena's `<_:label>` syntax provide. SPARQL §3.1.3's ban on blanks in DELETE templates exists because such labels could only denote fresh nodes; the reserved prefix removes that premise.

## Import fixes (found while auditing blank-node scoping)

- **TriG named-graph blank nodes scoped to their document**: named-graph blanks were minted from the bare label (`fdb-b0`) — colliding globally across every document imported into a ledger while *diverging* from the same document's default graph (`fdb-{ledger}-{t}-b0`). They now share the default graph's skolem key, so a label unifies across the default graph and GRAPH blocks of one document and stays distinct across documents.
- **Serial-import namespace codes allocated through the shared allocator**: the serial TriG/N-Quads commit path allocated namespace codes only in `state.ns_registry`, invisible to the spool's mid-parse code→prefix lookups. The silent empty-prefix fallback permanently wrote suffix-only strings (`name` instead of `http://schema.org/name`) into the index's predicate dict — bound-predicate patterns matched nothing on freshly imported TriG ledgers. Serial commits now mirror `parse_chunk`'s WorkerCache-over-shared-allocator discipline, and the fallback `debug_assert`s (warns in release) so any recurrence fails tests loudly instead of corrupting silently.

## Testing

- `it_stable_blank_nodes.rs` — 7 tests: in-place OWL-restriction edits via JSON-LD and SPARQL, node identity stable across edits, fresh-mint semantics preserved for ordinary labels, constant-vs-variable WHERE behavior
- `it_import.rs` — `import_trig_blank_label_document_scoped` (fails pre-fix) and `import_trig_bound_predicate_queryable` (fails pre-fix); `should_panic` unit test pins the spool prefix-miss detection
- W3C SPARQL eval: 283/327 before and after (zero regression, verified by stashing the changes and re-running)
- Full runs green: fluree-db-transact lib (277), grp_import (82), grp_transact (168), grp_query_sparql (258), edge annotations (78); clippy all-features clean

## Docs

- New "Editing Blank-Node Structures (Stable `_:fdb-` Ids)" section in `docs/transactions/update-where-delete-insert.md`, with pointers from `insert.md` and `concepts/iri-and-context.md`


aaj3f (approved):
Makes sense and looks good! 👍

PR #1434, opened 2026-07-05
feat: Aggregate netted branch diff in merge preview (include_changes)

bplatz:
## Summary

Adds an opt-in aggregate change set to `Fluree::merge_preview` — a git-diff-style rollup of the net facts a merge would apply, for merge-request review UIs (Fluree Solo). A branch with 40 commits that ultimately changes 12 facts now reviews as 12 facts instead of 40 commit-by-commit event logs.

Closes the `include_changes` feature request.

## What's included

**Netting engine** (`fluree-db-novelty`)
- `NetChangeAccumulator`: folds the source chain per **full fact identity** `(graph, s, p, o, dt, lang, list-index)` — deliberately not `Flake`'s own `Eq`/`Hash`, which ignores graph, and stricter than `(s,p,o,dt,g)`, which would collide language-tagged strings and list positions.
- `compute_delta_keys_and_changes`: one commit-chain walk serves both conflict keys and the change set, so requesting conflicts + changes never replays the source side twice.

**API** (`fluree-db-api`)
- `MergePreviewOpts` gains `include_changes`, `max_changes`, `changes_after_subject`.
- `MergePreview.changes` (`ChangeSummary`) carries exact `assert_count` / `retract_count` / `subject_count` (never truncated), subject-grouped resolved-flake entries ordered by full IRI, `truncated`, and `next_cursor`.

**HTTP** (`fluree-db-server`)
- `GET /merge-preview` gains `include_changes`, `max_changes` (hard-capped at 5,000), and `changes_after_subject` query params.

**CLI** (`fluree-db-cli`)
- `branch diff --changes / --max-changes / --stat / --changes-after` in local, tracked, and remote modes.

## Semantics

- **Net commit effect**: per fact, the net op is the newest in-range op; a fact survives only when its oldest and newest in-range ops agree. Create-then-delete and delete-then-restore churn never appears. A re-assert of a pre-range value nets as an assert (the merge does apply it).
- **Strategy-independent**: the change set is the raw source-vs-ancestor delta, before conflict resolution; conflicting keys resolve per the existing `conflicts.details`.
- **Pagination**: subjects ordered by full IRI (the cursor contract). `max_changes` counts flakes but cuts at subject boundaries — a subject's diff is never split across pages; the first subject of a page is returned whole even if it alone exceeds the cap, so pagination always progresses. `max_changes=0` is a cheap diff-stats mode (exact counts, no payload, no source-state load).
- **Cost**: same profile as the existing conflict computation — O(flakes since ancestor), one commit load per commit in the source divergence; caps bound response size, not compute. Each pagination page re-pays the replay (documented).

## Docs

- `docs/api/endpoints.md`: params, `changes` response object, examples.
- `docs/cli/server-integration.md`: Merge Preview Contract rule 11 (netting/pagination semantics for alternate server implementations).
- `docs/getting-started/rust-api.md`: `include_changes` example (also fixed two pre-existing non-compiling `MergePreviewOpts` literals).

## Testing

- 10 netting unit tests (cancellation both directions, same-commit pairs, graph/lang/list-position key separation, update both-sides, newest-representative).
- 10 API integration tests: fast-forward diff, churn nets to empty, update shows both sides, pure retract of ancestor fact, 3-page pagination round-trip (exact counts per page, no overlap, IRI order), stats-only mode, changes alongside conflict details (shared walk), cursor-without-include_changes rejection, multi-value predicate independence.
- 1 server HTTP test covering param passthrough, stats mode, and the 400 validation.
- Existing suites: 71 novelty, 34 merge-preview + 50 merge, 226 CLI — all passing; clippy `--all-features --all-targets -D warnings` clean on all four touched crates.

### DIFF
--- a/fluree-db-api/src/merge_preview.rs
+++ b/fluree-db-api/src/merge_preview.rs
@@ -20,9 +20,11 @@ use fluree_db_core::{
     ContentId, ContentStore, Flake,
 };
 use fluree_db_ledger::LedgerState;
-use fluree_db_novelty::compute_delta_keys;
+use fluree_db_novelty::{compute_delta_keys, compute_delta_keys_and_changes};
 use futures::{stream, StreamExt, TryStreamExt};
+use rustc_hash::FxHashSet;
 use serde::Serialize;
+use std::collections::BTreeMap;
 use std::sync::Arc;
 use tracing::Instrument;
 
@@ -31,6 +33,10 @@ pub const DEFAULT_MAX_COMMITS: usize = 500;
 
 /// Default cap on conflict keys returned in [`ConflictSummary::keys`].
 pub const DEFAULT_MAX_CONFLICT_KEYS: usize = 200;
+
+/// Default cap on change entries returned in [`ChangeSummary::entries`],
+/// counted in flakes.
+pub const DEFAULT_MAX_CHANGES: usize = 500;
 
 /// Knobs for [`crate::Fluree::merge_preview`].
 ///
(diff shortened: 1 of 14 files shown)


fluree-db-api/src/merge_preview.rs
```diff
@@ -457,10 +640,107 @@ impl crate::Fluree {
...
+        });
+    };
+
+    // Group by full subject IRI; BTreeMap gives the deterministic subject
+    // order that pagination cursors rely on.
+    let mut by_subject: BTreeMap<String, Vec<Flake>> = BTreeMap::new();
+    for f in net_flakes {
+        let subject = compactor.decode_sid(&f.s)?;
+        by_subject.entry(subject).or_default().push(f);
+    }
```
aaj3f: `compactor.decode_sid(&f.s)` runs **once per flake** during grouping, so a subject with K changed facts decodes its IRI K times, and every flake beyond `max_changes` is still decoded before being dropped (O(net_flakes) `format!` allocations). Consider grouping by `&Sid` (cheap, no alloc) into an `FxHashMap`, then decoding each *distinct* subject once and sorting by decoded IRI. Bounded, off the query hot path — a micro-opt, though, so take it or leave it

aaj3f (approved):
Neat addition!

fluree-db-api/src/merge_preview.rs
bplatz: Addressed in 1a5fd789fc3d0bb23f1a7419c4727dc746121293

PR #1436, opened 2026-07-06
[FIX] SPARQL Parsing of sub-`SELECT` as a Set-Operation Operand

### LINKED ISSUE #1435: SPARQL: sub-SELECT mis-parsed as a set-operation operand (UNION dropped; MINUS/OPTIONAL projection lost)
## Summary

A SPARQL **sub-SELECT (`{ SELECT ... }`) used as an operand of a set/graph operation is mis-parsed**, silently producing wrong results. Two symptoms were reported against `fluree/server:4.1.1` from the `fluree/azure-chat` NL→SPARQL work:

- **fluree/azure-chat#42** — `{ SELECT … } UNION { SELECT … }` returns only one arm (or 0 rows), not the union.
- **fluree/azure-chat#43** — `{ SELECT … } MINUS { SELECT … }` performs no subtraction and returns the full left operand.

Both are the **same root cause** in the SPARQL parser, and the defect is broader than those two shapes. Reproduced on `main` (HEAD `15e2a4b9`), on both novelty and durable-reindexed state.

## Root cause

`Parser::parse_group_graph_pattern` (`fluree-db-sparql/src/parse/query/pattern.rs`) only recognises a `{ SELECT ... }` sub-SELECT **inside its own nested-`{` branch**. Every *other* position that opens a group — `OPTIONAL`, `GRAPH`, `SERVICE`, the right arm of `MINUS`, each arm of `UNION` (via `parse_union_continuation`), `EXISTS`/`NOT EXISTS` (via the free `parse_group_graph_pattern`), and a bare `WHERE { SELECT … }` — consumes the `{` itself and then calls `parse_group_graph_pattern` positioned at the `SELECT`, where the leading `SELECT` is treated as an unexpected token.

Consequences:

- **UNION** — after parsing the left `{ SELECT … }` the parser never checked for a trailing `UNION` (unlike the plain-group branch right below it), so the `UNION` token hit `"UNION must follow a pattern"` and was discarded. The query collapsed into two independent subqueries lowered as a **conjunction**; at execution that is a correlated join on the shared projected variable, whose values are disjoint → only one arm survives (the smaller-cardinality one; ties → the left) or 0 rows. It is **not** "only the last arm," as azure-chat#42 hypothesised.
- **MINUS / OPTIONAL / GRAPH / SERVICE** — the sub-SELECT operand was parsed as a bare BGP with the `(expr AS ?v)` projection **dropped**. For MINUS the right arm then shares no bound variable with the left → nothing is subtracted; for OPTIONAL the projected/aggregate variable is lost.

The IR from lowering confirms it: `{SELECT} UNION {SELECT}` lowered to `[Subquery, Subquery]` (no `Union` node); `{SELECT} MINUS {SELECT}` lowered to `[Subquery, Minus(Triple×4)]` (right arm = raw triples, `Bind(?n)` projection gone).

## Reproduction

Data: 3 `ex:Person` (Alice/Bob/Carol), 2 `ex:Company` (Acme/Globex), 5 `ex:Country`, 2 `ex:Maker` (in France + Japan).

**azure-chat#42 (UNION):**
```sparql
SELECT DISTINCT ?n WHERE {
  { SELECT (?pn AS ?n) WHERE { ?p a ex:Person ; ex:name ?pn } }
  UNION
  { SELECT (?cn AS ?n) WHERE { ?c a ex:Company ; ex:name ?cn } }
}
```
Expected 5 (Alice, Bob, Carol, Acme, Globex); actual **2** (Acme, Globex).

**azure-chat#43 (MINUS):**
```sparql
SELECT DISTINCT ?n WHERE {
  { SELECT (?cn AS ?n) WHERE { ?c a ex:Country ; ex:cname ?cn } }
  MINUS
  { SELECT (?mn AS ?n) WHERE { ?c2 a ex:Country ; ex:cname ?mn . ?m a ex:Maker ; ex:inCountry ?c2 } }
}
```
Expected 3 (Brazil, Canada, Germany); actual **5** (no subtraction).

Broader symptoms (same cause): `{ plain } UNION { SELECT … }` drops the sub-SELECT arm; `OPTIONAL { SELECT (COUNT(…) AS ?c) … }` loses `?c` and duplicates rows.

## Fix

Centralise sub-SELECT detection at the top of `parse_group_graph_pattern` (the opening `{` is already consumed by the caller), so every position that admits a group admits the grammar's `GroupGraphPattern ::= '{' ( SubSelect | GroupGraphPatternSub ) '}'`; and run the trailing-`UNION` check for a sub-SELECT left arm too. The fix is entirely in the parser — fixed queries now compile to the **already-optimised** `UnionOperator` / sub-SELECT-`MINUS` plans (no query-engine change, no hot-path cost).

## Validation

- W3C SPARQL suite (before → after): **syntax-query 81 → 83** (`syntax-SELECTscope1`, a positive `{SELECT} UNION {SELECT}` syntax test, now passes), **subquery 11 → 12** (`sq12`, a bare `WHERE { SELECT … }` in CONSTRUCT, now passes), **negation 12 → 12** (the W3C negation suite has no sub-SELECT-as-MINUS-operand test — new `fluree-db-api` integration tests fill that gap). **Zero regressions.**
- New AST-level parser tests + end-to-end integration tests (UNION both arms + mixed + order-independence, MINUS, OPTIONAL) covering the azure-chat scenarios.
- JSON-LD query surface is unaffected (separate parser; `{query}`-in-union already worked and still passes).

A PR fixing this is on the way.

aaj3f:
Fixes #1435. Repairs the root cause behind **fluree/azure-chat#42** (sub-SELECT in `UNION`) and **fluree/azure-chat#43** (sub-SELECT in `MINUS`).

## The bug

A SPARQL sub-SELECT (`{ SELECT ... }`) used as an operand of a set/graph operation was mis-parsed, silently returning wrong results:

| Query shape | Expected | Before |
|---|---|---|
| `{ SELECT (…AS ?n) } UNION { SELECT (…AS ?n) }` | both arms | one arm only (or 0 rows) |
| `{ SELECT (…AS ?n) } MINUS { SELECT (…AS ?n) }` | difference | full left set (no subtraction) |
| `{ plain } UNION { SELECT (…AS ?n) }` | both arms | sub-SELECT arm dropped |
| `OPTIONAL { SELECT (COUNT(…) AS ?c) … }` | `?c` bound | `?c` lost, rows duplicated |

Silent (HTTP 200, no error), which is worse than a hard failure — NL→SPARQL / SQL→SPARQL generators emit `{ SELECT } UNION { SELECT }` and `{ SELECT } MINUS { SELECT }` routinely for "either/or" and "without" questions.

## Root cause

`Parser::parse_group_graph_pattern` (`fluree-db-sparql/src/parse/query/pattern.rs`) only recognised a `{ SELECT ... }` sub-SELECT **inside its own nested-`{` branch**. Every other position that opens a group — `OPTIONAL`, `GRAPH`, `SERVICE`, the right arm of `MINUS`, each arm of `UNION` (`parse_union_continuation`), `EXISTS`/`NOT EXISTS` (the free `parse_group_graph_pattern`), and a bare `WHERE { SELECT … }` — consumes the `{` itself and then calls `parse_group_graph_pattern` positioned at the `SELECT`, where the leading `SELECT` was treated as an unexpected token.

- **UNION**: the left sub-SELECT was pushed with no trailing-`UNION` check (the plain-group branch had one), so `UNION` hit `"UNION must follow a pattern"` and was discarded — the query collapsed into two independent subqueries lowered as a **conjunction**. At execution that becomes a correlated join on the shared projected variable; the arms' values are disjoint, so one arm survives (the smaller-cardinality one; ties → left) or 0 rows. (So it is *not* "only the last arm," as azure-chat#42 guessed.)
- **MINUS / OPTIONAL / GRAPH / SERVICE**: the sub-SELECT was parsed as a bare BGP with the `(expr AS ?v)` projection **dropped**. MINUS then shares no bound variable with the left → subtracts nothing; OPTIONAL loses the projected/aggregate variable.

Lowered IR confirms it — before: `{SELECT} UNION {SELECT}` → `[Subquery, Subquery]` (no `Union`); `{SELECT} MINUS {SELECT}` → `[Subquery, Minus(Triple×4)]` (projection gone).

## The fix (parser-only)

`fluree-db-sparql/src/parse/query/pattern.rs`:

1. Detect the `SubSelect` alternative at the **top** of `parse_group_graph_pattern` (the opening `{` is already consumed by the caller), so every caller implements the grammar rule `GroupGraphPattern ::= '{' ( SubSelect | GroupGraphPatternSub ) '}'`. This fixes MINUS/OPTIONAL/GRAPH/SERVICE/EXISTS/UNION-right and bare `WHERE { SELECT }` in one place.
2. Route the nested-`{` branch through `parse_group_graph_pattern` (which now handles both a group and a sub-SELECT) and run the trailing-`UNION` check for a sub-SELECT left arm too.

Error recovery for a malformed sub-SELECT is preserved (skip to the matching `}`).

## Performance

No query-engine change. Fixed queries now compile to the **same** `UnionOperator` the equivalent flat `UNION` always used, and to a proper sub-SELECT `MINUS` — i.e. the already-optimised plans. The *buggy* path was actually a more expensive per-row correlated `SubqueryOperator` join chain. So this is neutral-to-positive on speed, with no memory impact.

## Testing

- **Parser AST tests** (`fluree-db-sparql`): sub-SELECT in UNION (both arms + mixed group/sub-SELECT), MINUS right arm, OPTIONAL body. `assert_parses` also guards that the spurious `"UNION must follow a pattern"` diagnostic is gone.
- **Integration tests** (`fluree-db-api/tests/it_query_sparql_setop_subselect.rs`): the verbatim azure-chat 42 & 43 queries and near-neighbors, asserting end-to-end results (UNION == flat UNION; MINUS == FILTER NOT EXISTS; OPTIONAL binds the aggregate).
- **W3C SPARQL suite** (before → after): syntax-query **81 → 83** (`syntax-SELECTscope1`, a positive `{SELECT} UNION {SELECT}` syntax test), subquery **11 → 12** (`sq12`, bare `WHERE { SELECT }` in CONSTRUCT), negation **12 → 12** (no W3C test covers sub-SELECT-as-MINUS-operand — the new integration tests fill that gap). **Zero regressions.**
- Full `grp_query_sparql` (265) and `grp_query` JSON-LD (303) integration groups pass; `cargo clippy -p fluree-db-sparql --all-features -- -D warnings` clean.
- JSON-LD query surface is unaffected (separate parser; `{query}`-in-union already worked and still passes).

## Files

- `fluree-db-sparql/src/parse/query/pattern.rs` — the fix.
- `fluree-db-sparql/src/parse/query/tests.rs` — parser AST regression tests.
- `fluree-db-api/tests/it_query_sparql_setop_subselect.rs` (+ `grp_query_sparql.rs`) — integration regression tests.

### DIFF
--- a/fluree-db-api/tests/grp_query_sparql.rs
+++ b/fluree-db-api/tests/grp_query_sparql.rs
@@ -13,6 +13,8 @@ mod it_query_negation;
 mod it_query_sparql;
 #[path = "it_query_sparql_annotations.rs"]
 mod it_query_sparql_annotations;
+#[path = "it_query_sparql_setop_subselect.rs"]
+mod it_query_sparql_setop_subselect;
 #[path = "it_query_subquery.rs"]
 mod it_query_subquery;
 #[path = "it_query_unwind.rs"]
--- a/fluree-db-api/tests/it_query_sparql_setop_subselect.rs
+++ b/fluree-db-api/tests/it_query_sparql_setop_subselect.rs
@@ -1,0 +1,242 @@
+//! Regression tests: SPARQL sub-SELECT as an operand of a set operation.
+//!
+//! Fixes azure-chat #42 (sub-SELECT in UNION) and #43 (sub-SELECT in MINUS),
+//! plus the broader family (OPTIONAL, mixed arms) sharing the same parser root
+//! cause: `parse_group_graph_pattern` only recognised a `{ SELECT ... }`
+//! sub-SELECT inside its own nested-`{` branch, so a sub-SELECT used as a set-op
+//! operand was mis-parsed — UNION dropped the operator, MINUS/OPTIONAL dropped
+//! the `(expr AS ?v)` projection. See `fluree-db-sparql` parser tests for the
+//! AST-level guards; these assert the end-to-end query results.
+//!
+//! All inserts and queries are explicit with `@context` / `PREFIX`.
+
+use crate::support;
+use crate::support::{genesis_ledger, normalize_rows, MemoryFluree, MemoryLedger};
+use fluree_db_api::FlureeBuilder;
+use serde_json::json;
+
+/// 3 Person, 2 Company, 5 Country, 2 Maker (France + Japan have a maker).
+/// This is the exact dataset from the azure-chat repro.
+async fn seed_setop(fluree: &MemoryFluree, ledger_id: &str) -> MemoryLedger {
+    let ledger0 = genesis_ledger(fluree, ledger_id);
+    let insert = json!({
+        "@context": {"ex": "http://example.org/"},
+        "@graph": [
+            {"@id": "ex:alice",  "@type": "ex:Person",  "ex:name": "Alice"},
+            {"@id": "ex:bob",    "@type": "ex:Person",  "ex:name": "Bob"},
+            {"@id": "ex:carol",  "@type": "ex:Person",  "ex:name": "Carol"},
+            {"@id": "ex:acme",   "@type": "ex:Company", "ex:name": "Acme"},
+            {"@id": "ex:globex", "@type": "ex:Company", "ex:name": "Globex"},
+            {"@id": "ex:france",  "@type": "ex:Country", "ex:cname": "France"},
+            {"@id": "ex:germany", "@type": "ex:Country", "ex:cname": "Germany"},
+            {"@id": "ex:japan",   "@type": "ex:Country", "ex:cname": "Japan"},
+            {"@id": "ex:brazil",  "@type": "ex:Country", "ex:cname": "Brazil"},
+            {"@id": "ex:canada",  "@type": "ex:Country", "ex:cname": "Canada"},
+            {"@id": "ex:maker1", "@type": "ex:Maker", "ex:inCountry": {"@id": "ex:france"}},
+            {"@id": "ex:maker2", "@type": "ex:Maker", "ex:inCountry": {"@id": "ex:japan"}}
+        ]
+    });
+    fluree.insert(ledger0, &insert).await.unwrap().ledger
+}
+
+async fn sparql_rows(fluree: &MemoryFluree, ledger: &MemoryLedger, q: &str) -> serde_json::Value {
+    support::query_sparql(fluree, ledger, q)
+        .await
+        .expect("sparql query should succeed")
+        .to_jsonld(&ledger.snapshot)
+        .expect("to_jsonld")
+}
+
+/// azure-chat #42, verbatim: `{ SELECT (?pn AS ?n) } UNION { SELECT (?cn AS ?n) }`
+/// must return the union of both arms, not just one.
+#[tokio::test]
+async fn sparql_subselect_union_returns_all_arms() {
+    let fluree = FlureeBuilder::memory().build_memory();
+    let ledger = seed_setop(&fluree, "setop:union").await;
+
+    let q = r"
+        PREFIX ex: <http://example.org/>
+        SELECT DISTINCT ?n WHERE {
+          { SELECT (?pn AS ?n) WHERE { ?p a ex:Person ; ex:name ?pn } }
+          UNION
+          { SELECT (?cn AS ?n) WHERE { ?c a ex:Company ; ex:name ?cn } }
+        }
+    ";
+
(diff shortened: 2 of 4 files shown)


bplatz (approved):
👍

PR #1437, opened 2026-07-06
(SPARQL) Full W3C Suite in CI w/ Registers for Skip/Ignore

aaj3f:
## Summary

This PR makes the entire W3C SPARQL test suite (the `rdf-tests` submodule, ~1,420 tests) an enforced, durable CI gate, and carries the complete, adversarially-verified burn-down plan for the remaining gaps. Before this change, the `testsuite-sparql` CI job ran `cargo test --no-run` — it compiled the harness but **never executed a single test** — and every evaluation category was `#[ignore]`d.

Now `cargo test` in `testsuite-sparql/` runs **all 36 suites green in ~15s**, and CI runs exactly that on every PR. Every test either passes or appears in an explicit, root-cause-grouped skip register that is policed in both directions: an unregistered failure fails CI, and a registered test that starts passing **also** fails CI (stale entry). Registers can only shrink.

**This PR contains zero engine changes** — only the out-of-workspace `testsuite-sparql` crate, CI config, and docs. Hot-path risk is nil by construction. The engine fixes themselves live on 14 follow-up branches (see "What follows this PR").

## Reviewer's guide — where to spend attention

**Review carefully (the real surface, ~1.5k lines):**
- `testsuite-sparql/src/` — harness changes: `mf:UpdateEvaluationTest` implemented end-to-end over the public `sparql_update` transact surface; named graphs loaded as real named graphs (TriG via the transact builder); CSV/TSV expected-result comparison; `.rdf`/SRX result-parser fixes; the both-way register policing in `lib.rs::check_testsuite` (unexpected-pass detection, zero-test floor, infra-death rule, dead-entry detection).
- `testsuite-sparql/tests/registers/mod.rs` — the skip registers. Each entry carries a root-cause comment and the future PR that owns it.
- `.github/workflows/ci.yml` — the compile-only step replaced with the real run.

**Skim (analysis artifacts, ~6k lines):** `docs/audit/2026-07-sparql-testsuite-audit.md` (taxonomy + perf discipline) and `docs/audit/burn-down/` (nine per-cluster root-cause audits + `ROADMAP.md`, the verified implementation plan: PR slate, sequencing, decision register D-1..D-13, wave outcome records). These documents were produced under a two-stage process — parallel deep audits followed by adversarial verification that refuted 13 of 55 load-bearing claims before anything was implemented; §1.1 of the ROADMAP records every refutation.

## Notable content decisions

- **Skip-register policy**: entries are debt with a named owner, not permanent waivers — except the deliberately-permanent classes (protocol/HTTP suites = not applicable to an embedded engine; entailment regimes = documented non-goal, with the simple-entailment-answerable subset pinned green).
- **Decision register D-1..D-12 adopted** (ROADMAP §4) with a standing rule: any follow-up PR actioning a decision must present the options and rationale in its description.
- **CI tests the merge commit**, so the registers track main: when #1436 (sub-SELECT set-op fix) landed mid-effort, the stale-entry gate caught the four affected tests and the re-baseline commit documents exactly what moved — that mechanism working is the point of this PR.

## Fallout already handled on follow-up branches (not in this PR)

The suite immediately surfaced 10 production bugs, filed as #1438–#1447 with verified mechanisms (multi-op UPDATE silent data loss, `FILTER(true)` dropping all rows, `DATATYPE()` rejecting expressions, `GRAPH ?g` binding literals and leaking the default graph, a Turtle lexer rejection of valid `_:label.`, non-canonical `xsd:double` output, and more). Eight of the ten are already fixed on the follow-up branches below.

## What follows this PR

14 implementation branches (Waves 0–2 of the roadmap, ~230 register entries removed, all suite-green, all bench-gated) stack on this branch and become PRs after it merges, in the documented order: `pr-h1` first (register hygiene), then siblings (`pr-1`, `pr-l1`, `pr-l2`, `pr-x1`, `pr-pp`, `pr-2`, `pr-3`, `pr-u1`, `pr-base`, `pr-g1`, `pr-w15`), then the two stacked pairs (`pr-w2a` after `pr-1`; `pr-u2` after `pr-3`). Merge choreography, joint-entry rules, and the Wave-3+ plan are in `docs/audit/burn-down/ROADMAP.md`.

## Test plan

- `cd testsuite-sparql && cargo test` — 36 suites, 0 failed, 0 ignored (~15 s).
- `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings` clean.
- CI job runs the full suite with a 45-minute timeout backstop (actual runtime is minutes, dominated by build).
- No workspace crates touched.

### DIFF
--- a/testsuite-sparql/src/query_handler.rs
+++ b/testsuite-sparql/src/query_handler.rs
@@ -76,6 +76,23 @@ pub fn evaluate_update_evaluation_test(test: &Test) -> Result<()> {
         .clone()
         .context("UpdateEvaluationTest missing ut:request (update file URL)")?;
 
+    // Guard against a mis-parsed mf:result degrading to a trivially
+    // satisfiable "expected empty store": the result node must have exposed
+    // at least one recognized predicate (ut:data / ut:graphData / ut:result).
+    // A deliberate empty-store expectation always carries `ut:result
+    // ut:success` in the W3C manifests.
+    if test.result_data.is_none()
+        && test.result_graph_data.is_empty()
+        && !test.result_success
+        && test.result.is_none()
+    {
+        bail!(
+            "UpdateEvaluationTest mf:result parsed to an empty expectation \
+             (no ut:data, ut:graphData, or ut:result) — unrecognized result \
+             shape?\nTest: {test_id}"
+        );
+    }
+
     let descriptor = TestDescriptor::UpdateEval {
         test_id,
         request_url,
@@ -188,6 +205,10 @@ async fn fetch_state(fluree: &Fluree) -> Result<LedgerState> {
 /// TriG requires directives at document scope, so when wrapping a Turtle
 /// file's content in a `GRAPH { }` block its `@prefix`/`@base` lines must be
 /// hoisted out. W3C test data declares directives one per line.
+///
+/// ASSUMPTION: line-based detection. A multi-line string literal whose
+/// continuation line happens to start with `prefix`/`base` would be
+/// mis-hoisted; no W3C test data has that shape.
 fn split_turtle_directives(content: &str) -> (String, String) {
     let mut directives = String::new();
     let mut body = String::new();
@@ -283,12 +304,19 @@ async fn list_named_graphs(fluree: &Fluree, ledger: &LedgerState) -> Result<Vec<
     let SparqlResults::Solutions { solutions, .. } = results else {
         bail!("Graph enumeration returned non-SELECT results");
     };
+    // The engine currently binds `?g` as a plain string literal, not an IRI
+    // term, and also exposes the default graph under the ledger alias (both
+    // registered engine gaps — audit §8). Accept both term kinds so this
+    // enumeration keeps working when the engine is fixed, and exclude the
+    // alias-named default graph: only real named graphs count.
     Ok(solutions
         .into_iter()
         .filter_map(|sol| match sol.get("g") {
             Some(RdfTerm::Iri(iri)) => Some(iri.clone()),
+            Some(RdfTerm::Literal { value, .. }) => Some(value.clone()),
             _ => None,
         })
+        .filter(|name| name != TEST_LEDGER)
         .collect())
 }
 
--- a/testsuite-sparql/src/query_handler.rs
+++ b/testsuite-sparql/src/query_handler.rs
@@ -1,22 +1,31 @@
-//! `QueryEvaluationTest` handler: create an in-memory Fluree ledger, load
-//! test data, execute a SPARQL query, and compare against expected results.
+//! `QueryEvaluationTest` / `UpdateEvaluationTest` handlers: create an
+//! in-memory Fluree ledger, load test data (default + named graphs), execute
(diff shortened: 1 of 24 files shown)


testsuite-sparql/src/query_handler.rs
```diff
@@ -177,6 +398,120 @@ pub async fn run_eval_test(
...
+            actual,
+            Some(expected_url),
+        )?;
+    }
+
+    // 5. No unexpected non-empty named graphs.
+    //
+    // Note: this compares *non-empty* graphs only. Fluree does not track
+    // empty named graphs as first-class graph-store entries, so tests that
+    // distinguish CLEAR (graph remains, empty) from DROP (graph removed)
```
bplatz: The both-way register policing is nice, but it can only catch a *registered* test that passes or an *unregistered* test that fails — never an unregistered test that passes but shouldn't. The documented leniencies here open exactly that blind spot: CLEAR vs DROP is indistinguishable (empty named graphs not tracked), and an update erroring with `is_empty_transaction` is swallowed as a no-op (line ~433), so a wrongly-empty update passes whenever expected state == pre-state. If the specific W3C tests exercising these distinctions aren't in the register, they false-pass invisibly. Worth confirming those are registered even though they currently "pass".

testsuite-sparql/src/query_handler.rs
```diff
@@ -54,78 +63,282 @@ pub fn evaluate_query_evaluation_test(test: &Test) -> Result<()> {
...
+    // enumeration keeps working when the engine is fixed, and exclude the
+    // alias-named default graph: only real named graphs count.
+    Ok(solutions
+        .into_iter()
+        .filter_map(|sol| match sol.get("g") {
+            Some(RdfTerm::Iri(iri)) => Some(iri.clone()),
+            Some(RdfTerm::Literal { value, .. }) => Some(value.clone()),
+            _ => None,
+        })
+        .filter(|name| name != TEST_LEDGER)
```
bplatz: This excludes the default graph by exact alias string, assuming the engine surfaces it under exactly `w3c:test`. If the engine ever branch-qualifies or IRI-expands that name, every update-eval test with default-graph data will spuriously report an "unexpected non-empty named graph" and fail. Sturdier to subtract the known default-graph triples or match the engine's actual default-graph identifier rather than string-compare the alias.

bplatz (approved):
A couple comments on this one, great to have additional coverage!

testsuite-sparql/src/query_handler.rs
aaj3f: valid latent fragility, now fixed and proven. Low severity (it would've failed loud, not silent), but the coupling was invisible and I've changed that

testsuite-sparql/src/query_handler.rs
aaj3f: Valid structural insight. CLEAR/DROP/CREATE were already handled (registered); the open piece was net-zero vacuity, now enumerated (12 tests) and documented with a concrete watch-list. Nothing needs registering.

PR #1448, opened 2026-07-07
fix: close Cypher benchgraph gaps and add benchmark-critical fast paths

bplatz:
## Summary

Takes the Pokec benchmark to 35/35 supported queries and brings the hot benchmark shapes to seek/count-store parity. Three groups of changes:

### Cypher surface gaps
- Untyped single-hop patterns (`-->` / `-[e]->`) now follow relationships only (rdf:type and data properties excluded), matching the untyped var-length wildcard.
- Value-only bound relationship variables lower to base triple + optional annotation probe + coalesce, so plain-RDF edges match while reified parallel edges keep one row per annotation. COALESCE gained a binding-path arm so structured values survive BIND.
- Untyped `shortestPath` / `allShortestPaths` (`ShortestPathPattern.predicate` is now `Option<Sid>`), with per-hop predicates resolved post-hoc for `relationships(p)`.
- Unary minus folds at parse time (`SET n.p = -1`, negative inline properties).
- Bare `CREATE ()` / `CREATE (n)` asserts a hidden `db:Node` existence marker; `UNWIND range(...)` before a write clause reuses the `$param` batch desugar.
- **RETURN on write statements**: created-entity Sids are deterministic given a caller-supplied skolem txn id (`TxnOpts::skolem_txn_id`), so rows are reconstructed post-commit — on the HTTP route and the new `Fluree::transact_cypher_returning` API.
- Bare `MATCH (n)` whole-graph scans available behind `FLUREE_CYPHER_ALLOW_FULL_SCAN=1` (off by default).

### Performance
- **Stats-driven write WHERE planning**: `execute_where_streaming` previously planned with no statistics, so anchored MATCHes label-scanned. Now builds the same cached novelty-merged stats view as the read path (10k-user A/B: 82.7s → 25ms). Applies to every WHERE-driven write surface, not just Cypher.
- Property-join stars driven from the most selective bound object instead of hard-preferring rdf:type (labeled point lookup 4ms → 0.65ms; cost now constant in label size).
- New metadata-lane folds: whole-graph scalar aggregates from index directories (`count(n)` 27ms → 0.5ms), class-anchored aggregates and GROUP-BY histograms, group-key-filtered histograms.
- Equality-filtered OPTIONALs fold into required triples; anonymous hop chains fuse into frontier-BFS reachability; Cypher pattern IRIs encode to SIDs at lowering.

### Correctness hardening for the fast paths
- Fast paths gate on live novelty rather than the monotonic overlay epoch.
- Base per-graph property stats carry through incremental indexing.
- Distinct-count/aggregate folds decline on partial stats and scan-hidden predicates (`f:reifies*`).
- Specific-value lookups stay at seek speed under predicate novelty.

### Test wiring
Re-wires four `fluree-db-api` test files that were orphaned from CI (`autotests = false` with no `[[test]]` target): `it_query_cypher` (156 tests), `it_policy_cypher`, `it_csv_import`, `it_tpch_iceberg_bench`. Also updates the Cypher docs (support matrix, relationship-lowering semantics, configuration) and includes the Bolt protocol adapter design doc.

## Test plan
- [x] `cargo nextest run --workspace --all-features` (including the newly re-wired test targets)
- [x] Benchgraph Pokec suite: 35/35 supported, results verified against the general pipeline (including multi-valued property row semantics)
- [x] Write-path A/B on 10k-user in-memory and indexed pokec ledgers

### DIFF
--- a/fluree-db-query/src/fast_whole_graph_agg.rs
+++ b/fluree-db-query/src/fast_whole_graph_agg.rs
@@ -32,11 +32,28 @@
 //! (`#P1(n) × #P2(n)`), which changes duplicate-sensitive aggregates — the
 //! detector only admits the single-accessor shape.
 //!
+//! The class-anchored family generalizes the same folds to
+//! `MATCH (n:C) RETURN …` (anchor = `?n rdf:type <C>`): the subject universe
+//! becomes the class's instance count from the per-graph class stats, and
+//! property-derived folds additionally require the **containment proof**
+//! `class_property_flakes(C, P) == predicate_rows(P)` — equality means every
+//! `P`-bearing subject is a `C`, so predicate-scoped folds equal the
+//! class-restricted join. Both sides are current-state-exact at HEAD (class
+//! stats per issue #1266; rows from PSOT directories).
+//!
+//! Beyond the single-row scalar shape, `GROUP BY ?prop` + `COUNT(*)` (the
+//! Cypher histogram `RETURN n.age, COUNT(*)`) folds to a predicate-scoped
+//! POST run-length group count (values are physically contiguous in POST)
+//! plus one null-group row (`subjects − subj(P)` — the left join gives
+//! property-less subjects a single unbound-key row).
+//!
 //! The fold requires the strict metadata-lane gate ([`fast_path_store`]:
-//! single-ledger, root/no policy, no overlay, at HEAD) and declines when the
-//! graph carries predicates the variable-predicate scan hides (`f:reifies*`
-//! anywhere, the `f:` namespace in the default graph) — those facts are
-//! invisible to `?n ?p ?o` but not to the SPOT directories this fold reads.
+//! single-ledger, root/no policy, no overlay, at HEAD). The whole-graph
+//! anchor additionally declines when the graph carries predicates the
+//! variable-predicate scan hides (`f:reifies*` anywhere, the `f:` namespace
+//! in the default graph) — those facts are invisible to `?n ?p ?o` but not
+//! to the SPOT directories; a class anchor reads only class stats, so it is
+//! immune.
 
 use crate::binding::{Batch, Binding};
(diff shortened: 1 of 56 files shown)


fluree-db-query/src/fast_whole_graph_agg.rs
```diff
@@ -0,0 +1,698 @@
...
+            return Ok(true);
+        };
+        let hidden = fluree_db_core::is_reserved_reifies_predicate(&sid)
+            || (ctx.binary_g_id == 0 && sid.namespace_code == fluree_vocab::namespaces::FLUREE_DB);
+        if hidden && count_rows_for_predicate_psot(store, ctx.binary_g_id, p_id)? > 0 {
+            return Ok(true);
+        }
+    }
+    Ok(false)
+}
```
aaj3f, line 698: `graph_has_scan_hidden_predicates` walks the full predicate dictionary (`0..predicate_count()`) on every whole-graph-anchored aggregate that reaches the fold. It's directory-only and the per-predicate work is cheap, but on a store with a large dictionary this is a non-trivial fixed cost on a path meant to be sub-millisecond. Fine as-is; noting for awareness.

aaj3f (approved):
These are excellent perf wins! Glad to have them!

Only note, and honestly take-it-or-leave-it (though I'm excited to take our benchmarking harness and actually use it in real, durable, ways... we still haven't implemented the nightly or weekly benchmark runners, so they're not yet valuable bench regression guards--**though they could be with your help!!** 😆)
Many new hot-path fast paths ship with unit/integration coverage but **no criterion bench** (the diff touches no `benches/` dir). Per-PR CI only compiles benches, so this isn't a gate failure, but the repo convention (`docs/contributing/benches.md`, `regression-budget.json`) is that a hot-path change should carry a bench so the nightly job can catch a future regression. Consider a `query_whole_graph_agg` bench.

PR #1449, opened 2026-07-07
feat: Bolt protocol endpoint + Cypher performance campaign

bplatz:
Adds a Bolt protocol endpoint (Neo4j drivers → openCypher), moves Cypher to bare namespace-0 names by default.

## Bolt protocol server

- `fluree-db-bolt` protocol crate: PackStream codec + session state machine; listener in `fluree-db-server` (autocommit + explicit `BEGIN`/`COMMIT`/`ROLLBACK`), `Node`/`Relationship`/`Path` + temporal structures, write-`RETURN` entities as typed nodes.
- Session authentication with identity-derived policy enforcement.
- `bolt` feature on by default (listener binds only when `bolt_listen_addr` is set); first-class CLI flags on `server run/start/restart`; reference + guide docs.

## Bare namespace-0 names as the Cypher default

Cypher labels/types/property keys now live under namespace 0 (the empty prefix) instead of being forced under `http://example.org/`: zero-config durable names for LPG users, planner keeps full Sid-gating, and bare JSON-LD terms converge on the same names. RDF compat is the opt-in via `@context`/`@vocab` (ledger default or per-view). Placement rule is locked by test: no colon → whole name verbatim (never namespace-split); colon-containing backticked names are the RDF escape hatch.

## Performance campaign

Read/serve path, all query surfaces (SPARQL and JSON-LD ride the same substrate):

- **Shared leaf mmaps** through the leaflet cache — leaf files are immutable CAS artifacts; the per-query open+mmap+munmap cycle and whole-leaf `fs::read` copies were the dominant fixed cost of an indexed point lookup.
- **Cascaded within-leaflet seek**: bound point probes binary-search every bound sort-key prefix column (was: leading column only), making them leaflet-size-independent on all four index orders.
- **Galloping batched lookups** (SPOT/PSOT/OPST): sorted-merge the wanted-id list against leaflet batches instead of per-row membership tests; chunker gap-split reinstated at the measured break-even (4096 ids).
- **Bolt node-emit hydration**: batched sorted SPOT crawl engaged under live novelty with per-subject overlay dirty-split; shared `Arc<str>` strings through the encode path; adjacency no longer inlined into property maps.
- **Raw-id shortestPath lane**: bidirectional BFS in `s_id` space with batched frontier expansion (one galloping sweep per side per level); dictionary touched only at endpoints + final path. Per-node overlay dirty-split keeps novelty shortcuts/retracts/re-asserts and novelty-only endpoints exact (end-to-end test); views the lane can't serve (policy, unsummarizable overlay) keep the Sid search.
- Point-lookup floor: parsed-AST cache, reified-edge annotation probe gate, parallel wildcard JSON-LD prefetch.

## Fixes surfaced along the way

- Time-travel over fully retracted predicates regressed by the bound-object driver change: driver selection is rdf:type-first under replay (`to_t < index max_t`) so probes stay on the replay-correct SPOT sidecar lane. (The underlying rebuild-path hole — omitted empty partitions — predates this branch and remains open.)
- Narrow K≥2 hydration projections no longer lose their predicate filter to the wildcard flake cache.
- Specific-value lookups keep seek speed under predicate novelty; fast paths gate on live novelty rather than the monotonic overlay epoch; per-graph property stats carried through incremental indexing.

## Validation

- Full workspace suites green (api 2569 / query+indexer 1538 / server 414; the one recurring failure is the documented raft-liveness flake, passing in isolation).
- New coverage: NS-0 round-trip + RDF-compat (API view-level and server ledger-level), exotic backticked-name no-split rule, shortestPath novelty semantics (shortcut / base-edge retract / re-assert / novelty-only endpoint), cascade-seek unit tests.
- Probes for regression tracking: `bolt_node_emit_probe`, `cypher_point_floor_profile` (per-phase floor decomposition, `PROBE_ONLY`/`PROBE_USERS` filters).


fluree-db-bolt/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+pub mod value;
+
+pub use handshake::BoltVersion;
+pub use value::Value;
+
+/// Errors from decoding inbound bytes (malformed PackStream, oversized
+/// values, unknown message tags). These are protocol violations: the
+/// connection should fail the request and usually close.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DecodeError(pub String);
```
aaj3f, line 30: This is super minor, but generally in most `lib.rs` files, the repo convention has been a `thiserror` `Error` enum per library crate. Here it's a single-variant manual struct with correct `Display`/`Error` impls and no `anyhow`. Honestly, reasonable for a small codec lib crate, but just flagging for consistency. If decode-error variants ever need structure (e.g. distinguishing oversize vs. malformed-marker vs. truncated), converting to `thiserror` then would be cleaner.

fluree-db-query/src/shortest_path.rs
```diff
@@ -717,6 +846,387 @@ impl ShortestPathOperator {
...
+        static EMPTY: OnceLock<Arc<DirtyIds>> = OnceLock::new();
+        return Some(Arc::clone(EMPTY.get_or_init(|| {
+            Arc::new(DirtyIds {
+                subjects: FxHashSet::default(),
+                objects: FxHashSet::default(),
+            })
+        })));
+    }
+    let version = overlay.content_version()?;
+    let store_key = Arc::as_ptr(store) as usize;
```
aaj3f: Probably very unlikely but...

the `(content_version, g_id, store_ptr)` key is subject to A -> B -> A misreads (i.e. could be read as unchanged though in fact had changed twice): if a store is dropped and a new `BinaryIndexStore` is allocated at the same address with a coinciding `content_version` and `g_id`, a stale dirty-id set (resolved against the old store's subject-id space) could be served — and a wrongly-clean node would take the batched base sweep and miss novelty edges, yielding an incomplete shortest path. The codebase already solves exactly this elsewhere with a monotonic store-instance id (`NEXT_STORE_ID`, used by `LeafletCache::segment_ops_key` precisely "to make the key sound"). Suggest keying on that stable id instead of the raw pointer for parity and to close the ABA window.

fluree-db-binary-index/src/read/leaflet_cache.rs
```diff
@@ -353,6 +359,11 @@ impl CachedEntry {
             CachedEntry::R2(r2) => r2.byte_size(),
             CachedEntry::DictLeaf(bytes) => bytes.len(),
             CachedEntry::LeafDir(dir) => dir.byte_size(),
+            // A mapping is file-backed (reclaimable page cache), not heap:
+            // weigh it far below its length so it doesn't crowd out decoded
+            // batches, but keep it length-proportional so total mapped
+            // address space stays bounded by the cache budget (× 64).
+            CachedEntry::LeafMmap(mmap) => (mmap.len() / 64).max(16 * 1024),
```
aaj3f, line 366: Also probably an extreme edge case condition, but this could allow mmap failure in (admittedly very rare) scenarios. The under-weighing is well-reasoned (file-backed reclaimable pages, not heap) and keeps mapped address space bounded at ~64× the cache budget, which is fine on 64-bit. But each mapping is a distinct VMA, and at a large cache budget with many small/typical leaves the 16 KiB floor could admit tens of thousands of mappings, approaching Linux's default `vm.max_map_count` (65530) once combined with the process's other mappings — at which point `mmap` fails with ENOMEM. Worth a one-line note in the guide / an eye on it for large `leaflet_cache` budgets. The `NotFound→owned-read` fallback in `get_leaf_bytes_shared` means a failed mmap isn't fatal, but a non-NotFound `mmap` error there propagates rather than falling back.

aaj3f (approved):
Neat addition and some great proofs of perf wins

PR #1450, opened 2026-07-07
feat(iceberg): virtual-dataset info APIs + decimal-FK detection + subject-key fallback + secret redaction

aaj3f:
## What & why

Makes an R2RML/Iceberg **virtual (query-in-place) dataset** a first-class citizen of the dataset-info APIs and fixes deterministic generation for real Snowflake schemas. Four original changes from a downstream audit of fluree/solo's "generate a dataset from a catalog" flow, plus (added since the original body) the act-step engine capabilities, two virtual-query perf fixes, and a review-response wave:

1. **Virtual-aware `ledger_info`** — the dataset About/Data panels (stats, classes/properties, counts) were empty for a virtual dataset because `info`/`data-model` derive everything from a committed `LedgerState` a virtual graph doesn't have.
2. **Config-secret redaction (security)** — the thin graph-source-info stub serialized the resolved OAuth2 `client_secret` (a live PAT) to clients. A second instance of the same class (the SSE `ns-record` stream) was also leaking.
3. **Decimal-key FK detection** — FK inference was integer-only, so Snowflake surrogate keys (`decimal(38,0)`, i.e. `NUMBER(38,0)`) were silently dropped before the name match — no joins, no diagnostic.
4. **Subject-key auto-fallback** — a table with no verifiable identifier emitted *no subject*, so "select all tables" left several tables unsaveable.
5. **Act-step engine caps** (WP-DB2/WP-DB3) — a bounded row/column sampler over Iceberg tables (`sample_iceberg_rows` / `sample_column_values`) and `query_provisional_r2rml` (run real SPARQL against an unpersisted candidate mapping + connection, registering nothing global) — the LLM agent's data-peek and empirical-validation lanes.
6. **Virtual-query performance** — the browse crawl ("View Instances") and subject inspector no longer time out at scale: crawl routes through R2RML as a SINGLE budgeted scan, RefObjectMap refs template from the child row's FK under the crawl (no parent scans), bound-subject queries prune to the matching table, R2RML scans poll cancellation, and crawl OFFSET paginates (with a depth ceiling). Trace spans (DEBUG) cover the scan hot path for x-ray attribution.

### Virtual-aware `ledger_info` (`fluree-db-api/src/ledger_info.rs`, `fluree-db-server/src/routes/ledger.rs`)
- `LedgerInfoBuilder::execute` now falls back to `nameservice().lookup_graph_source()` on ledger-not-found and routes through a shared `build_graph_source_info`. The native committed-ledger path is **shape/value-identical** (one deliberate cosmetic delta from the typed-`LedgerInfo` refactor: nested `stats.properties`/`classes`/`datatypes` map keys re-emit in sorted order instead of source order — same keys and values, top-level order preserved, no consumer parses positionally).
- For Iceberg/R2RML, classes/predicates are derived from the compiled mapping (same resolver the query path uses, `FlureeR2rmlProvider::compiled_mapping`), and per-class/predicate counts come from the **Iceberg manifest `row_count`** (`preview_iceberg_table` at `StatsTier::Schema`) — metadata-only, **no Parquet scan** (cheaper than the native index path). Per-class property membership and FK ref-classes derive from the mapping; row counts fetch through ONE shared REST client with a bounded fan-out and a wall-clock budget (`FLUREE_ICEBERG_INFO_COUNT_BUDGET_MS`, default 10s), degrading to structure-only on timeout.
- The db-server thin stub `graph_source_info_json` is removed; both `/info` sites call the shared builder. Counts are `null` (not `0`) when unknown; the `source` block is the virtual-only addition (see shape below) and never carries credentials.

**Virtual `ledger_info` JSON shape** (the contract solo's panels consume; `stats.classes`/`stats.properties` mirror the native shape so existing consumers keep working):
```jsonc
{
  "ledger_id": "<gs_id>", "t": <snapshotId|null>,
  "ledger": { "alias": "<gs_id>", "t": <snapshotId|null>, "commit-t": null, "index-t": null,
              "flakes": <totalRows|null>, "size": 0, "named-graphs": [ /* urn:default */ ] },
  "graph": "urn:default",
  "stats": { "flakes": <totalRows|null>, "size": 0,
             "properties": { "<predIri>": { "count": <int|null>, "datatypes": { "<xsd|@id>": <int|null> } } },
             "classes":    { "<classIri>": { "count": <int|null>, "properties": { /* native shape */ } } } },
  "commit": null,
  "nameservice": { /* gs_record_to_jsonld — config REDACTED */ },
  "source": { "virtual": true, "type": "Iceberg"|"R2RML", "tables": ["ns.table", ...],
              "snapshot": <id|null>, "catalog": { "type": "...", "uri": "...", "warehouse": "...|null" },
              "table-row-counts": { "ns.table": <int>, ... } }   // NO credentials
}
```

### Secret redaction
- `redact_graph_source_config` (recursive; masks `client_secret`/`token`/`secret`/`password`/AWS keys/`default_val`, keeps an env-var *name*, and is a **byte-identical no-op** for secret-free configs so BM25/Vector/etc. are unchanged). Applied at every client-facing emission: `gs_record_to_jsonld`, the generic stub, and the SSE stream. The virtual builder structurally never emits auth.
- **Fails closed** (review hardening): input the redactor cannot parse as JSON is replaced with a placeholder, never echoed. A value-pattern backstop test serializes EVERY `AuthConfig` variant with sentinel secrets through the real storage path and asserts none survives redaction — with a wildcard-free `match` as an exhaustiveness canary, so adding an `AuthConfig` variant fails the build until it gets an explicit redaction decision.
- Deliberately **not** applied to the config type's `Serialize`: persistence (`IcebergGsConfig::to_json`) and query-time `.resolve()` need the real secret, and that round-trip is covered by existing tests. Redaction lives at the emission boundary instead.

### Decimal-key FK detection (`fluree-db-r2rml/src/emit/{heuristic,input,mod,diagnostic}.rs`, `fluree-db-api/src/graph_source/iceberg_generate.rs`)
- FK candidacy + PK indexing accept scale-0 decimals (`is_key_type`); `key_types_match` treats scale-0 decimals as matching across precisions, integers by exact width, and **never** cross-matches int↔decimal.
- An exact `child == parent.pk` name+type match joins on name∧type alone when bounds are absent (Snowflake often supplies none); range-containment disambiguates when several same-named parents survive — and (review response) **vetoes** the single-match join when BOTH sides carry complete `[min,max]` bounds and containment provably fails (surfaced as `UnresolvedFkCandidate` with both ranges; never a silently fabricated dangling join). Decimal-string stat bounds are parsed.
- A `*_KEY`/`*_ID` column skipped for a non-key type emits a `NonKeyTypeSkipped` diagnostic instead of failing silently.

### Subject-key auto-fallback
- New `subject_strategy` (`auto` default / `identifier` strict) on `EmitOptions` + per-table override. Under `auto`, a table with an unverifiable-but-present key uses it with a downgraded `SubjectKeyUnverified` warning; a truly keyless table gets a deterministic **composite** subject + a `SubjectKeySynthesized` warning (never a fabricated rownum). Override + `identifier_field_ids` paths stay strict.
- (Review response) A not-provably-non-null key adopted under `auto` — whether a declared identifier OR a name-fallback column — is **never indexed as an FK parent** (`index_as_pk = false`): saveability needs a subject; an FK parent needs a provably-non-null key. The per-table `primary_key` override remains the explicit, still-indexing assertion.
- (Review response) The synthesized composite templates over the **provably-non-null** flat columns when any exist — R2RML template expansion emits NO subject for a row with a NULL in any referenced column, so an all-columns composite silently dropped exactly the rows keyless tables tend to have. When no non-null column exists, the all-columns fallback now **discloses the NULL row drop** in its diagnostic.
- The per-table override accepts a **composite** `subject_key: ["A","B"]`.

### Act-step engine caps (`iceberg_sample.rs`, `ephemeral.rs`, `send_parquet.rs`)
- `sample_iceberg_rows` / `sample_column_values`: first-N rows/values, bounded to the first surviving row group via a `max_rows` budget threaded through the Parquet decode — never a full-file scan. The cost unit is the row group: an all-columns sample of a wide table decodes every column chunk of row group 0 (documented; no silent column cap).
- `Fluree::query_provisional_r2rml(conn, mapping_ttl, sparql)`: compiles a candidate mapping in memory and runs real SPARQL against `(mapping + connection)` through the same R2RML operator the persisted path uses — no graph source, no nameservice record, nothing global to unwind; REST catalogs only.

### Virtual-query performance + crawl correctness (`fluree-db-query/src/r2rml/{operator,rewrite}.rs`, `fluree-db-api/src/graph_source/crawl.rs`, `graph_query_builder.rs`, `query/builder.rs`)
- Browse crawl = **one budgeted wildcard scan**: the operator emits each subject's `rr:class`-derived `rdf:type` rows on the wildcard itself (native parity — also fixes `<iri> ?p ?o` / `?s ?p ?o` never returning the type triple), and the crawl regroup folds them into `@type`. No separate `?s a ?type` pattern: that was a REQUIRED join (dropping subjects whose map declares no class) and, unfused, the topmost budgeted scan starving the wildcard.
- A templated/column (non-constant) predicate under a variable-predicate wildcard is materialized per-row into `?p` (or the POM is skipped when NULL) — never a solution with a bound object and an unbound predicate.
- Crawl interception covers every terminal that can see a graph-source alias: all three `FromQueryBuilder` terminals AND both alias-path terminals (`execute_formatted` + `execute_tracked` — tracked was falling through to native hydration and returning `[]`).
- `trust_fk_refs` (crawl-only): RefObjectMap objects render as templated parent IRIs from the child row's own FK columns — no parent table scans; byte-identical for matched rows; gated to the true-wildcard crawl shape only.
- Bound-subject scans prune to TriplesMaps whose subject-template constant prefix matches the IRI (16-table fan-out → point lookup); R2RML scan loops poll cancellation at window/file/scan granularity; crawl OFFSET paginates on grouped subjects (best-effort ordering, documented) with a 10k depth ceiling that errors with guidance instead of scanning O(offset) remotely.
- Trace spans (DEBUG, `fluree_db_*` targets): `r2rml.scan_table` / `r2rml.load_table` / `iceberg.oauth_token` (no secret fields) / `iceberg.scan_plan` / `iceberg.parquet_read`.

## New generate API surface (companion solo PR wires it)
- `subject_strategy` on the generate request options + per-table override (`"auto"` default | `"identifier"`).
- Composite `subject_key` on the per-table override (precedence over single `primary_key`).
- Two diagnostic codes the review UI should render (and treat unknowns gracefully): `nonKeyTypeSkipped`, `subjectKeySynthesized`.
- Public act-step entries: `sample_iceberg_rows`, `sample_column_values`, `query_provisional_r2rml`.

## Testing
`cargo fmt --all -- --check` clean · `cargo clippy -p fluree-db-r2rml -p fluree-db-query -p fluree-db-api --all-targets --no-deps` clean · tests at current head: **fluree-db-r2rml + fluree-db-query 1,413/1,413**, **fluree-db-api (--features iceberg) 2,894/2,894** · the `enterprise.ttl` golden is **byte-identical** (all-integer-keyed with required identifiers → exercises no new branch = the no-regression proof) · live suites (`it_iceberg_{sample,provisional,generate}_live`) are double-gated out of CI (`#[ignore]` + `FLUREE_ICEBERG_LIVE=1` + PAT + now `ICEBERG_CATALOG_URI` — no account default committed) · CI note: the first full run on the merged head caught exactly one failure (a sync test building a `Fluree` after main's builder began spawning the cache-event listener) — fixed by running it under `#[tokio::test]`.

## Review response (bplatz + internal pre-review)
All six inline comments addressed in code (rdf:type on both wildcard shapes; templated-predicate binding; offset was already fixed and now has a depth ceiling; the required type-join is gone entirely; alias-path tracked terminal wired; nullable name-fallback key no longer indexed as FK parent), plus the internal review's blocking item (synthesized-subject NULL drop) and its optionals (range-refutation veto, fail-closed redaction + AuthConfig backstop, best-effort OFFSET docs, env-only live-test catalog URI, sampler cost docs, per-table error context). Full item-by-item disposition: `pr-1450-feedback-assessment.md`. Not adopted, with reasons stated there: IndexMap byte-identity restoration (claim softened instead), `deny_unknown_fields` (forward-compat; the exhaustiveness-canary test provides the forcing function), generate partial-success (solo-coordinated contract change).

## Companion / note
A fluree/solo PR re-pins this and wires the new generate surface. **Security note:** the *live* PAT exposure users see is served by solo's own query-lambda `handle_info` fallback — it closes when solo re-pins (so its `ledger_info` call becomes virtual-aware + redacted) **and** redacts its remaining fallback + proxy. This PR closes every db-side path.

**Stacked-PR note:** #1476's `operator.rs` bound-subject rdf:type hunk is superseded by the general fix here and must be dropped on restack (its unit test remains valid); #1484 restacks routinely.


fluree-db-query/src/r2rml/operator.rs
```diff
@@ -1526,10 +1540,31 @@ fn materialize_batch(
...
+            // Pure subject-only pattern. A plain `?s a ex:Class` scan (constrained
+            // by `class_filter`) emits the subject alone. A projected `?s a ?type`
+            // scan (`type_var`) instead emits one row per class the map declares,
+            // binding `?type` to that class IRI — the same subjects a bound-class
+            // scan visits, with the class projected rather than filtered. A map
+            // that declares no class produces no row for a `type_var` pattern (its
+            // subjects have no rdf:type triple).
+            match pattern.type_var {
+                Some(tv) => {
+                    for class_iri in triples_map.classes() {
```
bplatz, line 1732: `rr:class`/`rdf:type` triples are only emitted on this subject-only branch. The variable-predicate wildcard path below iterates `predicate_object_maps` only, so `<iri> ?p ?o` / `?s ?p ?o` never returns the `rdf:type -> class` triple.

fluree-db-query/src/r2rml/operator.rs
```diff
@@ -1543,12 +1578,21 @@ fn materialize_batch(
...
                 }
+                // Variable-predicate wildcard (`?s ?p ?o` / `<iri> ?p ?o`): bind
+                // `?p` to this POM's predicate IRI. A templated (non-constant)
+                // predicate leaves `?p` unbound for that triple.
+                if let Some(pv) = pattern.predicate_var {
+                    if let Some(pred_iri) = pom.predicate_map.as_constant() {
+                        row.push((pv, Binding::iri(pred_iri)));
+                    }
+                }
                 row.push((obj_var, object_binding));
```
bplatz: For a templated (non-constant) predicate, `?p` is left unbound but `?o` is still pushed here, yielding a solution with a bound object and an unbound predicate.

fluree-db-api/src/graph_source/crawl.rs
```diff
@@ -0,0 +1,335 @@
...
+    // The selection must be exactly a wildcard `["*"]` — explicit predicate lists
+    // and nested ref-crawls are handled elsewhere (fall back).
+    let spec = spec.as_array()?;
+    if spec.is_empty() || spec.iter().any(|v| v.as_str() != Some("*")) {
+        return None;
+    }
+    let where_clause = obj.get("where")?;
+    let limit = obj
+        .get("limit")
+        .and_then(JsonValue::as_u64)
```
bplatz, line 456: Only `limit` is extracted here; `offset` is never read or propagated into the flat query, so a crawl with a non-zero `offset` silently returns the first page.

fluree-db-api/src/graph_source/crawl.rs
```diff
@@ -0,0 +1,335 @@
...
+    flat_limit: Option<usize>,
+) -> JsonValue {
+    let mut where_patterns: Vec<JsonValue> = match where_clause {
+        JsonValue::Array(patterns) => patterns.clone(),
+        other => vec![other.clone()],
+    };
+    // `?s ?__crawl_p ?__crawl_o` — every (predicate, object) of the subject.
+    where_patterns.push(json!({ "@id": subject_var, CRAWL_PRED: CRAWL_OBJ }));
+    // `?s a ?__crawl_type` — the subject's declared class(es).
+    where_patterns.push(json!({ "@id": subject_var, "@type": CRAWL_TYPE }));
```
bplatz: `?s a ?type` is appended as a required pattern, so subjects whose triples-map declares no `rr:class` are inner-joined out of the crawl entirely. Multiple classes also produce a predicate x type cartesian that consumes the row budget.

fluree-db-api/src/graph_query_builder.rs
```diff
@@ -231,6 +231,32 @@ impl<'a, 'g> GraphQueryBuilder<'a, 'g> {
...
+        // A subgraph ("crawl") projection over an R2RML graph source cannot use
+        // native binary-index hydration (the source has no flakes); expand it
+        // through the R2RML operator instead. Only a wildcard `["*"]` crawl is
+        // handled here — any other shape returns `None` and falls through to the
+        // normal path unchanged.
+        #[cfg(feature = "iceberg")]
+        if let (Some((provider, table_provider)), Some(json)) = (r2rml.as_ref(), input.as_jsonld())
+        {
+            if view.graph_source_id.is_some() {
+                if let Some(expanded) = crate::graph_source::crawl::expand_wildcard_crawl(
```
bplatz: This wildcard-crawl expansion only runs in `execute_formatted`. `execute_tracked` has no equivalent, so the same crawl issued with tracking headers falls through to native hydration and returns empty.

fluree-db-r2rml/src/emit/heuristic.rs
```diff
@@ -231,40 +238,151 @@ fn build_table_draft(
...
+                    table.qualified_name(),
+                    Some(col.name.clone()),
+                    format!(
+                        "subject key '{}' chosen by name fallback; NOT provably non-null \
+                         (no `required`, no null_fraction==0) and uniqueness is unverifiable \
+                         metadata-only — emitted so the table is saveable",
+                        col.name
+                    ),
+                ));
+                SubjectKey::single(col.name.clone())
```
bplatz: Worth reconsidering: `SubjectKey::single` sets `index_as_pk = true`, so under `Auto` this not-provably-non-null, uniqueness-unverifiable column becomes a live FK parent. The `Identifier` path returns `none()` for the same nullable case precisely to avoid that, and the composite-synthesis path emits a usable subject with `index_as_pk = false`. FK joins from other tables would then resolve against a column that may be null/non-unique, silently changing join cardinality with no diagnostic. Saveability only needs a subject, not an FK-eligible PK — consider `index_as_pk = false` here.

bplatz (approved):
Some comments noted that should be reviewed, approving to not delay once you get a chance to take a look.

fluree-db-query/src/r2rml/operator.rs
aaj3f: Confirmed, and fixed for BOTH shapes in 1c0064f2e: the POM loop now emits each subject's `rr:class`-derived `rdf:type` row for any true wildcard (variable predicate, no predicate filter, no projected type-var), so `<iri> ?p ?o` (the subject inspector) and `?s ?p ?o` both return the type triple — native parity. The browse crawl also benefits: its regroup folds the `rdf:type` rows into `@type`, which let us drop the injected `?s a ?type` pattern entirely (see the crawl.rs reply). Covered by `bound_subject_wildcard_emits_rdf_type`, `var_subject_wildcard_emits_rdf_type_rows`, and `engine_e2e_wildcard_binds_predicate_var` (now asserting 12 rows = 3 subjects × (3 POMs + 1 type)). Note for the stack: #1476 carried a bound-subject-only version of this fix; its operator.rs hunk gets dropped on restack (its test stays — it passes against the general fix).

fluree-db-query/src/r2rml/operator.rs
aaj3f: Right — that was a malformed solution (bound object, unbound predicate). Fixed in 1c0064f2e: a new `materialize_predicate_from_batch` (mirrors the subject/object batch helpers) expands a templated/column predicate from the row and binds it into `?p`; when a referenced column is NULL the POM is skipped for that row — the triple doesn't exist — so no half-bound row can be emitted. Constant predicates keep the zero-cost `as_constant()` fast path. Test: `templated_predicate_binds_or_skips_never_half_bound`. One pre-existing residual left as-is: a bound-predicate pattern (`?s <pred> ?o`) still only matches constant predicate maps — conservative incompleteness that predates this PR.

fluree-db-api/src/graph_source/crawl.rs
aaj3f: This one was fixed on the branch before your review landed (a5528e880): offset is extracted, the flat triple budget covers `offset + limit` subjects, and `skip(offset).take(limit)` applies after grouping — `crawl_wildcard_offset_paginates_subjects` pins non-overlapping pages. As review follow-up we also hardened it in 1c0064f2e: a paging ceiling (10k subjects) now errors with guidance instead of quietly launching an O(offset) remote scan, and the ordering comment is rewritten as best-effort (first-seen scan order; not enforced across separately-executed requests — keyset/cursor is the real answer if this becomes a first-class pagination surface).

fluree-db-api/src/graph_source/crawl.rs
aaj3f: Both halves were real, and worse than they looked: the type-var merge into the wildcard only fired when the crawl's WHERE bound a class, so a plain unfiltered browse crawl kept the standalone REQUIRED type-var scan — dropping class-less subjects AND sitting topmost on the LIMIT budget. Fixed structurally in 1c0064f2e: the Wildcard crawl no longer injects `?s a ?type` at all. With the operator now emitting `rr:class`-derived `rdf:type` rows on the wildcard itself (see the operator.rs thread), the crawl is a SINGLE budgeted scan; the regroup folds `rdf:type` rows into `@type`; class-less subjects survive (type simply absent); and the multi-class cost is +1 row per class per subject instead of ×classes per (p,o) row. All 19 crawl unit/e2e tests (multi-class, vertical-partition, subject parity, scan-count, offset, LIMIT) pass through the single-scan plan.

fluree-db-api/src/graph_query_builder.rs
aaj3f: Partially fixed before the review (d4d993071 wired all three `FromQueryBuilder` terminals), but the terminal your comment anchors on — the alias path's `execute_tracked` — was indeed still unwired. Completed in 1c0064f2e: it now intercepts identically and returns 200 + empty tracking (the crawl has no fuel/policy stats), mirroring `FromQueryBuilder::execute_tracked`. The snapshot builder is deliberately not wired: graph sources aren't loadable into materialized snapshots, so a graph-source crawl can't reach it.

fluree-db-r2rml/src/emit/heuristic.rs
aaj3f: Agreed and adopted verbatim in 49badb146. The asymmetry was backwards — the declared-identifier path (stronger evidence) refused to index a nullable key while the name-fallback path (weaker evidence, just a naming convention) indexed it. Now both adopt the column as the subject (`SubjectKeyUnverified`, table stays saveable/browsable) with `index_as_pk = false`; the diagnostic says so and points at the per-table `primary_key` override as the explicit, still-indexing assertion. A child whose FK now goes unresolved surfaces as `UnresolvedFkCandidate` instead of silently joining through a nullable/unverified parent key. No effect on the live demo dataset (its tables carry `identifier_field_ids`, so name-fallback never fires there).

aaj3f:
Thanks for the pass — all six comments were valid and all are now addressed in code (49badb146…40af2f624; see the per-thread replies): rdf:type emitted on both wildcard shapes, templated predicates bound-or-skipped (never half-bound), the crawl reduced to a single budgeted scan with no required type join, the alias-path tracked terminal wired for crawl interception, and the nullable name-fallback key no longer indexed as an FK parent.

The same wave lands the internal pre-review's items — most notably: the synthesized composite subject now prefers provably-non-null columns (an all-columns composite silently dropped every row with a NULL anywhere — exactly the rows keyless tables have) with disclosure otherwise; single-match FK joins are vetoed when both sides carry bounds that refute containment; the config redactor fails closed and gained an every-`AuthConfig`-variant sentinel backstop test with an exhaustiveness-canary `match` (a new variant fails the build until it gets an explicit redaction decision).

Deliberately NOT adopted, with reasons: IndexMap byte-identity for native `/info` (the "byte-unchanged" claim is softened to shape/value-identical instead — nested stats keys re-emit sorted, no consumer parses positionally), `deny_unknown_fields` on the auth config (persistence forward-compat; the canary test provides the same forcing function), and generate partial-success (a contract change needing solo coordination — the dominant per-table failure already names the table, and the two remaining error sites now do too).

CI's first full run on the merged head also caught one real merge-interaction failure — a sync test building a `Fluree` after main's builder began spawning the cache-event listener — fixed under `#[tokio::test]`. PR body refreshed to cover the five post-body commits (act-step caps + perf) and this wave.

Stack note: #1476's operator.rs bound-subject rdf:type hunk is superseded by the general fix here and gets dropped on restack (its unit test remains valid).

PR #1451, opened 2026-07-07
feat(iceberg): add AWS Glue Data Catalog + S3 Tables catalog modes

aaj3f:
## Summary

Adds two **native-AWS-SDK** catalog modes to the Iceberg/R2RML graph-source path — **AWS Glue Data Catalog** (`--mode glue`) and **AWS S3 Tables** (`--mode s3tables`) — at **full parity** with the existing REST (Polaris / Snowflake Horizon) and Direct modes.

```bash
fluree iceberg map gs --mode glue     --table enterprise_dw.dim_store --region us-east-1 --r2rml m.ttl
fluree iceberg map gs --mode s3tables --table ns.tbl --table-bucket-arn arn:aws:s3tables:... --region us-east-1 --r2rml m.ttl
```

## Approach

Both modes resolve the table's `metadata_location` through the native AWS SDK (`aws-sdk-glue` `GetTable` / `aws-sdk-s3tables` `GetTableMetadataLocation`), then read metadata/manifests/data from S3 with the **ambient** credential chain — reusing the existing `direct.rs` reader and `from_default_chain`. **No `aws-sigv4`, no request signing, no REST prefix.**

**Why SDK-native rather than REST + SigV4:** empirically, for a normally S3-authorized principal, neither Glue DC (IAM mode) nor S3 Tables vends credentials, so ambient reads suffice — and staying on the AWS SDK keeps consistent signing/transport paths (prefer the SDK / override its HTTP client via `aws-smithy-http` rather than roll alternatives) instead of a hand-rolled REST+SigV4 client.

## Changes

- **Core:** `GlueSdkCatalogClient` + `S3TablesSdkCatalogClient` (`catalog/aws_sdk.rs`); `CatalogConfig::Glue`/`S3Tables` (serde `"glue"`/`"s3tables"`, variant ungated / SDK impl `aws`-gated); two `r2rml.rs` scan arms; `CatalogMode` + CLI + server `map`; `aws-sdk-glue`/`aws-sdk-s3tables` deps; `read_glue` example.
- **Full parity:** catalog browse, table preview, R2RML auto-generate, and mapping validate now work for glue/s3tables (were hard-rejected). `catalog_client()` dispatches to the SDK clients; preview fetches table metadata from S3 when the catalog returns no inline metadata; server request struct + dispatch widened. Query-time: glue/s3tables scans use the same per-query snapshot pin + cross-query `loadTable` cache as REST (`sdk_resolve_load_table`).
- **Tests:** env-gated live integration tests (`it_glue_sdk_reads.rs`, `it_glue_live.rs`) + a CI-level `metadata_location`-extraction unit test.
- **`storage-credentials` parsing fix (was a listed follow-up):** `VendedCredentials::from_load_table_response` now parses the standardized top-level `storage-credentials` array (apache/iceberg #10722) with spec-normative precedence over the legacy flat `config` map (longest-*usable*-prefix; remote-signing-only entries skipped), and the two duplicated REST `loadTable` parse sites are factored into one shared `parse_load_table_response`. This was a latent bug that would 403 any REST catalog vending via `storage-credentials`. Not hit by this PR's SDK-native path (ambient by design); benefits spec-compliant REST vending catalogs (Polaris/Unity/future Snowflake).

## Validation

- **Offline:** config + `iceberg_catalog` + server-route + api-lib (689) + iceberg (202, incl. `storage-credentials` fixtures) suites pass; `fmt`/`clippy` clean.
- **Live** (Glue + S3 Tables harness): `read_glue` exact counts (Glue 500 / 60,000, S3 Tables 200); full SPARQL e2e over Glue exact (300 stores / 60,000 order-line⋈product / 6 categories / 20,000 order⋈customer) incl. multi-table joins + a dangling-FK check; browse + preview live; the env-gated `it_glue_*` tests assert Glue 300 / S3 Tables 200 / browse / preview.

## Follow-up

- **#1456** — SDK-native Lake-Formation vended credentials for LF-governed Glue tables read by a *restricted* principal (the remaining "vended AWS" case; ambient reads 403 there). Filed as a de-risked spec — fail-closed on fine-grained LF policies, hybrid-catalog try-ambient-then-vend, `VendedS3Path` scope enforcement — produced by this PR's adversarial audit. S3 Tables LF-vending is deferred within that issue (managed storage needs the Glue federation integration).

### DIFF
--- a/fluree-db-api/src/graph_source/r2rml.rs
+++ b/fluree-db-api/src/graph_source/r2rml.rs
@@ -142,6 +142,116 @@ fn build_iceberg_filter(
     }
 }
 
+/// Which AWS-SDK catalog to build on a `loadTable` cache miss.
+enum SdkCatalog {
+    Glue {
+        region: Option<String>,
+        catalog_id: Option<String>,
+    },
+    S3Tables {
+        region: Option<String>,
+        table_bucket_arn: String,
+    },
+}
+
+impl SdkCatalog {
+    fn label(&self) -> &'static str {
+        match self {
+            SdkCatalog::Glue { .. } => "AWS Glue Data Catalog",
+            SdkCatalog::S3Tables { .. } => "AWS S3 Tables",
+        }
+    }
+
+    async fn build(&self) -> QueryResult<Box<dyn SendCatalogClient>> {
+        Ok(match self {
+            SdkCatalog::Glue { region, catalog_id } => Box::new(
+                fluree_db_iceberg::catalog::GlueSdkCatalogClient::new(
+                    region.as_deref(),
+                    catalog_id.clone(),
+                )
+                .await
+                .map_err(|e| {
+                    QueryError::Internal(format!("Failed to create Glue catalog client: {e}"))
+                })?,
+            ),
+            SdkCatalog::S3Tables {
+                region,
+                table_bucket_arn,
+            } => Box::new(
+                fluree_db_iceberg::catalog::S3TablesSdkCatalogClient::new(
+                    region.as_deref(),
+                    table_bucket_arn.clone(),
+                )
+                .await
+                .map_err(|e| {
+                    QueryError::Internal(format!("Failed to create S3 Tables catalog client: {e}"))
+                })?,
+            ),
+        })
+    }
+}
+
+/// Resolve an AWS-SDK catalog's `loadTable` with the same per-query snapshot pin
+/// (one Iceberg snapshot for the whole query — a correctness guarantee the raw
+/// per-scan SDK path lacked) and cross-query `loadTable` cache the REST path uses.
+/// Glue/S3Tables return `credentials: None`, so the credential-expiry gating is a
+/// no-op. The SDK client is built only on a genuine miss, so warm scans skip the
+/// per-scan credential-chain resolution + `GetTable`/`GetTableMetadataLocation`.
+async fn sdk_resolve_load_table(
+    session: &super::catalog_session::IcebergCatalogSession,
+    cache: &R2rmlCache,
+    graph_source_id: &str,
+    table_id: &fluree_db_iceberg::catalog::TableIdentifier,
+    kind: &SdkCatalog,
+) -> QueryResult<fluree_db_iceberg::catalog::LoadTableResponse> {
+    use super::catalog_session::{CachedLoadTable, IcebergCatalogSession};
+    let lt_key = IcebergCatalogSession::load_table_key(
+        graph_source_id,
+        &table_id.namespace,
+        &table_id.table,
+    );
+
+    if let Some(cached) = session.cached_load_table(&lt_key) {
+        debug!(namespace = %table_id.namespace, table = %table_id.table,
+            "loadTable pin hit (query-scoped)");
+        return Ok(cached);
+    }
+    let pinned = session.pinned_metadata_location(&lt_key);
+    let cross_query = if pinned.is_none() {
+        cache.get_rest_load_table(&lt_key)
+    } else {
+        None
+    };
+    let mut resp = if let Some(cq) = cross_query {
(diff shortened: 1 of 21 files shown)


fluree-db-iceberg/src/catalog/aws_sdk.rs
```diff
@@ -0,0 +1,290 @@
...
+            .database_list()
+            .iter()
+            .map(|d| d.name().to_string())
+            .collect())
+    }
+
+    async fn list_tables(&self, namespace: &str) -> Result<Vec<String>> {
+        let out = self
+            .client
+            .get_tables()
```
bplatz, line 106: These four list ops (`get_databases`/`get_tables` here, and `list_namespaces`/`list_tables` in the S3 Tables client) only read the first page. All are paginated — Glue defaults to 100 results + `next_token`, S3 Tables returns a continuation token — so browse silently truncates any catalog with >100 databases/tables. Loop on the token or use `.into_paginator().send()`.

fluree-db-api/src/graph_source/r2rml.rs
```diff
@@ -142,6 +142,116 @@ fn build_iceberg_filter(
...
+            SdkCatalog::Glue { .. } => "AWS Glue Data Catalog",
+            SdkCatalog::S3Tables { .. } => "AWS S3 Tables",
+        }
+    }
+
+    async fn build(&self) -> QueryResult<Box<dyn SendCatalogClient>> {
+        Ok(match self {
+            SdkCatalog::Glue { region, catalog_id } => Box::new(
+                fluree_db_iceberg::catalog::GlueSdkCatalogClient::new(
+                    region.as_deref(),
```
bplatz, line 169: The catalog client is built with only the mode `region`, but the S3 storage in the same scan arm (and the browse path in `catalog_client`) falls back `region → io.s3_region`. If a user sets only `io.s3_region`, `GetTable` resolves via the default-chain region while S3 reads use `io.s3_region` — mismatched, and it contradicts the `CatalogConfig::Glue.region` doc ("falls back to ... io.s3_region"). Thread `io.s3_region` into `SdkCatalog` so the fallback matches.

fluree-db-iceberg/src/credential/vended.rs
```diff
@@ -119,6 +119,65 @@ impl VendedCredentials {
...
+                        best = Some((len, creds));
+                    }
+                }
+            }
+            if let Some((_, creds)) = best {
+                return Ok(Some(creds));
+            }
+        }
+
+        // Fall back to the legacy flat top-level `config` map.
```
bplatz, line 172: Edge case in this fallback: if the only prefix-matching `storage-credentials` entry is remote-signing-only (no static `s3.*` keys) and a legacy top-level `config` carries static creds, we skip the matching entry and fall back to `config`'s static creds — which can contradict the catalog's intent to remote-sign that prefix. Rare shape, likely fine, but worth a note on the precedence.

bplatz (approved):
Several comments inline, approving so you don't have to wait after reviewing them.

aaj3f:
Backlog-sweep nudge: this has been approved since 7/7 but is now conflicting and ~7 weeks stalled, and #1456 (Lake-Formation vended credentials) is explicitly unstartable until it lands. The rebase is over the BYO-IAM credential work (#1500/#1503/#1514, e.g. `210cf4833`) — and for what it's worth, LF vending composes with the STS `AssumeRole` grant rail that landed there (`fluree-db-api/src/vended_credentials.rs`) rather than being replaced by it, so I don't think the stall has invalidated the design.

One thing the sweep flagged that raises the stakes a bit: the `storage-credentials`-array parse fix that #1456 describes as "resolved in #1451" is not on main (`rest.rs` is still `from_config_map`-only) — because this never merged. So that fix either lands with the rebase or needs splitting out and tracking on its own; right now it's in limbo where the issue tracker says it's handled and the code says otherwise.

Issue #1319, opened 2026-06-12
Integer literals from VALUES/expressions compare as xsd:long while stored integers are xsd:integer — same number is two distinct terms


bplatz:
## Summary

A plain integer literal in a query (`VALUES ?n { 5 }`) lowers to a binding with datatype **xsd:long**, while the same integer stored in a ledger materializes from the index as **xsd:integer**. Since term equality includes the datatype, the same number arriving from the two sources is treated as two distinct RDF terms.

## Repro

On any ledger containing `ex:a ex:n 5`:

```sparql
PREFIX ex: <http://example.org/ns/>
SELECT DISTINCT ?n { { ex:a ex:n ?n } UNION { VALUES ?n { 5 } } }
```

returns **two** rows:

```json
[{"n": {"type":"literal","value":"5","datatype":".../XMLSchema#integer"}},
 {"n": {"type":"literal","value":"5","datatype":".../XMLSchema#long"}}]
```

Per SPARQL, the bare literal `5` is `"5"^^xsd:integer`, so this should dedup to one row. The same mismatch makes joins/MINUS/GROUP BY on integer values miss across the two sources.

## Notes

- Independent of binding representation (encoded vs decoded) — it reproduces decoded-vs-decoded, e.g. against novelty-only data. Discovered while building `fluree-db-api/tests/it_mixed_representation.rs` (see the NOTE comment there excluding the integer case).
- Value-space comparison paths (FILTER `=`, numeric operators) are unaffected; this is about *term* identity surfaces: DISTINCT, GROUP BY keys, COUNT(DISTINCT), join unification, MINUS.
- Needs a canonicalization decision rather than a point fix: either lower bare integer literals to xsd:integer end-to-end, or canonicalize the int64 datatype family at the term-identity surfaces.

aaj3f:
See also #1343 (that issue closed as a duplicate of this one)

aaj3f:
[...]

PR #1452, opened 2026-07-07
test(sparql): burn-down Wave 0 — harness hygiene, parser accept-more, lexer/formatter, expression high-yield (−80 registers)

### LINKED ISSUE #1439: Variable-free FILTER eliminates every solution (FILTER(true) returns zero rows)
`FILTER(true)`, `FILTER(1=1)`, `FILTER(2 IN (1,2,3))` return **zero rows** through the public query API; the same filter with any variable reference works. Root cause: `required_vars = referenced_vars()` is empty, so the filter is "eligible" before any triple binds and is inlined against the seed row (`fluree-db-query/src/execute/where_plan.rs:697,:990`).

Silent wrong results on both the SPARQL and JSON-LD surfaces (shared IR). Planner-only fix: evaluate constant filters once per stream. Roadmap PR-X1 (D1).

### LINKED ISSUE #1440: DATATYPE()/LANG() reject non-variable arguments
`eval/rdf.rs:58-62` bails unless the argument is a bare `Expression::Var`; `LANG` (`eval/string.rs:115-134`) silently returns `""` for expression arguments. `FILTER(datatype(?a + ?b) = xsd:integer)` and `datatype("foo"@en)` fail. 29 W3C tests and any user composing expressions inside these builtins are affected.

Fix: evaluate the argument to a value, read its datatype/lang, type-error on non-literals — SPARQL-scoped per decision D-12 (the `@id` extension on the JSON-LD surface is deliberate and stays). Roadmap PR-X1 (D2/D2b).

### LINKED ISSUE #1444: Turtle import rejects _:label. (blank-node label immediately followed by statement dot)
`parse_blank_node_name` (`fluree-graph-turtle/src/lex/lexer.rs:477-490`) greedily consumes a trailing `.` into the label, then hard-errors, rejecting valid Turtle (`_:o6.`) that every other store ingests (per Turtle spec, PN_LOCAL must not end with a dot; the dot is the statement terminator). Blocks the four W3C json-res tests and any user data of this shape.

Fix constraint from adversarial review: use greedy-scan + trailing-dot **rewind**, not a lookahead loop — spec-valid `_:a..b` currently lexes and must remain byte-identical. Import-hot path: `insert_formats` + `import_bulk` benches gate the change. Roadmap PR-L1.

### LINKED ISSUE #1445: xsd:double serialization is non-canonical and inconsistent across output formats
Three different lexical forms today: SPARQL-JSON/XML render Rust `Display` (`"1000000"`), native CSV/TSV renders ryu (`"1000000.0"`), N-Triples/N-Quads export renders `{:E}` (`"1E6"` — missing the mandatory mantissa dot). The XSD canonical form is `"1.0E6"`.

Fix: one shared `canonical_xsd_double` helper routed through `sparql.rs`, `sparql_xml.rs`, `delimited.rs`, `LiteralValue::lexical()`, and `export.rs:1236/:1465`; JSON-LD numeric output stays a native JSON number (deliberate). User-visible output change — changelog note for downstream string-matchers; R2RML/Iceberg `lexical()` consumers flagged. Roadmap PR-L2.

aaj3f:
## Reviewer quick reference

**End-user API/contract changes:** (a) `xsd:double` RDF-lexical output is now the XSD canonical form (`"1000000"`/`"1E6"` → `"1.0E6"`) across SPARQL-JSON/XML, CSV/TSV, Turtle/N-Triples export, and `LiteralValue::lexical()` — downstream string-matchers must adjust (changelog note below); JSON-LD numeric output is deliberately unchanged. (b) SPARQL `DATATYPE()`/`LANG()` now accept expression arguments (additive), and `DATATYPE` on a non-literal is a type error **on the SPARQL surface only** (see Decisions). (c) Everything else in this wave is accept-more parsing or bug fixes — no previously-working query changes meaning.

**Performance verdict:** no hot-path regressions anywhere; all standard gates within `regression-budget.json`. One **must-accept output cost**: correct zero-length property-path answers are strictly larger result sets on shapes that previously returned wrong (incomplete) answers — details and numbers in Performance Notes below. This wave also *adds* a new pinned bench (`query_hot_property_path`).

**Production bugs found & fixed along the way** (not just W3C compliance): `FILTER(true)`/any variable-free filter silently returned zero rows (#1439); `DATATYPE()`/`LANG()` rejected composed expressions (#1440); Turtle import rejected valid `_:label.` (#1444); double serialization was non-canonical and internally inconsistent across formats (#1445); bare integers in VALUES minted `xsd:long` term identities (#1319 mechanism); and the W3C harness's own `.rdf` result parser dropped *bound* solutions on self-closing XML elements while keeping unbound ones.

**D-12 (scoping of strict SPARQL type errors) — actioned by the expression component (pr-x1, D2b).**
- *Decision point:* when `DATATYPE()` is applied to a non-literal (IRI/bnode), SPARQL 1.1 requires a type error; Fluree's JSON-LD surface deliberately returns the `@id` extension value instead.
- *Options considered:* (1) SPARQL-scoped strict errors, JSON-LD keeps its extension — **chosen**; (2) unified strictness across both surfaces — rejected: breaks a committed, test-pinned JSON-LD product behavior for zero correctness gain on that surface; (3) keep lenient behavior everywhere — rejected: fails W3C and leaves SPARQL spec-incorrect.
- *Why (1):* the surfaces share the IR but not the contract; the parity guideline gets an explicit recorded carve-out, and both behaviors are pinned by tests so neither can drift silently. (Full table: `docs/audit/burn-down/ROADMAP.md` §4.)

**D-13 (surfaced here, deliberately NOT actioned):** the collections work revealed that Turtle *ingest* stores object-position RDF collections as native `list_index` items — `rdf:first/rest` triples never exist in storage, so `basic#list-1..4` cannot green from parser work. Materialize-at-ingest vs translate-at-query vs documented divergence is an open team decision; the register entries say exactly this.

---

## Wave 0 of the W3C SPARQL burn-down (6 components, ≈−80 register entries)

First implementation wave of the burn-down plan in `docs/audit/burn-down/ROADMAP.md` (see PR #1437 for the enforcement model). Every component was implemented against its adversarially-verified cluster audit, is individually suite-green, and was integrated here with register conflicts resolved to the mechanical fixpoint dictated by the both-way CI gate. **Zero-to-low engine risk by design: this wave is harness fixes, parser accept-more, lexer/formatter corrections, and expression fixes on cold/erroring paths.** All bench gates within `regression-budget.json` budgets.

### Components

**1. Harness + register hygiene (`pr-h1`, −3).** Fixes the harness's `.rdf`/SRX result parsers: self-closing `Event::Empty` elements skewed the state stack (keeping unbound solutions and dropping bound ones — the mechanism was verified live against `result-sort-3.rdf`), plus missing `rdf:nodeID` blank-node value support. Applies the full register-comment correction batch so every entry names its root cause and owning future PR. Also commits the standalone-`[workspace]` marker for the harness (root path-exclusion doesn't reach git worktrees).

**2. Parser accepts valid syntax (`pr-1`, −29).** RDF collections `( … )` in triple patterns (desugared to `rdf:first/rest/nil`, guarded out of quoted-triple contexts per the RDF 1.2 grammar's negative tests), empty IRIREF `<>`, function-call `NIL` arglists, bare ORDER BY constraints, sub-SELECT placement, `VALUES ()`, property paths inside blank-node property lists, and a SPARQL 1.2 codepoint-escape pre-pass (`Cow` fast path — zero alloc when no escapes). **Notable finding:** `basic#list-1..4` do NOT green — Turtle *ingest* stores object-position collections as native `list_index` items, never `rdf:first/rest` triples; registered as decision D-13. Its P6 sub-SELECT work was later subsumed by main's #1436 and dropped cleanly. JSON-LD parity: `@list` ↔ first/rest equivalence test.

**3. Turtle bnode-dot lexer (`pr-l1`, −5).** `_:o6.` (blank-node label + statement dot) failed to lex (#1444). Fix is greedy-scan + trailing-dot **rewind** — the naive lookahead alternative was rejected because it broke spec-valid `_:a..b` (adversarial-verification catch, ROADMAP §1.1-9). Import-hot path: interleaved ABAB benches at n=100 (single runs measured ±11–15% noise on identical binaries); all within budget. Bonus green: entailment `owlds02` (its data was `_:x.`-shaped).

**4. Canonical `xsd:double` (`pr-l2`, −1).** One shared zero-alloc `canonical_xsd_double` helper (`fluree-graph-ir`), routed through every RDF-lexical site: SPARQL-JSON/XML, CSV/TSV (drops the `ryu` dep), `LiteralValue::lexical()`, and the N-Triples/N-Quads export sites that emitted dotless `1E6` (#1445). **Behavior change / changelog:** RDF-lexical double output is now the XSD canonical form (`1.0E6`); JSON-LD numeric output deliberately untouched (native JSON numbers) with a guard test. R2RML inherits the canonical form (spec-correct; flagged for that workstream).

**5. Expression high-yield fixes (`pr-x1`, −37).** Six per-defect commits, each individually suite-green: variable-free FILTER dropped ALL rows (`FILTER(true)` → 0 results, #1439 — plan-time fix, var-full paths byte-identical); `DATATYPE()`/`LANG()` rejected expression arguments (#1440); xsd:dateTime/date/time casts; `BNODE(label)` per-solution identity; regex `q` flag; bare integers in VALUES lowered to `xsd:integer` (#1319 — at lowering, `dt_compatible` untouched per the verified mechanism). **Decision transparency (D-12):** the DATATYPE-on-non-literal type error is SPARQL-scoped only; the JSON-LD `@id` extension behavior is deliberate and pinned by parity tests (options and rationale in ROADMAP §4). Fixing D2 unmasked the deeper D4 promotion defect on `tP-29/30` — honestly registered for PR-X2. BSBM benches at/inside the noise floor.

**6. Property-path fixes (`pr-pp`, −2).** pp16: the zero-length-path universe now spans ALL graph terms including literals (SPARQL §18.4), gated at plan level on the `ZeroOrMore`/`ZeroOrOne` modifiers — `*`/`+` traversal untouched; pp36: `empty_schema_with_len` guard for zero-column existence rows. Adds the `query_hot_property_path` bench + `regression-budget.json` entries (reconcile test passes). The slower closure numbers are spec-mandated larger output on previously-wrong shapes; `seq_plus` and BSBM unchanged.

### Verification
- Full W3C suite on the integrated branch: **36/36 suites, 0 failed, 0 stale.**
- `fluree-db-sparql` 601 unit tests; JSON-LD groups (`grp_query`, `grp_query_sparql`, `grp_transact`) green; fmt + clippy `-D warnings` clean.
- Integration-fixpoint commits document every register resolution (union-then-prune driven by the both-way gate).

Issues closed by this wave when merged: #1439, #1440, #1444, #1445 (+ #1319 mechanism half).

---

**Harness/registers (pr-h1):** harness-only; no engine code.

**Parser accept-more (pr-1):** 100% parse-time; zero runtime code touched. The SPARQL 1.2 codepoint-escape pre-pass uses a `Cow` fast path — zero allocation and a single scan when a query contains no escapes (the overwhelmingly common case).

**Turtle bnode-dot lexer (pr-l1) — import-hot path:** the fix deliberately keeps the existing greedy per-character scan byte-identical (no added lookahead, no backtracking); trailing dots are handled by a post-scan rewind, one extra check per *label*, not per character. Measurement was rigorous because single runs proved noise-dominated (±11–15% measured on *identical binaries*): interleaved ABAB at n=100 gave `import_bulk` +1.0%/−9.2% (noise band), turtle 100-nodes −0.3%, turtle 10-nodes ~+2.1% attributable to code layout — the changed function is *unreachable* by the bench corpus, which contains no `_:` tokens (corpus gap flagged for the bench backlog). All within the 5% budgets.

**Canonical double (pr-l2):** formatter-layer only, no query-path involvement. The shared helper post-processes `{:e}` in a fixed stack buffer (zero heap allocation); the CSV/TSV writer gained a write-into-buffer variant and *dropped* the `ryu` dependency.

**Expression high-yield (pr-x1):** the variable-free FILTER fix (D1) is plan-time classification in the planner — variable-bearing filter paths are byte-identical; all other defects are on cold or previously-erroring evaluation paths. BSBM before/after (quick profile, declared): q3 +1.9–3.4%, q5 −1.6–−0.3%, q9 +2.0–3.9%, bi/f2 +1.1–2.2% — same-code reruns on this box swing ±1–4%, so deltas are at/inside the noise floor and within the small-budget; the nightly full-profile bench is the backstop. Each defect landed as its own suite-green commit for bisectability.

**Property paths (pr-pp):** the `*`/`+` traversal fast path is untouched; the zero-length term universe is selected at plan level, only for the SPARQL `ZeroOrMore`/`ZeroOrOne` shapes, and pair buffers carry full terms only on that path. **Must-accept cost, stated plainly:** star_closure 2.37→6.06 ms and zero_or_one 1.69→6.27 ms in the new bench — these are NOT like-for-like comparisons: the pre-fix numbers computed *incomplete (wrong) results*, and SPARQL §18.4 requires every graph term (including literals) as a zero-length self-pair, so the correct output is strictly larger on both-ends-unbounded queries. The hot signals hold: seq_plus 43.50→43.87 ms (noise), `query_hot_bsbm` unchanged (q3 0.92 / q5 1.42 / q9 2.06 ms). This PR *adds* `query_hot_property_path` + `regression-budget.json` entries, so path-traversal performance is pinned from now on (`workspace_reconcile` passes).

## Issues closed on merge

Closes #1439
Closes #1440
Closes #1444
Closes #1445

Also fixes the reported mechanism of #1319 (bare VALUES/inline integers now lower to `xsd:integer`; the separate scan-path datatype-constraint question stays tracked there — left open deliberately rather than auto-closed).

*Mechanics note: GitHub fires these keywords when the PR merges into the default branch. In the stacked flow (#1437 merges first, this PR is then auto-retargeted to `main`), that happens exactly when these fixes actually reach `main`.*


fluree-db-query/src/property_path.rs
```diff
@@ -569,10 +579,97 @@ impl PropertyPathOperator {
...
+    /// touched by predicates outside the path. The `f:reifies` sidecar is
+    /// skipped (an implementation detail hidden from variable-predicate
+    /// reads); `rdf:type` triples are real data and are included.
+    ///
+    /// Runs one policy-filtered full scan of the active graph. Only reached
+    /// for both-variable unbounded `*`/`?` paths — already a full-closure
+    /// enumeration — never for the bound-endpoint traversals.
+    async fn zero_length_universe(&self, ctx: &ExecutionContext<'_>) -> Result<Vec<Binding>> {
+        let (db, overlay, to_t) = ctx.require_single_graph()?;
+        let flakes = range_with_overlay(
```
bplatz, line 650: `zero_length_universe` runs an unbounded full-graph SPOT scan with no `max_visited`/`ResourceLimit` bail — unlike every other traverse/closure path in this file — and it runs *before* the BFS, so that guard is bypassed. It fires for any both-unbound `*`/`?` regardless of path selectivity: `?x :rareProp* ?y` where `:rareProp` has 2 edges now reads the whole index into memory and emits ~(#distinct terms) self-pair rows. The result is §18.4-correct, but this needs a resource bound (and ideally to scope the scan to terms actually reachable by the path).

fluree-db-api/benches/query_hot_property_path.rs
```diff
@@ -0,0 +1,203 @@
...
+//! Mirrors `query_hot_bsbm.rs`: build once per scale, populate a
+//! file-backed ledger, full reindex behind the binary columnar index,
+//! then reuse the `GraphSnapshot` for all `b.iter` calls (warm-cache).
+//!
+//! ## Matrix
+//!
+//! The graph is `n_clusters` disjoint 8-node `ex:link`/`ex:p2` chains
+//! (cluster-bounded closures keep output linear in scale, so the bench
+//! measures per-node traversal cost, not a quadratic pair blow-up).
+//! Every node also carries one literal (`ex:name`), so the zero-length
```
bplatz, line 35: Coverage gap: every node carries an `ex:link` edge, so the zero-length universe ≈ the path domain — this bench never exercises a selective predicate over a large graph, the exact shape where the `zero_length_universe` scan amplifies. Combined with `query_hot_property_path` being budgeted looser than its siblings in `regression-budget.json`, the zero-length slowdown is absorbed rather than gated.

fluree-db-sparql/src/parse/query/modifier.rs
```diff
@@ -287,8 +287,20 @@ impl super::Parser<'_> {
...
-            // Not a valid order condition
-            return None;
+            // Bare Constraint form: `OrderCondition ::= … | ( Constraint | Var )`
+            // where `Constraint ::= BrackettedExpression | BuiltInCall |
+            // FunctionCall` — so `ORDER BY str(?o)` (no parentheses) is valid.
+            // Mirrors `parse_group_condition`'s bare-call branch: the call is
+            // self-delimiting, and a non-expression token (LIMIT / OFFSET /
+            // VALUES, `}`, EOF, …) ends the condition loop via restore.
+            let pos = self.stream.position();
+            match parse_expression(self.stream) {
```
bplatz, line 328: This bare-condition branch calls full `parse_expression`, but the grammar restricts the bare form to `Constraint | Var` (`Constraint = BrackettedExpression | BuiltInCall | FunctionCall`). `ORDER BY ?x - 1` parses as two keys — `ORDER BY ?x, (-1)` — silently dropping the arithmetic (the `-1` key is a no-op); same for `ORDER BY ?x + ?y`. These were parse errors before this PR; now they're silently-wrong sorts. Suggest restricting the bare form to a Constraint (bracketed expr / BuiltInCall / FunctionCall) rather than a full expression.

fluree-db-sparql/src/lex/lexer.rs
```diff
@@ -39,9 +39,15 @@ impl<'a> Lexer<'a> {
     ///
     /// Comments are identified by the lexer itself, so `#` characters inside
     /// string literals or IRIs are never treated as comments.
+    ///
+    /// Runs the SPARQL 1.2 codepoint-escape pre-pass first (§19.2:
+    /// `\uXXXX`/`\UXXXXXXXX` are unescaped over the whole query string
+    /// before tokenizing), so token spans index into the unescaped string
+    /// when escapes are present.
     pub fn tokenize_collecting(self, comments: &mut Vec<String>) -> Vec<Token> {
+        let unescaped = unescape_codepoints(self.input);
```
bplatz, line 48: Two issues with running the codepoint pre-pass here, before comment stripping:

1. Escapes inside `#` comments are decoded — a `\u000A` in a comment splits the line and can reject a previously-valid query (Jena/rasqal don't decode in comments).
2. An encoded backslash (`\u005C`) emits a real `\` that the downstream string/IRI lexer then re-interprets: `<http://ex/\u003E>` decodes to `<http://ex/>>`, truncating the IRI at the injected `>`. This may be the intended SPARQL 1.2 "as if literal" behavior, but it isn't pinned by a test.

bplatz (approved):
A few comments, approved pending your review.

fluree-db-query/src/property_path.rs
aaj3f: Confirmed — the scan had no cancellation poll and no bail, unlike every other loop in the operator, and it fired for selective predicates too. Fixed by polling `bail_if_cancelled` in the scan loop, so a large `?s :p* ?o` is now interruptible by the query timeout like the traversal paths.

I deliberately did *not* add a hard `max_visited` term cap. The zero-length universe is inherently O(graph): §18.4 (the pp16 fix) requires every graph term as a self-pair, including terms touched only by non-path predicates and literals — so a term cap would reject spec-valid queries on modestly large graphs (`max_visited` is always 10k and is a traversal-depth bound; the existing `star_closure` bench alone emits ~32k universe terms at medium). Time-based cancellation is the right axis for a full-graph enumeration. On scoping the scan to path-reachable terms: that would reintroduce the pp16 incompleteness, so I kept it full-graph. One honest residual — the `range_with_overlay` read still materializes the scan up front (same as the wildcard closure), so true input-streaming is a separate, broader change, noted in the doc comment.

fluree-db-api/benches/query_hot_property_path.rs
aaj3f: Agreed on both. The `ex:link`/`ex:p2`-everywhere graph made the zero-length universe scale with the path domain, so no scenario exercised a selective predicate over a large graph — exactly where the universe scan dominates. Added a `rare_star` scenario (`?s ex:rare* ?o` with a single fixed `ex:rare` edge): trivial closure, but the universe still scans every graph term, isolating that cost. Tightened the `query_hot_property_path` medium budget 5.0 to 3.0 to match the bsbm siblings. (Both numbers are still the acknowledged placeholders pending the first nightly baseline — the calibration follow-up will tune them.)

fluree-db-sparql/src/parse/query/modifier.rs
aaj3f: Correct — the bare branch used full `parse_expression` where the grammar's bare `OrderCondition` is `Constraint | Var` only. Restricted it (and `parse_group_condition`, same grammar restriction) to a call-shaped Constraint via `is_bare_constraint`, so `str(?o)` / `ex:fn(?o)` stay valid but `ORDER BY ?x - 1` no longer absorbs the `- 1` as a spurious second key. Added parser regression tests.

One nuance worth flagging: the parser doesn't currently reject trailing tokens after a complete query, so `ORDER BY ?x - 1` now parses as `ORDER BY ?x` with the `- 1` ignored, rather than a hard error (a *leading* bare arithmetic like `ORDER BY str(?x) - 1` does hard-error, since no condition parses). That removes the silent mis-sort you flagged; making trailing garbage itself a hard error is a broader, higher-blast-radius change across every query form, so I kept this scoped and left trailing-token strictness as a separate item. W3C syntax suites stay green with no register churn.

fluree-db-sparql/src/lex/lexer.rs
aaj3f: Dug into this against the W3C codepoint tests, and I think the pre-pass is behaving as the spec requires — with two genuinely-unpinned corners that I've now pinned.

On running before comment stripping: it has to be global-before-tokenizing. codepoint-esc-01 encodes an entire `ASK {}` in escapes, esc-07 uses escaped whitespace to separate the projection tokens, and esc-08 uses `\u0022` to open a string literal — none of those pass unless escapes resolve over the whole string before the tokenizer (and comment stripping is part of the tokenizer). So the comment-agnostic behavior is a consequence of that requirement rather than an independent bug. You're right that it diverges from engines that treat comments as opaque; I've documented it as a deliberate, known divergence and added a unit test so it can't drift.

On `<http://ex/\u003E>`: I believe the truncation is correct, by the same rule esc-08 pins — an escape that decodes to a delimiter acts as that delimiter (esc-08's `\u0022` opens a string; `\u003E` produces the `>` that closes the IRI). It's just a corner the suite doesn't cover, so I added a test pinning it. If we ever decide the IRI case should differ from the string case we'd need a spec basis, but today they're the same rule.

Net: no behavior change, two new pinning tests, and the pre-pass doc now spells out both corners.

PR #1453, opened 2026-07-07
feat(sparql): burn-down Wave 1 — validation passes, grammar tightening, authoritative diagnostics, reifier forms (−121 net registers)

### LINKED ISSUE #1438: SPARQL UPDATE: multi-operation (;) requests silently execute only the first operation
`parse_sparql` returns after parsing one operation with no trailing-token check (`fluree-db-sparql/src/parse/query/mod.rs:40-74,193-207`); `parse_and_lower_sparql_update` lowers that single body. A request like `INSERT DATA {...} ; DELETE WHERE {...}` commits the INSERT and **silently discards everything after the first `;`** — verified end-to-end through `graph().transact().sparql_update().commit()` with readback (W3C `basic-update#insert-05a` is a 5-op request that demonstrates it).

This is silent data loss for any client sending standard multi-op SPARQL UPDATE.

**Interim mitigation (decision D-10a, approved):** reject requests with trailing tokens with a loud parse error. **Full fix:** request-level sequential staging of ops within one atomic commit (roadmap PR-U2, gated on the Txn IR shape decision D-10b).

aaj3f:
## Reviewer quick reference

**End-user API/contract changes — this is the reject-more wave; read this list carefully.** SPARQL inputs that previously "worked" and now error (each spec-mandated, each with a changelog/migration note in its section): (a) `GROUP BY (expr)` with the raw expression reprojected — the field/#1362 shape — must alias the key (`GROUP BY (expr AS ?k)`); (b) Fluree's grouped-list projection extension is removed **from SPARQL only** (JSON-LD keeps it, parity-pinned); (c) BIND-scope, duplicate-alias, bnode-cross-scope, and GROUP-projection violations now reject; (d) stray/doubled dots and bare FILTER terms now reject (V1 is the highest-regression-risk item — users may rely on dot leniency; isolated commits for bisect/revert); (e) **parse errors are now authoritative at the API**: a query whose parse produced error diagnostics no longer executes a recovered AST; (f) multi-operation UPDATE requests error **loudly** instead of silently executing only the first operation (interim guard; real support lands in Wave 2); (g) blank nodes in DELETE forms are rejected on both the SPARQL and JSON-LD transact surfaces (`_:fdb-` stable ids exempt).

**Performance verdict:** entirely parse/validate-time and update-staging work — zero per-row engine code touched. BSBM measured anyway on the shared parse entry (numbers below): HEAD ≤ base on all three queries.

**Production bugs found & fixed along the way:** the API diagnostics-swallowing hole — `fluree.query` executed error-recovered ASTs, silently answering a *different question* than asked (this is the wave's most important product fix); five "green-by-recovery" W3C tests that only passed because of that hole (their real parser gaps are fixed here, not registered); the #1438 silent-data-loss window closed (loud guard, verified end-to-end).

**D-4 (reject-more behavior changes ship as hard errors) — actioned by pr-2 and pr-3.**
- *Options considered:* (1) hard error + changelog — **chosen**; (2) warning under a strictness flag — rejected: a flag means the spec-invalid query still silently executes for everyone who doesn't set it, perpetuating wrong results; (3) defer entirely — rejected: these validations gate ~65 W3C tests and the silent-wrong-results class of bugs.
- *Why (1):* accepted-but-misinterpreted queries produce wrong answers with no signal — strictly worse than a loud error; the added requirement (discovered during verification) is that error-severity diagnostics must **prevent AST production or execution**, otherwise the API swallows them and the "hard error" is theater. Migration notes list every newly-rejected shape.

**D-10a (fast-track the multi-op guard) — actioned by pr-3.**
- *Options considered:* (1) loud trailing-token error now, real `;`-loop support in Wave 2 — **chosen**; (2) wait for full multi-op support — rejected: every day in between, standard multi-op SPARQL UPDATE requests silently commit only their first operation (#1438 — silent data loss in production).
- *Why (1):* converts silent data loss into an explicit, actionable error with zero risk; the Wave-2 implementation replaces the guard in place.

**D-1 (SPARQL 1.2 triple-term strategy: accept-then-defer) — actioned by pr-w2a for the reifier forms.**
- *Options considered:* (1) accept-then-defer: parse everything the RDF 1.2 grammar allows, lower reifier-model-expressible forms to the existing edge-annotation IR, clean `not_implemented` for the rest — **chosen**; (2) reject-on-principle (documented-divergence register) — rejected: keeps 60+ syntax tests red for forms the engine can actually evaluate; (3) desugar triple terms to reifiers — rejected as semantically wrong: the 1.2 grammar itself distinguishes `BIND(<<( )>>)` (valid, a value) from `BIND(<< >>)` (invalid, a reifier) — desugaring would green syntax while corrupting eval semantics.
- *Why (1):* zero storage/index/perf risk, honest runtime errors for the deferred tail. *Known trade-off, stated plainly:* a user writing a deferred form gets a runtime `not_implemented` instead of a parse error.

---

## Wave 1 of the W3C SPARQL burn-down (4 components, ≈−123 register entries)

Stacks on the Wave-0 PR. This wave is the **reject-more + validation wave**: it makes the parser and API say "no" where the spec requires it. Every component carries the decision-transparency and changelog obligations for the behavior changes it ships (D-4 was adopted as hard-errors-with-changelog; the options considered are in each section and ROADMAP §4).

### Components

**1. Semantic validation passes (`pr-2`, −28).** V3 blank-node label scope across BGP boundaries, V4 GROUP BY/aggregate projection scope, V5 BIND scope, V6 duplicate/in-scope SELECT aliases — four validation passes that previously did not exist anywhere — plus the SPARQL 1.2 nested-aggregate and duplicated-VALUES-variable checks. The two AST walkers (free-variable collector, `contains_aggregate`) were written from scratch (the audit's claim that one existed was refuted). V5 lives **in the parser** by necessity: the single-pattern group simplification makes the legal and illegal BIND-scope shapes byte-identical as ASTs. **Behavior changes your review should consciously bless (both pinned by tests + in the migration note):** (a) `GROUP BY (expr)` + reprojection — the field/#1362 shape, literally W3C `agg08` — is now rejected; alias the key (`GROUP BY (expr AS ?k)`). (b) Fluree's SPARQL grouped-list projection extension is literally W3C `group06` — removed from the SPARQL surface; JSON-LD keeps it, with parity pins documenting the deliberate divergence.

**2. Grammar tightening + the diagnostics seam (`pr-3`, −15).** V1 dot-structure enforcement (same token flow, no backtracking), V2 FILTER-requires-Constraint, and the product-critical seam fix: **parse-error diagnostics are now authoritative at the API even when error recovery produced an AST** — previously `fluree.query` silently executed recovered ASTs that answered a different question than the user asked. That fix immediately unmasked five "green-by-recovery" tests, whose real gaps (bare ORDER BY constraints, VALUES row shape, sub-SELECT VALUES lowering) are fixed here rather than registered. Also owns the single trailing-token/EOF assertion: multi-op UPDATE requests stopped silently executing only their first operation and errored loudly instead (D-10a interim guard for #1438; verified end-to-end). **Decision transparency:** actions D-4 and D-10a; V1 is flagged as the highest-regression-surface change in the wave (users may rely on dot leniency) — isolated commits for easy bisect.

**3. UPDATE validation (`pr-u1`, −19).** Class D: `GRAPH` blocks inside `DELETE WHERE` now route through the existing Modify graph-scoped-template lowering (both prior rejections removed; triple-only fast path untouched). Class E: blank nodes in DELETE forms rejected across validator + lowering + **the JSON-LD transact surface** (which can express a bnode-in-delete — behavior change, in the migration note), with the stable `_:fdb-` ids exempted. Compliance-doc parity table gains the UPDATE/transact row.

**4. RDF 1.2 reifier-form parser (`pr-w2a`, −61).** Object-position/nested/standalone `<< s p o ~ r >>`, `~` variants, multi-reifier/multi-block annotation tails grouped into `AnnotationUnit`s per the RDF 1.2 attachment rule, property-path verbs in annotation bodies — all lowered to the existing `Pattern::EdgeAnnotation` machinery where evaluable, clean `not_implemented` where deferred. Kept the negative suite green **by design** (new VERSION short-string check; anon-annotation-in-DELETE validator). Includes the 3 `SPARQL12_VERSION` entries (`VERSION` already parsed; the failures were reified-triple bodies). **Decision transparency (D-1):** accept-then-defer was chosen over reject-on-principle and over desugar-to-reifier (the latter is semantically wrong: the 1.2 grammar distinguishes triple terms from reifiers); trade-off — deferred forms now fail at runtime with `not_implemented` instead of at parse. JSON-LD parity: byte-equal results between `@annotation` inserts queried via `rdf:reifies` and via the new syntax.

### Integration notes
- The wave-1 fixpoint commit prunes 56 tests that green only under the combined defenses (e.g. update-syntax tests covered by PR-2+PR-3+PR-U1 simultaneously; the pr-1∩pr-3 `syntax-order-07` and pr-2∩pr-3 `anonreifier-02` double-claims resolved mechanically by the gate).
- `tests.rs` was rebuilt deterministically from the clean parents after interleaved test-section hunks defeated naive merging; `SubSelect` gained pr-3's `values` field across all initializers; pr-2's walkers were migrated to pr-w2a's `AnnotationUnit` AST.

### Verification
Full W3C suite **36/36, 0 failed, 0 stale**; `fluree-db-sparql` 601 unit tests; `grp_query_sparql` 274, `grp_transact` 171; fmt + clippy clean.

Issues closed when merged: #1438's interim guard (full fix in Wave 2).

---

**Validation passes (pr-2):** run once per parse over the AST; the two new walkers are simple recursive visitors. No runtime/engine code touched; nothing per-row.

**Grammar tightening + API seam (pr-3):** V1/V2 preserve the existing token flow (a `dot_allowed` flag and a post-parse AST-shape predicate — no backtracking added to the happy path); the EOF assertion is a single end-of-parse check; the API seam fix is one branch at query submission. `query_hot_bsbm` before/after: 538/682/950 µs (HEAD) vs 579/735/1005 µs (base) for q3/q5/q9 — HEAD ≤ base across the board; later same-code reruns showed the box noisy under concurrent sessions, so read this as "no regression," not as an improvement claim. No per-row code touched.

**UPDATE validation (pr-u1):** update staging only — the query hot path is untouched by construction. GRAPH-in-DELETE-WHERE routes through the *existing* Modify graph-scoped-template lowering rather than new machinery; the triple-only DELETE WHERE fast path is preserved.

**Reifier-form parser (pr-w2a):** parse-time + lowering onto the existing `Pattern::EdgeAnnotation` operators — D-1's Option 1 was chosen precisely because it carries zero storage/index/perf risk. The grown `AnnotationTarget` IR variant is boxed to keep the pattern-enum width unchanged (enum width is the real hot-path exposure for IR types). Deferred positions cost nothing at runtime (they error at lowering).

## Issues closed on merge

None auto-closed by this PR — deliberately. #1438's window is closed here (silent data loss → loud error, verified end-to-end with a two-op request that previously committed only its first operation), but the *fix* — real multi-operation support — lands in the Wave 2 PR, which carries the `Closes #1438` keyword.

---

## Post-review updates (2026-07-11)

Both review passes (bplatz's inline comments + the local deep review) addressed in 7 follow-up commits; every gate re-run green: workspace nextest 2,154 + 2,412, full W3C suite 36/36 with 0 failed / 0 stale registers, `cargo fmt --all -- --check` clean, testsuite clippy `-D warnings` clean.

**Blocking items.** The two `cargo fmt` hunks are fixed. `WAVE3-HANDOFF-DOSSIER.md` is untracked from the tree (it entered as a merge-resolution add in the pr-w2a merge `e24694169`, invisible to a plain `git log -- <file>`; wave-2/wave-3 carry the identical blob unmodified, so the deletion propagates cleanly through future merge-ups). CI's absence was root-caused, not just retried: the PR was opened against `burndown/wave-0`, whose base didn't match ci.yml's `pull_request: branches: [main]`; Graphite's auto-retarget to main after wave-0 merged emits no `synchronize` event, so CI never re-evaluated. This push is the first event with base = main — the full workflow now runs.

**bplatz #1 — validation at the UPDATE seam.** `parse_and_lower_sparql_update` now runs the same `validate()` pass as the query seam, so validator-side UPDATE rules apply to production updates, not only inside the W3C harness. The strict validator deliberately over-rejects two documented, test-pinned DELETE WHERE behaviors (existential blank nodes; anonymous annotation tails), so the seam opts in via a new `Capabilities::delete_where_extensions` (default `false` = strict W3C — the harness and negative-syntax tests are untouched; `Capabilities` is now `#[non_exhaustive]`). **Migration note (newly rejected on the update surface):** (a) variables in `INSERT DATA` / `DELETE DATA` now error loudly — previously they lowered to never-bindable template variables whose flakes were silently skipped (a silent no-op); (b) update WHERE clauses are now validated by the same rules query WHERE clauses have enforced since the V4 pass — concretely, sub-SELECTs with `SELECT *` + `GROUP BY` or with ungrouped projected variables were previously accepted and executed inside an update WHERE and now reject up front (both pinned by seam tests). Everything else the seam's validate() rejects was already rejected at lowering (message source changes, behavior doesn't).

**bplatz #2 — multi-op AST suppression.** The parser now suppresses AST production for any unconsumed trailing input (extending the V5 `BindTargetAlreadyInScope` mechanism to queries and updates alike), so a caller of the public `lower_sparql_update_ast` that skips `has_errors()` can no longer stage a truncated request — the residual #1438 window is closed at the source. Public-contract note: `parse_sparql` no longer returns a recovered AST when trailing input remains.

**Optional items.** `rewrite_blank_nodes_to_vars` now reuses `BlankNodeVarNamer`'s non-lexable `_:`-scheme names — the previous `_fluree_bn_{label}` names were writable as user variables, so `?_fluree_bn_x` in the same request would silently unify with the rewritten existential — and the namer's anonymous arm mints `_:[]{N}`, fixing a pre-existing label-space collision (`_:b0` and `[]` used to map to the same variable and unify two independent existentials). The flagged validator↔lowering `fdb-` stable-id predicate divergence was investigated and **refuted**: `stable_blank_node_sid_from_label` is itself purely prefix-based, so both sides classify every label identically (`_:fdb-zzz` is a constant addressing a possibly-nonexistent stored node on both — a spec-correct no-op delete); a new predicate-parity test pins the two sides together so future drift fails loudly.

**Adjacent fix (found while verifying the naming change):** query lowering had the same anonymous-blank-node label-space collision (`_:b{N}` names minted at `vars.len()`), so `SELECT * WHERE { _:b2 <p> ?o . [] <p> ?o }` silently unified two independent existentials; anonymous names now use the unwritable `_:[]{N}` namespace on the query path too.

**Main merged in (fa6177b09):** main moved under the PR (wave-0's own review-feedback commit 99d384261), and its new ORDER BY/GROUP BY anti-absorption tests relied on trailing tokens being silently ignored — the exact leniency this wave removes. Reconciled inside the merge commit: those tests now assert the loud trailing-token rejection, which preserves their anti-absorption intent more strongly (absorption regressing would make the inputs parse cleanly again). All gates re-verified on the merged tree.

**Register arithmetic, for release notes:** the title's −123 is the removals count; net is **−121** — 123 entries removed, 2 honestly re-registered as loud-fail multi-op tests (`dawg-delete-insert-01b`, `update-reifier-08`, greened by PR-U2).


fluree-db-sparql/src/validate/mod.rs
```diff
@@ -100,24 +113,53 @@ impl<'a> Validator<'a> {
...
 
+    /// Shared validation for a query form's WHERE pattern: the capability
+    /// walk plus the query-only semantic passes (blank-node label scope —
+    /// update operations have their own blank-node rules and are exempt).
+    fn validate_query_where(&mut self, pattern: &GraphPattern) {
+        self.validate_graph_pattern(pattern);
+        bnode_scope::check_blank_node_scopes(pattern, &mut self.diagnostics);
+    }
+
     fn validate_update(&mut self, op: &UpdateOperation) {
```
bplatz, line 199: `validate_update` — and the UPDATE rules under it (`BlankNodeInDelete`, `AnonymousAnnotationInDelete`, `UnsupportedGraphInUpdate`) — is never reached on the production UPDATE path. `fluree-db-api`'s `parse_and_lower_sparql_update` only checks parse-level `has_errors()` and then lowers; it never calls `validate()`. Enforcement today relies entirely on the lowering-side mirrors (`reject_blank_nodes_in_delete_quad_pattern`, the `GraphName::Var` rejection, `resolve_reifier`'s anon/blank-reifier guards), so behavior is currently correct — but this is fragile in two ways: a W3C negative-update test can go green via the harness's own `validate()` call while production takes a *different* code path, and any future validator-only UPDATE rule would silently not apply to real updates. Suggest either wiring `validate()` into the update seam or documenting explicitly that lowering is the authoritative enforcement point for UPDATE.

fluree-db-sparql/src/parse/query/mod.rs
```diff
@@ -62,12 +62,61 @@ pub fn parse_sparql(input: &str) -> ParseOutput<SparqlAst> {
...
+                if matches!(ast.body, QueryBody::Update(_)) && stream.check(&TokenKind::Semicolon) {
+                    // Update ::= Prologue ( Update1 ( ';' Update )? )? — the
+                    // recursive Update may be empty, so one trailing ';'
+                    // after the last operation is valid SPARQL.
+                    stream.advance();
+                    if !stream.is_eof() {
+                        let span = stream.current_span();
+                        stream.add_diagnostic(Diagnostic::error(
+                            DiagCode::ExpectedToken,
+                            "multi-operation SPARQL UPDATE requests (';'-separated) are not \
```
bplatz, line 92: This multi-op guard (the #1438 silent-data-loss fix) pushes an error diagnostic but does **not** null the AST — unlike V5's `BindTargetAlreadyInScope`, which suppresses AST production. The in-tree seams (`parse_and_lower_sparql_update`, `parse_and_validate_sparql`) all check `has_errors()`, so it's loud where it matters. But the public `lower_sparql_update_ast` export has no internal guard: a caller that reads `.ast` without first checking `has_errors()` still gets a valid single-op AST and executes only the first operation — re-opening exactly the #1438 window this guard exists to close. Consider nulling the AST here (or guarding inside the lowering entry point) as belt-and-suspenders, given the failure mode is silent data loss.

bplatz (approved):
Couple comments inline.

fluree-db-sparql/src/validate/mod.rs
aaj3f: Wired in (1c3d845b8): `parse_and_lower_sparql_update` now runs the same `validate()` pass as the query seam before lowering, so validator-only UPDATE rules apply on the production path — proven at the seam by new `it_transact_update` tests, including the previously-silent vars-in-ground-DATA no-op, which now errors loudly. One wrinkle this comment surfaced: the strict validator deliberately over-rejects two documented, test-pinned DELETE WHERE behaviors (existential blank nodes; anonymous annotation tails), so the seam opts in via a new `Capabilities::delete_where_extensions` (`#[non_exhaustive]`; default stays strict so the W3C harness's negative-syntax tests are untouched). The lowering mirrors remain as defense-in-depth for direct `lower_sparql_update_ast` callers, and the sub-SELECT projection rules newly firing on update WHERE clauses (previously accepted and executed) are migration-noted in the PR body and pinned by 6e774d90f.

fluree-db-sparql/src/parse/query/mod.rs
aaj3f: Done (a7427231e): the guard — and the generic trailing-tokens case, queries included — now suppresses AST production through the same mechanism as V5, so any unconsumed input returns `ast=None` alongside the diagnostics, and a `has_errors()`-skipping `lower_sparql_update_ast` caller can no longer stage a truncated request. Keyed on a local flag at the guard rather than `DiagCode::ExpectedToken` (which ordinary error recovery uses at ~a dozen sites where a best-effort AST is still wanted for tooling). Parser tests pin `ast.is_none()` for every trailing-input shape; the grammar's legal single trailing `;` keeps its AST. The parenthetical alternative (guarding inside the lowering entry) turned out not to be cleanly possible — by the time `lower_sparql_update_ast` runs, the multi-op signal lives only in the discarded diagnostics — which is why the suppression sits at the parse entry.

PR #1454, opened 2026-07-07
feat(sparql): burn-down Wave 2 — GRAPH conformance (D-2), multi-op UPDATE (#1438), BASE resolution, Turtle-star ingest (−31+ registers)

### LINKED ISSUE #1317: `GRAPH <g>` leaks the default graph after `WITH <g> DELETE WHERE` on a not-yet-existing named graph followed by an upsert
## Summary

On the ledger-scoped query endpoint, after the sequence **(1)** `WITH <g> DELETE { ?s ?p ?o } WHERE { ?s ?p ?o }` on a **named graph that does not exist yet**, then **(2)** an `upsert` of TriG data into `<g>`, subsequent `GRAPH <g> { ?s ?p ?o }` queries return the **union of `<g>` and the default graph** instead of only the triples of `<g>`.

The data is stored correctly (the default graph is untouched, the named graph holds only its own triples), so this is a **read / graph-resolution** bug: named-graph isolation breaks for `<g>` after this specific write sequence.

## Environment

- Fluree Core `main` @ `fc0303fff` (2026-06-11)
- Single-server mode, local file storage (`fluree server run --storage-path store`)
- Endpoint: ledger-scoped `POST /v1/fluree/query/{ledger}`

## Steps to reproduce

The bug only manifests when the **default graph is large/complex enough to be indexed across multiple leaves + branches** (reliably reproduced with the GEMET thesaurus: ~81.9k triples → reindex stats `leaf_count: 12, branch_count: 4`). It did **not** reproduce with 3 triples, nor with 20 000 flat synthetic triples (index tree too shallow).

```bash
B=http://localhost:8090/v1/fluree

# 1. Ledger with a large default graph (here: SKOS + GEMET thesaurus, ~81895 triples)
curl -X POST $B/create -H 'Content-Type: application/json' -d '{"ledger":"repro:main"}'
curl -X POST $B/insert/repro:main -H 'Content-Type: application/json' --data-binary @skos.jsonld
curl -X POST $B/insert/repro:main -H 'Content-Type: application/json' --data-binary @GEMET_en_es_fr.jsonld
# (default graph is now indexed across multiple leaves/branches)

# 2. TRIGGER — wipe a NON-EXISTENT named graph, then upsert into it
curl -X POST $B/update/repro:main \
  -H 'Content-Type: application/sparql-update' \
  --data 'WITH <urn:doc:1> DELETE { ?s ?p ?o } WHERE { ?s ?p ?o }'

printf 'GRAPH <urn:doc:1> { <http://ex.org/docA> <http://ex.org/p> "X" . }' | \
curl -X POST $B/upsert/repro:main -H 'Content-Type: application/trig' --data-binary @-

# 3. Read the named graph
curl -X POST $B/query/repro:main \
  -H 'Content-Type: application/sparql-query' -H 'Accept: application/sparql-results+json' \
  --data 'SELECT (COUNT(*) AS ?c) WHERE { GRAPH <urn:doc:1> { ?s ?p ?o } }'
```

## Expected

`GRAPH <urn:doc:1>` contains **1** triple (`<http://ex.org/docA> <http://ex.org/p> "X"`).

## Actual

`GRAPH <urn:doc:1>` returns **~81896** triples — the 1 named-graph triple **plus the entire default graph**. A default-graph subject that was never written to the named graph resolves inside it:

```sparql
ASK { GRAPH <urn:doc:1> { <http://www.eionet.europa.eu/gemet/concept/100> skos:prefLabel ?l } }
# => true   (should be false)
```

The underlying storage is correct — the default graph is unchanged and the upserted triple is only in `<urn:doc:1>` at the storage level — so only the **read-time graph resolution** is wrong.

Once the leaked state has settled, every query shape agrees on it:

| Query on `GRAPH <urn:doc:1>` | Result | Expected |
|---|---|---|
| `SELECT (COUNT(*) AS ?c) WHERE { GRAPH <urn:doc:1> { ?s ?p ?o } }` | `81896` | `1` |
| `SELECT (COUNT(?s) AS ?c) WHERE { GRAPH <urn:doc:1> { ?s a skos:Concept } }` | `5573` | `0` |
| `ASK { GRAPH <urn:doc:1> { <…/concept/100> skos:prefLabel ?l } }` | `true` | `false` |
| `SELECT ?l WHERE { GRAPH <urn:doc:1> { <…/concept/100> skos:prefLabel ?l } }` | 3 default-graph labels | empty |

### Secondary observation: read-after-write inconsistency on the same path

Immediately after step (2) the upsert, the **first** `COUNT(*)` on `GRAPH <urn:doc:1>` transiently returns `1` (the correct, isolated value); re-running the identical query a moment later returns `81896` (the leaked value), and it stays there. So the same query on an unchanged ledger returns different results depending on timing — the leaked/indexed read path and the just-written novelty appear to be merged inconsistently for a short window. This may be the same root cause surfacing at two stages (write-time graph registration and read-time resolution).

## Scope / diagnostic (negative controls)

All on the same large-default-graph ledger:

| Sequence | `GRAPH <g>` result |
|---|---|
| Read a never-written named graph | ✅ empty |
| `upsert` into a fresh named graph **without** the preceding `WITH … DELETE` | ✅ isolated (only its own triples) |
| `WITH <g> DELETE WHERE { ?s ?p ?o }` alone (no upsert) | ✅ empty |
| `WITH <g> DELETE WHERE { ?s ?p ?o }` when the graph **already has content** + upsert | ✅ isolated |
| **`WITH <g> DELETE WHERE { ?s ?p ?o }` on a fresh graph, then upsert** | ❌ **returns `<g>` ∪ default graph** |

So the trigger is specifically **wipe-then-upsert on a not-yet-registered named graph**, on a ledger whose default graph spans a multi-level index. Wiping an already-existing graph is fine; a plain first upsert is fine.

## Impact

This breaks per-document named-graph workflows (one named graph per document, wiped-then-rewritten on each (re)publish). The first publish of any document corrupts that document's named graph so that every read of it also returns the whole default graph (the shared model / ontology), making per-document reads unusable.

## Suspected area

Likely in named-graph resolution on the indexed read path vs. `graph_registry` g_id assignment (`fluree-db-core/src/graph_registry.rs`, `FIRST_USER_GRAPH_ID = 3`): a `WITH <iri> DELETE WHERE` that references a not-yet-registered graph appears to leave `<iri>` resolving to / aliasing the default partition (`g_id = 0 / DEFAULT_GRAPH_ID`) at read time once the graph is materialized by the subsequent upsert. Same area as #1279 (single-db `GRAPH` resolution). It only surfaces once the default graph is indexed across multiple leaves/branches, so the novelty-only read path is unaffected.

### LINKED ISSUE #1438: SPARQL UPDATE: multi-operation (;) requests silently execute only the first operation

### LINKED ISSUE #1442: GRAPH ?g binds graph names as xsd:string literals and enumerates the default graph (W3C-breaking)
Two defects in `GraphOperator`:

1. `graph.rs:314-323` builds `Binding::Lit{String}` where an IRI term is required — `?g` comes back as a plain literal containing the IRI text (`{"type":"literal"}` instead of `{"type":"uri"}`).
2. `graph.rs:686-711` appends the ledger alias to the unbound-`?g` fan-out, exposing the default graph as a named graph. This is the deliberate #1279 extension, but it is W3C-breaking (`GRAPH ?g` must range over named graphs only).

**Decision D-2 (approved): W3C-by-default** — drop the implicit default-graph enumeration, keep explicit `GRAPH <alias>` addressing, no toggle. Three tests pin the current behavior and get updated: `it_query_dataset.rs:1591`, `:1752`, `it_upsert_duplicate_ids_repro.rs:228`. This is the tracking issue for that behavior change (roadmap PR-G1; options considered and rationale are in ROADMAP §4 D-2).

### LINKED ISSUE #1443: Bound GRAPH ?g returns nothing when ?g is a late-materialized EncodedSid (EXISTS/correlated contexts)
`extract_graph_iri_from_binding` (`graph.rs:157-172`) has no `EncodedSid`/`EncodedLit` arm, so a scan-sourced ref-valued `?g` silently matches nothing while string-valued rows survive (verified probe). Manifests inside `FILTER EXISTS { GRAPH ?g {...} }` and correlated shapes; top-level scan-bound `?g` works today, so the bug hides. Roadmap PR-G1 (BUG-5) — the regression test must use the EXISTS/late-materialized shape.

aaj3f:
## Reviewer quick reference

**End-user API/contract changes:** (a) **`GRAPH ?g` semantics change (D-2)** — `?g` binds as an IRI term (was: a string literal), and unbound `GRAPH ?g` no longer enumerates the ledger-alias default graph; explicit `GRAPH <alias>` addressing still works. Queries relying on the old implicit enumeration must name the alias. Three pinning tests updated; 2-reviewer sign-off requested on this component. (b) **Multi-operation UPDATE requests now execute fully and atomically** (previously: silent first-op-only, then Wave 1's loud error) — a new capability, closing #1438. (c) `BASE`-relative constant IRIs now resolve per RFC 3986 at query preparation (spec behavior; previously unresolved text). (d) Turtle-star asserting forms are newly accepted input; **documented divergence:** Fluree's edge-annotation model asserts the base triple of a reified statement (RDF 1.2's non-asserting reading belongs to the triple-terms epic).

**Performance verdict:** all changes are plan/lower/materialization-time or update-staging; the single-operation UPDATE path delegates byte-identically. BSBM at noise level on every component (numbers per component below); `transact_commit` measured marginally *faster*; import benches for the Turtle-star lexer within budget on a 6-round interleave.

**Production bugs found & fixed along the way:** `GRAPH ?g` binding graph names as `xsd:string` literals and leaking the default graph (#1442); bound `GRAPH ?g` silently matching nothing for late-materialized encoded bindings in EXISTS/correlated shapes (#1443); multi-op UPDATE silent data loss (#1438, fully); GRAPH-variable joins overwriting inner solutions instead of joining; a base-IRI resolution hole that five seemingly-unrelated test failures were hiding in.

**D-2 (`GRAPH ?g` goes W3C-by-default) — actioned by pr-g1.**
- *Options considered:* (1) drop the implicit default-graph enumeration, keep explicit alias addressing, no toggle — **chosen**; (2) an opt-in strictness flag — rejected: per-query flags on the GRAPH operator would put a mode branch into graph resolution, and a flag default preserves spec-wrong behavior for everyone; (3) keep the #1279 extension as-is — rejected: it makes `GRAPH ?g` results wrong per SPARQL (the default graph must not appear) and blocked ~14 W3C tests.
- *Why (1):* no W3C test — and no discovered consumer — names the alias implicitly; explicit addressing preserves the useful half of the extension; the three tests that pinned the old behavior are updated in this PR, and the changelog note covers migration. The #1279 history and the deviation from the audit's original seeding design (which regressed `graph-variable-scope` when implemented as written) are documented in the component section.

**D-10b (multi-op UPDATE staging model) — actioned by pr-u2.**
- *Options considered:* (1) `Vec<Txn>` with sequential staging against an evolving novelty overlay, folded into ONE atomic commit — **chosen**; (2) a new `UpdateRequest` IR wrapper — rejected: ripples through every Txn consumer (JSON-LD lowering, `pre_built_txn`, Cypher) for no semantic gain; (3) N separate commits, one per operation — rejected: a mid-request failure would leave a partially-applied request durably committed, violating the atomicity a client sending one request expects.
- *Why (1):* op N's WHERE must see op N−1's effects (SPARQL update-sequence semantics) while the request commits as a unit; the virtual-commit overlay provides the first property, the last-wins fold over the original base provides the second, and the single-op path is untouched.

**D-8 (TriG parser strategy) — partially constrained by pr-w15:** TriG-star inputs get clean, construct-specific deferred errors that do not foreclose the adopted try-the-existing-`parse_trig_phase1`-route-first plan for the harness's TriG needs.

---

## Wave 2 of the W3C SPARQL burn-down (4 components, ≈−31 direct + joint entries)

Stacks on the Wave-1 PR. This wave carries the two approved behavior-change decisions (D-2 GRAPH semantics, D-10 multi-op UPDATE) plus base-IRI resolution and Turtle-star ingest. Two components deliberately modify pinned engine semantics — flagged for 2-reviewer sign-off in their sections.

### Components

**1. Base-relative IRI resolution (`pr-base`, −6 + joints).** Extracts the Turtle parser's RFC 3986 resolver into dependency-free `fluree_vocab::iri` (behavior-identical delegation) and threads the query `BASE` through lowering: constant IRIs in patterns/GRAPH/paths/templates/VALUES, relative PREFIX namespaces, and an `IRI()`/`URI()` constant fold — all at prepare time, eval path stays base-free (variable-argument `IRI()` deliberately not resolved; documented). Ships `resolve_dataset_clause()` for the future within-ledger FROM work without lifting the guard. The harness now supplies the query-document base, mirroring its data-side `@base`. This PR exists because the audit's "BUG-4" was refuted — `graph-exist` actually failed on base resolution, exposing an ownership vacuum five tests were hiding in.

**2. GRAPH operator W3C conformance (`pr-g1`, −14 + joints) — DELIBERATE SEMANTICS CHANGE (D-2), 2-reviewer sign-off requested.** `?g` now binds as an IRI term (was: string literal); unbound `GRAPH ?g` no longer enumerates the ledger-alias default graph (the intentional #1279 extension — **removed**; explicit `GRAPH <alias>` addressing kept; no toggle). Three pinning tests updated. Join semantics: seed-when-produced + join-at-merge (the audit's unconditional-seeding design was refuted by W3C `graph-variable-scope` regressing — deviation documented); `EncodedSid`/`EncodedLit` extraction arms plus a `GraphVarCorrelated` EXISTS fallback (#1443 — Semijoin hash keys can never match across encodings; the regression test uses the EXISTS/late-materialized shape that actually pins it). Adopts the residual-audit defects A3/A4 (`subquery02/04`). Does NOT close #1317 (distinct defect; see its issue comment). **Changelog:** queries relying on implicit default-graph enumeration under `GRAPH ?g` must address the alias explicitly. FQL parity: three regression tests (same `GraphOperator`).

**3. Multi-operation UPDATE requests (`pr-u2`, −9) — closes #1438 (D-10).** Request-level `;` loop with accumulating prologue (one trailing `;` legal; empty/prologue-only request = valid no-op); cross-op blank-node label reuse is a parse error; **sequential staging within ONE atomic commit**: each op's WHERE sees the previous ops' effects via a virtual-commit overlay, folded last-wins over the original base — the single-op path delegates unchanged. Replaces Wave-1's loud-reject interim guard with real support. **Decision transparency (D-10b):** `Vec<Txn>` sequential staging was chosen over a new UpdateRequest IR and over N separate commits (atomicity per the SPARQL update-request semantics); options and rationale in the description body and ROADMAP §4. `insert-05a` + the same-bnode trio remain registered — their requests end in `DROP GRAPH` (Wave-3 PR-U3 territory). Benches: parse entry + commit path within noise (`transact_commit` marginally faster).

**4. Turtle-star ingest, asserting forms (`pr-w15`, −2 + re-baseline).** Reified triples (`<< s p o ~ r >>`) and annotation blocks (`{| |}`) in Turtle data now lex/parse and emit **byte-identical flakes to the JSON-LD `@annotation` path** via the shared `EdgeKey::to_reifies_facts_jsonld_compatible` encoder (the full-variant alternative would have emitted orphan flakes under cascade retracts — adversarial-verification catch). Fresh reifier per anonymous occurrence, never deduplicated by EdgeKey (`pattern-3-nomatch` is green precisely because two anonymous occurrences don't join); equivalence tests sit next to `edgekey_roundtrip_*`. `<<( )>>` triple terms and TriG-star get clean, specific deferred errors (D-8-compatible). **Documented divergence:** Fluree's edge-annotation model asserts the base triple of a reified statement; RDF 1.2's non-asserting semantics belong to the Option-1 epic. The 39 remaining eval-triple-terms entries were re-baselined against their REAL post-ingest errors (15 wave-2 syntax / 10 D-1 data / 5 serialization / 5 TriG / 2 harness / 2 misc). Import benches: 6-round interleave, all within budget (a 3-round phantom +11.5% dissolved at 6 rounds).

### Integration notes
The wave-2 fixpoint commit prunes 46 tests that green only in integration — the BASE+G1 joint entries (`pp34/pp35`, `exists03`), the G1-gated algebra/`optional-complex` tests, and the eq-graph second-lander interplay — all resolved by the both-way gate. `pr-u2`'s cross-op walker was migrated to the `AnnotationUnit` AST during integration.

### Verification
Full W3C suite **36/36, 0 failed, 0 stale**; merged-crate tests green (`fluree-db-sparql` 1178 incl. integration-migrated tests, transact, turtle, query); JSON-LD groups green; fmt + clippy clean.

Issues closed when merged: #1438 (fully), #1442, #1443.

---

**Base-IRI resolution (pr-base):** all resolution happens at lowering/preparation on constant IRIs — the evaluation path stays completely base-free, by design: variable-argument `IRI()` is deliberately NOT resolved (documented partial-spec choice; no W3C test exercises it) precisely so no per-row resolution exists anywhere. The resolver extraction into `fluree_vocab::iri` is behavior-identical delegation for the Turtle parser (its tests pin this).

**GRAPH conformance (pr-g1):** every fix is plan/seed/materialization-time — `?g` term-kind at binding materialization, enumeration-set construction at plan time, seed-when-produced decided from the subplan's produced-variable set, and the `GraphVarCorrelated` EXISTS fallback selected at strategy-choice time (its per-row evaluation applies only to a shape that previously returned silently-wrong empty results). BSBM: q3 −1.4%, q5 −1.0%, q9 +1.2%, bi/f2 −0.7% — noise-level, as predicted (BSBM never instantiates GRAPH; run for the shared context helpers).

**Multi-op UPDATE (pr-u2):** the single-operation path delegates to the existing code byte-identically — multi-op machinery engages only when a second operation exists. Touches the shared parse entry and the commit path, so both were measured: `query_hot_bsbm` q3 537→528 / q5 680→676 / q9 952→961 µs (noise); `transact_commit` populated 314.3→304.7 µs (marginally faster).

**Turtle-star ingest (pr-w15) — import-hot lexer:** star detection costs at most one extra peek on `<`/`{`/`~`; a byte-identical assertion over a star-free corpus pins that non-star lexing is unchanged. Interleaved A/B, 6 rounds: turtle 10txn_10nodes **+0.84%** (a 3-round phantom reading of +11.5% dissolved at 6 rounds — worth remembering when reviewing bench claims on this box), turtle 100nodes +2.87%, jsonld −0.5/−1.0%, import_bulk −7.5/−22% (noise-wide) — all within the 5% budgets. Flake emission reuses the existing `EdgeKey::to_reifies_facts_jsonld_compatible` encoder — no new storage shapes, no index changes.

## Issues closed on merge

Closes #1438
Closes #1442
Closes #1443

*Mechanics note: these keywords fire when this PR merges into the default branch — in the stacked flow, after #1437 and the prior waves merge and GitHub auto-retargets this PR to `main`. #1317 is deliberately NOT closed by this PR (distinct defect class — see the cross-reference comment on that issue).*


bplatz (commented):
A few issues from reviewing the wave-2 diff. Two don't land on changed lines, so noting them here:

- **`fluree-db-transact/src/parse/trig_meta.rs:616` (`resolve_iri`)** still does `format!("{base}{iri}")` instead of the RFC-3986 `fluree_vocab::iri::resolve_iri` this PR wired into the Turtle parser and SPARQL lowering. A document-IRI `@base` plus a sibling relative ref (e.g. `sibling.ttl`) resolves wrong, and this is the production TriG import/upsert path.
- **`docs/query/sparql.md:549`** still documents unbound `GRAPH ?g` as enumerating the ledger alias / default graph — the exact behavior D-2 removes here. Docs weren't updated in this changeset.

Inline comments below.

fluree-db-query/src/execute/where_plan.rs
```diff
@@ -329,6 +329,36 @@ pub(crate) enum ExistsFallbackReason {
...
+        | Pattern::Exists(inner)
+        | Pattern::NotExists(inner)
+        | Pattern::DefaultGraphSource { patterns: inner } => {
+            has_outer_correlated_graph_var(inner, outer)
+        }
+        Pattern::Union(branches) => branches
+            .iter()
+            .any(|b| has_outer_correlated_graph_var(b, outer)),
+        Pattern::Subquery(sq) => has_outer_correlated_graph_var(&sq.patterns, outer),
+        _ => false,
```
bplatz, line 384: The catch-all misses `Pattern::Service` (and `EdgeAnnotation`/`AnnotationTarget`), whose bodies are walked transparently by `produced_vars`. So `EXISTS { SERVICE <fluree:ledger:x> { GRAPH ?g { … } } }` with `?g` bound outside still selects `Semijoin`, re-opening the #1443 `Binding::Iri` vs `Sid`/`EncodedSid` hash mismatch this fallback exists to close.

fluree-db-sparql/src/parse/query/update.rs
```diff
@@ -9,6 +9,105 @@ use crate::ast::{GraphPattern, Iri};
...
+                    }
+                }
+            }
+        }
+    }
+
+    match op {
+        UpdateOperation::InsertData(insert) => from_elements(&insert.data.quads, out),
+        UpdateOperation::DeleteData(delete) => from_elements(&delete.data.quads, out),
+        UpdateOperation::DeleteWhere(dw) => from_elements(&dw.pattern.patterns, out),
```
bplatz: DELETE WHERE blank nodes are locally-scoped matching variables (per the lowering docs), not identity-bearing template nodes — feeding `dw.pattern` (and `modify.delete_clause`, L102) into the cross-op reuse check over-rejects. `DELETE WHERE { _:b1 :flag "a" } ; DELETE WHERE { _:b1 :flag "b" }` errors as a duplicate label though the two ops are independent. Only INSERT DATA / INSERT templates should feed this.

fluree-db-api/src/tx.rs
```diff
@@ -2031,6 +2031,171 @@ impl crate::Fluree {
...
+
+        let mut txn_meta: Vec<TxnMetaEntry> = Vec::new();
+        let mut last_ns_registry: Option<NamespaceRegistry> = None;
+
+        let op_count = txns.len();
+        let mut current = Some(ledger);
+        for (i, txn) in txns.into_iter().enumerate() {
+            // Captured before staging consumes the Txn; needed to advance
+            // the virtual state between operations.
+            let ns_delta = txn.namespace_delta.clone();
```
bplatz: `namespace_delta` is cumulative here (the shared registry is never reset between ops), so re-adopting it per op re-inserts every prior op's entries — O(n²) over an n-op request. Carrying each op's incremental delta only would make it O(n).

fluree-db-api/src/tx.rs
```diff
@@ -2031,6 +2031,171 @@ impl crate::Fluree {
...
+            let result = self
+                .stage_transaction_from_txn(state, txn, index_config, policy, tracker)
+                .await?;
+
+            for iri in &graph_iris {
+                if seen_graph_iris.insert(iri.clone()) {
+                    merged_graph_iris.push(iri.clone());
+                }
+            }
+            sim_registry.apply_delta(graph_iris.iter());
```
bplatz, line 2262: Per-op `apply_delta` assigns graph ids in arrival order, but `build_commit` assigns over the globally-sorted union, so `commit_record.graph_delta` is persisted with a different g_id↔IRI mapping than the registry the data actually uses. Live routing self-heals (it recomputes from the registry), but the on-disk commit metadata is wrong — e.g. `INSERT … GRAPH <…/zzz> ; INSERT … GRAPH <…/aaa>`.

fluree-db-transact/src/lower_sparql_update.rs
```diff
@@ -714,25 +714,59 @@ struct LiteralResult {
...
-/// * `opts` - Transaction options (branch, context, etc.)
+/// # Errors
 ///
-/// # Returns
+/// Returns `LowerError` if:
+/// - The AST body is not an UPDATE request (is a query)
+/// - WITH or USING clauses are present
+/// - Blank nodes appear in WHERE patterns
+/// - RDF-star quoted triples are used
+pub fn lower_sparql_update_request(
```
bplatz, line 752: Constant IRIs in UPDATE operations aren't resolved against `BASE`: `expand_iri` (L1842) handles only Full/Prefixed and never reads `prologue.base`, unlike the query path's `expand_iri_with`. `BASE <http://x/> INSERT DATA { GRAPH <g1> { <s1> <p1> "v" } }` stores literal `g1`/`s1`/`p1`, while the same relative IRIs in a SELECT now resolve. No test covers BASE + UPDATE.

fluree-db-transact/src/flake_sink.rs
```diff
@@ -261,6 +262,69 @@ impl GraphSink for FlakeSink<'_> {
...
+
+    /// Turtle-star reifier attachment → the durable `f:reifies*` bundle.
+    ///
+    /// Emits via [`EdgeKey::to_reifies_facts_jsonld_compatible`] — the SAME
+    /// shape the JSON-LD `@annotation` lowering produces (no
+    /// `f:reifiesDatatype` flake) — so Turtle-star and JSON-LD annotations
+    /// are bit-identical at the flake level and cascade retracts built from
+    /// `AttachmentNovelty` cancel either surface's assertions. The base
+    /// triple has already been emitted by the parser via `emit_triple`.
+    fn emit_reified_triple(
```
bplatz, line 284: `FlakeSink::emit_reified_triple` and `ImportSink::emit_reified_triple` duplicate the resolve → dt/lang → validate → `EdgeKey` → `to_reifies_facts_jsonld_compatible` bundle build (only the emit differs). A shared helper would keep the "bit-identical bundles" guarantee from drifting. Minor — mirrors the existing `build_flake`/`push_triple` split.

bplatz (approved):
Some comments inline to investigate.

fluree-db-query/src/execute/where_plan.rs
aaj3f: Fixed in 3f0e2ae6d — the walker now recurses into `Pattern::Service` bodies, `EdgeAnnotation`/`AnnotationTarget` bodies, and (a sibling gap the local pass found) expression-embedded pattern lists: `FILTER(EXISTS { … } && …)`, `BIND(EXISTS { … } AS ?x)`, and pattern comprehensions, via a new `Expression::for_each_embedded_patterns` traversal. Decision-level tests pin the bare shape, all three container shapes, and a negative control proving the Semijoin fast path survives for uncorrelated graph vars behind SERVICE.

fluree-db-sparql/src/parse/query/update.rs
aaj3f: Fixed in 7fa5988ba, taking exactly your reading: §19.6's label ban protects blank-node *minting* positions only, so the walker now collects INSERT DATA quad data and Modify INSERT templates, nothing else. DELETE-side blank nodes either act as locally-scoped existential variables (DELETE WHERE, per the documented extension) or are rejected outright downstream by the strict validator/lowering with the real error — which the label complaint used to mask. Your repro (`DELETE WHERE { _:b1 :flag "a" } ; DELETE WHERE { _:b1 :flag "b" }`) is pinned at parse level and executes end-to-end at the seam (each op deletes its own matches); W3C syntax-update-54 rejection is unchanged.

fluree-db-api/src/tx.rs
aaj3f: Acknowledged and deliberately deferred — filed as #1479. Realistic requests are small-n, and making it O(n) means carrying per-op *incremental* deltas through the Txn IR, whose cumulative shape is now documented/load-bearing. Worth flagging: investigating your graph_delta thread below surfaced a real correctness bug in this same loop's delta handling — the loop did change in a9106bb6e — but the cumulative shape (and thus this observation) still stands.

fluree-db-api/src/tx.rs
aaj3f: Your instinct here was right — and it ran deeper than metadata. The discriminating test you suggested (`INSERT … GRAPH <…/zzz> ; INSERT … GRAPH <…/aaa>`, commit, cold-reload, verify routing) found that the request **failed to commit at all**: the between-ops virtual apply and the final commit envelope only carried the *lowering-time* namespace deltas, but the namespace of a `GRAPH <iri>` name is allocated at *staging* time — so op 1's graph namespace never reached the virtual state (`InvalidGraph: flake references unknown graph Sid`), later ops' registries could diverge on code assignment (constructible cross-op Sid collisions), and the last-op-only envelope broke cold reloads. Fixed in a9106bb6e: the loop now threads each op's post-staging registry delta and re-adopts the union into the commit envelope. On the id-ordering half specifically: the persisted `graph_delta` *keys* are the staging-order sim ids while load-time registration re-derives ids from sorted `apply_delta` batches — but every consumer (live routing, commit replay, index metadata) reads the **values** and re-derives ids through the registry, so each instance is internally consistent; the new test pins routing both live and after a cold reload from disk.

fluree-db-transact/src/lower_sparql_update.rs
aaj3f: Fixed in 13a56e148 — verified end-to-end exactly as you described (`BASE <http://x/> INSERT DATA { GRAPH <g1> { <s1> <p1> "v" } }` stored literal `g1`/`s1`/`p1`). All 16 constant-IRI call sites funnel through the one leaf `expand_iri`, which now resolves with the same `fluree_vocab::iri` primitives as the query path's `expand_iri_with` (RFC 3986 §5 for relative refs, §4.1.1 for relative PREFIX namespaces, any-scheme absolutes pass through). Unit tests pin the RFC cases; a seam test pins that the same BASE-carrying document stores and finds identical IRIs on both surfaces, GRAPH names included. The deferred-WHERE half already resolved correctly via the query-side environment at staging time.

fluree-db-transact/src/flake_sink.rs
aaj3f: Extracted in 6418e998a: the resolve → dt/lang → validate → `EdgeKey` → bundle sequence now lives in `generate::flakes::reified_triple_bundle` (next to the `validate_value_dt_pair` guard both paths share); each sink keeps only its own error channel and emission, mirroring the `build_flake`/`push_triple` split you referenced.

aaj3f:
Review response for both reviews (bplatz's inline pass + the local from-scratch pass) — 12 commits on top of the `origin/main` merge, all items closed or explicitly filed.

**Blocking items from the request-changes review:**
1. **Anonymous mint collision** (`_:b{N}` vs user labels, silent data merge) → e1a0a351c: anonymous mints now use the `-b{N}` namespace (unlexable as a user label, still serializable as `fdb-{txn}--b{N}`), closing the pre-existing `[]` case too; Sid-level unit pin + end-to-end regression mixing `_:b1` with a `{| |}` reifier and a `[]` node.
2. **UPDATE ignores BASE** → 13a56e148 (see thread reply).
3. **#1443 encoded arms unpinned** → d865b9638 + an honest finding: the audit went a step further and proved the `EncodedSid`/`EncodedLit` extraction arms are currently *unreachable* from any plan shape (probed plain join, OPTIONAL, FILTER EXISTS, UNION, and the string variant on a real binary-indexed ledger — upstream operators materialize batches first). What ships: an indexed end-to-end test for the `GraphVarCorrelated` strategy (which no memory-backed suite exercised at all, W3C harness included), the arms kept as defense-in-depth, and their stale "never on a scan hot path" comment rewritten to state the reachability facts.
4. **Multi-op atomicity untested** → d865b9638: the mid-request-failure probe (op 2 fails at staging over op 1's data) now asserts error-as-a-unit, `t` unchanged, no trace; plus an indexed-ledger multi-op test driving the dangling-retract fold through a real novelty→index merge and a cold reload.
5. **DELETE WHERE over-rejection** → 7fa5988ba (see thread reply).
6. **Register double-counting** → 2ddb46c69: the 10 duplicated eval-triple-terms entries deduped to one cluster each, with attribution re-verified empirically (unregistered all 10, read the harness's actual failure reasons, re-added each once); stale cluster counts fixed; 36/36 suites green with both-way policing intact.

**Beyond the reviews** (found while writing the discriminating test bplatz suggested): a multi-op request writing to ≥2 new named graphs failed to commit outright, and a new-namespace-in-an-early-op request persisted a commit envelope that couldn't decode after reload — root cause and fix in a9106bb6e (staging-time namespace allocations now reach the virtual state, later ops' registries, and the commit envelope).

**Optional items landed:** trig_meta RFC 3986 resolution on the production TriG import path (cd8e8d823), where_plan recursion gaps incl. expression-embedded EXISTS (3f0e2ae6d), shared reified-triple bundle builder (6418e998a), the RFC 3986 trailing-slash resolver bug fixed with the six W3C IRI-resolution-08 fixture cases (dd2636e6b), unbound `GRAPH ?g` docs corrected for D-2 + a new multi-operation docs section (e3980628f, 8f5b4e157), per-op validation semantics documented and pinned via an inline-unique transient-violation test (8f5b4e157), merge-join edge-case and last-wins-BASE comments, `;;`-rejection and SPARQL-JSON `"type":"uri"` pins.

**Deferred with issues:** #1479 (O(n²) cumulative ns-delta re-adoption), #1480 (parser recursion depth ceiling), #1481 (lexer pin corpus + intra-prologue redeclared-prefix divergence between surfaces).

**Verification at HEAD (2e8bba826):** full nextest across all six touched crates `--all-features` — 5355/5355; W3C `make test` 36/36 with both-way register policing; CI-equivalent clippy clean; `cargo fmt --check` clean in both workspaces (the two pre-existing drift hunks folded in per review). The branch now includes the `origin/main` merge, so CI runs against the real merge result.

PR #1455, opened 2026-07-07
feat: openCypher write path, LOAD CSV, and expression/pattern surface

bplatz:
Closes a batch of openCypher feature gaps on top of `feature/bolt`, focused on the write path, `LOAD CSV`, and the expression/pattern surface. All new features fail with clear errors when unsupported — never a silently wrong result.

## Write path

- **`MERGE` completion.** Relationship `MERGE` (property patterns, rel-var `ON CREATE SET`, standalone `ON MATCH SET`); the upsert idiom (trailing `SET` and whole-map `SET n += $props`); and **per-row** node and relationship `MERGE … ON MATCH SET` — the `UNWIND $batch AS row MERGE …` shape that powers `LOAD CSV`.
- **Atomic two-stage staging.** A per-row `MERGE … ON MATCH SET` decomposes into two disjoint transactions (on-match vs. create, mutually-exclusive guards) that stage against the same base and land as a **single commit** — fail either stage, no commit. Threaded through all four write paths (owned autocommit, interactive Bolt, local consensus, Raft commit worker).
- **Multi-statement write scripts** (`;`-separated) on the transact API, and **`FOREACH`** over constant lists via parse-time unroll.

## LOAD CSV — `fluree load`

New CLI command streaming a local CSV into a ledger as batched per-row upserts (the `LOAD CSV` analog). The CLI holds the file, reads it client-side, and sends one transaction per batch — the server never sees the file, a path, or a URL. Per-row template in either language:

- `--cypher`: body rides in `UNWIND $batch AS row …`, columns read as `row.<col>`.
- `--jsonld`: batch injected as the update's `values` clause, one `?<column>` variable per CSV column.

Both hit the existing `POST /update/{ledger}` — no new server endpoint. The custom-server contract is documented in `docs/cli/server-integration.md`.

## Expression & pattern surface

- **Temporal constructors** `date()` / `datetime()` / `time()` / `duration()`, component-map constructors, and zero-arg `localdatetime()`.
- **Path enumeration**: free path values (`MATCH p = …`) and unbounded var-length binding.
- **Multi-hop fixed path values**, **property filters on bounded var-length relationships**, the **`NULL` literal** as a first-class value, and **keyword tokens as binding names** (`AS end`, `WITH count(*) AS count`).
- Fix: **`EXISTS` in projection** was collapsing to `false` inside the fused join block.

## Tooling & compatibility

- Bulk import for `.cypher` dump scripts.
- `apoc.meta.data` shim (LangChain `Neo4jGraph` schema introspection) and `db.*` / `dbms.*` procedure shims (`CALL db.labels() YIELD …`).
- Cypher schema DDL accepted as no-ops.

## Testing

Covered by `it_query_cypher.rs`, `it_query.rs`, `it_lower_cypher.rs`, and CLI integration tests (including the `fluree load` Cypher and JSON-LD paths). Docs updated: `docs/query/cypher.md`, `docs/reference/cypher-support-matrix.md`, `docs/cli/load.md`, `docs/cli/server-integration.md`.


fluree-db-cli/src/commands/load.rs
```diff
@@ -0,0 +1,288 @@
...
+                    } else {
+                        Value::String(cell.to_string())
+                    };
+                    obj.insert(header.clone(), value);
+                }
+                Value::Object(obj)
+            })
+            .collect(),
+    )
+}
```
aaj3f, line 231: This is really a comment for this + `fluree-db-cypher/src/params.rs:488-519` + `fluree-db-transact/src/lower_cypher_update.rs:1246-1250` and I think it could produce a silent wrong result on the LOAD CSV mainline.

A node MERGE fed by `UNWIND $batch` ends up lowering to an `InlineRows` VALUES join plus a single top-level `NOT EXISTS` guard evaluated once per row **against the one pre-write snapshot**, with no dedup of rows by the MERGE key. 

Failing input: `fluree load people.csv --cypher 'MERGE (n:Person {id: row.id}) SET n.name = row.name'` 

where two rows in the same batch (default 1000) share `id=5` and no `id=5` exists in the base — both rows pass the guard and the create branch fires twice, producing **two `Person` nodes with `id=5`** (skolemized--I just learned this word!--to distinct subjects per WHERE solution). Neo4j's sequential `LOAD CSV` upserts here. 

This directly contradicts the PR's "never silently wrong" claim and `docs/cli/load.md:80-91`'s upsert promise; `docs/cli/load.md:39-42` documents only *cross-batch* non-atomicity. 

Fix: dedup the batch rows by the MERGE key (last-wins) in `cypher_batch` so batch semantics match sequential `LOAD CSV`, or document the within-batch duplicate-key limitation prominently in `load.md` and add a within-batch dup-key integration case (the current `load_csv_upserts_via_cypher_template` only exercises cross-batch).

fluree-db-api/src/cypher_procedures.rs
```diff
@@ -0,0 +1,612 @@
...
+                        .entry(meta_type_name(tag))
+                        .or_insert(0) += count as i64;
+                }
+            }
+        }
+    }
+
+    // Source 2: novelty attribution. Pass A collects subject classes from
+    // novelty `rdf:type`; pass B attributes novelty property flakes to them.
+    if let Some(novelty) = overlay.and_then(|o| o.as_any().downcast_ref::<Novelty>()) {
```
aaj3f, line 254: (and same file lines 500-509)

`merged_stats`/`meta_data_rows` answer `db.labels` / `db.relationshipTypes` / `db.propertyKeys` / `db.schema.visualization` / `apoc.meta.data` from raw `snapshot.stats` merged with novelty; a policy-view overlay downcasts to `None` (the `None => indexed` arm), so the baked `InlineRows` values are never policy-filtered at execution. A read-restricted identity that can run Cypher therefore sees the entire label/relationship-type/property-key catalog and, via `apoc.meta.data`, per-label / per-property counts -- inconsistent with ordinary `MATCH` reads, which apply policy through the overlay. 

Fix: thread the caller's `PolicyContext` into the stats/novelty walk (or restrict the catalog to policy-visible predicates), or, if lenient-catalog parity with Neo4j is the deliberate choice, state it explicitly in the module doc (`cypher_procedures.rs:14-22`) and the security docs so it is a conscious, reviewed decision rather than an implicit leak.

fluree-db-core/src/temporal.rs
```diff
@@ -1986,6 +1986,252 @@ fn parse_seconds_to_micros(s: &str) -> Result<i64, String> {
...
+            let f = frac(
+                int("millisecond", 0)?,
+                int("microsecond", 0)?,
+                int("nanosecond", 0)?,
+            );
+            let tz = tz_suffix(func == "datetime");
+            Ok(format!(
+                "{y:04}-{mo:02}-{d:02}T{h:02}:{mi:02}:{s:02}{f}{tz}"
+            ))
+        }
```
aaj3f, line 2179: The `frac` closure composes fractional seconds as `format!(".{total:09}")` from a nanosecond total but never carries a ≥ 1 s total into the seconds field, and no range check rejects it. 

I reproduced it standalone: `datetime({year:2024, millisecond:1500})` composes `…T00:00:00.15Z` = **0.15 s, not 1.5 s** (off by a whole second and 10×); `millisecond:2500 → .25Z`. 

The strict re-parse accepts `.15`, so it neither errors nor rolls over. (The `duration()` path at `:2204` handles `milliseconds ≥ 1000` correctly via `s*1000 + ms` rollover — only `frac` is broken.) 

Fix: carry `total / 1_000_000_000` seconds into the seconds field before formatting the remainder, or range-validate the sub-second components and error on ≥ 1 s.

fluree-db-core/src/temporal.rs
```diff
@@ -1986,6 +1986,252 @@ fn parse_seconds_to_micros(s: &str) -> Result<i64, String> {
...
+                "months",
+                "weeks",
+                "days",
+                "hours",
+                "minutes",
+                "seconds",
+                "milliseconds",
+            ])?;
+            let months = int("years", 0)? * 12 + int("months", 0)?;
+            let days = int("weeks", 0)? * 7 + int("days", 0)?;
```
aaj3f: (and line 2117 and 2204)

`ms * 1_000_000`, `years * 12`, `weeks * 7`, `s * 1000` are unchecked i64 arithmetic on values taken straight from Cypher integer literals. `duration({years: 9223372036854775807})` overflows. Debug/test builds panic *during query lowering* (a DoS-shaped failure), release builds silently wrap to a wrong value (root `Cargo.toml [profile.release]` sets no `overflow-checks`). 

Fix: use `checked_mul`/`checked_add` and return a `LowerCypherError`/temporal error on overflow, as `const_list_value` already does for `range()`.

fluree-db-cli/src/commands/load.rs
```diff
@@ -0,0 +1,288 @@
...
+    template: Template,
+    batch_size: usize,
+    field_terminator: &str,
+    dirs: &FlureeDir,
+    remote_flag: Option<&str>,
+    direct: bool,
+) -> CliResult<()> {
+    if batch_size == 0 {
+        return Err(CliError::Usage("--batch-size must be at least 1".into()));
+    }
```
aaj3f, line 76: Lower-bounded (rejects 0) but no upper bound, so `--batch-size 100000000` is a user-inflicted OOM. Consider a sane cap or a doc note. It's minor though, as it's fully user-controlled.

fluree-db-api/src/cypher_procedures.rs
```diff
@@ -0,0 +1,612 @@
...
+                Expr::Lit(Literal::Null(span)),
+                int(0), // left
+                int(0), // right
+                int(0), // leftCount
+                int(0), // rightCount
+                Expr::List(other.clone(), span),
+                Expr::List(other, span),
+                lit("node"),
+            ]);
+        };
```
aaj3f, line 325: `apoc.meta.data` walks novelty twice with per-subject `BTreeSet`/`Sid` allocation; bounded by novelty size (not the full ledger), so acceptable, but a transient cost right after a large `fluree load` before reindex — consider a single-pass build if it shows up in profiles.

aaj3f (approved):
This is nicely done! I'm pretty thrilled by the Cypher parity you've helped us achieve

PR #1457, opened 2026-07-08
fix(policy): honor policy-class when identity is also present (same-ledger)

bplatz:
## Problem

A request carrying **both** `identity` and `policy-class` silently ignored the classes. In `build_policy_context_from_opts`, the same-ledger identity arm wins the priority chain and selects rules only from the identity node's own `f:policyClass` triples in the ledger. Two failure modes fall out:

- An identity with no subject node in the ledger (`IdentityLookupResult::NotFound`) yields an **empty policy set** — deny-all under default-deny.
- An identity that is not a resolvable IRI at all (bare emails, UUID subjects minted by application auth systems) can **never** carry `f:policyClass` triples, so the combination can never work for those callers.

Gateways that resolve grant-derived policy classes per request and forward them alongside the authenticated identity (the documented `fluree-identity` + `fluree-policy-class` header pair) depend on the classes being honored. Today that combination enforces nothing — it denies everything.

## Fix

New arm in the priority chain: when both fields are present, the **classes select the policy set** and the **identity is bind-only** — it resolves to populate `?$identity` for `f:query` rules and never drives rule selection. This mirrors the documented cross-ledger identity contract ("identities are bind-only under cross-ledger"). An unresolvable identity simply leaves `?$identity` unbound. Inline `opts.policy` still merges on top.

Identity-only and class-only behavior are unchanged.

## Tests

New integration test (`it_policy_tx::identity_with_policy_class_selects_class_policies`) covering three contracts:

1. Stored class-typed policies + a non-IRI identity + `policy-class` → insert allowed via the class-selected property whitelist
2. Same identity **without** `policy-class` → identity-mode stays default-deny (pins that the allow comes from the new arm, not a loosening of identity-mode)
3. Non-whitelisted property with the class selected → still denied

Full policy group: 60/60 passing.

### DIFF
--- a/fluree-db-api/src/policy_builder.rs
+++ b/fluree-db-api/src/policy_builder.rs
@@ -99,7 +99,8 @@ pub fn resolve_policy_source_g_ids(
 /// 2. **policy_class**: Query for policies of the given class types
 /// 3. **policy**: Parse inline policy JSON-LD
 ///
-/// Priority: identity > policy_class > policy
+/// Priority: (identity + policy_class: classes select, identity binds) >
+/// identity > policy_class > policy
 ///
 /// # Arguments
 ///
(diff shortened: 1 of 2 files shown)


fluree-db-api/src/policy_builder.rs
aaj3f, line 263: The priority comment ("`cross-ledger restrictions > identity > policy_class > policy > policy_values`") and the function-level doc at `policy_builder.rs:102` ("`Priority: identity > policy_class > policy`") were not updated to reflect the new combined arm, in which — when both are present — `policy_class` selects and `identity` is bind-only (i.e. class effectively outranks identity for the both-present case). For a security-relevant module where precise priority is the contract, consider adding one line noting the `(identity + policy_class)` case sits above identity-only. Non-blocking.

aaj3f (approved):
Good catch & good fix!

PR #1459, opened 2026-07-08
Raft implementation hardening

zonotope:
Builds on `fix/raft-query-peers` with a comprehensive correctness, durability, security, and liveness hardening pass over the Raft consensus layer, following a full review of the implementation. Scoped to the deployment assumptions: **reliable shared content storage, VPC-internal Raft/private listeners, external clients only via a load balancer.**

36 files, +3,591/−846. No behavioral change to non-Raft (single-node/peer) deployments except where noted; the Raft wire format is not yet deployed, so log/command representation changes carry no compatibility constraint.

## Correctness & data safety

- **Durable log-file removal and gap-safe recovery.** `truncate_from`/`purge_through` now fsync the log directory after their unlink batches, and `append` batches its per-entry data syncs behind a single directory fsync (2N → N+1 fsyncs). Because a crash mid-batch can leave tail renames durable out of order, `log_state` now reports the *contiguous prefix* above the purge cutoff (anchored at `cutoff+1`) rather than the max index — so openraft never sees a `last_log` past a missing entry. A dangling `current` snapshot pointer now hard-fails as corruption instead of silently booting fresh over a purged log, and superseded snapshots (plus stale `.tmp` files) are reclaimed after each snapshot write.
- **Orphaned-envelope releases.** Anonymous (keyless) submissions previously leaked their request-envelope blob in shared CAS on every commit; the state machine now releases it as the entry retires. Keyed entries continue to hold it until idempotency eviction.
- **Typed errors through the RPC stack.** Raft RPC responses now carry `postcard(Result<Resp, RaftError>)` so openraft's classifier can act on the peer's real outcome — restart snapshot streaming on `SnapshotMismatch`, stop on `Fatal` — instead of collapsing everything to an opaque 500 and treating it as transient.
- **Settled-vs-ambiguous outcome plumbing.** Unresolved proposals are distinguished from storage-layer errors and from settled rejections; commit blobs are retained unless the failure class is a guaranteed rejection; gateway-class (502–504) submission errors are no longer cached as terminal; already-stashed ferries are retried (via a refresh-and-re-stage path) rather than dropped, closing a queue-stall window.
- **Retraction & lineage guards.** CAS-ref commands respect the retracted tombstone; `drop_branch` (and its rename target `PurgeBranch`) refuse a branch that still has children across all nameservice backends, failing loud instead of stranding a child with a dangling parent pointer.
- **Cache coherence on write.** Commit-head advances are reported into the ledger cache's watermark synchronously and losslessly (not only via the lossy event bus), so a cache hit racing an apply can't serve state the replicated head has moved past.

## Security & DoS hardening

- **SSRF/loopback checks canonicalize IPs** (`to_canonical()`), closing the IPv4-mapped-IPv6 bypass (e.g. `[::ffff:169.254.169.254]` reaching the metadata service).
- **Request body limit enforced end to end.** The configured `--body-limit` now actually binds (the custom credential extractor honored no limit), and follower→leader forwarding refuses oversized bodies with 413 before buffering, sized from the same config.
- **Per-ledger admission map is bounded.** It previously grew one never-removed entry per distinct client-chosen ledger string (a memory-exhaustion vector, since admission runs before any existence check); idle per-ledger semaphores are now evicted, bounding the map to ledgers with in-flight submissions.
- **Auth layering** on the admin-protected write routes is documented and pinned by test (auth runs before leader-forward, so unauthenticated writes are rejected locally rather than relayed).
- **Hop-by-hop header handling** is consolidated into one RFC-2616 §13.5.1 classifier shared by both forwarders, closing drift where each stripped a different subset.

## Liveness & operations

- **Liveness measures replication lag as distance, not "did the match index move."** A follower inching forward one entry per sample while the leader gains thousands is now correctly demoted (its branch queues no longer back up indefinitely), via a configurable healthy-lag window on top of the existing time hysteresis.
- **Graceful shutdown on SIGTERM/SIGINT** with a bounded grace period and a release drain.
- **Index-head publishing retries against the new leader** during a leader change instead of failing.
- **Queue lifecycle fixes:** a fresh enqueue after an admin clear no longer eats a spurious `QueueCleared` abort; drained queues are removed from the map (fixing a slow leak and keeping the global-depth check proportional to live branches).

## State-machine determinism & wire safety

- **Postcard-safe stored status/config.** The HTTP-shaped status/config payloads (`#[serde(flatten)]` + `HashMap` extras) could not be postcard-encoded, so a reachable `push_config` would Raft-fatal the node. They now project through a `CanonicalValue` tree (sorted `BTreeMap`, no JSON text in the replicated layer); conversions are exhaustive-destructuring `From` impls so schema drift is a compile error, and CAS equality is structural.
- **Order-independent idempotency CIDs.** `canonical_body_bytes` no longer hashes payloads in `serde_json` insertion / `HashMap` iteration order, which was silently defeating idempotency dedup across retries and nodes.
- **Ledger IDs are canonicalized at the nameservice boundary** so retries that elide the default branch resolve identically.

## Performance

- Append fsync count cut from 2N to N+1 per batch (above); `apply_head` mutates the branch head-ref in place instead of cloning-and-rebuilding.


bplatz (approved):
Great!

PR #1460, opened 2026-07-08
feat: policy write verbs and multi-language conditions

bplatz:
## Summary

This series makes write intent first-class in the policy engine and unifies policy conditions across query languages. It addresses the POC findings that forced write policies into materialized property allow-lists:

1. **"May create new X" was inexpressible** — class targeting resolved from pre-transaction state, where a new subject has no class.
2. **`rdf:type` object values couldn't be constrained** — a holder whose allow-list included `rdf:type` could mint any class.
3. **`f:onProperty` was copy-by-value** — no way to reference a property set.

> **Stacked on #1457** (`fix/policy-class-with-identity`). Note the base branch's latest commits (the reworked #1457 fix + a `feature/cypher-import` merge) are not pushed yet, so the diff shows extra commits until the base is updated.

## Write verbs (`f:create` / `f:update` / `f:delete`)

`f:action` gains three verbs that classify each subject's **lifecycle** within the transaction — (exists pre, exists post): create (no, yes), update (yes, yes), delete (yes, no). Every staged flake inherits its subject's verb, so these are entity-lifecycle verbs, not flake-op verbs: changing a value (retract old + assert new) is entirely `f:update`; removing the subject outright is `f:delete`. Matches the SQL intuition (`UPDATE` can `SET NULL`; `DELETE` removes the row).

Verb policies get exact semantics on two axes:

- **Class targeting is pre ∪ post** — "may create new Leads" is one policy, and a deny can't be escaped by un-typing the subject in the same transaction.
- **`rdf:type` flakes match by the class they mint or remove** — a create grant scoped to Lead can never mint Contract, even smuggled alongside a legitimate Lead typing (`@type: ["ex:Lead", "ex:Contract"]` is denied). Under default-deny, class-scoped create grants are class-mint constraints, like SQL `GRANT INSERT ON <table>`.

**Bare `f:modify` keeps its legacy semantics** (all writes, pre-state class targeting). The documented immutable-records pattern — which relies on creates escaping class denies — is pinned by a test and behaves unchanged.

## Condition languages — every `f:query` is an ASK

- **SPARQL** (restores previously stashed work): `f:sparql` datatype, ASK/SELECT as existence checks, `$this`/`$identity` per SHACL-SPARQL convention, fail-closed on parse errors or CONSTRUCT/DESCRIBE/updates.
- **JSON-LD `ask` form**: `{"ask": <pattern>}` is the preferred spelling, normalizing onto the same path as the legacy `{"where": ...}` form (which is unchanged, permanently). Both keys together fails closed.
- **Cypher** (`f:cypher` datatype): read-only MATCH…RETURN, ≥1 row = condition holds; writes/schema/procedures rejected at build time, fail closed. `$this`/`$identity`/`$value`/`$op` arrive as idiomatic Cypher parameters (compare with `id(n)`/`elementId(n)`). Condition language is orthogonal to request protocol — a Cypher condition governs JSON-LD and SPARQL traffic identically, since enforcement is flake-level.
- **`$value` / `$op` bindings** (all languages): the flake's object and operation (`"assert"`/`"retract"`). Value gates like `ASK { FILTER($op = "retract" || $value = "qualified") }` constrain what may be written while exempting the retract half of a value change.
- **`f:queryState`** (`f:preState` default / `f:postState`): evaluates the whole condition against pre- or post-transaction state. "May create Leads owned by the requester" needs post-state — the `ex:owner` value is asserted in the same transaction. On the read path the two coincide, keeping dual-action policies portable. Deployed conditions keep their exact meaning (pre-state default).

## Fixes

- **Inline `opts.policy` merged inconsistently across selection arms** (reported): identity-only selection silently dropped inline policies (deny-all under default-deny); class-only treated stored-vs-inline as either/or. Inline policies now merge once, after the selection chain, in every mode. Same family as #1457.
- **Modify-set class targeting was stats-derived and unsound both ways**: a class-scoped allow didn't cover properties the class had never used in committed data (denied under default-deny despite an explicit allow), and the exclusive-property shortcut skipped the class check exactly when the staged flake was the first counterexample — a class-scoped allow permitted writing the "exclusive" property onto instances of other classes (confirmed end-to-end; the attack passed before this fix). Modify sets now index class policies by class (`PolicySet::by_class`): exact, predicate-agnostic, no stats dependency. This also makes cross-ledger modify class policies effective for the first time (the wire path has no stats handle, so they previously covered only `@id`/`rdf:type`).

## Testing

- 27 new integration tests across `it_policy_verbs.rs`, `it_policy_cypher_conditions.rs`, `it_policy_sparql.rs`, `it_policy_query_connection.rs`, `it_policy_tx.rs`, plus policy-crate unit tests.
- Regression tests for the two bug fixes were verified to **fail against the pre-fix code**.
- Behavior-pinning tests: legacy `f:modify` create-escape, pre-state default for conditions (the post-state test includes a control proving the selector — not evaluation order — switches the state), fail-closed paths for every language (garbage source, write statements, ambiguous `ask`+`where`, unbound identity).
- Full workspace `--all-features` run green (the only non-passes are the known LocalStack testcontainer timeouts, which need a Docker daemon).

## Docs

`policy-model.md` (action verbs, condition languages, bindings, `f:queryState`), `policy-in-transactions.md` (Write verbs section, verb-form immutable-records pattern, exact write-side class targeting), vocabulary reference.

## Deferred

- In-query state mixing (`GRAPH f:postState { ... }` inside one SPARQL condition) — needs per-pattern overlay switching in the execution context; the per-policy selector plus `$value` covers the practical cases.
- Schema-flake write bypass (`rdfs:subClassOf` self-minting under always-on policy entailment) — separate follow-up.


fluree-db-query/src/policy/executor.rs
```diff
@@ -59,24 +76,184 @@ impl<'a> QueryPolicyExecutor<'a> {
...
+                val: literal.clone(),
+                dtc: DatatypeConstraint::Explicit(dt_sid),
+                t: None,
+                op: None,
+                p_id: None,
+            },
+            None => Binding::Unbound,
+        },
+    }
+}
```
aaj3f: `binding_for_value` seeds the never-match unbound-identity sentinel as `Binding::Unbound` (UNDEF). The comment says "conditions on it never hold, failing closed" — which is true for a `FILTER($identity = …)` comparison, but **false for a positional binding**: in a pattern like `ASK { $identity <ex:user> $this }`, an UNDEF `$identity` is an *unconstrained* variable that matches any row (VALUES-UNDEF is "compatible with anything": `values.rs:108` → `join.rs` takes the scanned value). So the identity constraint vanishes and the condition holds for everyone. 

This is exactly this PR's canonical stored-policy idiom (`fluree-db-api/tests/it_policy_sparql.rs:59`), and the SPARQL/JSON-LD tests exercise it only with a *bound* identity. **Concrete exploit (reproduced against this branch's HEAD):** the `required` `f:view` SSN policy `ASK { $identity <ex:user> $this }`, queried by an **anonymous** request (`GovernanceOptions { policy_class: [EmployeePolicy], default_allow: false, identity: None }`), returned `[["…/alice","111-11-1111"]]` — Alice's `required`-gated SSN leaked to a caller with no identity. 

Same mechanism defeats **write** verb policies (e.g. "may create Leads where `$this ex:owner $identity`" authorizes creating Leads owned by anyone under an unbound identity — `stage.rs:1464` → `allow_modify_flake_write_async_detailed` → `combine_applicable_async_detailed` → same `build_policy_values_clause`/seeding). 

**Fix direction:** seed the unbound-identity marker as a concrete never-match `Binding::Sid` (the URN sentinel it already is, from `ensure_ground_identity` at `evaluate.rs:943-951`) rather than `Binding::Unbound`, so a positional binding finds no match (fail-closed) while `FILTER` comparisons stay false too; mirror the same on the JSON-LD null-injection path (`executor.rs:319-325`). Add unbound-identity fail-closed tests for **SPARQL and JSON-LD positional** conditions (today only Cypher's unbound case is covered — `it_policy_cypher_conditions.rs:175`). 

Note: the JSON-LD VALUES-UNDEF behavior pre-dates this PR, but this PR introduces the identical seeding for the new SPARQL surface, ships the positional idiom as the documented example, and extends it to the write path — so it is in scope here.

fluree-db-query/src/policy/executor.rs
```diff
@@ -59,24 +76,184 @@ impl<'a> QueryPolicyExecutor<'a> {
...
+        FlakeValue::Long(_) | FlakeValue::BigInt(_) => xsd_names::INTEGER,
+        FlakeValue::Double(_) => xsd_names::DOUBLE,
+        FlakeValue::Decimal(_) => xsd_names::DECIMAL,
+        FlakeValue::DateTime(_) => xsd_names::DATE_TIME,
+        FlakeValue::Date(_) => xsd_names::DATE,
+        FlakeValue::Time(_) => xsd_names::TIME,
+        _ => return None,
+    };
+    Some(Sid::new(XSD, name))
+}
```
aaj3f: `binding_for_value` / `default_literal_datatype` seed `$value` as `Binding::Unbound` for `FlakeValue` kinds with no "faithful" default datatype (`Json`, `Vector`, `GeoPoint`, the `Duration`/`GYear*` family, `Null`). A `$value` gate that binds `$value` **positionally** (e.g. `ASK { $this ex:stage $value }`) on a property of such a kind hits the same UNDEF-matches-anything path and is bypassed. Not executed, but it shares the finding-1 root cause; fixing finding 1's positional-UNDEF handling should cover this too. Add a `$value` gate test over a non-faithful datatype (e.g. a `@json` or vector object).

aaj3f (approved):
These are nice additions (I remember several times wishing for them in fluree history!). One note below re: possible exploit on identity

PR #1461, opened 2026-07-09
perf: batched frontier expansion for property-path traversals

bplatz:
## What

Property-path traversals (`PropertyPathOperator`) expanded each BFS node with its own index range query: one index descent, full `Flake` materialization, and a dictionary-backed `Sid` per neighbor — per node, per level. On expansion-shaped queries (transitive closures, multi-hop neighborhood reachability over indexed graphs) those per-node probes dominate end-to-end latency.

This PR gives the operator a batched raw-id lane built on the machinery the shortest-path operator already proved out:

- New shared `fluree-db-query/src/frontier.rs`: `PathNode`, the `DirtyIds` overlay summary, and the batched one-hop sweeps (typed PSOT, wildcard SPOT, inbound OPST), extracted from `shortest_path.rs`, which now consumes them.
- Each BFS level expands with a few galloping batched sweeps keyed by persisted `s_id` — no per-node index descents and no dictionary work in the loop.
- Results are emitted as `Binding::EncodedSid`, late-materialized downstream like binary-scan bindings. This part is load-bearing: resolving each result id back to a `Sid` (~2µs apiece) cost more than the batched traversal saved on large result sets, turning a win into a regression until removed.
- Overlay correctness stays per-node: any node the novelty overlay touches on the side being expanded — plus novelty-only nodes — takes the existing per-node Sid fallback, which merges novelty. Time travel is supported (sweeps honor `to_t`).

## Queries it helps

Anything that reaches the property-path operator with an endpoint bound:

- **SPARQL**: `p+` / `p*` in either direction, inverse `(^p)+`, alternations `(a|b)+`, and bound-both reachability filters — e.g. `skos:broader+` / `rdfs:subClassOf*` hierarchy walks over indexed data.
- **Cypher**: unbounded typed variable-length paths (`-[:T*]->`), untyped bounded ranges (`-[*1..3]->`), and fused anonymous-hop runs (`-->()-->(n)` under DISTINCT output).
- **JSON-LD/FQL** path queries, via the shared IR.

The lane declines — keeping the existing per-node traversal — for: no binary index store, active policy, multi-ledger datasets, eager materialization (reasoning views), composite sequence paths (`(a/b)+`), unsummarizable overlays, or a typed predicate that exists only in novelty.

## Results

Measured with `examples/path_expansion_probe.rs` (batched lane vs the prior per-node execution; 200k-node social graph, out-degree 16, scattered edges):

| query | result rows | batched | per-node |
|---|---|---|---|
| 2-hop expansion | 256 | 0.25 ms | 0.57 ms |
| 3-hop expansion | 4,096 | 3.2 ms | 9.6 ms |
| 4-hop expansion | 49,720 | 38.7 ms | 124.7 ms |

On a real social-network dataset (~10k nodes, ~122k edges), 4-hop distinct-neighborhood queries run ~5.6× faster end to end; the traversal-only portion improves ~8×. Row counts are identical between the two executions in every measurement.

## Tests

`fluree-db-api/tests/it_property_path_batched.rs` runs every path form differentially: base index + a novelty tail (retracts, novelty-only nodes, retract-and-re-assert churn, a novelty-only predicate) must produce rows identical to the fully reindexed ground truth, and lane engagement is asserted per query from the expand event (zero per-node fallback on a clean HEAD). Existing path/query suites and the W3C SPARQL property-path suite pass unchanged.

## Also included: update-path stats rebuild fix (fixed here after being pinned by bisection)

A separate commit fixes a write-path regression surfaced while validating this branch: `execute_where_streaming` fetched the cached stats view for **every update's WHERE clause**, and the cache key includes the overlay epoch — which advances on every commit. Under a write-heavy load every update therefore missed the cache and rebuilt the novelty-merged stats view, costing +60% on the internal WHERE of pattern-form deletes (`DELETE WHERE { <s> ?p ?o }`) and ~+6–10% end-to-end on delete-heavy update mixes, worsening with client concurrency. Inserts (no WHERE) and reads (stable epoch between them) were unaffected, which is why it hid in update benchmarks only.

A WHERE that is one plain triple has no join order to choose (index selection is fixed by the bound terms), so the stats view is now skipped for exactly that shape; multi-pattern WHEREs keep the stats-driven planning that anchored update MATCHes rely on. Internal-WHERE probe: 2.76–3.04s → 1.63–1.89s per 300 ops, back to the pre-regression baseline. W3C SPARQL update suite and the transact/query groups pass unchanged.


aaj3f (approved):
These are great improvement numbers! 

Gonna keep acting like automated benches are landing soon, so still going to beg that we update `fluree-db-query/benches` and the `regression-budget.json` so that these wins are preserved.

PR #1462, opened 2026-07-09
feat(sparql): burn-down Wave 3 — within-ledger FROM (D-3), UPDATE graph-management (D-5/6), equality/EBV/promotion (D-12, #1447), USING+GRAPH over-delete (#1441), triple-term syntax (−145 tests)

### LINKED ISSUE #1441: UPDATE USING combined with explicit GRAPH over-deletes from the default graph
`DELETE {...} USING <g3> WHERE { GRAPH <g2> {...} }` deletes rows it must not (W3C `dawg-delete-using-02a`: expected 5 triples to remain, 3 remain). Single-operation, verified; the explicit GRAPH block's scope is polluted by USING dataset selection (`fluree-db-transact/src/stage.rs:1382-1399` region).

Data loss in a supported UPDATE form. Roadmap PR-U6 (sequenced after the GRAPH-operator conformance PR).

### LINKED ISSUE #1447: Numeric aggregates silently skip non-numeric group members instead of erroring to unbound
`AVG`/`SUM` over a group containing a bnode/IRI/string return a value computed over only the numeric members (`aggregate.rs:737-755`, `binding_to_numeric` returning `None` is skipped). SPARQL requires a type error, making the aggregate unbound for that group (W3C `agg-err-01`). Silent wrong aggregates on mixed data, both query surfaces. Roadmap PR-X2 (poison-on-non-numeric accumulator state).

aaj3f:
# Wave-3 SPARQL W3C burn-down — within-ledger datasets, UPDATE graph-management, the equality/EBV/promotion lattice, and SPARQL 1.2 triple-term syntax

This is the capstone wave of the W3C SPARQL test-suite burn-down: seven thematically-coherent PRs integrated onto `burndown/wave-2`. It greens **146 W3C tests** to a suite that is 0-failed / 0-stale in both directions (`cd testsuite-sparql && cargo test` → `36 passed`), shrinking the skip register from 376 to 197 entry lines (343 → 197 unique tests — the tip register carries zero duplicate entries after the `c91f5556c` dedup). The wave is perf-neutral: the integrated A/B vs `burndown/wave-2` is within the 5% budget on all ten measured scenarios (worst +2.34%). The seven PRs are PR-G2 (within-ledger `FROM`/`FROM NAMED`), PR-W1 (nested-group FILTER scope + sub-SELECT correlation), PR-W2BC (SPARQL 1.2 triple-term syntax), PR-W2 (CONSTRUCT blank-node instantiation), PR-U3 (UPDATE graph-management verbs), PR-U6 (USING + explicit-GRAPH delete scoping), and PR-X2 (the equality/EBV/numeric-promotion lattice).

**Now based directly on `main`** — #1454 (Wave 2) has merged, and `origin/main` (through the #1450 Iceberg stack) is merged into this branch (`d072a95d6`), so the PR is conflict-free against `main`.

## Reviewer quick reference

**End-user API / contract changes.** SPARQL `SELECT … FROM <g> FROM NAMED <n>` against a single ledger's graphs now returns results instead of being rejected (PR-G2). Seven SPARQL UPDATE graph-management verbs — `LOAD`/`CLEAR`/`DROP`/`CREATE`/`COPY`/`MOVE`/`ADD` (+ `SILENT`/`INTO`) — now parse and execute, and the capability is exposed on the transact-builder surface (`Txn::clear_graph`/`drop_graph`/`copy_graph`/`clear_default_graph`, `GraphTransactBuilder::clear_graph`/`drop_graph`/`copy_graph`) (PR-U3). SPARQL/JSON-LD `=`/`!=`, effective-boolean-value, `AND`/`OR`, numeric promotion, `CONCAT`, and `SUM`/`AVG` are now spec-strict value semantics (PR-X2 — **DELIBERATE SEMANTICS CHANGE (D-12), 2-reviewer sign-off requested**) — this can return fewer rows for queries that relied on the old permissive behaviour; see the migration notes. `DELETE … USING <g> WHERE { GRAPH <h> … }` now scopes the explicit GRAPH block to the `USING NAMED` set (deletes fewer rows — a data-loss fix) (PR-U6). A `CONSTRUCT` whose template has a blank node now emits triples instead of an empty graph (PR-W2). SPARQL 1.2 triple-term syntax (`TRIPLE`/`SUBJECT`/`PREDICATE`/`OBJECT`/`isTRIPLE` and bare `<<( s p o )>>` values) now parses/validates and defers evaluation with a clean `not_implemented` (PR-W2BC). PR-W1 has no API surface change.

**Performance verdict: WITHIN BUDGET.** Integrated A/B vs `burndown/wave-2`, `FLUREE_BENCH_PROFILE=full` (n=100): all ten scenarios within the 5% budget, worst +2.34% (query `q5`). The whole wave — including the u3/u6 transact/staging path — is perf-neutral. Full table below.

**Production bugs found & fixed (beyond W3C compliance).** (1) `USING`-scoped delete with an explicit `GRAPH` block over-deleted from the default/other graph — a real data-loss bug, **Closes #1441** (PR-U6). (2) Predicate-grouped `COUNT` mis-tagged `xsd:long`; `SUM`/`AVG` silently averaged over the numeric subset of a mixed group; `AND`/`OR` aborted the whole expression on the first operand error; `CONCAT` silently coerced non-strings — **Closes #1447** (PR-X2). (3) A JSON-LD correlated sub-SELECT returned an extra non-conformant row when a correlation variable was bound only via an inner OPTIONAL (PR-W1). (4) PR-U2's cross-operation blank-node scope check over-rejected valid multi-op requests, and the W3C harness rejected a legitimate bare `mf:result []` empty-store expectation (both fixed in PR-U3).

## What this wave does (per PR)

**PR-G2 — within-ledger `FROM` / `FROM NAMED` (D-3, Option A).** Replaces the single-ledger dataset-clause rejection on the buffered query paths with within-ledger dataset construction: each clause IRI is resolved against the ledger's own graph registry (alias → default graph, registered named-graph IRI → that g_id), assembled into a single-snapshot `DataSetDb`, and executed through the existing `DatasetOperator`/`dataset_query.rs` path. A clause IRI that is not a graph in this ledger is still rejected (cross-ledger goes through the connection path). R2RML + within-ledger `FROM` is a documented unsupported combination (the R2RML dataset path is non-`Send`). Also fixes a mutual-async-recursion compile trap the delegation created — `query_dataset_with_options` gets a concrete boxed return (`Pin<Box<dyn Future … + Send>>`), dropping `cargo check` on the crate from 30+ min to ~7s.

**PR-W1 — nested-group FILTER scope + sub-SELECT correlation.** Two independent evaluation fixes. Family A: a single-child `{ }` group whose child is scope-sensitive (`FILTER`/`BIND`/nested `Group`) stays wrapped as an uncorrelated subquery rather than being flattened into the enclosing scope (so a nested FILTER no longer sees an enclosing-scope variable). Family B: a sub-SELECT correlation variable bound only via OPTIONAL/UNION is reconciled at merge (excluded from the per-row seed so it binds independently, then dropped on a conflicting double-binding) — the compatible-mapping join SPARQL §18.4 requires. Common self-produced-correlation and BGP/OPTIONAL paths (incl. BSBM) stay byte-identical.

**PR-W2BC — SPARQL 1.2 triple-term syntax (D-1, accept-then-defer).** The lexer emits a single contextual `TripleTermFn` token for the five function names (not reserved keywords), the expression parser resolves + arity-validates them, and a new `parse_triple_term_value` accepts `<<( s p o )>>` in the subject/object/`VALUES`/`BIND` positions RDF 1.2 allows; lowering rejects all deferred forms with `not_implemented`. The §2 negative-suite guardrails (no path/collection/reified-triple inside a term; value-context vs pattern-context subject rules; bare `<<( )>> .` is not a statement) keep every negative test rejected. No new evaluable capability — this is syntax acceptance only.

**PR-W2 — CONSTRUCT per-solution blank-node instantiation.** A CONSTRUCT template blank node (`[ … ]`, `_:a`, or a PR-1 collection cell) previously lowered to a never-bound variable and produced an empty graph. The fix records template blank-node variables at lowering (`ConstructTemplate.bnode_vars`) and mints a fresh blank node per solution row in the CONSTRUCT output path — shared within a row, distinct across rows — under a reserved `cst{n}` label prefix provably disjoint from stored (`fdb-`), `BNODE()` (`b{hex}`), and SERVICE blanks (a soundness hardening the isomorphism-based suite cannot catch).

**PR-U3 — SPARQL UPDATE graph-management verbs (D-5, D-6).** Grammar + AST + parse + lowering to a new `GraphMgmtOp` Transaction-IR directive + a whole-graph-scan staging primitive (`stage_graph_mgmt`). A named graph is a reserved `GraphId`; the execution primitive is a scan of a graph's flakes — clone-as-retraction for CLEAR/DROP, clone-with-rewritten-`g` for COPY/MOVE/ADD (datatype/lang/list-index ride along verbatim). CLEAR/DROP ALL|NAMED resolve their target set from the registry; COPY/MOVE compose over CLEAR with a set-difference so overlapping triples are never simultaneously asserted and retracted. The dispatch is a single `if txn.graph_mgmt.is_some()` check at the top of `stage()`, before any hot-path setup, so ordinary insert/upsert/update staging is byte-identical.

**PR-U6 — USING + explicit-GRAPH dataset scoping (class F, #1441).** SPARQL 1.1 §3.1.3: when `USING`/`USING NAMED` is present it defines the WHERE dataset exactly — the WHERE-visible named graphs are precisely the `USING NAMED` set (empty under a plain `USING <g>`). The bug treated an empty `USING NAMED` set as "no restriction" and registered every named graph, so `DELETE { … } USING <g3> WHERE { GRAPH <g2> { … } }` matched `g2` and over-deleted. The fix restricts the runtime dataset's named graphs to exactly the `USING NAMED` set whenever any `USING`/`USING NAMED` clause is present; the no-`USING` path (plain `DELETE WHERE { GRAPH <g> … }`) stays byte-identical. Prepare-time only; SELECT queries never reach it.

**PR-X2 — equality / EBV / numeric-promotion lattice (D-12). DELIBERATE SEMANTICS CHANGE — 2-reviewer sign-off requested.** Five IR/engine value-semantics fixes on the shared `fluree-db-query` evaluator: D4 numeric promotion (`ComparableValue` gains first-class `Float(f32)` so `integer < decimal < float < double` holds and `xsd:float` no longer widens to `xsd:double`); D-EBV (a fallible, datatype-aware bare-variable effective-boolean-value — numeric-zero/empty-string falsy; lang-tagged/foreign-datatype/IRI/ill-typed/unbound → a type error excluding the row); D5/D7 (datatype-aware three-valued `rdf_term_equal` for `=`/`!=`, with §17.2 three-valued `AND`/`OR`); D11 (`CONCAT` type-errors on a non-string argument); and aggregates (predicate-grouped `COUNT` → `xsd:integer`; `SUM`/`AVG` poison-to-unbound on a non-numeric member). The common int-int/iri-iri fast paths are kept first; the added per-row cost is one well-known-Sid tag-check on the xsd:double and xsd:string literal paths plus the `EqOutcome` wrapping on `=`/`!=`, measured neutral (`query_hot_bsbm_bi/f2` −0.51% on X2's own **standalone** branch-vs-wave-2 n=100 A/B; the **whole-wave integrated** f2 is +1.80% in the perf table below — both well inside the 5% budget, and the difference between the two is measurement, not an X2 cost, since no other wave-3 PR touches the FILTER/eval hot path f2 exercises).

## Decisions actioned in this wave

Per the standing transparency rule (ROADMAP §4), each decision point, the options weighed, and why the adopted option was chosen.

**D-1 — SPARQL 1.2 triple-term syntax: accept-then-defer (PR-W2BC).** Options: (4) accept + arity-validate the syntax, lower to `not_implemented` [adopted]; (3) reject-on-principle as a documented divergence; (2) desugar `<<( )>>` to the reifier model (dominated — a triple term is a value, a reifier is a resource, so `isTRIPLE`/`SUBJECT`/term-equality would silently diverge). Chosen Option 4: it makes Fluree parse standard SPARQL 1.2 today at zero engine/perf risk — exactly what a syntax-conformance suite rewards — while honestly deferring evaluation to the Option-1 first-class-value epic. The trade-off (a runtime `not_implemented` where there used to be a parse error) is named explicitly; failing fast at parse time is Option 3, a genuine either/or the team owns.

**D-3 — within-ledger `FROM`/`FROM NAMED` datasets (PR-G2).** Options: (A) engine within-ledger `DataSet` reusing the runtime dataset path [adopted]; (B) harness ledger-per-graph with `query_connection_sparql` (zero engine change). Both green the same 13 tests, so the tie-breakers are product value, reuse, and perf — all favour A. A fixes the actual user-facing gap (a real user can query one ledger's named graphs), extends #1279's within-ledger registry the natural way (a named graph *is* a g_id), and keeps a single snapshot (no cross-ledger provenance stamping, binary store stays enabled); B only satisfies the harness and spins up N ledgers per query.

**D-5 — remote (non-SILENT) LOAD: documented divergence (PR-U3).** Options: (a) parse LOAD, execute `SILENT` as a swallowed no-op, error on non-SILENT [adopted]; (b) add an opt-in HTTP fetch hook into the transact path. Chosen (a): it costs zero W3C coverage (the two eval tests LOAD a non-resolvable source under SILENT and expect the store unchanged; the syntax tests are parse-only), whereas (b) would drag an HTTP client and its failure/security surface into the embedded write path for no test benefit (ingesting a known document already has a first-class path — the insert API).

**D-6 — DROP ≡ CLEAR + the empty-named-graph model (PR-U3).** First half — DROP is implemented as CLEAR semantics now (the `GraphRegistry` is additive-only, and the two are indistinguishable to the harness, which compares/lists only non-empty graphs). Second half (a spotlighted product call) — whether a CLEAR'd/CREATE'd zero-triple named graph is enumerable by `GRAPH ?g`/FROM NAMED: Option A (persist a graph-existence record decoupled from flakes across storage/commit/query, greens `bindings#graph` + `agg-empty-group-count-graph`, −2) vs Option B (documented permanent divergence — a graph exists iff a flake carries its g_id) [adopted]. Chosen B: the capability is a query/dataset-model feature orthogonal to the UPDATE verbs, it is the same model fact that makes DROP≡CLEAR correct, and Option A's blast radius (three subsystems, a durable "empty graph" concept Fluree deliberately lacks) is disproportionate to −2 tests. Both tests stay registered with the full rationale.

**D-12 — SPARQL strict-type-error mode vs unified strictness (PR-X2).** Options: (a) a per-surface strictness flag threaded through evaluation; (b) one unified strict value-semantics path shared by all surfaces [adopted for equality/EBV/promotion]; (c) leave lenient. Chosen (b): equality/EBV/promotion are *value* semantics, correct on both the SPARQL and JSON-LD surfaces, and the parity tests assert the strict behaviour on both. Deliberate JSON-LD extensions are preserved by reading the binding directly rather than the surface (the lenient `From<&Binding>` EBVs stay for Cypher structural truthiness; PR-X1 already actioned the D2b `DATATYPE`/`LANG` `@id` carve-out).

D-11 (lexical-form preservation) is NOT actioned this wave — its tests (`group01`, the `sameTerm`/`distinct`/`dawg-str` lexical set) remain registered and are owned by the future PR-X3.

## Second-round review fixes (after the first-round response commits)

A second adversarial pre-review pass over the whole PR produced eight more commits, all included here:

- **`fix(sparql)` — triple-term subject + annotation tail no longer panics.** Parse-accepted input like `INSERT DATA { <<( :s :p :o )>> :q :o2 {| :a :b |} }` reached an `unreachable!` in the annotation-expansion pre-pass (it guarded QuotedTriple subjects but not TripleTerm); it now defers with the same clean `UnsupportedFeature` as every other deferred triple-term shape, with negative tests for the INSERT DATA and Modify-template paths.
- **`fix(query)` — NaN is never RDFterm-equal.** `"NaN"^^xsd:double = "NaN"^^xsd:double` was true via `numeric_cmp`'s bit-level total order (kept for ORDER BY); `=` now special-cases NaN per XPath op:numeric-equal (false; `!=` true), scoped to numeric operands, ORDER BY unchanged. Pre-existing bug surfaced by the X2 equality review.
- **`fix(sparql)` — `USING NAMED` without `USING` gives the WHERE an empty default graph** (§13.2.1 via §3.1.3). Previously the WHERE default fell through to the ledger's real default graph, so `DELETE { ?s ?p ?o } USING NAMED <h> WHERE { ?s ?p ?o }` deleted the whole default graph — the #1441 over-reach class one clause over. The no-USING and plain-USING paths are byte-identical; test proves fail-without-fix.
- **`fix(query)` — three evaluator edges:** unary minus now coerces cast results (`-xsd:float("1.5")` was unbound); CONCAT keeps a language tag carried on the VALUE (constants/STRLANG results — `CONCAT("a"@en, "b"@en)` was plain "ab") with case-insensitive tag comparison; ROUND/CEIL/FLOOR pass `xsd:integer` beyond i64 through instead of unbinding.
- **`feat(transact)` — builder `move_graph`/`add_graph`** complete the transfer-verb parity, and the reserved-graph guard now also fires on same-graph transfers (`COPY <…#config> TO <…#config>` is refused, not a blessed no-op).
- **`chore(query)` — blank-free CONSTRUCT templates skip the per-row bnode probes; `triple_term_fn_arity`'s catch-all is now a `debug_assert`.**
- **`test` — coverage adds:** three-valued AND dominance (both surfaces), SUM-poison pair, an indexed (reindexed, `range_provider`-asserted) SUM/AVG-poison twin, same-graph no-op transfers for all three verbs on both surfaces, multi-op requests mixing data + graph-management ops, CLEAR of an annotation-bearing graph, two-level nested sub-SELECT correlation + GROUP BY-keyed correlation, and per-row (not per-distinct-binding) CONSTRUCT blank minting.
- **`docs` — recorded the CONSTRUCT blank-node-template parity exemption** in `sparql-compliance.md` (the FQL construct template is a JSON-LD node map with no per-row fresh-blank syntax), plus stage-side notes: SILENT-with-missing-source still clears the destination, and O3 source-existence deliberately uses registry semantics vs the query surface's flake-carried D-6 model.

## Third-round: bplatz review response + the origin/main merge

All eleven review comments are resolved on the branch (per-comment replies inline). Three were already fixed by the second-round commits that landed just before the review posted (the triple-term panic guard, the `USING NAMED` empty-default fix, NaN equality). The six new commits (`8d8870ba1..2205252c5`, after the merge):

- **`fix(sparql)` — FROM/FROM NAMED reject the reserved system graphs.** The registry legs of within-ledger FROM resolution resolved `#config`/`#txn-meta` (the registry seeds them), exposing the governance graph where the `GRAPH <iri>` path blocks it; both legs now apply the same `>= FIRST_USER_GRAPH_ID` filter, with a 4-case rejection test.
- **`fix(query)` — strict typed-literal EBV on both eval paths; STRDT keeps Sid datatype args.** Cast/computed numerics (`xsd:float("1.5")` is a String tagged xsd:float) now have their §17.2.2 EBV on the BIND'd-binding path AND the function-call path (whose lenient conversion read every typed literal as truthy); ill-formed numeric/boolean lexicals are EBV false per rule 1. Fixing the tests exposed that STRDT's datatype argument (a Sid — constant IRIs lower through the encoder) was silently dropped, producing plain literals; Sid args are now decoded.
- **`fix(cypher)` — pin CALL imports as lateral seeds.** Merge fallout: main's newly-wired Cypher suite caught W1's Family-B rule breaking `CALL (p) { OPTIONAL MATCH … RETURN count(…) }` zero-match retention (openCypher imports are per-row bindings, not inferred correlations). `SubqueryPattern.pinned_vars` carries the explicit imports — set only by Cypher CALL lowering, so SPARQL/JSON-LD Family-B semantics (join-scope-1) are untouched.
- **`feat(sparql)` — CREATE GRAPH registers the graph.** CREATE recorded nothing, so `CREATE GRAPH <g>` then non-SILENT `COPY <g> TO <h>` hit O3's "source does not exist" error; CREATE now registers the IRI (empty, non-enumerable — D-6), with scoped exceptions so a registration-only zero-flake commit persists while ordinary no-op updates still elide. CREATE-of-existing stays a no-op (documented D-6 divergence), and the CLEAR/DROP `silent` flag carries a doc note on why it is never consulted.
- **`fix(sparql)` — unforgeable INSERT-template anon labels; FROM build-time set-dedup; typo-aware dataset rejection.** The transact-side counter minted template `[]` blanks as `_:b{n}` (a user's `_:b0` fused with them); now `_:[]{n}` per main's scheme. Repeated `FROM <g>` contributes one member (§13.2 set). The unresolved-IRI message now points at typos before the cross-ledger API.
- **`test(transact)` — pin the schema-flake CLEAR policy bypass; document the whole-graph novelty cliff.** A view-only default-deny identity CAN clear a schema-only graph (the `is_schema_flake` exemption) — pinned beside its rejected non-schema companion; `scan_graph_flakes` documents the one-commit novelty-doubling cliff for whole-graph ops. (Also verified, no change: `ReservedGraphTarget`/`SourceGraphNotFound` map to 400, not 500.)

**The merge itself** (`d072a95d6`) reconciled two same-problem-solved-twice collisions with wave-2's post-branch review commits now on main: the anonymous-blank hardening (main's `_:[]{N}` scheme adopted branch-wide; the wave-3 `#anon` counter dropped) and the eval-triple-terms register dedup (each side had removed the other's kept copy of six double-registered entries — restored once each; the suite's both-direction enforcement caught it). Register content is unchanged by the merge: same unique set, 0-failed / 0-stale.

## Register entries removed (146 unique / 179 entry lines; suite-driven, both directions)

- **PR-G2 (13):** `SPARQL10_QUERY_EVAL` `dawg-dataset-01..12b` (12) + `SPARQL11_CONSTRUCT` `constructwhere04`.
- **PR-W1 (3):** `SPARQL10_QUERY_EVAL` `filter-nested-2`, `dawg-optional-filter-005-not-simplified`, `join-scope-1`.
- **PR-W2BC (27):** `SPARQL12_SYNTAX_TRIPLE_TERMS_POSITIVE` → empty (`basic/bnode/compound/expr/inside/nested/subject/update`-tripleterm + `compound-all`).
- **PR-W2 (3):** `SPARQL10_QUERY_EVAL` `construct-3`, `construct-4` + `SPARQL11_CONSTRUCT` `constructlist`.
- **PR-U3 (68 unique tests / 91 register lines):** the remaining `SPARQL11_UPDATE` graph-management eval tests (`add/clear/copy/create/drop/load/move` + `-silent` variants, the 4 same-bnode joint tests) and the 23 double-registered `SPARQL11_SYNTAX_UPDATE_1` `test_*` (both copies deleted per the standing rule, so 23 tests remove 46 lines; that register is now empty). Per-PR unique counts sum to 145 (13 + 3 + 27 + 3 + 68 + 2 + 29); with the `SPARQL12_EVAL_TRIPLE_TERMS` intra-const dedup (`c91f5556c`) the wave removes **146 unique tests / 179 raw register lines** in total (independently recomputed from the git diff at tip; zero entries added).
- **PR-U6 (2):** `SPARQL11_UPDATE` `dawg-delete-using-02a`, `dawg-delete-using-06a` (empties the last class-F cluster).
- **PR-X2 (29):** `agg02`, `agg-err-01`, `concat02`, `strlang03-rdf11`, `dawg-langMatches-1/2/3/basic`, `type-promotion-03/04/05/21/29/30`, `{add,subtract,multiply,divide}-numbers-cast`, `unplus-2`, `unminus-2`, `dawg-bev-1..6`, `eq-2-1`, `eq-2-2`, `open-eq-04`.

## Query-surface parity

Per `docs/contributing/sparql-compliance.md` §"Query Surface Parity" (JSON-LD only; Cypher excluded — we do not own the openCypher grammar): every IR/engine-level fix carries a JSON-LD regression test alongside the SPARQL one, since the W3C submodule guards only the SPARQL surface. PR-X2 (value semantics), PR-W1 (scope/correlation — incl. the JSON-LD-only correlated-subselect bug fix), and PR-G2 (within-ledger dataset, a single construction over both surfaces) are IR/engine (case a) with JSON-LD tests in `it_query_expression_semantics.rs`, `it_query_filter_scope.rs`, and `it_query_dataset.rs`. PR-U3 exposes the new graph-management capability on the transact-builder surface in the same PR (`it_named_graphs.rs::test_graph_mgmt_transact_builder_parity`); a JSON-LD *document* key syntax is a deliberately-deferred surface-design decision. PR-U6 is a SPARQL `USING` surface fix with a JSON-LD `fromNamed` guard test on the shared machinery (the symmetric JSON-LD `from`-without-`fromNamed` restriction is a scoped follow-up). PR-W2 (CONSTRUCT bnode) and PR-W2BC (parse-only) are SPARQL-only surface (case c) — the FQL construct form has no blank-node template syntax, and accept-then-defer introduces nothing newly executable.

## Performance (integrated A/B vs `burndown/wave-2` @ 3040d8e6f)

`FLUREE_BENCH_PROFILE=full FLUREE_BENCH_SCALE=small` (real n=100; `query_hot_bsbm.rs:166` sets `sample_size` from the profile, so the env var is the real knob). Query benches whole-bench interleaved; insert/import benches per-scenario fine-interleaved (wave-2↔wave-3 back-to-back) to kill cross-run drift; self-gated on rustc=0; contended runs discarded. Δ% = (wave-3 − wave-2)/wave-2 (positive = wave-3 slower); budget 5% at `small`.

- query_hot_bsbm: q3 −2.27%, q5 +2.34% (worst), q9 +0.80% — CIs 0.4–0.6%.
- query_hot_bsbm_bi: f2 +1.80% — CI 0.4/0.6%.
- insert_formats: jsonld/100nodes −1.28%, jsonld/10nodes +0.40%, turtle/100nodes +0.10%, turtle/10nodes +0.98%.
- import_bulk: single_threaded −0.72%, default_threads +0.24%.

All within budget. An initial coarse-interleaved pass showed two apparent breaches (jsonld/100nodes +7.33%, import single_threaded +12.31%) that were foreign-build contention landing in the longer wave-3 windows, not a code regression: no wave-3 change touches the data-insert/bulk-import hot path (fluree-db-core + fluree-graph-turtle untouched; u3's shared-`stage()` change is a single per-transaction guard, byte-identical for inserts; u6's is behind `sparql_where`), and the clean fine-interleaved re-run collapsed both to −1.28% / −0.72% with tight CIs.

## Deferrals (staying registered, with per-entry pointer comments)

The equality cluster's remaining tests are blocked outside the equality/EBV/promotion lattice and stay registered with X2-owned notes: the D5b scan-path `EncodedLit` datatype-carry family (`eq-4`, `open-eq-02/05/06`, `dawg-lang-3`, `quotes-3/4`, `eq-graph-1/2/4` — bare-BGP object term-match, deliberately not risking the binary-index hot path), blank-node OUTPUT-identity isomorphism (`open-eq-07/08/10/11/12`), temporal timezone semantics (`eq-dateTime`, `date-1`), `dawg-langMatches-4` (LANG-of-IRI error propagation), `not-not` (ill-typed literal preservation, → PR-X3), and `agg-count-rows-distinct` (`COUNT(DISTINCT *)` needs a new `CountDistinctAll` IR aggregate + whole-row group plumbing — a perf-neutral post-wave-3 follow-up). The D-6 empty-named-graph divergence keeps `bindings#graph` + `agg-empty-group-count-graph` registered (PR-U3). D-11 lexical-form tests (`group01`, `sameTerm`/`distinct`/`dawg-str`) are PR-X3's. `nested-opt-1/2` (correlated-OPTIONAL independence) is PR-W1-OPT; `join-combo-2` + `optional-complex-2/3/4` are PR-G1 (GRAPH-variable) territory.

**Post-review scope note (`xsd:float` type-preservation):** first-class `xsd:float` type-preservation holds on the **`Binding::Lit`/query path** but **not** the **indexed/`Binding::EncodedLit` (binary-store) path** — a stored `xsd:float` that materializes via late binary-index decode is currently recovered as `xsd:double`, so arithmetic on such a binding yields `xsd:double`. This is the same perf-sensitive encoded-path class as the deferred D5b datatype-carry family (and also covers the encoded-path lang-EBV corner), tracked as a separate follow-up.

## Verification & harness/CI notes

Verified (including the first-round review fixes now merged): full W3C suite `36 passed; 0 failed; 0 ignored` (0 stale, both directions); `-p fluree-db-query -p fluree-db-api -p fluree-db-transact -p fluree-db-sparql` 0-failed (fluree-db-query 1229, fluree-db-transact 332, fluree-db-sparql 617, grp_query 360 / grp_query_sparql 287 / grp_misc 244 / grp_graphsource 113); `cargo fmt --all --check` clean. Clippy: this PR's diff is clean (`--no-deps` over the touched crates), and both CI clippy jobs are green (the main job runs `cargo clippy --all --all-features --all-targets` with no `-D warnings`; the `-D warnings` job is scoped to the `testsuite-sparql` workspace). A newer un-pinned clippy (1.97.0) surfaces a handful of pre-existing pedantic lints in untouched infra crates (bench-support / r2rml / planner / …) under a whole-workspace `-D warnings` run — pre-existing baseline lint-debt, not introduced by this wave, tracked separately.

Harness notes for reviewers: `testsuite-sparql` is an EXCLUDED separate cargo workspace — run the W3C suite with `cd testsuite-sparql && cargo test`, not top-level `cargo test`. A fresh `git worktree add` does NOT populate the `testsuite-sparql/rdf-tests` submodule — run `git submodule update --init` before the suite or every manifest spuriously fails. `fluree-db-api` uses grouped test bins (`grp_query`, `grp_query_sparql`, `grp_misc`, …), so scope with the full `-p fluree-db-api`; a per-file `--test <name>` matches nothing. The pre-existing `resolve_shapes_source_g_ids` dead-code warning in `fluree-db-api/src/tx.rs:506` is base noise (untouched by this wave; does not fire under clippy `--all-targets`).

## Migration / changelog

- Equality/EBV are spec-strict: comparing to an unrecognized/ill-typed-datatype literal, and the truth value of a numeric-zero / empty-string / non-boolean-typed literal, follow SPARQL §17 (type errors exclude rows; numeric zero and empty strings are falsy); `AND`/`OR` apply three-valued dominance; `CONCAT` and `SUM`/`AVG` raise type errors rather than coercing; numeric promotion preserves `xsd:float` (so `xsd:float(10)/4` serializes as a typed `xsd:float` literal, not a bare number). Queries relying on the old permissive behaviour may return fewer rows.
- New SPARQL UPDATE verbs (LOAD/CLEAR/DROP/CREATE/COPY/MOVE/ADD + SILENT/INTO). DROP ≡ CLEAR (empties, does not unregister); CLEAR/DROP DEFAULT|ALL retract default-graph schema too (scope schema into a named graph if it must survive); CREATE is a no-op; non-SILENT remote LOAD errors (no HTTP client in the transact path — use LOAD SILENT or the insert API); empty named graphs are not enumerable. Non-SILENT `ADD`/`COPY`/`MOVE` **error when the source graph does not exist** (§3.2) rather than treating it as empty — which, for COPY/MOVE, had silently emptied the destination on a typo'd source; `SILENT` opts into proceeding with the missing source treated as empty — which, per the spec's own shortcut equivalence (`DROP SILENT dest; INSERT … WHERE source`), **still clears the destination**; a SILENT transfer from a missing source is NOT a no-op. Graph-management verbs also **reject the reserved `#config`/`#txn-meta` system graphs** as target/source. New builder API: `GraphTransactBuilder`/`Txn` `clear_graph`/`drop_graph`/`copy_graph`/`move_graph`/`add_graph` (full parity with the SPARQL transfer verbs).
- `DELETE … USING <g> WHERE { GRAPH <h> … }` now follows §3.1.3 — the `GRAPH <h>` block matches only when `<h>` is in `USING NAMED`; updates relying on the old over-reach should add `USING NAMED <h>`.
- SPARQL `CONSTRUCT` templates with blank nodes now emit triples (a bug fix aligning with §16.2; no migration).
- SPARQL 1.2 triple-term functions and `<<( )>>` values now parse/validate and produce a lower-time `not_implemented` where they previously produced a parse error.
- **JSON-LD subquery correlation (review follow-up, PR-W1):** a JSON-LD `["query"]` sub-SELECT no longer *pins* a correlation variable that is bound only via an inner `OPTIONAL`/`UNION` — it is produced independently inside the subquery and reconciled against the parent at join time (SPARQL §18.4 compatible-mappings; fixes W3C `join-scope-1`). Self-produced join-key correlations are unchanged. Existing JSON-LD queries of that exact shape may return different rows (standard natural-join semantics rather than forced equality); this is a correctness fix and no query rewrite is required.

## Issues closed on merge

Closes #1441 (PR-U6 — the USING + explicit-GRAPH delete over-delete data-loss bug; PR-U6 greens `dawg-delete-using-02a`/`06a`, fully resolving it — the issue has no JSON-LD component). Closes #1447 (PR-X2 — aggregate / equality / EBV conformance). The symmetric JSON-LD `from`-without-`fromNamed` restriction PR-U6 identified is a *separate* scoped follow-up (file a new issue if it needs tracking), not part of #1441.

*Mechanics: the `Closes` keywords fire only when this PR merges into the repository's DEFAULT branch. In the stacked burn-down flow (Wave 3 → `burndown/wave-2` → `burndown/wave-1` → `main`), merging this PR into `burndown/wave-2` does NOT close #1441 or #1447 — GitHub auto-closes them only once the stack collapses to `main` and this PR is retargeted onto it. Until then the issue references stay linked but open.*


bplatz (commented):
Inline notes from a review pass. Two look like must-fix (system-graph exposure via FROM, and a reachable transactor panic on triple-term syntax); the rest are correctness gaps plus a few worth-a-glance items.

fluree-db-api/src/view/query.rs
```diff
@@ -549,10 +626,108 @@ impl Fluree {
...
+        if iri == db.snapshot.ledger_id.as_str() || iri == db.ledger_id.as_ref() {
+            return Ok(Some(Self::apply_graph_selector(
+                db.clone(),
+                &GraphSelector::Default,
+            )?));
+        }
+
+        // Registered named graph in this ledger (registry, with the same
+        // binary-store fallback `select_graph` uses).
+        let known = db.snapshot.graph_registry.graph_id_for_iri(iri).is_some()
```
bplatz: `known` is also true for the reserved system graphs — the registry seeds `#txn-meta` (g_id 1) and `#config` (g_id 2). So `FROM <urn:fluree:{ledger}#config>` resolves and exposes the config graph, which the plain `GRAPH <iri>` path blocks via `single_db_user_graph_id` (>= `FIRST_USER_GRAPH_ID`). Suggest gating `known` on `g_id >= FIRST_USER_GRAPH_ID`.

fluree-db-transact/src/lower_sparql_update.rs
```diff
@@ -406,6 +410,9 @@ fn subject_to_object(s: &SubjectTerm) -> Term {
         SubjectTerm::QuotedTriple(_) => {
             unreachable!("RDF-star quoted triples are rejected before annotation expansion")
         }
+        SubjectTerm::TripleTerm(_) => {
```
bplatz, line 429: This is reachable. `expand_annotated_triples` guards `QuotedTriple` but not `TripleTerm`, so an annotated triple-term subject parses and panics here instead of returning the accept-then-defer error, e.g. `INSERT DATA { <<( :a :b :c )>> :p :o {| :m :n |} }`. Add a matching `TripleTerm` guard next to the `QuotedTriple` one.

fluree-db-transact/src/stage.rs
```diff
@@ -1482,7 +1848,23 @@ async fn stream_where_into_accumulator(
...
+            // dawg-delete-using-02a/06a; #1441). Selecting all registered named
+            // graphs here (the `None` fallback below) is what over-deleted: the
+            // `GRAPH <g2>` probe reached g2 despite `USING <g3>`.
+            //
+            // `None` is returned ONLY when there is no `USING`/`USING NAMED`
+            // clause at all, so the ambient graph-store dataset (every
+            // registered named graph) applies — the case a plain
+            // `DELETE WHERE { GRAPH <g> { .. } }` (no USING) relies on. That
+            // no-USING path stays byte-identical to before.
+            if w.using_default_graph_iris.is_empty() && w.using_named_graph_iris.is_empty() {
```
bplatz, line 2121: This fixes the named-graph half; the default-graph half of §3.1.3 still looks off. With `USING NAMED <g2>` and no plain `USING`, `desired_where_default_graph_iris` is empty and `base_db` falls back to g_id 0 (the real default) rather than the empty graph, so `DELETE {..} USING NAMED <g2> WHERE {..}` can still read/delete the default graph. Worth verifying.

fluree-db-query/src/eval.rs
```diff
@@ -429,6 +431,153 @@ pub fn passes_filters(
...
+/// language-tagged or foreign-datatype literal has no EBV.
+fn lit_effective_bool(val: &FlakeValue, dtc: &DatatypeConstraint) -> Result<bool> {
+    match val {
+        FlakeValue::Boolean(b) => Ok(*b),
+        FlakeValue::Long(n) => Ok(*n != 0),
+        FlakeValue::Double(d) => Ok(!d.is_nan() && *d != 0.0),
+        FlakeValue::BigInt(n) => Ok(!n.is_zero()),
+        FlakeValue::Decimal(d) => Ok(!d.is_zero()),
+        FlakeValue::String(s) if is_xsd_string(dtc) => Ok(!s.is_empty()),
+        _ => Err(ebv_type_error()),
```
bplatz, line 643: A cast/computed `xsd:float` is a `FlakeValue::String` tagged xsd:float, which isn't `is_xsd_string`, so it falls to `ebv_type_error()` → false in FILTER. `lit_to_comparable` coerces such floats for comparison but EBV doesn't, so `FILTER(?f)` drops a truthy float. Coerce numeric-XSD string datatypes before the error fallthrough.

fluree-db-transact/src/lower_sparql_update.rs
```diff
@@ -837,6 +844,23 @@ pub fn lower_sparql_update(
...
             lower_modify(modify, prologue, ns, &mut vars, &mut bnodes, opts)?
         }
+        // Graph-management verbs. CLEAR/DROP/ADD/COPY/MOVE lower to a
+        // graph-management directive executed by whole-graph scan at staging
+        // time; CREATE and SILENT LOAD lower to an empty no-op transaction;
+        // non-SILENT remote LOAD is a documented divergence (D-5).
+        UpdateOperation::Clear(gm) | UpdateOperation::Drop(gm) => {
+            lower_clear_drop(gm, prologue, opts)?
+        }
+        UpdateOperation::Create(_) => {
```
bplatz: CREATE no-ops without registering the graph, so a subsequent non-SILENT `COPY <g> TO <h>` / ADD / MOVE hits the new O3 "source does not exist" error even though CREATE reported success — the two behaviors contradict. (Also: the CLEAR/DROP `silent` flag is parsed but never read.)

fluree-db-query/src/eval/compare.rs
```diff
@@ -414,6 +444,118 @@ fn cmp_values(left: &ComparableValue, right: &ComparableValue) -> Option<Orderin
...
+    }
+}
+
+/// SPARQL RDFterm-equal (`=`, and negated, `!=`): §17.4.1.7 and the §17.3
+/// operator mapping. Returns a three-valued outcome so an incomparable pair is
+/// a type error (excluding the row) rather than silently `false`/`true`.
+pub(crate) fn rdf_term_equal(a: &ComparableValue, b: &ComparableValue) -> EqOutcome {
+    // Value-comparable fast path: numeric promotion, string, boolean, temporal.
+    // No plain-string↔temporal coercion (equality, not ordering).
+    if let Some(ord) = cmp_values_inner(a, b, false) {
```
bplatz, line 541: This value-equality path routes through `numeric_cmp`, whose bit-fallback makes identical NaN bits compare Equal, so `?v = ?v` on `"NaN"^^xsd:double` returns true (should be false). Consider treating a numeric pair that isn't `partial_cmp`-orderable as Ne here.

fluree-db-transact/src/stage.rs
```diff
@@ -1018,6 +1030,360 @@ pub async fn stage(
...
+        match op {
+            GraphMgmtOp::Clear(target) => {
+                // Resolve the target to a set of (g_id, Option<graph Sid>) —
+                // `None` Sid = the default graph (g_id 0).
+                //
+                // N3 (documented footgun): CLEAR/DROP DEFAULT and CLEAR/DROP ALL
+                // retract the WHOLE default graph, including schema flakes
+                // (rdfs:Class, rdfs:subClassOf, …). That is spec-correct — CLEAR
+                // is an unconditional whole-graph retraction — but a one-line
+                // `CLEAR ALL` strips the ontology, and because `is_schema_flake`
```
bplatz, line 1154: Worth a glance / a pinning test: this N3 bypass lets a narrow-grant identity `CLEAR ALL` (or `MOVE DEFAULT TO <g>`) and wipe/re-home the ontology unblockably. `test_graph_mgmt_honors_modify_policy` uses a non-schema flake, so it doesn't cover this case. Might be the intended contract, but a test pinning it would help.

fluree-db-api/src/view/query.rs
```diff
@@ -549,10 +626,108 @@ impl Fluree {
...
+        if clause.to_graph.is_some() {
+            return Err(history_range_dataset_error());
+        }
+
+        let mut dataset = DataSetDb::new();
+        for iri in &clause.default_graphs {
+            let graph = self
+                .resolve_within_ledger_graph(db, iri)?
+                .ok_or_else(cross_ledger_dataset_error)?;
+            dataset = dataset.with_default(graph);
```
bplatz: Two minor things: repeated `FROM <g>` pushes the same graph twice → bag union, so solutions duplicate (spec wants RDF merge/dedup). And an unresolved within-ledger IRI (e.g. a typo) returns `cross_ledger_dataset_error`, telling the user to use the cross-ledger path — misleading for a local typo.

fluree-db-transact/src/stage.rs
```diff
@@ -1018,6 +1030,360 @@ pub async fn stage(
...
+/// filtering them would leave a "cleared" graph non-empty). Modify-policy is
+/// still enforced on the resulting flakes (see `stage_graph_mgmt`), so this is
+/// not a privilege escalation; it only means that under `default_allow` + a
+/// view restriction, a `CLEAR` can retract flakes an equivalent DELETE-WHERE
+/// (which only sees viewable rows) would not.
+async fn scan_graph_flakes(
+    ledger: &LedgerState,
+    g_id: GraphId,
+    tracker: Option<&Tracker>,
+) -> Result<Vec<Flake>> {
```
bplatz, line 1079: Not a correctness issue, just noting: `CLEAR ALL` / large COPY/MOVE materialize every flake into a Vec and only check `at_max_novelty` at entry, so one op can roughly double novelty in a single commit — a scale cliff for exactly the ops most likely to touch the whole store.

fluree-db-sparql/src/lower/term.rs
```diff
@@ -88,7 +88,17 @@ impl<E: IriEncoder> LoweringContext<'_, E> {
...
-                    let var_id = self.vars.get_or_insert(&format!("_:b{}", self.vars.len()));
+                    // A bare `[]` is a fresh non-distinguished variable. Mint an
+                    // unforgeable registry name from a dedicated monotonic
+                    // counter: `#` is outside PN_CHARS, so no user-written
+                    // `_:label` can ever collide with it. (The old `_:b{len}`
+                    // scheme was forgeable — a user `_:bN` with N == vars.len()
+                    // here mapped to the same var, merging two intended-distinct
+                    // template blanks; matches the `#bnpl…`/`#coll…` labeling the
+                    // parser already uses for property lists / collections.)
+                    let label = format!("_:#anon{}", self.anon_counter);
```
bplatz: Worth a quick glance: the unforgeable `#anon` counter added here isn't used by `lower_reifier_id` in `lower/annotation.rs`, which still mints `_:b{vars.len()}` — so a user-written `_:bN` could collide with an anonymous reifier blank. Looks like this path was missed by the hardening pass.

fluree-db-transact/src/error.rs
```diff
@@ -155,6 +155,33 @@ pub enum TransactError {
...
 
+    /// A SPARQL/builder graph-management operation (CLEAR/DROP/COPY/MOVE/ADD)
+    /// resolved a target, source, or destination to a reserved system graph:
+    /// `urn:fluree:{ledger}#config` (g_id 2) or `#txn-meta` (g_id 1). These
+    /// graphs are Fluree-internal — `#config` seeds SHACL/uniqueness governance
+    /// and cross-ledger rules, `#txn-meta` holds commit metadata — and are never
+    /// part of the W3C dataset, so they must never be a graph-management target.
+    /// Mirrors the cross-ledger resolver's reserved-graph guard.
+    #[error("graph-management operation targets reserved system graph <{graph_iri}>; refusing")]
+    ReservedGraphTarget {
```
bplatz, line 166: Worth confirming `ReservedGraphTarget` / `SourceGraphNotFound` map to a 4xx client error upstream, not a 500.

fluree-db-api/src/view/query.rs
aaj3f: Confirmed — the registry legs of `resolve_within_ledger_graph` lacked the `>= FIRST_USER_GRAPH_ID` filter the `GRAPH <iri>` path applies via `single_db_user_graph_id`. Fixed in `8d8870ba1`: both legs (registry + binary-store fallback) now gate on user graphs, so `FROM <…#config>` / `FROM NAMED <…#txn-meta>` fall through to the standard "not in this ledger" rejection; test covers all four clause/graph combinations plus the user-graph positives.

fluree-db-transact/src/lower_sparql_update.rs
aaj3f: Confirmed and empirically reproduced — this landed as `881a8e20c` (pushed shortly after your review was cut): `expand_annotated_triples` now guards `SubjectTerm::TripleTerm` beside the QuotedTriple guard, returning the same accept-then-defer `UnsupportedFeature` as every other deferred triple-term shape, with negative tests for the INSERT DATA and DELETE-template paths (plus the QuotedTriple twin).

fluree-db-transact/src/stage.rs
aaj3f: Confirmed — same conclusion independently: `USING NAMED <g2>` with no plain `USING` fell back to g_id 0, so a default-scoped WHERE read (and deleted from) the real default graph. Fixed as `5b20e07ae` (also just before your review posted): any `USING NAMED` with no plain `USING` gives the WHERE dataset an EMPTY default-graph member list (§13.2.1 via §3.1.3; WITH is likewise ignored for the WHERE when any USING form is present). The test proves fail-without-fix — the pre-fix code deletes the whole default graph.

fluree-db-query/src/eval.rs
aaj3f: Confirmed, and it went one level deeper: besides `lit_effective_bool` (the BIND'd-binding path), the function-call path used the lenient `Into<bool>` conversion, so `FILTER(xsd:float("0"))` was TRUE. Fixed in `f52d43847`: string-backed numeric/boolean-tagged literals parse their lexical form on both paths — well-formed values follow the numeric/boolean rule, ill-formed forms are EBV FALSE per §17.2.2 rule 1 (not a type error). The dispatcher routes only `TypedLiteral` results through the strict path; everything else keeps lenient truthiness, which the shared Cypher surface's structural semantics rely on (D-12 carve-out — a Cypher CALL test locks this). Fixing the tests exposed a second bug: STRDT's datatype arg arrives as a Sid, which `as_str()` doesn't expose, so `STRDT(?x, xsd:integer)` silently produced a PLAIN literal — same commit decodes Sid datatype args.

fluree-db-transact/src/lower_sparql_update.rs
aaj3f: Confirmed — fixed in `bf4b626a0`: CREATE now registers the graph IRI in the additive registry (the same `graph_delta` mechanism as a transfer destination), so `CREATE GRAPH <g>` followed by non-SILENT `COPY <g> TO <h>` is a legitimate empty source. This needed two plumbing exceptions for registration-only (zero-flake) commits — the tx-builder no-op shortcut and `build_commit`'s EmptyTransaction gate — both scoped to deltas containing an unregistered IRI, so ordinary no-op updates still elide. CREATE-of-existing stays a no-op rather than the §3.2.2 error (documented D-6 divergence), and the CLEAR/DROP `silent` flag now carries a doc note explaining why it is never consulted (a CLEAR/DROP of an unregistered graph is a D-6 no-op, so there is no error to suppress).

fluree-db-query/src/eval/compare.rs
aaj3f: Confirmed — landed as `f86642ce0` (same pre-review batch): `rdf_term_equal` special-cases NaN (Double or Float) against numeric operands → `Ne` before the value fast path, so `=` is false and `!=` true, per XPath op:numeric-equal. Scoped to equality: the bit-level total order is deliberately kept for ORDER BY stability, and NaN vs an unrecognized-datatype literal still reaches the TypeError arm. Unit + both-surface integration tests, constant and stored forms.

fluree-db-transact/src/stage.rs
aaj3f: Agreed it's the intended contract, and agreed it deserved a pin — added in `2205252c5`: a view-only default-deny identity CAN `CLEAR DEFAULT` a schema-only graph (the `is_schema_flake` exemption), sitting beside the companion test proving the same identity is REJECTED for non-schema flakes. Narrowing the exemption later now flips a test instead of moving silently.

fluree-db-api/src/view/query.rs
aaj3f: Both fixed in `dd86e74b3`: the within-ledger dataset build dedups default-graph members by resolved graph id (`FROM <g> FROM <g>` contributes one member — §13.2 set semantics; regression test compares row-for-row against single-FROM), and the rejection message now says to check the IRI against the ledger's registered graphs before reaching for the cross-ledger API. (The stacked #1483 additionally set-dedups triples across DISTINCT default members at execution time.)

fluree-db-transact/src/stage.rs
aaj3f: Agreed — documented on `scan_graph_flakes` in `2205252c5`: whole-graph ops materialize every flake and only check backpressure at commit entry, so one `CLEAR ALL`/large COPY can roughly double novelty in one commit; chunked staging for whole-graph ops is the named follow-up if the cliff is hit in practice. Left as a doc note rather than code in this PR.

fluree-db-sparql/src/lower/term.rs
aaj3f: Confirmed for the path you named — and it's resolved by the `origin/main` merge now on this branch: main's `5c7cde84b` (wave-2 review response) had independently hardened query-side anon labels — including `lower_reifier_id` — with the `_:[]{n}` scheme, which the merge reconciliation adopted branch-wide (the wave-3 `#anon` counter was dropped in its favor). Sweeping for stragglers found one more genuinely-forgeable instance neither pass caught: the transact-side `BlankNodeCounter` minted INSERT-template anon blanks as `_:b{n}`, so `INSERT { [] ex:q ?h . _:b0 ex:p ?h }` fused both into one node — fixed with the same `_:[]` scheme + regression test in `dd86e74b3`.

fluree-db-transact/src/error.rs
aaj3f: Verified, no change needed: `fluree-db-server/src/error.rs` maps `ApiError::Transact(_)` to 400 via the catch-all (`status_code()`), with 409 reserved for the retryable conflict variants (`CommitConflict`/`PublishLostRace`/`NamespaceConflict`) — so `ReservedGraphTarget` and `SourceGraphNotFound` both surface as 400 Bad Request with the stable `INVALID_TRANSACTION` error type.

aaj3f:
All eleven review comments are addressed on the branch — per-comment replies are threaded inline. Three were fixed by the second-round commits that landed just before the review posted (the triple-term panic, the `USING NAMED` empty-default gap, NaN equality); the remaining eight resolve in `8d8870ba1..2205252c5`, on top of an `origin/main` merge (`d072a95d6`) that makes the PR conflict-free against `main` (the PR body's new "Third-round" section has the full breakdown).

Two things from this response wave worth calling out beyond the inline threads:

- **The `origin/main` merge surfaced one real cross-branch regression**: main's newly-wired Cypher suite caught the wave-3 Family-B subquery rule (W1) breaking `CALL (p) { OPTIONAL MATCH … RETURN count(…) }` zero-match retention — openCypher imports are per-row bindings, not inferred correlations. Fixed in `4a2243faa` with an explicit `pinned_vars` lateral-import marker on `SubqueryPattern`, set only by the Cypher CALL lowering, so SPARQL/JSON-LD Family-B semantics (join-scope-1) are untouched.
- **The merge also reconciled a registers collision**: this branch and main's wave-2 review response had each deduplicated the eval-triple-terms register in opposite directions, so six double-registered entries lost both copies in the auto-merge — the suite's both-direction enforcement caught it immediately, and each is restored once under its correct blocker cluster (register content is otherwise unchanged: same unique set, 0-failed / 0-stale).

Verified on the tip: full W3C suite 36 passed / 0 failed / 0 stale; 5,173 scoped integration/unit tests passing (query/transact/api/sparql/cypher); `cargo fmt` clean on both workspaces; scoped clippy clean. The stacked #1483 has been rebased onto this head (its lint-chore cherry-pick deduplicated away against main's original, as its PR body predicted).

PR #1475, opened 2026-07-11
bench: virtual-dataset performance corpus, harness, baselines + audit (pre-work for the perf PR series)

aaj3f:
## What this is

Benchmark infrastructure + evidence base grounding the upcoming virtual-dataset (Iceberg/R2RML) performance PR series. **No engine-pathway code changes** — the only engine-adjacent change is implementing the reserved `BenchSpanLayer` stub in `fluree-bench-support`.

## Contents

**Audit series** (`docs/audit/2026-07-virtual-dataset-perf/`): `00-benchmark-brief` (2-page findings summary), `01-pathway-inventory` (every perf strategy on the R2RML/Iceberg path with trigger/bypass conditions, adversarially verified), `02-hypothesis-map` (H1–H8), `03-corpus-design` (26 BI questions → 54 queries), `04-findings-register` (grounded findings incl. 3 correctness bug classes), `05-diagnosis` (six deep-dives; master bottleneck: decode is file-count-bound at ~200ms/file × 7,670 files/fact table), `ROADMAP` (ranked 9-PR slate, correctness-first, each with corpus-grounded impact + DoD).

**Harness** (`fluree-bench-virtual`, bin `vbench`): 54-query tagged corpus; native/virtual targets; hot/cold protocols (true cold via home-scoped cache + subprocess); per-query pathway counters from trace spans; canonical result hashing against native-blessed oracles (`hash_gate` policy for nondeterministic-selection queries); budget-gated `compare` for PR regression checks (native budgets included — virtual wins can't silently regress native); self-contained HTML dashboard.

**Baselines** (committed): 52 expected-result oracles blessed from native; perf medians for native-SF01 and virtual-SF01 (hot + cold subset).

**`fluree-bench-support`**: `BenchSpanLayer` implemented (allowlisted span capture w/ late-recorded fields, in-memory sink, `file:` JSONL mode).

## Headline evidence (SF01 parity pair: identical data, native ledger vs live Snowflake Iceberg)

- 24 ok / 30 DNF(≥120–180s) of 54; completing queries: 17 hash-exact, 2 policy-exempt, 5 divergent → 3 root-caused correctness bug classes (silent-empty transitive paths, silent-empty subqueries, missing `rdf:type` on bound-subject wildcards).
[...]

PR #1476, opened 2026-07-11
fix(r2rml): PR-0 — loud-refuse silent-empty virtual sub-scopes (breaking: previously silent-empty queries now error)

aaj3f:
First PR of the virtual-dataset perf roadmap (`docs/audit/2026-07-virtual-dataset-perf/ROADMAP.md`, PR-0). Correctness before performance: converts silently-wrong empty results into loud, actionable errors.

> **Breaking behavior (release notes):** queries with a transitive property path (`+`/`*`) or subquery inside a `GRAPH` scope over an R2RML/virtual source previously "succeeded" with 0 rows; they now fail with `InvalidQuery` and guidance. The old behavior was the bug — no kill switch by design. Labeled `breaking-change` for the release-notes section.

## Fixes

**0a — Silent-empty guard (findings F1/F2).** Transitive property paths (`+`/`*`) and subqueries inside a GRAPH scope over an R2RML source aren't lowered to R2RML and silently evaluated against the source's empty native index — returning **0 rows as success** (q034: 0 vs native 4,514; q051: 0 vs 247). The whole-scope guard counted only top-level unconverted triples; sub-scopes escaped it. Now the scope fails with the same `InvalidQuery` shape. Patterns that only transform bound solutions (FILTER/BIND/VALUES/UNWIND) and self-routing search patterns are deliberately not flagged; sequence paths (`a/b`) are unaffected (decomposed upstream). No lowering attempted — that's a later capability PR.

**0b — Bound-subject wildcard `rdf:type` (finding F3) — landed on `main` ahead of this PR.** Originally part of this batch; during the #1450 review response it was generalized to both wildcard shapes (1c0064f2e, `operator.rs`, with unit test `bound_subject_wildcard_emits_rdf_type`) and reached `main` when #1450 merged. After the restack this PR's diff no longer carries it (`operator.rs` here is byte-identical to `main`); the corpus q042 hash-parity blessing rides in #1475. Recorded here for the roadmap paper-trail: `<iri> ?p ?o` emits the `rr:class`-derived type triple (q042: 24 rows, hash-parity with native).

## Review response (2026-07-13)

- **fmt (blocking):** `cargo fmt --all -- --check` is clean workspace-wide at HEAD (rustfmt from stable 1.97.0, what CI resolves). The flagged `operator.rs` drift left this PR's delta with the restack; the remaining violation (VALUES negative-case test) is fixed.
- **Duplicate error blocks:** the two copy-identical GRAPH-scope `InvalidQuery` blocks (seeded + batched paths) are now a shared `rewrite::unsupported_subscope_error`, placed next to the `R2rmlRewriteResult::unsupported` field whose contract it implements.
- **Kind dedup:** repeated kind names collapse in the message ("property path", not "property path, property path"), first-seen order preserved, unit-tested. Message text otherwise unchanged; the corpus harness gates on error *status*, so no blessing churn.
- **Fused-aggregate invariant:** comment added at the single-`[Pattern::R2rml]` shape gate documenting why the third `rewrite_patterns_for_r2rml` caller is decline-safe against non-lowered sub-scopes (falls back to the normal GRAPH path, which raises the loud error).
- **Release-notes surface:** retitled + `breaking-change` label (the repo's `.github/release.yml` categorizes by label).
- **Restack rider:** the #1475 review response added a corpus unit test pinning "no query declares virtual=error"; this PR's manifest deliberately flips q034/q051, so that test failed at this HEAD after the restack. Updated it to pin exactly `[q034, q051]` (both still `ok` on native).

## Validation (per the roadmap DoD)

At current HEAD (724d25584): `cargo fmt --all -- --check` clean; scoped clippy clean (`fluree-db-query`, `fluree-bench-virtual`); `fluree-db-query` 1254/1254; `fluree-bench-virtual` 33/33.

From the pre-restack validation (details in commit history, unchanged by the review response): `fluree-db-api` full suite (all grp_* bins) 0 failed; W3C SPARQL suite unregressed (3 pre-existing syntax-bin failures proven pre-existing by stash-and-rerun); corpus native 54/54 with 0 hash mismatches vs blessed oracles; virtual-SF01 targeted: q034/q051 fail loudly (`expected_status` enforced by the harness), q042 24 rows hash-parity with native; 3 native micro-query perf "violations" (1.1–1.2×, 1–27ms absolute) exonerated by base-binary A/B.

## Notes / follow-ups (non-blocking)

- Exotic `<iri> ?p <class-iri>` (bound-object) shape still misses the type row — noted in ROADMAP residuals.
- Harness follow-up: per-query budgets need an absolute floor (10% of 8ms is a 0.8ms tripwire) and baselines should be re-blessed same-day before gating.
- CI note (stack-wide): `ci.yml` triggers only on PRs whose base is `main`, so the gates above are local until the stack retargets. `fluree-bench-virtual` has one remaining clippy warning (`too_many_arguments`, #1475-scope, non-gating today).

Stacks on #1475 (harness/baselines/audit). Roadmap PR-1 (#1478, COUNT(*) manifest shortcut) follows.


aaj3f:
Consolidating at maintainer request. #1476 (PR-0 — loud-refuse of silently-empty virtual sub-scopes, a breaking correctness fix) is carried forward in #1507 (the R2RML/Iceberg virtual-dataset performance program), which brings the #1475–#1507 lineage forward as a single reviewable unit. Its verification of record is the program's corpus benchmark, published as [C2-bench-wave1.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C2-bench-wave1.md) (native == virtual, 0 hash mismatches). The pre-consolidation branch tip is preserved at tag `archive/pre-refactor-2026-07-21/perf_r2rml-pr0-correctness` — no history was rewritten. Closing in favor of #1507; discussion continues there.

PR #1477, opened 2026-07-11
bench+test: query correctness guardrails — differential harness, baseline gating, and four fast-path fixes

aaj3f:
## What this PR is

The safety net for upcoming query/ledger refactor work, plus the four live correctness issues it caught, in one self-contained, almost-entirely-additive PR. Nothing here changes a fast-path kernel; the four fixes make fast paths agree with the generic pipeline.

## Contents (11 commits, reviewable per-commit)

**Tooling (additive):**
1. `feat(query)` — planner fast-path kill switch (`set_fast_paths_disabled` / `FLUREE_DISABLE_QUERY_FAST_PATHS`): lets any query run down both the fast and generic paths in-process. Zero cost when unset; also a production bisection tool.
2. `test(api)` — differential fast-vs-generic harness (`it_differential_fastpath`, in `grp_query`): randomized base + trailing-novelty + time-travel data; asserts fast ≡ generic per shape. On first run against main it reported the three divergences fixed below.
3. `feat(bench)` — `bench-baseline` capture/compare bin, `fluree-bench-alloc` tracking allocator (`scenario_mem` = peak − scenario-start), and `query_overlay_matrix` (count/star/group-by × base/overlay/novelty × scales — the overlay lane previously had zero bench coverage; every existing hot bench runs at epoch 0).

**Fixes (each with its regression pin; all four verified by the harness):**
4. `fix(query)` — **AVG fast path is datatype-faithful to the generic pipeline**: exact `xsd:decimal` for integer inputs (was: f64 always). Two indexed test expectations that had pinned the old f64 output are updated in the same commit. *Review-response follow-up:* the `xsd:double` lane now **declines** to the generic pipeline entirely — a new mixed-magnitude harness case demonstrated the fast fold's order-dependence diverging (0.0 vs 1.03125, catastrophic cancellation), so per the fallback contract the lane bails rather than risk it; the harness case pins the decline-lane agreement.
5. `fix(query)` — **per-predicate COUNT answers from POST leaf-directory metadata, never from `StatsView` estimates**: fixes over-counting duplicate re-asserts (e.g. 50 vs correct 26) and a dropped `rdf:type` row under pure novelty; detector declines OFFSET (unsafe under outer re-wrapping). Also greens W3C `agg02` — its stale skip-register entry is removed.
6. `fix(query)` — **failed overlay translations keep the raw flake for all error kinds** (was: non-`Unsupported` errors were warn-logged and silently dropped). Brings the export/main-scan helper to parity with the range-provider path; lossless even under a degraded dictionary state (white-box test included).
7. `fix(query)` — **whole-graph-agg overlay lane bounds `to_t ≥ index_t`**: the lane reconciles an index-pinned base count with a `to_t`-bounded overlay delta, which is exact only at/above the index watermark; a public `GraphDb::as_of(t)` below it produced a wrong count (reproducer included as a regression pin).

**Gate + observability:**
8. `feat(query)` — `FastPathOutcome::{Proceed, Fallback(reason)}` emitted on the `fluree::fastpath` tracing target at the four planner gate sites and at fast-operator `open()` — first step toward EXPLAIN showing the actually-executed path.
9. `bench` — committed `guardrails-pre` baselines (24 scenarios, time + `scenario_mem`); compare-vs-self exits 0, a doctored baseline exits 1.
10. `ci(bench)` — per-PR cheap compare (overlay matrix + hot-BSBM at tiny/quick) against the committed baseline; nightly compare of the same subset. **Both metrics (time and peak_mem) enforce only when the committed baseline's `runner_class` matches the runner's; against a cross-class baseline both are `::warning::` advisory** — so today's `local`-class baseline annotates without gating, and committing a CI-class baseline (the `workflow_dispatch` capture job) flips both to enforcing with no code change. The ~30-min job is gated on a `bench-paths` check and skipped when a PR touches nothing perf-relevant. Deliberately small: the full-workspace nightly compare needs the bench timeout discussion first.
11. `test(api)` — FD-2 (MIN/MAX-string lex-order) marker removed: main's own `fast_min_max_string` fix is confirmed by the harness and now enforced against regression.

## Main reconciliation (post-review, 2026-07-27)

`main` was merged in (`27afaac0e`) with both conflicts resolved **semantically**, per review: `stats_query.rs` keeps this PR's exact leaf-directory COUNT but carries main's `45d6009bf` datatype fix (`xsd:integer`, §18.5.1.6) into the rewrite — nothing else from that commit's hunks is lost (diff-verified); the W3C register takes the union — **neither `agg02` nor `agg-err-01` is skipped**, and the register comment now records the honest history (agg02 had two independent faults: count *value*, fixed here; count *datatype*, fixed on main — the earlier note dismissing the datatype theory was wrong).

The investigation also found **both test guards were structurally blind to fast-path datatype faults**: `testsuite-sparql` builds memory-backed ledgers so indexed fast paths never fire (its datatype comparison is strict — it just never sees this code), and the differential harness formats through JSON-LD, which erases numeric datatypes. Hence `it_count_datatype_pin.rs`: an indexed, novelty-free ledger asserted through `to_sparql_json` with a **strict** `xsd:integer` check — mutation-verified (re-typing to `xsd_long` turns it red). That converts the datatype contract from convention to enforcement, in both directions.

## Verification (re-run post-merge on the reconciled head)

- Differential harness green with **zero** known-divergence markers (all three historical divergence classes enforced) + the new COUNT-datatype pin green.
- Full `-p fluree-db-api` suite green (testcontainers bins excluded locally — Docker; they run in CI). `testsuite-sparql` W3C: **36/36 post-merge with `agg02` and `agg-err-01` un-skipped**.
- Scoped clippy clean on touched crates (`fluree-db-query`, `fluree-db-api`, `fluree-bench-support`, `fluree-bench-alloc`); rustfmt clean.
- Compare-gate truth table proven by unit tests + bin runs: matched-class + breach ⇒ exit 1; matched-class clean ⇒ 0; mismatched-class breach (either metric) ⇒ advisory, exit 0.
- Bench: fixes are perf-neutral on the overlay matrix; overlay lanes verified genuinely active (novelty-condition full scans run ~8–105× slower than the overlay-merged lanes, as expected).

## Notes for reviewers

- The four fixes change **answers** for the affected shapes (they now match the generic pipeline and the W3C suite). If any consumer pinned the old outputs, the two updated indexed expectations in commit 4 show the pattern.
- **`query_hot_whole_graph_agg` bench sanity panic (triaged & resolved in this review response).** Not an engine bug — bench-side. The bench writes its Person/Company graph under `http://example.org/` but queried the `db()` view with bare Cypher identifiers (`MATCH (n:Person)`, `n.age`), which resolve through the view's default-context `@vocab`; `db()` copies no ledger default context onto the view, so every bare label/property resolved to namespace 0 and matched zero nodes — the class-anchored scenarios silently benchmarked empty results and the filtered-histogram sanity assert panicked (`null` row). Fixed with a one-line `with_default_context({"@vocab":"http://example.org/"})` in the bench setup (`-- --test` now exits 0 — `MATCH (n:Person)` counts 900; the `age >= 40` histogram yields 26 groups). Two caveats documented in BENCHMARKING.md: the per-PR gate only *compiles* benches (runtime panics surface nightly-only — a costed follow-up), and this bench's class-scenario numbers pre-fix are void (re-baseline the first time a committed baseline covers it; the `guardrails-pre` subset excludes it, so no collision).
- `regression-budget.json` gains rows for the new bench; the `workspace_reconcile` parity test stays green.


aaj3f:
Review response — all items addressed in 10 additive commits (`5dd086adb..48599c6d8`); no history rewritten.

**Blocking — `bench-compare` gate** (4 commits):
- `9fc19cec9` splits the compare: **memory breaches enforce** (your ±2.2% cross-machine measurement held up — the mem half stays a real gate), **cross-machine time is advisory** (`::warning::` annotations, exit 0), keyed off a new `runner_class` provenance field with auto-detection. Also closes the stale-sidecar edge you flagged (sidecar-only entries with no time entry this run are treated as not-rerun).
- `5d72eb084` re-captures `guardrails-pre.json` at the true merge-base (`2c5162014`, engine code without the four fixes; tooling overlaid), recording `git_sha` = merge-base + `runner_class` + host so provenance is machine-checkable — the dangling-sha capture you caught is gone.
- `29c9c17f4` adds a `workflow_dispatch` CI-class capture job (uploads an `ubuntu-latest` baseline as an artifact; committing one flips the time half to enforcing) and clears `target/fluree-bench-mem/` before CI compares.
- `b639fba2a` rewrites BENCHMARKING.md: the missing "Baselines: capture & compare" section, the directory+bin scheme, the reconciled gate phases, and your same-runner base-vs-HEAD design documented as the long-term option (not wired — it doubles per-PR build cost, the original eviction risk).

Proof of the split: compare-vs-self exits 0; doctored **mem** breach exits 1; doctored **time** breach under `runner_class=ci-ubuntu-latest` exits 0 with an advisory. Three new unit tests cover the three modes.

**Double-lane AVG — you were right, and it was worse than the comment claimed.** A new mixed-magnitude `xsd:double` harness case measured the fast fold at **0.0 vs the generic 1.03125** (catastrophic cancellation under order-dependent summation). The double lane now **declines** to the generic pipeline (`8d9c6b596`); the harness case pins the agreement (`4bb8f51c9`). The integer lane (the actual fix) is untouched — exact decimal, order-free.

**Should-fixes/nits** (`0bbfe7b6c`, `4bb8f51c9`, `5dd086adb`): `scenario_mem` now records true scenario-attributable peak (peak − reset point) — your read of `reset_peak` was correct; dead OFFSET wrap → `debug_assert!(offset.is_none())`; duplicate comment merged; env-override semantics documented on `set_fast_paths_disabled` **plus** the harness now asserts at startup that `FLUREE_DISABLE_QUERY_FAST_PATHS` is unset (the vacuous generic-vs-generic mode you worried about is now impossible, not just documented); global-switch hazard comment added; `stats_query` IO errors now decline to the planned fallback (verified nothing has been emitted at that point, so `Ok(None)` is safe).

**Not done:** renaming the bare `rustfmt` commit subject — it would rewrite pushed history and the response is deliberately all-additive.

**Bonus (rider):** the nightly `query_hot_whole_graph_agg` sanity panic you'd have hit next was triaged during this response — bench-side (`@vocab` missing from the queried view; class scenarios matched zero nodes) — and fixed here (`5ad2508cd`), with the per-PR compile-only bench gap documented (`48599c6d8`). Details in the updated PR description.

fluree-db-query/src/stats_query.rs
```diff
@@ -1,195 +1,157 @@
...
+        schema,
+        move |ctx| {
+            let Some(store) = fast_path_store(ctx) else {
+                return Ok(None);
+            };
+            let Some(counts) = exact_count_by_predicate_post(store, ctx.binary_g_id)? else {
+                return Ok(None);
+            };
+
+            let dt = WellKnownDatatypes::new().xsd_long;
```
bplatz: **Definite issue** — this line is the conflict with `main`, and it's a revert.

`main` changed exactly this to `xsd_integer` in `45d6009bf`:

> SPARQL COUNT yields xsd:integer (§18.5.1.6); the materialized and streaming aggregate paths already tag it so — this per-predicate fast-path must match or a COUNT re-typed as xsd:long leaks (agg02).

Because this PR rewrites the whole 1–195 region, git can't merge that one-line change in, and the natural resolution ("take the rewrite, it's newer work") puts `xsd_long` back.

What makes it worse than a plain revert: the register comment on this branch explicitly *rejects* the datatype theory —

> The old note guessed a datatype re-typing; the real fault was the stats-estimate count operator this PR replaced.

I don't think that's right; it reads like two independent faults in one test. This branch fixed the count *value* (stats estimate → POST directory), `main` fixed the count *datatype*, and each was sufficient to green agg02 at its own base. Post-merge with only the value fix, agg02 fails again on the datatype — while this PR has removed its skip entry, so nothing reports it as expected.

Resolution: keep the rewrite, carry `xsd_integer` into it, and re-run the W3C aggregates suite after the rebase rather than trusting either branch's pre-rebase 36/36.

testsuite-sparql/tests/registers/mod.rs
```diff
@@ -94,12 +94,13 @@ pub const SPARQL11_AGGREGATES: &[&str] = &[
     // empty named graphs; gated on decision D-6, expected to remain
     // registered after PR-G1 (1)
     "http://www.w3.org/2009/sparql/docs/tests/data-sparql11/aggregates/manifest#agg-empty-group-count-graph",
-    // expression/aggregate cluster (2) — PR-X2: agg-err-01 = aggregate must
-    // poison (unbind) on non-numeric group members; agg02 = COUNT(?var)
-    // re-typed as xsd:int — fix site UNCONFIRMED (it is NOT the
-    // group_aggregate finalize; probe before fixing)
+    // expression/aggregate cluster (1) — PR-X2: agg-err-01 = aggregate must
+    // poison (unbind) on non-numeric group members.
+    // (agg02 = `SELECT ?P (COUNT(?O) AS ?C) … GROUP BY ?P` now PASSES — fixed by
```
bplatz: **Definite issue** — the other half of the same conflict, in the opposite direction.

This branch correctly removes `agg02`, but it still lists **`agg-err-01`** below (it forked before `main` greened that one via the SUM/AVG poison fix in the same `45d6009bf`). `main`'s register has neither entry.

So the two conflicts pull opposite ways: `stats_query.rs` resolved toward this branch *loses* a fix, and `registers/mod.rs` resolved toward this branch *re-adds* a skip for a test that now passes — which quietly stops the suite from enforcing `main`'s poison fix. A skip entry for a passing test is the failure mode that doesn't announce itself.

Resolution here is just the union: drop both `agg02` and `agg-err-01`. Worth updating the parenthetical too, since the datatype explanation it dismisses turned out to be a real second fault.

.github/workflows/ci.yml
```diff
@@ -116,3 +116,55 @@ jobs:
...
         run: cargo test
+
+  bench-compare:
+    # Cheap per-PR perf gate. Runs the two cheapest benches at tiny/quick and
+    # compares against the committed baseline. MEMORY breaches FAIL the job
+    # (peak_mem is comparable cross-machine — reviewer measured ±2.2%); wall-clock
+    # TIME breaches are ADVISORY (::warning:: annotations, non-gating) because the
+    # committed baseline is runner_class=local and shared ubuntu-latest runners
+    # flap — a 10% time budget on a single PR run false-positives (BENCHMARKING.md,
+    # "Baselines: capture & compare"). Once a ci-* baseline is committed (via
```
bplatz: **Worth a look** — hard-failing on memory across machine classes, on every PR.

The time/memory split is well-reasoned and the `::warning::`-only treatment of time is clearly right. My hesitation is that memory is *gating* while the committed baseline is `runner_class=local` — so every PR compares a shared `ubuntu-latest` peak against a number captured on someone's laptop. The ±2.2% figure is one measurement between two specific machines; allocator behaviour, page size, and background load on shared runners can move a tracking-allocator peak by more than that, and the failure lands on an unrelated PR author with no obvious cause.

Given the comment already anticipates the fix ("once a ci-* baseline is committed … time enforces automatically"), it'd be safer to make memory advisory too until that ci-class baseline exists, then flip both to gating at once. That also avoids the asymmetry where the *less* portable signal is the one that blocks.

Separately worth surfacing to the team rather than deciding here: this adds a ~30-minute job to every PR. Reasonable for the value, but it's a shared-cost change that people should opt into knowingly.

bplatz (changes requested):
The four fixes are good work — but **both merge conflicts with `main` are semantic, not textual**, and resolving either in this branch's favor silently reverts a fix already on `main` (`45d6009bf`, "correct fast-path COUNT datatype and poison numeric aggregates", merged Jul 8 — after this branch forked). Details inline.

This branch is 216 commits behind `main`, and the conflict set is small (`stats_query.rs`, `testsuite-sparql/tests/registers/mod.rs`) — small enough that it looks like routine noise, which is why I want to flag it before the rebase happens.

What I verified as sound:

- **The AVG fix is right, and better than the description suggests.** The exact-`i64` lane with `checked_add` → decline on overflow is order-free, and declining `xsd:double` outright is the correct call — folding in POST order vs the generic pipeline's scan order genuinely can't be made to agree. Removing the Kahan state also makes `fold_row` cheaper than before, so hot-loop purity isn't just preserved, it improves.
- **`stamp_fast_path` is at `open()`, not per-row**, so the observability seed adds nothing to the kernels.
- **The non-numeric decline composes correctly with `main`'s aggregate poisoning.** `fold_row` returning `Ok(false)` on a non-numeric member hands off to the generic path, which now poisons per §18.5 — so the two changes agree rather than fighting.

One `worth a look` on the new CI job inline.

aaj3f:
Response to review — all three items adopted; `main` reconciled by merge (`27afaac0e`) + three commits (`605a9e8d0`, `5d9d69513`, `7e915bf19`). Additive on the prior history; no rewrites.

**`stats_query.rs` (definite) — you were right, and it's resolved exactly as you prescribed.** The rewrite stays; `45d6009bf`'s `xsd:integer` + §18.5.1.6 comment are carried into it, and the resolved file was diffed against that commit's hunks to confirm nothing else was lost in the rewritten region. Your two-fault reading was also correct — the register comment now records it plainly (count *value* fixed here, count *datatype* fixed on main, each sufficient to fail agg02), replacing the note that wrongly dismissed the datatype theory.

**`registers/mod.rs` (definite) — union taken.** Neither `agg02` nor `agg-err-01` is skipped; both pass in the post-merge W3C run (36/36 — re-run on the reconciled head rather than trusted from either side, per your note).

**What the investigation turned up, which you'll want to know:** the reason this branch's earlier 36/36 couldn't have caught the datatype fault is that *neither guard can see it* — `testsuite-sparql` builds every ledger memory-backed, so indexed fast paths never fire (its `terms_match` is strict on datatype; it just never sees fast-path output), and the differential harness formats through JSON-LD, which renders an integer literal as a bare number, so a re-typed count compares equal. Your "nothing reports it as expected" instinct was pointing at a real structural hole. New `it_count_datatype_pin.rs` closes it: indexed novelty-free ledger, `to_sparql_json`, strict `xsd:integer` assertion — mutation-verified (re-typing to `xsd_long` turns it red). The datatype contract is now enforced in both directions, independent of either suite's blind spot.

**CI gate (worth a look) — adopted, your way.** Memory now follows the same rule as time: **both metrics enforce only on a `runner_class` match**; against today's `local`-class baseline the whole job annotates and exits 0. When a `ci-ubuntu-latest` baseline is committed (the `workflow_dispatch` capture job exists), both flip to enforcing together — no flag day, no asymmetry where the less-portable signal gates. Truth table proven in unit tests and live bin runs (matched+breach=1 / matched-clean=0 / mismatched-breach=advisory-0). On job cost: a `bench-paths` check now skips the ~30-min job for PRs touching nothing perf-relevant; whether to keep the advisory compare per-PR at all (vs nightly-only until the CI-class baseline lands) is flagged for the team rather than decided here — the plan of record is to commit the CI-class baseline promptly after this merges, which makes the job a real gate.

aaj3f:
@bplatz -- ready for re-review at your convenience

bplatz (approved):
Both conflicts resolved as prescribed in 27afaac0e — rewrite kept with `xsd_integer` carried in, register has neither `agg02` nor `agg-err-01`, and the corrected comment records both faults. `it_count_datatype_pin` machine-enforcing the datatype on an indexed ledger closes the gap the memory-backed suite can't cover. Compare gate is now symmetric (both metrics advisory until a ci-* baseline lands) with the bench-paths filter addressing the shared-cost concern.

Re-verified post-merge as asked: aggregates suite 44/0 (2 documented ignores), query crate green. Also test-merged current `main` (54 commits ahead) — textually clean, and the suites pass on the merge result, so no new semantic conflicts from the duration fast-path or early-dedup work.

PR #1478, opened 2026-07-11
perf(r2rml): COUNT(*) manifest shortcut — provider table_row_count (PR-1)

aaj3f:
Roadmap PR-1 (`docs/audit/2026-07-virtual-dataset-perf/ROADMAP.md`). A bare, unfiltered `COUNT(*)` over one class of a virtual (R2RML/Iceberg) dataset now answers from the Iceberg manifest `record_count` sum instead of decoding every Parquet data file. On the SF01 corpus this turns the three fact COUNTs from DNF@180s into sub-5s **exact** counts:

| query | class | before | after (cold) | count |
|---|---|---|---|---|
| q036 | edw:Order | DNF@180s | 4.43s | 180,000 |
| q037 | edw:WebEvent | DNF@180s | 3.12s | 1,000,000 |
| q039 | edw:GlJournalEntry | DNF@180s | 2.76s | 250,000 |

Query-side detection shipped in the parent commit (`271c8b7a6`, behavior-neutral until this provider impl); this commit is the provider side.

## Soundness contract

`table_row_count` returns `Some(n)` only when the manifest count **provably equals** a full-scan `COUNT(*)` (pure `sound_manifest_row_count`, unit-tested without I/O):
1. **No delete manifests** — a merge-on-read positional/equality delete would make the sum an over-count → decline.
2. **Every subject/object non-null column provably zero-null** — `aggregate_column_stats`' coverage gate yields `Some(0)` only when every data file reported the stat and they sum to zero; absent/partial stats are *unknown* ⇒ decline (never treated as zero); unknown column or would-be-negative total likewise declines.
3. Empty `non_null_cols` = constant-subject mapping ⇒ sound with no null proof.

Any decline ⇒ `Ok(None)` ⇒ the caller runs today's delete/null-correct scan. Trait default is `Ok(None)`, so non-Iceberg providers always scan. Kill switch: existing `FLUREE_FUSED_R2RML_AGG` (off ⇒ full-scan path, verified live).

## Extraction + snapshot-pin preservation

The loadTable front-half of `scan_table_inner` is extracted into `load_table_context`, shared by scan and count. The per-query snapshot pin (`IcebergCatalogSession::cached_load_table` / `pinned_metadata_location` / `store_load_table`, via the query-scoped `self.session`) moved **verbatim**, so a COUNT and a scan in one query read the same pinned `metadata_location` ⇒ one snapshot. Scan-only concerns (the scan-start log, the Parquet disk cache) stay in `scan_table_inner` — a COUNT emits no phantom scan log and builds no unused cache.

## Gates (all six pass)

1. `fluree-db-query` full suite — green.
2. `fluree-db-api --features iceberg` full suite — green; 6 new `count_shortcut_*` tests (delete-manifest DoD centerpiece, nullable, stats-absent, unknown-column, constant-subject, clean-table exact).
3. `fluree-db-iceberg --lib --tests` — green (180). Bare `cargo test -p fluree-db-iceberg` fails only on the **pre-existing** aws-gated `examples/smoke_minio.rs` compile issue (proven pre-existing by stash-to-base; suggest a tiny follow-up gating the example on `#[cfg(feature="aws")]`).
4. Native corpus — 54/54 ok, **0 hash mismatches** (three runs; byte-identical through the refactor). Perf "violations" proven environmental: the violator count tracks machine load (14 under a concurrent `cargo test` → 4 idle → 10 under another build) and the violator *set* shuffles run-to-run; native never enters this provider.
5. Live virtual (SF01) — table above, `result_hash == native oracle` on all three; fused-scan sentinel q022 unaffected (took the scan path, shortcut correctly declined — `files_selected=1` + scan span present); GROUP-BY negative control q014 unaffected. Fire-vs-decline proven by **span counters** (fired ⇒ `files_selected=0` + no scan spans), stronger than log greps.
6. Kill switch — `FLUREE_FUSED_R2RML_AGG=0` restores the scan path on q036 (DNF vs 4.43s on).

## Note

q036's residual 4.4s is *entirely* `oauth_token` (0.52s) + `r2rml.load_table` (3.52s): PR-1 reduces fact COUNT to the cold-catalog REST floor — exactly roadmap PR-8's target.

Stacks on #1476 (PR-0 correctness) ← #1475 (bench harness) ← #1450.


aaj3f:
Consolidating at maintainer request. #1478 (PR-1 — the `COUNT(*)` manifest shortcut) is carried forward in #1507 (the R2RML/Iceberg virtual-dataset performance program), which brings the #1475–#1507 lineage forward as a single reviewable unit. Its verification of record is the program's corpus benchmark, published as [C2-bench-wave1.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C2-bench-wave1.md) (native == virtual, 0 hash mismatches). The pre-consolidation branch tip is preserved at tag `archive/pre-refactor-2026-07-21/perf_r2rml-pr1-count-manifest` — no history was rewritten. Closing in favor of #1507; discussion continues there.

PR #1482, opened 2026-07-12
perf(iceberg): eliminate per-file footer round-trips + raise decode concurrency (PR-2)

aaj3f:
Roadmap PR-2 (`docs/audit/2026-07-virtual-dataset-perf/ROADMAP.md`): the H1 decode-wall lever set, driven by the measurement in `06-per-file-cost.md` (committed here as phase 1).

## The finding (phase 1, measurement)

The ~200ms/file fixed cost behind the fact-scan DNF wall is **one thing: the Parquet footer read** — 2 sequential S3 range GETs per file that never warm, because (1) the in-memory footer cache caps at 64 entries vs 7,670 files per fact table, and (2) `read_metadata` range-reads footers from **source S3 even when the whole file is already in the local disk cache**. Actual decode: 0.13ms. Spawn/channel: 0.17ms. Arithmetic reconciles the wall exactly: 7,670 × 195ms ÷ 8 = 187s ≈ the observed DNF.

## The levers (phase 2)

**Lever A — footer-from-fetched-bytes** (`FLUREE_ICEBERG_FOOTER_FROM_CACHE`, default on): for files in the whole-file-fetch tiers, parse the footer from bytes we fetch anyway. A2: disk-cache hit ⇒ footer+data served locally. A1: cheap ≤32MB tier ⇒ one whole-file GET (3 S3 RTs → 1). **Sparse-range/large files keep today's footer-first path unchanged** — no forced whole-file fetches. Switch off restores prior behavior byte-for-byte (verified live: footer_read snaps back to ~208ms/file).

**Lever B — decode concurrency default** `min(cpus,8)` → `min(cpus,32)`: pointwise ≥ the old default on every core count (2-core Lambda unchanged); `FLUREE_ICEBERG_SCAN_CONCURRENCY` still wins; 32 is the measured contention ceiling (8→32 = 2.1× on a 91-file scan, mild creep past c=16).

## Live results (SF01, hash-verified against native oracles)

| measure | before | after |
|---|---|---|
| warm per-file total | 191ms | **0.67ms** (285×) |
| warm footer_read | 190,627µs | **21µs** |
| q011 (91-file fact scan) | 3.27s | **620ms** |
| q001 / q022 sentinels | 2.19s / 675ms | 1.06s / 440ms |
| **q046 (ORDER BY, 7,670 files)** | **DNF@180s** | **45.9s fully cold** (footer 20µs/file; residual = pure data-fetch GETs) |

q046 completing cold means the **fact-scan DNF class now completes**; warm repeats are seconds. Composes with PR-1 (COUNT from manifest) and points at PR-8 (the remaining catalog/REST floor).

## Gates

1–3. `fluree-db-query` / `fluree-db-api` / `fluree-db-iceberg --lib --tests` — green (161/…/full suites). One `grp_misc` failure fully attributed: `it_minmax_fast_path_fired` documents it must be "the only test in this binary" (thread-local tracing subscriber) but is bundled with 241 siblings — pre-existing order fragility perturbed by our added callsites, **237/237 pass single-threaded**, passes solo both ways, and base-vs-levers full-bin A/B confirms interleaving not behavior. Test-infra follow-up: use `set_global_default` capture per the known interest-cache pitfall, or restore its own bin.
4. Native corpus 54/54, **0 hash mismatches** (three runs). Perf noise per the environmental protocol (concurrent unrelated builds observed; q009/q010 outliers queued for idle re-verification — levers are structurally Iceberg-only).
5. Live virtual: table above; fire/decline and switch semantics proven via span counters (footer_read 21µs on ⇄ 208ms off).

Phase-1 sub-spans (`read_footer`/`fetch_bytes`/`decode`/`plan_columns`) are kept as permanent counters — they're the regression evidence for this exact class.

Stacks on #1478 (PR-1) ← #1476 (PR-0) ← #1475 (bench) ← #1450.


aaj3f:
Precision note on the kill-switch claim in the body/commit: `FLUREE_ICEBERG_FOOTER_FROM_CACHE=0` restores the footer path exactly (span-verified: `read_footer` 21µs → 207,898µs/file, the 2-round-trip behavior), but Lever B's concurrency default is deliberately NOT gated by that switch — so switch-off wall time is 1,860ms vs the true pre-PR 2,966ms on the same query. To reproduce pre-PR behavior fully, pin both `FLUREE_ICEBERG_FOOTER_FROM_CACHE=0` and `FLUREE_ICEBERG_SCAN_CONCURRENCY=8`. Also: the q046 45.9s cold figure is on a 16-core host (Lever B is core-bound, min(cpus,32) ⇒ 16-way here); a ≥32-core host or the env override reaches the projected ~23s.

aaj3f:
Consolidating at maintainer request. #1482 (PR-2 — per-file Parquet footer round-trip elimination + higher decode concurrency) is carried forward in #1507 (the R2RML/Iceberg virtual-dataset performance program), which brings the #1475–#1507 lineage forward as a single reviewable unit. Its verification of record is the program's corpus benchmark, published as [C2-bench-wave1.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C2-bench-wave1.md) (native == virtual, 0 hash mismatches). The pre-consolidation branch tip is preserved at tag `archive/pre-refactor-2026-07-21/perf_r2rml-pr2-decode-levers` — no history was rewritten. Closing in favor of #1507; discussion continues there.

PR #1483, opened 2026-07-12
Wave 4: resolve the eight deferred follow-ups from the #1462 review (#1467-#1474)

### LINKED ISSUE #1467: transact: reification-aware f:reifiesGraph re-homing for COPY/MOVE/ADD
**Deferred from PR #1462 review (O5).** Wave-3 shipped a fail-loud guard (commit `1cb3e8cb2`) rejecting COPY/MOVE/ADD of any graph whose source carries `f:reifies*` flakes. This tracks removing that guard with proper support.

Root cause: whole-graph re-homing (`fluree-db-transact/src/stage.rs`, Transfer arm ~:1233) rewrites only the flake-level `g`, leaving the `f:reifiesGraph` OBJECT pointing at the source graph → `EdgeKey::from_reifies_facts` (`fluree-db-core/src/edge.rs:422`) returns `GraphMismatch` → both readers (`hydration.rs:1676`, `indexer_attachment_provider.rs:334`) silently `continue`, so edge annotations vanish after a COPY/MOVE (base triples survive). Confirmed by reading; symmetric across default↔named.

Fix: make the Transfer re-home reification-aware, then drop the guard — named→named: rewrite each `f:reifiesGraph` OBJECT to the dest graph Sid; named→default: DROP the `f:reifiesGraph` flake; default→named: ADD an `f:reifiesGraph` flake for the dest.

Acceptance: annotation-bearing COPY/MOVE/ADD succeeds; `from_reifies_facts` decodes cleanly on the dest (no `GraphMismatch`); round-trip tests (annotated edge → COPY/MOVE to named + default → query `@annotation` / SPARQL `{| |}` → survival) replace the Wave-3 rejection test.

### LINKED ISSUE #1468: query: datatype-aware =/!= for stored language-tagged literals (§17.4.1.7)
**Deferred from PR #1462 review (O1); documented D5b limitation.** `lit_to_comparable` (`fluree-db-query/src/eval.rs:463`) maps a stored `LangTag(_)` literal to a bare `ComparableValue::String`, dropping the tag. So `FILTER(?a = ?b)` with stored `"x"@en`, `"x"@fr` collapses both to `String("x")` → **true** (should be false per §17.4.1.7), while the *constant* form `FILTER("x"@en = "x"@fr)` correctly returns false via the `(Lang,Lang)` arm in `compare.rs`. The `lang_parts`/`EqKind::Lang` machinery is dead for stored data, and PR-X2 newly makes constants and stored data disagree.

Why deferred: X2's documented bet (`eval.rs:458-462`) — carrying the tag through would break every string builtin (CONTAINS/REPLACE/…) that assumes plain `String`.

Fix direction: route stored lang-tagged literals to a LangTag-carrying `TypedLiteral` at the eval boundary while keeping the string builtins tag-transparent (or teach them to strip the tag).

Acceptance: stored `"x"@en = "x"@fr` → false; `"x"@en = "x"@en` → true; CONTAINS/REPLACE/etc. unaffected.

### LINKED ISSUE #1469: query: within-ledger FROM default-graph union should be a set-merge, not a bag
**Deferred from PR #1462 review (O6). Type: correctness / SPARQL §13.2 spec-conformance — NOT a mere optimization** (the *fix* is perf-constrained, the *issue* is a conformance gap; tag it accordingly so it isn't triaged as an optional perf tweak). `DatasetOperator::next_batch` (`fluree-db-query/src/eval/dataset_operator.rs:407-458`) concatenates default-graph members without dedup. Per SPARQL §13.2 the default graph of a dataset is the RDF *merge* (set) of the `FROM` graphs, so `FROM <g1> FROM <g1>` should not double-emit and `FROM <g1> FROM <g2>` should count a shared `(s,p,o)` once (the graph component is stripped in the merge). Current behavior is a bag-union → duplicate rows. Inherited (cross-ledger hits it too); PR-G2 makes it one-line reachable within a single ledger.

Why deferred: perf-sensitive — a correct fix needs per-row dedup at `(s,p,o)` triple-identity across members (hash-set or sort-merge), adding cost + memory on the multi-`FROM` path; the common no-`FROM`/single-graph path is untouched.

Acceptance: `FROM <g1> FROM <g1>` = single copy; shared triples counted once; single-graph/no-`FROM` path byte-identical + within the perf budget.

### LINKED ISSUE #1470: query: xsd:float first-class on the EncodedLit (binary-store) path
**Bonus finding from the PR #1462 review verification (not in the original review); caveated in the PR description.** PR-X2 made `xsd:float` first-class via `ComparableValue::Float`, but only on the `Binding::Lit`/constant path (`lit_to_comparable`, `eval.rs:452`). The EncodedLit path (`eval.rs:150-173`) decodes via `ComparableValue::try_from(&val)` → `Double` (the `TryFrom` has no `Float` tagging and ignores `dt_id`). So a stored `xsd:float` materializing as a late-decoded `Binding::EncodedLit` (binary-store) becomes `Double`, never `Float`: `?f + ?f` on an encoded float yields `xsd:double` not `xsd:float`. (B1's numeric-builtin fix isn't affected — `ABS` gets a `Double` and works — but the result *type* is wrong.)

Why deferred (and consistent with an existing, documented scope line): this is the SAME root cause as the deferred scan-path `EncodedLit` datatype-carry family the PR already labels D5b (`eq-4`, `open-eq-02/05/06`, `dawg-lang-3`, `quotes-3/4`, `eq-graph-*`) — the encoded/binary-index path deliberately doesn't carry datatype to avoid risking the hot path. So this is not a new hole; it's the arithmetic-typing face of that same deferral. **Consider making this the umbrella issue for the whole encoded-path datatype-carry cluster** (cross-reference the D5b register entries).

**Second manifestation, same root cause + same fix:** a lang-tagged literal on the encoded path reads as string-truthy in `comparable_effective_bool` instead of raising a type error (the documented untested EBV corner). Carrying `dt_id`/`lang_id` on decode resolves both this and the `xsd:float` typing — fold both in here so it doesn't resurface as a stray third issue.

Fix direction: carry `dt_id`/`lang_id` from the encoded value on the `EncodedLit` decode arm (tag `Float` for xsd:float; preserve lang for the EBV corner), within the binary-index perf budget.

Acceptance: `?f + ?f` on an encoded `xsd:float` yields `xsd:float`; a lang-tagged encoded literal's EBV is a type error (not string-truthy); encoded and Lit paths agree; perf within budget.

(Related PR-description action: the `xsd:float` caveat is being sharpened to state explicitly that type-preservation holds on the Lit/query path but NOT the indexed/EncodedLit path — so "first-class" doesn't oversell.)

### LINKED ISSUE #1471: query: three-valued error propagation in IN / NOT IN (§17.4.1.9)
**Deferred half of PR #1462 review O2** (the value-equality half shipped in commit `67eca544e`). `eval_in`/`eval_not_in` (`fluree-db-query/src/eval/logical.rs:109,143`) now use `rdf_term_equal`, but the §17.4.1.9 error rule isn't honored: a comparison that type-errors inside the list is swallowed (`:114`/`:148` demote `Err`→continue) and the expression returns `false`, where the spec requires `?x IN (list)` to be a *type error* when no element equals `?x` **and** at least one element comparison errored (e.g. `1 IN (<iri>)` → error, not false).

Fix direction: propagate a demotable comparison error out of `eval_in`/`eval_not_in` when no element definitively matches and ≥1 comparison errored.

Acceptance: §17.4.1.9 error-in-`IN` semantics; the value-equality behavior from PR #1462 unchanged.

### LINKED ISSUE #1472: testsuite harness: tighten the UpdateEvaluationTest empty-result guard
**Deferred from PR #1462 review (P3); zero live impact.** `evaluate_update_evaluation_test` (`testsuite-sparql/src/query_handler.rs:86`) bails only when `!result_present` AND all content signals are empty. A *present* but content-empty result (e.g. `mf:result [ ut:graphData <unresolvable> ]`, where `graphData` fails to resolve) has `result_present == true`, so the guard doesn't fire and the test is treated as an "expected empty store" — a false-pass risk against a coincidentally-empty store. No current W3C test is affected (real corpus uses a bare `[]`). Fix needs the manifest parser to expose predicate-presence on the result node (`result_present` is a bool today).

Acceptance: a present-but-content-empty `mf:result` with zero usable predicates is a parse/skip, not an expected-empty-store; bare `[]` still passes.

### LINKED ISSUE #1473: query: G2 within-ledger FROM path re-parses SPARQL up to 3× (cold-path)
**Deferred from PR #1462 review (N5, re-parse half).** `build_within_ledger_dataset` (`fluree-db-api/src/view/query.rs:97`) parses via `parse_and_validate_sparql` (:652) for every SPARQL query, then `parse_sparql_to_ir` (:124) parses again → ≥2× for ALL SPARQL (even no-`FROM`); the `FROM`/dataset route (:103) adds one more → up to 3×. Cold-path (per-query, not per-row) so not a measured hot-path regression, but avoidable.

Fix direction: parse once, thread the AST/IR through the within-ledger-dataset + validate + lower path.

Acceptance: a single parse per query on both the `FROM` and no-`FROM` paths; behavior unchanged.

### LINKED ISSUE #1474: infra: pin the Rust toolchain and/or clear pre-existing clippy-1.97.0 lint-debt
**Surfaced during PR #1462 review verification; NOT introduced by any wave-3 change. File STANDALONE (repo-wide hygiene) — do NOT bury it as a "#1462 follow-up," or it gets lost.** The repo has no `rust-toolchain.toml`, so local dev lands on the latest stable (currently clippy 1.97.0, 2026-07-07), which surfaces ~10-20 pre-existing pedantic lints (`question_mark`, `unneeded_wildcard_pattern`, redundant-ref-in-`format!`/`assert!`) across untouched crates: `fluree-bench-support/budget.rs:94`, `fluree-db-r2rml/materialize/term.rs:835`, `fluree-db-query/{sort.rs, execute/operator_tree.rs, explain.rs, r2rml/rewrite.rs, planner.rs}`, `fluree-db-iceberg/io/{parquet,send_parquet}.rs`, `transact/address.rs`.

These do **not** gate CI (`ci.yml:50` main clippy has no `-D warnings`; `ci.yml:108` `-D warnings` is scoped to `testsuite-sparql`), but a local `cargo clippy --all-targets -D warnings` fails on them. NB: clippy has a warm-cache false-negative — a member built as a *dependency* earlier in the same invocation isn't re-linted, so partial runs miss these.

Options: (a) add `rust-toolchain.toml` pinning a known-clean stable; (b) clear the lints (each a trivial rewrite) and add `-D warnings` to the main CI clippy job to prevent regrowth.

aaj3f:
# Wave 4: resolve the eight deferred follow-ups from the #1462 review

Originally stacked on #1462 (`burndown/wave-3`); #1462 has since merged, so this PR now targets `main` directly. Resolves all eight issues filed out of that review: closes #1467, closes #1468, closes #1469, closes #1470, closes #1471, closes #1472, closes #1473, closes #1474.

**Restacked twice onto the moving #1462 head** (post-review-fixes, then post-`origin/main`-merge). The lint-chore cherry-pick **deduplicated away on the second restack exactly as planned** — main now provides the original `2c5162014` through #1462's merge — so the branch is 8 issue commits + 3 review-response commits (see the new section below), with the toolchain-pin commit standing alone (its message records the history).

**Brought current with `main` post-approval** (`bda63c577` merge + a lint-fix rider): #1462's merge deleted the old base branch and left ~3 weeks of drift. The merge is keep-both except one semantic reconciliation — main's `e4b048486` membership-lane pin asserted bag-union cardinality over an identical two-ledger default union (8 rows); under #1469's set semantics that union RDF-merges and stays at the single-graph answer (2), so the two affected tests flip to pin the set answer (`dataset_membership_shape_set_merges_identical_graphs`, `dataset_multiple_default_graphs_set_merged_across_ledgers`). The operator guard from `e4b048486` is kept — still load-bearing for multi-graph *named* scopes, which never arm the set-dedup. The `-D warnings` gate this PR introduces also surfaced a batch of pre-existing lints across four drift crates (clippy stops at the first failing crate, so they peeled off one crate at a time), fixed in the rider commit per the dev-setup bump policy.

Each issue got a code-verified audit before design (issue claims checked against current code; three issue texts had stale refs or wrong framing — corrected below), an adversarially-reviewed plan, and an adversarially-reviewed implementation. One commit per issue; the two `chore` commits at the tip are a coupled pair (see #1474 below) — don't bisect between them.

## What each commit does

**`fix(transact): reification-aware f:reifiesGraph re-homing for COPY/MOVE/ADD (#1467)`** — Whole-graph Transfer now re-homes annotation bundles instead of rejecting them: named→named rewrites each `f:reifiesGraph` object to the dest graph Sid, named→default drops the flake, default→named synthesizes one. The wave-3 fail-loud guard is gone, replaced by a round-trip acceptance matrix (COPY/MOVE/ADD × all three graph-direction cases × JSON-LD `@annotation` + SPARQL `{| |}` survival, decode with no `GraphMismatch`, source intact for COPY / gone for MOVE). One new fail-loud case discovered during plan review ships with it: `ADD` where the same explicit reifier `@id` reifies a *different* edge in source and dest would silently merge bundles and drop both annotations — that is now detected and rejected. `clear_dest` uses a symmetric rehomed-contents model, so COPY onto an identical annotation leaves the already-correct dest anchor in place rather than churning retract+assert at the same `t`.

**`fix(query): carry language tag through lit_to_comparable (#1468)`** — Stored lang-tagged literals now reach FILTER evaluation as `TypedLiteral{LangTag}` instead of bare `String`, so stored `"x"@en = "x"@fr` is false and `"x"@en = "x"@en` is true (§17.4.1.7), matching what the constant path already did. The string builtins that must stay tag-transparent go through a narrow string-arg helper that accepts plain and lang-tagged strings only — deliberately *not* `as_str()`, which would have silently accepted IRIs and foreign-datatype literals where the spec requires type errors; regression tests pin IRI/foreign-dt args as type errors at every swept site. Builtins that were already tag-correct off the `Binding` (CONCAT, UCASE/LCASE, STRBEFORE/STRAFTER, LANG) are untouched.

**`fix(query): tag encoded xsd:float and error on encoded lang EBV (#1470)`** — The `EncodedLit` (binary-store) decode arm now consults `dt_id`: a stored `xsd:float` becomes `ComparableValue::Float`, so `?f + ?f` on an indexed float yields `xsd:float` (was `xsd:double`). The EBV arm errors on `lang_id != 0`, matching the constant/Lit path. The EBV half is a **consistency** fix, not a W3C-conformance claim — RDF 1.0's plain-literal reading of §17.2.2 would make lang-tagged EBV truthy, RDF 1.1's makes it an error; the two engine paths disagreeing with each other was the actual bug. Tests run against indexed fixtures (memory-backed ledgers never exercise this path).

**`fix(query): three-valued error propagation in IN / NOT IN (#1471)`** — Completes the O2 work from #1462 (value-equality shipped there; this ships the §17.4.1.9 error rule). The audit found the issue text undercounted: there were **two** swallow sites per function — `EqOutcome::TypeError` falling into a catch-all, and demotable eval errors — in both `eval_in` and `eval_not_in`. Now a definitive match still short-circuits, but no-match-with-≥1-errored-comparison is a type error (row-drop in FILTER, unbound in BIND, catchable by COALESCE), mirroring `eval_or`'s three-valued pattern. `1 IN (1.0)` stays true.

**`fix(testsuite): tighten UpdateEvaluationTest empty-result guard (#1472)`** — The manifest parser now exposes predicate-presence on the result node, so a *present* but content-empty `mf:result` (e.g. an unresolvable `ut:graphData`) is treated as a parse problem instead of "expected empty store" — closing a false-pass risk. Bare `[]` still means expected-empty. No W3C test outcome changes.

**`fix(query): default-graph FROM union is a set, not a bag (#1469)`** — Per §13.2 the dataset default graph is the RDF merge (a set) of the `FROM` graphs. The audit found the bag-union bug in **two** places, not one: `DatasetOperator::next_batch` and the `drain_count` COUNT(*) fast path. Both now dedup at `(s,p,o)` triple identity across default-graph members — armed only when the default union has ≥2 members, so no-FROM, single-FROM, and FROM-NAMED-only paths take exactly the prior code path (fused fast paths and pruned emit masks intact). Implementation review caught that ground/all-constant BGPs (empty solution schema) escaped the dedup and double-counted; the deduper now collapses the empty mapping to at most one solution per scan, with regression tests that fail without it. A `debug_assert` ties dedup arming to full variable-identity emission so a future dataset path can't silently over-dedup. History-mode unions are excluded (assert/retract rows must not merge). Known, pre-existing limitation (unchanged by this PR): property-path and SERVICE operators build their own scans and bypass `DatasetOperator`, so they are not deduped over a multi-default union.

**`refactor(api): parse SPARQL once on the within-ledger dataset path (#1473)`** — The audit confirmed (and slightly worsened) the issue's count: every SPARQL query parsed ≥2×, the FROM/dataset route up to 3×. All four SPARQL entry methods now parse+validate once and thread the AST through dataset-clause resolution and lowering. Behavior identical (validation-error messages included); a `cfg(test)` parse counter pins parse-once on both the FROM and no-FROM paths. The boxed-future `Send` boundary is preserved.

**`chore: rustfmt & clippy fixes for new rust 1.97`** — *(historical; no longer on the branch)* This was a cherry-pick of main's `2c5162014` (author preserved; patch-id verified identical), needed because wave-3 branched before it landed on main. It deduplicated away when the stack restacked onto the post-merge #1462 head, as predicted here.

**`chore(ci): pin rust-toolchain 1.97.0, honor pin in CI, deny warnings (#1474)`** — `rust-toolchain.toml` pins 1.97.0 (with rustfmt+clippy components); the five CI toolchain steps get explicit `@1.97.0` refs (`dtolnay/rust-toolchain@stable` ignores the toml — the floating `@stable` was the root cause of the surprise lint-debt this issue was filed about); the main clippy job now runs `-D warnings`. (The formerly-coupled cherry-pick has deduplicated away — see above — so this commit now stands alone.) One knock-on worth knowing: `release.yml` installs Rust via bare `rustup` (latest stable), but the new root `rust-toolchain.toml` now *transitively* pins release builds to 1.97.0 too — previously they floated with stable. No conflict (cargo-dist ships as a prebuilt binary), just a behavior change to be aware of.

## Review-response commits (from the pre-review of this PR)

Three commits at the tip resolve the pre-review's findings on this PR's own work:

- **`fix(query): re-tag encoded language-tagged literals so indexed =/!= are tag-aware`** — the pre-review's one blocker: #1468's tag carry stopped at the `Binding::Lit` path, so stored-lang `=`/`!=`/`IN` were tag-aware on the memory path and tag-blind on the **indexed** (`EncodedLit`) path — the production-typical one (`"x"@en = "x"@fr` compared equal; constant-vs-stored `FILTER(?l = "x"@fr)` returned zero rows). The decode arm now re-tags from `lang_id`, symmetric to #1470's FLOAT re-tag beside it; indexed twins (`seed_indexed`, `range_provider`-asserted) cover `=`/`!=`/IN/constant-vs-stored on both surfaces and fail without the fix.
- **`fix(query): set-merge multi-USING WHERE unions; string_arg accepts xsd:string; parse-once pinned end-to-end`** — the UPDATE-WHERE path now arms the #1469 `multi_default_graph` planning flag for `USING <g1> USING <g2>` (the operator-level dedup already self-armed, so this closes the `emit_is_full` soundness-coupling gap and declines bag-cardinality fast paths, with an end-to-end set-semantics test); `string_arg` and `str_arg_and_lang` now share one accept-set (an explicit `xsd:string` is a simple literal — CONTAINS and STRBEFORE previously disagreed on `STRDT(?x, xsd:string)`); and the #1473 parse-once guarantee is pinned END-TO-END through the entry methods, not just the primitives. Also corrects #1472's motivating example in the code comments (an IRI `ut:graphData` object resolves at parse time — the guard's real cases are a nested graphData missing `rdfs:label`, or non-content predicates).
- **`test(sparql): pin the review's untested re-homing and union-dedup scenarios`** — six coverage adds, all confirming inspected-only behavior: ADD same-edge/same-reifier succeeds (the collision guard's false-positive boundary), named→default MOVE/ADD, multi-hop COPY a→b→c, the **synthesized** default→named anchor surviving the durable-index seal, multi-FROM + OFFSET/LIMIT paging the deduplicated sequence, and a var-var join where both scans carry the cross-member duplicate (1×1, not 2×2). Plus two doc notes: the N3 schema-clobber note covers `COPY/MOVE TO DEFAULT`, and history mode is documented as the one place a multi-FROM union stays a bag.

## Issue-text corrections found during audit

- **#1470's "umbrella for the D5b cluster" framing is wrong** — most of that register family (`open-eq-02`, `eq-graph-*`, `quotes-3/4`, `dawg-lang-3`) is blocked on the *scan path* (`matches_datatype_constraint`), which no eval-path fix touches; `eq-4` additionally needs eval-path foreign-datatype carry and `open-eq-05/06` a constant-side carry, both deliberately deferred here. A correcting comment will be posted on the issue.
- **#1469** is also broken in `drain_count` (COUNT fast path), and `dataset_operator.rs` is not under `eval/`.
- **#1471** has two swallow sites per function, not one.
- **#1472's** motivating example (`[ ut:graphData <unresolvable> ]`) was imprecise — an IRI object resolves at parse time regardless of file existence and fails later at load; the guard's true catches are a nested `ut:graphData` blank missing its `rdfs:label`/`ut:graph`, or a result carrying only non-content predicates. The code comments now say so (the original commit message keeps the imprecise wording; corrected here rather than rewriting history).

## Verification

- Full workspace test suite, `testsuite-sparql` W3C harness (36 suites), `cargo fmt --all --check`, and CI-exact `cargo clippy --all --all-features --all-targets -- -D warnings` all green on the final tip (clippy verified cold after `cargo clean -p` on every crate in the lint table, ruling out the warm-cache false-negative).
- **Net W3C register delta: zero, by design** — none of these fixes green a registered W3C test as scoped (the greenable candidates all have second blockers, documented per-issue). The conformance value ships as new unit/integration tests, including JSON-LD parity twins for every SPARQL-expressible semantics change (shared-IR rule).
- No measured hot-path regression: the dominant eval arms are byte-unchanged (one added integer compare on the encoded decode/EBV arms); the #1469 dedup is statically skipped for every single-graph shape.


bplatz (approved):
Approving. The branch is green on everything I could measure locally: `cargo clippy --all --all-features --all-targets -- -D warnings` is clean, `cargo nextest` over `fluree-db-api` / `fluree-db-query` / `fluree-db-transact` with `--all-features` passes 4822 tests, and the full `testsuite-sparql` W3C run is 36/36 suites green with no stale register entries.

Nothing below blocks the merge — none of it is a reproduced failure. Please triage which of these you think are worth fixing here versus deferring or closing out; happy either way.

---

## 1. `BatchDeduper` turns a streaming path into an unbounded-memory one

`dataset_operator.rs:200-270`

Every `FROM <a> FROM <b>` query now accumulates a `Vec<Binding>` signature per distinct row into a `HashMap` that lives for the whole scan. `SELECT * FROM <g1> FROM <g2> WHERE { ?s ?p ?o }` goes from bounded-memory streaming to O(distinct triples) resident, with full `Binding` clones as keys and no fuel or memory accounting. `drain_count` compounds it: `COUNT(*)` over a union abandons the per-member count-only shortcut and materializes the entire distinct set.

The comment notes that hashing mirrors `DistinctOperator`, which is fair, but `DISTINCT` is user-requested and visible in the query text; this is implicit on any multi-`FROM`. Options, cheapest first: store a `u64`/`u128` hash rather than cloned bindings; charge fuel per insertion; document the cliff the way `scan_graph_flakes` documents its own (`stage.rs:1058-1063`).

## 2. The emit-mask / dedup coupling is enforced only by `debug_assert!`

`dataset_operator.rs:518`

`PlanningContext::multi_default_graph` forces `EmitMask::ALL` in `make_first_scan` only, but `DatasetOperator::scan` is built at seven sites. Five are safe — `optional.rs:434`, `where_plan.rs:2599` and `fast_group_count_firsts.rs` (x2) hard-code `ALL`, and `operator_tree.rs:2450`'s pruned mask sits behind the fused fast paths this PR declines. Two take a planner-computed mask that can prune:

- `join.rs:1435`/`1457` — `make_right_scan(self.right_emit, ...)`, fed straight from `emit_mask_for_triple`
- `property_join.rs:983` — the existence-only `{ s: true, p: false, o: false }` branch, which fails `emit_is_full()` by construction

I probed 11 query shapes over 2-member unions (plain join, LIMIT, DISTINCT, ORDER BY, COUNT, three-pattern, OPTIONAL, UNION, and property-join existence-only) with `open()` instrumented, and every one reported `emit_full = true` and returned the correct set-merged multiset. So this is latent, not live. But the assert compiles out in release, so if a future plan shape does prune, the symptom is silently wrong cardinality in production and a panic only under test.

Cheapest hardening if you want it: apply the same `if planning.multi_default_graph { EmitMask::ALL }` in `build_scan_or_join`'s `Some(left)` branch — one edit, covers both remaining sites — and/or promote the `debug_assert!` to a returned error.

## 3. The encoded lang re-tag cannot see overlay / ephemeral lang ids

`eval.rs:196-209`

The re-tag resolves through `ctx.binary_store`, the persisted `BinaryIndexStore`. `dict_overlay::decode_meta` is the one that handles `assign_lang_id()`-allocated ephemeral ids; the persisted store's `language_tags.get(lang_id - 1)` simply misses, returns `lang: None`, and the value falls through to the old tag-blind `String`. That would put `"x"@en = "x"@fr` back to `true` for a novelty-resident lang literal sitting on top of an indexed ledger — the exact #1468 bug, reappearing as a function of materialization.

`extract_lang_tag` (`string.rs:63-71`) has the identical shape, so if this is reachable it is one shared fix. A test that stages a language-tagged literal *after* indexing would settle it.

## 4. Smaller items

- **Per-row double allocation on the lang re-tag.** `decode_meta` clones the tag `String`, then `Arc::from(tag)` allocates again — two allocations per lang-tagged row per expression evaluation, on what the comment calls the hot decode arm. A `lang_id -> Arc<str>` cache on `ExecutionContext` would make it free and also cover `extract_lang_tag`.
- **`drain_count`'s dedup branch silently forgoes count-only.** Worth surfacing in `plan_details` so EXPLAIN says why.
- **`temporal_mode.rs:122`** — an orphaned sentence fragment is left in the doc comment (`/// full per-event stream rather than a deduplicated set.`).
- **`view/query.rs:836`** — `validate_sparql_for_view` still carries the inline `match &ast.body` that `sparql_ast_has_dataset` was extracted to replace, and its error string is now duplicated verbatim in `query_r2rml_with_options`. Three copies of one rule.
- **`stage.rs:1391`** — `Vec::with_capacity(src_flakes.len() + 1)` is under-sized for the default -> named synthesis case (one extra per reifier). Cosmetic.
- **No `docs/` changes.** Candidates: `docs/design/query-execution.md` for the §13.2 set-merge and its cost, dev-setup/contributing for the toolchain pin and a bump policy, and `CLAUDE.md` itself — its "CI runs" line now omits `-D warnings`, and its "Pre-existing Warnings" section is stale now that clippy `--all-features` is clean.

## 5. One question

`COPY <g1> TO <g2>` with an explicit reifier `@id` now leaves the same reifier subject carrying `f:reifiesSubject` / `f:reifiesPredicate` / `f:reifiesObject` in two graphs. The new ADD guard exists precisely because two bundles on one subject make `EdgeKey::from_reifies_facts` return `Duplicate` and drop both. `transfer_named_to_named_survives_reindex` proves the attachment indexer is graph-scoped — is the JSON-LD hydration reader too, or can it ever assemble a bundle by subject across graphs?

---

The test work here is unusually strong. `emit_is_full()` as a self-describing soundness check on the builder trait, the zero-column ground-pattern case (`seen_empty`) that is easy to miss entirely, the OFFSET/LIMIT test pinning that dedup sits *below* the slice, the reindex-level gate proving the re-homed anchor survives into the durable arena, and the thread-local parse counter asserting parse-once through the real entry methods rather than the primitives. The `IN` / `NOT IN` pending-error handling is right, including a definitive match dominating a pending error. And pairing the toolchain pin with `-D warnings` in the same commit is the correct order to do those two things in.

aaj3f:
Thank you for this, @bplatz — the probing here (especially instrumenting `open()` across eleven shapes for the emit-mask question) went well beyond what I'd hoped for from a non-blocking pass. Given how many PRs we have in flight right now, I leaned toward "take it in this PR" rather than spawning follow-ups, so everything below landed as three commits on the branch (`ed500bc60`, `b375f2f8a`, `9250b5cdd`). Per-item:

**1 (dedup memory).** Agreed on all three counts, and we took two of your three options: each *retained* row now charges a unit of fuel (so a fuel budget bounds the resident set), and the `BatchDeduper` doc now records the cliff the same way `scan_graph_flakes` records its own. We looked hard at the hash-only key and decided against it — a collision there doesn't degrade, it silently **drops a distinct row**, which felt like the wrong trade for a correctness-motivated feature; `DISTINCT` keeps full keys for the same reason. If the memory profile ever actually bites, I think the better lever is spilling or a bounded two-tier set, but I'd rather not speculate that machinery into existence now.

**2 (emit-mask coupling).** The `debug_assert!` is now a returned error, so the release-mode symptom becomes a loud failure instead of silently wrong cardinality. On the two prunable sites: the property-join existence-only branch now widens to emit the object whenever the active default union has ≥ 2 members (`property_join.rs` — schema-safe, since the consumer keys off `emit_obj` and just ignores the extra column). For the NLJ right-scan, `self.right_emit` turns out to be `EmitMask::ALL` from its one production constructor (`where_plan.rs:2624` — the test-only constructors are the other callers), so I left it alone and let the new error be the backstop for any future caller that prunes. If you traced a live path feeding `emit_mask_for_triple` into it that I've missed, very happy to be corrected there.

**3 (overlay/ephemeral lang ids).** This one we took seriously enough to settle empirically, per your suggestion: `sparql_stored_lang_equality_with_post_index_novelty` stages lang literals *after* indexing and runs mixed indexed/novelty operands through `=`, `!=`, constant comparison, and `LANG()` — and it passes, because any novelty drops those scans to the non-late-materialized `Lit` path, so an ephemeral `assign_lang_id()` id never escapes into an eval-visible binding today (they're scan-internal to the overlay-parallel fold path). So: latent, not live — same category as your `#2`. But your instinct about the shape was right, and we hardened it anyway: there's now a shared `ExecutionContext::lang_tag_for_id` and an unresolvable nonzero `lang_id` surfaces as an *unknown* value (unbound, demotable) rather than falling through to a tag-blind bare string — the one outcome we never want back.

**4a.** Folded into the same resolver: a per-query `LangTagCache` (shared across per-graph context derivations like `ConstSidCache`) memoizes `lang_id → Option<Arc<str>>`, so it's one store meta-decode per distinct id instead of a `String` + `Arc` allocation pair per row — and `extract_lang_tag` rides the same cache now (its `Binding::Lit` arm also clones the constraint's existing `Arc` instead of re-allocating).

**4b.** `plan_details` now reports `default_union_set_merge` when the dedup is armed, so EXPLAIN can say why `COUNT(*)` forgoes the per-member shortcut.

**4c.** Yep — that orphaned fragment was mine, from an earlier doc edit that inserted a block mid-sentence. Removed.

**4d.** Consolidated: `validate_sparql_for_view` now uses `sparql_ast_has_dataset`, and the error string lives in one `single_ledger_dataset_clause_error()` shared by both call sites.

**4e.** Sized exactly now (one synthesized anchor per reifier subject when the source is the default graph).

**4f.** `docs/design/query-execution.md` gained the §13.2 set-merge + cost-model paragraph, and `dev-setup.md` now documents the toolchain pin, the bump policy (toml + the five workflow refs move together, lints fixed in the same PR), the transitive effect on release builds, and retires the stale "1.75.0 or later" line. One caveat: there's no `CLAUDE.md` in the tree at this tip (or on `main`) — I suspect the one you're seeing is a local/untracked file on your side, which we sadly can't fix from here; if it's meant to be tracked, happy to bring it in with the corrected CI description.

**5 (the question).** Good question, and it's now pinned rather than argued: `copy_with_explicit_reifier_reads_scoped_per_graph` does `COPY <g1> TO <g2>` with an explicit reifier `@id` and then reads the annotation in *each* graph on both surfaces — the SPARQL `{| |}` annotation-tail and the JSON-LD hydration reader (via the `ledger#graph` composite selector). Both decode cleanly in both graphs. Mechanically: the hydration fallback's bundle scan (`format/hydration.rs:1755-1770`) runs `self.db.range(SPOT, s=ann_sid)` against the *graph-scoped view*, so it only ever sees the active graph's `f:reifies*` flakes — a subject's twin bundles can't co-mingle into a `from_reifies_facts` `Duplicate`. So the reader is graph-scoped the same way the attachment indexer is, and now a test will catch it if that ever changes.

One more thing that happened between your review and this push: #1462 merged and took `burndown/wave-3` with it, so this PR retargeted to `main` and picked up ~3 weeks of drift. The bring-current merge (`bda63c577`, plus a rider commit fixing pre-existing drift lints in four crates that the new `-D warnings` gate caught) is mostly keep-both, but it contains one semantic reconciliation you should eyeball, because it touches your `e4b048486`: `dataset_membership_shape_keeps_per_graph_multiplicity` pins the membership-lane guard with bag-union cardinality (identical two-ledger default union squares the mutual-knows join to 8), and under this PR's #1469 set semantics that same setup RDF-merges to one triple set and stays at the single-graph answer (2) — which is exactly what the merged tree returns. So the guard itself stays (it's still load-bearing for multi-graph *named* scopes, which never arm the set-dedup), and the test flips to pin the set answer, with the old 8 and the half-collapsed 4 both called out in the comment so the history isn't lost. `dataset_multiple_default_graphs_no_dedup` got the same treatment (renamed, and its loose `>= 2` tightened to exactly 2 — multi-ledger execution decodes bindings, so the dedup key compares by value rather than per-ledger sid; the cross-ledger unification working out of the box was a pleasant surprise). If you'd rather also have a bag-cardinality pin for the named-scope lane specifically, happy to add one — the guard comment covers it today but a test wouldn't hurt.

And thank you for the kind words on the test work — `emit_is_full()` and the parse counter in particular came out of exactly the kind of "what would silently break this later" pass you're doing here, so it's nice to have it land with a reviewer who reads for that.

PR #1484, opened 2026-07-13
perf(r2rml): prune star TriplesMap fan-out — all-members intersection + disjoint-class hint (PR-3)

aaj3f:
Roadmap PR-3 (`docs/audit/2026-07-virtual-dataset-perf/ROADMAP.md`): the F8 star over-scan. A same-subject star resolved TriplesMaps by its **base predicate alone**, so a typed dimension star sharing a common predicate (`edw:name`) fanned out to every name-bearing table — 6 loadTables + 6 scans for a 500-row single-table query, tripling cold latency on the most common query shape.

## The prunes (both behind `FLUREE_R2RML_STAR_TM_PRUNE`, default on)

**(a) All-members intersection.** Only TriplesMaps supplying *every* star predicate can produce a complete star row; resolution filters to that intersection. The load-bearing all-members-required assumption is pinned by test (OPTIONAL members never fuse into stars — verified + guarded).

**(b') Disjoint-class hint.** When the star's co-located class was *not* fused (class lives in a different TM) but `wildcard_class_fusion_is_safe` proves every non-class TM subject-template-**disjoint** from the class-declaring TMs, resolution prunes to the class TMs — the standalone class scan joined on subject would drop the pruned rows anyway. Resolution-only: `rdf:type` emission and materialization are untouched.

> **Design correction on the record:** the ROADMAP's original fix (b) — prune to class-declaring TMs whenever fusion is refused — was proven **unsound** under vertical partitioning (class triple and data POMs in different TMs sharing a subject template: pruning strips the TM that carries the bindings). Counterexample annotated in the ROADMAP; permanent regression test `class_prune_hint_refused_under_vertical_partition`. The (b') formulation reuses the existing template-disjointness predicate rather than new machinery.

**Crawl-side fan-out: no work needed** — already class-constrained since `81b0ec601` (#1450); residual 429 backoff remains roadmap PR-8.

## Live results (SF01, 16-core host)

| measure | before | after |
|---|---|---|
| q001 loadTable / scan_table spans | 6 / 7 | **1 / 2** |
| q001 cold wall | 14.18s | **2.94s (4.8×)** |
| switch off | — | 6 loadTables / 14.18s faithfully reproduced |

q022/q054 sentinels hash-OK; native corpus 55 queries, **0 hash mismatches**.

## q002/q042 mismatch: pre-existing, filed as F9

Both mismatch identically with the switch **on and off** ⇒ not this PR. Root cause (head-row diff): native compacts predicate/type IRIs to CURIEs (`edw:name`), virtual emits full IRIs — all values byte-identical after namespace folding. Unmasked by PR-0's row-count fix. Filed as **F9** in the findings register with the which-side-conforms product decision escalated (SPARQL 1.1 results-JSON favors full IRIs, but changing native output affects existing users). No harness masking — they stay honestly red until decided.

## Gates

query suite 1200/1200 (6 new tests: intersection prune, class-hint prune, vertical-partition refusal, optional-member guard, switch-off fan-out, disjointness edge); api all bins green; native 0 hash mismatches; live table above.

Stacks on #1482 (PR-2) ← #1478 (PR-1) ← #1476 (PR-0) ← #1475 ← #1450.

## Review follow-ups (7b216843c)

- **fmt gate (blocking):** cleared — `cargo fmt --all -- --check` green at HEAD via the per-PR rustfmt commits (`12102a075` + inherited stack passes).
- **F10 filed (review optional 1):** the pre-existing split-member star-*formation* gap (same-subject members split across template-sharing TMs ⇒ zero star rows; the data-data analog of the (b') counterexample) is now a registered residual — `04-findings-register.md` F10 + ROADMAP §PR-3 residual note + cross-refs at `tm_passes_star_prune` and the rewrite star-fusion loop. Future fix in the rewrite (refuse to fuse when no TM covers all members); preserved, not caused, by fix (a).
- **Switch-family helper (review optional 2):** all five `fluree-db-query` R2RML switches now route through one `env_switch_enabled` helper with a single falsy spelling set (`0/false/off/no`); the `fluree-db-api` crawl helper and `fluree-db-iceberg` footer switch already mirror it (cross-referenced — no common crate to share the symbol). The #1482 OnceLock-idiom aside had already landed in `0590613a7`.


aaj3f:
Consolidating at maintainer request. #1484 (PR-3 — the star `TriplesMap` fan-out prune) is carried forward in #1507 (the R2RML/Iceberg virtual-dataset performance program), which brings the #1475–#1507 lineage forward as a single reviewable unit. Its verification of record is the program's corpus benchmark, published as [C2-bench-wave1.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C2-bench-wave1.md) (native == virtual, 0 hash mismatches). The pre-consolidation branch tip is preserved at tag `archive/pre-refactor-2026-07-21/perf_r2rml-pr3-star-overscan` — no history was rewritten. Closing in favor of #1507; discussion continues there.

PR #1485, opened 2026-07-13
perf(r2rml): memoize dim parent lookups across child batches (PR-4)

aaj3f:
Roadmap PR-4 (`docs/audit/2026-07-virtual-dataset-perf/ROADMAP.md`) — with an **honest DoD relabel** discovered during gating.

## What it fixes: the within-operator per-child-batch parent rebuild (original H3)

The correlated R2RML scan rebuilt every RefObjectMap parent-lookup HashMap **per 1,000-row child batch** (`build_progress`), so a fact⋈dim query re-scanned its dimension parents once per batch. Fix: cache `ParentLookup`s across batches on the operator (`Arc`-shared, keyed by `LookupCacheKey`), **window-capped** — a lookup larger than one materialize window is used for its batch but *not retained*, so a fact-as-parent falls through to today's per-batch build and the cache cannot OOM. Kill switch `FLUREE_R2RML_PARENT_MEMO` (default on), read per-operator at construction so both regimes are deterministically testable.

| evidence | memo ON | memo OFF |
|---|---|---|
| q008 (fact⋈dim GROUP BY, 180K-row fact) | **ok, 8 scans, 55s** | **DNF@75s, 123+ scans** |
| q015 (fact-as-parent) | no OOM (RSS ~538MB), falls through | same scan count |
| native corpus | 55/55, 0 hash mismatches | — |

The on/off scan split is load-immune and self-attributing (only this cache + switch can produce it). q008's hash mismatch is **pre-existing** (identical at both switch states; F9-class — q008 was DNF at baseline so its hash was never verifiable before). CI enforcement: a scan-counting mock-provider test — 5 child batches ⇒ exactly **1** parent scan memo-on, **5** memo-off — so a future refactor that silently reintroduces the per-batch rebuild fails CI, not just a nightly bench. Query suite 1201/1201.

## What it explicitly does NOT fix: q050 (relabeled scope)

The first gate run **failed its original DoD**: q050 (dims-only correlated OPTIONAL) showed identical scan *rates* memo on/off. Root cause: the correlated OPTIONAL rebuilds the **entire operator tree per correlated row** (`build_correlated_optional_op` → seeded rebuild), resetting any operator-scoped cache — this also defeats the existing `scan_cache`. q050's fix is **PR-4b**: `inner_pattern_is_hash_join_safe` (optional.rs:1418) admits only `Triple/Filter/PropertyPath`, and since the R2RML rewrite converts OPTIONAL inners to `Pattern::R2rml`, every OPTIONAL over a virtual dataset falls to per-row rebuild. Widening that gate (soundness analysis for star/multi-valued shapes required) routes the OPTIONAL through the batched hash-left-join — one long-lived inner operator, inside which *this PR's memo* then applies as designed. The two compose.

Diagnosis trail, failed-gate report, and the decision tree are in the session record; the ROADMAP annotation carries the scope split.

Stacks on #1484 (PR-3) ← #1482 ← #1478 ← #1476 ← #1475 ← #1450.


aaj3f:
Correction that STRENGTHENS the evidence: the PR body's "q008's hash mismatch is pre-existing (F9-class)" is wrong — there is **no q008 hash divergence at all**. The reported mismatch was an artifact of an ad-hoc diagnostic script (`jq -r .result_hash` over a `run --out` JSONL reads the `kind:"meta"` header's `null` alongside the real hash, so the compare always failed; the real `vbench compare` gate reads per-record and was never red). Verified post-fix: q008's hash equals the blessed native oracle byte-for-byte across 4 clean runs (`3594be9f…ea81`), heads identical — deterministically correct, no F9 involvement (its `?region` is a string literal), no float-order drift. Clean-machine DoD wall: memo ON 55.8–58.8s. Harness note filed: ad-hoc JSONL parsing must filter `select(.query_id)` to skip the meta line.

aaj3f:
Consolidating at maintainer request. #1485 (PR-4 — dimension parent-lookup memoization across child batches) is carried forward in #1507 (the R2RML/Iceberg virtual-dataset performance program), which brings the #1475–#1507 lineage forward as a single reviewable unit. Its verification of record is the program's corpus benchmark, published as [C2-bench-wave1.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C2-bench-wave1.md) (native == virtual, 0 hash mismatches). The pre-consolidation branch tip is preserved at tag `archive/pre-refactor-2026-07-21/perf_r2rml-pr4-parent-memo` — no history was rewritten. Closing in favor of #1507; discussion continues there.

PR #1486, opened 2026-07-13
test: standalone it_minmax bin + aws-gate smoke_minio (vbench follow-ups)

aaj3f:
Test-infra follow-ups from the perf-series gate batteries (no engine code). Originally stacked on the perf series (#1482); that stack was consolidated and has fully landed on main, so this now targets `main` directly — clean merge-up of `origin/main`, no conflicts, and the effective diff is unchanged.

1. **`it_minmax_fast_path_fired` → standalone test bin.** Its thread-local `set_default` capture requires sole ownership of the process — its own doc header says so, and main currently bundles it into `grp_misc` (`grp_misc.rs:34-35`) while the file still opens with "Kept as the only test in this binary". A process-global subscriber is unsound here: the test asserts the **absence** of the shared fast-path decline event, which siblings also emit → nondeterministic false-fail under parallel `cargo test`. CI's nextest process-per-test model masks this, which is exactly why it's still latent — bare `cargo test` is what contributors reach for first. Main has since adopted this same standalone-bin convention for `it_count_datatype_pin`, so this just brings the minmax test in line with it.
2. **`smoke_minio` gated on `required-features=["aws"]`** — the example unconditionally imports the `aws`-gated `S3IcebergStorage` (`lib.rs:98`), so bare `cargo test -p fluree-db-iceberg` fails to compile under default features today; with the `[[example]]` gate it's green.

The binary-index reload flake that surfaced during the original gate batteries (`it_values_type_deleted_repro::values_bound_type_after_delete_incremental_index_timeline`) is tracked in #1489 and not addressed here.


bplatz (approved):
Approving — leaving notes below rather than blocking on any of them. Please decide which are worth folding in here versus deferring.

What I checked locally:

- `cargo test -p fluree-db-iceberg --no-run` on **main** fails with `could not compile fluree-db-iceberg (example "smoke_minio")`, 5 errors. On this branch it compiles. The break is real and the gate fixes it.
- `cargo check --workspace --all-targets` under default features is clean, so `smoke_minio` was the only instance of this class.
- `cargo test -p fluree-db-api --test it_minmax_fast_path_fired` passes standalone.
- `cargo clippy -p fluree-db-api -p fluree-db-iceberg --all-features --all-targets -- -D warnings` is clean.

The branch is two commits behind main; the `perf/r2rml-pr2-decode-levers` merge is already in main, so it is history noise rather than extra payload.

---

## 1. The iceberg gate closes the instance, not the class

`required-features = ["aws"]` matches the two examples directly above it and is clearly correct. But nothing in CI would have caught this and nothing guards the next one: both CI jobs run `--all-features`, which turns `aws` on and compiles `smoke_minio` fine. The break was only ever visible to someone typing a bare `cargo test -p fluree-db-iceberg`.

A `cargo check --workspace --all-targets` step under default features would close the class. Given the shared cache the marginal cost is close to zero. This is the one substantive addition I would suggest for this PR.

## 2. `docs/contributing/tests.md` needs the third standalone category

This is the item I would most like to see land here.

`tests.md:85-91` lists exactly two reasons a test stays out of a `grp_*` binary: feature-gated suites, and process-global env mutation. This PR adds a third — process-global tracing state — and documents it only in a source comment, in a file that the same doc four lines earlier tells contributors to add to a group by default.

Without that bullet the next person doing a grouping pass re-bundles it and we repeat this. One line in `tests.md` makes the invariant discoverable from where people actually look.

## 3. The new header comment's reason (2) does not hold

> the `!has_event("fast path declined")` ABSENCE assertion would capture a *sibling's* decline

I do not think this is possible. `span_capture::init_test_tracing` uses `tracing::subscriber::set_default()` — thread-local — and its own header says so explicitly ("not `set_global_default`, for test isolation"). libtest runs each test on its own thread, so a sibling's events go to that thread's `NoSubscriber` and never reach this `SpanStore`. Under `--test-threads=1` the `DefaultGuard` has already dropped before the next test starts. There is no path for cross-test capture.

Separately, I could not reproduce the flake: I restored the bundled arrangement (`it_minmax_fast_path_fired` back inside `grp_misc`) and ran `cargo test -p fluree-db-api --test grp_misc` six times — 251 passed, 0 failed, with `multilang_min_served_by_fast_path ... ok` on every run. Reason (1), the process-global callsite-interest cache, is a genuine `tracing` mechanism, but `Dispatch::new` rebuilds the interest cache and `tracing-core` re-evaluates callsites once a scoped default has been set, which is presumably why it does not bite here.

None of that makes the change wrong — a dedicated binary is cheap insurance for a test whose assertions are about instrumentation, and this arguably just restores what the file header always claimed. I would only trim the comment to what is demonstrable (the callsite-interest cache is process-global; the capture is thread-local, so any work pushed off-thread is invisible; isolation is cheap) and drop the sibling-decline claim, so nobody reasons from a mechanism that is not there.

---

Mechanics of the move itself all look right: `mod support;` plus `#![cfg(feature = "native")]` with no `required-features` matches the `it_cyclic_bgp_probe` convention exactly, and the `[[test]]` entry is in alphabetical position.

aaj3f:
Thank you for the careful pass here, @bplatz — especially for re-running the bundled arrangement and for checking the class claim against main yourself. All three notes are folded in:

1. **CI default-features check** — added in 406cd3809: `cargo check --workspace --all-targets` at the end of the `test` job. You're right that both jobs running `--all-features` made this whole class invisible; the step compiles exactly the shape a contributor's bare `cargo test -p <crate>` hits first. Clean locally on this branch, and your workspace check on main matching (`smoke_minio` as the only instance) is good confirmation it starts green.

2. **`tests.md` third category** — added in 95b24c14f, alongside the feature-gate and env-var bullets and with the same "bites bare `cargo test` only" caveat the env bullet carries. Agreed this was the load-bearing one — as written, the doc actively steers the next grouping pass into re-bundling it.

3. **The header's reason (2)** — you're right, and it's gone (b43c2dd84). Tracing both asserted messages to their emission sites: each has exactly one (`fast_path_common.rs:3731` / `:3742`), both inside `FastPathOperator::open()` on the emitting test's own current-thread runtime, after `compute` returns — so there is no path for a sibling's decline onto the capturing thread; tokio tasks don't migrate across runtimes, and rayon never executes them. The one genuine cross-thread channel I could find is narrower than what I'd written: events emitted *inside* `parallel_map_pooled` closures can land on another participant's subscriber via work-stealing — but neither asserted event is emitted there, so it can't produce the sibling-decline capture I claimed. The rewritten header keeps to that plus the callsite-interest rebuild race, which I think also squares the history: the one failure we actually observed was during the PR-2 gate batteries under heavy concurrent-build load, consistent with a rebuild-race window — and with your 6/6 clean runs on an idle machine.

In any event, thank you for keeping the comment honest — better nobody reasons from a mechanism that isn't there.

PR #1487, opened 2026-07-13
perf(r2rml): admit subject-driven R2RML leaves to the batched correlated-OPTIONAL (PR-4b)

aaj3f:
Companion to #1485 (PR-4), fixing the class PR-4 explicitly deferred: the **correlated-OPTIONAL per-row operator rebuild**.

## Mechanism

`inner_pattern_is_hash_join_safe` admitted only `Triple/Filter/PropertyPath` — and the R2RML rewrite converts OPTIONAL inner triples to `Pattern::R2rml` — so **every OPTIONAL over a virtual dataset** fell to the per-row build: the whole inner operator tree rebuilt per correlated row (q050: 430+ scans, DNF; each rebuild also reset every operator-scoped cache, which is why #1485's memo couldn't help here). This PR admits the **narrow** shape — subject-driven scalar-POM / single-valued RefObjectMap leaves; stars, type-var, wildcard-predicate, and bound-subject shapes stay per-row with pointers to the soundness analysis (`docs/audit/2026-07-virtual-dataset-perf/07-pr4b-batched-optional.md`). Kill switch `FLUREE_R2RML_BATCHED_OPTIONAL` (default on; off = per-row exactly).

Prerequisite (own commit, `df5cb353e`): `R2rmlPattern::referenced_vars` hardened to correct-by-construction (exhaustive destructure + completeness test) — the correlation-set computation depends on it, and the audit confirmed **no live omission** existed.

## Results (SF01, 16 cores)

| | before | after |
|---|---|---|
| q050 (dims-only OPTIONAL, 37.5K products) | DNF@120s, 430+ scans | **ok, 16.7s, 92 scans, hash == native oracle (30,000 rows)** |
| switch off | — | per-row DNF signature faithfully reproduced |
| native corpus | — | 54/54, 0 hash mismatches |

**q016 verified declining on the STAR exclusion** (its OPTIONAL inner fuses to a same-subject star) — checked from the admission path, not inferred; future widening targets star admission, and its inner is fact-scale anyway (H1 territory).

## Correctness evidence — two halves, deliberately

- **(A) Live oracle differential**: q050 batched result is hash-identical to the blessed native oracle on all 30K rows *including genuine OPTIONAL-misses* — covering the empty-bucket LEFT-JOIN partition path directly. Caveat honestly stated: SF01 has referential integrity by generation, so this run covers dangling-FKs only by code-path identity (a dangling FK and a filter-miss are indistinguishable at the partition layer).
- **(B) Hermetic differential (CI-enforced)**: `batched_equals_per_row_on_dangling_fk` drives `build_batch` and the per-row build head-to-head on a mock provider with a **genuine dangling FK** — row-for-row equality including the dangling row's miss. A live batched-vs-per-row comparison is impossible on q050 (the per-row side is the thing that DNFs), which is exactly why this hermetic artifact exists rather than a construction argument alone.

Suites: query 1204/1204, api all bins green. Composes with #1485: the parent memo now applies inside the single long-lived batched inner.

Stacks on #1485 (PR-4) ← #1484 ← #1482 ← #1478 ← #1476 ← #1475 ← #1450.


aaj3f:
Consolidating at maintainer request. #1487 (PR-4b — batched correlated-OPTIONAL admission for subject-driven R2RML leaves) is carried forward in #1507 (the R2RML/Iceberg virtual-dataset performance program), which brings the #1475–#1507 lineage forward as a single reviewable unit. Its verification of record is the program's corpus benchmark, published as [C2-bench-wave1.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C2-bench-wave1.md) (native == virtual, 0 hash mismatches). The pre-consolidation branch tip is preserved at tag `archive/pre-refactor-2026-07-21/perf_r2rml-pr4b-batched-optional` — no history was rewritten. Closing in favor of #1507; discussion continues there.

Issue #1266, opened 2026-06-02
Incremental indexing leaves per-(class, property) stats stale after retractions


bplatz:
## Summary

The per-`(class, property)` flake counts in the index stats
(`IndexStats.graphs[g].classes[*].properties[*].datatypes`) are not
current-state-accurate after an **incremental** refresh. A commit that only
*retracts* a property flake (with no new assertion of that property on the
subject) computes the correct decrement but never applies it, so the count stays
too high. The property-level counts (`graphs[g].properties[].count`) decrement
correctly; the class-scoped per-property counts do not.

## Root cause

- The stats hook records per-subject datatype deltas with **signed** values, so
  retractions decrement there: `id_hook.rs` (`subject_prop_dts ... += delta`).
- But the per-subject property **presence** set is **assert-only**: `id_hook.rs`
  (the branch gated on `rec.op`).
- The incremental class-stat merge in `build/incremental.rs` only applies a
  class's `class_prop_dts` deltas when that class appears in `class_properties`,
  which is derived from the assert-only presence set. For a retraction-only
  change the class is never revisited, so its decrement delta is dropped.

## Where it shows up

A fast path answers `COUNT(*)` of `?s rdf:type ?o1 . ?s P ?o2` directly from these
stats as `Σ_C Σ_dt classStat[C][P].count` — each `P`-flake on a `k`-typed subject
is attributed once per class, which equals the join's product-sum, so no scan is
needed. With stale counts this **over-counts** on any ledger that has had an
indexed retraction touching `P` on a typed subject.

## Current mitigation (stopgap, not ideal)

The fold is gated on `store.lex_sorted_string_ids()`, which is set only by bulk
import (`import.rs`) and cleared by any incremental refresh
(`incremental_root.rs`) — a reliable proxy for "class stats untouched by
incremental drift." So the fold fires only on pure bulk-import indexes (where the
full build produces exact counts) and defers to the always-correct merge
everywhere else. This enables the optimization for bulk-imported datasets but
disables it for any incrementally-updated ledger.

## Proposed fix

Make the incremental class-stat merge apply `class_prop_dts` (and lang/ref)
deltas for **every** class that has such a delta, not just classes with an
assertion in the current commit:

- Drive the per-class merge by the union of the delta-source maps
  (`class_properties ∪ class_prop_dts.keys ∪ class_prop_lang_deltas.keys ∪
  ref_edges.keys`).
- Apply the datatype deltas to the prior `ClassPropertyUsage.datatypes`
[...]

PR #1488, opened 2026-07-13
fix: attribute ref-class stats when an endpoint gains its first type

bplatz:
## Problem

Ref-class stats (the `ref-classes` / attributed-refs maps behind `/v1/{ledger}/data-model`) attribute a ref triple's target class when the ref is indexed and never revisit it across incremental index batches. If subject `S` gains its **first** `rdf:type` in a later transaction, all previously-indexed refs pointing at (or out of) `S` stay unattributed forever — the stats disagree with what a query returns, and data-model consumers draw an incomplete class graph. Observed in the field as e.g. `Paragraph -> sourceDocument: 3,065 id-refs, 0 attributed` after ~19K documents were typed retroactively; only a full reindex or a delete/re-assert "touch" of the refs recovered.

Reproduced locally on file storage — the storage backend is irrelevant.

## Root cause

The incremental Phase 3b ref-edge re-attribution (#1266) keys batch-local class membership on `base_subject_classes` alone. A subject typed for the first time in a batch has no base entry (no `rdf:type` hits in the base PSOT lookup), so the endpoint `resolve` closure routes it to the "stable external entity" branch, which resolves both base and net classes from the base index — where it has none. The net class set comes back empty and the `+1` side of the edge move is dropped. Both Case A (outgoing edges of newly-typed subjects) and Case B (inbound edges of newly-typed objects) are affected.

Class-to-class **re-types** were unaffected (the subject has a base entry), which is why the existing #1266 tests passed. The datatype/lang re-attribution stage was also unaffected (it reads both maps with `unwrap_or_default`), producing the telltale "N id-refs, 0 attributed" signature.

Edges typed in the *same* batch as the ref always worked (the forward `subject_ref_history` pass uses net classes) — matching the observed straggler batch that did attribute.

## Fix

Treat an endpoint as batch-local when it appears in **either** `base_subject_classes` or the net `subject_classes` map. First-time-typed endpoints now resolve base = {} and net = their new classes from the in-batch membership, so the edge lands in the new bucket. They also no longer get pushed into the external base-index `rdf:type` lookup, which could only ever come back empty for them — strictly fewer base-index probes.

No new scans: the base subject-property scan and reverse-OPST inbound scan already ran for these subjects (untyped-to-typed already lands in `changed_membership`); their results were just discarded at the last step.

## Tests

Adds first-type-arrives-later variants alongside the existing #1266 re-type tests in `it_indexing_stats.rs`:
- `ref_class_attributed_when_object_typed_later` — the field repro shape (2 inbound refs, then the target gets typed)
- `ref_class_attributed_when_subject_typed_later` — outgoing direction
- `ref_class_attributed_when_both_endpoints_typed_later` — attributes exactly once (Case A/Case B dedup)

Also verified end-to-end with the CLI: refs at t=1 -> index -> type at t=2 -> incremental index -> `fluree info --graph default` now shows the refs attributed where it previously showed `ref-classes: {}`.

### DIFF
--- a/fluree-db-api/tests/it_indexing_stats.rs
+++ b/fluree-db-api/tests/it_indexing_stats.rs
@@ -808,6 +808,302 @@ async fn ref_class_reattributed_after_both_endpoints_retype() {
         .await;
 }
 
+/// Untyped→typed (Case B variant): an OBJECT that had NO rdf:type when its
+/// inbound refs were indexed gains its FIRST type in a later batch. The edges
+/// must attribute to the new object class — a first-time-typed subject has no
+/// `base_subject_classes` entry, so this exercises the net-map branch of the
+/// endpoint resolver (a plain re-type does not).
+#[tokio::test]
+async fn ref_class_attributed_when_object_typed_later() {
+    let tmp = tempfile::TempDir::new().expect("tempdir");
+    let path = tmp.path().to_string_lossy().to_string();
+    let mut fluree = FlureeBuilder::file(path)
+        .build()
+        .expect("build file fluree");
+    let (local, handle) = start_background_indexer_local(
+        fluree.backend().clone(),
+        fluree
+            .nameservice_mode()
+            .as_arc_indexing_nameservice()
+            .expect("test fluree has writable nameservice"),
+        fluree_db_indexer::IndexerConfig::small(),
+    );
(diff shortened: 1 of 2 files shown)


zonotope (approved):
🧇

PR #1490, opened 2026-07-14
perf(r2rml): fuse GROUP-BY rollups over FK joins into the aggregate fold (PR-6)

aaj3f:
Roadmap PR-6 (`docs/audit/2026-07-virtual-dataset-perf/ROADMAP.md`; design + attribution in `10-pr6-rollup-fusedagg.md`): the last big slow-tail class. With the decode wall gone (#1482), the dominant real-world BI shape — `GROUP BY` a dimension attribute + `COUNT/SUM/AVG` over a fact⋈dim join — was **operator-bound**: the generic pipeline materialized 180K+ join bindings just to fold them into a handful of groups.

## What it does

Extends `FusedR2rmlAggregateOperator` to fold aggregates **directly from the fact column batches over a fact→dim FK chain**, resolving GROUP-BY dimension attributes through per-hop lookups (one small scan per dim, PR-4-style memoization) instead of materializing bindings. Two phases in one commit (an intermediate buildable split wasn't reconstructible after the in-place chain rewrite; revertability is via the sub-switch):

- **6a — 0/1-hop**: fact + one dim admitted at the `resolve_at_open` join-rejection site; group keys resolved via a sibling `build_attr_lookup` (FK → attribute value, not FK → subject IRI). Un-annotated string group keys default to `xsd:string` (R2RML natural mapping) — **join-path only** (extending it single-table exposed latent bug **F12**, filed in the register, unreachable today).
- **6b — k-hop chain**: `order_chain` orders leaves into a **linear** fact→dim₁→…→dimₖ chain — single-column ref-join per hop, no branch, no merge, **cycles decline** (tested). Dangling/NULL FK at **any** hop ⇒ that fact row drops from the rollup (R2RML no-triple semantics, same validity gate as materialization).
- **HAVING** admitted by wrapping (`HavingOperator` above the fused op, like ORDER BY/LIMIT today) when its lifted aggregates ⊆ projected aggregates; rejection case tested.

Kill switch **`FLUREE_FUSED_R2RML_AGG_JOIN`** (default on, standard `env_switch_enabled` family) gates only the join admission — off ⇒ the proven single-table fused path and the generic join pipeline, byte-identical. Also migrates `FLUREE_FUSED_R2RML_AGG` to the switch family (fixes the exact-"0"/"false" parsing inconsistency documented in the WP1 audit).

## Live results (SF01, hash == native oracle on every query)

| query | shape | before | after |
|---|---|---|---|
| q008 | 2-hop (Order→Customer→Geography) revenue by region | ~45–50 s | **609 ms (~75×)** |
| q009 | 2-hop + HAVING | ~50 s | **591 ms** |
| q032 | 1-hop on-hand by store | 70–79 s | **294 ms–2.6 s** |
| q025 | 1-hop + HAVING | 1.8 s | ~1.1 s |
| q010 / q014 | multi-key / single-table | 1.8 s / — | ~0.5 s / unchanged path |
| q022 | **single-table sentinel** | — | **byte-identical** (validated after an attribution round caught a first-cut regression; fix scoped join-only) |

## Validation

Post-restack onto the review-updated base (`67daa5686`): query suite **1227/1227** (new: chain-order/cycle-decline, HAVING admit/reject, string-key default, decimal-key regression); api suite full green; **native corpus 54/54, 0 hash mismatches**; W3C suite green (36/36 register groups, 0 failures); fused-vs-generic A/B verified per query with decline-path debug attribution (the 6a gate initially showed no win — root cause was a `group-key-datatype-unsupported` decline on un-annotated string keys, found by instrumenting the decline path, fixed, re-validated). BSBM native budgets deferred to an idle-machine run before merge-to-main (changes are R2RML-graph-source-gated; native corpus already gates the native path).

**Dropped from this PR:** the F9 CURIE-parity rider — its one-line formatter fix proved a no-op (root cause is a namespace-map gap: virtual snapshots never register the R2RML mapping's vocab prefixes, so the compactor has nothing to compact against). Re-scoped in the findings register; stays in the tail queue.

Stacks on #1487 (PR-4b) ← #1485 ← #1484 ← #1482 ← #1478 ← #1476 ← #1475 ← #1450.

---

## Admission contract (precise, matches the code)

**Admitted.** A `GRAPH { … }` over an R2RML graph source feeding a GROUP-BY (or implicit) aggregation of multiset `COUNT`/`SUM`/`AVG`, whose WHERE lowers to a **linear `fact → dim₁ → … → dimₖ` foreign-key chain**. `order_chain` requires exactly one root (the fact), one ref-join per hop, and a full single-visit walk — **branch, merge (diamond), cycle, and disconnected shapes decline**. Each hop is a **single-column FK**. GROUP-BY keys are scalar attribute columns on the **terminal** dimension (an un-annotated column takes R2RML's natural `xsd:string`, **join path only**). Aggregates fold from **fact** columns / fact decimal expressions. **HAVING** is applied by a wrapping `HavingOperator`, admitted only when its lifted aggregate is in the SELECT projection.

**Excluded → generic pipeline** (byte-identical with the sub-switch off): a FILTER over the join; branch/merge/cycle joins; composite FK; IRI/ref group keys or keys off a non-terminal dim; aggregates over a dim column; DISTINCT aggregates; the single-table path with an un-annotated string key (kept byte-identical — see **F12**); and a non-lowered sub-scope (`PropertyPath`/`Subquery`), which declines so the generic GRAPH path raises the loud unsupported-subscope error.

**Row semantics.** A dangling or NULL FK at any hop, or a NULL terminal group attribute, drops the fact row — matching the R2RML/inner-join row-drop exactly.


aaj3f:
Consolidating at maintainer request. #1490 (PR-6 — fused GROUP-BY rollups over FK joins) is carried forward in #1507 (the R2RML/Iceberg virtual-dataset performance program), which brings the #1475–#1507 lineage forward as a single reviewable unit. Its verification of record is the program's corpus benchmark, published as [C2-bench-wave1.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C2-bench-wave1.md) (native == virtual, 0 hash mismatches). The pre-consolidation branch tip is preserved at tag `archive/pre-refactor-2026-07-21/perf_r2rml-pr6-rollup-fusedagg` — no history was rewritten. Closing in favor of #1507; discussion continues there.

PR #1493, opened 2026-07-14
perf(r2rml): admit the same-subject STAR to the batched OPTIONAL (PR-4c, q016 — last corpus DNF)

aaj3f:
## PR-4c — admit the same-subject STAR to the batched OPTIONAL (q016, the last corpus DNF)

Stacked on PR-8b (`perf/r2rml-pr8b-innerjoin-memo`).

q016 (`?o a Order ; orderId ?oid . OPTIONAL { ?sh edw:order ?o ; edw:shipStatus ?st }` LIMIT 5000) DNF'd at 180 s. Its OPTIONAL inner is a **same-subject STAR** on `?sh` (FactShipment) whose correlation var `?o` is the OBJECT of a ref member, so PR-4b's narrow admission (a subject-driven single-object leaf — exactly q050) declined it → the per-row `OptionalBuilder::build` rebuild → ~445–517 FactShipment scans.

This widens `inner_pattern_is_hash_join_safe` to also admit a same-subject star (`r2rml_star_is_hash_join_safe`), so q016 takes the batched hash-left-join.

### Sound to admit (design doc §a–b)

- **The correlation set is COMPLETE for a star.** `R2rmlPattern::referenced_vars` exhaustively destructures the struct (a new field is a compile error) and surfaces every star-member object var — the P1 audit, **already landed + tested as PR-4b's precursor**. So a star's `?o`/`?st` are in the partition-key input; no mis-partition. `star_constraints`/`class_prune_hint` carry no var. No change to `referenced_vars` or the seed/partition core.
- **Cartesian multiplicity ≡ per-row** for LEFT-JOIN (the same leaf materialization runs both paths; not the excluded row-multiplying subquery). The object-side `?o` is seeded BOUND (`corr_var_only_triple_object` is Triple-only) — sound, only de-optimized. Kept EXCLUDED: `type_var`, wildcard `predicate_var`, bound subject. Sub-switch `FLUREE_R2RML_BATCHED_OPTIONAL_STAR` (default on within the PR-4b family).

### Result + the scan-count attribution

| | before | after |
|---|---|---|
| q016 | DNF @ 180 s | **ok**, 5000 correct rows, **~39 s hot** |
| q016 `scan_table` | 445–517 | **182** = 2 FACT_ORDER (outer, **collapsed**) + 180 FACT_SHIPMENT (inner) |

An initial gate criterion expected the inner to collapse to single-digit scans. It didn't — and the attribution (an env-gated per-scan table tag) showed why: the outer driving side IS collapsed (2 scans), but the batched hash-left-join drives the seed in **windows** and re-scans the inner table per window. That is **not** a PR-4c regression — the *shipped* PR-4b does the same (q050: **92** inner scans, 9.3 s, shipped on DNF-flip + parity). The single-digit criterion was calibrated against an idealized scan-once model, not the shipped sibling's baseline; the resolution was checking that baseline rather than forcing either number. Filed as **F14** (`04-findings-register.md`): *batched R2RML OPTIONAL inner re-scans the main table per seed window* — affects both PR-4b and PR-4c; fix class = consume the whole seed in one inner scan + in-memory hash-join (a PR-4d candidate; prize q016 ~39 s→seconds, q050 ~9.3 s→~1 s). A properly-scoped follow-up, **not a blocker** — uniform with what already shipped.

### Gate

- **Differential** `batched_equals_per_row_on_object_correlated_star`: ONE mock (Shipment→Order ref + shipStatus scalar), FOUR required orders asserted row-for-row batched≡per-row — the CARTESIAN (order matched by 2 shipments), the LEFT-JOIN miss (0), a valid single (1), and a NULL star member (the star-conjunction drops the row — both paths agree). `r2rml_star_admission` pins admitted/excluded shapes. PR-4b tests still pass; full query lib **1231/0**.
- **Switch matrix**: `FLUREE_R2RML_BATCHED_OPTIONAL_STAR=0` ⇒ q050 byte-identical (PR-4b unaffected).
- q016 **rows-parity** (its `rows_only` hash-gate) + native corpus **54/54, 0 hash mismatches** + W3C + suites green. (The q050 native `SLOW` flag is F13 stale-baseline drift, already base-A/B-exonerated.)

Design note: `docs/audit/2026-07-virtual-dataset-perf/13-pr4c-optional-star.md`. This closes the corpus — the last DNF.


aaj3f:
Consolidating at maintainer request. #1493 (PR-4c — same-subject star admission to the batched OPTIONAL) is carried forward in #1507 (the R2RML/Iceberg virtual-dataset performance program), which brings the #1475–#1507 lineage forward as a single reviewable unit. Its verification of record is the program's corpus benchmark, published as [C2-bench-wave1.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C2-bench-wave1.md) (native == virtual, 0 hash mismatches). The pre-consolidation branch tip is preserved at tag `archive/pre-refactor-2026-07-21/perf_r2rml-pr4c-optional-star` — no history was rewritten. Closing in favor of #1507; discussion continues there.

Issue #1496, opened 2026-07-14
Display-profile rendering: compact Binding::Iri like stored/stamped IRIs (W3C surfaces settled by #1630)


aaj3f:
## Summary

While working on Iceberg/R2RML (virtual-dataset) pathway optimizations, we discovered that the **native SPARQL-results-JSON formatter renders IRIs provenance-dependently**:

- IRIs that come from **stored data** (`Binding::Sid`) are **compacted to CURIEs** using the query's `PREFIX`/`@context` declarations (via `compact_id_sid → compact_id_iri`).
- IRIs that are **constructed, federated, GRAPH-bound, or search-produced** (`Binding::Iri`) are rendered as **full IRIs**, *even when a declared prefix matches*, because the `Binding::Iri` arm of the formatter is written raw (`fluree-db-api/src/format/sparql.rs:333` streaming, `:487` DOM).

So two bindings with the **identical IRI value** can render differently based solely on their provenance. Example (a query declaring `PREFIX edw: <http://ns.fluree.dev/edw#>`): a stored `edw:name` predicate renders `edw:name`, but `BIND(IRI("http://ns.fluree.dev/edw#name") AS ?p)` renders `http://ns.fluree.dev/edw#name`.

As part of the virtual-dataset work we are aligning the **virtual** path to native's *current* internal model (scoped, so we do not change native output). **This issue raises the broader question: should both native and virtual change — and to what?** It is a product/spec decision, deliberately deferred out of the perf branches.

## (a) Where native produces `Binding::Iri` (the provenance-dependent, currently-raw sites)

All five are native-reachable (verified by workspace enumeration):

| site | trigger | reachable via |
|---|---|---|
| `fluree-db-query/src/eval/value.rs:522` | `ComparableValue::Iri` not encodable to a SID (code comment names `UUID`, `IRI()`) | `BIND(IRI("…"))` / `UUID()` / constructed IRIs |
| `fluree-db-query/src/sparql_results.rs:84,88` (via `fluree-db-api/src/remote_service.rs:153`) | remote SPARQL-Results `uri`/`bnode` term | `SERVICE <…> { }` federation |
| `fluree-db-query/src/graph.rs:343,440` (via `Binding::iri()`) | the `?g` graph-variable value | `GRAPH ?g { }` projection |
| `fluree-db-query/src/bm25/operator.rs:615` | full-text hit IRI not encodable to a SID | BM25 search (cross-ledger/unencodable hit) |
[...]

PR #1499, opened 2026-07-14
fix(format): CURIE-align virtual graph-source IRIs in sparql_json (F9)

aaj3f:
## What

Virtual (R2RML / Iceberg graph-source) queries rendered predicate and `@type` IRIs as **full IRIs** in SPARQL-results JSON, while native compacted them to **CURIEs** from the query's `PREFIX` declarations. This closes that divergence (finding F9) so virtual output matches native.

**q002 / q042 flip — now byte-identical to the native oracle:**

| query | virtual before (raw) | virtual after (this PR) | native oracle | rows |
|---|---|---|---|---|
| q002 | `8cd3964a…` | `d65ea385…` | `d65ea385…` | 8 |
| q042 | `0615a095…` | `33b91b54…` | `33b91b54…` | 24 |

Head row-diff verified: **every** changed cell is exactly full-IRI → CURIE — `<edw:openDate>`, `<edw:name>`, `<edw:channel>`, `<edw:storeId>`, `<edw:storeType>`, `<edw:geography>`, `<edw:regionManager>`, and the `rdf:type` object `<edw:Store>`. The `rdf:type` predicate stays a full IRI (no `rdf:` prefix declared), the `data.fluree.dev/edw/` reference objects stay full (that namespace isn't declared), and every literal is unchanged — all matching the oracle. **No baseline re-bless:** the `expected/` oracles were blessed from native and already carry the CURIE form, so virtual now simply matches them.

## Why (mechanism)

SPARQL-results-JSON compaction is `@context`/`PREFIX`-driven: `ContextCompactor::compact_id` keys **only** on the query's declared prefixes and never consults the namespace-code map. Native predicates arrive as `Binding::Sid` and run through `compact_id_sid → compact_id_iri` (compacts); virtual predicates arrive as `Binding::Iri` (a raw graph-source string) and were written **verbatim** at `format/sparql.rs` (streaming `write_term` and DOM `format_binding`), bypassing the compactor. The fix routes `Binding::Iri` through the same `compact_id_iri` native already uses — so parity holds by using the identical compaction call.

(This corrects an earlier "namespace-map gap" hypothesis: native renders `rdf:type` and `data.fluree.dev/edw/` references full *despite* both being in the ledger namespace map, precisely because compaction tracks declared prefixes, not the map.)

## Scope

- Change is gated three ways: `from_graph_source && FLUREE_R2RML_CURIE_ALIGN && format == SparqlJson`. Native results (`Binding::Sid`) are untouched; only virtual/graph-source `Binding::Iri` in `sparql_json` is affected.
- `QueryResult.from_graph_source` is set in exactly one place — `query_view_with_r2rml_options`, from the resolved view's `graph_source_id`. Native `query_with_options` and mixed-dataset paths leave it false (raw, unchanged); the native-ledger-reached-with-`.with_r2rml()`-but-no-mapping edge is documented at the set site.
- Other output formatters (JSON-LD, SPARQL XML, typed, delimited) share the compactor but receive the flag as `false` and keep raw graph-source IRIs — that cross-format consistency work is tracked as register **F16**.
- **Kill switch `FLUREE_R2RML_CURIE_ALIGN`** (default on) forces the pre-F9 raw rendering — powers the revert differential and gives production a revert lever.

Design alternatives were refuted at review gates: encoding at the R2RML operator (hot-path namespace work) and emitting `Binding::IriMatch` (unsound — it carries a load-bearing originating-ledger `primary_sid` that `eq_for_join` compares; virtual IRIs have none). The broader question of whether **native** should also compact its constructed/federated/GRAPH IRIs (a provenance-dependent-rendering finding) is deferred to fluree/db#1496.

## Blast radius — exactly q002 + q042

Among **full-hash-gated** queries, the switch moves only q002 and q042 — rerun-proven. The raw ON-vs-OFF differential listed more, which decompose cleanly: `hash_gate=rows_only` queries (unordered `LIMIT`) vary run-to-run by subset nondeterminism, not the switch (they project only literals — the switch is a definitional no-op); every other full-gated query was re-run with the switch off on a healthy network and is byte-identical to its switch-on hash. A transient Snowflake network reset during the original OFF phase errored a block of queries (confirmed transport, not the switch); all completed cleanly on rerun.

## Gates

- q002/q042 flip green vs the native oracle (heads verified full-IRI→CURIE only).
- Full virtual corpus ON compare: 54 records, **0 hash mismatches**, 0 perf violations (proper per-query gates).
- Native 54/54: 0 mismatches. W3C SPARQL suite: green. `fluree-db-api` lib: 664/664 including three new tests — a context-driven-compaction mechanism proof (namespace map irrelevant; `rdf:type` stays full), a flag-off-raw / flag-on-CURIE scoping guard on the same `Binding::Iri` row, and a DOM↔streaming parity assertion with a `Binding::Iri` row so both arms must flip together (closing the diagnosed partial-fix no-op loophole).

Stacked on `perf/r2rml-pr5-scan-topk`.


aaj3f:
Consolidating at maintainer request. #1499 (F9 — CURIE-alignment of virtual graph-source IRIs in `sparql_json`) is carried forward in #1507 (the R2RML/Iceberg virtual-dataset performance program), which brings the #1475–#1507 lineage forward as a single reviewable unit. Its verification of record is the program's corpus benchmark, published as [C2-bench-wave1.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C2-bench-wave1.md) (native == virtual, 0 hash mismatches). The pre-consolidation branch tip is preserved at tag `archive/pre-refactor-2026-07-21/fix_f9-virtual-curie` — no history was rewritten. Closing in favor of #1507; discussion continues there.

PR #1501, opened 2026-07-14
perf(r2rml): PR-4d batched-OPTIONAL seed coalescing (F14) — one inner scan for the whole driving side

aaj3f:
## What

The batched correlated-OPTIONAL hash-left-join (`optional.rs`) seeded and scanned its inner **once per OUTER (driving) batch**. When the driving side emits many small windowed batches (a fact scan under a LIMIT budget), the inner table is re-scanned per window — the **F14** residual left after PR-4b/4c collapsed the per-*row* rebuild. PR-4d coalesces the whole (bounded) driving side into **one** seed so the inner is scanned once.

## Mechanism

`OptionalOperator::next_batch` called `build_batch` once per required batch. New `pull_coalesced_required` concatenates required batches (column-wise, via `Batch::schema_arc()`) up to a cap into one combined batch, so `build_batch` seeds the inner over the whole distinct correlation set at once. Gated to the batched hash-join-safe path via `OptionalBuilder::supports_seed_coalescing()` (default false; `PlanTreeOptionalBuilder` = true iff every inner pattern is hash-join-safe) — so **only** the batched-OPTIONAL family (q016/q050) changes; every other OPTIONAL keeps the streaming single-batch pull.

- **Kill switch:** `FLUREE_R2RML_OPTIONAL_SEED_COALESCE` (default on). Off ⇒ single-batch pull, byte-identical.
- **Cap:** `FLUREE_R2RML_OPTIONAL_SEED_COALESCE_CAP` (default 512K rows). A LIMIT-bounded driving side exhausts below the cap (`… LIMIT k` coalesces exactly the ≤k rows into one scan); beyond the cap it degrades to cap-sized windows (still ≫ the pre-PR-4d one-per-outer-batch).

## Gate — REQUIRED per-head full-corpus baseline (priming + 3-rep, cache-thrashed, per-query manifest timeout)

**Oracle correctness: 54 records, 0 hash mismatches, 0 perf violations.** (The prior mini-gate compared nothing to the blessed oracles; this is the real check.)

| query | metric | pre / OFF | post / ON (3-rep median, in-protocol) |
|---|---|---|---|
| **q016** | `scan_table` | 182 | **3** |
| q016 | wall | 43.1 s | **17.47 s** (2.5×) |
| q016 | rows / oracle | 5000 | 5000, hash-clean |
| **q050** | `scan_table` | 63 | **5** |
| q050 | wall (3-rep median) | 1.24 s (F9 head) | **1.11 s** (marginally *faster*) |
[...]

PR #1502, opened 2026-07-15
perf(r2rml): F20 RefObjectMap-target resolution prune — q031 72s→5.2s (kill the shared-predicate fan-out)

aaj3f:
## F20 — RefObjectMap-target resolution prune (kill the shared-predicate fan-out)

Stacked on `perf/r2rml-pr4d` (#1501). Kills q031's over-scan: a star subject `?p` bound solely as one `RefObjectMap` object (parent class C) has its downstream single-predicate star's TriplesMap resolution constrained to C, so `?p edw:name ?pn` no longer fans out to all six `edw:name`-bearing dims.

### q031 — honest numbers (cache-thrashed, full-corpus 3-rep protocol)

| | load_table.n | scan_table.n | rows | wall |
|---|---:|---:|---:|---|
| OFF (pre-F20 fan-out) | 7 | 1448 | 5000 | 154.0 s |
| ON (F20 prune) | 2 | 243 | 5000 | **5.21 s** (3-rep median, in-protocol) |

**72 s → 5.21 s (~14×)**, a deterministic fan-out kill (7→2 loads, 1448→243 scans) — the load/scan collapse is cache-independent; the wall is the measured cache-thrashed 3-rep median. `rows_only` parity (row count ON==OFF; hashes differ by unordered-LIMIT nondeterminism, expected).

### Residual (measured, not inferred — doc 20 §8)

The 243 surviving scans decompose (per-scan trace, reverted) as 2 fact + 1 deduped DIM_PRODUCT parent lookup + **241 budgeted DIM_PRODUCT MAIN re-scans** of `?p edw:name` (`row_budget=Some(5000)` alone disqualifies the `scan_cache`; `filters=0`, not parent lookups). An A/B differential collapsing those 241→3 measured **Δ −0.5 s** — they cost nothing. A query-scoped small-dim main-scan cache is therefore **measured-and-rejected**. q031's residual is the fact-scan + materialize + `loadTable` floor — the **same floor as q016**-post-#1501; the isolated first-ask still pays ~2–3 s of `loadTable` (amortized to load=0 in corpus order). Named levers to close ≤3 s (not in this PR): the loadTable-METADATA cache + PR-2a / materialization.

### The verdict chain (six evidence-forced flips — preserved in the register)

q031 cost six flips: PR-5-suspected → bisect-refuted → memo×limit-pushdown → cold-floor loadTable-pin → **measurement: RefObjectMap-target fan-out (F20)** → the residual measurement above (small-dim cache rejected). Along the way **F19** (the "query-scoped parent-memo doesn't engage" theory) was **refuted** — the memo already shares on q031's path (`with_active_graph` clones the Arc; confirm-experiment 1× shared vs 5× fresh-ctx control); its one sound residual fix (`with_graph_ref` memo-clone) is filed and deferred, affecting no corpus query.

### Soundness (the load-bearing part)

The prune reuses the existing `class_prune_hint` verbatim (zero operator change). It fires only when BOTH:
- **Invariant A (join-var provenance):** `?p`'s only binding in scope is exactly one `RefObjectMap` member (unique parent class C). Computed via **`Pattern::produced_vars`** (not `referenced_vars`) — so a `BIND` target or a path endpoint (which *rebind* `?p`) decline, while `FILTER`/`MINUS`/`(NOT) EXISTS` (which only *constrain*) do not. One principled choice; future IR pattern types inherit correct treatment. Declines on: a second object binding, a UNION/VALUES-bound `?p`, a second `RefObjectMap` with a different parent, a mixed ref/non-ref predicate, an ambiguous parent class, or a standalone class assertion on `?p`.
- **Invariant B (template-disjointness):** `wildcard_class_fusion_is_safe(C)` — the F10 vertical-partition guard, verbatim; a map sharing C's subject template that also maps the predicate declines.

Cross-scope pre-bound `?p` (seeded outside the scope) is sound **by conjunction**: the in-scope ref member restricts `?p` to class-C FK objects regardless of the prune (engine-tested via CountingProvider: valid product 1→1 row; non-product Store IRI → 0 rows, identically pruned or not).

**Kill switch:** `FLUREE_R2RML_REF_TARGET_PRUNE` (default on) — off is byte-identical to the pre-F20 fan-out, as is every DECLINE path.

### Tests + gate

12 tests: 10 hermetic rewrite (positive + BIND/UNION/VALUES/2nd-binder/different-parent/template-share/class-assert declines + FILTER-no-decline) + 2 engine-level (cross-scope conjunction). Full `fluree-db-query` lib green (1247). Full-corpus baseline at this head: **54 records, 0 hash mismatches, 0 perf violations**; the 3 expected-error queries unaffected; 44/51 ok-queries ≤3 s.

CI note: this PR's base is `perf/r2rml-pr4d`, not `main`, so repo CI does not fire on it (stacked-PR convention).


aaj3f:
Consolidating at maintainer request. #1502 (F20 — the RefObjectMap-target resolution prune) is carried forward in #1507 (the R2RML/Iceberg virtual-dataset performance program), which brings the #1475–#1507 lineage forward as a single reviewable unit. Its verification of record is the program's corpus benchmark, published as [C2-bench-wave1.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C2-bench-wave1.md) (native == virtual, 0 hash mismatches). The pre-consolidation branch tip is preserved at tag `archive/pre-refactor-2026-07-21/perf_r2rml-q031-refprune` — no history was rewritten. Closing in favor of #1507; discussion continues there.

Issue #1500, opened 2026-07-14
Iceberg on operator-owned S3: use inline metadata, fail closed on vended downgrade, hydrate secret refs, hedge the disk-cache key


aaj3f:
## Summary

Four engine-side changes needed to support reading Snowflake-managed Iceberg tables whose data files live in an **operator-owned S3 bucket**, via either credential path — while holding the line that **performance must not degrade under any credential configuration**.

Two of the four are unconditional wins that stand on their own merits, independent of any customer:

1. **Use the inline `metadata` the catalog already hands us** — saves ~1 S3 round-trip per table, in every configuration, on the *first* cold touch.
2. **Fail closed on the silent vended→ambient downgrade** — a security fix that also establishes the invariant the disk cache's safety argument depends on.
3. **`ConfigValue::SecretRef` + `hydrate()`** — the engine half of the credential-rotation fix.
4. **A scope discriminant on the disk cache key** — a ~10-line hedge that makes per-graph-source identity (mechanism B) a one-liner later instead of a re-architecture.

All anchors verified at `a8de70e0f64454e4e3c736f203a9de4499dece45` (the fully-optimized head, PR-5 tip).

---

## 1. Use `load_response.metadata` instead of re-fetching it from S3

**We are issuing an S3 GET per table for bytes the catalog already gave us.**

`LoadTableResult.metadata` is a spec field returned inline on a successful `loadTable`, and the engine already parses and retains it — `fluree-db-iceberg/src/catalog/rest.rs:377-383`, `catalog/mod.rs:31-36`:

> "so callers can read the full table schema/snapshot without a second S3 fetch"

But the query path never reads it. `r2rml.rs:1343-1390` goes in-memory cache → disk cache → **`storage.read(metadata_location)`** — a fresh S3 GET. The intended consumers are named at `rest.rs:377-379` ("metadata preview"); the read path was simply never wired up.

**Live-verified against Snowflake Horizon** (account `ABACYOU-PP85756`, customer-owned external volume, `2026-07-14`): `loadTable` returns **3,438 bytes** of inline `metadata` with full `schemas`, `current-snapshot-id`, and 2 `snapshots`. So `rest.rs:378`'s claim ("Snowflake Horizon / Polaris include it") is confirmed live, not just asserted in a comment.

[...]

PR #1503, opened 2026-07-15
perf(iceberg): loadTable-METADATA cache — zero-REST first-ask (q031 72s→~6s arc) + F19 residual

aaj3f:
## loadTable-METADATA cache — persisted credential-free pointer + lazy storage (+ F19 residual)

Stacked on `perf/r2rml-q031-refprune` (#1502). Closes the isolated first-ask ~2–3s `loadTable` REST/OAuth cost that F20/PR-4d left on q031/q016 (on top of the shared fact-scan/materialize floor). 4 commits, +906/−10.

### The win (deterministic gate — GREEN, both runs)
Warm-disk / cold-memory, fresh process per query. The COUNTERS are the deterministic gate; walls vary with cache warmth (two runs shown — pre- and post- the pin fix, which perturbed nothing):
| q | pointer cache ON | OFF (old eager path) |
|---|---|---|
| q031 | **load 0 / oauth 0** / 5000 rows / 5.9–6.5s | load 2 / oauth 1 / 5000 / 10.3–14.9s |
| q016 | **load 0 / oauth 0** / 5000 rows / 17.0–25.6s | load 2 / oauth 1 / 5000 / 20.6–28.5s |

When a query's metadata (slice-2), scan files (slice-2), and every Parquet file (DiskArtifactCache) are all disk-resident, it now issues **zero loadTable GET / OAuth / S3**. OFF is byte-identical to the old path; row parity holds. q031's first-ask is **72s → 5.9s** across this arc (F20 + this). Full-corpus baseline: 54/54 oracle 0-mismatch, 44/51 ok ≤3s (corpus-order loads stay amortized, so full-run walls track #1502 — this closes the per-query first-ask the amortized baseline hid).

### Mechanism
**(a) Persisted credential-free pointer** (`disk_catalog_cache.rs`): a `lt_key → metadata_location` entry (same Envelope + `CACHE_FORMAT_VERSION` discipline). Payload is the location + `snapshot_ms` + `cached_at_ms` ONLY — no creds/token/config. Switch `FLUREE_ICEBERG_LOADTABLE_PTR_CACHE`; TTL `FLUREE_ICEBERG_LOADTABLE_PTR_TTL_SECS` (300s, 0=off).

**(b) `LazyS3Storage`** (new `lazy_storage.rs`): impls `SendIcebergStorage` over a tokio `OnceCell` + a deferred builder. The pointer rung in `load_table_context` serves `metadata_location` (pointer) + metadata (caches) with no S3, and defers storage; the reader only forces the builder on a real S3 miss. The eager REST client is built up front — construction does **no** network; the OAuth token exchange rides the first catalog op (the GET), so deferring the GET zeroes both counters (**the gate's `oauth_token.n=0` enforces this forever**). The `'static` builder captures owned Arc handles (the provider holds `fluree: &'a`, which can't cross the Parquet `tokio::spawn`) and runs `resolve_rest_load_and_storage` — the SAME session-pin ladder as the eager path (no duplicate; full slice-1 reuse incl. the cross-query put). `LazyS3Storage` is `Clone` via an Arc-shared cell (`SendParquetReader` requires `S: Clone`); clones share one build.

### Security (creds never persisted)
Hermetic `pointer_persists_no_credential_bytes` asserts the on-disk pointer has exactly `{metadata_location, snapshot_ms, cached_at_ms}` and no credential/token/secret/access_key/session/sig bytes. Vended creds are fetched in-memory, per process, only when the lazy handle is forced (a real S3 miss).

### Snapshot-pin correctness (review fix, commit 6750286df)
The pointer rung consults `session.pinned_metadata_location` BEFORE the disk pointer (a prior touch's pin wins), and registers a **location-only** pin on a hit — so a query that touches the same table twice reads ONE snapshot even if the pointer's TTL expires between touches. Kept in a separate `location_pins` map so `cached_load_table` never serves a credential-less entry (which would build ambient-cred storage); the eager path reloads only creds and converges on the pinned location. Hermetic `location_pin_keeps_query_on_one_snapshot_across_pointer_ttl_expiry`.

### TTL staleness semantics
The pointer (unlike content-addressed entries) CAN go stale on a table commit, so it carries a TTL = the max cross-process snapshot staleness for a latest-snapshot read — the same tradeoff class the existing 60s in-memory cross-query cache accepts, extended deliberately, consistent with the disk-cache-is-steady-state ruling. A stale pointer never returns WRONG data for the snapshot it names; it can only lag a newer snapshot within the TTL.

### as_of_t forward-guard + wiring obligation
Iceberg snapshot time-travel is deliberately unimplemented on the scan path (`grep r2rml-as-of-t`), so the caller passes `min_snapshot_ms=None` (latest). The guard (`pointer_is_usable`) + hermetic `pointer_ttl_and_as_of_rider` exist; doc-21 §4 + the register carry the obligation: whoever implements time-travel MUST thread the requested snapshot ms into `get_metadata_location`.

### F19 residual (ride-along, commit ebcab7594)
`with_graph_ref` now clones the R2RML parent-memo Arc (its key is store-disambiguated `(graph_source_id, parent_tm, cols, as_of_t)`, unlike `const_sid_cache` which stays fresh for the store switch). Closes the F19 residual; hermetic `parent_memo_survives_with_graph_ref_rebuild`. No corpus query exercises SERVICE/multi-source R2RML, so gate-neutral.

### Tests
lazy_storage 3/3 (deferred-not-invoked, 16-concurrent-forces→1-build, no-panic error path); disk_catalog_cache 8/8; catalog_session 7/7; full `-p fluree-db-api --lib` 765/0; it_graph_source_r2rml 53/53. All fluree-db-api test targets compile.

CI note: base is `perf/r2rml-q031-refprune`, not `main`, so repo CI does not fire (stacked-PR convention).


aaj3f:
Heads-up from the sibling branch `feat/external-s3-iceberg` (BYO-IAM engine work for #1500, forked at the same base `eee643572` / #1502): there is a **known semantic interaction** to reconcile when these two land, beyond the expected `r2rml.rs` textual conflicts.

That branch adds §2 fail-closed credential handling: in the REST arm, `vended_credentials == true` + `load_response.credentials == None` now **errors** (`err:catalog/CredentialsNotVended`, 403) instead of silently downgrading to ambient AWS credentials. The decision matrix lives in one shared `decide_credential_source()` used by both the scan and preview paths.

If this PR's zero-REST first-ask path synthesizes/reconstructs a loadTable response with `credentials: None` for a vended-configured source, the composed branches would 403 every cached first-ask (loud, not silent — but broken). Whoever rebases second should make the cached path either (a) carry an explicit "credentials come later / from cache" marker through the decision, or (b) join behind the same decision matrix the way the existing per-query pin and 60s loadTable caches do (they preserve `credentials` verbatim, which is what keeps the §2 guard from false-firing on today's cache hits).

Related, in your favor: that branch deliberately kept the disk `metadata` cache layer as the fallback for responses carrying `metadata: None` (only inline-vending REAL loadTable responses bypass it), so a zero-REST path still has its cross-process metadata source.

aaj3f:
Consolidating at maintainer request. #1503 (the loadTable-METADATA cache — a credential-free pointer + lazy storage for a zero-REST first-ask) is carried forward in #1507 (the R2RML/Iceberg virtual-dataset performance program), which brings the #1475–#1507 lineage forward as a single reviewable unit. Its verification of record is the program's corpus benchmark, published as [C2-bench-wave1.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C2-bench-wave1.md) (native == virtual, 0 hash mismatches). The pre-consolidation branch tip is preserved at tag `archive/pre-refactor-2026-07-21/perf_r2rml-loadtable-meta-cache` — no history was rewritten. Closing in favor of #1507; discussion continues there.

PR #1508, opened 2026-07-15
Fix duplicate triples from re-asserted facts in the V6 incremental indexer

### LINKED ISSUE #1412: [BUG] Re-inserting existing triples yields duplicate triples + assert/retract churn; upsert of existing triples registers no new transaction
**Source:** Fluree Solo internal QA (surfaced via Solo, appears engine-level)
**Reporter:** Laurent Bégin
**Reported:** 2026-06-25
**Severity:** Critical (data integrity)

## Summary
Re-inserting the same triples leaves the KG with **duplicate triples**, and the transaction history shows an **assert / assert / retract churn** on those triples. Separately, **upsert** of existing triples registers **no new transaction** while **insert** does (re-asserting existing triples) — likely the same underlying mechanism.

## Environment
Solo stack `lbegin-test.fluree-emea` (Radio Canada / CBC, `KG05`), which uses `fluree-db-api`. Surfaced through the Solo Data/Model UI.

## Symptoms
1. **Duplicate triples.** After multiple inserts of the same triples, an instance ends up with duplicated object values — e.g. "Québec (province)" shows property `générique` (`publishing#BT`) with the same object IRI `b32e4947-…` repeated ~16×, and `groupe` (`t3#inGroup`) values repeated many times.
2. **Retract churn in history.** Transaction `t:476` (42 changes) shows a repeating **assert / assert / retract** pattern per value (e.g. `générique` +,+,−; each `groupe` value +,+,−).
3. **Not reproducible with a single triple.** Laurent could not reproduce by inserting the same *single* triple repeatedly — only with multi-triple inserts.
4. **Upsert vs insert asymmetry (related question).** Upserts of existing triples do **not** register as new transactions (the API returns the last transaction info); inserts **do** register new transactions that re-assert existing triples. End result is the same (no net data change), but the transaction log / dedup behavior differs.

## What we need from the DB team
- Confirm whether this is **engine-level** insert/novelty dedup behavior or a **Solo-transact payload-construction** issue (i.e. Solo emitting each triple twice).
- Clarify the **expected insert vs upsert semantics** for already-present triples.
- Determine why duplicates **persist in the indexed/materialized view**.

## References
- Slack — report (fluree-internal #mr1-qa-feedback): https://fluree-internal.slack.com/archives/C0BASKBSNTS/p1782379252869899
- Slack — related upsert/insert question: https://fluree-internal.slack.com/archives/C0BASKBSNTS/p1782379136620289
- Screenshots (internal): instance duplicate values; history `t:476` assert/retract pattern.

_Filed in fluree/db per triage; surfaced via the Solo QA workflow (Slack channel above)._

zonotope:
Re-inserting triples that already exist in the database left the indexed view with duplicate rows: the same subject-property-object appearing many times (observed ~16× in the reporting ledger), along with spurious assert/retract churn in transaction history views. Live queries against in-memory novelty were correct; the duplicates appeared only after an incremental index build and compounded with each subsequent one. A full rebuild silently healed them, which made the bug look intermittent.

Root cause: commit files are event logs of user assertions, not state deltas — re-asserting a present fact is a legitimate event that records a new transaction, and every layer that materializes state from those events must apply RDF set semantics itself. Three of the four materializers did (in-memory novelty via `NoveltyFactState`, the query overlay via `resolve_overlay_ops`, and the full rebuild via `KWayMerge::next_deduped`), but the V6 incremental pipeline replayed raw commit events with no lifecycle resolution: `resolve_incremental_commits_v6` sorted records without collapsing them, and `merge_novelty`'s two-pointer walk assumes at most one op per fact identity, so each duplicate assertion event materialized as an additional physical row in the FLI3 leaves.

### Set-semantics dedup in the incremental resolve (the core fix)

`dedup_fact_lifecycles` runs after the `(g_id, SPOT)` sort in `resolve_incremental_commits_v6` and resolves each fact's in-window event sequence to a single final op, mirroring the in-memory novelty rules exactly so the updated index materializes the same state a live query resolves: a re-assert of a currently-asserted fact is a no-op (the earlier assert and its `t` survive, matching `NoveltyFactState`'s skip rule), any other op becomes the identity's winner, and same-`t` ties resolve retract-wins (matching `resolve_overlay_ops`).

Fact identity is the resolved V2 identity `(o_type, o_i)` — the identity `merge_novelty`, replay, and the query overlay all compare by — rather than raw v1 fields. This matters because `OTypeRegistry::resolve` collapses the datatype dict id for 1:1 object kinds (a boolean is `xsd:boolean` whatever datatype IRI the commit carried) and consults `lang_id` only for `rdf:langString`, so datatype-IRI variants and stray language tags on one fact dedup together while genuine language tags and list positions stay distinct. The dedup compacts in place behind a write cursor: a window with no duplicate facts copies nothing, and multi-record groups resolve through a reused scratch buffer in O(k log k).

### Correct stats deltas via base presence

The signed stats hooks (per-property/datatype/graph flake counts and rdf:type class counts) previously derived ±1 deltas from each record's op against base-seeded counts, which double-counts re-asserts and, once lifecycles collapse, applies unbalanced deltas for retract-then-reassert and assert-then-retract windows. The Phase 2 SPOT merge now reports which novelty ops matched an existing base row (`MergeOutput::matched`, threaded through `update_leaf` → branch update → `Phase2TaskOutput`, collected for the SPOT order only), and the Phase 3b feed uses a new `IdStatsHook::on_record_with_base_presence`: +1 only for an assert of a base-absent fact, −1 only for a retract of a base-present one, 0 otherwise. Monotone accumulators (HLL sketches, touched-property sets, `last_modified_t`) still observe every record. The from-scratch event-replay feeds (rebuild/import) keep the unchanged legacy `on_record`. This also fixes a pre-existing bug: every re-assert of an indexed fact used to permanently inflate its property count.

### No history entries for non-transitions

The history sidecar is a transition log: time-travel replay undoes a retract entry by marking the fact present before it, so a retract entry with no paired assert anywhere makes replay fabricate the fact's presence at every `t` before the retraction. `merge_novelty` no longer writes a history entry for a retract of a fact the leaf never materialized (the `Greater` arm and the post-rows drain loop), and the rebuild path skips retract-winner lifecycles whose identity has no assert anywhere in the commit log. This fixes both the ghost introduced by lifecycle collapsing (a fact asserted and retracted within one window deduped to a bare retract) and a pre-existing ghost when a user retracts a fact that never existed.

## Known limitations and follow-ups

- A fact asserted and retracted within a single indexing window is invisible to index-served time travel during its brief lived interval (that interval was never materialized in any published leaf; before this branch it was "visible" only via the buggy ghost row). The live novelty view retains those intermediate ops until the window is indexed. Restoring full in-window fidelity requires a transition-log rework of the history sidecars; the full analysis, design, risks (including mixed-vintage sidecar compatibility), and testing strategy are captured in `history-transition-log-plan.md`.
- The full rebuild path resolves a re-asserted fact to the highest-`t` assert while novelty and the incremental path keep the first assert's `t`, so the two build paths can materialize different row `t` for the same event log. Pre-existing, out of scope here; folded into the plan file as a blocking design decision for the rework.
- Already-published leaves keep their duplicate rows, corrupted stats, and stale bare-retract history entries until affected ledgers run a full reindex; this branch stops new damage.

Fixes #1412

### DIFF
--- a/fluree-db-indexer/src/build/incremental.rs
+++ b/fluree-db-indexer/src/build/incremental.rs
@@ -390,6 +390,20 @@ struct Phase2Task {
     kind: Phase2TaskKind,
 }
 
+/// Graph-scoped V2 fact identity: `(g_id, s_id, p_id, o_type, o_key, o_i)`.
+type FactIdentityKey = (u16, u64, u32, u16, u64, u32);
+
+fn fact_identity_key(rec: &RunRecordV2) -> FactIdentityKey {
+    (
+        rec.g_id,
+        rec.s_id.as_u64(),
+        rec.p_id,
+        rec.o_type,
+        rec.o_key,
+        rec.o_i,
+    )
+}
+
 enum Phase2TaskUpdate {
     Default { leaf_entries: Vec<LeafEntry> },
     Named { branch_cid: ContentId },
@@ -404,6 +418,9 @@ struct Phase2TaskOutput {
     replaced_sidecar_cids: Vec<ContentId>,
     new_leaf_count: usize,
     fetch_stats: Phase2FetchStatsSnapshot,
+    /// Novelty ops whose identity matched an existing base row (SPOT task
+    /// only; fresh-graph tasks have no base rows to match).
+    matched: Vec<RunRecordV2>,
 }
 
 #[allow(clippy::too_many_arguments)]
@@ -434,6 +451,9 @@ async fn execute_phase2_task(
             .leaflet_rows
             .max(1)
             .saturating_mul(config.leaflets_per_leaf.max(1)),
+        // Fact identity is order-independent; collect base-row matches from
+        // the SPOT task only (they feed the Phase 3b stats deltas).
+        collect_matched: order == RunSortOrder::Spot,
     };
 
     let started = Instant::now();
@@ -472,6 +492,7 @@ async fn execute_phase2_task(
                 replaced_sidecar_cids: meta.replaced_sidecar_cids,
                 new_leaf_count: meta.new_leaf_count,
                 fetch_stats,
+                matched: meta.matched,
             }
         }
         Phase2TaskKind::DefaultFresh => {
@@ -499,6 +520,7 @@ async fn execute_phase2_task(
                 replaced_sidecar_cids: Vec::new(),
                 new_leaf_count,
                 fetch_stats: Phase2FetchStatsSnapshot::default(),
+                matched: Vec::new(),
             }
         }
         Phase2TaskKind::NamedExisting { branch_cid } => {
@@ -537,6 +559,7 @@ async fn execute_phase2_task(
                 replaced_sidecar_cids: meta.replaced_sidecar_cids,
                 new_leaf_count: meta.new_leaf_count,
                 fetch_stats,
+                matched: meta.matched,
             }
         }
         Phase2TaskKind::NamedFresh => {
@@ -573,6 +596,7 @@ async fn execute_phase2_task(
                 replaced_sidecar_cids,
                 new_leaf_count,
                 fetch_stats: Phase2FetchStatsSnapshot::default(),
+                matched: Vec::new(),
             }
         }
     };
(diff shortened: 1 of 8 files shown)


bplatz (approved):
Awesome, thanks!

PR #10, opened 2020-10-30
Feature/syncto improvements

bplatz:

PR #1514, opened 2026-07-17
feat(virtual): serverless Lambda-usability + BYO-IAM — admission widenings, capacity guards, typed memory budget (consolidates #1505 + #1514)

### LINKED ISSUE #1500: Iceberg on operator-owned S3: use inline metadata, fail closed on vended downgrade, hydrate secret refs, hedge the disk-cache key

aaj3f:
feat(virtual): serverless Lambda-usability + BYO-IAM — admission widenings, capacity guards, typed memory budget (consolidates #1505 + #1514)

This is one of five surviving pull requests in the consolidated virtual-dataset review program, which collapses the #1475–#1529 lineage into five reviewable units at the maintainers' request. #1514 makes R2RML/Iceberg virtual-dataset chat queries usable in the serverless (Solo / AWS Lambda) scenario and folds in the BYO-IAM (operator-owned S3) engine work; it carries the program's single most important native call-out, the R3 typed memory budget. It is based on #1507 (so its own diff is exactly #1505 + #1514), and reviewers rely on the pasted gate record rather than CI. Reading order across the program: #1520 (standalone MoR correctness fix) → #1507 (performance program) → **#1514 (this PR)** → #1528 (big-Iceberg-audit implementation) → #1529 (native-twin materialize builder). The outward-facing audit record is on the `docs/virtual-dataset-audit-2026-07` branch — the serverless envelope this epic targets is in [C1-constraint-envelope.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C1-constraint-envelope.md) and the memory-budget finding the R3 change protects against is in [V2-membudget-verification.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/V2-membudget-verification.md).

## Original PR body

_Scope expanded by consolidation; the tip PR's own change (the Lambda-usability epic) is the narrative below, and it folds in the BYO-IAM engine work (#1505) — rolled up under "Consolidated scope."_

Makes Iceberg/R2RML virtual-dataset chat queries "reasonably usable" in the solo/Lambda scenario. Grounded in the Lambda-reality assessment (`docs/audit/2026-07-lambda-reality/`) and two rounds of post-deploy forensics (`~/fluree/db/lambda-round2-*.md`, `lambda-round3*-*.md`). Three waves: C1-C5 (the original epic), E1-E2 (round-2 fusion admission), R3-A/B (round-3 capacity protection).

## Component checklist

- [x] **C1 — `DatasetOperator` forwards the `LIMIT` budget / top-k to members.** Every chat SPARQL is a `FROM <ledger>` dataset query; the dataset operator inherited the no-op budget hooks — a `LIMIT 20` never reached the R2RML scan (observed live: 8M-row full scan). Switch `FLUREE_R2RML_DATASET_BUDGET`.
- [x] **C2 — manifest-backed info-stats routing.** The 900s `get_data_model` runaway. **Live-verified fixed: 4s.**
- [x] **C3 — mid-scan cancellation** in the R2RML drain helpers. **Live-verified: scan-bound queries cancel cleanly at the deadline.**
- [x] **C4 — intra-file row-group parallelism** for single/few-file tables. Switch `FLUREE_ICEBERG_ROWGROUP_PARALLELISM`.
- [x] **C5 — fused-aggregate admission for the deployed rollup shapes** (adversarially gated): plain-literal group-key default matching materialization by construction, guarded by O1 (`star_constraints` decline), O2 (single-source-view dataset guard with `(ledger, to_t, policy)` dedup), Q2 (lang/IRI group-key decline, both fold paths).
- [x] **E1 — shared group-key predicate fusion** (round-2: the `GROUP BY category` 2-scan timeouts). Fuse when a class-declaring map co-locates the predicate and all other maps are subject-template-disjoint (structurally provable). Vertical-partition/overlap cases still refuse.
- [x] **E2 — join+flag rollups fuse with constraints APPLIED** (round-2: the `isCurrent true` join timeouts). Dim-side constraints filter during FK-map build; fact-side ride the fold; anything unresolvable declines to materialize (enumeration adversarially verified total). **Live: 52s → 0.45s with the identical constrained count.**
- [x] **R3-A/B — unified `ctx.checkpoint()`: cancellation + typed memory budget in the join/aggregate phases** (round-3: OOMs at 344s/818s ran 2-4.5× PAST the deadline in post-scan phases with zero polls — 813s of log silence on the worst; OOMs at 156-170s died BEFORE any deadline, where cancellation can't help). One mechanism: a query-scoped memory counter lives on the shared cancellation handle (every operator of a query reaches the SAME counter — a join build and a downstream fold share one budget), and `checkpoint()` = cancellation check first, then budget — wired at the 7 buffering sites (hash-join build/probe, both aggregate folds, and the fused-fold's map builds, so any future interior-FACT map is covered from day one). The ~85 non-buffering scan drains keep plain cancellation checks deliberately. Typed `MemoryBudgetExceeded` → 507, distinct from the 408 timeout; an external cancel (deadline / RSS watchdog) can never be masked as a 507 (cancel-first ordering, adversarially verified). Core crate carries pure mechanism (never enforces); enforcement lives only in the query engine. Also makes solo's deployed RSS memory-watchdog bite inside the fold — in the round-3b window, 3 hard OOMs slipped past that watchdog exactly where these checkpoints were missing. **Knobs**: `FLUREE_QUERY_MEMORY_BUDGET_BYTES` overrides; `0` disables; default = ~78% of the *detected* container/system memory (cgroup v2 → v1 → /proc/meminfo; 8GiB fallback) — proportional to the machine, not a fixed ceiling. There is **no engine-imposed time limit anywhere**: these are polls only; deadlines are armed solely by callers.

- [x] **W4-1 — constant-key scan-filter pushdown + datatype coercion** (round-3b: a point lookup full-scanned 120M+36M rows into a hard OOM). A constant-object key equality now pushes as a scan filter even when the pattern carries other predicates (previously only a LONE constant pushed; any second predicate demoted it to a post-read residual). String literals coerce to the column's type ONLY on canonical round-trip (lexical-eq ⟺ typed-eq; `"01"`/`"1.0"`/non-numeric decline the push — residual stays the sole authority, no over-prune possible).
- [x] **W4-2 — mixed fact+dim GROUP BY keys fuse** (round-3b `#7`: a 36M two-hop rollup grouped by dim-year + fact-shipMethod declined → 156s materialize → deadline). Group keys now resolve per-key to their binding pattern (fact column or terminal-dim attribute; a var bound in ≥2 patterns is a cross-source equality and DECLINES — adversarially derived rule). Live: the corpus analogues run at ~110ms fused.
- [x] **W4-1b — const-object folds onto the co-located crawl wildcard** (the deployed detail-crawl shape: `?x key "v" . ?x ?p ?o`). The key constant and the wildcard previously ran as SEPARATE JOINED SCANS — the constant never pruned the crawl. The fold attaches it as a subject-level constraint; the executor treats a variable-predicate wildcard as a true wildcard at all sites (a mid-review ship-blocker: the first cut produced an INEXECUTABLE plan returning subject-only rows — caught by the layered review, fixed, and pinned by an execution hermetic asserting bound `?p`/`?o` plus corpus sentinel q068).
- [x] **F-16 amendment — scalar-gated `trust_fk_refs` restore + wildcard TM-prune guard.** W4-1b's constraint fold had disabled the FK-template ref shortcut by construction (catch `#10`) and would have TM-pruned vertically-partitioned wildcard subjects (latent wrong-results). The shortcut is restored for folded wildcards ONLY when every folded constraint is scalar — an IRI constraint would let a DANGLING FK's templated render falsely satisfy the existence check (catch `#11`), so non-scalar keeps the sound parent-scan path.

## What this changes for the deployed workload (honest attribution)

1. **The production rollup timeouts are fixed by this PR** (single-table SUM/COUNT, shared-predicate `category`, flagged join rollups).
2. **Fact⋈dim rollups were already fused in the deployed engine** — not what timed out; no credit claimed.
3. **Three latent correctness bugs in already-shipped code closed** (unguarded fused-join flag over-count; blind lang/IRI string default on both fold paths).
4. **The heavy multi-FACT class (120M⋈36M joins) now fails fast and typed instead of burning 10GB for 13 minutes.** Round-3 analysis: that class is unfusable by design (five orthogonal gates); the make-it-succeed levers (dim-scan semi-join, mixed fact+dim group keys, LIMIT-through-joins, spilling) are scoped follow-ups. R3-A/B is capacity protection, deliberately not a throughput claim.
5. **The deployed detail-crawl ("show me everything about X") is fixed for correctness and pruning; its ref parent-scan cost splits by invocation path.** Chat SPARQL runs via `query_from` where `trust_fk_refs` is OFF — parent scans for ref rendering predate this PR there and remain (bounded by the R3 guards; F-16 stays open for parent-lookup bounding on that path). The browse-crawl API path (`trust_fk_refs` ON) gets the full win: pruned child + scan-free ref render via the restored scalar-gated shortcut, evidenced by the integration hermetic (landing pre-merge) since no SPARQL corpus query can exercise the trust path.

## Verification

- fluree-db-query lib 1277 green (TDD tripwires per wave); vbench bin green; scoped clippy clean throughout.
- **Full 65-member live corpus at every wave head — 0 hash mismatches in every run** (R3 head `33a7d540b` inclusive — both the separate-mechanism head and the unified refactor gated identically, zero behavioral change). New members validate every deployed failure shape: q060-q065 (budget/flag/join/sort/shared-predicate/join+flag paths), oracles blessed from fusion-OFF runs.
- Perf-gate violations in every run ⊂ the documented loadTable-GET latency set {q002,q004,q024,q030,q043} (F20/F21) — known-cause, non-fused queries, run-to-run drift.
- **Wave-4 gate (68 members, live SF01): 0 hash mismatches.** New members: q066/q067 (mixed fact+dim keys, COUNT and COUNT+intSUM — 105-116ms fused), q068 (folded-crawl sentinel — 200 rows via the FROM path, validating the ship-blocker executor fix end-to-end). Perf violations: q004/q022/q024/q030/q043 — telemetry-verified as the documented loadTable-GET latency class (q022 is a NEW member of that drift family: its wall growth is 100% catalog latency — loadTable 1.09s + scan-plan 1.48s vs ~60ms of actual data work, scan shape identical to baseline; no wave-4 code path touches it).
- Wave-4 provenance: implemented across three agent generations plus a team-lead-implemented final amendment, with 11 substantive adversarial-review catches across the epic — three of them (an inexecutable plan, a disabled-by-construction shortcut, and a dangling-FK over-match) in the W4-1b lineage alone, none reaching a pushed branch.
- **Memory-accounting scoping (from the adversarial passes, kept honest)**: per-row byte estimates under-count large heap-strings (long IRIs, big literals, GROUP_CONCAT) — "over-count-safe" is scoped to small-binding builds like the deployed DW workload. RSS can also grow in untracked scan-stage buffers; the layering is: **engine budget fires first for tracked join/aggregate accumulation; an RSS-level watchdog (solo ships one) remains the necessary backstop for untracked growth.** The counter is query-lifetime-cumulative (sums sequentially-freed operators) — strictly over-count-safe; a per-operator scope is a scoped follow-up only if legitimate multi-join queries ever trip it. The engine's 507 body carries a distinctive "memory budget exceeded: used > budget" marker so the two guards are distinguishable.
- Provenance note: q064's native oracle bless was executed by an accidentally-resumed stale agent session and independently re-verified (triple-consistent) before acceptance.
- Repo CI does not trigger on non-main base branches — this record is the verification for the PR; full CI fires when the stack moves toward main.

## Not in this PR (follow-ups, evidence-backed)

- **Pre-merge**: the shortcut-fires integration hermetic (the committed F-16 closure evidence — the corpus structurally cannot exercise `trust_fk_refs`).
- **F-16 (open)**: parent-lookup bounding for ref-bearing crawls on the chat/`query_from` path (bound the lookup to the child-pruned FK set, or a solo-side decision to enable `trust_fk_refs` for chat crawls — a product call on browse semantics).
- **VALUES lowering** (round-3b `#9`, diagnosis complete): VALUES → IN-set scan filter riding the parked ref-value-set backend; the diagnosed query would then fuse end-to-end.
- **Make-it-succeed for the heavy class**: dim-scan semi-join front-end (reusing the parked ref-value-set IN-set→Iceberg-prune backend, 86.6% measured); mixed fact+dim group-key fold extension (live 36M evidence: a 2-hop date rollup grouped by dim-year + fact-shipMethod declines today); LIMIT-through-joins budget forwarding (`HashJoin` inherits the no-op and drains its build fully); bounded/spilling aggregation (large, whole-cloth).
- **Two open diagnostics**: a VALUES-over-graph-source query that hangs in *planning* (183s pre-scan silence); a constant-key point lookup (`?ol ex:orderLineKey "1"`) that failed to bind as a scan constraint and full-scanned 120M+36M rows.
- Single-table flagged rollups (family-B single-table `star_constraints` application); `SELECT DISTINCT` folds; solo-side companions (see `solo-round2-brief.md`, incl. differentiating the RSS-watchdog 507 body).

Base: `feat/external-s3-iceberg`.

## Consolidated scope

The diff of this survivor is exactly #1505 followed by #1514. #1505's pre-consolidation tip is preserved at `archive/pre-refactor-2026-07-21/feat_external-s3-iceberg` and #1514's at `archive/pre-refactor-2026-07-21/feat_lambda-usability`.

| PR | What it did |
|---|---|
| #1505 | BYO-IAM on operator-owned S3: use the REST catalog's inline `loadTable` metadata (no S3 GET for handed bytes), fail-closed vended→ambient credential handling with typed 403s (`err:storage/AccessDenied`, `err:catalog/CredentialsNotVended`), a `ConfigValue::SecretRef` + async hydration path for credential rotation, session-cached S3 clients in Direct mode, and a config-time `verify_storage_access` probe. Closes #1500 / #1497 / #1498. Its review-critical commit reconciles four latent semantic breaks that a clean rebase onto the loadTable-metadata cache would otherwise have shipped silently, by unifying the credential-decision and hydration seams into single call sites. |
| #1514 | The Lambda-usability epic (this PR's tip): the C1–C5 admission/cancellation/parallelism/fusion work, the E1–E2 round-2 fusion admissions, the R3-A/B unified cancellation-plus-typed-memory-budget capacity protection, and the W4-1/W4-1b/W4-2 round-3b pushdown and mixed-key fusion, plus the F-16 scalar-gated `trust_fk_refs` amendment. |

## Verification of record

CI does not fire on a stacked base, so the verification of record is the in-body gate record of the two constituents; both were built and measured against a standing Snowflake + AWS rig.

For #1514: the full 65-member live corpus gated at every wave head with 0 hash mismatches, including the R3 unified-refactor head `33a7d540b` (the separate-mechanism and unified heads gate identically — zero behavioral change), and the wave-4 68-member live SF01 gate reported 0 hash mismatches. The only perf-gate violations in any run fall inside the documented loadTable-GET catalog-latency set (q002/q004/q022/q024/q030/q043), which are non-fused, known-cause, and network-attributed rather than engine regressions. For #1505: an independent adversarial review over the full diff returned SHIP with zero blockers — it recomputed the disk-cache golden hash, byte-diffed the legacy error strings, verified resolved secrets are never persisted / fingerprinted / logged, and confirmed the 401→refresh path is untouched; its one should-fix (a redactor doc-comment) landed in the same branch. The live rig verified the vended and BYO-IAM (zero-identity-permission role, bucket-policy-only access) paths end-to-end, both AWS AccessDenied message variants surviving verbatim to the CLI and the 403 JSON, and the zero-REST composition holding under BYO-IAM.

The serverless constraint envelope this epic targets — the deployed cap, heartbeat, and deadline, and the query-class × wall matrix — is documented in [C1-constraint-envelope.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C1-constraint-envelope.md). The memory-budget blind spot that R3 closes (the scan-path and post-scan buffering that could grow to ~4.4–11 GB unaccounted) is verified independently in [V2-membudget-verification.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/V2-membudget-verification.md), and the per-core throughput ceiling that frames the "capacity protection, not throughput" posture is recomputed in [RT2-redteam-empirics.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/RT2-redteam-empirics.md).

## Native-impact appendix

This survivor carries the program's single most consequential native call-out. It is foregrounded here.

- **R3-A/B typed memory budget — the disclosed native behavior change.** A unified `ctx.checkpoint()` (cancellation check first, then budget) is wired at seven engine-wide buffering sites — the hash-join build and probe, both aggregate folds (`group_aggregate`), and the fused-fold map builds — with a query-scoped memory counter living on the shared cancellation handle, so every operator of a query reaches the same counter (a join build and a downstream fold share one budget). A native query performing a large hash-join or group-aggregate now trips a typed `QueryError::MemoryBudgetExceeded` → **507 (distinct from a 408 timeout)** at roughly 78% of the *detected* container/system memory (cgroup v2 → v1 → `/proc/meminfo`, with an 8 GiB fallback) **instead of OOMing the process**. This is real, intended capacity protection, not a silent change. Cancel-first ordering guarantees an external cancel or deadline can never be masked as a 507. Knobs: `FLUREE_QUERY_MEMORY_BUDGET_BYTES` overrides the budget and `=0` disables it; there is no engine-imposed time limit anywhere — the checkpoints are polls only, and deadlines are armed solely by callers. Verified: the full 65-member live corpus gated 0 hash mismatches at every wave head, including the R3 unified-refactor head `33a7d540b`. Honest scope (kept from the adversarial passes): per-row byte estimates under-count large heap-strings, so "over-count-safe" is scoped to small-binding builds like the deployed DW workload; the counter is query-lifetime-cumulative (also over-count-safe); and an RSS-level watchdog remains the backstop for untracked scan-stage buffers. The full mechanism — the seven buffering sites, the 507 response code distinct from the 408 timeout, the cancel-first ordering, and the knobs — is set out per the original #1514 PR record (https://github.com/fluree/db/pull/1514), in its R3-A/B component-checklist item, and the memory-budget blind spot it closes is verified independently in [V2-membudget-verification.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/V2-membudget-verification.md).
- **C1 `DatasetOperator` LIMIT-budget / top-k forwarding — a second generic-operator budget extension, living here.** `DatasetOperator` is a generic operator on every `FROM <ledger>` query, so this is a distinct `set_row_budget` extension from the F17 UNION/BIND forwarding in #1507 and the OPTIONAL forwarding in #1528. It is inert without a LIMIT budget, and it is switch-gated by `FLUREE_R2RML_DATASET_BUDGET`. It is called out here so the generic-operator budget story is complete across the program, and it is documented per the original #1514 PR record (https://github.com/fluree/db/pull/1514), as component-checklist item C1.
- Everything else in the epic is on the R2RML/Iceberg virtual path: the manifest-backed info-stats routing (C2), mid-scan cancellation (C3), intra-file row-group parallelism (C4), the C5/E1/E2 fused-aggregate admissions, and the W4-1/W4-1b/W4-2 pushdown and fold widenings. The three latent correctness bugs the epic closes are inside the fused-R2RML path. The BYO-IAM work (#1505) is entirely in the Iceberg storage / credential layer.

Confirmed negatives across the span (git-checked): no `arrow` / `datafusion` / `parquet` dependency bump, and no `.github/workflows` or `rust-toolchain` edits.


fluree-db-core/src/cancellation.rs
```diff
@@ -121,6 +148,51 @@ impl QueryCancellation {
...
     }
+
+    /// Record `bytes` of retained query memory into the shared counter. Callers record
+    /// at the points where a retained buffer (a hash-join build table, a GROUP BY map,
+    /// a fused dim-map) grows, so the counter tracks the query's live post-scan memory.
+    /// Monotonic and intentionally conservative — over-counting can only trip the guard
+    /// on a query already near its ceiling. No-op on a disabled handle.
+    #[inline]
+    pub fn record_alloc(&self, bytes: usize) {
+        if let Some(inner) = &self.inner {
```
bplatz, line 164: **Definite issue** — the monotonic counter over-counts operators that are *rebuilt*, not just ones that are sequentially freed.

The doc frames the conservatism as summing sequentially-freed operators, which is the benign case. The one that isn't benign is re-execution: `GraphOperator` builds a fresh inner subplan per parent batch (`graph.rs:357`, `:536`), and the batched-OPTIONAL path rebuilds similarly. If a hash join sits in that inner subplan, every rebuild re-records its full build table while the previous one has already been dropped — so the counter grows linearly in parent batches while actual peak memory stays flat at one build.

With N driving batches the counter reports ~N× the true peak. On a Lambda-sized container (budget = 78% of, say, 1 GiB) a correlated GRAPH query that never exceeds a few hundred MB can trip a typed `MemoryBudgetExceeded` → 507. That's a false abort on a legitimate query, and it lands on native queries too — `hash_join` and `group_aggregate` are shared operators and the guard is default-on.

This is different from the disclosed scoping note. "Cumulative across sequential operators" is bounded by plan size; cumulative across *re-executions* is bounded by data size, which is the thing the budget is supposed to be measuring.

Cheapest fix that keeps the mechanism: have the correlated rebuild sites reset or scope the counter per inner execution, rather than making it fully per-operator. Worth at least confirming whether a rebuild-heavy corpus member exists — I don't think the 68-member gate covers this shape, since it'd show up as a spurious 507 rather than a hash mismatch.

fluree-db-query/src/r2rml/fused_aggregate.rs
```diff
@@ -867,6 +900,72 @@ impl FusedR2rmlAggregateOperator {
...
+    /// per-var source chain-index, or `None` to decline the whole fuse.
+    fn route_group_key_sources(chain: &[&R2rmlPattern], group_by: &[VarId]) -> Option<Vec<usize>> {
+        let last = chain.len().checked_sub(1)?;
+        let mut out = Vec::with_capacity(group_by.len());
+        for gv in group_by {
+            let mut src: Option<usize> = None;
+            for (i, p) in chain.iter().enumerate() {
+                if Self::predicate_for_var(p, *gv).is_some() {
+                    if src.is_some() {
+                        return None; // ≥2 sources: cross-source equality
```
bplatz, line 938: **Definite issue (sequencing)** — this PR widens admission into a fold that still has an open under-count.

The cross-source decline rule here is right, and the W4-2 per-key resolution reads correctly. The problem is what it widens *into*: `insert_dim_gkeys` is unchanged on this branch, and it still treats an equal-value duplicate dim join-key as harmless. It isn't — the map holds one entry and is probed once per fact row, while the reference pipeline emits one row per matching dim row, so `COUNT(*)` comes back short. I've flagged that on #1507.

C5, E1, E2 and W4-2 together admit substantially more shapes to the fused path — shared-predicate group keys, join+flag rollups with constraints applied, mixed fact+dim keys. Each of those is a new shape reaching a fold whose dim-side duplicate handling is known-wrong, so this PR increases the blast radius of that bug rather than being independent of it.

Not asking for a fix here — asking that this land after the `insert_dim_gkeys` fix on #1507, and that the widened shapes get re-checked against a dim with a non-PK join key once it's in. The 68-member SF01 gate can't catch it: unique PKs mean the duplicate case never arises in the corpus, which is exactly why 0 hash mismatches isn't evidence on this point.

fluree-db-query/src/hash_join.rs
```diff
@@ -665,6 +665,16 @@ impl HashJoinOperator {
...
         while let Some(batch) = build.next_batch(ctx).await? {
+            // The build side is drained FULLY here in open() — for a multi-million-row
+            // FACT build that is minutes of accumulation into `table`/`wildcard_rows`
+            // with no poll of its own (the round-3 OOM driver). Account this batch's
+            // contribution to the query-scoped memory counter, then checkpoint: a
+            // deadline abort surfaces a clean `Cancelled`, and an over-budget build —
+            // inherently un-LIMIT-able, every build row is needed for correctness —
+            // surfaces a typed `MemoryBudgetExceeded` BEFORE the runtime OOMs.
+            // Batch-granularity, approximate (rows × cols × est/binding).
+            ctx.record_alloc(batch.len() * ncols * crate::context::BINDING_EST_BYTES);
```
bplatz, line 676: **Worth a look** — `ncols` here is the build *schema* width, so the estimate scales with declared columns rather than retained ones.

`BINDING_EST_BYTES` is a flat 64 regardless of what the binding actually holds, which the body already acknowledges under-counts long IRIs and GROUP_CONCAT state. Combined with `ncols` being schema width, the per-batch estimate can be off in both directions at once on a wide, string-heavy build — under on the strings, over on the narrow columns.

That's fine for the stated purpose (catching a runaway before the OOM), and I wouldn't hold the PR for it. Worth knowing when the 507s start arriving, though: the number in `MemoryBudgetExceeded { used_bytes }` isn't close enough to real bytes to debug against, so it'd help if the error or the span made clear it's an estimate.

bplatz (changes requested):
Two issues inline, one of them a sequencing constraint rather than a defect in this diff.

The engine-wide surface I checked holds up: all 7 `checkpoint()` sites are batch-granular (hot-loop purity respected), `checkpoint()` is cancel-first so an external cancel can't be masked as a 507, `query_memory_budget_bytes` is `OnceLock`-cached so there's no per-call cgroup read, `DatasetOperator` concatenates without dedup (so C1's per-member budget is sound, and the `distinct_above_dataset_absorbs_budget` test pins the boundary), per-member top-k is sound, and W4-1's `coerce_scalar_for_pushdown` gate is genuinely tight — `n.to_string() == *s` rejects `"01"`/`"+1"`/`"1.0"`, out-of-i64 `xsd:integer` declines rather than wrapping, and every decline direction is toward the residual.

fluree-db-core/src/cancellation.rs
aaj3f: Confirmed — and reproduced before fixing: a hermetic test with two sequential inner builds on one cancellation handle showed the second build falsely aborting at the summed estimate against a budget the true peak never approached. Your distinction (cumulative-across-plan is bounded by plan size; cumulative-across-re-executions is bounded by data size) is exactly right and is now quoted in the fix's invariant comments. Fixed in 5b331043b at the rebuild boundary: the counter is snapshotted before each correlated inner build and the delta released after the inner drains, at all three rebuild sites — `execute_in_graph` (per-row), `execute_in_graph_batched`, and the batched-OPTIONAL driver — using the release primitive ported byte-identically from the audit line (which forbids releasing live persistent builds; the snapshot-delta pattern satisfies it by construction). The repro is now a permanent regression test (`d1_rebuild_boundary_release_no_false_abort`, with a negative control). On your corpus question — confirmed, and stronger than framed: no harness sets the budget env at all, so the 68-member gate ran at the ~78%-of-container default and could not have tripped a spurious 507 on any shape; 0 spurious aborts was structurally not evidence.

fluree-db-query/src/r2rml/fused_aggregate.rs
aaj3f: Agreed on the sequencing, now moot in the good direction: the `insert_dim_gkeys` fix landed three rungs earlier on #1507 (0b4269a67) and this branch has merged it up (5b5966446), so the widenings here sit on top of the decline rather than the bug. New regression tests exercise a NON-unique dim join key against the widened admission (`join_fold_declines_on_non_pk_dim_join_key` (control: unique key fuses; non-PK duplicate declines to the generic pipeline)) — with one precision on scope: of the widened shapes, W4-2 mixed fact+dim keys and dim-side E2 lower through the join fold and could reach `insert_dim_gkeys`; E1 shared-predicate, C5, and fact-only E2 lower through the single-scan fold and structurally cannot. The tests target the former; the boundary is now stated in the test module so the distinction survives.

fluree-db-query/src/hash_join.rs
aaj3f: Fixed in 5b331043b at the labeling level you suggested: the abort message now reads "used ~N B (estimated)", and the span key is renamed to `mem_used_bytes_est` (in the process it also got properly declared — it was previously recorded but unregistered). The field name stays for API stability. The underlying estimate remains schema-width × flat per-binding cost as you describe; making it byte-accurate is deliberately out of scope for a guard whose job is catching runaways before the OOM.

aaj3f:
Both definite items and the worth-a-look are addressed: the rebuild over-count is reproduced, fixed at the rebuild boundary with the audit line's release primitive, and pinned by a regression test (5b331043b); the widening now lands after the dup-key decline via the merge-up (5b5966446) with non-unique-key regression tests for the shapes that can actually reach the join fold; and the budget abort is labeled as an estimate. This PR gets no repo CI on its stacked base, so the local gate run is recorded in the PR's verification appendix: core 714/0, query 1340/0, api 2772/0 at 5b331043b, re-run at 44af11841.

bplatz (approved):
All three items resolved at 44af11841, verified locally:

- **D1**: fixed at the rebuild boundary as suggested — `QueryCancellation::release` (saturating, with the persistent-build invariant doc) + snapshot-delta at all three sites (`execute_in_graph` per-row, `execute_in_graph_batched`, batched-OPTIONAL `build_batch`). I re-derived the soundness independently: every `record_alloc` site in the crate lives inside a correlated inner tree (hash-join build, GROUP BY maps, fused-dim maps) and the outer drain structures never charge, so the released delta is exactly the dropped inner's transient charge — and the pattern composes for nested rebuilds. Early `?`-exit skips the release, which only over-counts on an aborting path. `cancellation.rs` diffs empty against the audit line. The regression test's negative control proves the un-released sum would have tripped the pinned budget, so it can't pass vacuously. Also confirmed the corpus point: no harness sets the budget env, so the 68-member gate structurally couldn't surface this.
- **D2**: satisfied structurally — base is the f17 branch and 5b5966446 brings in the `insert_dim_gkeys` decline, so the widenings cannot land ahead of it. `join_fold_declines_on_non_pk_dim_join_key` covers the re-check with a unique-key control that fuses, and the stated join-fold/single-scan boundary (W4-2 + dim-side E2 reach `insert_dim_gkeys`; E1/C5/fact-only E2 cannot) matches the code.
- **W**: "(estimated)" in the Display, field doc'd as approximate with the name kept stable, span key renamed `mem_used_bytes_est` — and properly declared, fixing that the record was previously a silent no-op.

Re-verified: `cargo test -p fluree-db-query` → 1293 passed; core cancellation tests incl. the two new release tests pass; D1/D2 regression tests pass individually. Stacked base gets no repo CI, consistent with the verification appendix.

PR #1516, opened 2026-07-17
feat(cli): fluree model — governance tooling for access, entities, and class hierarchy

bplatz:
## What

Adds a `fluree model` command family: governance tooling that compiles declared intent into ordinary JSON-LD transactions against the target ledger. No new server API — the commands work identically against local ledgers, `fluree-db-server`, and hosted stacks. Three facets:

- **`model access enable|show`** — a stateless policy scaffolder. Profiles `read` / `write` / `intake` compile into a policy class (the assignment unit grants and tokens carry) plus one or two thin policies built on the engine's write verbs (`f:create`/`f:update`/`f:delete`). Options: `--class` (the governed class → `f:onClass`), `--property` (column-narrowed writes via `f:onProperty`), `--connected <path>` (relationship-gated read; the SPARQL property path is stored verbatim as an `@path` context term inside `f:query`), `--policy-class` (override the assignment-unit IRI), `--space` + `--remote` (attach the policy class to a hosted stack's space grant, merging never clobbering), `--dry-run`.
- **`model entity define|show`** — entity definitions as SHACL node shapes, with a property spec mini-language (`"<iri> string required in[a,b]"`) and `--closed` (with the `rdf:type` carve-out). Defining a shape activates transaction-time validation for its class.
- **`model class define|show`** — `rdfs:Class` / `rdfs:subClassOf` hierarchy; with entailment enabled the hierarchy flows through both query and policy, which is why it lives under governance tooling.

## Design notes

- **Stateless by design**: there is no stored intent language and nothing to sync. The compiled artifacts are small and self-describing; hand-editing a policy is legitimate authorship, not drift.
- **Atomic replace for owned nodes**: re-running `enable` installs via ONE transaction that wildcard-deletes the two owned policy ids (`{policy-class}/view`, `{policy-class}/write`) and inserts the fresh compilation. Merge-upsert is unsafe here: the policy loader gives `f:allow` precedence over `f:query`, so a stale `f:allow: true` would silently disable a newly added `--connected` gate. Owning both ids on every run also makes profile switches exact (write → read revokes the write policy). `entity define` gets the same treatment for the shape + property-shape children (a dropped `sh:closed`/`sh:minCount` doesn't survive a re-run); the entity class node itself stays additive since it's shared authorship with `model class define`.
- **Naming**: the access command's target flag is `--class` (it compiles to `f:onClass`, and `model class define --class` takes the same IRI for the same concept); `--policy-class` names the assignment unit by its role.
- **Verb exactness**: class targeting matches pre ∪ post state and `rdf:type` writes match the class they mint, so a Lead grant can never create a Contract — no property allow-list needed. The property surface of a class belongs to its SHACL shape.

## Tests

- Unit tests for all three compilers plus the replace-transaction builder (21 in `fluree-db-cli`).
- New engine integration tests pinning the contracts the scaffolder depends on:
  - `scaffolder_write_profile_grants_class_ownership_only` — the exact emitted shape, stored and selected via `policy-class`, grants class ownership and nothing else;
  - `policy_fquery_with_path_context_term` — an `@path` context term inside a stored `f:query` survives verbatim and enforces the path;
  - `policy_replace_swaps_allow_for_path_gate` — the exact replace transaction the CLI emits actually narrows a previously allow-all read profile to a relationship gate (the failure mode the replace semantics exist to prevent).

## Docs

`docs/cli/model.md` (full command reference incl. re-run semantics and the three class-shaped roles), indexed in `docs/SUMMARY.md` and the CLI command table; picked up automatically by the embedded `fluree docs` corpus.


fluree-db-cli/src/commands/model/entity.rs
```diff
@@ -0,0 +1,440 @@
...
+            }
+        }
+    }
+    Ok(out)
+}
+
+/// Last path segment of an IRI, for stable child-node ids.
+fn local_name(iri: &str) -> &str {
+    iri.rsplit(['/', '#', ':']).next().unwrap_or(iri)
+}
```
aaj3f, line 125: [as well as `:113` (child id), and `:251-254` (owned set)]

I think there's a silent validation-correctness bug. Child property-shape ids are minted as `{shape}/{local_name(path)}` where `local_name` is the last `/`,`#`,`:` segment of the property IRI, with no dedup/collision check. Two property specs whose IRIs share a final segment collapse to the same `@id`, and JSON-LD unions them into one node. 

Traced through the real SHACL compiler: `sh:path` is last-write-wins (`fluree-db-shacl/src/compile.rs:648`) so **one property's path is silently dropped from validation**, while `sh:datatype` accumulates (`compile.rs:667`) so the survivor gets two datatypes → an unsatisfiable shape. 

Concrete repro: 

```
model entity define Lead \
  --property "https://schema.org/name string required" \
  --property "https://example.org/name integer"
```

It produces one merged `Lead/shape/name`; one `name` is unenforced and the survivor requires every value be **both** `xsd:string` and `xsd:integer`, so every instance is rejected — with **no CLI error**. Common vocabularies collide on the final segment (`name`, `email`, `id`, `status`, `title` across schema.org/foaf/custom), so this is reachable in ordinary multi-namespace models. 

Fix is a one-liner: detect a duplicate child id across specs and return `CliError::Usage`. (A stricter reviewer could hold this as Request-changes; I'm leaving the overall verdict at Approve-with-nits because the trigger is narrow and the fix is trivial — but it should land before merge.)

fluree-db-cli/src/commands/model/entity.rs
```diff
@@ -0,0 +1,440 @@
...
+            }
+            if p.iri_kind {
+                node[format!("{SH}nodeKind")] = json!({"@id": format!("{SH}IRI")});
+            }
+            if p.required {
+                node[format!("{SH}minCount")] = json!(1);
+            }
+            if !p.values_in.is_empty() {
+                node[format!("{SH}in")] = json!({"@list": p.values_in});
+            }
```
aaj3f, line 196: a spec combining `iri` with `in[...]` (or with a datatype token) emits contradictory constraints: `sh:in` values are serialized as string **literals** while `sh:nodeKind sh:IRI` requires an IRI value, so nothing validates; likewise `iri` + a datatype emits both `sh:datatype` and `sh:nodeKind`. 

`parse_property_spec` also accepts two datatype tokens (last wins) and silently drops empty values in `in[a,,b]`, and whitespace inside `in[a, b]` fails with a cryptic `unknown token 'in[a,'`. 

Consider rejecting the mutually-exclusive combinations and tightening the `in[...]` tokenizer.

fluree-db-cli/src/commands/model/entity.rs
```diff
@@ -0,0 +1,440 @@
...
+    // `model class define` (hierarchy, label), so it stays additive.
+    let mode = resolve_mode(dataset, remote, dirs, direct).await?;
+    let shape_iri = format!("{}/shape", entity.trim_end_matches('/'));
+    let mut owned: Vec<String> = specs
+        .iter()
+        .map(|p| format!("{shape_iri}/{}", local_name(&p.path)))
+        .collect();
+    owned.insert(0, shape_iri);
+    let owned_refs: Vec<&str> = owned.iter().map(String::as_str).collect();
+    replace_nodes(&mode, &graph, &owned_refs).await?;
```
aaj3f: Very minor but worth considering that there are scenarios that may leave orphaned child shapes on property removal. The owned/replace set is derived from the **current** specs only, so re-running `define` with a property dropped never deletes that property's old child shape. 

Verified against the compiler that this does not break validation, as orphan children carry no `@type sh:NodeShape` and aren't reachable from the (fully-replaced) parent's `sh:property`. So, the dropped constraint does stop firing and closure narrows correctly, but dangling triples (and the `sh:in`/`sh:ignoredProperties` RDF-list blank nodes, which a wildcard-delete-by-subject doesn't cascade to) accumulate on every drop/rename. 

The `:243-248` comment ("property-shape children are OWNED … replaced atomically") over-claims for the removed-child case. If we want a true replace, query the shape's existing `sh:property` child ids and union them into the owned set; otherwise we can soften the comment.

fluree-db-cli/src/commands/model/class.rs
```diff
@@ -0,0 +1,274 @@
...
+        node[RDFS_LABEL] = json!(l);
+    }
+    if !subclass_of.is_empty() {
+        node[RDFS_SUBCLASS_OF] = json!(subclass_of
+            .iter()
+            .map(|p| json!({"@id": p}))
+            .collect::<Vec<_>>());
+    }
+    node
+}
```
aaj3f, line 145: [as well as `mod.rs:69-86` `upsert`]

`class define` cannot clear the previous `subClassOf` parent (or a label) because `compile` omits `rdfs:subClassOf` when the parent list is empty, and `upsert` only replaces *listed* properties, so there is no way to remove the final parent (no `--no-subclass`). 

This is a footgun precisely because, with RDFS entailment, a stale class parent widens every grant on that parent (`class.rs:70-75`). 

It's semi-surfaced by the printed "replaces the parent set" message but can't actually be executed to empty.

fluree-db-cli/src/commands/model/entity.rs
aaj3f, line 1: Relevant to lines 101-107, 149, 251-256

`entity define --label` is non-idempotent: the class node is inserted-but-not-owned (additive), so re-running with a changed `--label` adds a second `rdfs:label` rather than replacing it (unlike `class define`, which upserts). 

Minor, but a user attempting to re-label would leave two labels on the class.

fluree-db-cli/src/commands/model/entity.rs
```diff
@@ -0,0 +1,440 @@
...
+    let mut tokens = spec.split_whitespace();
+    let path = tokens
+        .next()
+        .ok_or_else(|| CliError::Usage("empty --property spec".into()))?
+        .to_string();
+    if !(path.starts_with("http://") || path.starts_with("https://") || path.starts_with("urn:")) {
+        return Err(CliError::Usage(format!(
+            "--property must start with an absolute IRI (got '{path}')"
+        )));
+    }
```
aaj3f: The absolute-IRI check is duplicated three times. Here, at `:194-201` and in `class.rs:173-181`

fluree-db-api/tests/it_policy_verbs.rs
```diff
@@ -729,3 +729,126 @@ async fn post_state_condition_sees_staged_flakes() {
...
+                "f:action": [
+                    {"@id": "f:create"},
+                    {"@id": "f:update"},
+                    {"@id": "f:delete"}
+                ],
+                "f:onClass": {"@id": "http://example.org/ns/Lead"},
+                "f:allow": true
+            }
+        ]
+    });
```
aaj3f: `scaffolder_write_profile_grants_class_ownership_only` hand-copies the byte-exact compiled shape rather than calling `access::compile()` (understandable — importing the CLI compiler into a `fluree-db-api` test would invert the crate dependency). A one-line comment cross-referencing `access::tests::write_profile_emits_class_view_and_verb_policy` would flag the coupling so a future `compile()` change doesn't drift from this copy.

aaj3f (approved):
A few inline notes and then just some test gaps is you want to cover them: 
- no test for the child-id collision (the should-fix bug above)
- none for the drop-a-property owned-set boundary
- none for the contradictory-token / `in[a,,b]` parser cases
- `class.rs` has no test for the can't-clear-last-parent limitation or `show`'s policy-class exclusion

fluree-db-cli/src/commands/model/entity.rs
bplatz: Addressed in 8c1174fd97a2cab97d378e5439778619a01a3217 — duplicate child ids are now rejected up front with a `CliError::Usage` naming both colliding IRIs and the shape id.

fluree-db-cli/src/commands/model/entity.rs
bplatz: Addressed in 8c1174fd97a2cab97d378e5439778619a01a3217 — `iri`+datatype, a second datatype, and `iri`+`in[...]` are now rejected; empty values in `in[a,,b]` error instead of being dropped. For `in[a, b]` I kept the whitespace tokenizer but the error now says spaces aren't allowed inside `in[...]` rather than the cryptic unknown-token message.

fluree-db-cli/src/commands/model/entity.rs
bplatz: Addressed in 8c1174fd97a2cab97d378e5439778619a01a3217 — went with the true-replace option: `define` now queries the shape's existing `sh:property` children and unions them into the owned set, so a dropped property's child shape is deleted. The RDF-list blank nodes behind `sh:in`/`sh:ignoredProperties` are the one residue a wildcard delete-by-subject can't cascade to; the ownership comment now says so.

fluree-db-cli/src/commands/model/class.rs
bplatz: Addressed in 8c1174fd97a2cab97d378e5439778619a01a3217 — added `--clear-subclass-of` (clap-enforced conflict with `--subclass-of`), which atomically deletes every `rdfs:subClassOf` edge while writing the node. Kept empty `--subclass-of` meaning "leave parents alone" so a label-only re-run can't silently wipe the hierarchy.

fluree-db-cli/src/commands/model/entity.rs
bplatz: Addressed in 8c1174fd97a2cab97d378e5439778619a01a3217 — when `--label` is given, the transaction now deletes the class node's existing `rdfs:label` in the same atomic update (targeted property replace); the node otherwise stays additive.

fluree-db-cli/src/commands/model/entity.rs
bplatz: Addressed in 8c1174fd97a2cab97d378e5439778619a01a3217 — hoisted to `mod.rs::require_absolute_iri`; there was a fourth copy in `access.rs` that's deduped too.

fluree-db-api/tests/it_policy_verbs.rs
bplatz: Addressed in 8c1174fd97a2cab97d378e5439778619a01a3217

bplatz:
Test gaps from the review are covered in 8c1174fd97a2cab97d378e5439778619a01a3217: child-id collision rejection, drop-a-property owned-set boundary, the contradictory-token and `in[a,,b]`/`in[a, b]` parser cases, the clear-parents transaction (which replaces the can't-clear-last-parent limitation), and `show`'s policy-class exclusion.

PR #1518, opened 2026-07-18
fix: Cypher field-report round — reindex OOM, EXPLAIN, =~, IRI round-tripping, HTTP bulk import, arena sealing

bplatz:
Addresses the full KB-on-FlureeDB field report (Cypher/Bolt integration): the reindex-query OOM, interactive-latency tooling, surface gaps, the silent write failure, and the bulk-import sidecar.

## Performance: reindex-query OOM / 35s (the headline)

`UNWIND $uris MATCH (s:INDEXED {value: uri})-[p]->(o:INDEXED)` against ~72k nodes / ~190k reified edges pegged CPU and OOM-killed a 16 GB server; LIMIT 20 still took 35s.

Root cause: a value-only relationship binding lowers to `OPTIONAL { EdgeAnnotation }`; on reified ledgers without a sealed annotation arena, the expanded chain evaluated per row and drove from `f:reifiesPredicate` — enumerating ~(sidecar / #relationship-types) candidate reifiers per row and existence-checking each with its own scan (~25ms/row).

- **Batched annotation OPTIONAL probe**: answers each required batch with three set-wise `f:reifies*` planned scans (overlay-merged, policy-filtered) hash-indexed once per query run, then per-row lookups. Falls back to the per-row path for history/dataset/multi-ledger contexts.
- Verified at the reported scale: full 54,600-row result **35s → 1.1s**, LIMIT 20 **0.9s**, flat memory.
- Supporting: the OPTIONAL hash-join lane admits single-source `DefaultGraphSource` chains; the forward-arena probe accepts per-row (variable) predicates; structural selectivities for `f:reifies*` bound-object lookups; a reusable `cypher_unwind_probe` example pins the shape.

## Annotation arena sealing (write-only ingest trap)

Annotated insert + index left `annotation_index = None` permanently — three cooperating defects:
1. The attachment provider returned "delta unknown" when the ledger wasn't resident in the `LedgerManager` (the norm for write-only ingest); the build defensively dropped the arena and stamped the sticky bit, blocking the bootstrap scan forever. Fixed with a **transient** (never-cached) backend load whose commit replay yields the complete event history.
2. A full rebuild under `Augment` coverage discarded a previously-sealed arena; root assembly now receives the previous root's CID and merges the prior arena's events with the delta (the incremental Phase 3d contract).
3. `Fluree::reindex` read the NsRecord before quiescing the background indexer, so a racing publish left it rebuilding from a stale record. Re-fetched after cancel + wait-for-idle.

Pinned by a four-cell seal matrix including a background-race reproduction; the deliberate post-defensive-drop scan-fallback contract is unchanged.

## Server-side bulk import over HTTP (drops the customer's sidecar)

The negotiated `/import-upload` handshake now accepts raw source data — the same formats as `fluree create --from` (TTL/NT/NQ/TriG/JSON-LD/JSONL±gz/zst, plus CSV and Cypher dumps converted server-side via shared `fluree-db-api::import_source` helpers the CLI also uses). Discovery advertises `source-upload` + formats; refused on Raft-replicated servers (the pipeline publishes nameservice heads outside the replicated log). `fluree create <ledger> --remote <name> --from data.ttl|dump.cypher` rides mint→PUT→complete→poll. Imports with annotations get a follow-up reindex so the arena is sealed on arrival.

Also surfaced and fixed: the import commit paths held tracing `EnteredSpan` guards across store awaits, making the pipeline future `!Send`.

## Cypher surface

- **EXPLAIN**: `POST /explain/{ledger}` with `application/cypher` (raw or `{cypher, params}` envelope; `$param` substitution matches execution), plus `fluree query --cypher --explain` locally and over `--remote`. Verified live end-to-end.
- **`=~` regex operator**: whole-string Neo4j semantics (non-capturing-group anchoring so alternations and runtime patterns behave; inline flags pass through), wired in read filters and write-path WHERE.
- **Round-trippable names**: `labels()`/`type()`/`keys()` and rendered node/relationship names now follow the `db.labels()` rule — strip the configured `@vocab`, otherwise return the full IRI. A relationship typed `http://…/kb#order` no longer comes back as `"order"`.

## Correctness: silent no-op on wrong Content-Type

A Cypher envelope posted as `application/json` fell through to the JSON-LD update parser, produced a zero-clause transaction, and returned HTTP 200 with a real commit envelope. Updates with neither `insert` nor `delete` are now rejected (422), and a misrouted `{"cypher": …}` body gets a 400 pointing at `Content-Type: application/cypher`.

## Docs

`docs/api/endpoints.md` (Cypher explain, source uploads), `docs/query/cypher.md` (explain, `=~`, naming rule), `docs/cli/server-integration.md` (remote Cypher explain + source-upload contracts), `docs/cli/create.md`.

## Not addressed (deliberate)

- Multi-clause MERGE composition — real architectural constraint (one snapshot horizon per transaction); sequential statements remain the documented workaround.
- The reported ~18s topic-page latency — needs field profiling; the relationship-binding fix plus arena sealing likely covers most of it, and Cypher EXPLAIN now exists to attribute the rest.


anh-h (approved):
This addresses everything that was really problematic for me, and also saves me from testing a couple of KB-side workarounds I had half-drafted for the relationship-binding issue. Thanks for turning this around so fast.

PR #1520, opened 2026-07-18
fix(iceberg): fail closed on merge-on-read delete files (silent stale-row risk)

aaj3f:
fix(iceberg): fail closed on merge-on-read delete files (silent stale-row risk)

This is one of five surviving pull requests in the consolidated virtual-dataset review program, which collapses the #1475–#1529 lineage into five reviewable units at the maintainers' request. #1520 is the standalone correctness fix and the only member based on `main`, so it receives full CI on its own branch; it is meant to land first, ahead of the four stacked survivors. Reading order across the program: **#1520 (this PR)** → #1507 (R2RML/Iceberg virtual-dataset performance program) → #1514 (serverless Lambda-usability + BYO-IAM) → #1528 (big-Iceberg-audit implementation) → #1529 (native-twin materialize builder). The durable, outward-facing audit record these PRs draw on is published on the `docs/virtual-dataset-audit-2026-07` branch — for this PR start with [00-MASTER-AUDIT.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/00-MASTER-AUDIT.md) §2 (finding F-AUD-1) and the independent verification in [V1-mor-verification.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/V1-mor-verification.md).

## Forest map

This is **PR-SAFE-MOR**, the first (and only main-based) implementation PR of the big Iceberg audit. It closes audit finding **F-AUD-1** (silent merge-on-read staleness).

| This PR | Finding | Tier | Where it sits in the program |
| --- | --- | --- | --- |
| Fail-closed MoR delete-file guard | [F-AUD-1](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/00-MASTER-AUDIT.md) | Tier-0 (ship regardless) | Base = `main` (portable, gets real CI). The other Tier-0/1/2 items are branch-coupled and stack on the `perf/audit-tier012` integration branch (see #1528). |

Audit references: [00-MASTER-AUDIT.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/00-MASTER-AUDIT.md) (§2 F-AUD-1, §6 Tier-0) · [V1-mor-verification.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/V1-mor-verification.md) (the verified gap + fix sites).

## The bug

Fluree's virtual Iceberg read path **recognizes** merge-on-read (MoR) position/equality delete files (`content=1` manifests) but never **applies** them: delete manifests are dropped with a `debug!` log and only the live data files are read (`manifest_list.rs`, `planner.rs`, and no delete reader anywhere in `io/`). Consequences, both silent:

- A query over a MoR-maintained table **returns deleted rows as if live**.
- The manifest `COUNT(*)` / row-total shortcut **over-counts** by exactly the delete cardinality (`stats.rs` sums `record_count`, never subtracting deletes).

Copy-on-write deletes are **not** affected — status=`DELETED` manifest entries are already dropped correctly, so CoW scans exactly ADDED+EXISTING files. The gap is strictly MoR delete *files*.

## Exposure (why this matters now)

- **Snowflake-managed v2 (what the smoke datasets are today): SAFE.** V1 measured the deployed `fl-svl-iceberg-smoke-use2` tables as format-v2, append-only, `total-delete-files/position/equality = 0` — Snowflake's current copy-on-write default. The bug is **not** triggered by today's data.
- **External writers: MoR now.** Athena `DELETE` writes position delete files by default; Flink/CDC upserts write equality deletes by design; Spark is one table-property away. The BYO-IAM segment (#1505) is the highest-exposure surface.
- **Snowflake, imminent.** `ICEBERG_MERGE_ON_READ_BEHAVIOR=AUTO` already means MoR for v3 (deletion vectors) and for **all** externally-managed tables. A pending Snowflake BCR bundle (`2026_03` / bcr-2279) would flip SF-managed **v2** to MoR positional deletes by default. **The `2026_03` date is WEB-sourced and unverified** — verify the BCR before scheduling against it; the *gap* is V1-verified, the *date* is not.

Severity: **latent-but-certain** — no delete files in today's smoke data, but MoR arrives with certainty down at least four independent paths. The failure mode (silent wrong answers, no error/log/metric) is the worst class for a database, which is why this ships fail-closed regardless of the trigger date.

## The guard (fail-closed, with an escape hatch)

Refuse — never silently mis-answer — any Iceberg read whose snapshot carries delete files, until MoR application is implemented. Two independent signals so a snapshot cannot slip through:

1. **Snapshot summary counters** (`total-delete-files` / `total-position-deletes` / `total-equality-deletes`) — a zero extra-I/O check; the summary is already in memory. New `Snapshot` accessors mirror `deleted_records()`.
2. **Manifest list** — a `content=1` delete manifest present even when the summary omits/under-counts it. The scan planners now parse the manifest list *with* deletes and count them (the list is read anyway).

Both scan planners (`ScanPlanner` + the Send-safe `SendScanPlanner` used by the server) and the stats/COUNT preview consumer (`iceberg_catalog.rs`) inherit the guard, so scan results **and** row/COUNT totals are covered. On refusal the error is typed (`IcebergError::MergeOnReadDeletes`) and actionable — it names the table, the counters/evidence, why (deletes not applied → deleted rows / over-count), and the override switch.

### Switch | test | behavior ledger

| Switch | Default | Effect | Coverage |
| --- | --- | --- | --- |
| `FLUREE_ICEBERG_ALLOW_MOR_DELETES` | **off** (guard active) | Truthy (`1`/`true`/`yes`/`on`) disables the guard: delete files ignored, read proceeds, **results may include deleted rows / over-count**, warns once per table. | `mor_guard` unit tests (both arms) + planner + stats fixture tests |

Field revert is per-mechanism via the switch (no redeploy needed). Docs: added to `docs/graph-sources/iceberg.md` (Limitations + a new "Environment switches" table). There is no `SWITCHES.md` on `main`; the harness PR (PR-HARNESS) folds this row into the regenerated registry.

## Tests (the fixture that did not exist before)

V1 flagged that **no MoR delete fixtures existed**, so CI proved nothing about this class. This PR adds real `content=1` delete fixtures (in-code Avro) that drive the guard end-to-end:

- **Snapshot accessors** — present/absent counters (`snapshot.rs`).
- **Guard logic** (`mor_guard.rs`) — position>0 refuses; equality>0 refuses; zero/absent proceeds; override proceeds; error message names the table + switch; pure truthy-parsing.
- **Planner, real Avro** (`planner.rs`) — a manifest list with a `content=1` entry drives `plan_scan()` to `MergeOnReadDeletes`; a summary-counter snapshot refuses before any I/O; a delete-free snapshot plans normally.
- **Stats/COUNT path** (`stats.rs`) — `read_snapshot_data_files` detects deletes, then the guard refuses (default) / proceeds (override), mirroring the `iceberg_catalog.rs` consumer exactly.

The override "skipped under override" arm is proven at the guard-function level rather than through the planner, to avoid mutating the shared process env in a parallel test run (documented in-code).

## Verification record

All run locally in this branch's worktree (base = `main` @ `7581f0ac8`). **CI runs live on this PR** (base=main fires `fmt`/`clippy`/`test`/`testsuite-sparql`, even while draft).

| Gate | Command | Result |
| --- | --- | --- |
| fmt | `cargo fmt -p fluree-db-iceberg -p fluree-db-api -- --check` | clean |
| clippy (scoped, `-D warnings`) | `cargo clippy -p fluree-db-iceberg --all-features --all-targets --no-deps -- -D warnings` | clean (no pre-existing lints in this crate) |
| iceberg tests | `cargo test -p fluree-db-iceberg --all-features` | **200 passed**, 0 failed (was 188; +12 new) |
| api tests (compile surface) | `cargo test -p fluree-db-api --features iceberg` | all bins pass, 0 failed (102 ignored = live-Snowflake) |
| featureless / default build | `cargo build -p fluree-db-iceberg --no-default-features` + default | both clean |
| api clippy (edited file) | `cargo clippy -p fluree-db-api --features iceberg --no-deps` | no findings in `iceberg_catalog.rs` |

Perf arm: not applicable (this is a correctness guard, not an optimization) — no vbench claim.

## Update — review follow-ups (commit `c5f7feaff`)

Two review follow-ups, behavior-preserving:

- **Shared planner guard helpers.** The identical guard block in `ScanPlanner` and `SendScanPlanner` is extracted into two `mor_guard` helpers — `ensure_summary_scannable` (zero-I/O summary check before the manifest-list read, returns the override flag) and `ensure_manifests_scannable` (the post-read belt-and-suspenders) — so the fail-closed check cannot drift between the two planners. Same checks, same order, same override.
- **Production COUNT-path refusal test.** A COUNT / `table_row_count` query routes through the production `SendScanPlanner`; #1520 originally tested only the runtime-agnostic `ScanPlanner`. Added a hermetic test proving `SendScanPlanner::plan_scan()` fails closed on a delete-bearing snapshot (`MergeOnReadDeletes`), so the guard is not merely transitive. Both new helpers are also unit-tested directly.

Gates re-run at this head: fmt clean; `clippy -p fluree-db-iceberg -D warnings` clean; iceberg 203 tests pass; `cargo test -p fluree-db-api --features iceberg` 0 failures. CI re-fires on this push (base=main).

Note (residual, out of scope for this PR): the `/info` `table_row_counts` display (`ledger_info.rs::fetch_virtual_table_row_counts`) derives counts from the snapshot summary `total-records`, which over-counts a merge-on-read table. It is a metadata display, not a query answer, and is not guarded here — flagged for the audit follow-up if a guarded `/info` count is wanted.

## Consolidated scope

#1520 folds in no other pull requests — it is the standalone Tier-0 correctness fix and stands exactly as scoped. Its two commits are the guard (`771f6e88b`) and the review-follow-up test plus shared planner helpers (`c5f7feaff`). The pre-consolidation tip is preserved at tag `archive/pre-refactor-2026-07-21/fix_iceberg-mor-delete-guard`; nothing was rewritten.

## Verification of record

CI fires on this PR because its base is `main`, so `fmt`/`clippy`/`test`/`testsuite-sparql` run against the change directly — it is the one survivor with real CI on its own diff, and the in-body gate table records the local reproduction.

The independent adversarial diff review is recorded under R-1520 in [pr-reviews-impl.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/pr-reviews-impl.md): verdict SHIP on main — the guard is complete there (both live planners plus the stats/preview consumer guarded, the in-memory scan-files cache safe by construction, and the branch-only disk cache does not exist on `main`). The same review flagged that the guard was incomplete on the stacked perf line, where a cached scan-files hit-arm rebuilds scan tasks around `plan_scan`; that gap was closed on the audit line and is recorded under R-CACHE-ARM in the same file (verdict CLOSED — delete-flag on the cached entry, re-check on both hit arms, and a `CACHE_FORMAT_VERSION` bump so pre-guard entries miss and recompute through the guard). That fix travels into the program via #1528, so the guard is complete on both the main-based and stacked paths.

The finding itself and its fix sites are verified independently in [V1-mor-verification.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/V1-mor-verification.md) with live AWS metadata measurement, and the exposure timeline is summarized in [00-MASTER-AUDIT.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/00-MASTER-AUDIT.md) §2.

## Native-impact appendix

**Zero native surface.** Every edit is in `fluree-db-iceberg` (`planner.rs`, `send_planner.rs`, `stats.rs`, `mor_guard.rs`, `snapshot.rs`) plus `fluree-db-api/iceberg_catalog.rs`, and each is reached only through the Iceberg scan / stats path — never through a native SPARQL query. The one switch, `FLUREE_ICEBERG_ALLOW_MOR_DELETES`, defaults off (guard active) and is an inverted-polarity escape hatch. Coverage is 203 iceberg tests including real `content=1` Avro delete fixtures. There is nothing on this PR for a native-pathway reviewer to check.

## Landing and hand-off notes

Land #1520 first. It shares its fail-closed guard content with #1528's line (the same guard was re-applied on the audit integration branch). The two guard commits carry different git patch-ids only because they were cut against different base trees, but their added and removed lines are byte-identical, so once #1520 merges to `main`, #1528's guard portion resolves as an identical-content merge and the audit line's cache-arm guard (`CACHE_FORMAT_VERSION` 2→3) rides on top cleanly.

The `/info` `table_row_counts` over-count noted above as out of scope is picked up downstream: #1528 carries the rider that flags delete-bearing tables into an `mor-approximate-tables` list (upper-bound counts), so this PR's deferral is closed there rather than left open.

### DIFF
--- a/fluree-db-api/src/graph_source/iceberg_catalog.rs
+++ b/fluree-db-api/src/graph_source/iceberg_catalog.rs
@@ -242,6 +242,13 @@ pub(crate) const DISTINCT_COUNT_WARNING: &str =
     "distinct_count (NDV) is unavailable from metadata alone; it requires column \
      profiling and is deferred to PR-5.";
 
+/// Caveat pushed onto a preview that proceeded past the fail-closed MoR guard
+/// under the `FLUREE_ICEBERG_ALLOW_MOR_DELETES` override: its counts include
+/// deleted rows and are therefore upper bounds. Shared by both preview tiers.
+const MOR_UPPER_BOUND_WARNING: &str =
+    "Table has merge-on-read delete files; row/value/null counts are upper bounds \
+     (position/equality deletes are not subtracted).";
+
 /// Which statistics tier a preview should compute.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "lowercase")]
@@ -584,6 +591,42 @@ pub(crate) fn table_schema_from_metadata(
     })
 }
 
+/// Run the zero-I/O snapshot-summary MoR guard (audit F-AUD-1) that BOTH preview
+/// tiers share, BEFORE any summary-derived count is surfaced. The Schema tier
+/// fills `row_count` from `total-records` and the Stats tier aggregates over the
+/// live data files; neither subtracts merge-on-read deletes, so both over-count a
+/// MoR snapshot. Mirrors the two scan planners (`ensure_summary_scannable`), so
+/// the preview can no longer return silent over-counts the planners would refuse.
+///
+/// Returns the operator override flag (`FLUREE_ICEBERG_ALLOW_MOR_DELETES`) so the
+/// Stats-tier manifest-list backstop reuses it and the env is read exactly once.
+/// `table_display` is the W2 "qualified (location)" identifier. Extracted so the
+/// two-tier guard is unit-testable without a live catalog.
+fn preview_summary_guard(metadata: &TableMetadata, table_display: &str) -> Result<bool> {
+    match metadata.current_snapshot() {
+        Some(snapshot) => Ok(fluree_db_iceberg::ensure_summary_scannable(
+            snapshot,
+            table_display,
+        )?),
+        // No current snapshot ⇒ no counts to over-count; still surface the
+        // override flag for the (delete-free) Stats manifest path.
+        None => Ok(fluree_db_iceberg::mor_deletes_allowed()),
+    }
+}
+
+/// Whether a proceeded preview must caveat that its counts are upper bounds: true
+/// iff the read proceeded under the override AND either delete signal fired — the
+/// manifest-list (`has_delete_files`) or the snapshot summary. The summary arm
+/// covers a delete manifest that mis-parsed as data (F-AUD-1 sub-claim (a)), the
+/// gap that made the preview's two "signals" effectively one.
+fn preview_counts_are_upper_bounds(
+    allow_mor: bool,
+    has_delete_files: bool,
+    summary_deletes: bool,
+) -> bool {
+    allow_mor && (has_delete_files || summary_deletes)
+}
+
 /// Preview an Iceberg table's schema (Tier-A) and, at `tier = Stats`, its
 /// per-column statistics (Tier-B).
 ///
@@ -613,19 +656,41 @@ pub async fn preview_iceberg_table(
         ))
     })?;
 
+    // W2: name BOTH the catalog-qualified table and its storage location, so an
+    // operator grepping logs finds the same identifier the scan planners emit
+    // (they pass the location) plus the friendlier qualified name.
+    let table_display = format!("{} ({})", table.qualified(), metadata.location);
+
+    // D1 (audit F-AUD-1): run the summary MoR guard for EVERY tier before any
+    // summary-derived count is surfaced. Returns the override flag reused by the
+    // Stats-tier manifest backstop below.
+    let allow_mor = preview_summary_guard(metadata, &table_display)?;
+
     let mut schema = table_schema_from_metadata(&table, metadata)?;
 
     match tier {
-        StatsTier::Schema => Ok(TablePreview {
-            schema,
-            stats_completeness: StatsCompleteness {
-                tier: "schema".to_string(),
-                manifests_read: 0,
-                had_column_bounds: false,
-                has_delete_files: false,
-            },
-            warnings: Vec::new(),
-        }),
+        StatsTier::Schema => {
+            // Schema tier reads no manifests (`has_delete_files` stays false), but
+            // its `row_count` comes from `total-records`, which over-counts a MoR
+            // snapshot. Under the override the guard proceeded, so caveat the count.
+            let mut warnings = Vec::new();
+            let summary_deletes = metadata
+                .current_snapshot()
+                .is_some_and(fluree_db_iceberg::summary_indicates_deletes);
+            if preview_counts_are_upper_bounds(allow_mor, false, summary_deletes) {
+                warnings.push(MOR_UPPER_BOUND_WARNING.to_string());
+            }
+            Ok(TablePreview {
+                schema,
+                stats_completeness: StatsCompleteness {
+                    tier: "schema".to_string(),
+                    manifests_read: 0,
+                    had_column_bounds: false,
+                    has_delete_files: false,
+                },
+                warnings,
+            })
+        }
         StatsTier::Stats => {
             let mut warnings = vec![DISTINCT_COUNT_WARNING.to_string()];
 
@@ -677,24 +742,27 @@ pub async fn preview_iceberg_table(
                     schema.total_bytes = schema.total_bytes.or(Some(agg.total_bytes));
 
                     if has_delete_files {
-                        // Fail closed (F-AUD-1): the aggregated row/value/null
-                        // counts sum live data-file records and do NOT subtract
-                        // merge-on-read deletes, so they over-count. Refuse unless
-                        // the operator opted out via FLUREE_ICEBERG_ALLOW_MOR_DELETES.
-                        let allow_mor = fluree_db_iceberg::mor_deletes_allowed();
+                        // Manifest-list backstop (F-AUD-1): the aggregated
+                        // row/value/null counts sum live data-file records and do
+                        // NOT subtract merge-on-read deletes, so they over-count.
+                        // Reuse the hoisted override flag (env read once) and raise
+                        // the TYPED error (W3) so the 409 discriminant survives
+                        // instead of flattening to a 400 "Invalid configuration".
                         fluree_db_iceberg::ensure_no_delete_manifests(
                             usize::from(has_delete_files),
-                            &table.qualified(),
+                            &table_display,
                             allow_mor,
-                        )
-                        .map_err(|e| crate::ApiError::config(e.to_string()))?;
-                        // Only reached under the override: surface the caveat.
-                        warnings.push(
-                            "Table has merge-on-read delete files; aggregated row/value/null \
-                                 counts are upper bounds (position/equality deletes are not \
-                                 subtracted)."
-                                .to_string(),
-                        );
+                        )?;
+                    }
+                    // Reachable only under the override once a delete signal fired;
+                    // caveat the counts regardless of WHICH signal saw them (the
+                    // summary arm covers a delete manifest that mis-parsed as data).
+                    if preview_counts_are_upper_bounds(
+                        allow_mor,
+                        has_delete_files,
+                        fluree_db_iceberg::summary_indicates_deletes(snapshot),
+                    ) {
+                        warnings.push(MOR_UPPER_BOUND_WARNING.to_string());
                     }
 
                     (manifests_read, agg.had_column_bounds, has_delete_files)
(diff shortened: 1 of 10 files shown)


fluree-db-api/src/graph_source/iceberg_catalog.rs
```diff
@@ -677,6 +677,18 @@ pub async fn preview_iceberg_table(
                     schema.total_bytes = schema.total_bytes.or(Some(agg.total_bytes));
 
                     if has_delete_files {
+                        // Fail closed (F-AUD-1): the aggregated row/value/null
+                        // counts sum live data-file records and do NOT subtract
+                        // merge-on-read deletes, so they over-count. Refuse unless
+                        // the operator opted out via FLUREE_ICEBERG_ALLOW_MOR_DELETES.
+                        let allow_mor = fluree_db_iceberg::mor_deletes_allowed();
+                        fluree_db_iceberg::ensure_no_delete_manifests(
```
bplatz, line 751: **Definite issue** — this path implements only one of the guard's two signals.

Both planners check the summary counters *and* the manifest list. Here only the manifest-list signal (`has_delete_files`) is consulted; `ensure_no_summary_deletes` is never called, even though `snapshot` is in scope at `:635`.

That matters because the manifest-list signal fails **open**:

```rust
// manifest_list.rs:249-252
let content = match get_field("content") {
    Some(AvroValue::Int(i)) => ManifestContent::from_avro(*i),
    _ => ManifestContent::Data,
};
```

Any encoding other than a bare Avro `Int` — a `["null","int"]` union, a `long` — silently reclassifies a delete manifest as a data manifest. The `"default": 0` in `MANIFEST_LIST_SCHEMA_V2` doesn't rescue it either: that const is only used by tests (`manifest_list.rs:390`), while production parses with `Reader::new`, i.e. writer schema only, so the declared default never applies at read time.

On the planners the summary check backstops exactly this case. Here there is no backstop, so a delete-bearing table returns over-counted stats with no error and no warning.

The v2 spec makes `content` a required int, so this is defense-in-depth rather than something I'd expect to fire on today's data — but it is the one place where the two-signal design isn't actually two signals.

Fix is to add the summary check **outside** the `if has_delete_files` block (it needs to run precisely when `has_delete_files` is wrongly `false`):

```rust
let allow_mor = fluree_db_iceberg::mor_deletes_allowed();
fluree_db_iceberg::ensure_no_summary_deletes(snapshot, &table.qualified(), allow_mor)
    .map_err(|e| crate::ApiError::config(e.to_string()))?;
```

One knock-on: the `warnings.push(...)` caveat below is currently reachable only under the override, which is correct — but if the summary arm fires while `has_delete_files` is false, you'd want that warning on the override path too.

fluree-db-iceberg/src/metadata/snapshot.rs
```diff
@@ -86,6 +86,32 @@ impl Snapshot {
             .get("deleted-records")
             .and_then(|s| s.parse().ok())
     }
+
+    /// Get the total number of merge-on-read **delete files** from the snapshot
+    /// summary (`total-delete-files`), if present. A value `> 0` means the
+    /// snapshot carries position/equality delete files that Fluree does not yet
+    /// apply — see [`crate::mor_guard`].
+    pub fn total_delete_files(&self) -> Option<i64> {
```
bplatz, line 94: **Worth a look** — these accessors fail open on malformed input.

`.parse().ok()` turns a present-but-unparseable counter into `None`, which `ensure_no_summary_deletes` then reads as `0` via `unwrap_or(0)` — i.e. "no deletes." The `.max(0)` there also clamps a negative to `0`. Both resolve toward *proceed* inside a guard whose entire premise is to resolve toward *refuse*.

The manifest-list check backstops this on the planners, so it's low-stakes there. It's only genuinely unbackstopped on the preview path (see the `iceberg_catalog.rs` comment). If you'd rather the summary arm carry its own weight, having a present-but-unparseable value refuse — rather than read as zero — would make it fail closed independently.

fluree-db-iceberg/src/scan/planner.rs
```diff
@@ -199,6 +199,12 @@ impl<'a, S: IcebergStorage> ScanPlanner<'a, S> {
         // Determine projection
         let (projected_field_ids, projected_columns) = self.resolve_projection(schema)?;
 
+        // Fail closed on merge-on-read delete files (F-AUD-1): the scan reads
+        // only live data files and never applies deletes, so a MoR snapshot
+        // would silently return deleted rows. Cheap zero-I/O check first.
+        let allow_mor =
+            crate::mor_guard::ensure_summary_scannable(snapshot, &self.metadata.location)?;
```
bplatz, line 206: **Worth a look** — the guard names the table's storage location, not the table.

Both planners pass `&self.metadata.location`, so the operator-facing error reads ``Refusing to read Iceberg table `s3://bucket/dw/fact_orders` ``. The preview path passes `table.qualified()` and produces `dw.fact_orders`, so the same refusal reads differently depending on which path tripped it.

Worth noting that `error_message_is_actionable` doesn't catch the difference — it calls the guard directly with a literal `"dw.dim_customer"`, so it asserts against a string that neither planner actually produces. Not wrong, just not covering the real call sites' output.

The location is arguably the more actionable identifier of the two; the main thing is that they agree, so an operator grepping logs finds both.

fluree-db-api/src/graph_source/iceberg_catalog.rs
```diff
@@ -677,6 +677,18 @@ pub async fn preview_iceberg_table(
...
+                            usize::from(has_delete_files),
+                            &table.qualified(),
+                            allow_mor,
+                        )
+                        .map_err(|e| crate::ApiError::config(e.to_string()))?;
```
bplatz: **Worth a look** — `ApiError::config` for what isn't a configuration problem.

Nothing is misconfigured when this fires: the config is correct and the table simply has delete files. Flattening it into a config error likely surfaces to the caller as a 4xx configuration complaint, which points the operator at the wrong thing — and it drops the typed `MergeOnReadDeletes` discriminant that the rest of the PR went to the trouble of introducing.

bplatz (approved):
Approving — the design is right, and I verified the substantive claims locally rather than taking the gate table on faith: `cargo test -p fluree-db-iceberg --all-features` gives 203 passed / 0 failed, and `cargo clippy -p fluree-db-iceberg --all-features --all-targets` is clean.

Three things I checked independently and can confirm:

- **No bypass on the scan path.** Every production entry point (`r2rml.rs:1074` and `:1117`, `iceberg_sample.rs:143`, `ephemeral.rs:179`) routes through `SendScanPlanner::plan_scan`, so all of them inherit the guard.
- **The scan-files cache-hit arm is genuinely safe on `main`,** as claimed. It rebuilds tasks without re-entering `plan_scan`, but the key is `metadata_location`, which Iceberg versions per commit, and entries are only ever written from the guarded miss arm — so a cached entry can only correspond to a snapshot that already passed the guard.
- **`send_read_snapshot_data_files` has exactly one production consumer** (`iceberg_catalog.rs:651`), and it is guarded.

One **definite issue** and three **worth a look** items inline. The definite one is the metadata-preview path implementing only one of the two signals — it's a small fix, and I'd take it before merge since it leaves open the same silent over-count this PR exists to eliminate.

Two non-blocking notes:

- The `/info` `table_row_counts` deferral is fine to defer, but worth flagging for #1528 that it is *structurally* unable to detect deletes rather than just missing a call: `fetch_virtual_table_row_counts` is metadata-only by design (`loadTable` summary, never reads the manifest list), so closing it needs a different mechanism than this guard, not a guard call.
- This converts silently-wrong into hard-fail for anyone already querying Athena/Flink MoR tables. Agree that's the correct call and the escape hatch covers it, but it probably deserves a release note — the failure is immediate and total rather than gradual.

fluree-db-api/src/graph_source/iceberg_catalog.rs
aaj3f: Confirmed — and it extended one tier further than flagged: the Schema tier surfaces `total_records` with no delete detection at all (`iceberg_catalog.rs:540-543` → returned at `:619-628` guard-free), so a summary check added only to the Stats arm would have left the same over-count reachable. Fixed in a3068f826 by hoisting a shared `preview_summary_guard` that runs before BOTH tiers (planner pattern — env read once, override flag reused by the existing manifest check inside `if has_delete_files`), with the upper-bound caveat now pushed from either signal on the override path. Regression tests cover Stats + Schema tiers for summary-reported deletes, manifest-only deletes with a silent summary, and override-with-warning. One nit on the write-up for the record: the call the preview was making is `ensure_no_delete_manifests` (the manifest-signal guard) — the substance of the comment was exactly right, the named function just differs from the one at the site.

fluree-db-iceberg/src/metadata/snapshot.rs
aaj3f: Confirmed and fixed in 83c13d9b7, with one deliberate placement choice: the fail-closed flip lives in the guard (`ensure_no_summary_deletes` now reads the raw summary map; a PRESENT key that fails integer parse — or parses negative — refuses with the malformed key/value named in the evidence text; ABSENT keys still proceed, since that arm is what the manifest-list check backstops), rather than in the shared accessors. Reason: the accessors' `Option<i64>` shape is consumed for display on the audit line (`/info`), where "unparseable" must surface as deletes-present rather than a refusal — the guard-level flip gives both consumers the right behavior, and the audit-line copy is reconciled to the same semantics in 8effa91e3. Unit tests: `"garbage"`, `"-1"`, absent-key-proceeds. Spec check for the false-positive question: snapshot `summary` is `map<string,string>` with string-encoded integer counters, so no conforming writer emits a non-integer — refusal has effectively zero false-positive risk.

fluree-db-iceberg/src/scan/planner.rs
aaj3f: Fixed in 83c13d9b7: the preview now renders `qualified (location)` so both identifiers appear in one refusal, planners keep the location they have in scope, and an operator grepping logs finds either form. `error_message_is_actionable` no longer asserts a hand literal — it goes through the real call-site shapes for both the planner and preview renderings.

fluree-db-api/src/graph_source/iceberg_catalog.rs
aaj3f: Fixed in a3068f826: new feature-gated `ApiError::Iceberg(#[from] IcebergError)` preserves the typed discriminant end-to-end; `MergeOnReadDeletes` maps to 409 (nothing is misconfigured — the table simply has delete files), Display carries the guard message verbatim so every substring consumer survives. Downstream check as part of this fix: solo's only consumer of the preview error wraps it as its own fixed 400 and discards the db status code, so the 400→409 change is invisible there while the preserved message keeps any text-based handling working; surfacing the 409 in solo would be a solo-side follow-up. Tests assert 409, both identifiers, and that the message is no longer an "Invalid configuration".

aaj3f:
All four review items are addressed at this head (83c13d9b7 + a3068f826): the preview guard now covers both tiers including the Schema-tier gap that had no detection at all, summary counters fail closed on malformed input at the guard level, the refusal identifier is unified across planners and preview with the test retargeted to real call-site output, and the preview returns a typed 409 instead of a 400 config error with the message text preserved for downstream substring consumers (solo verified unaffected). Per the release-note suggestion: no CHANGELOG file exists in-repo — release notes are generated from PR titles via `.github/release.yml` — so the `breaking-change` label is now applied, which files this PR's title under Breaking Changes in the next release's notes. The `/info` structural note is handled on the audit line exactly as you anticipated (zero-I/O summary counters, no guard call).

PR #1522, opened 2026-07-19
perf(r2rml): coverage widenings — scan-side pruning/budget/IO + fold-side aggregate admission (audit Tier-2)

aaj3f:
## Forest map

PR 3 of 4 in the big-iceberg-audit implementation phase (audit working set `audit-2026-07/00-MASTER-AUDIT.md` §6 Tier-2; bundling decision `audit-2026-07/decisions/DEC-001-pr-bundling.md` — both in the team's audit set, not committed to the repo). It bundles the eight Tier-2 coverage widenings — every one an ADMISSION gap where a shape misses a fast-path gate and falls to full per-row RDF materialization (F-AUD-15 ceiling; C2b measured each cliff). The PR is rigidly sectioned along the DEC-001 soundness seam: **Section 1 (scan-side)** = bounds/prune/budget/IO (does the prune drop only provably-non-matching data?), **Section 2 (fold-side)** = aggregate/stats (over-count guards, term parity). Between them it **closes/advances F-AUD-5, F-AUD-6, F-AUD-7, F-AUD-8, F-AUD-11** and audit **items 12 + 14**. Every mechanism ships with a kill switch, corpus member(s) with blessed native oracles, and the local gate record (§Verification) — CI does NOT fire here (base ≠ main), so the local reproduction is the verification of record and the live virtual re-bless lands in PR-HARNESS (PR 4).

Base = `perf/audit-mem-guards` (PR-SAFE-MEM, #1521). Sibling stack: PR-SAFE-MOR (#1520, on main) → integration `perf/audit-tier012` → PR-SAFE-MEM (#1521) → **this** → PR-HARNESS (leaf, re-bless).

---

## Section 1 — scan-side (bounds / prune / budget / IO)

The soundness question is "does the prune / early-termination drop only provably-non-matching data / preserve completeness?" C2b already proved the machinery works when admitted (ASK pruned 7,670→0 files; FILTER-decimal pruned 7,670/7,670); these five widenings admit more shapes to it.

**[7] FILTER-IN / VALUES → `Expression::In`** (`966224d6b`, F-AUD-5). `ScanCmpOp::In` + `ScanValue::Set`, emitted from `FILTER … IN` and single-var `VALUES` over an exactly-one-scalar-POM; whole-or-nothing per member; cap `FLUREE_R2RML_IN_PUSHDOWN_MAX`=64. Switch `FLUREE_R2RML_IN_PUSHDOWN`. Scoped: FK-IRI IN (probe-02) still declines to prune (the FK-key template reversal is a documented follow-on) — q069 ships as a parity member; the scalar prune win is q070.

**[8] ASC scan-side top-k** (`b58c9c2d3`, F-AUD-6). Admits ASC for a REQUIRED (non-nullable) column only (the nulls-first soundness subtlety that declined ASC); `TopKBound`/`plan_topk_read` generalized worst-first (DESC regression tests preserved). Switch `FLUREE_R2RML_TOPK_ASC`. Member q071.

**[10] Timestamp manifest pushdown** (`1a3d0d104`, F-AUD-11). `ScanValue::Timestamp{micros,tz}` + `LiteralValue::TimestampTz`, frame-matched at build (UTC↔timestamptz, naive↔timestamp). MANIFEST-LEVEL ONLY (no row-group Int64 arm — parquet logical-unit ambiguity). Switch `FLUREE_ICEBERG_TIMESTAMP_STATS`. Member q072.

**[11] Row-budget forwarding** (`5db220551`, F-AUD-7). OPTIONAL forwards its LIMIT budget to the required side (sound: ≥1 output/required row — closes probe-04's 68,828× read amplification). **MINUS DECLINES** (does NOT forward): its anti-join output ⊆ primary with multiplicity ≤1, so a finite primary budget can under-produce — the brief's "MINUS forwards" is unsound. FILTER/SORT/DISTINCT/HAVING/subquery get explicit observable declines. Switch `FLUREE_R2RML_BUDGET_OPTIONAL`. Members q073 (cliff) + q074 (control).

**[12] `read_ranges` parallel GETs** (`26bde327a`, B1-AppD/F-AUD-16). Wired the zero-caller `read_ranges` into the coalesced sparse-fetch loops; fixed a latent bug — the dormant fn used completion-order `buffer_unordered` (would have paired wrong bytes to wrong ranges) → order-preserving `buffered`. Switch `FLUREE_ICEBERG_PARALLEL_RANGE_GETS`. No new member by design (the cold corpus members exercise it; PR-HARNESS's cold subset is its evidence).

---

## Section 2 — fold-side (aggregate admission)

The soundness question is entirely different — over-count guards and term parity. Both fold widenings ride the EXISTING `FLUREE_FUSED_R2RML_AGG` master switch as WIDENINGS (no new switch); per A2's lesson this is called out explicitly: **switch-OFF reverts the widenings too** (back to the pre-PR full-materialize decline).

**[9] MIN/MAX fused admission** (`c3c5f4c4e` + tie-break fix `e7c32b9c8`, F-AUD-8). `AggregateFn::Min`/`Max` now fold: `Fold::MinMax` materializes ONLY the candidate object term (no subject IRI, no per-row BindingRow) through the same `materialize_object_from_batch` + `LiteralEncoder` path the FILTER fold uses, keeping the running extreme by `compare_bindings` with the SAME tie-break as `min_by`/`max_by` (MIN keeps first-on-ties, MAX keeps last-on-ties — so an equal-but-differently-rendered extreme like a double `+0.0`/`-0.0` resolves to the exact element the generic aggregate returns) — **byte-parity with the generic `agg_min`/`agg_max` by construction**, but STREAMING (O(1) memory) instead of buffering every value. Scoped to numeric (int/long/double/decimal) + date/timestamp PLAIN-LITERAL columns (string collation, lang/IRI, and un-annotated → xsd:string decline). Grouped + ungrouped. No manifest/column-stats shortcut in this PR (fold-only — that ungrouped-stats shortcut is a documented follow-on; deletes + filters complicate the per-file bounds). Members: q075 (ungrouped), q076 (grouped, 4 channels), blessed native oracles.

**[9b] Filtered-COUNT fused admission — the q038 class** (`c3c5f4c4e`, C2 / D-c5 slice-1.5; the highest single-query payoff in the corpus). q038 = a COUNT with a constant-object constraint (`isCurrent true`) that blocks BOTH the manifest shortcut AND the fused fold → full materialize at ~7k rows/s (57.6s virtual, 301× native, DNF at SF20). The fused single-table path previously DECLINED any `star_constraints` to avoid an over-count (the D-c5 O1 ship-blocker); it now RESOLVES each constraint to a per-row scalar-column check and APPLIES it in the fold, via the SAME `resolve_star_constraint_checks` / `row_satisfies_constraints` machinery the join path (E2) already uses. **The D-c5 guard discipline is honored exactly:** (a) the decline is removed ONLY for shapes the fold now applies constraints for — a constraint that doesn't resolve to a scalar column (a RefObjectMap object) still declines, never a silent over-count; (b) parity is pinned at two levels — the fused-vs-materialized equality is the corpus native oracle (q038 + the new multi-constraint q077 = isCurrent∧segment=Enterprise = 50038, plus the grouped q022/q061 over-count tripwires, which now FUSE with the constraint applied, oracle unchanged), and the AND-semantics + NULL-column drop are hermetic unit tests (`multi_constraint_requires_all_to_match`, `e2_row_satisfies_boolean_flag`); (c) the COUNT(*) manifest shortcut stays DECLINED for a constraint-bearing plan (it requires `fact_constraints.is_empty()` — the delete-blind `record_count` cannot see per-row matches). Predicted telemetry shift on q038 (LOCAL native check; live virtual confirmation is PR-HARNESS's re-bless): scan_table n=1 fused, wall from ~57.6s virtual → seconds class.

**[14] `get_data_model` beyond `t==0` + the `/info` MoR rider** (`50368671e`, C1 residual + #1520 F-AUD-1). C2's manifest-backed `/info` route was gated on an empty native shell (`t==0`), so `get_data_model` lost the virtual model the moment a virtual-dataset shell took any native write. Routing is now per-member: a graph-source-registered ledger serves its manifest-backed stats regardless of native `t` — an empty shell serves virtual-only (unchanged); a HYBRID (`t>0` + graph source) MERGES the two members (native ledger/commit metadata authoritative + native-only classes preserved; the graph-source member's classes overlaid by IRI — a UNION, never a SUM, so no double-count — with its `source`/snapshot block attached), fail-safe to the virtual model on a deserialize miss. New default-on switch `FLUREE_R2RML_INFO_MEMBER_ROUTING` (off = the strict t==0 reroute); the `FLUREE_ICEBERG_INFO_COUNT_BUDGET_MS` time-budget is unchanged. RIDER: `fetch_virtual_table_row_counts` derived per-table counts from the snapshot-summary `total-records`, which OVER-COUNTS a merge-on-read table; it now flags such a table (via `mor_guard::summary_indicates_deletes`, zero I/O) into a new `mor-approximate-tables` list on the `source` block, so the count is an honest upper bound rather than silently authoritative — matching the existing `StatsCompleteness.has_delete_files` convention; consumers that ignore the new key (the /data-model + MCP readers, all null-safe per C-adversarial O4) are unaffected. Hermetic tests for hybrid routing, the class-union merge, and the MoR flag.

---

## Verification

CI does NOT fire on this PR (base ≠ main; the workflow's `pull_request: branches:[main]` filter matches the target). The full gate record is committed at `docs/audit-impl/cov-scan-gates.md` (scan-side + fold-side sections) and is the **verification of record**; the **perf arm is ADVISORY until PR-HARNESS re-blesses** at the final integrated head (the committed vbench baseline is 62% gate-blind — C2/RT2). The correctness/hash arm is live: all new members carry blessed native oracles.

Fold-side run (this PR's delta), all green:
- **fmt** clean (the pre-existing `hash_join.rs` base drift reverted, per db-verify-gotchas).
- **clippy** `-p fluree-db-query --lib --tests --bins --no-deps` → 0 (also fixed a pre-existing scan-side `semicolon_if_nothing_returned` deny at `operator_tree.rs:3401`, so the whole branch is clippy-green); `-p fluree-db-api --features iceberg --all-targets --no-deps` → 0 (one pre-existing non-gating `question_mark` warning on unchanged code).
- **test** `-p fluree-db-query` (lib 1306 + all it_*/grp_* bins) → 0; `-p fluree-db-api --features iceberg` → 0 (incl. the 3 new item-14 tests); `-p fluree-bench-virtual --bins` → 33 pass, `shipped_corpus_is_valid` = **77 members**.
- **native oracles** blessed via `vbench baseline --expected --targets native-sf01` (local ledger, no creds): q075/q076/q077. q077's native run took 133s (dev build — a native-planner cost for a doubly-constrained COUNT; the fused/virtual path applies both constraints in one scan, which is the point).

---

## Residuals (carried to PR-HARNESS / follow-ons)

- **FK-IRI IN-set pruning** (F-AUD-5, probe-02/q069): scalar-column IN lowers; reversing the subject/FK template to recover key values is a distinct follow-on. q069 documents the gap as a parity member.
- **MINUS budget forwarding**: ruled UNSOUND (anti-join absorption) — shipped as an explicit documented decline, not a widening.
- **String MIN/MAX**: declines (collation) — numeric/temporal only this PR.
- **Manifest/column-stats MIN/MAX shortcut**: the true ceiling-buster (answer without a data-file scan) — deferred (per-file min/max bounds are complicated by deletes/filters); this PR is fold-only.
- **Per-ledger native+virtual merge (item 14)**: the merge takes a UNION with the graph source winning IRI collisions (sound for the deployed shell + disjoint-hybrid cases; documented). A genuinely-overlapping same-IRI hybrid is not the deployed topology.
- **SWITCHES.md**: this PR adds `FLUREE_R2RML_INFO_MEMBER_ROUTING`; the six scan-side switches + this one are pending SWITCHES.md rows in PR-HARNESS. Items 9/9b ride the existing `FLUREE_FUSED_R2RML_AGG` (widening — documented in the gate doc + Section 2).

### DIFF
--- a/docs/audit-impl/cov-scan-gates.md
+++ b/docs/audit-impl/cov-scan-gates.md
@@ -1,0 +1,74 @@
+# PR-COVERAGE — scan-side gate record
+
+Branch `perf/audit-coverage` (base = `perf/audit-mem-guards` @ 1441e0038). Scan-side
+section: audit items 7, 8, 10, 11, 12. Recorded by the scan-side implementer; the
+fold-side implementer appends its section below. The single PR body cites this file.
+
+CI does not fire on this branch (base ≠ main); this is the local reproduction of
+record. The **perf arm is ADVISORY until PR-HARNESS re-blesses** at the final
+integrated head — the committed perf baseline is 62% gate-blind (C2/RT2). The
+correctness/hash arm is live: new members carry blessed native oracles (below).
+
+Toolchain: env `cargo`/`clippy` (no repo pin). Two known-pre-existing quirks handled
+per `db-verify-gotchas`: (1) `cargo fmt` also touches `hash_join.rs` (pre-existing
+drift) — reverted, left out of this branch; (2) the `smoke_minio` **example** fails to
+compile without `--features aws` (its own doc requires aws) and is broken on the base
+too — excluded from the scoped gates via `--lib --tests --bins`.
+
+## Scan-side gates (verbatim)
+
+- **fmt:** `cargo fmt` then reverted the unrelated `hash_join.rs` hunk → the 26
+  changed source files are fmt-clean (`cargo fmt --check` clean on them).
+
+- **clippy (query):** `cargo clippy -p fluree-db-query --lib --tests --bins --no-deps`
+  → no warnings/errors in any item-7/8/10/11 file (env clippy's pre-existing
+  workspace lints are out of scope per db-verify-gotchas; CI main clippy has no
+  `-D warnings`).
+
+- **clippy (api, iceberg):** `cargo clippy -p fluree-db-api --features iceberg
+  --all-targets --no-deps` → exit 0.
+
+- **clippy (iceberg, aws):** `cargo clippy -p fluree-db-iceberg --features aws
+  --lib --tests --bins --no-deps` → clean for item-10/12 files (examples excluded:
+  `smoke_minio` is aws-doc'd and pre-existing-broken).
+
+- **test (query):** `cargo test -p fluree-db-query --lib` → **1306 passed; 0 failed**.
+  Full `cargo test -p fluree-db-query` (incl. grp_* integration bins) → exit 0.
+
+- **test (iceberg, aws):** `cargo test -p fluree-db-iceberg --features aws --lib`
+  → **246 passed; 0 failed**.
+
+- **test (api, iceberg):** `cargo test -p fluree-db-api --features iceberg` → exit 0
+  (live/`#[ignore]` tests skipped; hermetic tests pass).
+
+- **test (bench-virtual):** `cargo test -p fluree-bench-virtual --bins` → exit 0;
+  the 7 corpus meta-tests (incl. `shipped_corpus_is_valid` = 74 members,
+  `rows_only_hash_gate_marks_nondeterministic_limits`) pass.
+
+## New corpus members + blessed native oracles (item 7/8/10/11)
+
+Promoted from the A3 probe battery; run `vbench run --targets native-sf01` (local
+ledger `/Users/ajohnson/vbench/.fluree`, NO credentials — uses the on-disk iceberg
+cache), blessed via `vbench baseline --expected`. **Native oracles only** — the live
+virtual gate is PR-HARNESS's job.
+
+| member | item | probe | native rows | hash (12) | exec | hash_gate |
+|--------|------|-------|-------------|-----------|------|-----------|
+| q069 filter_in_fkref       | 7  | probe-02  | 1100 | 67a2035630da | from | full |
+| q070 scalar_values_glaccount | 7 | probe-03 | 0    | e3b0c44298fc | from | full |
+| q071 asc_topk_order_total  | 8  | probe-01  | 10   | 7dc1c655cb0c | from | full |
+| q072 timestamp_range_webevent | 10 | probe-07 | 5000 | 498045e221ce | from | rows_only |
+| q073 optional_budget_order_customer | 11 | probe-04 | 50 | b86545e2b0e2 | from | rows_only |
+| q074 limit_budget_control  | 11 (control) | probe-04b | 50 | 6e0f430cd272 | from | rows_only |
+
+Item 12 (`read_ranges` parallel GETs) adds NO member by design — the existing cold
+corpus members exercise the coalesced-fetch path; PR-HARNESS's cold-subset re-run is
+its evidence.
+
+## Notes carried to the PR body / SWITCHES.md (PR-HARNESS)
+
+New kill switches (all default-on): `FLUREE_R2RML_IN_PUSHDOWN`,
+`FLUREE_R2RML_IN_PUSHDOWN_MAX` (=64), `FLUREE_R2RML_TOPK_ASC`,
+`FLUREE_ICEBERG_TIMESTAMP_STATS`, `FLUREE_R2RML_BUDGET_OPTIONAL`,
+`FLUREE_ICEBERG_PARALLEL_RANGE_GETS`.
+</content>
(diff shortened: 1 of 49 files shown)


aaj3f:
Consolidating at maintainer request. #1522 (the eight Tier-2 coverage widenings — scan-side pruning/budget/IO and fold-side aggregate admission) is carried forward in #1528 (the big-Iceberg-audit implementation), which brings the audit program forward as a single reviewable unit. Its verification of record is recorded under R-1522 in [pr-reviews-impl.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/pr-reviews-impl.md) (verdict SHIP; every mechanism admission-only, the in-engine FILTER/SortOperator authoritative), alongside the whole-stack live gate of 77 records / 0 hash mismatches / 0 perf violations. The pre-consolidation branch tip is preserved at tag `archive/pre-refactor-2026-07-21/perf_audit-coverage` — no history was rewritten. Closing in favor of #1528; discussion continues there.

PR #1524, opened 2026-07-19
Rework history sidecars as transition logs

zonotope:
## Problem

The V6 index has no TSPO index; time travel reconstructs past states by walking history sidecar (FHS1) entries backward from the current leaf via `replay_leaflet`, which reads an assert entry as "this fact was absent immediately before this `t`" and a retract entry as "this fact was present immediately before this `t`". The writers did not honor that contract. Incremental indexing deduplicated each window's events down to one winner per fact and discarded the losers, and the merge synthesized retract+assert pairs for replaced rows, so sidecars recorded a mix of real transitions, fabricated events, and nothing at all for facts whose state changed mid-window.

This produced wrong `AS OF` answers served from the index, verified end to end before the rework:

- A fact deleted and re-added within one indexing window showed as present during the gap between the delete and the re-add (the in-window retract was discarded, so replay never saw the absence).
- A fact created and deleted within one window vanished from history entirely (no row survives and neither event reached the sidecar, so its brief life was unqueryable).
- A fact retracted twice in one window recorded the second retract's `t`, so `AS OF` queries between the two retracts wrongly showed it present.

The full rebuild path had the mirror-image problem: it wrote every surviving event to the sidecar, including the winner assert that also materializes as the row, so `history` queries — which read base rows and sidecar entries as independent flakes — returned duplicate asserts.

## Approach

The sidecar is now a transition log, governed by two invariants that live in the new `fluree-db-binary-index/src/format/transitions.rs` module doc:

- **Transition-log invariant**: writers record exactly the events that change a fact's presence. No-op events (re-assert of a present fact, retract of an absent one) are never written, because replay would invert them into false state changes.
- **Row-assert conservation**: a fact's materialized assert lives either as its base row or as a sidecar entry, never both (history queries would emit duplicates) and never neither (replay would fabricate presence before the fact's birth). The final assert of a resolution becomes the row; writers that consume an existing row migrate its assert into the sidecar first.

Both writer paths resolve presence through one shared state machine, `resolve_transitions`: sort a fact identity's events by `(t, op)` with retract-first ties, fold presence from the base state, keep only the ops that flip it, and pop a final assert out as the materialized row.

### Superseded event stream

`dedup_fact_lifecycles` still picks one winner per fact identity per window, but now returns the replaced events as a second `superseded` stream instead of discarding them (no-op re-asserts are still dropped entirely — they are not state changes). `IncrementalNovelty` carries `superseded`/`superseded_ops` alongside `records`/`ops`, and Phase 2 routes the superseded stream per (graph, order) through all four sort orders into the branch update.

### Merge

`merge_novelty` takes the superseded stream as a third cursor. On a matched row (Equal arm), the consumed row's assert migrates into the sidecar at the row's own `t`, then the row's ops, superseded events, and the winner resolve through `resolve_transitions` against base-present; the synthesized retract+assert pair is gone. Unmatched winners (Greater arm and drain) resolve against base-absent, so a winner retract with no prior presence writes nothing. `assemble_history` is now a straight newest-first sort and concat; the boundary dedup machinery (`FactKeyV3`, adjacent-assert dedup) is removed because conservation makes it unnecessary.

### Rebuild parity

`index_build` resolves each fact identity's full event sequence through the same `resolve_transitions` call (base-absent, since a rebuild sees the entire log), so rebuild and incremental agree on which events are transitions. A rebuilt row carries the earliest assert of the fact's final presence run.

### Stream co-routing invariant

Every superseded identity must travel in the same slice as its winner or the merge resolves against the wrong base state. The `StreamSlices` type pairs the winner and superseded slices structurally, established at the two slicing constructors (`slice_streams_to_leaves`, `slice_streams_to_leaflets`) instead of debug-asserted at each of the four consumption points.

### Stats

Row counts previously assumed every novelty assert added a row. The merge now reports base-presence matches (`MergeOutput::matched`, collected on the SPOT order only) so stats apply deltas instead of drifting on replaced rows.

## Semantics decisions

Two behaviors surfaced by review are blessed and documented in the `transitions.rs` module doc:

- **Index-served history is transition-grade.** `history` queries over the index report state changes; no-op re-asserts are queryable only from the commit log, which remains the authoritative event record. Event-grade audit belongs to commit replay, not sidecars.
- **Visible `t` is window-granular.** A row's `t` is when the fact most recently became present, observed at indexing-window granularity: a re-assert that crosses an indexing boundary replaces the row at the re-assert's `t` (matching the live novelty overlay), while a rebuild sees the whole log as one window and normalizes to the birth `t`. A reindex may therefore change a re-asserted fact's reported `t`; presence at every `t` is unaffected.

## Known accepted gaps

- Fresh-graph first windows (`DefaultFresh`/`NamedFresh`) build leaves directly rather than through the merge, so their superseded events are dropped with a `tracing::warn!`; in-window lifecycles in a graph's first window are invisible to time travel. Documented in the plan file; reachable beyond genesis via roots with fewer than four orders.
- Sidecars written before this change may carry duplicate asserts until the next rebuild; the read path tolerates them and a rebuild normalizes.
- Sidecar growth for hot facts between rebuilds is unbounded; compaction is deferred.

## Testing

- `replay_property_tests.rs`: an exhaustive property harness that runs every op chain up to length 4 across both base states and 11 query targets through the real pipeline (dedup → routing → merge or rebuild → encode → decode → `replay_leaflet`) and checks presence and row `t` at every target against an independent reference fold, for both the incremental and rebuild paths.
- `transitions.rs` unit tests pin the state machine's edge cases, including base-present with zero transitions and same-`t` tie ordering.
- `it_replay_gap_check.rs`: end-to-end regressions for the three original wrong-answer scenarios, asserting replayed state through the real branch merge and replay path.
- Merge and rebuild unit tests pin row-assert migration, synthesized-pair suppression, first-retract-wins, and never-asserted-retract suppression.
- Downstream suites green: fluree-db-api `grp_query_history` 35/35, fluree-db-query, `grp_query_sparql`, `grp_transact`, full workspace build, fmt, clippy.


fluree-db-indexer/src/run_index/build/novelty_merge.rs
```diff
@@ -352,17 +319,24 @@ pub fn merge_novelty(input: &MergeInput<'_>) -> MergeOutput {
...
-            new_history.push(record_to_hist_entry(nov, op));
         }
         ni += 1;
     }
 
+    debug_assert_eq!(
+        si,
+        input.superseded.len(),
+        "every superseded event's identity must have a winner in this leaflet's novelty"
+    );
```
bplatz, line 342: Not a blocker, but worth considering: if the co-routing invariant is ever violated in release, the cursor in `gather_identity_events` never consumes the stranded events and they're silently dropped from history — this assert only catches it in debug builds. Since the check is O(1), a release-mode `tracing::error!` (or hard error) here would surface the failure instead of losing time-travel data quietly.

Related: the invariant's soundness rests on an unstated premise — all novelty/superseded `t`s exceed the base watermark, so no boundary key (taken from existing rows) can sort between two same-identity events. Worth stating on the `StreamSlices` doc.

bplatz (commented):
Not blockers, but two things worth considering:

- Inline comment on the co-routing check in `novelty_merge.rs`.
- `docs/design/index-format.md` (line ~265) still describes Region 3 as an "optional operation log". This PR redefines it as a strict transition log with row-assert conservation — those semantics currently live only in the `transitions.rs` module doc. A pointer from index-format.md would keep the format doc accurate.

bplatz (approved):

PR #1526, opened 2026-07-21
fix: keep history entries in the same leaflet as their row

bplatz:
## Summary

The V3 `LeafWriter` attached each pushed history entry to whatever leaflet was mid-buffer at push time. In the rebuild push order a fact's transition entries are pushed immediately before its materialized row, so when that row opened a new predicate/`o_type` segment (a new leaflet), the entries were stranded in the **previous** leaflet's sidecar segment.

Time-travel replay is per-leaflet, so the row's own leaflet then carried an empty history segment and any AS OF query before the fact's final assertion silently dropped it — a correctness bug for historical queries after a full index rebuild.

## Fix

Buffer pushed history in `pending_history` and commit it only once the row's leaflet is known:

- `commit_all_pending` on `push_record` (after any segment-boundary flush) and on `finish` — entries follow their row into the leaflet that actually holds it.
- `commit_pending_for_current_segment` (segment-key matched) just before a boundary flush, so a non-materialized fact that ends a segment still lands in that segment's leaflet.

Only the full-rebuild path (`index_build`) pushes history through this writer; all incremental paths use `push_record` only, so nothing else relied on the old eager-attach behavior.

## Tests

Adds two regression tests that both failed before the fix and pass after:

- `fluree-db-binary-index/tests/it_history_sidecar_placement.rs` — `LeafWriter` unit level.
- `fluree-db-indexer/tests/it_rebuild_history_placement.rs` — end-to-end through `build_index`.

Each builds a fact whose row opens a new PSOT predicate segment, then replays that segment AS OF a time before the fact's final assertion and asserts the fact is still visible.

## Docs

Documents the **history co-location** invariant in `docs/design/index-format.md`.

### DIFF
--- a/fluree-db-binary-index/src/format/leaf.rs
+++ b/fluree-db-binary-index/src/format/leaf.rs
@@ -96,9 +96,19 @@ pub struct LeafWriter {
     // History accumulation.
     sidecar_builder: HistSidecarBuilder,
     /// Whether a sidecar segment has been started for the current leaflet buffer.
-    /// Reset to false on flush_leaflet(). Set to true on first push_history_entry()
-    /// or on flush_leaflet() itself (which starts a segment for the encoded leaflet).
+    /// Reset to false on flush_leaflet(). Set to true when history is committed to
+    /// the current leaflet, or on flush_leaflet() itself (which starts a segment
+    /// for the encoded leaflet).
     current_segment_started: bool,
+    /// History entries pushed but not yet committed to a leaflet segment.
+    ///
+    /// In the build push order, a fact's transition entries are pushed immediately
+    /// before its materialized row (`push_record`). Buffering them here — rather
+    /// than eagerly attaching to whatever leaflet is mid-buffer — lets us commit
+    /// them to the leaflet that actually holds their row, even when that row opens
+    /// a new predicate/o_type segment (and thus a new leaflet). See
+    /// `commit_pending_for_current_segment` / `commit_all_pending`.
+    pending_history: Vec<HistEntryV2>,
 
     // Segmentation tracking.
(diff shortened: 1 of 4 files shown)


fluree-db-binary-index/src/format/leaf.rs
```diff
@@ -177,6 +192,11 @@ impl LeafWriter {
             RunSortOrder::Spot => {}
         }
 
+        // This record's transition entries (and those of any preceding
+        // non-materialized facts in the same gap) belong to the leaflet that
+        // holds this record — the current one, post any segment-boundary flush.
+        self.commit_all_pending();
+
```
zonotope, line 202: This call can start a new sidecar segment, but it comes after the outgoing leaflet was flushed at line 176, but before the leaf threshold check at line 209. If the outgoing leaflet flush pushed the leaf over the threshold, then the new sidecar segment started here will still be open with no leaflet to pair with when the sidecar builder is taken when the leaf is flushed. That segment's data will then be lost. 

`flush_leaf` also doesn't reset `current_segment_started`, so it will still be true when the next leaf starts a fresh builder, so any history for that leaf committed before the first leaflet flush for that leaf will also be dropped. 

The leaf threshold check should be called right after the leaflet flush and `flush_leaf` should also reset `current_segment_started`.

zonotope (approved):
There is a bug here that could corrupt history with a relatively quick fix, but everything else looks fine.

fluree-db-binary-index/src/format/leaf.rs
bplatz: Thanks, addressed in ec4fab0525b4cc34836ceb49b6fd72981e316d8b

PR #1527, opened 2026-07-21
feat: Enrich Cypher transaction commit result for write parity

bplatz:
## Problem

`commit_cypher_transaction` returned only the committed `t`, and both write staging helpers ran `Tracker::disabled()`. As of v4.1.3, `WritePlan` gained a `Sequential` variant for multi-clause statements that can only be driven through the `CypherTransaction` API — but that API couldn't report the signals a transactor reads off `stage().execute()`: `commit_id`, fuel (billing), and indexing state (`needed`, `novelty_size`). Without fuel, multi-clause writes can't be billed; without the indexing signals, novelty accumulates and never triggers background reindexing.

## Changes

- **Meter fuel.** `begin_cypher_transaction` now takes `TrackingOptions` and pins a shared `Tracker` on the transaction. Since `Tracker` accumulates atomically and fuel is charged at staging time, fuel sums automatically across every write statement and every clause of a sequential write — no per-clause bookkeeping. Both `Tracker::disabled()` sites now use the transaction's tracker.
- **Enrich the commit result.** `commit_cypher_transaction` returns `TransactResultRef` — the same struct the autocommit `execute()` path yields — so callers map one result shape. Carries the final `commit_id` + `t`, the flake count summed across all staged statements, the indexing signals (`needed`, `novelty_size`), and the accumulated fuel tally. Per-statement `RETURN` rows continue to flow via `CypherTxnWriteOutcome`.
- **Warning cleanup.** Gate the `format_binding_with_result` re-export to the `iceberg` feature (its only consumer), clearing an unused-import warning under default features.

Callers updated: `bolt.rs` (begin + commit) and the bolt transaction test. Added tests covering the fuel tally + indexing signals on the enriched result, and that opting out of tracking leaves the tally empty.

## Testing

- `it_cypher_seq_writes` — 20/20 pass (incl. 2 new).
- Server bolt/cypher suite — 48/48 pass.
- `cargo clippy -p fluree-db-api -p fluree-db-server --all-features --all-targets` clean; warning fix verified under both default and `iceberg` features.

### DIFF
--- a/fluree-db-api/src/cypher_txn.rs
+++ b/fluree-db-api/src/cypher_txn.rs
@@ -42,7 +42,10 @@
 use crate::cypher_write::{self, WritePlan};
 use crate::error::ApiError;
 use crate::view::GraphDb;
-use crate::{Fluree, GovernanceOptions, Result, Tracker};
+use crate::{
+    CommitReceipt, Fluree, GovernanceOptions, IndexingStatus, Result, Tracker, TrackingOptions,
+    TransactResultRef,
+};
 use fluree_db_core::ContentId;
 use fluree_db_ledger::LedgerState;
 use fluree_db_nameservice::{CasResult, RefKind, RefValue};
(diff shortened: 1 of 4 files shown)


fluree-db-api/src/cypher_txn.rs
```diff
@@ -379,11 +400,48 @@ impl Fluree {
...
+            // ledger, so a prior commit — hence a base head id — always exists.
+            let commit_id = txn
+                .base_head
+                .as_ref()
+                .and_then(|r| r.id.clone())
+                .ok_or_else(|| {
+                    ApiError::internal(
+                        "cannot commit an all-reads Cypher transaction on an uncommitted ledger",
+                    )
+                })?;
```
zonotope: This is inconsistent with the rest of the system, and a regression. `create_ledger` just registers a new ledger with the nameservice but doesn't write a genesis commit and its `commit_head_id` is `None`. Users can then add data after the fact with no error. Plus, before this pr an all reads commit here just returned `Ok(0)` instead of erroring. I think it would be more consistent to go back to that behavior and allow commits on such a ledger after only reads to just return what the autocommit noop paths return with `t == 0` instead of erroring.

zonotope (approved):
One suggestion, but lgm

fluree-db-api/src/cypher_txn.rs
bplatz: Thanks for the catch, addressed in 0e3cd7442f7902aa4ff701ce93c1dab6fb9fae13

PR #1523, opened 2026-07-19
test(bench): re-bless at the audit head + resilience hermetics + SWITCHES.md regeneration (audit Tier-1/hygiene)

aaj3f:
## PR-HARNESS — re-bless at the audit head + resilience hermetics + SWITCHES.md regeneration

Final PR (4 of 4) of the big-iceberg-audit implementation phase (DEC-001). Restores
**harness honesty**: the perf baseline is re-blessed at the true integrated head (the live
verification of record), the one uncovered resilience path gets a regression test, every
mechanism has a documented kill switch, and the vbench bless path can no longer pin a
timeout — or an abort — as a budget.

Leaf on `perf/audit-coverage` (#1522) → `perf/audit-mem-guards` (#1521) →
`perf/audit-tier012` (merge-up #1508) → the audit line. Links:
`audit-2026-07/00-MASTER-AUDIT.md`, `audit-2026-07/decisions/DEC-001-pr-bundling.md`,
`audit-2026-07/C2-bench-wave1.md`.

### Forest map — what this closes

| Audit finding | This PR |
|---|---|
| **F-AUD-18** perf baseline stale, gate blind on 62% (28 members @180000 + 2 @120000, DNF-caps-as-budgets) | Never-bless-a-non-completion guard (`is_unblessable_wall`) + full-corpus re-bless (native + LIVE virtual, spliced) |
| **F-AUD-19** kill-switch hygiene (E1 unswitched; SWITCHES.md omits ~8 levers) | Retro-switch E1 (`FLUREE_R2RML_SHARED_PREDICATE_FUSION`); full SWITCHES.md regeneration; document W4-1/W4-1b exemptions |
| **F-AUD-20** resilience zero regression coverage | C3 mid-sweep-cancel drain test (the one gap; R3-B/PR-8-429/C2-info already covered by #1521/#1522) |
| **R-1522 rider** | `count_shortcut_eligible` predicate + decline test — **landed upstream as `3ece996fd`**; my redundant commit dropped on rebase |

### THE LIVE GATE — verification of record for the whole stack

`vbench compare --run <merged> --gate` → **77 records, 0 hash mismatch(es), 0 perf
violation(s), exit 0.** Correctness passes end-to-end at the final integrated head
(#1521 mem-fix + #1522 coverage + this PR). The drift-set 300% overrides absorbed the
loadTable-GET network variance. **Accepted residual:** those 4×-baseline (300%-over)
[...]

PR #1528, opened 2026-07-21
perf(virtual): big-Iceberg-audit implementation — correctness/capacity guards, coverage widenings, harness re-bless, browse-parity, family-C (consolidates the audit program + #1525)

aaj3f:
perf(virtual): big-Iceberg-audit implementation — correctness/capacity guards, coverage widenings, harness re-bless, browse-parity, family-C (consolidates the audit program + #1525)

This is one of five surviving pull requests in the consolidated virtual-dataset review program, which collapses the #1475–#1529 lineage into five reviewable units at the maintainers' request. #1528 carries the big-Iceberg-audit implementation: the memory-budget guard, the Tier-2 coverage widenings, the harness re-bless and switch registry, the browse-parity package, and the family-C filter-join at the tip — together with the retraction-resurrection merge-up and the merge-on-read cache-arm guard that live on its integration base. It is the deepest and widest survivor, and it carries the one native-observable edit in the whole program that no switch can revert; that edit is foregrounded in the native-impact appendix below. Reading order across the program: #1520 (standalone MoR correctness fix) → #1507 (performance program) → #1514 (serverless Lambda-usability + BYO-IAM) → **#1528 (this PR)** → #1529 (native-twin materialize builder). The outward-facing audit record is on the `docs/virtual-dataset-audit-2026-07` branch — start with [00-MASTER-AUDIT.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/00-MASTER-AUDIT.md) §6 (the recommendations ladder) and the implementation reviews in [pr-reviews-impl.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/pr-reviews-impl.md).

## Original PR body

_Scope expanded by consolidation; the tip PR's own change (the family-C filter-join) is the narrative below. The four audit PRs plus the retraction-resurrection merge-up and the MoR cache-arm guard folded in are rolled up under "Consolidated scope."_

## Forest map

Two deployed production queries on `enterprise-byo-test-10` DNF (>180s) on the virtual (R2RML/Iceberg) dataset. P4 ([parity/P4-famc-probe.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/parity/P4-famc-probe.md)) traced BOTH to one line — the join-path blanket filter decline in `fused_aggregate.rs::resolve_join_at_open` — and DEC-003 §1/§7 authorized the fix (AJ, 2026-07-21). This PR lifts that guard and ports the single-table `FilterPlan` machinery to the join path. All the join fold, F1 combine, and E2/W4-2 machinery already exist beneath (base `perf/browse-parity`); this is the finite widening P4 scoped.

The two production shapes (adapted to the SF0.1 bench star, verbatim from P4):

```sparql
# Q1 / q086 — open tickets by segment (FACT-SIDE INEQUALITY FILTER). Measured virtual: ~10.3s (declined→materialize).
SELECT ?segment (COUNT(?ticket) AS ?openTickets)
WHERE { ?ticket a edw:SupportTicket ; edw:status ?status ; edw:customer ?cust .
        ?cust edw:segment ?segment .
        FILTER(?status != "Closed") }
GROUP BY ?segment ORDER BY DESC(?openTickets)

# Q2 / q087 — below-reorder by category (VAR-TO-VAR FILTER, COUNT + 2×AVG). Measured virtual: ~5.0s (declined→materialize).
SELECT ?category (COUNT(?snap) AS ?belowReorderCount) (AVG(?onHand) AS ?avgOnHand) (AVG(?reorder) AS ?avgReorder)
WHERE { ?snap a edw:InventorySnapshot ; edw:onHandQty ?onHand ; edw:reorderPoint ?reorder ; edw:product ?prod .
        ?prod edw:category ?category .
        FILTER(?onHand < ?reorder) }
GROUP BY ?category ORDER BY DESC(?belowReorderCount)
```

## Mechanism

`route_filter_source` routes the FILTER to the single chain pattern that owns EVERY referenced variable:
- **FACT-owned** → carried in `Resolved.filter`, applied per fact row in `next_batch`. This is the exact machinery the single-table path already uses, extracted into the shared `build_filter_plan` so a filtered join is byte-parity with the single-table fold. (The fact row-validity already null-drops the filter's member columns before the filter runs.)
- **TERMINAL-dim-owned** → applied per dim row during the FK→GKey map build (`row_passes_filter_plan`), so a fact row probing a filtered-out dim key drops — equivalent to applying the residual post-join because the dim attrs are functionally determined by the dim PK.

Q2's COUNT + two AVG folds need no new aggregate machinery — `resolve_agg_folds` already folds several from the fact scan; only the filter port is new.

## Semantics — the D-c5 bar (a wrong count is the one unacceptable outcome)

The fused filter matches the MATERIALIZED `FilterOperator` by construction: the materialized operator (`filter.rs`), `passes_filters` (`eval.rs`), and the fused `FilterPlan` all evaluate the SAME `PreparedBoolExpression` through `eval_to_bool_non_strict`, which demotes a can-demote expression error to `false` (row excluded).

A NULL/absent column value in a comparison is an unbound BGP member ⇒ the row is **EXCLUDED** (e.g. `?status != "Closed"` with NULL status does NOT count as "not Closed"; `?onHand < ?reorder` with a NULL on either side is excluded). On the fact side this is enforced by the R2RML star row-validity (`validity_cols`), which fires **before** the filter — fact filter vars are always fact object vars, so a null member drops the row first (BGP parity, identical to single-table). On the dim side `row_passes_filter_plan` enforces the same null-drop inline (a null column short-circuits to `false` before the boolean eval), because the dim scan does not otherwise null-check a non-group filter column — this also keeps `!BOUND`/COALESCE-style filters sound. **Semantics-divergence check (task item 2): none.** The single-table plan already matches the materialized operator (same primitive), so it is reused verbatim; nothing to report.

Tested: `family_c_row_passes_filter_plan_null_excludes_and_compares` exercises `!=`-with-NULL and var-to-var with a NULL on each side with null-bearing hermetic fixtures (SF01 has no null STATUS/onHand rows, so the corpus members cover the non-null path and the unit tests cover the null path).

## Admission / decline enumeration (each a soundness line, all in code)

ADMIT: a filter whose vars all resolve to scalar columns on ONE pattern that is the FACT or the TERMINAL dim.

DECLINE (`Ok(None)` → materialize): a var-free/constant filter; a var bound as an object on ≥2 patterns (cross-source value-equality) or a filter spanning ≥2 patterns (fact AND dim, or two dims); an INTERIOR-dim owner (v1, symmetric with the interior-dim group-key decline); a `RefObjectMap` FK / template / constant / multi-valued object var (`build_filter_plan` → `object_map_for_var`); EXISTS/NOT-EXISTS/subquery filters and any non-bare-FILTER shape (detection only captures a single `Pattern::Filter(expr)`).

## Switch

Rides `FLUREE_FUSED_R2RML_AGG_JOIN` as a widening (program switch policy: widenings ride their family switch; OFF reverts). With the join sub-switch OFF, `resolve_join_at_open` is never reached, so a filtered join reverts to the materialize path.

## Gates (all local; CI does not fire on a non-main base) — record in `docs/audit-impl/family-c-gates.md`

- fmt: my files clean (`fused_aggregate.rs`, `corpus.rs`); the pre-existing `hash_join.rs` drift untouched.
- clippy `-p fluree-db-query --all-targets --no-deps`: 0 new lints (only the 5 pre-existing doc-list warnings at `fused_aggregate.rs:112-116`).
- `cargo test -p fluree-db-query`: 1319 lib (+4 family-C tests) + all integration bins, 0 failed.
- `cargo test -p fluree-bench-virtual --bins`: 36 passed (corpus validates at 87 members; smoke covers all tags).
- no-native featureset: `cargo check -p fluree-db-api --no-default-features --features aws,iceberg,shacl` → PASS.
- native oracles blessed offline (native-sf01, no creds): q086 hash `69006c4ffd39…` (3 rows), q087 hash `d7aba410bb32…` (10 rows) — byte-identical to P4's measured native.

### LIVE (virtual-sf01, single-flight, 1 rep, PAT in-memory) — `compare --gate`: 4 records, **0 hash mismatches, 0 perf violations**

| q | shape | before | fused wall | rows | hash == native oracle |
|---|-------|--------|-----------|------|-----------------------|
| q086 | Q1 fact-side `!=` FILTER | ~10.3s materialize | **258 ms** (~40×) | 3 | yes |
| q087 | Q2 var-to-var FILTER, COUNT+2×AVG | ~5.0s materialize | **158 ms** (~32×) | 10 | yes |
| q065 | regression: join+flag (no FILTER) | fused | 310 ms | 3 | yes (unchanged) |
| q038 | regression: F1 filtered-COUNT | fused | 28 ms | 1 | yes (unchanged) |

Both new members collapse from seconds to sub-second AND hash-match the native materialized oracle byte-for-byte (q087's AVG decimals included) — the fused fold fired and is exactly correct. The two regression sentinels are unchanged.

## Residuals (still decline → materialize)

Interior-dim filters (v1); filters spanning fact AND dim (or two dims); filters over a `RefObjectMap` FK / non-scalar object; EXISTS/subquery filters. Each is a sound decline (falls back to the materialize path), not a wrong answer.

## Consolidated scope

The four audit PRs below stack in order onto an integration base branch that begins with a merge of `main` (bringing the retraction-resurrection fix) and then re-applies the merge-on-read cache-arm guard; the family-C filter-join is the tip. Each superseded PR closes as consolidated-at-maintainer-request, its pre-consolidation tip preserved under `archive/pre-refactor-2026-07-21/<branch>`. The merge-up and cache-arm guard never had their own PR and are anchored by the tags `archive/pre-refactor-2026-07-21/perf_audit-tier012` and `archive/pre-refactor-2026-07-21/audit_mor-cache-arm`.

| Item | What it did |
|---|---|
| `main` merge-up (#1508 fix) | The integration base's first commit merges `main`, bringing the F-AUD-2 fix for retraction-resurrection (a hard/CoW delete resurrected by a re-index in an un-indexed window). Folded in via the base branch, not a separate PR. |
| MoR cache-arm guard | Re-applies the fail-closed merge-on-read guard on the cached scan-files hit-arms that the stacked perf line adds around `plan_scan`, with a delete-flag on the cache entry and a `CACHE_FORMAT_VERSION` 2→3 bump so pre-guard entries miss and recompute. This completes the #1520 guard on the stacked path. Folded in via the base branch, not a separate PR. |
| #1521 | Closes the memory-budget blind spot on the non-aggregate scan path (F-AUD-3): a saturating `release` decrement primitive, a per-query budget divisor, and scan-window plus fact-as-parent memory accounting, so a wide crawl or point-lookup that previously OOMed now aborts typed. |
| #1522 | Eight Tier-2 coverage widenings — scan-side (FILTER-IN / VALUES pushdown, ASC scan-side top-k, timestamp manifest pushdown, OPTIONAL row-budget forwarding, parallel range GETs) and fold-side (MIN/MAX fused admission, filtered-COUNT admission, `/info` native+virtual member routing) — each an admission gap that previously fell to full per-row materialization. |
| #1523 | Re-blesses the perf baseline at the true integrated head, adds the one uncovered resilience hermetic (mid-sweep cancel), retro-switches the shared-predicate fusion, and regenerates the switch registry — restoring harness honesty (the bless path can no longer pin a timeout or an abort as a budget). |
| #1525 | The browse-parity package: native-shape crawl node-document output (FK refs as `{"@id": …}`, typed-json honored, booleans as real bools), constant-IRI select-map routing onto the bound-subject machinery, a loud refusal for a dropped VALUES clause, a typed `err:r2rml/UnsupportedPattern` refusal envelope, property-var crawl budget forwarding, an unmapped-class short-circuit, the cfg-orphan build fix, and the F1 constrained-COUNT fusion. |
| #1528 | The family-C filter-join (this PR's tip): admit row-level FILTERs to the fused join aggregate — the production DNF class. |

## Verification of record

CI does not fire on a non-main base, so the verification of record is the final live gate plus the six independent adversarial reviews.

The whole-stack live gate at the final integrated head reported **77 records, 0 hash mismatches, 0 perf violations** (`vbench compare --run <merged> --gate`, from the re-bless in #1523) — correctness passes end-to-end. The native re-bless went 54→77 entries with all 74 oracles reproduced exactly (no native regression), and the harness bless guard now refuses to pin a DNF or an abort as a budget. The stale-baseline analysis and the run JSONLs are in [C2-bench-wave1.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C2-bench-wave1.md) and [data/](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/data/virtual-rebless-merged.jsonl).

The six adversarial diff reviews are recorded in [pr-reviews-impl.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/pr-reviews-impl.md): R-1521 (memory-budget guard) verdict SHIP with every residual bounded, documented, and in the over-count-safe direction; R-CACHE-ARM (the perf-line cache-bypass fix) verdict CLOSED — the guard is now complete on the stacked path as it already was on `main`; R-1522 (the eight widenings) verdict SHIP with no blocking finding, every mechanism admission-only and the in-engine FILTER/SortOperator authoritative; R-1523 (the harness/trust layer) verdict SHIP — the gate can no longer pin a timeout or abort as a budget; R-PARITY (#1525 browse-parity) verdict SHIP with the native-untouched rule holding for every cluster; and R-1528 (family-C) verdict SHIP — null semantics sound on all four attack sub-points, no admitted-and-wrong shape. Family-C is live-verified: q086 and q087 collapse from seconds to sub-second and hash-match the native materialized oracle byte-for-byte.

The correctness findings these PRs implement are verified independently: the merge-on-read gap in [V1-mor-verification.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/V1-mor-verification.md), the memory-budget blind spot in [V2-membudget-verification.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/V2-membudget-verification.md), and the browse-parity serialization and shape chains in [parity/P1-serializer-verify.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/parity/P1-serializer-verify.md), [parity/P3-shape-matrix-empirical.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/parity/P3-shape-matrix-empirical.md), and the two production-DNF shapes in [parity/P4-famc-probe.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/parity/P4-famc-probe.md). The coverage gaps each widening closes were reproduced live in [C2b-bench-wave2-probes.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C2b-bench-wave2-probes.md).

## Native-impact appendix

Foregrounded first: the one edit in the whole program that a switch cannot revert.

- **`BINDING_EST_BYTES` 64 → `size_of::<Binding>()` = 88 in `context.rs` is shared/native and NOT switch-revertible.** After this change, native fold and join memory accounting counts 88 bytes per binding regardless of `FLUREE_SCAN_MEM_ACCOUNTING`. It is defended as conservative — over-count, fail-late-not-wrong — but it is not proven byte-neutral: a native query under memory pressure aborts marginally earlier than before. It is verified only by the compile-time canary `binding_est_bytes_is_at_least_binding_stack_size` (the constant is `>= size_of::<Binding>()`), with no native corpus leg. This is the single native-observable edit in the whole program that no switch can revert, and reviewers should treat it as such.
- **#1521 extends the R3 memory surface, all native-safe.** A saturating `release` decrement primitive is added to `fluree-db-core/src/cancellation.rs` (git-confirmed 58 insertions; inert for native — only the R2RML window-release pairing calls it). A budget-share divisor lands in `execute/runner.rs` (`FLUREE_QUERY_BUDGET_SHARE_DIV`, default **1** = full budget per query = today's behavior, gated on `div>1 && memory_limit().is_none()` so it is a no-op for native). Scan-window and parent-build accounting land in the R2RML `operator.rs` (gated `FLUREE_SCAN_MEM_ACCOUNTING`, default on, R2RML-scan-only; off is byte-identical except that the unconditional pull-loop `check_cancelled` is upgraded to `checkpoint`, whose outcome is byte-identical).
- **#1522 OPTIONAL `set_row_budget` forwarding — a third generic-operator budget extension.** It forwards the LIMIT budget on the generic `OptionalOperator` to the required/outer side only. Switch `FLUREE_R2RML_BUDGET_OPTIONAL` (default on); off swallows the budget and is byte-identical. MINUS, FILTER, SORT, DISTINCT, HAVING, and subquery explicitly decline (documented, observable spans) — MINUS declines because its anti-join output can under-produce under a finite primary budget.
- **#1522 item-14 / #1525 `/info` member-routing is a genuinely native-reachable path (gated).** `merge_virtual_into_native` in `ledger_info.rs` sits on the native-reachable `/info` path; it unions classes and properties by IRI (a graph-source member replaces a native collision, never sums; it keeps the top-level native `stats.flakes`/`size`). Switch `FLUREE_R2RML_INFO_MEMBER_ROUTING` (default on); off restores the pre-mechanism behavior byte-identically. Minor caveat: an IRI-form mismatch would double-LIST (not double-count). The MoR rider on this path flags delete-bearing tables into an `mor-approximate-tables` list (upper-bound counts) — this is where #1520's deferred `/info` over-count is closed.
- **#1525 Cluster-A crawl formatter — one shared default-path delta: boolean rendering.** `format_node_object_binding` has four callers, all on the graph-source crawl path, so no native or hydration path reaches it. But the shared JSON-LD default-format literal arm (`format/mod.rs`) is touched: `xsd:boolean` strings now coerce to real bools (the D3 fix). That boolean rendering is the sole shared/default-path behavioral delta; every non-boolean literal is untouched (`coerce_bool_literal` returns `None`), and it is regression-tested (`crawl_boolean_literal_renders_as_bool`).
- **#1525 cfg-orphan fix is a positive native-surface fix.** It restores `build()`'s `#[cfg(native)]` gating and the correct feature-gating on `with_secret_resolver` that the #1505 SecretRef insertion had dropped — build-time feature-gating only, no runtime path, restoring no-native builds. It adds a new permanent gate, `cargo check -p fluree-db-api --no-default-features --features aws,iceberg,shacl`, which CI should adopt.
- **The remaining edits cannot fire for native.** #1525's canon key-sort is bench/test-harness only (zero production impact; element order preserved). #1525's planner FULL_SCAN estimate (E1) is confined to the `Pattern::R2rml` arm — native produces `Pattern::Triple`, so it is never reached, and it is gated. Every fold widening (#1522 MIN/MAX and filtered-COUNT, #1525 F1, and this PR's family-C filter-join) lives in `fused_aggregate.rs`, which returns `None` the instant any non-R2rml pattern is present — it cannot fire for native. The review confirmed this PR's own diff is only `fused_aggregate.rs` (+625), reusing the shared `eval_to_bool_non_strict` / `can_demote_in_expression` without modifying them. #1523 is bench tooling and docs; its native re-bless reproduced all 74 oracles exactly.

Confirmed negatives across the span (git-checked, name-only diff clean): no `arrow` / `datafusion` / `parquet` dependency bump, and no `.github/workflows` or `rust-toolchain` edits.

## Landing and hand-off notes

Land #1520 first. #1520 and this line both carry the fail-closed merge-on-read guard; the added and removed lines are byte-identical (only the git patch-id differs, because the two were cut against different base trees). Once #1520 merges to `main`, this PR's guard portion resolves as an identical-content merge and the cache-arm guard (`CACHE_FORMAT_VERSION` 2→3) rides on top cleanly. A reviewer diffing this PR will see the `main` merge-up as a large block on the integration base; it is the retraction-resurrection fix, folded in via the base branch and anchored by the archive tags named in the consolidated-scope roll-up, not lost work.


fluree-db-query/src/r2rml/fused_aggregate.rs
```diff
@@ -704,9 +807,20 @@ fn insert_dim_gkeys(
...
+    // The FK→GKey map assumes the parent JOIN KEY is UNIQUE — true for a proper
+    // star schema (the RefObjectMap's parent columns are the dim's surrogate PK),
+    // but NOT guaranteed for a hand-written mapping whose parent columns are a
+    // non-key subset. A duplicate parent join key means the materialized inner
+    // join FANS OUT (one fact row matches multiple dim rows), which a single-entry
+    // per-key map cannot represent: conflicting group attrs would mis-attribute
+    // (last-wins), and even EQUAL group attrs would UNDER-COUNT the fan-out. So any
+    // duplicate parent key DECLINES the fused plan (caller returns `Ok(None)` →
+    // materialize), the conservative posture the whole operator takes when a shape
+    // is outside what it can fold soundly. Returns `false` on any duplicate (was
```
bplatz, line 820: **Definite issue (sequencing, not code)** — this fix is correct, and it's four PRs downstream of the bug it fixes.

This is exactly the defect I raised on #1507: the equal-value duplicate was a latent fan-out under-count, and declining on *any* duplicate parent key is the right resolution. The comment even documents the reasoning better than I did.

The problem is placement. Under the stated merge order (#1520 → #1507 → #1514 → #1528), `main` carries the under-count for the duration of three merges, and #1514 *widens* fused-aggregate admission on top of it (C5/E1/E2/W4-2), so the exposure grows before it shrinks. The same is true of the cache-arm guard in this PR, which fixes the MoR override-leak I raised on #1507's persisted scan-files cache.

Both are small, self-contained commits. Please cherry-pick them onto #1507 — the `insert_dim_gkeys` change is a one-line behavior flip plus the comment, and the cache-arm work is the delete flag + `guard_cached_scan_files` + the `CACHE_FORMAT_VERSION` bump. That way no intermediate state of `main` ships a known silent-wrong-answer, and this PR keeps only the work that's genuinely its own.

If cherry-picking is awkward against the branch topology, the alternative is to land #1528 ahead of #1514 — but forward-porting the two fixes is cleaner and lets #1507 be correct on its own merits.

fluree-db-iceberg/src/mor_guard.rs
```diff
@@ -0,0 +1,259 @@
...
+}
+
+/// Whether the snapshot **summary** counters report any merge-on-read delete
+/// files (`total-delete-files` / `-position-deletes` / `-equality-deletes` > 0).
+/// Zero extra I/O — the summary is already in memory. Used both by
+/// [`ensure_no_summary_deletes`] and by the planner to stamp the resulting
+/// [`crate::scan::ScanPlan::has_delete_manifests`] flag so cached scan-file
+/// selections carry the signal downstream.
+pub fn summary_indicates_deletes(snapshot: &Snapshot) -> bool {
+    snapshot.total_delete_files().unwrap_or(0) > 0
```
bplatz: **Worth a look** — this fails open on a malformed counter, and `/info` is now the one consumer with no backstop.

`total_delete_files()` and friends are `.parse().ok()`, so a present-but-unparseable counter becomes `None` → `unwrap_or(0)` → `> 0` is false → "no deletes." On the planners that's fine, because the manifest-list count is OR'd in right after. But `fetch_virtual_table_row_counts` calls this on its own, and that path is metadata-only by design — there is no manifest list to fall back to.

So the failure mode is a table that silently *doesn't* appear in `mor-approximate-tables` and has its over-counted row total presented as exact. That's a quieter version of the same class this whole thread is about: the operator gets a number with no signal that it's an upper bound.

Cheapest fix is at the accessors — treat present-but-unparseable as "deletes present" rather than absent, so the summary arm fails closed on its own and every consumer inherits it. Absent stays `0`, which is correct.

fluree-db-core/src/cancellation.rs
```diff
@@ -152,15 +155,37 @@ impl QueryCancellation {
...
+    /// for an allocation whose drop point the caller can prove — e.g. a scan window
+    /// that has been handed off downstream and is about to drop, so a streaming scan
+    /// of N sequentially-freed windows accounts only the resident one rather than
+    /// their all-time sum. Saturating, so a double- or over-release can never
+    /// underflow the counter. It is a CALLER INVARIANT that `release` is paired only
+    /// with a matching non-persistent `record_alloc`: persistent buffers (join /
+    /// aggregate builds, retained lookups) must never be released, or the guard would
+    /// under-count live memory. No-op on a disabled handle.
+    #[inline]
+    pub fn release(&self, bytes: usize) {
```
bplatz, line 179: **Worth a look** — the mechanism is right, but it doesn't reach the case that most needs it.

Saturating so it can't underflow, and the caller invariant is stated clearly. It currently has one caller (`r2rml/operator.rs:1507`, scan windows), which fixes streaming-window accumulation.

What it deliberately excludes — persistent join/aggregate buffers — is where I think the real over-count is. I've raised this on #1514: `GraphOperator` builds a fresh inner subplan per parent batch (`graph.rs:357`, `:536`), and batched-OPTIONAL rebuilds similarly, so a hash join inside one re-records its whole build on every rebuild while the previous is already dropped. The counter grows linearly in driving batches against a flat true peak.

A rebuilt build table isn't really a "persistent buffer" — its drop point is exactly as provable as a scan window's, it's just at inner-subplan teardown rather than batch handoff. If the rebuild sites released on teardown, this mechanism would cover it without weakening the invariant for genuinely retained buffers. Worth deciding whether that's in scope here or a follow-up, but I don't think it should stay unaddressed — it's a false-abort path on native queries with the guard default-on.

bplatz (changes requested):
The code here is the strongest in the program — my only substantive ask is about **where these fixes live**, not what they do.

All three headline items verified correct:

- **Cache-arm guard (R-CACHE-ARM).** `has_delete_manifests` is the OR of *both* signals (`summary_indicates_deletes(snapshot) || delete_manifests > 0`) in both planners — I specifically went looking for a summary-only stamp, which would have let an override-cached delete-bearing selection come back with the flag `false`. It isn't there. `guard_cached_scan_files` is called from both hit arms (`r2rml.rs:1916` in-memory, `:1950` disk) and unit-tested on both sides.
- **`/info` `mor-approximate-tables`.** Using the zero-I/O summary counters is the right mechanism — I'd flagged on #1520 that this path is structurally metadata-only and can't reach the manifest list, so a guard call wouldn't have worked. This does the reachable thing instead.
- **ASC top-k (F-AUD-6).** The soundness gate is real: `r2rml.rs:2062` filters on `!tk.ascending || field.required`, and a declined directive falls through to the parallel scan rather than proceeding. The worst-first `OrdKey` reorientation is correct for both directions.

**The ask:** two of the defects this PR fixes are live in #1507, which merges three PRs earlier. Detail inline.

Two `worth a look` items inline as well.

fluree-db-query/src/r2rml/fused_aggregate.rs
aaj3f: Done, with one compile-reality adjustment to the exact mechanics. (a) The `insert_dim_gkeys` decline is forward-ported to #1507 (0b4269a67), byte-identical with this branch's version down to the in-body comment, so nothing diverges at merge. (b) The cache-arm work could not ride down as-is: #1507's rung predates the MoR guard entirely (no `mor_guard.rs`, no override — those arrive with #1520 off newer `main`), so the cherry-pick does not compile there. #1507 instead received a guard-independent equivalent (0247c7934): same `PersistedScanFiles` schema and version bump byte-identical to this branch's, manifest-list-only detection mirroring its COUNT path, and hit arms that decline flagged entries outright. That closes the override-persistence leak standalone at #1507 — no intermediate `main` ships it in any order — and this branch's override-aware `guard_cached_scan_files` remains the final form, superseding the conservative arm at merge. The convergence merge on this branch (8effa91e3) reconciles the shared regions so that supersession is an auto-resolve rather than a maintainer conflict.

fluree-db-iceberg/src/mor_guard.rs
aaj3f: Confirmed and fixed at the source PR as suggested: #1520's guard now fails closed on present-but-unparseable (and negative) counters at the guard level (83c13d9b7 there), and this branch's copy is reconciled to the same semantics in 8effa91e3 — with the placement nuance that the flip lives in the shared predicate rather than the raw `Option<i64>` accessors, so `/info` inherits "deletes-present" for a malformed counter and lists the table as approximate instead of silently presenting an over-count as exact.

fluree-db-core/src/cancellation.rs
aaj3f: Agreed that it shouldn't stay unaddressed, and it hasn't: fixed at #1514 (5b331043b), which is where the counter itself lives — rebuild-boundary releases (snapshot before inner open, release the delta after drain) at both GraphOperator sites and the batched-OPTIONAL driver, using this branch's release primitive ported byte-identically so the two rungs converge. Your framing that a rebuilt build table's drop point is exactly as provable as a scan window's is the argument the site comments now make. Regression test included; the false-abort repro it pins is recorded in the PR #1514 thread.

aaj3f:
The placement ask is executed (with the B2 compile-reality adjustment detailed inline — the guard doesn't exist at #1507's rung, so it got a guard-independent equivalent instead of the literal cherry-pick), both worth-a-look items are fixed at their source rungs (#1520 fail-closed counters, #1514 rebuild-boundary releases), and this branch's convergence merge (8effa91e3) reconciles every shared region — dup-key decline, cache schema, hasher, guard copy, release sites — so the stack's remaining merges auto-resolve. Gates re-run green at the new head.

bplatz (approved):
All three items resolved: the two fixes now live at #1507/#1514 with this branch containing them via the convergence merges (supersession auto-resolves — verified `insert_dim_gkeys` and `summary_delete_evidence` hash-identical across rungs), and the fail-closed counter semantics reach every consumer — the fail-open accessors have zero remaining production call sites, and the /info test pins the malformed-counter scenario. mor_guard, /info, iceberg (250+), and query (1323) suites all pass locally.

One non-blocking nit: the new `smoke_minio` example is missing a `[[example]] required-features = ["aws"]` stanza in `fluree-db-iceberg/Cargo.toml` (its two siblings have one), so `cargo test -p fluree-db-iceberg` at default features fails to compile the example target — and CI can't see it since it runs `--all-features`. Fine to fold into a later commit.

PR #1422, opened 2026-07-02
feat: materialize Iceberg/R2RML graph sources into native ledgers + tracking worker

christophediprima:
## Summary

Adds the ability to **materialize an Iceberg / R2RML graph source into a native Fluree ledger** and
keep it fresh with a **background tracking worker**. Querying a graph source directly re-reads and
re-joins the raw Iceberg rows on every request; materializing into a native ledger gives deduped,
indexed, time-travelable state that queries at native speed and refreshes incrementally.

Five commits:

1. **Append-only incremental scan + snapshot-window helpers** (`fluree-db-iceberg`).
2. **R2RML incremental scan + snapshot helpers** (`fluree-db-api`).
3. **Materialize into native ledgers + tracking worker + endpoints** (`fluree-db-api`, `fluree-db-server`).
4. **Tombstone deletion + latest-by-key dedup** for materialization.
5. **Refreshable Google metadata-server catalog auth** — so long-running tracking against a Google
   Iceberg REST catalog keeps authenticating past the ~1h a static bearer lasts.

Builds on the existing Iceberg graph-source support on `main`.

## Motivation

A graph source exposes an Iceberg table as read-on-query triples. That's ideal for ad-hoc access,
but not for a dataset you query repeatedly:

- **No dedup.** Iceberg tables written by streaming/CDC pipelines are **append-only** — many row
  versions per entity. A graph source surfaces all of them; there's no "current state" of a subject.
- **Re-read + re-join every query.** Multi-pattern queries (`?s a X ; p1 ?a ; p2 ?b …`) join the raw
  append-only rows on each request, which blows up combinatorially and re-does the scan every time.
- **No native indexes / time travel** over the sourced data.

Materializing solves all three: collapse the append-only rows to **one node per subject** (upsert),
write them into a native ledger (native indexes, SPARQL, history), and **refresh incrementally** off
the source's snapshot log. The tracking worker automates the refresh.

## Changes

- **`fluree-db-iceberg` — incremental scan + snapshot windows** (`scan/planner.rs`,
  `scan/send_planner.rs`, `metadata/table.rs`, `config.rs`): plan a scan over only the data files
  **added between two snapshots** (append-only incremental), plus helpers to resolve the current
  snapshot and the delta from a watermark. Falls back to a full scan when an incremental window
  isn't safe (e.g. a non-append snapshot). + unit tests.

- **`fluree-db-api` — R2RML materialization** (`graph_source/r2rml_materialize.rs`): read
[...]

PR #1529, opened 2026-07-21
feat(materialize): fluree materialize — native-twin builder from virtual R2RML/Iceberg datasets

aaj3f:
feat(materialize): fluree materialize — native-twin builder from virtual R2RML/Iceberg datasets

This is one of five surviving pull requests in the consolidated virtual-dataset review program, which collapses the #1475–#1529 lineage into five reviewable units at the maintainers' request. #1529 implements `fluree materialize`: a bulk builder that turns a virtual R2RML-over-Iceberg graph source into a native Fluree ledger ("twin"), verifies it against the source, and writes it as a pack or a local ledger. It is the last member of the reading order. Since the consolidated body below was written, the previously-open live gate has been RUN AND PASSED (an independent full-triple diff against a natively-imported ledger at full 35.2M scale came back identical — see the acceptance section directly below), and a measured performance fix slate has landed on the branch; the sections below the acceptance report are the original consolidated body, kept intact for review continuity. Reading order across the program: #1520 (standalone MoR correctness fix) → #1507 (performance program) → #1514 (serverless Lambda-usability + BYO-IAM) → #1528 (big-Iceberg-audit implementation) → **#1529 (this PR)**. The outward-facing audit record is on the `docs/virtual-dataset-audit-2026-07` branch — the fact base for materializing Iceberg into native ledgers is in [C3-materialization-facts.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C3-materialization-facts.md) and the strategy framing (with the snapshot-keyed correctness lemma) is in [D1-strategy-options.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/D1-strategy-options.md).

## 2026-07-29 addendum — performance fix slate + full-scale acceptance (commits `74780fd25`…`476d08192`)

A stage-by-stage performance audit of this builder against the repo's own bulk-import path (`fluree create --from`) found that bounded-by-design had been conflated with slow-by-implementation, and a measured fix slate landed here as eight commits on this branch (per the maintainers' one-PR consolidation rule). The audit's baseline, at equal bounds (`--parallelism 2 --memory-budget-mb 1024`, the 35,238,778-triple sf01 dataset over a Snowflake-managed REST catalog): build+publish 348.7s, and a verify phase that double-buffered the whole graph in RAM even in quick mode — 16.5GB RSS at 6.1M-triple dims scale, projected 95–155GB at 35M, i.e. certain OOM on the build host.

The slate, in commit order on this branch: a merge-up of the consolidated program base (`74780fd25`); produce-phase instrumentation spans (`5e429a9fd`); **O2 — memory-bounded verify in both modes** plus decimal value-canonicalization in the parity compare (`9649c3e4d`); **O5 — concurrent table pin + Pass-1 FK pre-index** bounded by `--parallelism` (`2caec9194`); **O6 — budget honesty**: sub-2GB budgets now scale the chunk size instead of silently flooring at 128MB, and the FK parent index is charged against the budget with a typed `ParentIndexBudgetExceeded` instead of living outside it (`03787651a`); **O4 — per-row allocation cleanup** in the enumerator hot path, byte-identical across the r2rml exact-output suite (`0e6cf38bb`); **O1 — a parallel produce side**: `--parallelism` sync worker threads render/intern/encode with private sinks while the non-`Send` Iceberg scans stay on the async driver, with an in-commit differential test asserting p=1 and p=2 produce identical triple sets, contiguous chunk indices, and exactly one stamped final chunk (`e77bebc15`); **O3a — the `OwnedTriple` intermediate buffer removed**, terms interned directly from the enumerator into the chunk sink (`06e00004f`); and **a dateTime value-canonicalization fix** in the verify compare (`476d08192`, receipt below). Review-response commits (bplatz, 2026-07-31): `95389a66f` + `d1a5d0904` + `82b44631a` — full-verify rewritten as a single linear index pass (the O(n²) OFFSET paging is gone), verify spool moved off possibly-tmpfs `/tmp` to under `.fluree` with a `--tmp-dir` override, the 64KiB stamp cap now enforced fail-loud (`StampTooLarge`), `read_stamp` requires the `MATERIALIZE_NS` namespace (foreign-ns stamp forgeries rejected), post-O3a dead code removed, `--verify` help states quick-mode sample strength, the MoR env constant imported rather than hand-copied, and `docs/cli/materialize.md` + SUMMARY + graph-sources docs added per the repo docs mandate. Final review commit `5602938b4` — **duplicate parent join keys**: `fluree materialize` now DECLINES by default a source whose FK parent join key maps to more than one parent (typed error naming table/join-column/count), overridable with `--allow-duplicate-parent-keys`. Rationale: the prior last-wins pick was run-to-run NONDETERMINISTIC — the parent index consumes `scan_table`'s `buffer_unordered` stream, so when a key's colliding rows span data files an I/O-completion race chose the winner, and baking one draw violated the twin's mapping-hash / stamp reproducibility contract. The fix ships a deterministic keep-min tie-break (lexicographically smallest parent subject) shared by the twin builder and the virtual query path, so both agree and both are reproducible; an override-built twin records the distinct ambiguous-key count in its stamp (distinct-key counting deliberately, since a raw collision tally would itself be order-dependent). The true RefObjectMap fan-out (one edge per matching parent) is a tracked follow-up.

**Shared-path changes** (correcting an earlier claim in this body that `ParsedChunk.txn_meta` was the only one). Two functions on the shared text-import path change: 1. `ParsedChunk.txn_meta` — self-diff byte-identical for every text import (a data chunk emits empty txn_meta; only the twin's final stamp chunk populates it). 2. `ImportConfig::effective_chunk_size_mb` (previously unlisted) — the sub-2GB branch now derives chunk size from the budget instead of always applying the 128MB floor, resizing chunks for EVERY text import (Turtle/TriG/JSON-LD) with `memory_budget_mb < 2048`. At 512MB / max_inflight=2: 128MB → ~51MB chunks (~2.5× commit count, different ledger shape, same input). Default budget ≈80% RAM, so most hosts land ≥2048MB and are unaffected; what changes is containers under ~2.5GB and any explicit `--memory-budget-mb` below 2GB (which the machine-safety posture encourages, and every finalizer-touching test uses at 256MB). Sizing only — commit contents/index identical, only chunk boundaries and count differ. 3. `build_parent_lookup` in fluree-db-query (the VIRTUAL query path) changes last-wins → deterministic keep-min on a duplicate parent join key — output-affecting ONLY on the duplicate-key shape that is nondeterministic today (an I/O-order-dependent arbitrary pick becomes a stable one); no effect on unique-key joins, which is the entire corpus. A ninth commit closes the recorded O5(c) residual (`fa96388d3`): tables are scheduled into topological waves so a non-cyclic FK parent is scanned exactly ONCE — its key→subject index is built during its own emit scan and frozen per-wave for the parallel workers — eliminating the separate pre-index pass that made cold fact tables the dominant wall; only cyclic/self-referential parents keep the pre-index fallback, and the quick gate's independent source re-scan is deliberately kept (its samples are chosen post-publish, so reusing build-pass statistics would let the build self-certify).

Measured acceptance at full 35.2M scale (record: the A1/A2/B-REMEASURE documents of the materialize-perf audit series): **verify now completes bounded where it previously OOM'd** — the entire quick-verified 35M run peaked at 1.41GB RSS (dims scale: 16.46GB → 1.24GB, 13.3× less); a clean-load warm-cache build ran 56.0s wall with encode spread across both workers (the cold-source wall remains dominated by fact tables being scanned for Pass-1 indexing, Pass-2 emission, and the verify oracle — the recorded O5(c) follow-up on this branch, not a separate PR); and the **independent oracle gate that kept this PR draft is now closed**: a full-triple streaming diff (same-renderer N-Triples export of both sides → external sort → set compare, ~1.2GB RSS) between the materialized twin and the blessed natively-imported `enterprise-sf01` ledger returned **zero differences in either direction at 35,238,778 triples** — the enumerator is proven correct against an oracle that is not itself.

Two verify-gate false-reject classes were found by that acceptance run and fixed in the parity compare's value canonicalization: decimals (stored E-notation vs the oracle's plain `f64` rendering) and dateTimes (storage pads fractional seconds to six digits, `…T00:16:42.000000Z`, while the oracle's formatter drops the fraction when zero — the gate now normalizes insignificant fractional-second zeros and zero UTC offsets on both sides). Both were fail-CLOSED failures — the gate refused correct twins and dropped them loudly, never the reverse. Known residuals, recorded here rather than deferred to other PRs: a non-zero-UTC-offset dateTime is deliberately not collapsed to its UTC instant (the schema under test is all-UTC); graph-source resolution is CWD-relative (`./.fluree`) and does not honor `FLUREE_HOME` for sources — a UX wart worth a look; and whether a first-class skip-verify mode for operators running their own independent diff deserves productizing is an open maintainer question.

## Original PR body

_Base re-pointed by consolidation onto the family-C branch beneath it; no other PR's scope folded in — this survivor remains the standalone materialize builder, original text below._

## `fluree materialize` — a native twin from a virtual (R2RML-over-Iceberg) dataset

Implements the materialization design's first deliverable: bulk-build a native Fluree ledger ("twin") from a virtual R2RML-over-Iceberg graph source, verify it against the source, and write it as a `.flpack` pack or a local ledger. This is the categorical fix identified by the materialization assessment — it ends the per-query cold-parquet storm and the admission-gate tail-chase for datasets that warrant a native copy, with guaranteed 1:1 native query semantics.

First-principles per the AJ directive — **zero anchoring to PR #1422**; no transact-replay. The build reuses the existing bulk `create --from` finalize → spool-sort → k-way-merge index → publish pipeline verbatim, pointed at the R2RML scan stream instead of a TTL parser.

Design + assessment record: [C3-materialization-facts.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C3-materialization-facts.md) (machinery + arithmetic) and [D1-strategy-options.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/D1-strategy-options.md) (the strategy space).

### What a reviewer should walk

The build is one new producer branch feeding an unchanged downstream. Read it in dependency order:

- **Enumerator (Chunk A, prior):** `fluree-db-r2rml/src/materialize/graph.rs` — the engine-free `TriplesMap → (s,p,o)` whole-graph enumerator. Composes the SAME per-term helpers the query operator uses (term/datatype fidelity by construction), dims-first FK resolution.
- **Streaming driver + adapter (Chunk B, prior):** `fluree-db-api/src/materialize.rs` — scans each table through the enumerator, encodes byte-budgeted `ParsedChunk`s via `ImportSinkObserver`, ships them into the existing finalizer channel (`ImportSource::Virtual`).
- **C2 `ac21c0711`** — per-table snapshot capture: `IcebergCatalogSession.snapshots` recorded at the single `scan_table_inner` caller of `load_table_context` (first-writer-wins), `pinned_tables(gs)`, `FlureeR2rmlProvider::build_watermark` override. `fluree-db-api/src/graph_source/{catalog_session.rs,r2rml.rs}`.
- **C3 `fdcbc11f7`** — pin-all pre-pass (all snapshots pinned up front → cross-table skew bounded to seconds) + `WatermarkStamp` assembly (watermark vector + SHA-256 mapping hash + builder version) + fail-loud empty-watermark guard.
- **D `ee7897e1e`** — persist the completion stamp in the twin's FINAL commit: `ParsedChunk.txn_meta` (empty → byte-identical for every text import) + `finalize_parsed_chunk` consume; the stamp is ns-encoded through the sink and rides the last non-empty commit via a one-chunk lookahead (completion-marker semantics: a valid twin is one whose head commit carries a parseable stamp). `fluree-db-transact/src/{import.rs,import_sink.rs}`, `fluree-db-query/src/r2rml/provider.rs`.
- **E `743f5e4ae`** — the post-build parity gate `verify_twin`: class counts + stratified per-subject sample (`quick`, always-on) and a full-triple diff (`full`), enumerator-as-oracle vs a native wildcard query, both rendered to identical N-Triples.
- **F+G `bf27d545a`** — the `fluree materialize` CLI + MoR build-time UX. `fluree-db-cli/src/commands/materialize.rs`.

### Correctness proofs (all hermetic; no live runs in this PR)

- **Engine-equivalence** — `materializer_driver_matches_engine_wildcard`: the driver's triple set is byte-identical to the query engine's `?s ?p ?o` wildcard over the same scan.
- **Queryable twin fidelity** — `materialize_builds_queryable_native_twin`: a real indexed+published twin round-trips class counts, an FK edge to the parent IRI, an `xsd:decimal` (value+datatype), and a lang tag (value+`@en`).
- **Stamp survives pack/restore** — `materialize_stamp_survives_pack_restore`: `archive_ledger → restore_ledger` preserves the head-commit stamp byte-for-byte and `read_stamp` parses it identically.
- **Parity gate** — `verify_twin_passes_on_faithful_twin` (pass, `quick`+`full`) and `verify_twin_fails_on_corrupted_twin` (an injected instance fails on both the class-count check and the full diff).

### Parity-gate value-semantics (ratified)

A parity gate asks "same graph in **value** space." `"5.00"^^xsd:decimal` and `"5"^^xsd:decimal` are one value; comparing pre-intern lexical (enumerator) against post-intern canonical (twin) is a diff artifact, not a data defect. The gate strips trailing fractional zeros for xsd numeric datatypes on **both** sides. Where lexical renderings genuinely differ for the same value, the **twin's native rendering is contract-correct by the parity direction** (native is the fixed contract) — so any client-observable decimal-lexical drift on cutover is the virtual path shedding a deviation, not the twin introducing one. The lead-run live native-sf01 full-triple diff must carry the same value-level canonicalization.

### Gate record

`fmt` clean; scoped `clippy` 0 new warnings; `cargo check -p fluree-db-api --no-default-features --features aws,iceberg,shacl` (no-native featureset) compiles; CLI compiles with and without `iceberg`. Tests: fluree-db-transact lib 303, fluree-db-query lib 116, fluree-db-api materialize lib 6, `it_graph_source_r2rml` 60 (2 ignored ext-infra), fluree-db-cli materialize helpers 5.

### Machine-safety posture

The CLI default is **co-resident-tolerant** (fixed `parallelism=2`, `memory_budget_mb=512`, never own-the-box auto-sizing); `--memory-budget-mb` / `--parallelism` raise it explicitly, `--max-performance` opts into host auto-sizing on a cleared machine. Every finalizer-touching test runs with `threads=2` + `memory_budget_mb=256`. Verification uses publish-then-verify-then-drop: the gate needs a queryable twin, so a failed gate drops the twin — nothing unverified stays announced. (Deliverable-2 rider: solo's alias repoint must sequence strictly AFTER gate-pass.)

### Residuals (deferred, tracked)

- **C-2 memory optimization** — the FK parent index stores `key → parent-subject-term` (general); the exists-set + apply-parent-template form saves memory when parent subject-template columns ⊆ the join key. Guarded by `MaterializeStats.ref_edges` per-(child,predicate) counts. Not trivial → left as a guarded-threshold note.
- **Two-stage producer** — the virtual import requires a multi-thread tokio runtime (`block_on` off a dedicated thread); a runtime-flavor-agnostic async-task → bridge-thread producer is deferred until a caller needs it.
- **`--output s3`** — not yet wired in the file-backed CLI (rejected up front). The direct-S3 CAS publish path exists (designed and verified CAS-safe); it needs an S3-backed home the CLI does not construct.
- **Per-property counts** — skipped-with-note in the gate; the manifest-sound non-nullable fast path is a real-Iceberg optimization.
- **Bounded stratified sample** — the sample currently enumerates the source; a `read_task_sample`-bounded sample is the scale follow-up.
- **Deliverable 3 (freshness)** — snapshot-diff delta-pack shipping builds on this watermark; out of scope here.

**Package-end gate (lead-run, not in this PR):** the live native-sf01 full-triple diff and any live build against Snowflake. No live runs were performed here.

### Warehouse-root direct resolution (follow-on, AJ live-gate registration-gap fix)

`bf27d545a`→`6fcf7fc64`. The local CLI's Direct mode was single-table, blocking registration of a multi-table R2RML source over a catalog-less bucket copy (a Snowflake-managed Iceberg database whose table dirs carry random suffixes, `fact_order.UIHGsQex/`). Now a Direct `--table-location` may be a **warehouse ROOT**: auto-detected when its leaf dir doesn't name the requested table, each `rr:tableName` resolves to its own dir under the root via one session-cached S3 LIST (`SendIcebergStorage::list_dir`), matched by the pure `match_warehouse_table_dir` (`<name>.<suffix>/` or bare `<name>/`, case-insensitive, namespace-stripped). Ambiguity/miss fails loud naming the candidates. Single-table direct is byte-unchanged (self-diff), no new flag / config change (rule documented in `iceberg map --help`). Registration shape: `fluree iceberg map <name> --mode direct --table-location s3://bucket/warehouse/dw --s3-region <r> --r2rml <ttl>` — pure ambient IAM, zero catalog/OAuth. Hermetic: 5 match tests; iceberg lib 251, api it_graph_source_r2rml 62 green. The live S3 LIST is part of the lead-run gate.

## Consolidated scope

#1529 folds in no other pull requests. It implements the native-twin builder set out in the materialization design record and was built first-principles, reusing the existing bulk-ingestion finalize → spool-sort → k-way-merge → publish pipeline pointed at the R2RML scan stream. Its pre-consolidation tip is preserved at tag `archive/pre-refactor-2026-07-21/feat_materialize-builder`; nothing was rewritten. Its base is re-pointed onto the family-C branch beneath it (`perf/family-c-filter-join`), which is a sibling of the fork point, so the diff renders builder-only while GitHub notes the base is some commits ahead.

## Verification of record

Every proof in this PR is hermetic; the live gate is deferred and is the basis for the draft status.

The hermetic proofs, all green: `materializer_driver_matches_engine_wildcard` (the driver's triple set is byte-identical to the query engine's `?s ?p ?o` wildcard over the same scan); `materialize_builds_queryable_native_twin` (a real indexed and published twin round-trips class counts, an FK edge, an `xsd:decimal`, and a lang tag); `materialize_stamp_survives_pack_restore` (archive→restore preserves the head-commit stamp byte-for-byte); and the parity-gate pair `verify_twin_passes_on_faithful_twin` and `verify_twin_fails_on_corrupted_twin` (an injected instance fails on both the class-count check and the full diff). The independent adversarial review is recorded under R-1529 in [pr-reviews-impl.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/pr-reviews-impl.md): verdict SHIP (draft) with no blocking finding — the shared text-import path is byte-untouched, the completion stamp rides the structurally-final commit so partial builds are detectable, the CLI's `Box::leak` is shell-only and one-per-process, and every fail-loud guard fires. The review's load-bearing observation, which a reader must internalize: the parity gate is an ingestion-integrity gate, not an enumerator-correctness gate — its oracle is the enumerator, so a fault in the enumerator itself is only caught by the deferred live native-sf01 diff.

The build/serve/sync arithmetic and the pack substrate anchors are in [C3-materialization-facts.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/C3-materialization-facts.md); the native-twin-promotion strategy and its snapshot-keyed correctness lemma are in [D1-strategy-options.md](https://github.com/fluree/db/blob/docs/virtual-dataset-audit-2026-07/D1-strategy-options.md).

## Native-impact appendix

- **One shared-path field, self-diff byte-identical for every text import.** `ParsedChunk.txn_meta` plus a `finalize_parsed_chunk` consume land on the shared bulk import / transact path (`fluree-db-transact/src/import.rs` and `import_sink.rs`). Every text-import `ParsedChunk` site sets `txn_meta = Vec::new()` (`parse_chunk`, the ndjson constructor, and `finalize`), and `is_virtual = config.virtual_source.is_some()` is false for all file and remote imports; the new `ImportSource::Virtual` arm is inserted before the remote-parallel path with no existing body touched. The `import.rs` diff is 138 insertions / 1 deletion (the only change to an existing line is `is_channel_fed` gaining `|| is_virtual`). The `build_watermark` trait default is empty, so non-materialize providers are unaffected, and `record_snapshot` is fire-and-forget first-writer-wins with no extra I/O. Verified: transact lib 303 + import 10 green, and `materialize_stamp_survives_pack_restore`.
- **The rest is new code reusing shared helpers.** The new modules (`fluree-db-r2rml/src/materialize/graph.rs`, `fluree-db-api/src/materialize.rs`, the CLI, and the warehouse-root `list_dir`) compose the shared per-term helpers the query operator uses without modifying them.

Confirmed negatives (git-checked): no `arrow` / `datafusion` / `parquet` dependency bump, and no `.github/workflows` or `rust-toolchain` edits.

## Draft status and open gates

This PR is deliberately a draft. Two gates are open and both must be run before any live cutover relies on the twin:

- **The live native-sf01 full-triple diff — now CLOSED (2026-07-29)**: run at full 35.2M scale against the natively-imported ledger, zero differences in either direction (see the addendum above). The enumerator-correctness gap the hermetic gate could not cover is discharged.
- **The warehouse-root direct-resolution S3 LIST arm** is exercised only by a live probe against a real catalog-less bucket copy; the match logic is hermetically tested (5 cases) but the live LIST has not been run.

Until both are exercised, the builder invites hermetic review as a draft with the live parity gate stated as pending; it should not be flipped ready on the hermetics alone.


aaj3f:
Flipped to ready: the draft-status reason (the unrun independent live gate) is discharged — see the 2026-07-29 addendum at the top of the body for the fix slate, the full-scale acceptance numbers, and the zero-diff oracle verdict at 35,238,778 triples. Reading order still applies: this reviews on top of #1554 (the landing PR for the audit program content).

fluree-db-api/src/import.rs
```diff
@@ -347,12 +361,25 @@ impl ImportConfig {
...
+        } else {
+            // Sub-2GB budgets (O6): the flat 2GB overhead term underflowed to 0 and
+            // the 128MB clamp floor then applied REGARDLESS of budget — so a
+            // 512MB-budget import used the same 128MB chunk as a 2GB one, and a
+            // 128MB chunk under a 512MB budget blows the budget once the ~2.5x
+            // parse/inflight expansion is counted. Instead reserve ~40% for run
+            // buffers + index build and divide the rest across the working set;
+            // floor at 16MB (still worth a commit), cap at 128MB (continuous with
+            // the large-budget floor at the 2GB boundary).
+            let working = budget_mb as f64 * 0.6 / denominator;
```
bplatz, line 393: **Shared-path behavior change that the native-impact appendix does not list.**

The appendix says the only shared-path change is `ParsedChunk.txn_meta`, "self-diff byte-identical for every text import." This rewrite of `effective_chunk_size_mb` changes chunk sizing for EVERY import whose budget is under 2048 MB — Turtle/TriG/JSON-LD included, not just materialize.

Concretely, at `memory_budget_mb = 512` with `max_inflight = 2`: before, `512.saturating_sub(2048) = 0` → `clamp(128, 768)` = **128 MB**. After: `512 * 0.6 / 6` = **51 MB**. That is ~2.5x the commit count and a different ledger shape for the same input.

The default budget is 80% of RAM, so most hosts land above 2048 and are unaffected. What IS affected: containers under roughly 2.5 GB, and anyone who passes `--memory-budget-mb` explicitly — which the PR's own machine-safety posture actively encourages, and which every finalizer-touching test in this repo already does (`memory_budget_mb = 256`).

The fix itself reads correct to me — the old floor genuinely ignored the budget below 2 GB and the comment explains exactly why. The ask is just that the appendix name it as a second shared-path change with its blast radius, rather than the byte-identical claim standing unqualified.

fluree-db-api/src/materialize.rs
```diff
@@ -0,0 +1,2757 @@
...
+    const PAGE: usize = 200_000;
+    let file = std::fs::File::create(path)
+        .map_err(|e| R2rmlError::Materialization(format!("twin spool create: {e}")))?;
+    let mut writer = std::io::BufWriter::new(file);
+    let mut class_count: BTreeMap<String, u64> = classes.iter().map(|c| (c.clone(), 0)).collect();
+    let type_p = format!("<{}>", fluree_vocab::rdf::TYPE);
+    let mut offset = 0usize;
+    loop {
+        let q = format!(
+            "SELECT ?s ?p ?o WHERE {{ ?s ?p ?o }} ORDER BY ?s ?p ?o LIMIT {PAGE} OFFSET {offset}"
```
bplatz: **OFFSET paging over the whole twin is O(n^2), and moves the memory cost onto the query side.**

At 35M triples this is ~176 pages, and each page asks the engine for a full `ORDER BY ?s ?p ?o` over the entire twin and then skips `M` rows. Each page also materializes 200k bindings as `serde_json::Value` via `to_sparql_json`.

So the boundedness O2 won is real on the verify side (external sort, `SORT_RUN_LINES` runs) but the twin-read side is neither bounded nor linear. The addendum's 35.2M full-triple diff reads like an external same-renderer N-Triples export rather than a run of `verify_twin_full`, so I do not think this specific path has been exercised at that scale — worth confirming.

An index-ordered scan or the existing export path would give a single linear pass; `spool_twin_ntriples` is most of the way to re-implementing one already.

fluree-db-api/src/materialize.rs
```diff
@@ -0,0 +1,2757 @@
...
+    provider: &P,
+    graph_source_id: &str,
+) -> Result<ParityReport, MaterializeError>
+where
+    P: R2rmlProvider + R2rmlTableProvider,
+{
+    let mapping = provider.compiled_mapping(graph_source_id, None).await?;
+    let classes = mapping_classes(&mapping);
+
+    let dir = std::env::temp_dir().join(format!(
```
bplatz: **`std::env::temp_dir()` may be tmpfs, which would undo the bounded-memory rewrite.**

Full mode writes two complete N-Triples renderings of the graph plus their sorted runs here — tens of GB at the scales this builder targets. On many Linux hosts `/tmp` is tmpfs and therefore RAM-backed, so the verify that was just rewritten to be memory-bounded would put it all back in memory, on exactly the large datasets where it matters.

There is also no override. Suggest defaulting to the import's spool dir (or somewhere under `.fluree`) and adding a `--tmp-dir`. `TmpDirGuard` cleans up on drop but not on SIGKILL, which is a second reason not to leave GBs in a shared `/tmp`.

fluree-db-api/src/materialize.rs
```diff
@@ -0,0 +1,2757 @@
...
+const STAMP_PRED_SAMPLE_SEED: &str = "sampleSeed";
+
+/// ns-encode the completion stamp into `txn_meta` entries, interning each
+/// predicate through `sink` so its namespace code is published in this chunk's
+/// namespace_delta (and thus resolves when the commit is read back). The
+/// watermark vector is stored as one deterministic JSON string (sorted by table)
+/// under `materialize:watermark`. The whole stamp must fit the 64 KiB txn_meta
+/// cap; a schema with a very large table count would need a compacter encoding
+/// (documented residual).
+fn encode_stamp(
```
bplatz, line 570: **The 64 KiB cap this doc comment asserts is not enforced on this path.**

`MAX_TXN_META_BYTES` is checked in exactly two places — `fluree-db-transact/src/parse/txn_meta.rs:407` and `parse/trig_meta.rs:1398` — both of which validate txn-meta arriving from user TriG/JSON-LD. `encode_stamp` builds `TxnMetaEntry`s directly and hands them to `finalize_parsed_chunk`, so neither validator runs.

The watermark JSON grows with table count times `metadata_location` length; S3 metadata URIs plus JSON framing run roughly 350 bytes per table, so ~190 tables. That is not an unreasonable warehouse.

Every other guard on this path is fail-loud with a typed error (`EmptyWatermark`, `ParentIndexBudgetExceeded`). This one should be too, rather than a comment describing a cap and a downstream failure mode that is not stated.

fluree-db-api/src/materialize.rs
```diff
@@ -0,0 +1,2757 @@
...
+    ));
+    Ok(entries)
+}
+
+/// Reader-side of the completion stamp: extract it from a commit's `txn_meta`, if
+/// present. A twin is valid iff a head-walk finds a commit whose txn_meta yields
+/// `Some` here (all three stamp fields present). Keys on the stable predicate
+/// LOCAL NAMES (the namespace code is per-ledger); the local names are
+/// materialize-specific, so requiring all three together identifies the stamp.
+pub fn read_stamp(txn_meta: &[TxnMetaEntry]) -> Option<WatermarkStamp> {
```
bplatz: **This ignores the namespace that `intern_meta_predicate` was added to publish.**

The reason `GraphSink::intern_meta_predicate` exists (and the reason `encode_stamp` routes the stamp predicates through the chunk's own sink) is so their namespace codes land in this commit's `namespace_delta` and resolve when the commit is read back. Then `read_stamp` matches on bare local names, so any commit carrying `builderVersion` + `mappingHash` + `watermark` string entries in ANY namespace parses as a valid twin stamp.

Given the stated contract is "a twin is valid iff a head-walk finds the stamp," it seems worth resolving the ns code and requiring it to be `MATERIALIZE_NS`. The information is already there by construction — the mechanism was built and then not used on the read side.

fluree-db-api/src/materialize.rs
```diff
@@ -0,0 +1,2757 @@
...
+/// triples (via [`ImportSinkObserver`]) instead of parsing Turtle text. `t` is
+/// the caller-assigned transaction number (`chunk_idx + 1`), matching the parse
+/// path's contract.
+///
+/// `stamp` is `Some` for exactly ONE chunk — the twin's FINAL commit — and its
+/// completion stamp is ns-encoded into this chunk's `txn_meta` through the SAME
+/// sink, so the stamp predicates' namespace codes publish in this commit's
+/// namespace_delta. Every other chunk passes `None`, leaving `txn_meta` empty
+/// (byte-identical to a text-import chunk).
+pub fn build_virtual_chunk(
```
bplatz: **Vestigial after O3a — `triples` is now always empty.**

Since the O3a commit removed the `OwnedTriple` intermediate, the only remaining call is `build_virtual_chunk(&[], ctx, t, idx, Some(&stamp))` at line 1006. So the `triples` parameter is dead on every path, `ImportSinkObserver` exists solely to serve it (its only use is the loop at 335-337 that never iterates), and the `OwnedTriple` alias exists only to name the dead parameter.

All three are `pub`, so no dead-code warning fires and nothing will surface this later — precisely the rot the repo's dead-code policy is aimed at. Suggest `build_stamp_chunk(ctx, t, idx, stamp)` and deleting `ImportSinkObserver` + `OwnedTriple`, or keeping them with the `#[expect(dead_code)]`-style why/when comment the policy asks for.

fluree-db-api/src/materialize.rs
```diff
@@ -0,0 +1,2757 @@
...
+impl ParityReport {
+    /// Human-readable summary of the failing checks (empty when passed).
+    pub fn failures(&self) -> Vec<&ParityCheck> {
+        self.checks.iter().filter(|c| !c.outcome.passed()).collect()
+    }
+}
+
+/// Subjects per class fully compared by the stratified sample (deterministic:
+/// the lexicographically first N subjects of each class).
+const SAMPLE_SUBJECTS_PER_CLASS: usize = 3;
```
bplatz, line 1231: Worth surfacing in `--help`: the default gate's sample strength is three subjects per class. Combined with the class-count check that is a reasonable smoke test, but "quick verify" reads stronger than it is, and the CLI already prints an honesty note about the shared oracle — the sample size belongs in the same breath.

Separately, `seeded_offset` can put the window near the end of a class, so `twin_sample_subject_iris` issues `ORDER BY ?s LIMIT 3 OFFSET N` with N close to the class size — a full sort and skip per class. Fine at the sizes tested; the same shape as the full-mode paging note above.

fluree-db-r2rml/src/materialize/graph.rs
```diff
@@ -0,0 +1,1447 @@
...
+            let subject = match materialize_subject_from_batch(&tm.subject_map, batch, row)? {
+                Some(s) => s,
+                None => continue,
+            };
+            for cols in &col_sets {
+                if let Some(key) = get_join_key_from_batch(cols, batch, row) {
+                    entry
+                        .entry(cols.clone())
+                        .or_default()
+                        .insert(key, subject.clone());
```
bplatz: Last-wins on a duplicate parent join key. This MATCHES the generic query operator (`R2rmlScanOperator`'s `ParentLookup` is also a plain `HashMap<Vec<String>, RdfTerm>`), so the twin agrees with the virtual path and the parity gate is right to pass — no divergence introduced here.

The reason it is still worth a note: the fused aggregate deliberately DECLINES this shape (`insert_dim_gkeys` returns false on any duplicate, with a comment explaining that a single-entry-per-key map cannot represent the fan-out), and materializing bakes the last-wins result into a permanent native ledger where fixing the operator no longer fixes the data. If the intent is that the twin is the durable artifact, a loud decline here — matching the fused path's posture — may be worth more than silent agreement with the deviation.

fluree-db-cli/src/commands/materialize.rs
```diff
@@ -0,0 +1,336 @@
...
+
+use crate::cli::{MaterializeOutput, MaterializeVerify};
+use crate::context;
+use crate::error::{CliError, CliResult};
+use fluree_db_api::server_defaults::FlureeDir;
+
+/// The env var read by fluree-db-iceberg's fail-closed merge-on-read guard. A
+/// stable public contract (`fluree_db_iceberg::mor_guard::ALLOW_MOR_DELETES_ENV`);
+/// hard-coded here so the CLI need not take a direct dependency on that crate.
+const ALLOW_MOR_DELETES_ENV: &str = "FLUREE_ICEBERG_ALLOW_MOR_DELETES";
```
bplatz: The comment calls this a stable public contract, but a hard-copied literal with no compile-time link is the drift hazard the repo handles elsewhere with `// SYNC:` comments at both ends (the OTEL init pair in `telemetry.rs` / `main.rs`, and the Cypher `"5.4.0"` version string). A `// SYNC:` here and on `mor_guard::ALLOW_MOR_DELETES_ENV` would at least make a rename visible to whoever does it.

Note also that this file is `#![cfg(feature = "iceberg")]`, and under that feature `fluree-db-iceberg` is already in the tree via `fluree-db-api` — so importing the constant may be cheaper than the comment suggests.

bplatz (approved):
Approving. Please walk the inline notes below and decide which belong in this PR versus a follow-up — a few of them (1, 4) I think are worth folding in here; the rest are your call.

## Verified locally

- `cargo clippy --all --all-features --all-targets -- -D warnings` — clean.
- `cargo nextest run -p fluree-db-api -p fluree-db-r2rml -p fluree-db-cli -p fluree-db-transact -p fluree-db-iceberg --all-features --no-fail-fast` — 4373 passed, 0 failed.

## The one finding that is not inline

**No docs.** This adds a new top-level `fluree materialize` command with no `docs/cli/materialize.md` and no `docs/SUMMARY.md` entry; every one of the other ~43 CLI commands has both. `docs/graph-sources/r2rml.md` and `docs/graph-sources/iceberg.md` are the natural home for the twin / watermark / completion-stamp concept and for the new warehouse-root `--table-location` resolution rule, which the body says is documented only in `iceberg map --help`. Per the repo's docs mandate this belongs in the same changeset.

## Process note

The base `audit-program/land-on-main` is 45 commits and ~10k lines ahead of main, so this diff is ~6.4k of the ~16.5k that actually reaches main through it. Worth stating explicitly for whoever sequences the landing.

## What is strong

The wave-scheduled index-during-emit design is the right shape, and claiming a chunk `idx` only AFTER securing the first batch — so an empty chunk can never leave a gap that stalls the in-order consumer forever — is exactly the kind of detail that is easy to get wrong and was gotten right. Publish-then-verify-then-drop is sound: I confirmed `create ... import` rejects a ledger that already has commits, so the hard drop can only ever hit a ledger this build just created, and the `:branch` strip closes the one way that guard could have been defeated. The verify-default honesty message printed to the operator is the right instinct. And apart from the chunk-sizing note, the shared-path footprint really is as small as the appendix claims.

fluree-db-api/src/import.rs
aaj3f: You're right on both counts here — the appendix's "byte-identical" claim was written before the O6 commit landed and I never went back to qualify it. I've reworked that section of the body to name `effective_chunk_size_mb` as a second shared-path change with your blast-radius numbers (128MB → ~51MB chunks at 512MB budget, ~2.5× commit count, sizing-only — contents/index identical), and `95389a66f` adds the same disclosure as a comment at the site so the next reader of `import.rs` doesn't need the PR body to know text imports under 2GB budgets are affected. Thanks for actually working the arithmetic — the 512MB example is now the canonical one in both places.

fluree-db-api/src/materialize.rs
aaj3f: Good catch — on a tmpfs host that would have quietly un-done the entire bounded-memory story at exactly the scales it was built for. `d1a5d0904` defaults the verify spool under the target's `.fluree` storage area instead of `std::env::temp_dir()`, and adds `--tmp-dir` for operators who want it elsewhere. The SIGKILL-residue point is documented alongside (the guard cleans on drop; a hard kill leaves files, and having them under `.fluree` at least makes them discoverable rather than lost in a shared `/tmp`).

fluree-db-api/src/materialize.rs
aaj3f: Agreed, and folded in per your call: `d1a5d0904` enforces `MAX_TXN_META_BYTES` fail-loud before finalize with a typed `StampTooLarge` naming actual-vs-cap (and the table-count driver, since watermark size scales with tables × metadata-location length), consistent with `EmptyWatermark`/`ParentIndexBudgetExceeded`. There's a boundary test that builds a watermark just over the cap. Your ~190-table arithmetic is what convinced me this wasn't theoretical — that's a real warehouse.

fluree-db-api/src/materialize.rs
aaj3f: This one stung a little because, as you said, the mechanism was built and then not used on the read side. `d1a5d0904` fixes it: `read_stamp` now resolves the stamp predicates' namespace codes via the commit's `namespace_delta` and requires `MATERIALIZE_NS` — a commit carrying `builderVersion`/`mappingHash`/`watermark` local names in any other namespace no longer parses as a twin stamp. There's a forged-foreign-ns rejection test plus the round-trip.

fluree-db-api/src/materialize.rs
aaj3f: Cleaned up in `d1a5d0904`: `build_stamp_chunk(ctx, t, idx, stamp)` replaces the empty-slice call, and `ImportSinkObserver` + the `OwnedTriple` alias are deleted rather than annotated — you're right that `pub` was going to keep this invisible forever.

fluree-db-api/src/materialize.rs
aaj3f: `--verify` help now states the quick-mode sample strength (three subjects per class + class counts) in the same breath as the shared-oracle honesty note. On the `seeded_offset` sort-and-skip shape: acknowledged as the same class as the full-mode paging note — the full-mode side is gone entirely (see the other thread), and the per-class variant is bounded by class count at sizes where it doesn't bite; left in place with a note rather than rewritten.

fluree-db-r2rml/src/materialize/graph.rs
aaj3f: Your instinct here turned out to be right for a stronger reason than durability. We put this one through a careful adversarial pass before deciding, and Claude's determinism analysis found the decisive fact: the last-wins pick was *run-to-run nondeterministic* — the parent index consumes `scan_table`'s `buffer_unordered` stream, so when a duplicate key's colliding rows span data files, I/O completion order chooses the winner. Two runs over the same pinned snapshots could bake different twins, which breaks the mapping-hash/stamp reproducibility contract outright — so "silent agreement with the virtual path" wasn't actually agreement with any one graph.

`5602938b4` lands the resolution: materialize now declines duplicate parent join keys by default with a typed error naming table/join-column/count, `--allow-duplicate-parent-keys` overrides (the MoR-guard pattern), and — the part I'd call the real fix — a deterministic keep-min tie-break (lexicographically smallest parent subject) now shared by BOTH the twin builder and the virtual path's `build_parent_lookup`, so the two agree *and* both are reproducible regardless of scan order. An override-built twin records its distinct ambiguous-key count in the stamp (distinct keys deliberately — a raw collision tally is itself order-dependent). The true RefObjectMap fan-out (one edge per matching parent) is now tracked as #1583; since twins are rebuildable, re-materialization heals any override-built twin once that lands.

fluree-db-cli/src/commands/materialize.rs
aaj3f: You were right that the import is cheaper than the comment claimed — under `#[cfg(feature = "iceberg")]` the crate's already in the tree, so `d1a5d0904` just imports `ALLOW_MOR_DELETES_ENV` through the `fluree-db-api` re-export and the hand-copied literal is gone.

aaj3f:
Thanks for this review, @bplatz — the local-verification legwork (full clippy + the 4,373-test run) on top of the inline notes is much appreciated. Everything is now folded in across `95389a66f` / `d1a5d0904` / `82b44631a` / `5602938b4`: both of your fold-in picks (the chunk-sizing disclosure — the body appendix now names it as a second shared-path change with your blast-radius numbers — and the fail-loud stamp cap), the full-verify rewrite as a single linear pass (your suspicion was correct that the 35M diff had gone through the external export pipeline, not `verify_twin_full`), the tmpfs-safe spool + `--tmp-dir`, the namespace-checked `read_stamp`, the dead-code cleanup, the `--help` honesty note, the imported env constant, and — per the repo docs mandate you flagged — `docs/cli/materialize.md` + the `SUMMARY.md` entry + the twin/watermark/stamp concept and warehouse-root resolution rule in the graph-sources docs.

The duplicate-parent-key note grew into the most substantive change of the set: an adversarial pass found the last-wins pick was run-to-run nondeterministic (I/O-completion order picks the winner across data files), so beyond your decline suggestion it now ships a deterministic keep-min tie-break shared by the twin builder and the virtual path — detail in that thread, fan-out tracked as #1583.

On your process note re: the base branch — stated explicitly for whoever sequences the landing: #1554 (the `audit-program/land-on-main` PR) lands first, and this PR then auto-retargets to `main` with its diff collapsing to the builder-only ~6.4k. The base was refreshed against current main today so #1554 reviews conflict-free.

fluree-db-api/src/materialize.rs
aaj3f: Confirmed — your read was exactly right, and worth stating plainly: the 35.2M full-triple diff in the addendum was the *external* same-renderer export → sort → comm pipeline, not `verify_twin_full`; the OFFSET-paged path had never been exercised past test scale. `d1a5d0904` replaces it with a single linear index-ordered pass (reusing the export-path scan shape feeding the existing external-sort/k-way machinery, which is more or less the "spool_twin_ntriples is most of the way there" observation implemented), and counts now come from bounded COUNT queries rather than full materialization. There's a hermetic multi-chunk twin test on the new path. The per-class seeded sampling in quick mode still does an `ORDER BY ?s LIMIT 3 OFFSET N` per class — same shape you flagged in the other note, bounded at 16 classes here — noted in-code rather than rewritten; happy to fold it into a follow-up if it ever grows teeth.

Issue #1532, opened 2026-07-22
History queries: unbound ?t / ?op silently return null instead of erroring


bplatz:
A history query (`from` + `to`) that selects `?t` / `?op` **without binding them** succeeds, returns history rows, and leaves both columns null. Asserts and retracts become indistinguishable, with no error or warning.

## Repro

```sparql
PREFIX ex: <http://example.org/>
SELECT ?name ?t ?op
FROM <mydb:main@t:1>
TO <mydb:main@t:latest>
WHERE { ex:alice ex:name ?name . }
```

Returns 3 rows (the full change log) with `?t: null, ?op: null` on every row.

The correct query needs the RDF-star metadata bindings:

```sparql
WHERE { << ex:alice ex:name ?name >> f:t ?t ; f:op ?op . }
```

The JSON-LD equivalent has the same shape: `@t` / `@op` annotations on the value object are required, and omitting them yields nulls.

A related case fails more quietly still — filtering on unbound history metadata returns an empty result set:

```sparql
WHERE { ?s ?p ?o . FILTER(?t >= 10 && ?t <= 15) }   # → []
```

## Why it matters

This is the natural first guess at the syntax, and it is indistinguishable from success. It is what let the time-travel cookbook drift for as long as it did — every example in it used this shape and nobody noticed, because nothing errors.

Note the engine already errors cleanly on an adjacent mistake: a constant object in the quoted triple (`<< ex:alice ex:name "Alice" >> f:t ?t`) is rejected with a diagnostic. The unbound case deserves the same treatment.

## Proposal

In history mode, when a projected variable is neither bound by a pattern nor produced by a metadata annotation, and its name suggests history metadata (or more robustly: when the query is a history-range query and *no* metadata bindings appear anywhere), reject with a diagnostic pointing at the `f:t` / `f:op` (SPARQL) or `@t` / `@op` (JSON-LD) form.

A warning on the response rather than a hard error is also defensible, since the rows themselves are correct — only the metadata is missing.

## Current behaviour is pinned

`fluree-db-api/tests/it_query_history_sparql.rs::history_metadata_vars_without_bindings_are_null` locks the current behaviour as a regression test. Update it — do not delete it — when this changes.

Docs describing the sharp edge: `docs/guides/cookbook-time-travel.md` (Gotchas), `docs/concepts/time-travel.md` (Entity History (SPARQL)).

bplatz:
Closing this — the premise was wrong, and the pushback on it was right.

[...]

PR #1533, opened 2026-07-22
docs: Correct time-travel and history query documentation

bplatz:
The time-travel cookbook had drifted badly from the implementation — most of its examples were not runnable. Audited every documented form against a real ledger and rewrote what was wrong.

## What was broken

| Doc | Problem |
|---|---|
| `cookbook-time-travel.md` | `fluree history` shown taking a SPARQL query string; it takes an **entity IRI** and builds the query itself. Every `fluree history` example was invalid. |
| `cookbook-time-travel.md` | HTTP examples used `?ledger=` and `?t=` URL params. Neither exists — `SparqlParams` has only `query` / `default_context` / `default_graph_uri`. |
| `cookbook-time-travel.md` | Multi-ledger example used an invented `{alias: {ledger, t}}` map plus `@graph` aliases. |
| `cookbook-time-travel.md` | `FILTER(?t >= 10 && ?t <= 15)` against a never-bound `?t` — silently returns `[]`. |
| `cookbook-time-travel.md` | Claimed `?op` renders as `assert` / `retract`. It is `xsd:boolean`; only the CLI *table* renderer shows `+` / `-`. |
| `concepts/time-travel.md` | `["filter", "(= ?op \"retract\")"]` matches nothing — contradicted `jsonld-query.md`, which already says string constants are not accepted. |
| `cli/history.md` | Missing `--remote`; did not say the argument is an entity IRI rather than a query. |

The cookbook also never mentioned the RDF-star form that history queries actually use, nor `@recorded:`.

`docs/query/sparql.md` and `docs/query/jsonld-query.md` were already correct and are essentially untouched.

## SPARQL coverage

Several history patterns had been shown only in JSON-LD. The cookbook now carries SPARQL for all of them — variable-predicate (all properties of an entity), variable-subject (one property across subjects), retractions-only as a complete query rather than a bare `FILTER` fragment, range scoping, and a `GROUP BY` aggregate over history rows.

## What is now documented that was not

- SPARQL `FROM ... TO ...` is **connection-scoped only** — rejected on `/v1/fluree/query/{ledger}` (`routes/query.rs:2687`) and on `/v1/fluree/stream/query` (`routes/stream_query.rs:306`).
- History mode rejects reasoning.
- The quoted triple's object must be a variable.
- `<< s p o >>` (flake metadata) vs `<<( s p o )>>` (RDF 1.2 reification) do not compose.
- The cookbook now separates point-in-time travel from history queries up front — they are different capabilities that were being run together.

## On unbound `?t` / `?op`

I originally filed #1532 arguing the engine should error when a history query selects `?t` / `?op` without binding them. That was wrong and the issue is closed.

The same shape in a non-history query behaves identically — `SELECT ?name ?zzz ?t ?op` returns null for all three unbound variables. This is just SPARQL 1.1: selecting a variable the WHERE clause never binds is legal and yields unbound. Erroring would deviate from the spec, reserve `?t` as a magic name (breaking anyone using it for "type"/"total"), and be arbitrary — there is no principled reason to reject `?t` but accept `?zzz`. We already have real syntax for this, and the engine already errors cleanly on genuine misuse of it.

It was a documentation failure, not an engine one. The gotchas section now explains the semantics rather than implying a defect.

## Root cause of the drift

The SPARQL history lowering path (`lower/rdf_star.rs`) had parser-level tests but **no end-to-end coverage** — the five existing `it_query_history*` / `it_query_time_travel*` files are all JSON-LD. Nothing would have caught the docs describing syntax that does not work.

Adds `it_query_history_sparql.rs` (10 tests) covering every SPARQL form the cookbook now shows: metadata binding, the `;`-continued form, variable predicate, variable subject, boolean `f:op` filtering, retractions-only, narrowed ranges, `GROUP BY` aggregates, rejection of constant quoted-triple objects, and the unbound-variable semantics above.

## Also

`fluree history --format` advertised `tsv` in its clap help, which the command rejects at runtime. Dropped from the help text.

## Verification

Every example in the rewritten cookbook was executed against a real ledger — including the multi-ledger `from` array with per-ledger `@t:` suffixes and both structured `from` shapes. All cross-reference anchors checked.

`cargo clippy -p fluree-db-cli -p fluree-db-api --all-features --all-targets -- -D warnings` clean; `grp_query_history` 45/45 passing.

### DIFF
--- a/.fluree-memory/repo.ttl
+++ b/.fluree-memory/repo.ttl
@@ -1108,6 +1108,23 @@ mem:fact-01kp4g1wa8yyty4sd2p03k6bg9 a mem:Fact ;
     mem:artifactRef "fluree-graph-json-ld/src/expand.rs" ;
     mem:createdAt "2026-04-13T22:40:33.608461+00:00"^^xsd:dateTime .
 
+mem:fact-01ky55dwckmwazppht36vrn5hk a mem:Fact ;
+    mem:content "SPARQL history queries use the legacy RDF-star form: `FROM <l@t:1> TO <l@t:latest>` + `<< s p ?o >> f:t ?t ; f:op ?op`. The carve-out fires ONLY when the quoted-triple subject is reifier-less AND the predicate is f:t/f:op; anything else takes the RDF 1.2 reification path. Inner object must be a variable (BIND target). `f:op`/`@op` is xsd:boolean — \"assert\"/\"retract\" strings match nothing. Selecting ?t/?op WITHOUT the bindings does NOT error: it returns history rows with null metadata (issue #1532; pinned by history_metadata_vars_without_bindings_are_null). `FROM..TO` is rejected on the ledger-scoped and streaming endpoints — connection-scoped `/v1/fluree/query` only." ;
+    mem:tag "docs" ;
+    mem:tag "history" ;
+    mem:tag "rdf-star" ;
+    mem:tag "sparql" ;
+    mem:tag "time-travel" ;
+    mem:scope mem:repo ;
+    mem:artifactRef "docs/guides/cookbook-time-travel.md" ;
+    mem:artifactRef "fluree-db-api/tests/it_query_history_sparql.rs" ;
+    mem:artifactRef "fluree-db-query/src/eval/fluree.rs" ;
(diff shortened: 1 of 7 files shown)


zonotope (approved):
🥍

PR #1534, opened 2026-07-22
Redact graph source config secrets

### LINKED ISSUE #1497: `fluree info` prints OAuth2 `client_secret` in plaintext — ConfigValue redacts Debug but not Serialize

zonotope:
## Problem

`fluree info <gs>` and `fluree iceberg info <gs>` printed the full graph-source config as pretty JSON — including `catalog.auth.client_secret`, i.e. a live Snowflake PAT / OAuth2 client secret or bearer token. `ConfigValue` has a redacting `Debug` impl, but every CLI print path serializes via serde, which bypasses it, so the secret landed in terminal scrollback, `script` logs, and CI output. The server routes already redact through `redact_graph_source_config`; the CLI's local print paths were the unpatched surface. (A redacting `Serialize` on the type itself is not an option: persistence requires lossless serialization — CI-locked tests assert the stored config carries the secret for query-time catalog auth. The type-level fix, `ConfigValue::SecretRef`, is tracked separately in #1500 §3.)

## Fix

All CLI graph-source rendering now goes through a single new module, `fluree-db-cli/src/graph_source_display.rs`, which routes every config through the existing fail-closed redactor before printing. Local nameservice records are converted with `build_generic_graph_source_info` — the same redacted DTO the server's `/info` route emits (now `pub` in `fluree-db-api`) — so local and remote sources render identically through one renderer, and redaction happens at the API layer rather than being re-implemented per print site. Remote responses are re-redacted client-side as defense-in-depth against unredacting servers; the redactor is idempotent, so already-clean responses render unchanged.

The refactor also collapses duplication this bug was hiding in: the two print functions existed verbatim in both `info.rs` and `iceberg.rs`, and the source-type label mapping existed three times across the workspace (CLI ×1 after its own dedup, server ×1, plus the private `graph_source_type_label` in `fluree-db-api`). All of it now resolves to the one API helper, so a future label or field change cannot drift between the CLI, the server, and the API.

## Hardening beyond the original leak

- **Container-shaped secrets** (`fluree-db-api`): `redact_json_secrets` only masked scalar values under secret keys; a secret hidden in a container shape (`{"token": ["tok-1"]}`, `{"client_secret": {"value": "..."}}`) passed through unredacted. Containers under a secret key are now redacted with inverted strictness — every scalar leaf inside is masked except the `env_var`/`java_property` reference identifiers — closing the gap for hand-authored or out-of-band records while preserving the env-var debugging affordance. This hardens every consumer of the shared redactor (server `/info`, events route, CLI). Complements #1504, which tracks the key-name denylist failing open.
- **Fail-closed on unparseable configs**: a stored config that does not parse now renders as `[redacted:unparseable]` (matching the server-side semantics) instead of being silently omitted — an operator no longer concludes a misparsed (and possibly secret-bearing) config doesn't exist, and its raw bytes are never echoed.
- **The redaction property is a tested invariant**: the print functions are thin wrappers over `render_*` functions returning `String`, and tests feed secret-bearing records and server responses through the real render seam, asserting the secrets are absent, `[redacted]` is present, and non-secret identifying fields (client id, scope, catalog URI) survive. A refactor that drops the redaction call now fails the suite instead of silently reintroducing the leak.

## User-visible output changes

- `client_secret`, `token`, and other secret values print as `"[redacted]"` everywhere (the point of the PR).
- The unified renderer prints `Retracted:` whenever the field is present, so remote output for generic (BM25/Vector/Geo) sources gains that line; local output keeps it.
- `Index ID:` always prints, showing `(none)` when the source has no index; previously remote output omitted the line entirely.
- Empty configs are omitted consistently: `"{}"`, `"{ }"`, and `"{\n}"` all suppress the Configuration section (previously only the byte-exact `"{}"` did, via a string comparison in the server DTO builder).
- Unparseable configs render a Configuration section containing `[redacted:unparseable]` instead of nothing.

## Verification

Unit tests cover the render seam end-to-end (secret masking local and remote, env-var name preservation, fail-closed unparseable, non-secret configs rendering in full) plus new API-side tests for container-shaped secrets and whitespace-variant empty configs; the existing per-variant redaction canaries and lossless-storage assertions all still pass, including under `--features iceberg`. Verified live against the issue's reproduction: `fluree iceberg map --oauth2-client-secret "$PAT" ...` followed by `fluree info <gs> --direct` and `fluree iceberg info <gs>` prints `"client_secret": "[redacted]"` with zero occurrences of the raw secret, while the stored nameservice record still carries it (persistence remains lossless). Clippy is clean and `cargo fmt --check` passes across the touched crates.

Closes #1497

### DIFF
--- a/fluree-db-api/src/ledger_info.rs
+++ b/fluree-db-api/src/ledger_info.rs
@@ -1001,7 +1001,10 @@ fn is_secret_config_key(key: &str) -> bool {
 /// "default_val": "..."}`) keeps its `env_var` name (safe, aids debugging) while
 /// the inline `default_val` fallback is masked by the recursive walk. Returns
 /// `true` when anything was redacted.
-fn redact_json_secrets(value: &mut JsonValue) -> bool {
+///
+/// Value-level form of [`redact_graph_source_config`], for callers that
+/// already hold a parsed JSON tree.
+pub fn redact_json_secrets(value: &mut JsonValue) -> bool {
     let mut redacted = false;
     match value {
         JsonValue::Object(map) => {
--- a/fluree-db-api/src/ledger_info.rs
+++ b/fluree-db-api/src/ledger_info.rs
@@ -995,12 +995,26 @@ fn is_secret_config_key(key: &str) -> bool {
     SECRET_CONFIG_KEYS.iter().any(|s| *s == lower)
 }
 
-/// Recursively replace secret-bearing scalar values with `"[redacted]"`.
+/// Keys inside a secret-keyed reference object that identify where the secret
+/// comes from rather than what it is (the `ConfigValue::Dynamic` wire shape);
+/// their string values are preserved during subtree redaction.
+const SECRET_REF_SAFE_KEYS: &[&str] = &["env_var", "java_property"];
+
+fn is_secret_ref_safe_key(key: &str) -> bool {
+    let lower = key.to_ascii_lowercase();
+    SECRET_REF_SAFE_KEYS.iter().any(|s| *s == lower)
+}
+
+/// Recursively replace secret-bearing values with `"[redacted]"`.
 ///
-/// A secret held as an env-var reference object (`{"env_var": "...",
-/// "default_val": "..."}`) keeps its `env_var` name (safe, aids debugging) while
-/// the inline `default_val` fallback is masked by the recursive walk. Returns
-/// `true` when anything was redacted.
+/// A scalar under a secret key is masked outright. A container under a secret
+/// key is redacted with inverted strictness by [`redact_secret_subtree`]:
+/// every scalar leaf inside is masked except reference-identifier strings
+/// (`env_var`, `java_property` — safe, they aid debugging), so a secret held
+/// as `{"env_var": "...", "default_val": "..."}` keeps its `env_var` name
+/// while the inline `default_val` fallback is masked, and a secret hidden in
+/// an unrecognized container shape (`{"token": ["..."]}`) cannot escape.
+/// Returns `true` when anything was redacted.
 ///
 /// Value-level form of [`redact_graph_source_config`], for callers that
 /// already hold a parsed JSON tree.
@@ -1009,9 +1023,15 @@ pub fn redact_json_secrets(value: &mut JsonValue) -> bool {
     match value {
         JsonValue::Object(map) => {
             for (k, v) in map.iter_mut() {
-                if is_secret_config_key(k) && !v.is_object() && !v.is_array() && !v.is_null() {
-                    *v = JsonValue::String("[redacted]".to_string());
-                    redacted = true;
+                if is_secret_config_key(k) {
+                    if v.is_object() || v.is_array() {
+                        if redact_secret_subtree(v) {
+                            redacted = true;
+                        }
+                    } else if !v.is_null() {
+                        *v = JsonValue::String("[redacted]".to_string());
+                        redacted = true;
+                    }
                 } else if redact_json_secrets(v) {
                     redacted = true;
                 }
@@ -1025,6 +1045,37 @@ pub fn redact_json_secrets(value: &mut JsonValue) -> bool {
             }
         }
         _ => {}
+    }
+    redacted
+}
+
+/// Redact every scalar leaf beneath a secret-keyed container, keeping only
+/// reference-identifier strings ([`SECRET_REF_SAFE_KEYS`]) and nulls.
+fn redact_secret_subtree(value: &mut JsonValue) -> bool {
+    let mut redacted = false;
+    match value {
+        JsonValue::Object(map) => {
+            for (k, v) in map.iter_mut() {
+                if is_secret_ref_safe_key(k) && v.is_string() {
+                    continue;
+                }
+                if redact_secret_subtree(v) {
+                    redacted = true;
+                }
+            }
+        }
+        JsonValue::Array(arr) => {
+            for v in arr.iter_mut() {
+                if redact_secret_subtree(v) {
+                    redacted = true;
+                }
+            }
+        }
+        JsonValue::Null => {}
+        _ => {
+            *value = JsonValue::String("[redacted]".to_string());
+            redacted = true;
+        }
     }
     redacted
 }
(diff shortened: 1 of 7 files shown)


fluree-db-api/src/ledger_info.rs
```diff
@@ -995,20 +995,43 @@ fn is_secret_config_key(key: &str) -> bool {
...
+/// Keys inside a secret-keyed reference object that identify where the secret
+/// comes from rather than what it is (the `ConfigValue::Dynamic` wire shape);
+/// their string values are preserved during subtree redaction.
+const SECRET_REF_SAFE_KEYS: &[&str] = &["env_var", "java_property"];
+
+fn is_secret_ref_safe_key(key: &str) -> bool {
+    let lower = key.to_ascii_lowercase();
+    SECRET_REF_SAFE_KEYS.iter().any(|s| *s == lower)
+}
+
```
aaj3f, line 1007: **Optional (residual risk, already tracked).** `SECRET_CONFIG_KEYS` is a key-name allowlist, so a secret stored under a key not on the list still prints in the clear — the container-shape hardening only triggers when the *outer* key is already a secret key.

This is the documented "fails open" behavior tracked in #1504, and this PR correctly brings the CLI to parity with the server rather than solving it here. No change requested; flagging only so the guarantee's edge is explicit in the review record.

fluree-db-api/src/ledger_info.rs
```diff
@@ -1026,6 +1049,37 @@ fn redact_json_secrets(value: &mut JsonValue) -> bool {
     redacted
 }
 
+/// Redact every scalar leaf beneath a secret-keyed container, keeping only
+/// reference-identifier strings ([`SECRET_REF_SAFE_KEYS`]) and nulls.
+fn redact_secret_subtree(value: &mut JsonValue) -> bool {
+    let mut redacted = false;
+    match value {
+        JsonValue::Object(map) => {
+            for (k, v) in map.iter_mut() {
```
aaj3f, line 1058: **Optional (note, not a defect).** `redact_secret_subtree` preserves a string value under any key named `env_var`/`java_property` at any depth inside a secret subtree.

That's the intended debugging affordance (those keys hold a *name*, not a value) and matches `ConfigValue::Dynamic`, so it's correct — worth only a mental note that a self-inflicted config storing a literal secret under a key literally named `env_var` would preserve it.

aaj3f (approved):
**✅ Approve.**

Well-executed credential-disclosure fix. The CLI's `info`/`iceberg info` print paths serialized stored graph-source configs with `to_string_pretty` and no redaction, dumping live OAuth2 client secrets and bearer tokens to the terminal, while the server route already redacted.

The fix consolidates all rendering into one module that routes through the existing fail-closed redactor, deletes the duplicated leak sites, and hardens the shared redactor for container-shaped secrets and unparseable configs — with the redaction property captured as a tested invariant so a future refactor that drops the call fails the suite.

I verified the denylist covers the two secrets that actually serialize (`token`, `client_secret`), that the env-var preservation matches the `ConfigValue` wire shape precisely, that no other CLI config-print path was missed, and that lossless persistence is untouched. The depth is right: it fixes at the API layer so CLI and server share one redactor.

**Adherence to repo commitments**
- **Patterns/abstractions:** ✔ Extends `redact_graph_source_config`/`build_generic_graph_source_info`; collapses three duplicated helpers into one.
- **Performance (speed first, memory second):** ✔ CLI display path only, off the query engine — no regression risk.
- **Testing:** ✔ New render-seam tests + API container-shape/empty-config tests; ran locally at HEAD → **147 passed, 0 failed**. Existing per-variant canaries and lossless-storage assertions still pass.
- **Conventions:** ⚠️ Clippy-clean (incl. 23 denied lints) and `fmt`-clean; only gap is the PR title / intermediate commit messages not using the `fix(cli):` conventional-commit prefix that feeds release notes. No AI-attribution trailers.

Verified locally at branch HEAD: traced secret-key coverage against the real `AuthConfig`/`ConfigValue` serialization, swept the CLI crate for missed print paths, confirmed no feature-gating coverage trap, re-ran the targeted redaction/render tests.

Recommend merge.

PR #1535, opened 2026-07-22
feat: Add auth token and model access disable CLI commands

bplatz:
## Summary

Addresses an internal user report on CLI gaps around credential handling and access-profile lifecycle:

- **`fluree auth token [--remote <name>]`** — prints exactly the stored access token to stdout for `.env`-style scripting (`FLUREE_TOKEN=$(fluree auth token)`). Previously the only way to get a token was `fluree config list`, which dumps the entire config including refresh tokens for every remote. Diagnostics (expired-token warning) go to stderr; refresh tokens are never printed by any command.

- **`fluree model access disable`** — the inverse of `enable`. Identifies the policy class the same way it was enabled (`--profile` + `--class` derives the default `{class}/access/{profile}` id, or `--policy-class` for overrides), deletes the policy class node and both owned policy nodes in one delete-only transaction (delete-if-exists, so re-running is a no-op), and supports `--dry-run`. With `--space`/`--remote` it first detaches the policy class from the space grant, mirroring enable's ordering rationale. It never posts an empty `policyClasses` list: on hosted stacks a class-less grant reads as instance-wide access, so dropping the last class would widen the grant instead of revoking it — the CLI leaves the grant unchanged (fail-closed stale reference) and says to delete the grant itself.

- **`fluree auth status` honest flow label** — status previously echoed the config discriminator `oidc_device` even when login fell back to authorization-code + PKCE, contradicting the "IdP does not support device flow" message in the same session. Login now records the flow that actually ran in a new optional `login_flow` field (cleared by logout and manual-token logins), and status shows `Auth type: oidc` plus `Login flow: device code` / `authorization code + PKCE`. The config discriminator is unchanged for back-compat.

- **`model access enable` grant message** — the final line now labels the reported access level and class count as the grant's merged state (cumulative across enables), not the profile just attached.

## Testing

- New unit tests: policy-class resolution (derivation, override, missing-identification errors), delete-only disable transaction shape, grant class extraction
- New integration tests: `auth token` prints exactly the token (stdout composability contract), errors without a stored token; `model access disable` dry-run output and identification requirement
- `cargo test -p fluree-db-cli` (151 lib + 118 integration) and `cargo clippy --all-targets -D warnings` on both touched crates pass

## Docs

- `docs/cli/auth.md`: `auth token` section, updated status output, security notes
- `docs/cli/model.md`: disable semantics including the last-policy-class grant guard

### DIFF
--- a/fluree-db-cli/src/config.rs
+++ b/fluree-db-cli/src/config.rs
@@ -1036,6 +1036,83 @@ mod tests {
         assert_eq!(result, fluree_dir.join("storage"));
     }
 
+    /// Regression guard for the read/write asymmetry of the remote config:
+    /// reads deserialize via serde, but writes go through the hand-rolled
+    /// `toml_edit` builder in `write_sync_config_toml` — a field missing
+    /// from that builder round-trips to `None` with green CI everywhere
+    /// else. Every `RemoteAuth` field is constructed here WITHOUT
+    /// `..Default::default()`, so adding a field breaks this test at
+    /// compile time and forces both the writer branch and this assertion
+    /// to be updated together.
+    #[tokio::test]
+    async fn remote_auth_round_trips_every_field_through_toml_store() {
+        use fluree_db_nameservice::RemoteName;
+        use fluree_db_nameservice_sync::{
+            OidcLoginFlow, RemoteAuth, RemoteAuthType, RemoteConfig, RemoteEndpoint,
+            SyncConfigStore,
+        };
+
+        let tmp = tempfile::tempdir().unwrap();
+        let store = TomlSyncConfigStore::new(tmp.path().to_path_buf());
+
+        let auth = RemoteAuth {
+            auth_type: Some(RemoteAuthType::OidcDevice),
+            token: Some("access-tok".into()),
+            issuer: Some("https://idp.example.com".into()),
+            client_id: Some("fluree-cli".into()),
+            exchange_url: Some("https://api.example.com/auth/exchange".into()),
+            refresh_token: Some("refresh-tok".into()),
+            scopes: Some(vec!["openid".into(), "offline_access".into()]),
+            redirect_port: Some(8400),
+            login_flow: Some(OidcLoginFlow::AuthCodePkce),
+        };
+        store
+            .set_remote(&RemoteConfig {
+                name: RemoteName::new("origin"),
+                endpoint: RemoteEndpoint::Http {
+                    base_url: "https://api.example.com/v1/fluree".into(),
+                },
+                auth: auth.clone(),
+                fetch_interval_secs: Some(30),
+            })
+            .await
+            .unwrap();
+
+        let read = store
+            .get_remote(&RemoteName::new("origin"))
+            .await
+            .unwrap()
+            .expect("remote persisted")
+            .auth;
+
+        assert_eq!(read.auth_type, auth.auth_type);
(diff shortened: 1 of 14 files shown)


fluree-db-cli/src/config.rs
```diff
@@ -686,6 +687,13 @@ impl TomlSyncConfigStore {
                     if let Some(port) = auth.redirect_port {
                         auth_table.insert("redirect_port", Value::from(i64::from(port)).into());
                     }
+                    if let Some(flow) = auth.login_flow {
```
aaj3f, line 690: **Optional (regression guard for a documented footgun).** The `login_flow` persistence is correct, but it depends on this hand-rolled `toml_edit` writer staying in sync with the serde-read form — the exact asymmetry your own new memory entry flags as a "silently drops the field" trap.

There's no test that round-trips a `RemoteAuth { login_flow: Some(..) }` through `set_remote` → `get_remote`, so the next added field (or a rename of these enum strings) could reintroduce the silent-drop with green CI.

A small round-trip test over the config store would lock in what the memory documents. Not blocking — this PR handled the gotcha correctly.

aaj3f (approved):
**✅ Approve.**

Two well-executed CLI additions plus two honesty fixes. `auth token` is the right shape for `.env` scripting — exactly the access token on stdout, diagnostics on stderr, refresh token never printed — and pinned by an integration test asserting the composability contract. `model access disable` is the careful one: I verified its fail-closed grant-detach never POSTs an empty `policyClasses` list (which on a hosted stack would widen the grant to instance-wide instead of revoking), that it detaches before deleting policies, that re-runs are no-ops (delete-if-exists), and that `resolve_policy_class` mints the identical id `enable` derives so it deletes the right nodes. The `login_flow` honesty fix is correctly threaded through the hand-rolled TOML writer that the codebase's own memory warns is a silent-drop trap, with the written enum strings matching serde's snake_case read form.

- **Patterns/abstractions:** ✔ Extracts a shared `GrantsApi` (resolve/space_grant/upsert) that both attach and detach reuse, rather than duplicating the grants-endpoint plumbing.
- **Performance (speed first, memory second):** ✔ CLI/config paths only, off the query engine. No regression.
- **Testing:** ✔ Unit (policy-class resolution, delete-only txn shape, access-transition provenance, class extraction) + integration (`auth token` exact-token + error, disable dry-run + identification); `cargo test -p fluree-db-cli` green in CI. ⚠️ No round-trip test for `login_flow` persistence (inline note).
- **Conventions:** ✔ `feat:`/`fix:`/`docs:` conventional commits; clippy `-D warnings` clean both touched crates; `cli/auth.md` + `cli/model.md` updated; no AI-attribution.

Verified locally at branch HEAD: traced the token stdout/stderr split and refresh-token non-use, the empty-`policyClasses` fail-closed guard and detach/delete ordering, the `resolve_policy_class`↔`enable` id parity, and the `login_flow` hand-rolled-writer wiring + serde round-trip; PR CI fully green.

Recommend merge.

bplatz:
Added the suggested round-trip guard in `config.rs` tests — with one twist beyond the ask: the `RemoteAuth` is constructed exhaustively (no `..Default::default()`), so adding a field now fails this test at compile time, forcing the toml_edit writer branch and the assertions to be updated together. Also covers the second write cycle (read → modify → rewrite), which is the pass where a writer-omitted field actually disappears.

PR #1537, opened 2026-07-22
fix(proxy-storage): resolve advanced index kinds over a remote mount

### LINKED ISSUE #1536: Remote mount reverse-parser drops advanced index kinds (stats, spatial, history, graph-sources)
## Summary

The remote-mount CAS reverse-parser `cid_and_ledger_from_address` (`fluree-db-nameservice-sync/src/proxy_storage.rs`) recognizes only the core commit + index-tree content kinds. A mounted ledger's **advanced** index artifacts fail to resolve over the mount — the parser returns `None` and the fetch raises `"Cannot derive CID from address"`:

| Kind | Forward `content_path` layout |
|---|---|
| `StatsSketch` (planner HLL/NDV) | `{ledger}/index/stats/{h}.hll` |
| `SpatialIndex` | `{ledger}/index/spatial/{h}.bin` |
| `HistorySidecar` (time-travel) | `{ledger}/index/objects/history/{h}.fhs1` |
| `GraphSourceSnapshot` (BM25/vector) | `graph-sources/{ledger}/snapshots/{h}.gssnap` |
| `GraphSourceMapping` (R2RML) | `graph-sources/{ledger}/mapping/{h}.ttl` |

Basic JSON-LD/SPARQL works (commit + index roots/branches/leaves/dicts resolve); anything that drives those artifacts — stats-based query planning, time-travel/history, spatial, and full-text/vector graph-source queries — hard-fails over a remote mount.

A second, coupled gate compounds it: the origin object endpoint's `is_allowed_object_kind` allowlist (`fluree-db-server/src/routes/storage_proxy.rs`) 404s these same kinds, so even a correct parser wouldn't get the bytes.

## Two classes of "missing"

- **5 recoverable kinds** (above) have distinct forward `content_path` layouts, so the reverse parser can be made exhaustive. **This is the fix.**
- **4 edge-annotation kinds** (`AnnotationForward/Reverse` × `Branch/Leaf`) are forward-*lossy*: `content_path` routes all four to a shared `{prefix}/blob/{h}.bin` catch-all with no codec discriminator, so no reverse parser can recover them without a forward-layout change (+ reindex of existing annotated ledgers). Out of scope here.

## Fix

Add the 5 recoverable kinds to the reverse-parser (handling the asymmetric `graph-sources/…` layout, where the ledger is not the leading path segment) and widen the origin `is_allowed_object_kind` allowlist to match, with a table-driven round-trip test over every `content_path` kind that also pins the annotation gap.

_Surfaced dress-rehearsing fluree/solo's remote dataset catalog (solo#818) end-to-end._

aaj3f:
Closes #1536.

## Problem

The remote-mount CAS reverse-parser `cid_and_ledger_from_address` recognized only the core commit + index-tree content kinds, so a mounted ledger's advanced index artifacts — planner statistics (HLL/NDV), spatial index, history sidecars, and graph-source snapshots/mappings — failed to resolve (parser → `None` → "Cannot derive CID from address"). Surfaced dress-rehearsing fluree/solo's remote dataset catalog (solo#818) end-to-end.

## Change (scope updated per review)

1. **Reverse-parser** (`fluree-db-nameservice-sync/src/proxy_storage.rs`): add all 5 recoverable kinds. Four share the `{ledger}/…` layout; the two graph-source kinds lead with a literal `graph-sources` segment, so the match yields an `alias_start` (0 normally, 1 for graph-sources). Per review: the gs arms now carry an explicit contract comment — the recovered "ledger" is a **graph_source_id** (what `content_path` was given at write time), which the object endpoint cannot resolve yet.
2. **Origin allowlist** (`fluree-db-server/.../storage_proxy.rs`): widen `is_allowed_object_kind` to the **three ledger-scoped** advanced kinds only — `StatsSketch`, `SpatialIndex`, `HistorySidecar`. Review established the graph-source half was inert server-side (`resolve_block_ledger` → `ns.lookup()` deliberately skips gs records → 404 at step 3; and the token-scope semantics for multi-dependency graph sources are an open design decision), so the gs kinds stay excluded and their serving is split out to **#1539**. The doc also states the Filtered-tier asymmetry (`is_allowed_block_kind` excludes all advanced kinds) is deliberate per the v1 all-or-nothing design.
3. **Shared layout const**: `GRAPH_SOURCES_PATH_SEGMENT` in `fluree-db-core` next to `content_path`, used by the forward layout and the reverse-parser; `fluree-db-nameservice::STORAGE_SEGMENT_GRAPH_SOURCES` aliases it — forward and reverse can no longer drift.
4. **Re-export** `CODEC_FLUREE_SPATIAL_INDEX` + `CODEC_FLUREE_GRAPH_SOURCE_MAPPING` from `fluree-db-core`.

The 4 edge-annotation kinds stay unrecoverable **by design** (`content_path` routes them to a shared `blob/.bin` catch-all with no codec discriminator); a table-driven test pins that gap.

## Tests

- Parser: 5 per-kind round-trips + multi-segment graph-source case + table-driven exhaustiveness guard (mutation-verified).
- **Object-endpoint** (new, per review — the mount E2E rides `POST /storage/block`, which is deliberately unchanged): stats/spatial/history round-trip 200 byte-identically through `GET /storage/objects/{cid}` with correct kind headers; graph-source kinds pinned 404 **even with the bytes present in storage and a `storage.all` token**; plus a unit pin of `is_allowed_object_kind` itself (a re-widened list could still 404 downstream in the integration test, so the gate is asserted directly).
- `cargo test`: nameservice-sync 81, server lib 130, proxy_integration 29, core 699, nameservice 155 — all green; clippy + fmt clean.

## Follow-ups

- **#1539** — serve graph-source artifacts over the object endpoint: dual-value resolution (gs id for address reconstruction; real ledger(s) for authz + serving gate) and the explicit decision on which dependency ledger's scope governs. Client side is ready in this PR.
- **fluree/solo#841**: the Solo origin's own allowlist (`is_cas_servable`) mirrors this PR — widened to the same three kinds, gs kinds pinned excluded in lockstep.
- Fetchable ≠ works: planner stats should degrade, not hard-fail, when absent — worth one E2E per advanced feature over a live mount.


fluree-db-server/src/routes/storage_proxy.rs
```diff
@@ -618,6 +623,11 @@ fn is_allowed_object_kind(kind: ContentKind) -> bool {
             | ContentKind::IndexBranch
             | ContentKind::IndexLeaf
             | ContentKind::DictBlob { .. }
+            | ContentKind::StatsSketch
+            | ContentKind::SpatialIndex
+            | ContentKind::HistorySidecar
+            | ContentKind::GraphSourceSnapshot
+            | ContentKind::GraphSourceMapping
```
bplatz: The three ledger-scoped kinds (StatsSketch / SpatialIndex / HistorySidecar) look complete with this widening — their addresses carry a real ledger alias, so `resolve_block_ledger` → auth → serving gate → `content_address` all work as they do for leaves.

For the two graph-source kinds, though, I don't think a request survives step 3 of this handler. Snapshots/mappings are written under a **graph_source_id**, not a ledger id (e.g. `bm25.rs` passes `graph_source_id` to `content_address`), so the client-side parser recovers that id and sends it as `?ledger=`. `resolve_block_ledger` resolves via `ns.lookup()`, which intentionally returns `Ok(None)` for graph-source records (#1369 — see `file.rs` `lookup()` and the `lookup_skips_graph_source_records` test), and the `DictBlob` fallback doesn't apply → 404 before this allowlist is consulted. Unless there's another resolution path I'm missing, the graph-source half of the fix is inert server-side — worth a look before merge.

If it does need fixing here, note it's more than swapping in `lookup_any()`: `effective_ledger` currently serves two conflicting purposes. Address reconstruction below (`content_address(method, kind, &effective_ledger, ...)`) must use the *graph-source id* to rebuild the `graph-sources/…` path, while `is_authorized_for_ledger` and `effective_serving_cached` need a *real ledger* (token `fluree.storage.ledgers` scopes hold ledger aliases — as written, a gs fetch would only ever pass with a `storage.all` token — and the serving gate resolves a ledger's `f:serveBlocks` posture). So the resolution step likely needs to return both values: authorize/gate against the graph source's source ledger, keep the gs id for the address. That also implies a small deliberate semantics decision — which ledger's storage scope governs a graph source's artifacts — that seems better made explicitly than fallen into.

fluree-db-server/src/routes/storage_proxy.rs
```diff
@@ -601,13 +601,18 @@ pub struct ObjectQuery {
...
 /// Content kinds allowed through the CID object endpoint.
 ///
-/// All replication-relevant kinds are served (commits, txns, config, and
-/// index artifacts). Only `GarbageRecord` (internal GC metadata) is excluded.
+/// All replication- and mount-relevant kinds are served: commits, txns, config,
+/// and every index artifact a remote mount may resolve from the index root —
+/// the core index tree (roots/branches/leaves), dict blobs, and the advanced
+/// index kinds (planner statistics, spatial, history sidecars, and graph-source
+/// snapshots/mappings). `GarbageRecord` (internal GC metadata) and the
+/// edge-annotation arenas are excluded — neither is fetched over a mount.
```
bplatz: Adjacent asymmetry worth one sentence here (or in the PR description): the `POST /storage/block` path's allowlist (`is_allowed_block_kind` in `fluree-db-api/src/block_fetch.rs`) still excludes all five of these kinds, so a mount in `ProxyReadMode::Filtered` continues to fail on advanced artifacts exactly as before. I believe that's intentional per the v1 all-or-nothing serving design (the filtered tier is transport-only today, and the api-side test asserting `!is_allowed_block_kind(GraphSourceSnapshot)` pins it) — but stating it as deliberate would keep someone from "fixing" the mismatch casually later. Worth a look to confirm the intent matches.

fluree-db-nameservice-sync/src/proxy_storage.rs
```diff
@@ -111,51 +114,71 @@ fn cid_and_ledger_from_address(address: &str) -> Option<(ContentId, String)> {
...
-        (CODEC_FLUREE_DICT_BLOB, n - 3, false)
+        (CODEC_FLUREE_DICT_BLOB, 0, n - 3, false)
+    } else if n >= 5 && parts[0] == "graph-sources" && parts[n - 2] == "snapshots" {
+        // GraphSourceSnapshot: graph-sources/{name}/{branch}/snapshots/{hash}.gssnap
+        // The ledger is not the leading segment here, so the alias slice
+        // starts at 1 (after the literal `graph-sources` prefix).
+        (CODEC_FLUREE_GRAPH_SOURCE_SNAPSHOT, 1, n - 2, true)
+    } else if n >= 5 && parts[0] == "graph-sources" && parts[n - 2] == "mapping" {
+        // GraphSourceMapping: graph-sources/{name}/{branch}/mapping/{hash}.ttl
+        (CODEC_FLUREE_GRAPH_SOURCE_MAPPING, 1, n - 2, true)
```
bplatz: These arms match the forward layouts in `content_path` exactly (verified against `storage.rs`), and the `alias_start` handling is right for multi-segment names.

One naming caveat that ties into the server-side question: for these two kinds, the "ledger" this function recovers is actually a **graph_source_id** — that's what `content_path` was given at write time. Callers (and the server's `?ledger=` param) treat it as a ledger alias, which is where the object endpoint's resolution appears to break down (see comment there). A short comment here noting the recovered alias is a graph-source id for these arms would make that contract visible.

Optional hardening: the literal `"graph-sources"` now lives in three places (here, `content_path`, and `STORAGE_SEGMENT_GRAPH_SOURCES` in `fluree-db-nameservice`). A shared const next to `content_path` in core would keep forward and reverse from drifting — the exhaustiveness test below already catches drift at test time, so this is purely nice-to-have.

fluree-db-nameservice-sync/src/proxy_storage.rs
```diff
@@ -704,6 +727,109 @@ mod tests {
...
+    /// distinct address MUST survive the mount reverse-parser (forward → reverse
+    /// recovers the same CID + ledger). The four annotation kinds are a KNOWN
+    /// forward-side gap — `content_path` routes all four to the shared
+    /// `{prefix}/blob/{hash}.bin` catch-all, which carries no codec
+    /// discriminator, so no reverse parser can recover them until the forward
+    /// layout gives them distinct paths. This test pins both facts: adding a
+    /// kind (or giving annotations distinct paths) without updating the parser
+    /// fails here instead of silently shipping a mount hole.
+    #[test]
+    fn test_all_content_path_kinds_roundtrip_or_are_documented_gaps() {
```
bplatz, line 794: This exhaustiveness guard is a great addition — pinning the annotation-kind gap so it can't regress silently is exactly the right move.

Coverage gap on the other side of the wire, though: nothing exercises the new kinds through `GET /storage/objects/{cid}`, which is the endpoint this PR changes. The existing proxy_integration graph-source test (`test_storage_proxy_rejects_graph_source_cids`) targets `POST /storage/block`, whose allowlist is untouched — so "mount E2E stays green" doesn't tell us the new kinds are actually served. Suggest object-endpoint tests per kind: stats/spatial/history should 200 and round-trip bytes; graph-source should pin whatever behavior is decided (my reading is it currently 404s in `resolve_block_ledger` — see the comment on the allowlist). A stats-over-mount case would have been enough to catch the graph-source issue's shape, so this feels like cheap insurance.

bplatz (commented):
Reviewed with a focus on whether each of the five kinds actually round-trips end-to-end over a mount. The parser work checks out — I verified every new arm against the `content_path` layouts in `fluree-db-core/src/storage.rs`, and the `alias_start` refactor handles the asymmetric `graph-sources/` layout correctly, including multi-segment ledger names. The exhaustiveness-guard test pinning the annotation gap is a genuinely nice addition.

My one substantive concern (detail inline on the allowlist): I don't believe the two graph-source kinds work end-to-end yet. Their addresses carry a *graph-source id*, and `resolve_block_ledger` on the object endpoint only resolves real ledgers — `ns.lookup()` deliberately returns `None` for graph-source records (#1369) — so those fetches appear to 404 at step 3 before the widened allowlist ever matters. The three ledger-scoped kinds (stats, spatial, history sidecar) look complete and correct.

If that reading is right, I'd suggest either (a) splitting the graph-source kinds out and landing the ledger-scoped three now — they're valuable on their own — or (b) extending the server-side resolution/authorization for graph sources in this PR, with object-endpoint tests proving it. Related: the mount E2E that stays green exercises `POST /storage/block`, which this PR doesn't change, so it doesn't cover the changed endpoint either way (inline note on the tests).

Happy to be wrong on the graph-source path — if there's a resolution step I'm not seeing, a quick pointer (or better, an object-endpoint test for a `.gssnap` CID) would settle it.

bplatz (approved):
Some inline comments / suggestions

fluree-db-server/src/routes/storage_proxy.rs
aaj3f: Confirmed — you read it exactly right, and thanks for the precise trace. I verified all three legs: the `lookup()` guard in file.rs returns `Ok(None)` for graph-source records, bm25/r2rml mint under `graph_source_id`, and `GraphSourceRecord.dependencies` is a `Vec<String>` — so the governing-scope question is genuinely multi-answer.

Took option (a) in 32606c5d3: the allowlist is narrowed to StatsSketch / SpatialIndex / HistorySidecar, and gs serving is split out to #1539. That issue carries your dual-purpose `effective_ledger` analysis (gs id for `content_address` reconstruction, real ledger(s) for `is_authorized_for_ledger` + `effective_serving_cached`) and your point that as drafted a gs fetch could only ever have passed under `storage.all` — the semantics decision is named there so it gets made explicitly.

fluree-db-server/src/routes/storage_proxy.rs
aaj3f: Agreed it's intentional — the filtered tier is transport-only today, and the v1 all-or-nothing design keeps advanced kinds out of it.

Stated explicitly now (32606c5d3): the `is_allowed_object_kind` doc names the `is_allowed_block_kind` asymmetry as deliberate, and the PR description carries the same sentence, so nobody "fixes" the mismatch casually.

fluree-db-nameservice-sync/src/proxy_storage.rs
aaj3f: Both done in 32606c5d3.

The two gs arms now carry the contract comment: the recovered "ledger" is a **graph_source_id** (what `content_path` was given at write time), the object endpoint can't resolve it today, and parsing it remains forward-compatible with #1539's server side.

The literal now lives once: `GRAPH_SOURCES_PATH_SEGMENT` in fluree-db-core next to `content_path`, used by the forward layout and this parser, with `STORAGE_SEGMENT_GRAPH_SOURCES` in fluree-db-nameservice aliased to it — forward and reverse can no longer drift.

fluree-db-nameservice-sync/src/proxy_storage.rs
aaj3f: Right on both counts — the green mount E2E rode `POST /storage/block` and proved nothing about the changed endpoint.

Added in 32606c5d3, all through `GET /storage/objects/{cid}`: stats/spatial/history round-trip 200 byte-identically (kind headers asserted); the `.gssnap` 404 is pinned **even with the bytes present in storage under their gs id and a `storage.all` token** (mapping kind too).

Also a direct unit pin of `is_allowed_object_kind` itself — a re-widened list could still 404 downstream in the integration test for the wrong reason (unresolvable gs id, missing bytes), so the gate is asserted on its own.

aaj3f (commented):
Response to the graph-source concern: you were right, and it's addressed in 32606c5d3.

Verified server-side before changing anything — `ns.lookup()`'s guard in file.rs deliberately returns `Ok(None)` for graph-source records, so a gs fetch died at step 3 regardless of the allowlist; and `GraphSourceRecord.dependencies` is a `Vec`, which makes "whose storage scope governs a gs artifact" a real design decision rather than a `lookup_any()` swap.

Went with your option (a): `is_allowed_object_kind` now serves only the three ledger-scoped kinds, and graph-source serving is split out to #1539 — which captures the dual-value resolution shape (gs id for address reconstruction, real ledger(s) for authz + the serving gate) and the explicit dependency-scope decision. The parser arms stay, with the graph_source_id contract documented inline, so the client already forms the right request the day the server learns to answer it.

Also took the test suggestion: object-endpoint coverage per kind (details on the inline thread), plus a direct unit pin of the allowlist. And mirrored the narrowing on the Solo origin's `is_cas_servable` (fluree/solo#841) so both origins move in lockstep.

PR #1541, opened 2026-07-24
Substitute property-position variables in datalog rules

### LINKED ISSUE #1531: Datalog rule is not replacing variables in property-position with bound values
I tried using a datalog rule to insert property-value pairs defined in the graph,
but I found when I have an insert pattern like this:
```json
{
  ...,
  "insert": {"@id": "?foo", "?bar": "?baz"}
}
```
then only `?foo` and `?baz` seem to be properly replaced, and the property that's inserted is literally named '?bar'.

zonotope:
Issue #1531 reported that variables in property position of a datalog rule's `insert` pattern were not replaced with their bound values: a rule like `{"insert": {"@id": "?foo", "?bar": "?baz"}}` substituted `?foo` and `?baz` but emitted the literal string `?bar` as the property of the derived fact.

The root cause was in the JSON-LD rule parser, not the substitution logic. `parse_node_pattern` in `fluree-db-query/src/datalog_rules.rs` unconditionally resolved every node-map key as an IRI, so a `"?bar"` key was encoded as a `RuleTerm::Sid` for the literal IRI `?bar`. Everything downstream already supported predicate variables — the where-clause matcher binds them, `instantiate_pattern` substitutes them, and the SPARQL rule-lowering path (`fluree-db-api/src/sparql_lang.rs`) maps them correctly — the JSON-LD parser just never produced one. This broke both clauses: a property variable in `where` matched nothing (its literal-IRI predicate exists on no triples), and one in `insert` derived facts under the bogus literal predicate.

## Changes

- `fluree-db-query/src/datalog_rules.rs`:
  - New `parse_predicate_term` helper: a `?`-prefixed node-map key parses as `RuleTerm::var`; anything else expands and resolves as an IRI. Used at all three pattern-construction sites in `parse_node_pattern`, so predicate variables now work in both `where` and `insert` clauses.
  - Dropped `parse_object_value`'s two already-unused parameters, which the change orphaned.
  - A predicate or subject variable that ends up bound to a non-Sid value (e.g. `?p` bound to a string via an object position, then reused as a predicate) now resolves to a new `TermResolution::Incompatible` variant and the pattern matches nothing for that binding row, instead of returning an error. Previously that error propagated out of rule execution and caused the reasoning layer to silently discard every datalog-derived fact for the query; a literal can never occupy subject/predicate position, so an empty join is the correct semantics. A literal constant in those positions (a malformed rule) still errors.
  - Removed the `?`-prefix branch in `expand_iri` that the parser change made unreachable (both remaining callers check for variables before calling it).

## Tests

- `datalog_rule_insert_property_variable_substituted` — the issue's verbatim shape: `?rel` bound to a predicate IRI via an object reference in `where`, used as the property in `insert`; the derived fact must appear under the resolved predicate.
- `datalog_rule_where_property_variable_binds_and_substitutes` — a copy-properties rule binding `?prop` in `where` and reusing it in `insert`.
- `datalog_rule_literal_bound_property_variable_does_not_abort_other_rules` — a rule whose predicate variable binds to a string must derive nothing itself while a sibling rule's derivations survive.
- A unit test pinning that a literal-bound subject/predicate variable resolves to `Incompatible` rather than an error.

Fixes #1531

### DIFF
--- a/fluree-db-query/src/datalog_rules.rs
+++ b/fluree-db-query/src/datalog_rules.rs
@@ -769,12 +769,12 @@ async fn match_pattern_with_bindings(
     // variable bound to a non-Sid value can never occupy that position in a
     // flake, so the pattern matches nothing for this binding row.
     let Some((subject_sid, subject_var)) =
-        resolve_term_with_bindings(&pattern.subject, bindings)?.split()
+        resolve_term_with_bindings(&pattern.subject, bindings).split()
     else {
         return Ok(Vec::new());
     };
     let Some((predicate_sid, predicate_var)) =
-        resolve_term_with_bindings(&pattern.predicate, bindings)?.split()
+        resolve_term_with_bindings(&pattern.predicate, bindings).split()
     else {
         return Ok(Vec::new());
     };
@@ -862,8 +862,9 @@ enum TermResolution {
     Bound(Sid),
     /// Unbound variable, to be bound from each matched flake
     Unbound(Arc<str>),
-    /// Variable bound to a non-Sid value: no flake can carry it in this
-    /// position, so the pattern matches nothing (an empty join, not an error)
+    /// A literal constant, or a variable bound to a non-Sid value: no flake
+    /// can carry a literal in subject/predicate position, so the pattern
+    /// matches nothing (an empty join, not an error)
     Incompatible,
 }
 
@@ -878,18 +879,22 @@ impl TermResolution {
     }
 }
 
-/// Resolve a subject- or predicate-position term using existing bindings
-fn resolve_term_with_bindings(term: &RuleTerm, bindings: &Bindings) -> Result<TermResolution> {
+/// Resolve a subject- or predicate-position term using existing bindings.
+///
+/// Total: a literal constant (`RuleTerm::Value`) — a plausible typo for an
+/// unprefixed IRI like `{"@id": "Alice"}` — can no more occupy subject or
+/// predicate position than a variable bound to a literal, so it resolves to
+/// `Incompatible` (that pattern matches nothing) rather than erroring, which
+/// would abort the whole fixpoint and drop every rule's derivations.
+fn resolve_term_with_bindings(term: &RuleTerm, bindings: &Bindings) -> TermResolution {
     match term {
-        RuleTerm::Sid(sid) => Ok(TermResolution::Bound(sid.clone())),
+        RuleTerm::Sid(sid) => TermResolution::Bound(sid.clone()),
         RuleTerm::Var(var) => match bindings.get(var.as_ref()) {
-            Some(BindingValue::Sid(sid)) => Ok(TermResolution::Bound(sid.clone())),
-            Some(_) => Ok(TermResolution::Incompatible),
-            None => Ok(TermResolution::Unbound(var.clone())),
+            Some(BindingValue::Sid(sid)) => TermResolution::Bound(sid.clone()),
+            Some(_) => TermResolution::Incompatible,
+            None => TermResolution::Unbound(var.clone()),
         },
-        RuleTerm::Value(_) => Err(QueryError::InvalidQuery(
-            "Literal value not allowed in subject/predicate position".to_string(),
-        )),
+        RuleTerm::Value(_) => TermResolution::Incompatible,
     }
 }
 
@@ -1327,7 +1332,21 @@ mod tests {
         let mut bindings = Bindings::new();
         bindings.insert(Arc::from("?p"), BindingValue::String("blue".to_string()));
 
-        let resolved = resolve_term_with_bindings(&RuleTerm::var("?p"), &bindings).unwrap();
+        let resolved = resolve_term_with_bindings(&RuleTerm::var("?p"), &bindings);
+        assert!(matches!(resolved, TermResolution::Incompatible));
+        assert!(resolved.split().is_none());
+    }
+
+    #[test]
+    fn resolve_term_literal_constant_is_incompatible() {
+        // A literal constant in subject/predicate position (e.g. a typo like
+        // `{"@id": "Alice"}` parsed as a string value) can never match a
+        // flake; resolution must report an empty join, not an error, so the
+        // authoring mistake cannot abort the whole fixpoint.
+        let bindings = Bindings::new();
+        let literal = RuleTerm::Value(RuleValue::String("Alice".to_string()));
+
+        let resolved = resolve_term_with_bindings(&literal, &bindings);
         assert!(matches!(resolved, TermResolution::Incompatible));
         assert!(resolved.split().is_none());
     }
--- a/fluree-db-query/src/datalog_rules.rs
+++ b/fluree-db-query/src/datalog_rules.rs
@@ -405,25 +405,16 @@ fn parse_node_pattern(
             continue;
         }
 
-        // Resolve predicate IRI
-        let predicate_iri = expand_iri(key, context)?;
-        let predicate_sid = resolve_iri(&predicate_iri, snapshot)?;
+        let predicate = parse_predicate_term(key, context, snapshot)?;
 
         // Parse object(s)
         match value {
             JsonValue::Array(arr) => {
                 for item in arr {
-                    let obj = parse_object_value(
-                        item,
-                        context,
-                        snapshot,
-                        patterns,
-                        &subject,
-                        &predicate_sid,
-                    )?;
+                    let obj = parse_object_value(item, context, snapshot, patterns)?;
                     patterns.push(RuleTriplePattern {
                         subject: subject.clone(),
-                        predicate: RuleTerm::Sid(predicate_sid.clone()),
+                        predicate: predicate.clone(),
                         object: obj,
                     });
                 }
@@ -441,7 +432,7 @@ fn parse_node_pattern(
                 // Add pattern linking parent to nested subject
                 patterns.push(RuleTriplePattern {
                     subject: subject.clone(),
-                    predicate: RuleTerm::Sid(predicate_sid.clone()),
+                    predicate: predicate.clone(),
                     object: nested_subject.clone(),
                 });
 
@@ -460,7 +451,7 @@ fn parse_node_pattern(
                 let obj = parse_term(value, context, snapshot)?;
                 patterns.push(RuleTriplePattern {
                     subject: subject.clone(),
-                    predicate: RuleTerm::Sid(predicate_sid.clone()),
+                    predicate: predicate.clone(),
                     object: obj,
                 });
             }
@@ -470,14 +461,27 @@ fn parse_node_pattern(
     Ok(())
 }
 
+/// Parse a node-map key into a predicate term: a `?`-prefixed key is a
+/// variable; anything else expands and resolves as an IRI.
+fn parse_predicate_term(
+    key: &str,
+    context: &JsonValue,
+    snapshot: &LedgerSnapshot,
+) -> Result<RuleTerm> {
+    if key.starts_with('?') {
+        Ok(RuleTerm::var(key))
+    } else {
+        let iri = expand_iri(key, context)?;
+        Ok(RuleTerm::Sid(resolve_iri(&iri, snapshot)?))
+    }
+}
+
 /// Parse an object value, handling nested structures
 fn parse_object_value(
     value: &JsonValue,
     context: &JsonValue,
     snapshot: &LedgerSnapshot,
     patterns: &mut Vec<RuleTriplePattern>,
-    _parent_subject: &RuleTerm,
-    _predicate_sid: &Sid,
 ) -> Result<RuleTerm> {
     match value {
         JsonValue::Object(nested) if nested.contains_key("@id") => {
--- a/fluree-db-query/src/datalog_rules.rs
+++ b/fluree-db-query/src/datalog_rules.rs
@@ -23,12 +23,14 @@ use fluree_db_core::value::FlakeValue;
 use fluree_db_core::{GraphDbRef, LedgerSnapshot, Sid};
 use fluree_db_reasoner::{
     BindingValue, Bindings, CompareOp, DatalogRule, DatalogRuleSet, DerivedFactsBuilder,
-    FrozenSameAs, RuleFilter, RuleTerm, RuleTriplePattern, RuleValue,
+    FrozenSameAs, ReasoningBudget, ReasoningDiagnostics, RuleFilter, RuleTerm, RuleTriplePattern,
+    RuleValue,
 };
 use fluree_vocab::namespaces::FLUREE_DB;
 use serde_json::Value as JsonValue;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
+use std::time::Instant;
 
 use crate::error::{QueryError, Result};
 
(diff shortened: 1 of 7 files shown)


fluree-db-query/src/datalog_rules.rs
```diff
@@ -847,26 +856,37 @@ async fn match_pattern_with_bindings(
...
+fn resolve_term_with_bindings(term: &RuleTerm, bindings: &Bindings) -> Result<TermResolution> {
+    match term {
+        RuleTerm::Sid(sid) => Ok(TermResolution::Bound(sid.clone())),
+        RuleTerm::Var(var) => match bindings.get(var.as_ref()) {
+            Some(BindingValue::Sid(sid)) => Ok(TermResolution::Bound(sid.clone())),
+            Some(_) => Ok(TermResolution::Incompatible),
+            None => Ok(TermResolution::Unbound(var.clone())),
+        },
         RuleTerm::Value(_) => Err(QueryError::InvalidQuery(
             "Literal value not allowed in subject/predicate position".to_string(),
```
bplatz: Literal constants in subject position still abort the entire fixpoint, which is the same failure mode this PR fixes for bindings.

`parse_term` (:523) maps any colon-less, non-`?` string to `RuleTerm::Value`, so a where-clause `{"@id": "Alice"}` — a plausible typo for `ex:Alice` — lands here. The error propagates out of `execute_rule_matching` (:1223) and is swallowed at `reasoning_prep.rs:361`, dropping every rule's derivations, not just this rule's.

Confirmed locally: a ledger with a working grandparent rule plus one rule using `"@id": "Alice"` returns zero rows. A literal can no more be a subject than a predicate, so `Incompatible` seems right here too — or drop the rule at parse time, as `extract_datalog_rules` already does for parse failures.

fluree-db-query/src/datalog_rules.rs
```diff
@@ -766,9 +765,19 @@ async fn match_pattern_with_bindings(
...
-    let (predicate_sid, predicate_var) = resolve_term_with_bindings(&pattern.predicate, bindings)?;
+    // Resolve pattern terms using existing bindings. A subject/predicate
+    // variable bound to a non-Sid value can never occupy that position in a
+    // flake, so the pattern matches nothing for this binding row.
+    let Some((subject_sid, subject_var)) =
+        resolve_term_with_bindings(&pattern.subject, bindings)?.split()
+    else {
+        return Ok(Vec::new());
+    };
+    let Some((predicate_sid, predicate_var)) =
```
bplatz, line 850: Unbound predicate variables make the `(None, None)` arm of `choose_index_and_match` (:1048) reachable from JSON-LD rules for the first time — a full SPOT scan of the ledger, per its own comment.

`execute_rule_matching` doesn't reorder patterns, so a rule whose first where pattern is `{"@id": "?s", "?p": "?o"}` full-scans on every fixpoint iteration, against a growing overlay. This path has no fuel tracker and no time or fact budget; `max_iterations` (:1257) is the only bound. Worth rejecting an all-unbound leading pattern, or hoisting a bound pattern to the front.

(Unrelated, but docs/query/datalog-rules.md:335 advertises max facts / max time / max memory budgets that don't exist.)

bplatz (commented):
Fix looks right — `parse_predicate_term` is the correct choke point, and the `?` branch in `expand_iri` really is unreachable now (both remaining callers check for variables first). fmt and clippy clean; `grp_reasoning` and `grp_query_reason` pass, including the four new tests.

Two inline notes, plus:

- `docs/query/datalog-rules.md` should be updated in this changeset. The Where clause (L47) and Insert clause (L79) sections enumerate the supported pattern shapes and don't mention property-position variables. The insert bullet "Variables are replaced with the bound values from `where`" currently reads as a description of the old broken behavior.
- Minor: `DatalogRule::new` (fluree-db-reasoner/src/datalog.rs:147,155) builds `depends_on`/`generates` by filtering on `RuleTerm::Sid`, so variable predicates contribute nothing to either. `generates` has no consumers and `depends_on` only feeds the sort key at :287, so under a fixpoint this is cosmetic — but a var-predicate rule now always sorts first as zero-dependency, which is backwards for the copy-properties shape in the new test.

fluree-db-query/src/datalog_rules.rs
```diff
@@ -470,14 +461,27 @@ fn parse_node_pattern(
...
 }
 
+/// Parse a node-map key into a predicate term: a `?`-prefixed key is a
+/// variable; anything else expands and resolves as an IRI.
+fn parse_predicate_term(
+    key: &str,
+    context: &JsonValue,
+    snapshot: &LedgerSnapshot,
+) -> Result<RuleTerm> {
+    if key.starts_with('?') {
```
aaj3f, line 473: Real location: `fluree-db-query/src/datalog_rules.rs:1046-1052` and `fluree-db-query/src/execute/reasoning_prep.rs:322` — neither is in this diff. Anchoring here because this line is what makes them reachable.

*Blocking.* **CRITICAL: this makes an unbudgeted, unbounded full-index scan broadly reachable, on the one reasoning path that has no fact or time ceiling.**

An unbound predicate variable in a leading `where` pattern resolves to `(None, None)` in `choose_index_and_match` (`:1046-1052`), whose own comment reads `// Nothing bound: scan all (expensive!)`. `execute_rule_matching` (`:607-630`) does not reorder patterns, so a rule beginning `where [{"@id": "?s", "?p": "?o"}, …]` full-scans SPOT once per rule per fixpoint iteration, against an overlay that grows every iteration.

What makes this blocking rather than a known wart is the asymmetry sitting fifteen lines away in the same function. `compute_derived_facts` hands OWL2-RL a real budget — `ReasoningOptions { budget: reasoning_budget(reasoning), .. }` at `execute/reasoning_prep.rs:250` — and when that budget caps it emits a `tracing::warn!` whose own comment calls it "a correctness event, not a perf detail" (`:262-275`). The datalog call two branches below gets `MAX_DATALOG_ITERATIONS = 100` (`:322`, `:337`, `:346`) and nothing else: no fact cap, no wall-clock cap, no diagnostics. 100 iterations x a full ledger scan per rule is not a bound anyone would sign off on if it were written down — and `docs/query/datalog-rules.md:334` currently tells users the caps exist.

Worth noting that none of the three new tests exercise this shape — I checked each one's index route. `datalog_rule_insert_property_variable_substituted` binds both predicates explicitly; `datalog_rule_where_property_variable_binds_and_substitutes` binds `?other` in its first pattern before using `?prop`, so the wildcard pattern runs SPOT with a bound subject; `datalog_rule_literal_bound_property_variable_does_not_abort_other_rules` likewise binds `?x` first. All three take the safe route. The feature's expensive shape — a leading pattern with nothing bound — is the one with no coverage.

I want to be precise about what is and isn't new here: the `(None, None)` arm was already reachable from **SPARQL**-defined rules, since `ref_to_rule_term` (`fluree-db-api/src/sparql_lang.rs:199-207`) has always mapped `Ref::Var` into predicate position via `triple_to_rule_pattern:240`. This PR does not invent the hazard; it moves it from a seldom-used rule dialect to the front-end nearly every rule author uses, which is what makes the missing budget worth fixing now.

Please reuse the existing mechanism rather than standing up a second one — `fluree_db_reasoner::ReasoningBudget` already exists, is already constructed by `reasoning_budget()` at `reasoning_prep.rs:176`, and already has a suite at `fluree-db-api/tests/it_reasoning_budget.rs`. Thread it into `execute_datalog_rules_with_query_rules` and surface a `capped` diagnostic the way OWL2-RL does. A cheap complementary guard: hoist a pattern with at least one bound term to the front of `where_patterns`, or reject an all-unbound leading pattern at parse time.

fluree-db-query/src/datalog_rules.rs
```diff
@@ -470,14 +461,27 @@ fn parse_node_pattern(
     Ok(())
 }
 
+/// Parse a node-map key into a predicate term: a `?`-prefixed key is a
+/// variable; anything else expands and resolves as an IRI.
+fn parse_predicate_term(
```
aaj3f, line 468: *Praise.* (Covers `:466-477`.)

One named helper shared by all three pattern-construction arms rather than three inline `starts_with('?')` checks — and the PR then actually *deletes* the `?` branch in `expand_iri` and the two orphaned `parse_object_value` parameters instead of leaving them as dead defensive code.

I traced both remaining `expand_iri` callers (`:474`, `:520`); the removed branch is genuinely unreachable.

fluree-db-query/src/datalog_rules.rs
```diff
@@ -847,26 +856,37 @@ async fn match_pattern_with_bindings(
...
-                match binding {
-                    BindingValue::Sid(sid) => Ok((Some(sid.clone()), None)),
-                    _ => Err(QueryError::InvalidQuery(format!(
-                        "Variable {var} bound to non-SID value in subject/predicate position"
-                    ))),
-                }
-            } else {
-                Ok((None, Some(var.clone())))
-            }
+/// Resolution of a subject- or predicate-position term against existing bindings
```
aaj3f, line 933: *Praise.* (Covers `:859-880`.)

`TermResolution` with an explicit `Incompatible` variant and a `split()` adapter is the right shape — much better than threading a third state through `Option<Option<Sid>>`.

The doc comment explains *why* an empty join is the correct semantics, and the early return at `:765-777` sits before index selection, so an incompatible row costs no scan.

aaj3f (approved):
Approving but some probably should-fix notes below.

The diagnosis and the fix are both right: the JSON-LD rule parser was the only layer that did not understand predicate variables, `parse_predicate_term` is the correct single choke point, and converting a literal-bound subject/predicate binding from a hard error into an empty join fixes a genuinely nasty bug where one malformed rule erased every rule's derivations for the entire query. The two behavior changes have cleanly orthogonal test coverage — I mutated each one separately and each reddened exactly its own pair of tests, so these are real guards rather than decoration. 

The one finding I'd double-click on, though, is not the fix but its blast radius: it moves an unbudgeted full-ledger-scan-per-rule-per-iteration from a seldom-used rule dialect to the front-end everyone uses, on the one reasoning path with no fact or time ceiling, while OWL2-RL is handed a real `ReasoningBudget` fifteen lines earlier in the same function and the docs already speak to caps that do not exist. Threading the existing budget through `execute_datalog_rules_with_query_rules` (plus bplatz's docs update) would clear this though.

The parser fix itself is correct, minimal, and lands at the right choke point — I verified it by mutation, not by reading, and could not break the substitution logic on variable capture, multiple predicate variables, nested patterns, insert/where scope, or fixpoint termination. The one blocking finding is about the blast radius of the capability it unlocks, not about the substitution.

- **Performance (speed first, memory second):** ✖ **CRITICAL.** The PR's own code is parse-time and allocation-neutral — the `Incompatible` early return even saves an index scan — but it makes the `(None, None)` scan-all arm broadly reachable with no fact/time budget behind it.
- **Testing:** ⚠️ No SPARQL-side coverage for a shared-engine semantics change that the SPARQL front-end can already reach.
- **Conventions:** ⚠️ `docs/query/datalog-rules.md` was not updated — its budget and topological-sort claims are now conspicuously untrue.

---

**Non-blocking, but please file: a pre-existing bug this feature walks straight into.** *Real location: `fluree-db-query/src/datalog_rules.rs:353-372` and `:664-685`, neither in this diff.* Not blocking this PR — I confirmed it is pre-existing and independent of it — but it is the reason the obvious mitigation for the blocking finding isn't available today. **Rule FILTERs cannot compare against an IRI, and the exclusion form fails open silently.**

`parse_filter_term` (`:353-372`) never expands or resolves IRIs: a colon-bearing token falls through every branch to `RuleValue::String`, so `ex:knows` becomes the *string* `"ex:knows"`. Meanwhile `resolve_filter_term` (`:664-685`) renders a `BindingValue::Sid` as `FlakeValue::String(sid.name.to_string())` — the **local name only**, `"knows"`. `compare_values` (`:710-731`) then compares `"knows"` against `"ex:knows"`.

So `["filter", "(= ?p ex:knows)"]` is **always false** and the rule silently derives nothing, while `["filter", "(!= ?p ex:password)"]` is **always true** — an author writing a copy-properties rule who tries to exclude a sensitive predicate gets a no-op filter with no error and no warning. That is the failure direction that actually hurts. The only thing that does match is the bare local name, `(= ?p knows)`, which is namespace-blind across `ex:knows`, `foaf:knows`, and anything else named `knows`.

To be clear that this is not this PR's doing: `flake_value_to_binding` (`:985`) maps `FlakeValue::Ref` to `BindingValue::Sid` for **object** position too, so `["filter", "(!= ?o ex:bob)"]` over an ordinary object variable is broken the same way on `main` today. The fix is to resolve colon-bearing filter operands through `expand_iri` + `resolve_iri` into `RuleValue::Ref` and compare `Sid` bindings as `FlakeValue::Ref` — or, minimally, to reject a resolvable-IRI filter operand loudly rather than silently treating it as a string.

**SPARQL-side parity coverage (`fluree-db-api/tests/it_datalog_rules_sparql.rs`, untouched by this PR).** *Optional (parity, rubric §4).* The `TermResolution::Incompatible` change is shared engine code below both front-ends, and per the anchors above the SPARQL dialect could already put a variable in predicate position. So a SPARQL `CONSTRUCT { ?s ?p ?o } WHERE { … }` rule exercises exactly the path this PR changed and gains no coverage — that file still has its original three tests. This isn't one of §4's SPARQL-only exemptions. One SPARQL analogue of `datalog_rule_literal_bound_property_variable_does_not_abort_other_rules` would close it.

**`docs/query/datalog-rules.md:334`** *(no diff line).* *Optional, seconding bplatz.* Beyond the Where/Insert sections he named, line 334 claims termination "is guaranteed by the budget controls (max iterations, max facts, max time, max memory)". Only max-iterations exists for datalog; the other three are the OWL2-RL budget. Given the blocking finding, this line is actively misleading and should either be corrected or made true.

**`docs/query/datalog-rules.md:339`** *(no diff line).* *Optional.* "Rules are topologically sorted by their predicate dependencies" does not describe `compute_execution_order` (`fluree-db-reasoner/src/datalog.rs:279-287`), which sorts by `depends_on.len()` under its own comment `// For now, just use insertion order (can be improved later)`.

**`fluree-db-reasoner/src/datalog.rs:147-152`** *(no diff line).* *Optional, confirming bplatz.* I independently checked his "cosmetic" claim and it holds, with one correction to the reasoning: `generates` has no readers anywhere in the workspace, and `depends_on` has **two** consumers, not one — the sort key at `:287` and `dependent_predicates()` at `:291`. The second is itself dead (no call sites; the only `iter_in_order` use is `datalog_rules.rs:1217`), so the conclusion is unchanged and a var-predicate rule sorting first as zero-dependency is harmless under a naive fixpoint. Worth a comment on `depends_on` saying so, since the field name implies load-bearing semantics it does not have.

**`fluree-db-query/src/datalog_rules.rs:1238`, `:1264`** *(not in the diff).* *Optional, pre-existing but amplified.* The fixpoint's dedup key allocates a `String` per derived flake per iteration via `format!("{:?}", flake.o)`, and `:1264` rebuilds the entire derived overlay each iteration with `builder.extend(all_derived.values().cloned())` — O(n²) clones across the fixpoint. Not introduced here, but the copy-properties shape this PR enables is exactly what makes derived-fact counts large enough for it to matter. Follow-up issue, not a change in this PR.

PR #1542, opened 2026-07-24
Add regression coverage for sub-SELECT aggregate correlation joins

### LINKED ISSUE #1388: Sub-SELECT correlation var produced by an inner aggregate is not join-enforced
## Problem

When a sub-SELECT exposes a correlation variable that it produces via an **aggregate** (or BIND/OPTIONAL) rather than a top-level triple, the natural join with the enclosing pattern is not enforced on that variable. Rows that should be filtered out by the join survive with the parent's binding.

### Repro

```sparql
SELECT ?s
WHERE {
  ?s :p ?c .
  {
    SELECT ?s (COUNT(?x) AS ?c)
    WHERE { ?s :r ?x }
    GROUP BY ?s
  }
}
```

Data: `ex:a :p 2 ; :r ex:x1, ex:x2` (COUNT=2, matches `:p`) and `ex:b :p 5 ; :r ex:y1` (COUNT=1, does NOT match `:p`=5).

Expected: only `ex:a` (the `:p ?c == COUNT` equality must hold).
Actual: both `ex:a` and `ex:b` returned — the `?c` equality is never checked.

### Root cause

In `fluree-db-query/src/subquery.rs`, the materialize-once hash join keys only on `join_keys` (correlation vars the subquery produces via a top-level required triple). A correlation var produced by an aggregate is excluded from `join_keys`, and the non-clobbering merge (`subquery.rs:~420-430`) keeps the parent's value without an equality check. So the join degenerates to the other keys (here `?s`) only.

### Scope / status

- **Pre-existing**, not a regression: verified the identical wrong output on `origin/main` as well as on the PR branch (`fix/groupby-datatype-expr-1362`).
- Surfaced during review of PR #1374 (https://github.com/fluree/db/pull/1374#discussion_r... aaj3f). Deliberately left out of that PR's scope.

### Suggested fix direction

Either include aggregate/BIND-produced correlation vars in the post-join equality check (filter merged rows where the parent binding ≠ the subquery binding for such vars), or extend the hash key to cover them. Needs care for the unbound-compatible case.

zonotope:
Issue #1388 reported that a sub-SELECT exposing a correlation variable through an aggregate (e.g. `(COUNT(?x) AS ?c)`) failed to enforce the natural join with the enclosing pattern: the hash join keyed only on self-produced correlation vars, so rows whose aggregate value conflicted with the parent's binding leaked through unchecked.

The underlying defect no longer reproduces on `main`. It was fixed by c0f63e4c3 ("fix(sparql): reconcile OPTIONAL-produced subquery correlation vars"), which partitions correlation variables into `join_keys` (self-produced by a top-level required triple or property path, used as the hash-join key) and `reconcile_vars` (everything else, checked for compatibility at merge time per SPARQL §18.4). An aggregate alias is not self-produced, so it lands in `reconcile_vars` and the merge now drops rows whose aggregate value conflicts with the parent's binding — exactly the enforcement #1388 asked for. That commit's message and tests only covered the OPTIONAL/UNION case (W3C `var-scope-join-1`), leaving the aggregate shape untested, which is presumably why the issue remained open.

This PR adds the missing regression coverage so the aggregate-correlation behavior is pinned. No production code changes.

## Changes

- New test file `fluree-db-api/tests/it_query_subselect_correlation.rs` with six end-to-end tests over the issue's dataset (`ex:a` with `ex:p 2` and COUNT 2, `ex:b` with `ex:p 5` and COUNT 1, `ex:c` with no `ex:p` and COUNT 3):
  - The issue's verbatim repro: aggregate-produced `?c` is join-checked, returning only `ex:a`.
  - The pattern-order-swapped variant (sub-SELECT written before the parent triple) returns the identical result.
  - A scalar (ungrouped) aggregate correlation var: no parent matches the total count, so the result is empty.
  - The unbound-compatible case flagged in the issue: a parent `?c` left unbound by OPTIONAL is compatible and adopts the subquery's count, while a bound conflict still drops the row.
  - A BIND-produced correlation var, which is likewise not self-produced and must reconcile; also documents that the non-aggregate sub-SELECT emits one row per body match, so the duplicate `ex:a` rows are correct multiset join semantics.
  - JSON-LD `["query", ...]` surface parity for the issue shape (`"(as (count ?x) ?c)"` with `groupBy`).
- Wired the file into `fluree-db-api/tests/grp_query_sparql.rs`.

Closes #1388

### DIFF
--- a/fluree-db-api/tests/grp_query_sparql.rs
+++ b/fluree-db-api/tests/grp_query_sparql.rs
@@ -19,6 +19,8 @@ mod it_query_sparql_parse_errors;
 mod it_query_sparql_setop_subselect;
 #[path = "it_query_subquery.rs"]
 mod it_query_subquery;
+#[path = "it_query_subselect_correlation.rs"]
+mod it_query_subselect_correlation;
 #[path = "it_query_unwind.rs"]
 mod it_query_unwind;
 #[path = "it_query_values.rs"]
--- a/fluree-db-api/tests/it_query_subselect_correlation.rs
+++ b/fluree-db-api/tests/it_query_subselect_correlation.rs
@@ -181,8 +181,20 @@ async fn sparql_subselect_bind_projection_multiset() {
 }
 
 /// JSON-LD `["query", ...]` surface parity for the issue-#1388 shape: the
-/// correlated (per-row seeded) path must enforce the same merge-time check on
-/// the aggregate-produced `?c`.
+/// aggregate-produced `?c` must be reconciled at merge time, the same as the
+/// SPARQL sub-SELECT above.
+///
+/// NOTE: on this fixture this runs `join_mode=true`, NOT the per-row seeded
+/// route. The sub-SELECT builds `uncorrelated=false, est_rows=None`, so
+/// `must_materialize_once` is false and the cardinality guard
+/// (`child.estimated_rows().is_none_or(|n| n >= SUBQUERY_MATERIALIZE_MIN_PARENT_ROWS)`,
+/// subquery.rs) resolves the `None` estimate to join-mode — like every other
+/// test here. The reconcile check is load-bearing in join-mode too, so this
+/// still exercises it (disabling the check turns this test red). The per-row
+/// seeded route — the one c0f63e4c3 singles out as needing the seed fix — is
+/// covered by the Family-B tests in `it_query_filter_scope.rs`; a parent whose
+/// child reports a concrete estimate below the threshold would flip this test
+/// onto it.
 #[tokio::test]
 async fn jsonld_subquery_aggregate_correlation_var_is_joined() {
     let fluree = FlureeBuilder::memory().build_memory();
--- a/fluree-db-api/tests/it_query_subselect_correlation.rs
+++ b/fluree-db-api/tests/it_query_subselect_correlation.rs
@@ -145,20 +145,22 @@ async fn sparql_subselect_aggregate_correlation_var_unbound_parent_is_compatible
     );
 }
 
-/// End-to-end correctness for a BIND-produced correlation variable: `?c` is
-/// bound by `BIND(2 AS ?c)` in the sub-SELECT and the result must agree with
-/// the parent's `?c`. The sub-SELECT emits `(ex:a, 2)` once per `ex:r` link,
-/// so `ex:a` appears twice (multiset join semantics); `ex:b` (parent `?c` = 5)
-/// does not agree and is dropped.
+/// End-to-end (multiset-semantics) assertion for a BIND'd sub-SELECT
+/// projection. `?c` is bound by `BIND(2 AS ?c)` and the result must agree with
+/// the parent's `?c`: the sub-SELECT emits `(ex:a, 2)` once per `ex:r` link, so
+/// `ex:a` appears twice (`normalize_rows` sorts but does not dedupe, so the
+/// duplicate has teeth); `ex:b` (parent `?c` = 5) does not agree and is dropped.
 ///
-/// NOTE: unlike the aggregate cases above, this shape eliminates `ex:b`
-/// *before* the merge-time reconcile check — disabling that check (subquery.rs
-/// `process_parent_batch`) leaves this test green — so it is an end-to-end
-/// correctness assertion, not a regression guard for the reconcile mechanism.
-/// The aggregate cases here, and the OPTIONAL Family-B tests in
-/// `it_query_filter_scope.rs`, pin the reconcile merge itself.
+/// NOTE: this does NOT exercise the merge-time reconcile check. On this
+/// fixture the sub-SELECT plans UNCORRELATED — the `SubqueryOperator` builds
+/// with `correlation_vars=[]` / `reconcile_vars=[]` (`?c` never becomes a
+/// correlation variable), so `reconcile_vars.is_empty()` short-circuits the
+/// check at `subquery.rs:472` and `ex:b` is dropped by the outer `?s ex:p ?c`
+/// join instead. Disabling the reconcile check leaves this test green. The
+/// reconcile mechanism itself is pinned by the aggregate cases above and the
+/// OPTIONAL Family-B tests in `it_query_filter_scope.rs`.
 #[tokio::test]
-async fn sparql_subselect_bind_correlation_var_end_to_end() {
+async fn sparql_subselect_bind_projection_multiset() {
     let fluree = FlureeBuilder::memory().build_memory();
     let ledger = seed_counts(&fluree, "subsel-corr:bind").await;
 
@@ -174,7 +176,7 @@ async fn sparql_subselect_bind_correlation_var_end_to_end() {
     assert_eq!(
         normalize_rows(&rows),
         normalize_rows(&json!([["ex:a"], ["ex:a"]])),
-        "BIND-produced ?c must be join-checked against the parent, got {rows}"
+        "BIND'd ?c must agree with the parent, keeping ex:a twice (multiset), got {rows}"
     );
 }
 
--- a/fluree-db-api/tests/it_query_subselect_correlation.rs
+++ b/fluree-db-api/tests/it_query_subselect_correlation.rs
@@ -1,0 +1,199 @@
+//! Regression tests: correlation variables a sub-SELECT produces via an
+//! aggregate (or BIND) must still be join-checked against the parent's binding
+//! (issue #1388).
+//!
+//! When a sub-SELECT exposes a correlation variable through `(COUNT(?x) AS ?c)`
+//! rather than a top-level triple, the variable is not a hash-join key (the
+//! subquery does not self-produce it in its WHERE body), so the natural join
+//! must be enforced at merge time — a row whose aggregate value conflicts with
+//! the parent's binding is dropped, while an unbound side stays compatible and
+//! adopts the bound side's value, per SPARQL §18.4 compatible mappings.
+//!
+//! All inserts and queries are explicit with `@context` / `PREFIX`.
+
+use crate::support;
+use crate::support::{genesis_ledger, normalize_rows, MemoryFluree, MemoryLedger};
+use fluree_db_api::FlureeBuilder;
+use serde_json::json;
+
+/// `ex:a` has `ex:p 2` and two `ex:r` links (COUNT = 2, matches).
+/// `ex:b` has `ex:p 5` and one `ex:r` link (COUNT = 1, does not match).
+/// `ex:c` has no `ex:p` and three `ex:r` links (parent side unbound).
+async fn seed_counts(fluree: &MemoryFluree, ledger_id: &str) -> MemoryLedger {
+    let ledger0 = genesis_ledger(fluree, ledger_id);
+    let insert = json!({
+        "@context": {"ex": "http://example.org/"},
+        "@graph": [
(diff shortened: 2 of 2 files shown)


fluree-db-api/tests/it_query_subselect_correlation.rs
```diff
@@ -0,0 +1,199 @@
...
+          ?s ex:p ?c .
+          { SELECT ?s ?c WHERE { ?s ex:r ?x . BIND(2 AS ?c) } }
+        }
+    ";
+
+    let rows = sparql_rows(&fluree, &ledger, q).await;
+    assert_eq!(
+        normalize_rows(&rows),
+        normalize_rows(&json!([["ex:a"], ["ex:a"]])),
+        "BIND-produced ?c must be join-checked against the parent, got {rows}"
```
bplatz: This one doesn't pin what the doc comment says it pins. With the reconcile merge check disabled it still returns exactly `[ex:a, ex:a]` — the only test in the file that stays green — so `ex:b` is being eliminated by something other than merge-time reconciliation, most likely the per-row seed path.

Still a valid end-to-end assertion, but as written it doesn't cover the mechanism, and it would not catch a regression in the reconcile check. Worth either retitling it to what it actually exercises, or reshaping it so the BIND value can only be reconciled at merge time.

fluree-db-api/tests/it_query_subselect_correlation.rs
```diff
@@ -0,0 +1,199 @@
...
+          ?s ex:p ?c .
+          { SELECT (COUNT(?x) AS ?c) WHERE { ?any ex:r ?x } }
+        }
+    ";
+
+    let rows = sparql_rows(&fluree, &ledger, q).await;
+    assert_eq!(
+        normalize_rows(&rows),
+        normalize_rows(&json!([])),
+        "no parent ex:p equals the total count 6, got {rows}"
```
bplatz: Verified this does discriminate today (returns `[ex:a, ex:b]` with the check disabled), so no objection. But an empty-result assertion passes for any reason the query returns nothing. Adding an entity with `ex:p 6` to the fixture would make this assert a row instead of the absence of rows, and would catch a future regression where the scalar-aggregate subquery path itself stops producing.

bplatz (commented):
Checked that these actually discriminate, since the whole value of a test-only PR is that they'd fail without c0f63e4c3. Reverse-applying that commit conflicts with later changes to `subquery.rs`, so I instead disabled just the reconcile merge check (`subquery.rs:472`, forced to skip) in a scratch worktree and re-ran: 5 of the 6 fail, with the aggregate cases returning `[ex:a, ex:b]` — the issue's exact wrong output. Good coverage, and the scalar-aggregate case is not the vacuous empty assertion it first looks like (it returns `[ex:a, ex:b]` unchecked).

The BIND test is the exception — see inline. Full `grp_query_sparql` passes (294).

One note on closing #1388: the issue names three producers — aggregate, BIND, OPTIONAL. Aggregate is well covered here; the sub-SELECT-side OPTIONAL producer isn't in this file. The PR body says c0f63e4c3 covered it in `it_query_filter_scope.rs`, which is fine, but a pointer in the module doc would make the closure auditable from one place.

fluree-db-api/tests/it_query_subselect_correlation.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    let rows = sparql_rows(&fluree, &ledger, q).await;
+    assert_eq!(
+        normalize_rows(&rows),
+        normalize_rows(&json!([["ex:a", 2], ["ex:c", 3]])),
+        "unbound parent ?c must adopt the subquery count; bound conflicts drop, got {rows}"
+    );
+}
+
+/// A BIND-produced correlation variable inside the sub-SELECT is likewise not
```
aaj3f: *Should fix.* (Covers `:141-144`.)

Building on @bplatz's observation that this test stays green with the reconcile check disabled: the reason is not the per-row seed path, it's that **`?c` never becomes a correlation variable at all**. I instrumented the `SubqueryOperator` construction site and this test is the only one that builds with `uncorrelated=true, correlation_vars=[], reconcile_vars=[]` — so `reconcile_vars.is_empty()` short-circuits the check at `subquery.rs:472` before it looks at anything, and `ex:b` is eliminated elsewhere in the plan. The doc comment's claim that a BIND-produced var "is likewise not self-produced by a triple and must be reconciled at merge time" is therefore describing a code path this query does not enter.

That matters more than a wording nit, because this is the one test in the file carrying the #1388 issue's "or BIND" producer, and as written it would not catch a regression in the reconcile logic.

Two options: retitle it to what it actually is (an end-to-end multiset-semantics assertion for a BIND'd sub-SELECT projection — still worth having), or reshape it so `?c` is genuinely a correlation var the subquery does not self-produce. If you take the second, verify the fix landed by re-running with `if false && !self.reconcile_vars.is_empty()` at `subquery.rs:472` and confirming this test goes red along with the other five.

fluree-db-api/tests/it_query_subselect_correlation.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    let rows = sparql_rows(&fluree, &ledger, q).await;
+    assert_eq!(
+        normalize_rows(&rows),
+        normalize_rows(&json!([["ex:a"], ["ex:a"]])),
+        "BIND-produced ?c must be join-checked against the parent, got {rows}"
+    );
+}
+
+/// JSON-LD `["query", ...]` surface parity for the issue-#1388 shape: the
```
aaj3f, line 204: *Should fix (doc), optional (coverage).* (Covers `:166-168`.)

The doc comment says this pins "the correlated (per-row seeded) path", but on this fixture it does not. Instrumentation shows it builds with `uncorrelated=false, est_rows=None`, so `must_materialize_once` is false and the cardinality guard at `subquery.rs:236-238` — `child.estimated_rows().is_none_or(|n| n >= SUBQUERY_MATERIALIZE_MIN_PARENT_ROWS)` — resolves `None` to **join-mode**. All six tests in this file run `join_mode=true`; the per-row seeded route has zero coverage here.

This is not a live bug: I forced the per-row route (`join_mode = eligible && correlation_vars.is_empty()`) and all six still passed. But it does mean the JSON-LD test is currently a route-duplicate of the SPARQL test above it rather than the second-route parity case its comment advertises — and the per-row route is the one c0f63e4c3's message singles out as needing the seed fix, i.e. the route where the merge check alone is insufficient.

Minimum fix is one line of comment accuracy. If you want the route actually covered, a parent whose child reports a concrete estimate below 8 flips the branch.

fluree-db-api/tests/it_query_subselect_correlation.rs
```diff
@@ -0,0 +1,199 @@
+//! Regression tests: correlation variables a sub-SELECT produces via an
```
aaj3f, line 1: *Optional.* (Covers the module doc at `:1-12`.)

The commit c0f63e4c3 this file pins has **two** mechanisms, and only one is pinned here. Reverting the per-row seed exclusion at `subquery.rs:593-596` (`self.join_keys` → `self.correlation_vars`) while leaving the merge check intact leaves **all six tests green**; the same mutation turns `it_query_filter_scope.rs::subselect_optional_bound_correlation_var_reconciles_to_empty` red (`got [["ex:paul","ex:george",null]]`, expected `[]`). So the seed half is covered, but in a different test binary (`grp_query`, not `grp_query_sparql`).

This lines up with @bplatz's auditability point about closing #1388 from one place. Since the module doc already explains the join_keys/reconcile_vars split, naming both mechanisms and pointing at `it_query_filter_scope.rs` for the seed-exclusion + OPTIONAL-producer half would make the closure verifiable without re-deriving it.

Worth doing before `Closes #1388` lands, since the issue names three producers and this file demonstrably exercises the reconcile mechanism for one of them.

fluree-db-api/tests/it_query_subselect_correlation.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    let rows = sparql_rows(&fluree, &ledger, q).await;
+    assert_eq!(
+        normalize_rows(&rows),
+        normalize_rows(&json!([])),
+        "no parent ex:p equals the total count 6, got {rows}"
+    );
+}
+
+/// Unbound-compatible merge: `ex:c` has no `ex:p`, so its parent `?c` is
```
aaj3f, line 144: *Praise.* (Covers `:116-139`.)

This is the strongest test in the file and the one to preserve through any reshuffling. It is the only one that asserts a bound value rather than a subject list, and it covers all three SPARQL §18.4 outcomes in a single assertion — equality keeps (`ex:a`), conflict drops (`ex:b`), unbound-adopts (`ex:c`).

Under the disabled-check mutation it fails with `[["ex:a",2],["ex:b",5],["ex:c",3]]`, which is a much more diagnostic failure message than a bare subject list.

fluree-db-api/tests/grp_query_sparql.rs
```diff
@@ -19,6 +19,8 @@ mod it_query_sparql_parse_errors;
 mod it_query_sparql_setop_subselect;
 #[path = "it_query_subquery.rs"]
 mod it_query_subquery;
+#[path = "it_query_subselect_correlation.rs"]
```
aaj3f, line 22: *Praise.* (Covers `:22-23`.)

Wiring is correct and alphabetically placed, which is the trap worth checking first in this crate: `fluree-db-api` sets `autotests = false`, so an `it_*.rs` that no `grp_*.rs` declares compiles nowhere and runs never (cf. the ~166 dead tests in `it_query_cypher.rs`).

Verified reachable: `cargo nextest run -p fluree-db-api --test grp_query_sparql --all-features -E 'test(subselect_correlation)'` → 6 run, 6 passed, 288 skipped.

aaj3f (approved):
Tests-only, correctly wired, and the coverage is real rather than decorative — I re-ran @bplatz's mutation independently (merge reconcile check forced off at `subquery.rs:472`) and confirm 5 of 6 go red with the aggregate cases returning issue #1388's exact wrong output `[ex:a, ex:b]`, and I hand-derived all six expected values from the fixture at `it_query_subselect_correlation.rs:19-36` — every one is arithmetically correct, including the scalar total of 6 and the two-row `ex:a` multiset (`normalize_rows` sorts but does not dedupe, so that assertion has teeth). 

Any review nits I have are all about the file describing itself accurately: the BIND test at :141 builds with `correlation_vars=[]` so it never enters the reconcile path its comment claims. The JSON-LD test at :166 runs `join_mode=true` rather than the "per-row seeded path" it advertises. And the seed-exclusion half of c0f63e4c3 is pinned in a different binary entirely. 

None of that blocks merge — no production code changes and nothing here is a correctness bug — but a regression file whose comments misstate the mechanism will mislead whoever next refactors `subquery.rs`, which is exactly the reader this file exists for, so I'd land the comment corrections first.

Recommend merge after the two doc-comment corrections at :141-144 and :166-168 — no code change required for either.

Context for the inline comments: I am extending @bplatz's pass rather than repeating it. I independently reproduced his mutation (merge reconcile check disabled at `subquery.rs:472` → 5 of 6 red, aggregate cases returning `[ex:a, ex:b]`) and confirm every number in it, including that the scalar-aggregate case at :96 is not a vacuous empty assertion. The three findings below are the ones his pass did not reach — the *mechanism* behind the green BIND test, a second uncovered half of the fix, and the execution route all six tests actually take.

PR #1543, opened 2026-07-24
feat(BM25): embedded BM25 full-text search over the standard query path + a bm25 CLI

christophediprima:
## Summary

Two related, independent changes that make Fluree's BM25 full-text search usable without a second process or the Rust API:

1. **Embedded `f:searchText` now executes over the normal query route** (`POST /v1/fluree/query`), not only via the dedicated `query_connection_with_bm25` API method or the standalone `fluree-search-httpd` service.
2. **A `fluree bm25` CLI** to create, list, sync, and drop BM25 indexes (the only management surface that ships today is the Rust API — there is no HTTP endpoint for index management).

Both are self-contained and could land independently; they are grouped here because they are the two halves of "use BM25 from a running Fluree server."

---

## 1. Embedded BM25 over `/v1/fluree/query`

A query with an `f:searchText` / `f:graphSource` where-clause parses correctly (the JSON-LD parser lowers it to a `Pattern::IndexSearch`) but, on the standard query route, failed at execution with `BM25 IndexSearch requires ExecutionContext.bm25_search_provider or bm25_provider (not configured)`. The query is valid; the execution context that route builds simply never wired a BM25 provider. The only code that wired one was `query_connection_with_bm25` / `query_dataset_with_bm25`, whose sole callers were tests — the server never called them — so embedded search was unreachable over HTTP. (The docs describe embedded search as the default deployment mode and flag the missing in-server wiring as a "near-term gap".)

**Fix.** Wire `FlureeIndexProvider` as the `bm25_provider` on the `ExecutionContext` at the sites the HTTP route funnels through: `execute_dataset_into_with_r2rml` and `execute_dataset_tracked_with_r2rml` (the dataset path used by `POST /v1/fluree/query`), and the shared `view_context_config!` macro (the single-graph view path), for parity. Each creates a `FlureeIndexProvider::new(self)` — a zero-cost `&Fluree` wrapper — declared before the context so it outlives the borrow, mirroring exactly how the R2RML providers are already threaded through those same sites. The provider is only consulted by the `IndexSearch` operator, so non-search queries are unaffected; this is the same provider `query_dataset_with_bm25` already used, just made available on the standard path.

**Verified end-to-end.** `POST /v1/fluree/query` with a BM25 search block now returns ranked `?doc`/`?score` bindings instead of a 400 — confirmed on a real index (`"nature"` → two profiles, scored 3.78 and 2.68). A new integration test, `bm25_embedded_search_via_plain_query_connection`, runs `f:searchText` through the plain `query_connection` (the path the HTTP route uses) and asserts scored results — it would have returned the "not configured" error before this change.

**Not covered (deliberately):** SPARQL. The `f:` search block is recognized only by the JSON-LD parser (`parse/node_map.rs`); the SPARQL parser never constructs a `Pattern::IndexSearch`, so a SPARQL query with those predicates matches them as ordinary triples and silently returns nothing. Adding SPARQL search is a separate parser/lowering change in `fluree-db-sparql`. Vector (`f:queryVector`) is also out of scope here: its provider impl is `#[cfg(feature = "vector")]` and not in the default server build, so wiring it would be dead code in the shipped binary — a one-line follow-up once that feature is enabled.

---

## 2. `fluree bm25` CLI

BM25 index management is Rust-API-only (`Bm25CreateConfig` + `create_full_text_index`, `sync_bm25_index`, `drop_full_text_index`) with no HTTP route. This adds a thin CLI wrapper that runs in-process against local storage, so it works under `docker exec` against a running server's data directory (native storage uses per-file locks, not a whole-store exclusive lock).

- **`fluree bm25 create --name <n> --ledger <l> -f <index-query.json>`** — build a BM25 index over a source ledger from an indexing query.
- **`fluree bm25 list [--stale]`** — tabular view (NAME / BRANCH / SOURCE LEDGER / INDEX_T / LEDGER_T / STALE), the same look-and-feel as `fluree list`. `--stale` prints `name:branch` per line so a maintenance loop can be scripted: `for i in $(fluree bm25 list --stale); do fluree bm25 sync --index "$i"; done`.
- **`fluree bm25 sync --index <gs>`** — incremental sync up to the source ledger's head; prints the watermark `old -> new` and upserted/removed counts (falls back to a full resync when no affected subjects are detected).
- **`fluree bm25 drop --index <gs>`** — retract the index in the nameservice and delete its snapshot blobs.

---

## Tests

Full BM25 integration suite green (`fluree-db-api --test grp_graphsource`, incl. the new `bm25_embedded_search_via_plain_query_connection`); `cargo fmt --check` and `clippy -D warnings` clean on both `fluree-db-api` and `fluree-db-cli`. The branch is based on current `main` and compiles + tests against it (not just the fork's older base).

### DIFF
--- a/docs/indexing-and-search/bm25.md
+++ b/docs/indexing-and-search/bm25.md
@@ -110,6 +110,8 @@ The analyzer is not configurable — it always uses the English pipeline for con
 
 BM25 search is expressed with the `f:` namespace predicates in a query's `where` clause. It is a **JSON-LD (FQL) feature** and executes over the standard query endpoint — `POST /v1/fluree/query` — so search hits can be joined with ledger data in a single query. It is **not** available via SPARQL: the SPARQL parser does not recognize the `f:` search block, so those predicates would be matched as ordinary triples and return nothing.
 
+> **Policy-enforced search must be scoped to a single ledger.** A search whose `from` names more than one ledger conservatively returns **no hits** when any of those ledgers enforces a policy — not filtered hits. Per-flake readability cannot be resolved across a multi-graph scope, so the check hides everything rather than risk leaking. This is fail-closed and therefore safe, but silent: the same query returns rows with policies off and zero rows with them on, and no error explains the difference. Scope policy-enforced searches to one ledger.
+
 ### JSON-LD Query Syntax
 
 BM25 search is integrated into Fluree's query system via the `f:` namespace predicates:
@@ -326,7 +328,7 @@ The `fluree bm25` command group manages indexes without writing Rust. Each comma
 | `fluree bm25 create --name <n> --ledger <l> -f <index-query.json>` | Build a BM25 index over a source ledger from an indexing query |
 | `fluree bm25 list [--stale]` | List indexes with their source ledger, `INDEX_T`/`LEDGER_T`, and staleness. `--stale` prints `name:branch` per line for scripting |
 | `fluree bm25 sync --index <gs>` | Incrementally sync an index up to its source ledger's head (prints the watermark `old -> new` and upserted/removed counts) |
-| `fluree bm25 drop --index <gs>` | Retract the index and delete its snapshot blobs |
+| `fluree bm25 drop --index <gs> --force` | Retract the index and delete its snapshot blobs (`--force` required — drop is destructive) |
 
 A maintenance loop syncs every stale index:
 
--- a/docs/indexing-and-search/bm25.md
+++ b/docs/indexing-and-search/bm25.md
@@ -17,7 +17,7 @@ BM25 is a probabilistic ranking function that scores documents based on query te
 
 ## Creating a BM25 Index
 
-BM25 indexes are created via the Rust API using `Bm25CreateConfig`. There are no HTTP endpoints for index management yet — indexes are managed programmatically.
+BM25 indexes are managed via the Rust API (`Bm25CreateConfig`) or the `fluree bm25` CLI — there is no HTTP endpoint for index management. The CLI (`fluree bm25 create/list/sync/drop`) runs in-process against local storage, so it works under `docker exec` against a running server's data directory; see [CLI Management](#cli-management) below.
 
 ### Basic Index
 
@@ -107,6 +107,8 @@ Fluree uses a default English analyzer that applies:
 The analyzer is not configurable — it always uses the English pipeline for consistency.
 
 ## Querying BM25 Indexes
+
+BM25 search is expressed with the `f:` namespace predicates in a query's `where` clause. It is a **JSON-LD (FQL) feature** and executes over the standard query endpoint — `POST /v1/fluree/query` — so search hits can be joined with ledger data in a single query. It is **not** available via SPARQL: the SPARQL parser does not recognize the `f:` search block, so those predicates would be matched as ordinary triples and return nothing.
 
 ### JSON-LD Query Syntax
 
@@ -315,11 +317,28 @@ The storage format is selected automatically based on the storage backend:
 - **File storage**: V3 single-blob format (optimized for local filesystem)
 - **Memory / S3 / object store**: V4 chunked format (enables selective loading and caching)
 
+## CLI Management
+
+The `fluree bm25` command group manages indexes without writing Rust. Each command runs in-process against local storage (no server round-trip), so it works under `docker exec` against a running server's data directory (native storage uses per-file locks, so this is safe alongside a running server).
+
+| Command | Description |
+|---------|-------------|
+| `fluree bm25 create --name <n> --ledger <l> -f <index-query.json>` | Build a BM25 index over a source ledger from an indexing query |
+| `fluree bm25 list [--stale]` | List indexes with their source ledger, `INDEX_T`/`LEDGER_T`, and staleness. `--stale` prints `name:branch` per line for scripting |
+| `fluree bm25 sync --index <gs>` | Incrementally sync an index up to its source ledger's head (prints the watermark `old -> new` and upserted/removed counts) |
+| `fluree bm25 drop --index <gs>` | Retract the index and delete its snapshot blobs |
+
+A maintenance loop syncs every stale index:
+
+```bash
+for i in $(fluree bm25 list --stale); do fluree bm25 sync --index "$i"; done
+```
+
 ## Deployment Modes
 
 ### Embedded Mode (Default)
 
-In embedded mode, the BM25 index is loaded and searched within the same process as Fluree. This is the default behavior.
+In embedded mode, the BM25 index is loaded and searched within the same process as Fluree. This is the default behavior, and it is what serves an `f:searchText` query over the main server's `POST /v1/fluree/query` route — the query execution context wires a local index provider, so no separate search service is required.
 
 ### Remote Mode
 
--- a/fluree-db-api/src/view/dataset_query.rs
+++ b/fluree-db-api/src/view/dataset_query.rs
@@ -678,6 +678,13 @@ impl Fluree {
                 (None, None, None, None, None)
             };
 
+        // Wire the BM25 index provider so embedded `f:searchText` graph-source
+        // queries execute in-process (parity with `query_dataset_with_bm25`).
+        // It is a zero-cost `&Fluree` wrapper and is only consulted by the
+        // `IndexSearch` operator, so non-search queries are unaffected. Declared
+        // before `config` so it outlives the borrow held by the context.
+        let index_provider = crate::FlureeIndexProvider::new(self);
+
         let config = ContextConfig {
             tracker: if tracker.is_enabled() {
                 Some(tracker)
@@ -688,6 +695,7 @@ impl Fluree {
             dataset: Some(&runtime_dataset),
             policy_enforcer: primary.policy_enforcer().cloned(),
             r2rml: Some((r2rml.provider, r2rml.table_provider)),
+            bm25_provider: Some(&index_provider),
             binary_g_id: primary.graph_id,
             binary_store,
             dict_novelty,
@@ -818,12 +826,19 @@ impl Fluree {
                 (None, None, None, None, None)
             };
 
+        // Wire the BM25 index provider so embedded `f:searchText` graph-source
+        // queries execute in-process (parity with `query_dataset_with_bm25`).
+        // Zero-cost `&Fluree` wrapper, only consulted by the `IndexSearch`
+        // operator; declared before `config` so it outlives the context borrow.
+        let index_provider = crate::FlureeIndexProvider::new(self);
+
         let config = ContextConfig {
             tracker: Some(tracker),
             cancellation: options.cancellation.clone(),
             dataset: Some(&runtime_dataset),
             policy_enforcer: primary.policy_enforcer().cloned(),
             r2rml: Some((r2rml.provider, r2rml.table_provider)),
+            bm25_provider: Some(&index_provider),
             binary_g_id: primary.graph_id,
             binary_store,
             dict_novelty,
--- a/fluree-db-api/src/view/dataset_query.rs
+++ b/fluree-db-api/src/view/dataset_query.rs
@@ -679,10 +679,17 @@ impl Fluree {
             };
 
         // Wire the BM25 index provider so embedded `f:searchText` graph-source
-        // queries execute in-process (parity with `query_dataset_with_bm25`).
+        // queries execute in-process over the standard query route.
         // It is a zero-cost `&Fluree` wrapper and is only consulted by the
         // `IndexSearch` operator, so non-search queries are unaffected. Declared
         // before `config` so it outlives the borrow held by the context.
+        //
+        // `bm25_provider` (the "legacy" index-provider slot), NOT the
+        // "preferred" `bm25_search_provider` that `ContextConfig` steers you
+        // toward: view-policy enforcement on a hit only works in index-provider
+        // mode. Search-provider mode has no local flakes and fails closed
+        // (`operator.rs`), so it would return zero rows for every
+        // policy-enforced query. Do not "upgrade" this.
         let index_provider = crate::FlureeIndexProvider::new(self);
 
         let config = ContextConfig {
@@ -827,9 +834,16 @@ impl Fluree {
             };
 
         // Wire the BM25 index provider so embedded `f:searchText` graph-source
-        // queries execute in-process (parity with `query_dataset_with_bm25`).
+        // queries execute in-process over the standard query route.
         // Zero-cost `&Fluree` wrapper, only consulted by the `IndexSearch`
         // operator; declared before `config` so it outlives the context borrow.
+        //
+        // `bm25_provider` (the "legacy" index-provider slot), NOT the
+        // "preferred" `bm25_search_provider` that `ContextConfig` steers you
+        // toward: view-policy enforcement on a hit only works in index-provider
+        // mode. Search-provider mode has no local flakes and fails closed
+        // (`operator.rs`), so it would return zero rows for every
+        // policy-enforced query. Do not "upgrade" this.
         let index_provider = crate::FlureeIndexProvider::new(self);
 
         let config = ContextConfig {
--- a/fluree-db-api/src/view/mod.rs
+++ b/fluree-db-api/src/view/mod.rs
@@ -93,11 +93,17 @@ macro_rules! view_context_config {
             None
         };
         // Wire the BM25 index provider so embedded `f:searchText` graph-source
-        // queries execute in-process on the single-graph view path too (parity
-        // with `query_dataset_with_bm25` and the dataset path). Zero-cost
-        // `&Fluree` wrapper, only consulted by the `IndexSearch` operator, so
-        // non-search queries are unaffected. Declared before `$cfg` so it
-        // outlives the borrow the context holds.
+        // queries execute in-process on the single-graph view path too.
+        // Zero-cost `&Fluree` wrapper, only consulted by the `IndexSearch`
+        // operator, so non-search queries are unaffected. Declared before `$cfg`
+        // so it outlives the borrow the context holds.
+        //
+        // `bm25_provider` (the "legacy" index-provider slot), NOT the
+        // "preferred" `bm25_search_provider` that `ContextConfig` steers you
+        // toward: view-policy enforcement on a hit only works in index-provider
+        // mode. Search-provider mode has no local flakes and fails closed
+        // (`operator.rs`), so it would return zero rows for every
+        // policy-enforced query. Do not "upgrade" this.
         let __index_provider = $crate::FlureeIndexProvider::new($self);
         let $cfg = ContextConfig {
             tracker: Some($tracker),
--- a/fluree-db-api/src/view/mod.rs
+++ b/fluree-db-api/src/view/mod.rs
@@ -92,11 +92,19 @@ macro_rules! view_context_config {
         } else {
             None
         };
+        // Wire the BM25 index provider so embedded `f:searchText` graph-source
+        // queries execute in-process on the single-graph view path too (parity
+        // with `query_dataset_with_bm25` and the dataset path). Zero-cost
+        // `&Fluree` wrapper, only consulted by the `IndexSearch` operator, so
+        // non-search queries are unaffected. Declared before `$cfg` so it
+        // outlives the borrow the context holds.
+        let __index_provider = $crate::FlureeIndexProvider::new($self);
         let $cfg = ContextConfig {
             tracker: Some($tracker),
             cancellation: $options.cancellation.clone(),
             policy_enforcer: __db.policy_enforcer().cloned(),
             r2rml: $r2rml,
+            bm25_provider: Some(&__index_provider),
             binary_store: __db.binary_store.clone(),
             binary_g_id: __db.graph_id,
             dict_novelty: __db.dict_novelty.clone(),
--- a/fluree-db-api/tests/it_graph_source_bm25.rs
+++ b/fluree-db-api/tests/it_graph_source_bm25.rs
@@ -728,6 +728,74 @@ async fn bm25_query_connection_with_idx_pattern() {
     );
 }
 
+/// Regression: embedded `f:searchText` must resolve through the STANDARD query
+/// path (`query_connection`), not only the dedicated `query_connection_with_bm25`.
+///
+/// This is exactly what the HTTP `POST /v1/fluree/query` route executes (via
+/// `query_from()`). Before the dataset/view execution context wired a
+/// `FlureeIndexProvider` as `bm25_provider`, this returned
+/// `InvalidQuery("BM25 IndexSearch requires ExecutionContext.bm25_search_provider
+/// or bm25_provider (not configured)")`. It must now succeed.
+#[tokio::test]
+async fn bm25_embedded_search_via_plain_query_connection() {
+    let fluree = FlureeBuilder::memory().build_memory();
+
+    let ledger_id = "bm25/plain-qc:main";
+    let ledger0 = support::genesis_ledger(&fluree, ledger_id);
+    let tx = json!({
+        "@context": { "ex":"http://example.org/" },
+        "@graph": [
+            { "@id":"ex:doc1", "@type":"ex:Doc", "ex:title":"Rust programming guide", "ex:author":"Alice" },
+            { "@id":"ex:doc2", "@type":"ex:Doc", "ex:title":"Rust and WebAssembly", "ex:author":"Bob" },
+            { "@id":"ex:doc3", "@type":"ex:Doc", "ex:title":"Python for beginners", "ex:author":"Charlie" }
+        ]
+    });
+    let _ledger = fluree.insert(ledger0, &tx).await.expect("insert failed");
+
+    let index_query = json!({
+        "@context": { "ex":"http://example.org/" },
+        "where": [{ "@id":"?x", "@type":"ex:Doc", "ex:title":"?title" }],
+        "select": { "?x": ["@id", "ex:title"] }
+    });
+    let cfg = Bm25CreateConfig::new("plain-qc-search", ledger_id, index_query);
+    let created = fluree
+        .create_full_text_index(cfg)
+        .await
+        .expect("create index failed");
+
+    let idx_query = json!({
+        "@context": { "ex":"http://example.org/", "f": "https://ns.flur.ee/db#" },
+        "from": ledger_id,
+        "where": [
+            {
+                "f:graphSource": &created.graph_source_id,
+                "f:searchText": "rust",
+                "f:searchLimit": 10,
+                "f:searchResult": {"f:resultId": "?doc", "f:resultScore": "?score"}
+            },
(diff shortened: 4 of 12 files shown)


fluree-db-api/src/view/dataset_query.rs
```diff
@@ -688,6 +695,7 @@ impl Fluree {
             dataset: Some(&runtime_dataset),
             policy_enforcer: primary.policy_enforcer().cloned(),
             r2rml: Some((r2rml.provider, r2rml.table_provider)),
+            bm25_provider: Some(&index_provider),
```
aaj3f, line 705: *Blocking.*

**This wiring site has zero test coverage, and so does its twin at `:841`.** I verified by mutation (each experiment re-run after `touch`ing the changed sources, with a confirmed recompile every time, so no stale-binary artifact): setting `bm25_provider: Some(&index_provider)` back to `None` at line 698 leaves every test in this PR green; same for line 841; only mutating the `view/mod.rs:107` macro site turns `bm25_embedded_search_via_plain_query_connection` red.

So the one new test exercises the single-graph view path exclusively, and the two dataset-path changes — 15 of the ~23 non-doc library lines in this PR, and the path the PR description names as the one serving `POST /v1/fluree/query` — are unguarded.

This is not hypothetical drift. The exact failure this PR fixes is a *missing struct field initializer*, which is the single easiest thing for a future refactor of `dataset_query.rs` to drop while CI stays green. The HTTP route would silently regress to the same `InvalidQuery("BM25 IndexSearch requires ExecutionContext.bm25_search_provider or bm25_provider (not configured)")` that motivated this PR.

The good news is the wiring **is** correct and covering it is cheap. I wrote a scratch test that forces the dataset path with a two-ledger `from`; it passes on HEAD and fails with the pre-fix error when both dataset sites are mutated. Roughly:

```rust
// Multi-ledger `from` => dataset execution path (execute_dataset_*).
let idx_query = json!({
    "@context": { "ex":"http://example.org/", "f": "https://ns.flur.ee/db#" },
    "from": [l1, l2],
    "where": [{
        "f:graphSource": &created.graph_source_id,
        "f:searchText": "rust",
        "f:searchLimit": 10,
        "f:searchResult": {"f:resultId": "?doc", "f:resultScore": "?score"}
    }],
    "select": ["?doc", "?score"]
});
let result = fluree.query_connection(&idx_query).await
    .expect("DATASET path must serve f:searchText");
```

Please add that as a second test alongside the existing one.

fluree-db-api/tests/it_graph_source_bm25.rs
```diff
@@ -728,6 +728,74 @@ async fn bm25_query_connection_with_idx_pattern() {
     );
 }
 
+/// Regression: embedded `f:searchText` must resolve through the STANDARD query
```
aaj3f, line 731: *Blocking.*

**No policy regression test guards the path this PR opens.** The suite's policy test (`bm25_search_enforces_view_policy_on_indexed_flake`, line 903) drives `query_connection_with_bm25` — the API that was *already* reachable before this PR. This PR's whole effect is to make the `IndexSearch` operator reachable from the standard, policy-enforced query route, so that route is exactly where policy enforcement now needs pinning.

I read the enforcement path (`operator.rs:247-265` → `search_readability.rs`) and it looks correct, so I am not claiming a live leak. But "correct today, untested" on a newly-widened surface where the failure mode is *returning documents a policy says the caller cannot read* is the wrong risk to carry.

Please add a policy-enforced variant of the new test — ideally on the dataset path, since `search_hit_readable` behaves differently there (`ActiveGraphs::Many` with >1 graph routes to `conservative_multi_graph`, which hides hits whenever any active graph has a policy).

fluree-db-cli/src/cli.rs
```diff
@@ -1125,6 +1145,84 @@ pub enum Commands {
...
+        #[arg(long)]
+        k1: Option<f64>,
+
+        /// BM25 b (document-length normalization, 0..=1). Default 0.75.
+        #[arg(long)]
+        b: Option<f64>,
+    },
+
+    /// Drop (retract) a BM25 full-text index and delete its snapshots.
+    Drop {
```
aaj3f, line 1190: *Blocking (convention).*

**`fluree bm25 drop` deletes data with no confirmation flag, breaking this CLI's own convention for destructive commands.** It retracts the nameservice record *and* deletes snapshot blobs immediately.

Every sibling does better: dropping an Iceberg-family graph source requires `--force` (`cli.rs:2648`, comment "Required flag to confirm deletion"), and so does `fluree drop` for a ledger (`cli.rs:382`, enforced at `commands/drop.rs:15` with "use --force to confirm deletion of '{name}'").

BM25 is in the same category by this PR's own words — `commands/bm25.rs:3` opens "BM25 is a Fluree *graph source* (like Iceberg/R2RML)". A rebuildable index is a softer loss than a ledger, but re-creating one over a large corpus is expensive, and the asymmetry is surprising precisely because the adjacent graph-source drop does gate. Suggested fix:

```rust
Drop {
    /// Index graph-source alias to drop (e.g. "silver-search:main").
    #[arg(long)]
    index: String,

    /// Required flag to confirm deletion
    #[arg(long)]
    force: bool,
},
```

fluree-db-api/src/view/dataset_query.rs
```diff
@@ -678,6 +678,13 @@ impl Fluree {
                 (None, None, None, None, None)
             };
 
+        // Wire the BM25 index provider so embedded `f:searchText` graph-source
+        // queries execute in-process (parity with `query_dataset_with_bm25`).
+        // It is a zero-cost `&Fluree` wrapper and is only consulted by the
+        // `IndexSearch` operator, so non-search queries are unaffected. Declared
+        // before `config` so it outlives the borrow held by the context.
+        let index_provider = crate::FlureeIndexProvider::new(self);
```
aaj3f, line 693: **This bypasses the purpose-built opt-in that already exists for exactly this.** `FromQueryBuilder::with_index_providers()` (`query/builder.rs:888`, doc comment "Enable BM25/Vector index providers") sets `GraphSourceMode::IndexProviders`, and `QueryCore::validate()` still rejects that mode outright at `query/builder.rs:151` with *"Index provider mode (.with_index_providers()) is not yet supported by query builders; use fluree.query_connection_with_bm25() or fluree.query_dataset_with_bm25() instead"*.

After this PR the API tells users the opposite of the truth in both directions: calling `.with_index_providers()` — the documented way to turn BM25 on — hard-errors, while *not* calling it silently gives you working BM25. Whatever you decide about unconditional wiring, that stub error message is now stale and actively misleading. Either implement `GraphSourceMode::IndexProviders` to set the provider (and make the unconditional wiring the default), or delete the mode and its builder methods. Leaving a booby-trapped opt-in in place is the one outcome worth avoiding.

To be clear on the broader altitude question: the PR does **not** invent a parallel BM25 dispatch. It fills the existing `ContextConfig.bm25_provider` slot that `Bm25SearchOperator` already reads, in the same place and shape as the existing R2RML providers. On rubric §6 that is extending the shared mechanism, not special-casing. (`SearchBackend` in `fluree-search-service/src/backend/mod.rs:37` is the standalone service's backend abstraction and is not the in-engine dispatch — it is not what this PR should have used.)

fluree-db-api/src/view/dataset_query.rs
```diff
@@ -678,6 +678,13 @@ impl Fluree {
                 (None, None, None, None, None)
             };
 
+        // Wire the BM25 index provider so embedded `f:searchText` graph-source
+        // queries execute in-process (parity with `query_dataset_with_bm25`).
```
aaj3f: The comment justifies choosing `bm25_provider` as "parity with `query_dataset_with_bm25`", which undersells it. The real reason is that view-policy enforcement on a hit **only works in index-provider mode**: per `operator.rs:243-257`, search-provider mode has no local flakes and fails closed, hiding every hit whenever a policy is active. So wiring the field the docs call "preferred" (`context.rs:120-123`) would make BM25 return zero rows for every policy-enforced query.

Since `context.rs` actively steers a future reader toward the "preferred" provider, please record the policy reason here so nobody "upgrades" this and silently breaks search under policy.

fluree-db-cli/src/cli.rs
```diff
@@ -1125,6 +1145,84 @@ pub enum Commands {
...
+
+        /// Branch for the index graph source (default "main").
+        #[arg(long, default_value = "main")]
+        branch: String,
+
+        /// Inline indexing query (FQL / JSON-LD).
+        #[arg(short = 'e', long = "query")]
+        query: Option<String>,
+
+        /// Read the indexing query from a file (`-` for stdin).
```
aaj3f: *Nit.*

The help text says ``Read the indexing query from a file (`-` for stdin)``, but `-` is not handled: `resolve_input` wraps it as `InputSource::File("-")` and `read_input` calls `std::fs::read_to_string("-")` (`fluree-db-cli/src/input.rs:43`), so `-f -` fails with `failed to read -`. There is no `-` special-case anywhere in `input.rs`.

Implicit stdin does work (pipe with neither `-e` nor `-f`), so either drop the "`-` for stdin" clause or say "pipe via stdin". Note the two other stdin mentions in this file (`cli.rs:2349`, `:2560`) use a different, actually-handled `@-` convention.

fluree-db-cli/src/commands/bm25.rs
```diff
@@ -0,0 +1,253 @@
+//! BM25 full-text search index commands.
```
aaj3f, line 1: *Nit.*

253 new CLI lines land with no tests, while `fluree-db-cli/tests/integration.rs` carries 276 tests covering essentially every other command group (I confirmed zero of them mention bm25).

`run_list`'s staleness computation in particular has real logic worth pinning — especially the bare-alias fallback at `commands/bm25.rs:84-87` that retries a dependency lookup with an implicit `:main`.

fluree-db-api/src/view/mod.rs
```diff
@@ -92,11 +92,19 @@ macro_rules! view_context_config {
         } else {
             None
         };
+        // Wire the BM25 index provider so embedded `f:searchText` graph-source
```
aaj3f, line 95: *Praise.*

Declaring the provider before the `ContextConfig` so it outlives the borrow, with a comment saying exactly that, is the kind of note that survives a refactor. Doing it identically at all three sites, matching the existing R2RML threading, is right.

docs/indexing-and-search/bm25.md
```diff
@@ -108,6 +108,8 @@ The analyzer is not configurable — it always uses the English pipeline for con
 
 ## Querying BM25 Indexes
 
+BM25 search is expressed with the `f:` namespace predicates in a query's `where` clause. It is a **JSON-LD (FQL) feature** and executes over the standard query endpoint — `POST /v1/fluree/query` — so search hits can be joined with ledger data in a single query. It is **not** available via SPARQL: the SPARQL parser does not recognize the `f:` search block, so those predicates would be matched as ordinary triples and return nothing.
```
aaj3f, line 111: *Praise.*

Documenting plainly that BM25 is JSON-LD-only and that **SPARQL silently returns nothing** because the parser matches `f:` predicates as ordinary triples is exactly the honest failure-mode disclosure that saves a user a day.

I verified the claim: `fluree-db-sparql` never constructs `Pattern::IndexSearch` (its only two mentions, `lower/mod.rs:365` and `lower/construct.rs:164`, are skip-arms). Equally good is narrowing the "near-term gap" in `docs/graph-sources/bm25.md` to the remote-delegation half rather than quietly deleting it.

aaj3f (changes requested):
@christophediprima thank you for another excellent PR!

This fixes a real and gap — embedded BM25 was documented as the default deployment mode while being unreachable over HTTP — and it fixes it at the right altitude, filling the existing `ContextConfig.bm25_provider` slot exactly where the R2RML providers are already threaded rather than bolting in a parallel dispatch. The choice of the "legacy" provider over the "preferred" one is also correct, since only index-provider mode can enforce view policy on a hit, though the code comment credits the wrong reason. The CLI is a genuinely thin wrapper that reimplements nothing, and the docs are honest about what still does not work. 

My one set of requests is not to the design; it is that the majority of the library change is untested. Mutation testing shows both `dataset_query.rs` sites can be reverted to `None` with the entire suite still green — and that is the path the PR itself names as serving `POST /v1/fluree/query`. Add a multi-ledger dataset test and a policy-enforced variant (I confirmed both are ~60 lines and do go red on mutation), gate `bm25 drop` behind `--force` like its Iceberg sibling, and this is a clear approve.

- **Patterns/abstractions:** ⚠️ Extends the existing `Bm25IndexProvider` slot correctly and invents no parallel construct — but leaves `FromQueryBuilder::with_index_providers()` as a stub that hard-errors while the capability it advertises now works without it (`query/builder.rs:151`).
- **Performance (speed first, memory second):** ✔️ No hot-path regression. `FlureeIndexProvider` is one `&Fluree` reference with no allocation, stored once per query in a struct already built once per query, and read only by the `IndexSearch` operator. Noted for the record: omitting `f:searchLimit` yields `top_k(usize::MAX)`, which scores the whole corpus via `score_all()` with no WAND early termination (`scoring.rs:400-405`) — pre-existing, but newly reachable from the public route.
- **Testing:** ✖️ The new test is correctly wired into `grp_graphsource.rs:16-17` and does run, but covers only 1 of 3 wiring sites (proven by per-site mutation). No policy test on the newly opened path; no server-level HTTP test; no CLI test for 253 new lines.

Blocking on the dataset-path test, the policy test, and the `bm25 drop --force` gate before merge.

christophediprima:
Thank you — the mutation testing was the useful part of this review, and it turned out to be more useful
than either of us expected: **your suggested fix only closes half the gap.**

I added the multi-ledger dataset test you sketched and confirmed it goes red exactly as you predicted.
Then I re-ran mutation **per site**, and the two `dataset_query.rs` sites are not reachable by the same
query — a multi-ledger `from` isn't what selects between them, the builder method is:

| Entry point | Lands in | Wiring site |
|---|---|---|
| `query_connection`, `query_from().execute_formatted()` | `execute_dataset_internal_with_r2rml` → `execute_dataset_into_with_r2rml` | `:698` (`:705` after this change) |
| `query_from().execute_tracked()` | `execute_dataset_tracked` → `execute_dataset_tracked_with_r2rml` | `:841` (`:855` after this change) |

So the one dataset test leaves the tracked site green. There are now three tests, and each of the three
wiring sites is killed by exactly one of them — no overlap:

```
dataset_query.rs:705 -> None   ⇒ bm25_embedded_search_via_dataset_path            FAILED (others green)
dataset_query.rs:855 -> None   ⇒ bm25_embedded_search_via_tracked_dataset_path    FAILED (others green)
view/mod.rs macro    -> None   ⇒ the two plain-route tests                        FAILED (others green)
```

Everything else you asked for:

- **Policy test** on the newly-opened path — added, driving plain `query_connection`. See the thread for
  why I put it there rather than on the dataset path.
- **`bm25 drop --force`** — gated, matching the wording of both siblings.
- **`with_index_providers()`** — deleted, along with the mode. Note it was **five** methods across two
  files, not one, with zero callers anywhere in the repo.
- **Provider comment** — rewritten at all three sites to record the policy reason rather than "parity".
- **Stdin help text** — fixed.
- **CLI tests** — added, seven of them, including a mutation-checked guard on the bare-alias fallback.

The only thing I've left alone is the `f:searchLimit` → `score_all` observation, which read as noted-for-
the-record rather than a request — happy to add a doc note if you'd rather it not go unrecorded.

fluree-db-api/src/view/dataset_query.rs
christophediprima: Fixed, but your remedy needed extending — see the table in my top-level reply. A multi-ledger `from`
reaches `execute_dataset_into_with_r2rml` only; `execute_dataset_tracked_with_r2rml` needs
`query_from().execute_tracked()`. Your scratch test (which I used essentially verbatim, thank you) kills
`:698`; I added `bm25_embedded_search_via_tracked_dataset_path` for `:841`.

I verified this the same way you did — per-site mutation, each experiment `touch`ed and recompiled — and
the matrix is in the top-level reply. The three tests now kill the three sites 1:1.

fluree-db-api/tests/it_graph_source_bm25.rs
christophediprima: Added `bm25_embedded_search_enforces_policy_via_plain_query_connection`: a control query over the plain
route proving both rust docs match, then the same search under a `f:onProperty ex:title` policy scoped to
Alice-authored docs — `doc1` survives, `doc2`'s hit is dropped.

I put it on the plain `query_connection` route rather than the dataset path, and I want to flag the
trade-off rather than bury it. Per your own note, `ActiveGraphs::Many` with >1 graph routes to
`conservative_multi_graph`, which hides hits whenever *any* active graph carries a policy. If that is
right, a dataset-path policy test can only assert "nothing came back" — which passes just as happily if
enforcement broke entirely, i.e. it is a weaker guard than the single-graph one. The plain route gives the
fine-grained assertion (one doc in, one doc out) on the surface this PR actually widens.

**I have not verified the multi-graph behaviour myself** — that is your observation, not my measurement.
If you want the dataset variant as well, I'll add it as an explicit pin of "BM25 under policy across >1
ledger returns nothing", which seems worth recording as a documented limitation in its own right.

fluree-db-cli/src/cli.rs
christophediprima: Fixed exactly as suggested — `force: bool` with the same `/// Required flag to confirm deletion` doc as
`cli.rs:2648`, enforced in `run_drop` with the same message shape as `commands/drop.rs:15`
(`"use --force to confirm deletion of '{index}'"`).

fluree-db-api/src/view/dataset_query.rs
christophediprima: Agreed, and I took the delete option: the mode, the `validate()` arm, `set_index_providers()`, and the
builder methods are gone.

Two notes from doing it. It was **five** `with_index_providers()` methods, not one —
`query/builder.rs:254`, `:540`, `:888` and `graph_query_builder.rs:106`, `:434` — and **zero callers**
anywhere in the repo: no tests, no docs, no internal use. Implementing the mode instead would have made it
a no-op flag, since this PR turns the capability on unconditionally at all three sites; an opt-in with
nothing to opt into is the same booby trap in a different shape.

Strictly this removes public methods from a published crate. I think that's fine and worth saying out
loud: `validate()` rejected the mode outright, so every existing call already hard-errors — no working
caller can break.

fluree-db-api/src/view/dataset_query.rs
christophediprima: Fixed at all three sites. The comment now reads:

> `bm25_provider` (the "legacy" index-provider slot), NOT the "preferred" `bm25_search_provider` that
> `ContextConfig` steers you toward: view-policy enforcement on a hit only works in index-provider mode.
> Search-provider mode has no local flakes and fails closed (`operator.rs`), so it would return zero rows
> for every policy-enforced query. Do not "upgrade" this.

fluree-db-cli/src/cli.rs
christophediprima: Fixed — the help now reads "Read the indexing query from a file (or pipe it via stdin)". I did not add
`@-` support; implicit stdin already works and adding a second convention to one subcommand seemed worse
than describing the one that exists.

fluree-db-cli/src/commands/bm25.rs
christophediprima: Added — seven end-to-end tests in `tests/integration.rs`, same `assert_cmd` + temp-dir style as the rest of
the file.

Both bits of logic you singled out are pinned:

- **Staleness.** A commit after the index is built flips `STALE` to `YES` and makes `--stale` emit the
  alias; `sync` clears it again. That is the maintenance loop, so it should not be able to regress quietly.
- **The bare-alias fallback.** `bm25 create --ledger docs` persists the dependency verbatim as `docs` while
  ledger records are keyed `docs:main`, so the lookup only matches through the implicit-`:main` retry. I
  checked this test earns its place the same way you checked the wiring sites: deleting the `.or_else(...)`
  makes exactly this test fail and no other. Without the retry `LEDGER_T` renders `-`, the index reports
  itself never-stale, and a `--stale`-driven loop skips it forever.

Plus the empty-list message, the fresh-index listing, `drop` refusing without `--force` and succeeding with
it, and the two `create` input rejections (a name containing `:`, and malformed JSON).

docs/indexing-and-search/bm25.md
```diff
@@ -315,11 +317,28 @@ The storage format is selected automatically based on the storage backend:
...
+## CLI Management
+
+The `fluree bm25` command group manages indexes without writing Rust. Each command runs in-process against local storage (no server round-trip), so it works under `docker exec` against a running server's data directory (native storage uses per-file locks, so this is safe alongside a running server).
+
+| Command | Description |
+|---------|-------------|
+| `fluree bm25 create --name <n> --ledger <l> -f <index-query.json>` | Build a BM25 index over a source ledger from an indexing query |
+| `fluree bm25 list [--stale]` | List indexes with their source ledger, `INDEX_T`/`LEDGER_T`, and staleness. `--stale` prints `name:branch` per line for scripting |
+| `fluree bm25 sync --index <gs>` | Incrementally sync an index up to its source ledger's head (prints the watermark `old -> new` and upserted/removed counts) |
+| `fluree bm25 drop --index <gs>` | Retract the index and delete its snapshot blobs |
```
aaj3f: *Optional (worth fixing before merge).* The CLI reference table documents `fluree bm25 drop --index <gs>`, but commit `43e48a11` made `--force` mandatory, so the documented command now hard-errors with `use --force to confirm deletion of '<gs>'`.

This is self-inconsistency introduced inside the PR: the docs commit (`ec569110`) shipped the table two commits before the `--force` gate landed, and the response round didn't circle back. A reader copy-pasting from the docs hits the error on their first try.

```diff
-| `fluree bm25 drop --index <gs>` | Retract the index and delete its snapshot blobs |
+| `fluree bm25 drop --index <gs> --force` | Retract the index and delete its snapshot blobs (`--force` required) |
```

fluree-db-api/src/view/dataset_query.rs
```diff
@@ -688,6 +702,7 @@ impl Fluree {
             dataset: Some(&runtime_dataset),
             policy_enforcer: primary.policy_enforcer().cloned(),
             r2rml: Some((r2rml.provider, r2rml.table_provider)),
+            bm25_provider: Some(&index_provider),
```
aaj3f, line 705: *Optional (documentation).* On the multi-ledger dataset path this PR newly opens, any active view policy silently yields **zero** search hits rather than filtered hits, and this limitation isn't stated in the user-facing docs.

The mechanism is `search_readability.rs:74-76` — `conservative_multi_graph` returns "not readable" whenever any of >1 active graphs carries a policy — combined with the single-ledger fast path at `dataset_query.rs:171-172`, which means the dataset path is only ever entered with >1 ledger. So the conservative branch is the *only* branch a policy-enforced dataset search can take.

I confirmed this empirically: the same `f:onProperty` policy that keeps doc1 and drops doc2 on the plain route drops both on a two-ledger `from`, returning 2 rows without the policy and 0 rows with it, silently and with no error. This is fail-closed and therefore not a leak, and it pre-dates the PR on `query_dataset_with_bm25` — but this PR is what makes it reachable from `POST /v1/fluree/query`, where "my search returns nothing once I enable policies" is a confusing outcome to debug.

A sentence in the Querying section of `docs/indexing-and-search/bm25.md` would cover it — e.g.: *a policy-enforced `f:searchText` query spanning more than one ledger conservatively returns no hits, because per-flake readability cannot be resolved across a multi-graph scope; scope such searches to a single ledger.* Commenting here because `search_readability.rs` is not in this diff.

aaj3f (approved):
All three blocking asks are fixed, and I verified each by mutation rather than by reading the response.

**(A) Dataset-path test coverage — verified fixed.** The round-1 gap was that both `dataset_query.rs` wiring sites had zero coverage. Per-site mutation now gives a clean 1:1 matrix, each experiment run after `touch`ing the source with a confirmed recompile: reverting `dataset_query.rs:705` to `bm25_provider: None` fails exactly `bm25_embedded_search_via_dataset_path` (3 passed, 1 failed); reverting `:855` fails exactly `bm25_embedded_search_via_tracked_dataset_path` (3 passed, 1 failed); reverting `view/mod.rs:113` fails exactly the two plain-route tests (14 passed, 2 failed). The author's diagnosis that a multi-ledger `from` does not select between the two dataset sites — the builder method does — is correct, and `execute_tracked()` is indeed the only route to `execute_dataset_tracked_with_r2rml`.

**(B) Policy regression test — verified fixed, and the placement reasoning is right.** Short-circuiting `search_hit_readable` (`fluree-db-query/src/search_readability.rs:37`) to `return Ok(true)` turns `bm25_embedded_search_enforces_policy_via_plain_query_connection` red, so the test pins the actual enforcement mechanism rather than an incidental empty result. Your `conservative_multi_graph` reasoning also checks out against the code, and it is in fact stronger than stated: `search_readability.rs:74-76` hides every hit when any of >1 active graphs carries a policy, *and* `dataset_query.rs:171-172` short-circuits a single-ledger `from` to the view path — so the dataset path is only ever reachable with >1 ledger, and a selective-enforcement assertion there is impossible to write, not merely weak. The plain route is the only placement that can carry a real assertion. I confirmed the resulting behavior empirically (control 2 rows, policy 0 rows on a two-ledger query); the one gap left is that this limitation isn't documented, which I've raised as an optional item.

**(C) `bm25 drop --force` — verified fixed, exact-pattern.** The flag (`cli.rs:1195-1197`) carries the same `/// Required flag to confirm deletion` doc comment as both siblings (`cli.rs:388` for `drop`, `:2656` for `iceberg drop`), and the enforcement in `commands/bm25.rs:236-241` reproduces `commands/drop.rs:15-19` verbatim, error type and message shape included. `bm25_drop_requires_force` asserts the refusal, that the index survives the refusal, and that `--force` succeeds.

**(D) `with_index_providers()` deletion — verified complete.** `git grep` for `with_index_providers`, `IndexProviders`, and `set_index_providers` returns zero hits across all tracked files, so nothing dangles. Clippy is clean both with `--all-features --all-targets` and under default features, confirming the `GraphSourceMode` variant removal doesn't break the non-`iceberg` build. Removing a builder mode whose entire behavior was to return an error is the right call over making it work.

Also verified: (E) the provider-slot comment is present at all three sites and its claim is accurate (`bm25/operator.rs:248-257`); (F) the help-text fix is correct — `input.rs:43` opens `-` as a literal file, so the old "`-` for stdin" was wrong while the implicit stdin path at `:31-33` does work; (G) the seven CLI tests pass by name and have teeth — inverting the staleness comparison (`commands/bm25.rs:89`) fails both staleness tests, and removing only the `:main` retry (`:87`) fails exactly `bm25_list_resolves_a_branchless_dependency_alias`, matching your stated PR claim precisely.

Recommend merge once the two one-line docs touch-ups and the trailer removal are done — none of them gate the code, which is correct and now genuinely well covered.

zonotope:
It looks like all of the outstanding issues have been addressed so I am merging. Thanks @christophediprima!

PR #1545, opened 2026-07-24
Limit recursion depth on turtle imports

### LINKED ISSUE #1480: parser(turtle): recursion depth ceiling for nested constructs
Reified-triple nesting recurses with no depth guard (2 bytes/level via `<<`), joining the pre-existing unguarded `[ … ]` property-list and `( … )` collection recursion class. The test harness runs with `RUST_MIN_STACK=8MB`; production stacks don't. One global depth ceiling across the three recursive productions would close the class.

From the PR #1454 local review (optional item).

zonotope:
Issue #1480 reported that the Turtle parser's recursive constructs — blank-node property lists (`[ … ]`), collections (`( … )`), and reified triples (`<< … >>`) — recurse without a depth limit, so a small adversarial document can overflow the stack. A Rust stack overflow is uncatchable and aborts the whole process, making this a denial-of-service vector on any surface that parses untrusted Turtle. The bug was confirmed by reproduction before the fix: a ~1MB document of 100,000 nested property lists aborted the test process with SIGABRT.

## Changes

- `fluree-graph-turtle/src/error.rs`: new `pub const MAX_NESTING_DEPTH: u32 = 128`, documented alongside the existing `MAX_INPUT_BYTES` input bound. The three constructs share one counter, so mixed nesting is bounded too. 128 matches common practice (serde_json's default recursion limit) and is far beyond real-world Turtle nesting.
- `fluree-graph-turtle/src/parser.rs`: a `nesting_depth` field and a `with_nesting` closure wrapper that owns the entire invariant in one place — the ceiling check (erroring with a positioned parse error past the limit), the increment, and an unconditional decrement — so call sites cannot increment without the matching decrement, and the counter stays correct even if parse-error recovery is ever added. Applied at all four recursion entry points: `parse_blank_node_property_list`, `parse_collection`, `parse_collection_as_list` (the object-position indexed-list variant, a fourth entry beyond the issue's list of three — caught because an unguarded outer level let collections nest one past the ceiling), and `parse_reified_triple`. Guarding these four bounds every recursion cycle: reified-triple nesting through the subject/object positions loops back through `parse_reified_triple`, and annotation bodies are already capped at one level by the star gate.
- Documentation: a `### Parser Limits` subsection in `docs/transactions/turtle.md` covering both the nesting ceiling and the pre-existing 4 GiB input bound, a resource-limits note under the Turtle entry in `docs/reference/compatibility.md`, and a `# Limits` section in the crate docs with intra-doc links to the two constants.

## Tests

Five cases in the parser's test module, with single-`format!`/`str::repeat` input generators matching the Cypher crate's recursion-test idiom:

- `nesting_at_the_ceiling_parses` — all three constructs parse at exactly depth 128, pinning the inclusive boundary against a stricter-by-one guard regression.
- `nesting_past_the_ceiling_errors_cleanly` — all three constructs fail at depth 129 with the "maximum depth" parse error (the reified case runs through the star-capable `StarSink`, since the default sink rejects star forms outright).
- `deeply_nested_property_lists_do_not_overflow_the_stack` — the original 100k DoS shape returns the clean error. This is not redundant with the depth-129 case: under a regression where the ceiling check stops short-circuiting before the recursive descent, 129 frames still parse or error cleanly while only the 100k input reproduces the overflow.
- `sibling_nesting_does_not_accumulate_depth` — 512 sibling nestings (deliberately more than the 128 ceiling) parse fine, proving only the nesting chain counts.
- One lexer subtlety the generators encode: bare `()` lexes as a single `Nil` token, so nested collections must carry a real innermost item to count levels correctly.

Fixes #1480

### DIFF
--- a/docs/reference/compatibility.md
+++ b/docs/reference/compatibility.md
@@ -148,6 +148,10 @@ Fluree parses Turtle 1.1:
 - Collections
 - Blank nodes
 
+Resource limits: recursive constructs (property lists, collections, reified
+triples) may nest at most 128 levels deep, and a single parse accepts at most
+4 GiB of input; either limit produces a clean parse error.
+
 **Specification:** https://www.w3.org/TR/turtle/
 
 ### JSON Web Signature (JWS)
--- a/docs/transactions/turtle.md
+++ b/docs/transactions/turtle.md
@@ -329,6 +329,18 @@ riot --validate data.ttl
 }
 ```
 
+### Parser Limits
+
+The parser enforces two resource limits; input past either is rejected with a
+parse error rather than ingested:
+
+- Blank-node property lists (`[ ... ]`), collections (`( ... )`), and reified
+  triples (`<< ... >>`) may nest at most 128 levels deep. Wide structures are
+  unaffected — a flat list of any length counts as one level; only the nesting
+  chain is limited.
+- A single parse call accepts at most 4 GiB (`u32::MAX` bytes) of input. Bulk
+  import splits files into chunks well below this automatically.
+
 ## Performance Tips
 
 ### 1. Use Batch Import
--- a/fluree-graph-turtle/src/error.rs
+++ b/fluree-graph-turtle/src/error.rs
@@ -54,6 +54,14 @@ impl TurtleError {
 /// wrapped offset that would later panic when slicing the source.
 pub const MAX_INPUT_BYTES: usize = u32::MAX as usize;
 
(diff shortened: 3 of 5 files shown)


bplatz (approved):
Verified locally: 106 tests pass, clippy and fmt clean. Guard placement covers every recursion cycle, flat collections don't consume depth, and the ceiling holds with room to spare on the stack sizes this actually runs on.

PR #1546, opened 2026-07-26
fix: gate WHERE-level early dedup on aggregate duplicate-insensitivity

bplatz:
## Problem

On a 6-hop existential chain (every intermediate variable bound but never projected), the pipeline carries full duplicate multiplicity through every join. On PrimeKG (8.7M flakes):

```sparql
SELECT (COUNT(DISTINCT ?p3) AS ?n) WHERE {
  ?dis pkg:diseaseName "epilepsy" ; pkg:associatedWithProtein ?p1 .
  ?p1  pkg:participatesInPathway ?pw .
  ?pw  pkg:hasPathwayProtein ?p2 .
  ?p2  pkg:targetedByDrug ?d1 .
  ?d1  pkg:interactsWithDrug ?d2 .
  ?d2  pkg:targetsProtein ?p3 .
}
```

The final join consumed a **10.17M-row intermediate carrying only 4,086 distinct `?d2` values** (2,490× redundancy) — 74.7s, while a hand-rewritten nested `SELECT DISTINCT` version returned the identical answer in 185ms. Join order was already optimal; the problem was purely that nothing collapsed duplicates once a variable went dead.

## Root cause

The WHERE planner already has per-step early dedup (trim dead columns via live-var analysis, then insert a streaming `DistinctOperator`), but it was gated on `query.output.is_distinct()` — i.e. `SELECT DISTINCT` only. That gate is wrong in **both** directions:

1. **Too narrow:** an aggregate query whose aggregates are all duplicate-insensitive (`COUNT(DISTINCT …)`, `MIN`, `MAX`, `SAMPLE`, any `DISTINCT`-marked aggregate) can't observe WHERE-output multiplicity either, but never got the dedup.
2. **Unsound:** `SELECT DISTINCT (COUNT(?x) AS ?c)` *did* fire the dedup and returned wrong counts — confirmed on main with a failing test before the fix (returned 2 where the correct count is 4). The outer DISTINCT dedups *result* rows after aggregation; it does not license pre-aggregation dedup.

## Fix

`build_operator_tree_inner` now computes `where_dedup_safe`:

- **grouping present** → every aggregate must satisfy the new `AggregateFn::duplicate_insensitive()` (any `DISTINCT`-marked variant, plus `MIN`/`MAX`/`SAMPLE`)
- **no grouping** → `SELECT DISTINCT`, as before

The flag is threaded through the WHERE planner under that name (renamed from `distinct_query`). No change to the dedup mechanism itself, subquery planning, fast paths, or the hash-join planner.

## Results (PrimeKG, release build)

| Query | Before | After | Result |
|---|---|---|---|
| Chain as written | 74,660 ms | **188.1 ms** | 2341 ✓ |
| Hand-dedup rewrite | 184.9 ms | 188.3 ms | 2341 ✓ |

The plan now shows a `DistinctOperator` after each join, collapsing every intermediate to its distinct live tuples (193 → 464 → 5,790 → 4,949 → 4,086 rows instead of 1 → 17 → 85 → 1,700 → 15,300 → 10.17M).

Dedup is only inserted at steps where trimming actually dropped a dead variable, so star shapes and queries whose variables stay live are untouched, as are all plain-`COUNT`/`SUM` aggregates.

## Testing

- Plan-shape guards (`it_query_explain.rs`): dedup fires for `COUNT(DISTINCT)`/`MAX` chains; blocked for plain `COUNT`, mixed aggregates, and `SELECT DISTINCT` + plain `COUNT` (exactly one outer `DistinctOperator`)
- Correctness fixtures in `it_query_sparql.rs` (fan-in chain: plain count keeps multiplicity, distinct count unchanged) with JSON-LD twins in `it_query_aggregates.rs`
- Unit test partitioning every `AggregateFn` variant by duplicate-insensitivity
- Full W3C SPARQL compliance suite: 36/36 suites pass (aggregates + grouping included)
- 1,213 `fluree-db-query` unit tests, `grp_query` (380), `grp_query_sparql` (290), `grp_query_reason` (100), `it_query_cypher` (228), `it_query_sparql_indexed` (85) — all green
- clippy `--all-features --all-targets -D warnings` clean

Design doc updated: `docs/design/query-execution.md` § "WHERE-level early dedup".

### DIFF
--- a/docs/design/query-execution.md
+++ b/docs/design/query-execution.md
@@ -162,12 +162,24 @@ streaming `DistinctOperator` so duplicates collapse before the next join.
 `build_operator_tree_inner`): collapsing duplicate rows is only legal when
 downstream cannot observe WHERE-output multiplicity —
 
-- **grouping present** → every aggregate must be *duplicate-insensitive*:
-  `AggregateFn::duplicate_insensitive()` = any `DISTINCT`-marked aggregate
-  (`COUNT/SUM/AVG/MEDIAN/VARIANCE/STDDEV/GROUP_CONCAT/collect DISTINCT`) plus
-  `MIN`/`MAX`/`SAMPLE`. A single plain `COUNT`/`SUM`/… blocks dedup for the
-  whole WHERE.
+- **grouping present** → *both* of:
+  - every aggregate must be *duplicate-insensitive*:
+    `AggregateFn::duplicate_insensitive()` = any `DISTINCT`-marked aggregate
+    (`COUNT/SUM/AVG/MEDIAN/VARIANCE/STDDEV/GROUP_CONCAT/collect DISTINCT`) plus
+    `MIN`/`MAX`/`SAMPLE`. A single plain `COUNT`/`SUM`/… blocks dedup for the
+    whole WHERE.
+  - no variable may pass through the grouping stage *raw*: every var in
+    `VariableDeps::required_aggregate_vars` must be a `GROUP BY` key or an
+    aggregate output. A non-key variable that survives grouping is emitted as
+    a per-group **list** (`Binding::Grouped`, the JSON-LD grouped-projection
+    feature — SPARQL rejects the shape), and that list observes row
+    multiplicity. This clause also covers the `GROUP BY ?g` case with *no*
+    aggregation stage, where the aggregate check alone is vacuously true.
 - **no grouping** → the query itself must be `SELECT DISTINCT`.
+
+Both clauses are checked against `variable_deps`; when it is `None`
+(wildcard/boolean/construct) dedup is off anyway, since projection pushdown —
+and therefore the live-var trimming that triggers dedup — is disabled.
 
 An outer `SELECT DISTINCT` over an aggregate query does **not** license WHERE
 dedup — it dedups *result* rows after aggregation, while a plain `COUNT` under
@@ -176,7 +188,11 @@ correctness bug fixed alongside the aggregate-aware gate).
 
 The dedup is inserted only at steps where trimming actually dropped a dead
 variable, so queries whose variables all stay live (e.g. same-subject stars
-feeding the projection) pay nothing. Subqueries dedup at their own boundary
+feeding the projection) pay nothing. Note the memory trade: `DistinctOperator`
+holds an uncapped, non-spilling hash set of the distinct rows seen at each
+insertion point, so an aggregate query that previously streamed (e.g. a
+`MAX`-only chain whose intermediates are large and already near-distinct)
+becomes resident-memory-bound for no gain. The gate errs toward the speed win. Subqueries dedup at their own boundary
 (`apply_solution_modifiers` applies the subquery's `DISTINCT`) rather than
 per-step.
 
--- a/docs/design/query-execution.md
+++ b/docs/design/query-execution.md
@@ -141,3 +141,42 @@ needs the history sidecar, defers). Two lanes split on whether novelty is presen
   materialized through a novelty-aware graph view (the persisted-dict
   `encoded_sid` form would not resolve them).
 
+## WHERE-level early dedup (projection + distinct between joins)
+
+Deep existential chains (`?a p1 ?b . ?b p2 ?c . ?c p3 ?x` where only `?x` is
+projected/aggregated) can carry compounding duplicate multiplicity: once `?a`
+is dead, every distinct `?b` is repeated once per `?a` that produced it, and
+each join multiplies the redundancy into the next hop. On a real 6-hop
+biomedical query this reached 10.17M intermediate rows carrying 4,086 distinct
+values (2,490× redundancy, ~400× slowdown).
+
+The WHERE planner (`fluree-db-query/src/execute/where_plan.rs`,
+`build_sequential_join_block` / `build_sequential_triple_chain`) counters this
+with **early dedup**: at each join step it computes the live-variable set
+(post-WHERE required vars ∪ vars referenced by not-yet-executed
+patterns/filters/binds), trims dead columns from the join output
+(`with_out_schema`), and — when dedup is licensed — wraps the step in a
+streaming `DistinctOperator` so duplicates collapse before the next join.
+
+**Soundness gate** (`where_dedup_safe`, computed in
+`build_operator_tree_inner`): collapsing duplicate rows is only legal when
+downstream cannot observe WHERE-output multiplicity —
+
+- **grouping present** → every aggregate must be *duplicate-insensitive*:
+  `AggregateFn::duplicate_insensitive()` = any `DISTINCT`-marked aggregate
+  (`COUNT/SUM/AVG/MEDIAN/VARIANCE/STDDEV/GROUP_CONCAT/collect DISTINCT`) plus
+  `MIN`/`MAX`/`SAMPLE`. A single plain `COUNT`/`SUM`/… blocks dedup for the
+  whole WHERE.
+- **no grouping** → the query itself must be `SELECT DISTINCT`.
+
+An outer `SELECT DISTINCT` over an aggregate query does **not** license WHERE
+dedup — it dedups *result* rows after aggregation, while a plain `COUNT` under
+it still observes pre-aggregation multiplicity (this exact miswiring was a
+correctness bug fixed alongside the aggregate-aware gate).
+
+The dedup is inserted only at steps where trimming actually dropped a dead
+variable, so queries whose variables all stay live (e.g. same-subject stars
+feeding the projection) pay nothing. Subqueries dedup at their own boundary
+(`apply_solution_modifiers` applies the subquery's `DISTINCT`) rather than
+per-step.
+
--- a/fluree-db-api/tests/it_query_aggregates.rs
+++ b/fluree-db-api/tests/it_query_aggregates.rs
@@ -715,3 +715,78 @@ async fn aggregates_arithmetic_over_min_max_jsonld() {
         normalize_rows(&json!([["sup:partA", 4], ["sup:partB", 4]]))
     );
 }
+
+/// JSON-LD parity for WHERE-level early-dedup semantics (see the SPARQL
+/// twins in `it_query_sparql.rs`): a fan-in chain where two subjects reach
+/// the same intermediate node, so the intermediate relation carries
+/// duplicate rows once the head var is projected away.
+///
+///   a1 --p1--> b1 --p2--> {10, 20}
+///   a2 --p1--> b1
+async fn seed_chain_fan_in(fluree: &MemoryFluree, ledger_id: &str) -> MemoryLedger {
+    let ledger0 = genesis_ledger(fluree, ledger_id);
+    let insert = json!({
+        "@context": {"ex": "http://example.org/ns/"},
+        "@graph": [
+            {"@id": "ex:a1", "ex:p1": {"@id": "ex:b1"}},
+            {"@id": "ex:a2", "ex:p1": {"@id": "ex:b1"}},
+            {"@id": "ex:b1", "ex:p2": [10, 20]}
+        ]
+    });
+    fluree
+        .insert(ledger0, &insert)
+        .await
+        .expect("insert chain fixture")
+        .ledger
+}
+
+/// `count-distinct` over the fan-in chain: 2 distinct ?x values. The
+/// duplicate-insensitive aggregate licenses WHERE-level early dedup; the
+/// result must be identical either way.
+#[tokio::test]
+async fn count_distinct_chain_fan_in_jsonld() {
+    let fluree = FlureeBuilder::memory().build_memory();
+    let ledger = seed_chain_fan_in(&fluree, "agg/dedup-cd:main").await;
+
+    let query = json!({
+        "@context": {"ex": "http://example.org/ns/"},
+        "select": ["(as (count-distinct ?x) ?c)"],
+        "where": [
+            {"@id": "?a", "ex:p1": "?b"},
+            {"@id": "?b", "ex:p2": "?x"}
+        ]
+    });
+    let result = support::query_jsonld(&fluree, &ledger, &query)
+        .await
+        .expect("query");
+    let rows = result.to_jsonld(&ledger.snapshot).expect("jsonld");
+    assert_eq!(rows, json!([[2]]), "2 distinct ?x values: {rows}");
+}
+
+/// `selectDistinct` + plain `count` over the fan-in chain: the outer DISTINCT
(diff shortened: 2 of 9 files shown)


fluree-db-query/src/execute/operator_tree.rs
```diff
@@ -2938,12 +2938,27 @@ fn build_operator_tree_inner(
...
 
+    // WHERE-level early dedup (project away dead vars, collapse duplicate rows
+    // between joins) is sound only when downstream cannot observe row
+    // multiplicity. With a grouping phase, that is decided by the aggregates
+    // alone: every one must be duplicate-insensitive (COUNT/SUM/AVG/… DISTINCT,
+    // or MIN/MAX/SAMPLE). An outer SELECT DISTINCT does NOT license it — it
+    // dedups result rows *after* aggregation, so a plain COUNT under it still
+    // observes pre-aggregation multiplicity. Without grouping, SELECT DISTINCT
+    // is exactly the license.
+    let where_dedup_safe = match query.grouping.as_ref() {
```
aaj3f, line 2953: *Blocking.*

The gate consults `g.aggregates()` and nothing else, but aggregates are not the only consumer of WHERE-output multiplicity under a grouping phase. `GroupByOperator` renders a **projected variable that is neither a GROUP BY key nor an aggregate/bind output** as `Grouped(Vec<Binding>)` containing *all* values in the group, duplicates included (`fluree-db-query/src/groupby.rs:1-6`).

That per-group list is a duplicate-**sensitive** consumer, and it is a deliberate, pinned Fluree JSON-LD feature — `fluree-db-api/tests/it_query_grouping.rs:10-13` calls it "a long-standing Fluree feature" and `:107` pins it precisely so a refactor cannot silently break it. This PR silently turns those lists from multisets into sets whenever a WHERE variable dies mid-chain.

Two things make it reachable. Under `Grouping::Implicit` or `Explicit` with an all-duplicate-insensitive aggregate, the gate returns `true` and never looks at the projection. And under `Grouping::Explicit { aggregation: None }` (a dedup-only `GROUP BY ?g`), `g.aggregates()` is an **empty iterator**, so `.all(…)` is *vacuously* `true` and dedup turns on with no aggregate having licensed it at all.

Minimal repro, JSON-LD surface, run against both base and this branch. Fixture: `a1,a2,a3 --ex:p1--> b1 --ex:p2--> 10` and `a4 --ex:p1--> b2 --ex:p2--> {20,30}`. Query `{"select": ["?x", "(as (max ?x) ?m)"], "where": [{"@id":"?a","ex:p1":"?b"},{"@id":"?b","ex:p2":"?x"}]}`. **main** returns `[[[10,10,10,20,30],30]]`; **this branch** returns `[[[10,20,30],30]]`. The dedup-only-`GROUP BY` form shows the same thing: `{"select":["?b","?x"], "groupBy":["?b"]}` over `a1,a2 --p1--> b1 --p2--> 10` gives `[["ex:b1",[10,10]]]` on main and `[["ex:b1",[10]]]` here.

The existing pinned test does not catch it because its fixture has no fan-in, so nothing is ever deduped. Note the result now depends on an optimizer decision: delete the unused `?a` from the WHERE and the list changes shape. This is SPARQL-invisible (SPARQL rejects ungrouped projection under `GROUP BY`), so it is a JSON-LD-surface-only regression — which is exactly the parity direction the PR's own new tests were written to cover.

Suggested fix — I applied this, forced a rebuild, and re-ran the suites: **683/683 still green**, both positive plan-shape guards (`explain_chain_count_distinct_inserts_early_dedup`, `explain_chain_max_inserts_early_dedup`) still pass, so the PrimeKG headline optimization is untouched (its projection is the aggregate output), and the repro returns main's answer again.

```rust
let where_dedup_safe = match query.grouping.as_ref() {
    Some(g) => {
        let aggregates_ok = g
            .aggregates()
            .all(|spec| spec.function.duplicate_insensitive());
        // A projected var that is neither a GROUP BY key nor an
        // aggregate/bind output comes back as a per-group LIST
        // (JSON-LD grouped projection) — that list observes row
        // multiplicity, so dedup would silently drop entries.
        let mut visible: HashSet<VarId> = g.group_by_vars().collect();
        visible.extend(g.aggregates().map(|s| s.output_var));
        visible.extend(g.binds().map(|(v, _)| *v));
        let no_grouped_lists = query
            .output
            .projected_vars()
            .is_some_and(|p| p.iter().all(|v| visible.contains(v)));
        aggregates_ok && no_grouped_lists
    }
    None => query.output.is_distinct(),
};
```

Treating a `None` projection (wildcard) as unsafe costs nothing: with no projection pushdown, `required_where_vars` is `None`, `live_vars` is `None`, and the dedup never fires anyway. Please also add a JSON-LD regression test with a fan-in fixture next to `jsonld_groupby_ungrouped_select_projects_grouped_list`, since that test as written cannot fail on this.

docs/design/query-execution.md
```diff
@@ -141,3 +141,42 @@ needs the history sidecar, defers). Two lanes split on whether novelty is presen
...
+(post-WHERE required vars ∪ vars referenced by not-yet-executed
+patterns/filters/binds), trims dead columns from the join output
+(`with_out_schema`), and — when dedup is licensed — wraps the step in a
+streaming `DistinctOperator` so duplicates collapse before the next join.
+
+**Soundness gate** (`where_dedup_safe`, computed in
+`build_operator_tree_inner`): collapsing duplicate rows is only legal when
+downstream cannot observe WHERE-output multiplicity —
+
+- **grouping present** → every aggregate must be *duplicate-insensitive*:
```
aaj3f: *Blocking (small).*

The doc's gate description says "grouping present → every aggregate must be *duplicate-insensitive*", which is technically true but leaves a reader unable to predict the two cases that actually bite: a grouping phase with **no aggregation stage** (`GROUP BY ?g` alone) satisfies it vacuously and enables dedup, and the gate does not consider ungrouped projected variables at all.

Once the finding on `operator_tree.rs:2949` is fixed, please state both explicitly — the vacuous-`GROUP BY` path and the grouped-list exclusion — so the next author extending this gate knows what the license actually covers.

fluree-db-query/src/ir/grouping.rs
```diff
@@ -114,6 +114,20 @@ impl AggregateFn {
...
+    /// are collapsed to one — i.e. it observes only the *set* of values, not
+    /// their multiplicity. True for every `DISTINCT`-marked variant plus
+    /// `Min`/`Max`/`Sample` (order statistics / arbitrary pick).
+    ///
+    /// This is the soundness condition for WHERE-level early dedup: when
+    /// every aggregate in the grouping phase is duplicate-insensitive, the
+    /// planner may project away dead variables and collapse duplicate rows
+    /// between joins without changing results (see
+    /// `where_plan::build_sequential_join_block`).
+    pub fn duplicate_insensitive(&self) -> bool {
```
aaj3f, line 132: *Optional.*

`duplicate_insensitive()` is a non-exhaustive `matches!` delegating to another non-exhaustive `matches!` (`is_distinct()`, `:101`). The default is `false` = duplicate-sensitive, which is the correctness-safe direction and the right choice — worth saying so in the doc comment.

But the consequence is that a future `AggregateFn` variant silently loses the optimization with no signal, and the unit test at `:353` is a hand-written enumeration that will still pass. An exhaustive `match self { … }` here would make a new variant a compile error, which is the repo's usual bar for a soundness-critical partition.

fluree-db-query/src/execute/where_plan.rs
```diff
@@ -1527,12 +1527,12 @@ fn build_sequential_join_block(
...
             pending_binds = new_binds;
             pending_filters = new_filters;
-            operator = if ctx.distinct_query && pruned_vars.as_ref().is_some_and(|s| !s.is_empty())
-            {
-                Some(Box::new(DistinctOperator::new(child)))
-            } else {
-                Some(child)
-            };
+            operator =
+                if ctx.where_dedup_safe && pruned_vars.as_ref().is_some_and(|s| !s.is_empty()) {
```
aaj3f, line 1531: Real location: `fluree-db-query/src/distinct.rs:24-26` ("Memory usage grows with the number of unique rows seen").

*Optional (memory).* This PR extends early dedup from `SELECT DISTINCT` queries (which were going to materialize a dedup hash set anyway) to `MIN`/`MAX`/`SAMPLE`/`COUNT(DISTINCT)` aggregate queries, which previously streamed with O(1) planner-side memory.

`DistinctOperator` keeps an uncapped, non-spilling `HashMap` of every distinct row per insertion point. On the motivating shape this is a massive win; on a `MAX`-only chain whose intermediates are large and already near-distinct it is new resident memory for no gain.

Worth a sentence in the design doc so the trade is on the record — the speed win is not in question.

(Commenting here because `fluree-db-query/src/distinct.rs` is not in this diff.)

fluree-db-api/tests/it_query_explain.rs
```diff
@@ -448,3 +448,141 @@ async fn explain_logical_plan_preserves_compound_structure() {
...
+        "a duplicate-sensitive aggregate in the mix must block early dedup: {physical}"
+    );
+}
+
+/// `SELECT DISTINCT` over a plain `COUNT`: the outer DISTINCT dedups *result*
+/// rows after aggregation and must NOT license WHERE-level dedup (that was the
+/// bug — the intermediate relation was collapsed before the count). Exactly
+/// one `DistinctOperator` may appear: the outer one above the aggregation.
+#[tokio::test]
+async fn explain_select_distinct_plain_count_only_outer_distinct() {
```
aaj3f, line 574: *Praise.*

`explain_select_distinct_plain_count_only_outer_distinct` asserting **exactly one** `DistinctOperator` (rather than `>= 0` or `> 0`) is the test that actually pins the bug: it distinguishes "the outer result-row DISTINCT survived" from "no dedup at all".

I confirmed it has teeth in both directions — forcing the gate to `true` fails it, forcing it to `false` leaves it passing while the two positive guards fail. Please keep that exact-count assertion if the plan shape is ever refactored.

fluree-db-api/tests/it_query_aggregates.rs
```diff
@@ -715,3 +715,78 @@ async fn aggregates_arithmetic_over_min_max_jsonld() {
...
 }
+
+/// JSON-LD parity for WHERE-level early-dedup semantics (see the SPARQL
+/// twins in `it_query_sparql.rs`): a fan-in chain where two subjects reach
+/// the same intermediate node, so the intermediate relation carries
+/// duplicate rows once the head var is projected away.
+///
+///   a1 --p1--> b1 --p2--> {10, 20}
+///   a2 --p1--> b1
+async fn seed_chain_fan_in(fluree: &MemoryFluree, ledger_id: &str) -> MemoryLedger {
```
aaj3f, line 726: *Praise.*

Adding JSON-LD twins for the SPARQL fixtures is the right call for a shared-IR fix and it paid off: I ran both against the base engine and **both** fail with `[[2]]` vs `[[4]]`, proving the released bug was reachable from both front-ends, not just SPARQL.

I worked the fixture's arithmetic by hand (4 chain rows, 2 distinct `?x`) — the expectations are correct.

aaj3f (approved):
This is a genuinely important fix and the analysis behind it is right: the old `query.output.is_distinct()` gate was wrong in both directions, and the unsound direction has been shipping silently wrong `COUNT`s since the v4 baseline — I confirmed on the base engine that `SELECT DISTINCT (COUNT(?x) AS ?c)` returns 2 where the answer is 4, on both the SPARQL and JSON-LD surfaces. The altitude is right (it generalizes the existing license rather than special-casing the slow query), the `AggregateFn` classification is correct for all 13 variants with a correctness-safe default, the HAVING / ORDER BY / sub-SELECT / mixed-aggregate placement holes I probed are all genuinely closed, and the tests have teeth in both directions. 

I think really just one thing I'd strongly suggest before merge: the gate inspects aggregates but not the projection, so a projected non-key, non-aggregate variable — which Fluree's JSON-LD surface deliberately renders as a per-group *list* — silently loses its duplicates, and a dedup-only `GROUP BY` enables dedup via a vacuously-true `.all()` on an empty aggregate set. I have a verified eight-line fix that keeps all 683 tests green and leaves the PrimeKG optimization intact; with that plus the corresponding doc sentence, this is a clear approve.

- **Performance (speed first, memory second):** ✔ speed / ⚠️ memory. Hot path (`fluree-db-query` planner) is touched; **no plan shape loses dedup that previously had it correctly** — the only ON→OFF flip is the unsound `SELECT DISTINCT` + duplicate-sensitive-aggregate case. Gate cost is O(#aggregates) once at plan time, nothing per-row. Memory caveat: `DistinctOperator`'s uncapped hash set now applies to aggregate queries that previously streamed. No bench covers this path (`fluree-db-query/benches/` is `vector_math.rs` only), so CI cannot detect this optimization silently regressing.
- **Testing:** ⚠️ Unit + integration tests that genuinely run and genuinely fail when the gate is broken (19/683 fail with the gate forced on; both positive guards fail with it forced off). Missing: any coverage of the ungrouped-projection / dedup-only-`GROUP BY` interaction, which is where the regression lives.

Maybe worth flagging in the release notes that this fixes silently wrong aggregate results present in v4.0.0 through v4.1.4.

bplatz:
@aaj3f Confirmed the grouped-list regression on both repros (implicit-grouping and dedup-only `GROUP BY`) — main returns the multiset, this branch returned the set. Fixed in 2618b01f7, with a slightly wider gate than suggested:

```rust
let keys: HashSet<VarId> = g.group_by_vars().collect();
let agg_outputs: HashSet<VarId> = g.aggregates().map(|s| s.output_var).collect();
let no_raw_passthrough = variable_deps.as_ref().is_some_and(|d| {
    d.required_aggregate_vars.iter().all(|v| keys.contains(v) || agg_outputs.contains(v))
});
```

`required_aggregate_vars` is what the grouping stage's *output* must carry, so it already folds in projection, HAVING, post-binds and ORDER BY. That closes one hole the projection-only version leaves: a post-bind whose *expression* reads a non-key var passes a bind-output whitelist but still reads a grouped list. `None` deps stays unsafe, as you noted, at no cost.

Added two JSON-LD regression tests with a fan-in fixture next to the pinned one; both fail with the gate reverted. Doc now states the vacuous-`GROUP BY` path, the grouped-list exclusion, and the `DistinctOperator` memory trade. `duplicate_insensitive()` is now an exhaustive match.

CI-parity locally: clippy clean, 9831/9832 nextest (the one failure is a LocalStack port-mapping flake), explain guards and the PrimeKG shape untouched. Will call out the v4.0.0–v4.1.4 wrong-aggregate fix in the release notes.

PR #1547, opened 2026-07-26
Limit recursion depth in the SPARQL and JSON-LD parsers

zonotope:
The SPARQL query parser and the JSON-LD expansion/context parser were both recursive-descent over untrusted input with no depth ceiling, so a small adversarial payload could overflow the stack. A Rust stack overflow is uncatchable and aborts the whole process, making each a denial-of-service vector:

- **SPARQL** is reachable from every unauthenticated read-query endpoint (`parse_sparql_to_ir` feeds every query view, plus SPARQL UPDATE). `ASK { { { { … } } } }`, `FILTER((((…))))`, deeply nested property paths, `<< … >>` quoted triples, or `<<( … )>>` triple terms all overflow at parse time.
- **JSON-LD** expansion recurses over the value tree. Text inputs are incidentally shielded by serde_json's default 128-level deserialization limit, but a programmatically built `JsonValue` bypasses that and reaches the unguarded recursion.

This branch adds a shared per-parser depth ceiling (128, matching serde_json's default and well beyond real-world nesting) and closes every recursion cycle. It was found and hardened via adversarial review: the initial guards missed four cycles of the same crash class, all now covered and regression-tested (two were confirmed to SIGABRT before the fix).

## SPARQL (`fluree-db-sparql`)

- `TokenStream` gains a `depth` field and `MAX_PARSE_DEPTH = 128`, with `try_enter_recursion`/`leave_recursion` primitives, a `Result`-lane `with_recursion_guard` closure wrapper (expressions, property paths), and a diagnostic-lane `Parser::with_recursion_guard` (pattern/term parsers) that emits a new `DiagCode::NestingTooDeep` (S010). The counter lives on the stream so it spans EXISTS sub-parsers that share it, and mirrors the call stack (not the token position) so backtracking needs no snapshotting.
- Guards at every recursion entry: `parse_group_graph_pattern` (nested groups plus OPTIONAL/GRAPH/SERVICE/MINUS/UNION arms), `parse_expression`, `parse_unary_expr` (stacked unary operators, which bypass `parse_expression`), `parse_blank_node_property_list`, `parse_collection`, `parse_path_alternative` (grouped paths), and — added during review — `parse_quoted_triple` (`<< … >>`) and `parse_triple_term_value` (`<<( … )>>`). The RDF-star forms parse fully before lowering rejects them, so the crash was at parse time.
- The Result-lane guard records the S010 diagnostic at the source, so the specific depth code reaches the output regardless of how each of ~9 downstream sites maps the returned `Err` string (several turn it into a generic `ExpectedToken`; the GROUP BY/ORDER BY speculative sites drop it entirely on backtrack).
- `add_diagnostic` is capped at `MAX_DIAGNOSTICS = 256`: a deeply-nested query that trips the ceiling would otherwise retain one heap-allocated diagnostic per surplus token during error recovery (O(input) diagnostic memory on attacker-controlled input). Error recovery was measured to be linear, not quadratic, so no separate CPU-DoS remains.

## JSON-LD (`fluree-graph-json-ld`)

- `MAX_EXPAND_DEPTH = 128` with a shared `guard_depth(idx)` helper. Expansion has no explicit depth counter, but the pre-existing `idx` path vector grows at least one element per recursion edge, so its length bounds the depth. The guard is called at both recursion entry points — `expand_node_internal` and `parse_node_value` — because `@list`/`@set` container chains self-recurse through `parse_node_value` without re-entering `expand_node_internal`, and would otherwise bypass a guard placed only at the former.
- Scoped-`@context` parsing (`ParsedContext::parse`) is a separate recursion cycle — wrapped `@context`, sequential context arrays, and a term definition's own `@context` all recurse through it. The public signature is unchanged; a private `parse_guarded(base, context, depth)` core threads a counter through the two internal helpers and refuses past the same ceiling.
- `JsonLdError::NestingTooDeep` is a unit variant with the ceiling formatted into the message from the const, and `MAX_EXPAND_DEPTH` is `pub(crate)` (it has no external consumers), matching the SPARQL side.

## Tests

- SPARQL: exact-boundary (128 parses, 129 errors), all recursion families error cleanly at adversarial depth (groups, FILTER parens, stacked unary, property lists, collections, grouped paths, `<< >>`, `<<( )>>`), a mixed-construct test proving the shared counter, a wide-but-shallow sibling test, and a diagnostic-count-cap test. The depth assertion requires the S010 code (previously a loose "maximum depth" substring, since the Result lane lacked the code).
- JSON-LD: below-ceiling parses unchanged; property-chain, `@list`/`@set`, and scoped-`@context` chains each error with `NestingTooDeep`; and a test pinning that deeply nested text is still rejected at serde_json deserialization (the shield the expansion guard backs up).
- The two RDF-star SPARQL cycles and both JSON-LD cycles were each confirmed to overflow the stack with their guard disabled and to pass with it restored.

## Notes

Cypher already had an equivalent guard (`MAX_PARSE_DEPTH`, `enter_recursion`/`leave_recursion`); a shared cross-crate helper was considered and rejected — the three parsers have incompatible state and error types, and this branch's closure-wrapper form is a deliberate improvement over Cypher's manual enter/leave pairing.


fluree-graph-json-ld/src/expand.rs
```diff
@@ -202,6 +202,7 @@ fn parse_node_value(
     idx: &[JsonValue],
     strict: bool,
 ) -> Result<Vec<JsonValue>> {
+    guard_depth(idx)?;
```
aaj3f, line 205: *Optional, but the highest-value item here.*

This guard is load-bearing and **completely uncovered** — I removed it and all 13 JSON-LD depth tests stayed green.

The reason is that `nested_containers` (`expand.rs:1045`) interposes a property key at every level, so the chain alternates `parse_node_value` → `expand_node_internal` and is caught by the *other* guard at `expand.rs:558`. The shape this guard actually defends is a `@list` nested directly inside a `@list`, with no property key between levels, which self-recurses in `parse_node_value` without ever re-entering `expand_node_internal` — exactly the case the doc comment describes.

I verified both directions: with the guard, `{"p": {"@list": {"@list": … }}}` 20 000 deep returns `NestingTooDeep` cleanly; with only the `expand_node_internal` guard, the same input overflows the stack and SIGABRTs. As it stands a future refactor can delete this line and CI stays green while reopening the DoS.

Suggested fixture to add alongside `nested_containers`:

```rust
/// A `@list`/`@set` nested DIRECTLY inside another, with no property key
/// between levels: this self-recurses through `parse_node_value` without
/// re-entering `expand_node_internal`, so only the guard there stops it.
fn direct_container_chain(depth: usize, key: &str) -> JsonValue {
    let mut v = json!({"@id": "http://example.org/leaf"});
    for _ in 0..depth {
        v = json!({ key: v });
    }
    json!({ "http://example.org/p": v })
}
```

Build the fixture (and drop it) on a large-stack thread — `serde_json::Value` has recursive `Clone` and `Drop`, which is what the existing test's "depth is kept moderate" comment is working around.

fluree-graph-json-ld/src/expand.rs
```diff
@@ -525,13 +526,36 @@ fn get_context(node_map: &Map<String, JsonValue>) -> Option<&JsonValue> {
...
 }
 
+/// Maximum expansion depth for nested JSON-LD values. Expansion recurses
+/// over the value tree through [`expand_node_internal`] and
+/// [`parse_node_value`], and while text inputs are already bounded by
+/// serde_json's default 128-level deserialization recursion limit, a
+/// programmatically built `JsonValue` bypasses that — without this guard,
+/// deep nesting would overflow the stack and abort the process (a Rust
+/// stack overflow is not catchable).
+pub(crate) const MAX_EXPAND_DEPTH: usize = 128;
```
aaj3f, line 540: *Optional.*

The effective ceiling is not uniformly 128. Each `@list`/`@set` level pushes two `idx` entries (the property key, then the `@list` marker), so container chains are refused at **64** levels of user-visible nesting while plain node chains get 128 — I measured both.

That is defensible (the guard bounds frames, not user-facing levels), but the constant's doc comment reads as a flat promise of 128, and nothing pins the 64. Worth one sentence in the doc comment so the next reader doesn't treat a container-chain rejection at 64 as a bug.

fluree-db-sparql/src/parse/stream.rs
```diff
@@ -11,6 +11,23 @@ use crate::lex::{Token, TokenKind};
...
+/// Maximum recursion depth for nested parser productions — group graph
+/// patterns, expressions, property paths, blank-node property lists, and
+/// collections all recurse through the parser, so unbounded nesting lets a
+/// small hostile query overflow the stack and abort the process (a Rust
+/// stack overflow is not catchable). Past this ceiling parsing fails with a
+/// [`DiagCode::NestingTooDeep`](crate::diag::DiagCode::NestingTooDeep)
+/// diagnostic. All productions share one counter, so mixed nesting is
+/// bounded too; it lives on the stream so it spans sub-parsers (EXISTS
+/// groups) that share it.
+pub(crate) const MAX_PARSE_DEPTH: u32 = 128;
```
aaj3f, line 23: *Optional, consistency.*

After #1545 lands alongside this, the repo has four front-end depth ceilings with three names, three visibilities, and two values: `MAX_PARSE_DEPTH = 128` `pub(crate)` here, `MAX_EXPAND_DEPTH = 128` `pub(crate)` in `fluree-graph-json-ld/src/expand.rs:536`, `MAX_NESTING_DEPTH = 128` **`pub`** in `fluree-graph-turtle/src/error.rs` (on `main`), and `MAX_PARSE_DEPTH = **256**` private in `fluree-db-cypher/src/parse/stream.rs:15`.

The PR body mentions Cypher's guard but not that it is twice as permissive. I agree with rejecting a shared cross-crate helper — the state and error types genuinely differ — but the *number* is a security parameter, not an implementation detail, and Cypher being the odd one out looks unintentional. Either align Cypher to 128 or note why it earns 256.

fluree-db-sparql/src/parse/query/term.rs
```diff
@@ -251,9 +261,14 @@ impl super::Parser<'_> {
                     }
                     return Some(Verb::Path(path));
                 }
-                Err(_) => {
-                    // Restore position and try simple predicate
+                Err(msg) => {
```
aaj3f, line 264: *Praise, and worth keeping.*

This is the only non-guard behavior change in the diff, and it is the right call: surfacing the path parser's message instead of falling through to the generic "expected predicate or property path" is what makes a depth-exceeded verb report S010 rather than a misleading syntax error.

I confirmed both the old and new branches return `None`, so this is a message change and not a control-flow change, and the W3C syntax suite (36/36) agrees.

fluree-db-sparql/src/parse/stream.rs
```diff
@@ -31,7 +53,52 @@ impl TokenStream {
...
+    /// Run `f` one recursion level deeper, erroring past
+    /// [`MAX_PARSE_DEPTH`]. Owns both sides of the depth bookkeeping for the
+    /// `Result`-based parsers (expressions, property paths).
+    ///
+    /// On refusal it records the [`DiagCode::NestingTooDeep`] diagnostic
+    /// directly, so the specific depth code reaches the output regardless of
+    /// how the caller maps the returned `Err` string (several sites turn it
+    /// into a generic `ExpectedToken`, and the `GROUP BY`/`ORDER BY`
+    /// speculative sites drop it entirely on backtrack).
+    pub(crate) fn with_recursion_guard<T>(
```
aaj3f, line 85: *Praise.*

Recording the S010 diagnostic at the point of refusal, rather than trusting ~9 downstream sites to preserve the `Err` string, is the detail that makes this user-visible.

I verified end-to-end: at depth 129 the S010 diagnostic is first in the list, and `fluree-db-api/src/query/helpers.rs:466-480` refuses the recovered AST whenever error-severity diagnostics exist, so the caller gets "nesting exceeds the maximum depth of 128" and never executes a truncated query.

aaj3f (approved):
This closes a real, uncatchable-abort DoS on both remaining unguarded front-ends, and it does so by extending the mechanism the repo already chose for Cypher and Turtle rather than inventing a parallel one. I did not take the completeness claim on trust: I built a call graph over all 165 parser functions and recomputed the strongly-connected components with the eight guard functions removed, and every recursion cycle in `fluree-db-sparql/src/parse/` disappears — the guards form a feedback vertex set, so there is no unguarded path, including the RDF-star cycles the PR says adversarial review turned up late. Putting the counter on `TokenStream` instead of `Parser` is the subtle decision that makes it hold, since EXISTS spins up a fresh `Parser` over the same stream and a parser-local counter would have reset on every `FILTER EXISTS { … }`. On the question of whether 128 is actually low enough for a production stack: measured in a release build, the worst legitimate shape at the ceiling costs about 512 KB (SPARQL FILTER parens, ~4 KB/level) and JSON-LD expansion about 384 KB — roughly a 4× margin on a 2 MB tokio worker stack, so the ceiling does its job where it matters. 

The one note really considering before merge is not a defect in the shipped code but in what protects it: the `@list`/`@set` guard at `expand.rs:205` is load-bearing yet no test covers it, which I confirmed by removing it and watching all 13 JSON-LD depth tests stay green while a direct `@list`-in-`@list` chain SIGABRTs.

- **Testing:** ⚠️ 798/798 pass and the SPARQL guards have real teeth (removing the quoted-triple guard SIGABRTs the test), but the JSON-LD `parse_node_value` guard is uncovered — the existing container fixture is caught by the other guard.

---

**Follow-up: the JSON-LD *query* parser is the remaining unguarded front-end.** *Optional, scope/follow-up, and the answer to the SPARQL↔JSON-LD parity question.* The finding is in `fluree-db-query/src/parse/where_clause.rs:150` and `:512`, which this PR does not touch.

The parity rule pairs the SPARQL front-end with the **JSON-LD query** front-end, and it is worth being precise that `fluree-graph-json-ld` — the crate this PR hardens — is the JSON-LD *document* expansion/context crate on the import and transaction path, **not** the JSON-LD query parser.

That parser lives in `fluree-db-query/src/parse/`, and it has the same unguarded shape this PR is fixing: `parse_where_array_element` (`where_clause.rs:150`) and `parse_subquery_patterns` (`where_clause.rs:512`) are mutually recursive — reachable through nested `optional`, `union`, `minus`, `exists`, and `graph` blocks, and through `node_map::parse_node_map` for nested node objects — and I confirmed there is no depth counter anywhere in that directory.

It is shielded in practice on the wire path by exactly the mechanism this PR's body cites for JSON-LD documents: a query arriving as HTTP text is capped at 128 nesting levels by serde_json's deserialization limit before the parser sees it. But the programmatic path bypasses that shield in precisely the way the PR describes — a `JsonValue` assembled in-process (`fluree-db-mcp`, the CLI, reasoner- or policy-generated queries) reaches the recursion directly.

So after this merges, three of the four front-ends have an explicit guard and the JSON-LD query parser is the remaining one relying solely on serde_json. Not this PR's job, but it belongs in a follow-up issue so the hardening pass isn't recorded as finished.

PR #1549, opened 2026-07-29
perf: eliminate per-flake staging cost in bulk upserts

bplatz:
## Problem

Bulk JSON-LD upserts staged at ~8 ms/flake (STAGING-PERF-BUG report: 35–70 s per ~6k-flake transaction against a 5 MB / 451k-flake ledger, with essentially all time inside `stage()` and `retractions ≈ 0`).

## Root cause

`generate_upsert_deletions` issues one bound-subject point query per (subject, predicate) tuple to find existing values to retract. For subjects **not yet in the persisted subject dictionary** — the dominant case when bulk-upserting new entities — `BinaryScanOperator` cannot resolve an `s_id`, sets `unresolved_bound_subject_iri`, and degrades to a **full PSOT predicate-partition scan with per-row subject-IRI decoding**. On the repro ledger that's ~56k row-decodes per subject for `rdf:type` and the multi-valued `blockingKey` alone; a 600-subject upsert spent 13.27 s of 13.29 s staging in this loop, producing zero retractions.

Resolved subjects were never the problem: the same 3,000-tuple loop over existing subjects runs in ~40 ms (~13 µs per point probe).

## Fix

**Commit 1 — subject-existence pre-check in `generate_upsert_deletions`:** group predicates by subject and resolve existence once per subject:
- persisted presence via the subject reverse dictionary (`find_subject_id_by_parts` + full-IRI `find_subject_id`), authoritative under the canonical namespace-encoding invariants;
- novelty presence via one Sid-space overlay walk per referenced graph (no dictionary translation, works even when dict-novelty is uninitialized).

Subjects present in neither cannot have values to retract and skip their per-predicate queries entirely. Everything else keeps the existing query path.

**Commit 2 — cache the persisted `p_id → Sid` table on `BinaryIndexStore`:** every scan `open()` rebuilt the full predicate table (one dict resolve + namespace encode per predicate in the ledger), and the `binary_range` materialization loops re-resolved the predicate per emitted row. The table is now built once per store instance (`OnceLock<Arc<[Sid]>>`; stores are immutable once shared behind `Arc`) and shared; novelty-only predicates keep their ephemeral p_ids in a per-operator overlay map. Removes an O(predicate-count) fixed cost per scan open that becomes significant for wide-schema ledgers.

## Results (repro clone, 600 subjects / 6,600 flakes)

| scenario | deletion phase | staging total |
|---|---:|---:|
| new subjects, before | 13,270 ms | 13,292 ms |
| new subjects, after | **0.9 ms** | **3.5 ms** |
| existing subjects (values changed) | ~40 ms | ~43 ms |

Verified on the repro ledger that all three subject classes retract correctly: persisted subjects (values replaced, no duplicates), novelty-only subjects, and new subjects (skipped, `skipped_subjects` debug field added for prod diagnosis).

## Tests

New `it_transact_upsert_indexed.rs` pins replace-semantics across persisted + novelty-only + new subjects in a single upsert against a file-backed indexed ledger, re-checked after the next index build. Existing transact/query/index/history integration groups and unit tests pass.

### DIFF
--- a/fluree-db-api/tests/it_transact_upsert_indexed.rs
+++ b/fluree-db-api/tests/it_transact_upsert_indexed.rs
@@ -39,6 +39,38 @@ async fn name_values(
     normalize_rows(&rows)
 }
 
+const GRAPH_ALPHA: &str = "http://example.org/graphs/alpha";
+const GRAPH_BETA: &str = "http://example.org/graphs/beta";
+const GRAPH_GAMMA: &str = "http://example.org/graphs/gamma";
+
+async fn graph_name_values(
+    fluree: &fluree_db_api::Fluree,
+    ledger: &fluree_db_api::LedgerState,
+    ledger_id: &str,
+    graph_iri: &str,
+    subject: &str,
+) -> Vec<String> {
+    let query = json!({
+        "@context": ctx(),
+        "from": format!("{ledger_id}#{graph_iri}"),
+        "where": {"@id": subject, "ex:name": "?n"},
+        "select": "?n"
+    });
+    let results = fluree
+        .query_connection(&query)
+        .await
+        .expect("named-graph query");
+    let results = results.to_jsonld(&ledger.snapshot).expect("to_jsonld");
+    let mut vals: Vec<String> = results
+        .as_array()
+        .expect("array")
+        .iter()
+        .filter_map(|v| v.as_str().map(str::to_string))
+        .collect();
+    vals.sort();
+    vals
+}
+
 /// One upsert touching a persisted subject, a novelty-only subject, and a
 /// brand-new subject must replace values for the first two and plainly insert
 /// the third — before and after the next index build.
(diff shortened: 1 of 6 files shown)


fluree-db-transact/src/stage.rs
```diff
@@ -2618,65 +2633,99 @@ async fn generate_upsert_deletions(
...
+                Some(tg) => txn
+                    .graph_delta
+                    .get(&tg)
+                    .and_then(|iri| ledger.snapshot.graph_registry.graph_id_for_iri(iri)),
+            };
+            (txn_g, ledger_g)
+        })
+        .collect();
+
+    // Novelty presence: one overlay walk per referenced graph, collecting which
```
aaj3f: *Optional (performance).* The novelty-presence walk is unconditional and eager, so an upsert whose subjects all resolve in the persisted dictionary now pays a full novelty walk it previously did not.

Mechanism: the walk is built before the subject loop for every upsert with a binary store, and the skip test at `:2743-2748` consults the novelty set *first* (`!novelty_present…is_some_and(…) && !subject_in_base(subject)`). So the O(novelty-in-graph) cost is paid even when the dictionary would have answered every subject on its own.

Concrete window: a single-subject upsert against an already-persisted subject on a ledger whose novelty has grown. Before this PR that was one ~13 µs point probe; after, it is that probe plus a walk of every novelty flake in the graph, each hashed into a `HashSet<&Sid>` probe. Under the shipped defaults this is free — `DEFAULT_REINDEX_MIN_BYTES = 100` (`server_defaults.rs:31`) means novelty is trimmed after essentially every commit — but the bound the comment appeals to is `reindex_max_bytes`, which is 20% of system RAM floored at 64 MB (`server_defaults.rs:48-62`). That is the ceiling that applies when indexing lags or is manual, and at that scale the walk is tens of milliseconds to seconds.

Suggested fix: make the set lazy rather than eager — check `subject_in_base` first (a point lookup) and build the per-graph novelty set only on the first subject that misses the dictionary. Bulk-new-subject upserts still do exactly one walk per graph, so the headline win is untouched, while all-persisted upserts do none. Worth also softening the comment at `:2662-2664`: "bounded by reindex backpressure" is accurate about the mechanism but reads as "small," and the actual bound is 20% of RAM.

fluree-db-api/tests/it_transact_upsert_indexed.rs
```diff
@@ -0,0 +1,162 @@
...
+    });
+    let rows = query_jsonld_formatted(fluree, ledger, &query)
+        .await
+        .expect("query");
+    normalize_rows(&rows)
+}
+
+/// One upsert touching a persisted subject, a novelty-only subject, and a
+/// brand-new subject must replace values for the first two and plainly insert
+/// the third — before and after the next index build.
```
aaj3f, line 76: *Optional (coverage).* The new test covers only the default graph, but named-graph handling is the part of `generate_upsert_deletions` the diff restructured most: the graph-id translation was hoisted out of the loop into `ledger_g_for_txn_g` (`stage.rs:2636-2656`), a new early `continue` was added for a named graph absent from the ledger registry (`:2737-2740`), and `effective_g_id` is now derived before the skip rather than after it. The skip keys the novelty set per ledger graph while `subject_in_base` is graph-agnostic.

The existing named-graph upsert test does not cover this: `it_named_graph_isolation.rs:370 pre_index_upsert_isolates_named_graphs` is explicitly a genesis-time test ("This runs at genesis (no binary index)", `:416`), so `binary_store.is_some()` is false and the new skip never engages. I could not find any post-index named-graph upsert test.

Suggested: add a second case to this file that indexes, then upserts the same subject IRI into two named graphs, asserting that each graph's values are replaced independently and neither leaks into the other. The graph-agnostic `subject_in_base` makes the skip conservative here rather than wrong, so I expect this to pass — which is exactly why it is cheap to pin.

fluree-db-transact/src/stage.rs
```diff
@@ -2579,32 +2579,47 @@ fn inline_values_to_pattern(values: &InlineValues) -> Result<Pattern> {
...
+/// upserts of new entities that turned staging into minutes of work producing
+/// zero retractions.
+///
+/// Named graph support: retractions are created in the same graph as the insert
+/// templates to ensure proper cancellation with assertions.
 async fn generate_upsert_deletions(
     ledger: &LedgerState,
     txn: &Txn,
     new_t: i64,
     graph_sids: &std::collections::HashMap<u16, Sid>,
```
aaj3f, line 2596: *Optional (perf evidence).* A hot-path change with no bench coverage. No bench in `fluree-db-api/benches` or `fluree-db-query/benches` exercises upsert at all — `transact_commit.rs` benches `insert_turtle_with_opts`, and `regression-budget.json` has no staging entry — so the nightly regression gate cannot catch a future regression in this function, and the PR's own 13.3 s → 0.9 ms number has no standing home. An `upsert`-shaped variant alongside `transact_commit` (new subjects and persisted subjects as separate inputs, since they exercise different paths) would give the win a durable guard. Commenting here because `fluree-db-api/benches/` has no file in this diff.

fluree-db-binary-index/src/read/binary_index_store.rs
```diff
@@ -1449,6 +1455,25 @@ impl BinaryIndexStore {
         self.dicts.predicates.resolve(p_id)
     }
 
+    /// Shared persisted `p_id → Sid` table, built once per store on first use.
```
aaj3f, line 1458: *Praise.* The `OnceLock` staleness argument in the doc comment is correct and I verified it rather than taking it: the store has exactly two `&mut self` methods (`set_ns_split_mode`, `augment_namespace_codes`) and both are reachable only from load-time sites holding the store by owned `&mut` before `Arc::new`; the post-commit `refresh_index` path deliberately does not re-sync (`ledger/loading.rs:31-38`). Stating the invariant in the comment where a future reader will need it is the right call for a cache whose failure mode would be silent.

fluree-db-transact/src/stage.rs
```diff
@@ -2618,65 +2633,99 @@ async fn generate_upsert_deletions(
...
-            fluree_db_query::execute_pattern(ledger.as_graph_db_ref(0), &query_vars, pattern)
-                .await?
-        };
-
-        // Convert each result to a retraction flake in the appropriate graph.
-        // Here we use the txn-local g_id to look up the graph Sid (flake.g).
+        // Retraction flakes carry the graph Sid (flake.g), looked up by the
+        // txn-local g_id. Resolved before any skip so broken graph delta/sid
+        // wiring still surfaces as an error.
         let graph_sid: Option<Sid> = match graph_id {
```
aaj3f, line 2712: *Praise.* Resolving `graph_sid` *before* the skip, with a comment saying why, keeps broken graph delta/sid wiring loud instead of letting a new fast path swallow it. That is the easy thing to get wrong when hoisting work out of a loop.

aaj3f (approved):
This is really nice to have! And nice to remove wasted work rather than needing a new fast path.

The remaining items are optional: the novelty walk is eager where it could be lazy, named-graph-plus-indexed has no test despite being the most-restructured branch, and no bench covers upsert staging at all.

- **Patterns/abstractions:** ✔ Generalizes an existing per-open computation to a per-store one and honors the engine's fail-open convention on lookup errors; no parallel mechanism introduced.
- **Performance (speed first, memory second):** ✔ with one noted window. Large net win on the target shape plus an O(predicate-count) fixed cost removed from every scan `open()`; the one regression window is a small upsert against large novelty, non-blocking under shipped defaults, cheap to close by making the walk lazy. No bench covers this path either way.
- **Testing:** ✔ New integration test is wired into `grp_transact` and runs, and it goes red under mutation of *both* halves of the pre-check — not merely present. Gap: no named-graph × indexed case.
- **Conventions:** ✔ Two focused `perf(...)`-style commits with substantive multi-line bodies carrying the measurements; no AI-attribution trailers; `cargo fmt --all -- --check` clean and `cargo clippy --all-targets` silent on all three changed crates.

Recommend merge. Note that PR #1553 is stacked on this branch, so this one must land first.

fluree-db-transact/src/stage.rs
bplatz: Addressed in bb2a042f8 — the skip now consults the base dictionary first and builds the per-graph novelty set only on the first subject that misses it, so all-persisted upserts pay no walk while bulk new-subject upserts keep exactly one walk per graph. Comment reworded to name reindex_max_bytes as the real ceiling.

fluree-db-api/tests/it_transact_upsert_indexed.rs
bplatz: Addressed in b4b4020cb — added upsert_indexed_named_graphs_replace_independently: post-index upsert of the same subject IRI into two named graphs (replaced independently, no leakage), plus a brand-new subject and a never-registered graph in the same transaction, asserted both from novelty and after the next index build.

fluree-db-transact/src/stage.rs
bplatz: Agreed but deferring to a follow-up PR — an upsert-shaped bench beside transact_commit with new-subject and persisted-subject variants plus a regression-budget entry is a bigger lift, and #1553 is stacked on this branch. The two variants will also guard both sides of the now-lazy pre-check.

bplatz:
@aaj3f Thanks for the careful review — all three optional items are handled:

- **Lazy novelty walk**: addressed in bb2a042f8. The skip now checks the base dictionary point-probe first and builds the per-graph novelty set only on the first subject that misses it, so all-persisted upserts pay no walk while bulk new-subject upserts keep exactly one walk per graph. The "bounded by reindex backpressure" comment is reworded to name `reindex_max_bytes` as the actual ceiling.
- **Named-graph × indexed coverage**: addressed in b4b4020cb. `upsert_indexed_named_graphs_replace_independently` upserts the same subject IRI into two named graphs post-index (replaced independently, no cross-graph leakage), alongside a brand-new subject and a never-registered graph in the same transaction — asserted from novelty and again after the next index build.
- **Upsert bench**: agreed, deferring to a follow-up PR (new-subject and persisted-subject variants beside `transact_commit`, plus a regression-budget entry) so stacked #1553 isn't held up. The two variants will guard both sides of the now-lazy pre-check.

PR #1550, opened 2026-07-29
bench: baseline host/corpus identity, phase-share gating, and noise floors

aaj3f:
# PR-bench: baseline extensions (host class, corpus identity, phase shares, noise floor)

> Builds on the `BaselineFile` introduced by **#1477**, which has now merged —
> this branch is rebased onto `06f1c10cf` (#1477's merge-reconcile head), which is
> in `main`'s history, so it merges cleanly with no rebase outstanding.
>
> The last commit is a **disclosed rider**, deliberately separable: a one-line
> `Cargo.lock` refresh (`fluree-bench-virtual` 4.1.3 → 4.1.4). That entry was
> stale on `main` itself — any plain `cargo build` on a main-derived branch
> rewrites it — and building this branch is what surfaced it. Drop or cherry-pick
> it independently; it carries no dependency-resolution change.

## Why

#1477 gave the workspace a real baseline capture/compare tool. This PR closes the
four gaps between that tool and what the `fluree rdf` conversion effort needs from
it (`riot-analog-bench-strategy.md` §6b, F5/F6), and corrects a doc that had
drifted from the workflows it describes.

The gaps share a theme: **a compare run that emits a number which looks like a
result and is not one.** Today the tool will happily compare a laptop against a CI
runner, a Turtle corpus against an N-Triples one, or a single flaky sample against
another, and print a tidy Δ% table either way.

### 1. Host block, and a refusal instead of a silent downgrade

`runner_class` becomes `host.class` inside a block that also records os, arch,
cpu model, and physical/logical cores (probed via `sysctl` on Darwin,
`/proc/cpuinfo` on Linux, falling back to `available_parallelism`).

A class names a **comparability set**, not a machine — two hosts share one when a
human has decided their absolute numbers may be compared. A mismatch is now a
**refusal (exit 2)**, not a quiet flip to advisory. `--allow-host-mismatch` opts
back into the advisory posture.

The derived default is `{os}-{arch}`, and the docs say plainly that it is coarse
and not a promise: an M1 and an M4 both derive `macos-aarch64`. Any host whose
numbers are meant to gate should set `FLUREE_BENCH_HOST_CLASS` explicitly.
`FLUREE_BENCH_RUNNER_CLASS` is still read as the pre-rename spelling, so no
existing workflow breaks.

### 2. Corpus identity — the expensive silent failure

A new `meta` sidecar (`target/fluree-bench-meta/`, same shape as #1477's `mem`
sidecar) carries per scenario: `sha256`, `byte_len`, `element_count`,
`input_syntax`, `output_syntax`, `thread_count`. Declarable per entry, or once at
file level and inherited.

A mismatch in hash, syntax, or thread count is a **refusal with no override**.
Unlike the host check there is no posture in which comparing different inputs is
meaningful, and the resulting table is indistinguishable from a real regression.
`name` and `element_count` are deliberately *not* part of the identity — two paths
to identical bytes are the same corpus.

**SHA-256, not BLAKE3.** blake3 is not in the workspace; `sha2` already is. A
corpus is hashed once per capture (streamed in 1 MiB chunks), so the throughput
difference does not pay for a new dependency. **No new workspace dependencies.**

> **Nothing produces phases or corpus blocks yet.** `meta.rs` is Tier-1
> scaffolding for the conversion lane: schema, recorder, and gate are in place,
> but no bench in the workspace calls `meta::record_scenario`, so no phase or
> corpus row appears in any report this PR produces. The share gate and the
> corpus refusal go live with the first producer. That is deliberate — the gate
> has to exist before the benches that depend on it, and it ships with full test
> coverage rather than waiting.

### 3. Per-phase timings and share drift

Per-entry `{phase: {ns, share_pct}}`, with the share derived from the ns at record
time (`ScenarioMeta::with_phase_ns`) so the two can never disagree.

- **Absolute `phase_time`** gates only within a host class, like `time` and
  `peak_mem`.
- **`phase_share` gates regardless of host class.** A share is a ratio within one
  run, so the machine cancels. This is the only signal a cross-host comparison
  keeps, and it is the one that catches a regression an aggregate number hides.

Only share *growth* gates. Shares sum to 100, so gating both directions would fail
a clean improvement twice — once for the phase that got better, once for its
complement. Shrink is reported as `improved` so the whole picture stays visible.

**One-sided phases are gated on share alone.** Comparing only same-named phases
is itself a hiding place: move work into a phase the baseline never had and the
old phase's share falls — reading as an improvement — while the new phase
carrying the cost produces no row. So a phase new in the current run breaches
when it claims more than the drift threshold, and a phase that has vanished
breaches when it *used to*. Rows are marked `(NEW)` / `(VANISHED)`.

Here is the case that motivates the whole feature, from the end-to-end smoke run —
whole-scenario time is inside its noise floor and clean, while the phase breakdown
localizes a 28% parse regression masked by a write improvement:

```
scenario [metric]                                    status     baseline     observed        Δ   budget
convert/ttl_to_nq/small [time]                           ok     1.000 ms     1.025 ms   +2.50%     3.0%
convert/ttl_to_nq/small::parse [phase_time]        EXCEEDED   600.000 µs   768.750 µs  +28.12%     3.0%
convert/ttl_to_nq/small::parse [phase_share]       EXCEEDED        60.0%        75.0%  +15.0pp    5.0pp
convert/ttl_to_nq/small::read  [phase_time]              ok   100.000 µs   102.500 µs   +2.50%     3.0%
convert/ttl_to_nq/small::read  [phase_share]             ok        10.0%        10.0%   +0.0pp    5.0pp
convert/ttl_to_nq/small::write [phase_time]        improved   300.000 µs   153.750 µs  -48.75%     3.0%
convert/ttl_to_nq/small::write [phase_share]       improved        30.0%        15.0%  -15.0pp    5.0pp
```

### 4. Noise floor

`capture --accumulate` folds repeat runs into one file: each pass appends the fresh
criterion mean to a per-scenario sample vector and recomputes median + MAD. At
n ≥ 5 the gate compares **medians** instead of single-run means, and widens the
wall-clock budget to `max(budget_pct, 3 × MAD / median)`.

Criterion's own confidence interval describes iterations *within* one run and says
nothing about run-to-run spread — which is exactly what shared-runner flap is. MAD
rather than standard deviation because one outlier run is the common failure here
and barely moves MAD.

Three bounds, each of which inverts the gate if dropped: the floor **only widens**
(a tighter-than-noise budget can only false-alarm); it is **wall-clock only**,
never touching `peak_mem` (an allocator peak is near-deterministic, so a noisy
clock is no reason to forgive an allocation regression); and it is **baseline
only** (otherwise the worse a run behaves, the more it is forgiven).

`accumulate` itself refuses to pool samples across host classes or corpora, and
both guards live in the library rather than the CLI. Accumulation is the one
operation that *destroys* the evidence of a mismatch — two populations blended
into one median/MAD, with the provenance overwritten in the same breath — so a
direct caller must not be able to skip the check.

### 5. BENCHMARKING.md corrections

Every claim checked against the merged main + #1477 tree. Five were wrong:

| Claim | Reality |
|---|---|
| `.github/workflows/bench-nightly.yml` (in "Where benches live") | Never existed; the file is `bench.yml` |
| "lands in bench-5" / "lands in bench-nightly PR" | Both landed |
| Bench table lists 12 benches | 18 `[[bench]]` entries are registered — six missing |
| `BenchProfile::Full` = "Nightly CI" (`runtime.rs`) | Every CI job pins `quick`/`tiny`, nightly included |
| `bench-baselines/README.md`: "enforces memory across any runner class, time advisory" | False since 5d9d69513 made the gate symmetric — stale on #1477's own branch |

Note on the "runs on every PR" and "dedicated 4-core runners" claims from the
original brief: **#1477 already removed both** in its rewrite of the CI-gate
section. They are gone from the merged result; nothing to do.

The six missing bench rows are `query_overlay_matrix`, `query_hot_bsbm_bi`,
`query_hot_property_path`, `query_hot_whole_graph_agg`, `annotation_hydration`,
`annotation_planner` (five are missing on main; #1477 adds the first). Each row is
written from the bench's own module docs. The table now also states outright that
nothing enforces it — unlike the `[[bench]]` ↔ budget reconcile — which is why it
rots.

## Sample baseline (schema 2)

```json
{
  "schema_version": 2,
  "label": "convert-pre",
  "git_sha": "c7d82614d",
  "captured_at": "2026-07-28T23:54:15Z",
  "profile": "quick",
  "scale": "small",
  "host": {
    "class": "bench-m8gd",
    "name": "ip-10-0-3-7",
    "os": "linux",
    "arch": "x86_64",
    "cpu_model": "AMD EPYC 9R14",
    "physical_cores": 16,
    "logical_cores": 32
  },
  "entries": {
    "convert/ttl_to_nq/small": {
      "mean_ns": 990000.0,
      "median_ns": 990000.0,
      "mem": { "peak_bytes": 24117248, "total_allocated_bytes": 310378496 },
      "noise": {
        "n": 5,
        "median_ns": 1000000.0,
        "mad_ns": 10000.0,
        "samples_ns": [980000.0, 990000.0, 1000000.0, 1010000.0, 1020000.0]
      },
      "corpus": {
        "name": "dblp.ttl",
        "sha256": "aaaa1111bbbb2222cccc3333dddd4444eeee5555ffff6666aaaa7777bbbb8888",
        "byte_len": 1048576,
        "element_count": 40000,
        "input_syntax": "turtle",
        "output_syntax": "nquads",
        "thread_count": 8
      },
      "phases": {
        "read":  { "ns": 100000.0, "share_pct": 10.0 },
        "parse": { "ns": 600000.0, "share_pct": 60.0 },
        "write": { "ns": 300000.0, "share_pct": 30.0 }
      }
    }
  }
}
```

Every added block is optional. A schema-1 file still loads: `runner_class` maps to
`host.class`, a bare `host` string maps to `host.name`, and missing
`noise`/`corpus`/`phases` simply switch off their respective rules. The committed
`bench-baselines/guardrails-pre.json` is **unmodified** and still compares. A
baseline claiming a *newer* schema version is a refusal — silently ignoring fields
we don't understand is how a gate stops gating.

## Gating semantics, complete

| metric | units | noise floor applies | enforced within host class | enforced across host classes |
|---|---|---|---|---|
| `time` | ns | yes | yes | no (advisory) |
| `peak_mem` | bytes | **no** | yes | no (advisory) |
| `phase_time` | ns | yes | yes | no (advisory) |
| `phase_share` | percentage points | no | yes | **yes** |

The noise floor is wall-clock-only and baseline-only. It never touches
`peak_mem` — an allocator peak is near-deterministic, so a noisy clock is no
reason to forgive an allocation regression — and the current run's own spread
never widens its own budget, or the worse a run behaves the more it is forgiven.

Phase rows also carry a presence marker. A phase present on both sides compares
normally; one new in the current run breaches when its share exceeds the drift
threshold (`(NEW)`); one that has vanished breaches when it *used to*
(`(VANISHED)`). Neither one-sided case emits a `phase_time` row.

| condition | outcome | exit |
|---|---|---|
| within budget | pass | 0 |
| enforced breach | fail | 1 |
| host class differs, no `--allow-host-mismatch` | refuse | 2 |
| corpus sha256 / syntax / thread count differs | refuse | 2 |
| baseline `schema_version` newer than the binary | refuse | 2 |

## CI impact: none, by construction

The per-PR `bench-compare` and nightly `bench-gate` compare a `host_class=local`
baseline on `ci-ubuntu-latest`, which the new default would refuse. Both steps now
pass `--allow-host-mismatch` explicitly, reproducing #1477's exact behaviour
(annotations, exit 0). The change is that the downgrade is written down, with a
comment naming the condition for removing the flag.

`bench-capture` gains a `capture_samples` dispatch input (default `1`, unchanged
behaviour). At `≥ 5` it reruns the subset and accumulates, producing the median +
MAD a baseline needs before it can gate through runner flap.

## Files

| File | What |
|---|---|
| `fluree-bench-support/src/meta.rs` | **new** — `CorpusInfo`, `PhaseTiming`, `ScenarioMeta`, sidecar IO, SHA-256 |
| `fluree-bench-support/src/baseline.rs` | `HostBlock`, `NoiseStats`, `CompareOptions`, `FatalMismatch`, `Metric`, refusal + share + noise gating, schema-1 read path |
| `fluree-bench-support/src/bin/bench_baseline.rs` | `--host-class`, `--allow-host-mismatch`, `--share-drift-pp`, `--accumulate`, `--clean-sidecars`; exit code 2; per-metric rendering |
| `fluree-bench-support/src/runtime.rs` | corrected profile docs (`Full` is not the CI profile) |
| `fluree-bench-support/src/lib.rs`, `Cargo.toml` | `meta` module; `sha2` (existing workspace dep) |
| `.github/workflows/{ci,bench}.yml` | explicit `--allow-host-mismatch`; `capture_samples`; clear the meta sidecar dir |
| `BENCHMARKING.md`, `bench-baselines/README.md`, `fluree-bench-support/README.md` | corrections + new semantics |

## Verification

```
cargo nextest run -p fluree-bench-support                    122 passed, 0 skipped
cargo test -p fluree-bench-support --test workspace_reconcile  1 passed
cargo clippy -p fluree-bench-support --all-targets --no-deps   clean, 0 warnings
cargo fmt --check                                              clean
```

**57 of the 121 lib tests are new** (measured against this branch's base at
`06f1c10cf`, which has 64). Every gating rule has one: host mismatch fatal /
allowed / matched / forced-advisory, corpus mismatch by hash, by syntax, by thread
count, file-level vs entry-level precedence, one-sided declaration, share growth /
shrink / within-threshold / configurable-threshold / cross-host enforcement, MAD
arithmetic, floor widening, floor not tightening, floor not hiding a real move,
sub-threshold sample counts, accumulate arithmetic, and the schema-1 read path
including the exact shape of the committed `guardrails-pre.json`.

Also exercised end-to-end against the real binary: the committed schema-1 baseline
(refusal, then advisory under the flag), a captured schema-2 baseline accumulated
to n=5, a share-drift breach under a passing aggregate, a corpus swap, and
cross-host share enforcement.

## Deviations from the §6b spec

1. **SHA-256 instead of BLAKE3** — blake3 is not in the tree; `sha2` is. The spec
   permits this ("prefer no new deps").
2. **Share drift gates growth only, not absolute movement.** Gating `|Δ|` would
   double-flag every real improvement, since shares sum to 100. Both directions are
   still reported.
3. **`--allow-host-mismatch` exists at all.** The spec says compare hard-errors on
   host mismatch; making that unconditional would red every PR today, since the
   only committed baseline is cross-class. The refusal is the default and the
   escape hatch is explicit, per-invocation, and documented — strictly louder than
   #1477's silent downgrade.
4. **`thread_count` is part of corpus identity**, though the spec listed it as a
   recorded field. A K=8 run compared against a K=16 baseline is the same class of
   silent failure as a corpus swap. Its refusal message is distinct so a reader can
   tell a config mismatch from a data mismatch.

## Open questions

1. **Should `guardrails-pre.json` be re-stamped to a specific host class?** It was
   captured on Apple silicon but records `host_class=local`. Re-stamping it
   `macos-aarch64` would make a Mac dev's local compare *enforce* — but would also
   assert that any two Macs are comparable, which is false. Left as `local`
   deliberately; it is the honest value.
2. **What `host_class` string does the official publication locus use?** H-10
   approved `m8gd.4xlarge`; the strategy's F7 wants publication tooling to refuse
   non-official classes. This PR provides the field; the allowlist is not
   implemented and needs the canonical string.
3. **Where should milestone convert baselines live** — `bench-baselines/convert/`
   per §6b, with rolling per-change baselines gitignored? Not created here (no
   convert benches exist yet).

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -12,6 +12,16 @@ on:
   schedule:
     - cron: "0 7 * * *" # 07:00 UTC daily
   workflow_dispatch:
+    inputs:
+      capture_samples:
+        description: >-
+          Repeat runs to fold into the captured baseline. 1 is a single
+          measurement with no noise estimate; >= 5 records a median + MAD, which
+          is what a baseline intended to GATE on shared runners needs — the
+          compare gate widens each budget to 3*MAD/median, and ubuntu-latest
+          flap is the whole reason the per-PR compare is advisory today.
+        required: false
+        default: "1"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -39,9 +49,10 @@ jobs:
     env:
       FLUREE_BENCH_PROFILE: quick
       FLUREE_BENCH_SCALE: tiny
-      # Baseline is runner_class=local, so the nightly compare enforces memory and
-      # treats time as advisory until a ci-* baseline is committed (bench-capture).
-      FLUREE_BENCH_RUNNER_CLASS: ci-ubuntu-latest
+      # Baseline is host_class=local, so the nightly compare treats time and
+      # memory as advisory until a ci-* baseline is committed (bench-capture).
+      # (FLUREE_BENCH_RUNNER_CLASS is still read as the pre-rename spelling.)
+      FLUREE_BENCH_HOST_CLASS: ci-ubuntu-latest
     steps:
       - name: Free disk space
         run: |
@@ -85,11 +96,12 @@ jobs:
       - name: Compare captured subset against committed baseline
         if: always()
         run: |
-          rm -rf target/fluree-bench-mem
+          rm -rf target/fluree-bench-mem target/fluree-bench-meta
           cargo bench -p fluree-db-api --bench query_overlay_matrix
           cargo bench -p fluree-db-api --bench query_hot_bsbm
           cargo run -p fluree-bench-support --bin bench-baseline -- \
-            compare --baseline bench-baselines/guardrails-pre.json
+            compare --baseline bench-baselines/guardrails-pre.json \
+            --allow-host-mismatch
 
   bench-capture:
     # On-demand CI-class baseline capture (review item 1c). Captures the cheap
@@ -104,7 +116,7 @@ jobs:
     env:
       FLUREE_BENCH_PROFILE: quick
       FLUREE_BENCH_SCALE: tiny
-      FLUREE_BENCH_RUNNER_CLASS: ci-ubuntu-latest
+      FLUREE_BENCH_HOST_CLASS: ci-ubuntu-latest
     steps:
       - name: Free disk space
         run: |
@@ -123,15 +135,27 @@ jobs:
 
       # Same cheap subset and scales (tiny+small) as the committed baseline, so
       # the artifact is a drop-in ci-* replacement once a human commits it.
+      #
+      # Each pass reruns the benches and folds the fresh criterion means into the
+      # same file with --accumulate, which is what turns N runs into one median +
+      # MAD. Passes must NOT be parallelised and the sidecars are cleared each
+      # time: a sample has to be an independent measurement of the same thing.
       - name: Capture cheap subset (CI-class)
         run: |
-          rm -rf target/fluree-bench-mem
-          for scale in tiny small; do
-            FLUREE_BENCH_SCALE=$scale cargo bench -p fluree-db-api --bench query_overlay_matrix
-            FLUREE_BENCH_SCALE=$scale cargo bench -p fluree-db-api --bench query_hot_bsbm
+          samples="${{ github.event.inputs.capture_samples || '1' }}"
+          rm -f bench-baselines/guardrails-pre-ci.json
+          for pass in $(seq 1 "$samples"); do
+            echo "::group::capture pass $pass/$samples"
+            rm -rf target/fluree-bench-mem target/fluree-bench-meta
+            for scale in tiny small; do
+              FLUREE_BENCH_SCALE=$scale cargo bench -p fluree-db-api --bench query_overlay_matrix
+              FLUREE_BENCH_SCALE=$scale cargo bench -p fluree-db-api --bench query_hot_bsbm
+            done
+            cargo run -p fluree-bench-support --bin bench-baseline -- \
+              capture --label guardrails-pre-ci --out bench-baselines/guardrails-pre-ci.json \
+              --accumulate
+            echo "::endgroup::"
           done
-          cargo run -p fluree-bench-support --bin bench-baseline -- \
-            capture --label guardrails-pre-ci --out bench-baselines/guardrails-pre-ci.json
 
       - name: Upload CI-class baseline artifact
         uses: actions/upload-artifact@v4
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -143,6 +143,14 @@ jobs:
       - name: Capture cheap subset (CI-class)
         run: |
           samples="${{ github.event.inputs.capture_samples || '1' }}"
+          # `seq 1 0` runs zero passes, so a 0 (or any non-number) would produce
+          # no baseline at all while the job still reported success. Fail loudly
+          # instead — a capture job that captures nothing is worse than one that
+          # errors, because the artifact step would then upload nothing quietly.
+          if ! [[ "$samples" =~ ^[1-9][0-9]*$ ]]; then
+            echo "::error::capture_samples must be a positive integer, got '$samples'"
+            exit 1
+          fi
           rm -f bench-baselines/guardrails-pre-ci.json
           for pass in $(seq 1 "$samples"); do
             echo "::group::capture pass $pass/$samples"
@@ -162,3 +170,6 @@ jobs:
         with:
           name: guardrails-pre-ci-baseline
           path: bench-baselines/guardrails-pre-ci.json
+          # Default is `warn`, which lets the job go green having produced no
+          # artifact — the exact silent success this job exists to avoid.
+          if-no-files-found: error
--- a/BENCHMARKING.md
+++ b/BENCHMARKING.md
@@ -187,9 +187,12 @@ captured at the branch's **merge-base** (see `bench-baselines/README.md`).
 **Compare** — after a change, rerun the benches and compare:
 
 ```bash
+# --allow-host-mismatch is required against the committed baseline, which is
+# host_class=local; without it compare refuses (exit 2) rather than compare
+# absolute numbers across machines. Drop it once you have a same-class baseline.
 cargo run -p fluree-bench-support --bin bench-baseline -- \
-    compare --baseline bench-baselines/guardrails-pre.json \
-    [--only <substr>] [--share-drift-pp 5] [--allow-host-mismatch] [--advisory]
+    compare --baseline bench-baselines/guardrails-pre.json --allow-host-mismatch \
+    [--only <substr>] [--share-drift-pp 5] [--advisory]
 ```
 
 Each scenario present in both runs is checked against `baseline × (1 +
@@ -208,12 +211,30 @@ share.
 — exit 2, no verdict — unless `--allow-host-mismatch` downgrades them to
 `::warning::` annotations. `--advisory` forces the downgrade even on a match.
 
+> **No bench emits phases yet.** The `meta` sidecar and everything below about
+> phases is Tier-1 scaffolding for the `fluree rdf` conversion lane: the schema,
+> the recorder, and the gate are in place, but no bench in the workspace calls
+> `meta::record_scenario` today, so no phase or corpus row appears in any current
+> report. The share gate goes live with the first producer. Corpus identity is in
+> the same state — the refusal rules are real, they simply have nothing to check
+> until a bench declares what it read.
+
 **Share drift always enforces**, because a share is a ratio within a single run:
 the machine cancels. Only share *growth* gates; shares sum to 100, so gating both
 directions would fail a clean improvement twice, once for the phase that got
 better and once for its complement. This is the metric that catches the
 regression an aggregate number hides — a scenario 2% slower overall whose parse
 share moved 60% → 75% has a parse regression masked by a write improvement.
+
+**A phase that appears or disappears is gated on its share alone.** Comparing
+only same-named phases is how a regression hides: move work into a phase the
+baseline never had and the old phase's share *falls*, reading as an improvement,
+while the new phase carrying the cost produces no row. So a phase new in the
+current run breaches when it claims more than the drift threshold, and a phase
+that has vanished from the current run breaches when it *used to* claim more than
+the threshold — work that large did not evaporate, it moved somewhere unmeasured.
+Such rows are marked `(NEW)` / `(VANISHED)` and carry no `phase_time` row, since
+there is no ratio to report against a missing side.
 
 **A corpus mismatch is always a refusal, with no override.** Benches that read an
 input record its SHA-256, byte length, element count, input/output syntax, and
@@ -244,10 +265,25 @@ done
 ```
 
 With ≥ 5 samples the baseline records a median and a MAD, the gate compares
-medians instead of single-run means, and each budget widens to
-`max(budget_pct, 3 × MAD / median)`. The floor only ever *widens* an envelope: a
-quiet scenario keeps its declared budget. MAD rather than standard deviation
-because one outlier run is the common failure here and barely moves it.
+medians instead of single-run means, and the wall-clock budget widens to
+`max(budget_pct, 3 × MAD / median)`. MAD rather than standard deviation because
+one outlier run is the common failure here and barely moves it.
+
+Three bounds on that, each of which exists because dropping it inverts the gate:
+
+- **Widening only.** A budget tighter than the measured noise can only produce
+  false alarms; a quiet scenario keeps its declared budget.
+- **Wall-clock only.** The floor applies to `time` and `phase_time`, never to
+  `peak_mem`. A tracking allocator's peak is near-deterministic for a given
+  workload, so widening the memory budget by a *timing* MAD would let a real
+  allocation regression through on the strength of a noisy clock.
+- **Baseline only.** The current run's own spread never widens its own budget.
+  Otherwise the worse a run behaves, the more it is forgiven.
+
+`capture --accumulate` refuses to pool samples across host classes or across
+corpora. Blending two populations into one median and MAD destroys the evidence
+that they were ever different, and the accumulated file inherits the fresh run's
+provenance — so there would be nothing left for a later `compare` to catch.
 
 ### Why the gate has two phases
 
--- a/BENCHMARKING.md
+++ b/BENCHMARKING.md
@@ -317,11 +317,24 @@ whose numbers are meant to gate should set `FLUREE_BENCH_HOST_CLASS` explicitly
 
 Two ways to reach phase 2:
 
-1. **Committed CI-class baseline (implemented).** Run `bench.yml`'s
-   `workflow_dispatch` `bench-capture` job with `capture_samples: 5`, download
-   the artifact, commit it, and remove `--allow-host-mismatch` from `ci.yml`'s
-   compare step. Its `host_class=ci-ubuntu-latest` matches CI and its noise floor
-   absorbs runner flap, so both metrics enforce.
+1. **Committed CI-class baseline (implemented).** Note the **filename change** —
+   the capture job writes `guardrails-pre-ci.json`, but both compare jobs read
+   `guardrails-pre.json`. Committing the artifact under its own name leaves CI
+   reading the old `host_class=local` file, which without
+   `--allow-host-mismatch` refuses every run. So:
+
+   ```bash
+   # 1. Run bench.yml's workflow_dispatch `bench-capture` job with
+   #    capture_samples: 5, and download the artifact.
+   # 2. REPLACE the committed baseline — do not add it alongside:
+   mv guardrails-pre-ci.json bench-baselines/guardrails-pre.json
+   # 3. Drop `--allow-host-mismatch` from ci.yml's and bench.yml's compare steps.
+   ```
+
+   The replaced file's `host_class=ci-ubuntu-latest` then matches CI and its
+   noise floor absorbs runner flap, so both metrics enforce. (Keeping the
+   `-ci` suffix instead would work only if you also repoint every
+   `--baseline` path; one rename is the smaller change.)
 2. **Same-runner interleaved base-vs-HEAD (documented future option).** Measure
    merge-base and HEAD in the *same* job on the *same* runner and compare those,
    instead of any committed cross-machine file — the only design that survives
--- a/BENCHMARKING.md
+++ b/BENCHMARKING.md
@@ -40,13 +40,24 @@ FLUREE_BENCH_TRACING=1 cargo bench -p fluree-db-api --bench insert_formats
 |---|---|---|---|
 | `FLUREE_BENCH_PROFILE` | `quick` \| `full` | `quick` | sample-count + warmup discipline |
 | `FLUREE_BENCH_SCALE` | `tiny` \| `small` \| `medium` \| `large` | `small` | per-bench input size |
+| `FLUREE_BENCH_HOST_CLASS` | any name | `{os}-{arch}` | comparability class for baseline capture/compare |
+| `FLUREE_BENCH_HOST` | any name | `$HOSTNAME` | host name recorded as provenance |
 | `FLUREE_BENCH_TRACING` | `1` (or unset) | unset | install a stderr tracing subscriber |
 | `FLUREE_BENCH_RUNTIME` | `multi` (or unset) | single-threaded | tokio runtime shape |
 | `RUST_LOG` | tracing-subscriber filter | `info` when `FLUREE_BENCH_TRACING=1` | tracing levels per crate |
 
+Those defaults are what a bare `cargo bench` on a laptop uses. **Both CI jobs
+pin `FLUREE_BENCH_PROFILE=quick` and `FLUREE_BENCH_SCALE=tiny`** — including the
+nightly, which is quick-profile despite the name — so a nightly number is not a
+`full`-profile number. `FLUREE_BENCH_RUNNER_CLASS` is still read as the
+pre-rename spelling of `FLUREE_BENCH_HOST_CLASS`.
+
 ## Current benches
 
-Hand-maintained; add new entries when introducing a bench file.
+Hand-maintained, and reconciled against the `[[bench]]` declarations plus
+`regression-budget.json` — `cargo test -p fluree-bench-support --test
+workspace_reconcile` enforces those two against each other, but nothing enforces
+this table, so it is the one that rots. Add a row when you add a bench file.
 
 | Crate | Bench file | Topic |
 |---|---|---|
@@ -60,6 +71,12 @@ Hand-maintained; add new entries when introducing a bench file.
 | `fluree-db-api` | `reindex_incremental.rs` | Orchestrator's incremental path via `Fluree::trigger_index(...)` over delta novelty |
 | `fluree-db-api` | `novelty_replay.rs` | Cold reload with `without_indexing()` so populate stays in novelty; scaled by commit count |
 | `fluree-db-api` | `query_hot_bsbm.rs` | Warm-cache SPARQL: BSBM-shape Q3/Q5/Q9 against a reindexed ledger |
+| `fluree-db-api` | `query_hot_bsbm_bi.rs` | BSBM Business-Intelligence "bowtie" (BI-1 F2): join-ordering tie-break between two equally-selective anchors |
+| `fluree-db-api` | `query_hot_property_path.rs` | Hot-cache SPARQL property paths, one scenario per execution mode of the operator (`*` closure, sequence, `?`, …) |
+| `fluree-db-api` | `query_hot_whole_graph_agg.rs` | Cypher aggregate folds from `fast_whole_graph_agg` (whole-graph + class scalars, histograms) against a linear-cost pipeline baseline |
+| `fluree-db-api` | `query_overlay_matrix.rs` | The same query shapes at three ledger conditions — base / overlay / novelty — so the columnar+novelty merge lane has coverage |
+| `fluree-db-api` | `annotation_hydration.rs` | `inject_annotations` hydration cost: index scan vs sealed annotation arena |
+| `fluree-db-api` | `annotation_planner.rs` | Planner direction for `f:reifies*` edge-annotation queries: arena-informed row counts vs HLL-only stats |
 | `fluree-db-query` | `vector_math.rs` | SIMD vs scalar dot/L2/cosine micro-bench |
 | `fluree-db-spatial` | `spatial_bench.rs` | S2 covering build + within/intersects/radius latency |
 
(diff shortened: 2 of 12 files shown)


bplatz (commented):
@aaj3f I did a defect-focused pass over the gate and workflow changes. I found two fail-open paths that look merge-blocking (missing corpus metadata and invalid phase shares), plus several accumulation/provenance and input-hardening issues. Details are inline; I left this as a comment review rather than changing the approval state.

fluree-bench-support/src/baseline.rs
```diff
@@ -349,47 +1168,118 @@ pub fn compare(
...
             continue;
         }
-        let budget_pct = budget_for_id(budget, id);
 
-        if base.mean_ns > 0.0 {
-            let (change_pct, status) = classify(base.mean_ns, cur.mean_ns, budget_pct);
+        // Corpus identity gates everything else about this scenario: if the two
+        // runs read different inputs, none of their numbers are comparable and
+        // emitting rows for them would be worse than emitting nothing.
+        if let (Some(base_corpus), Some(cur_corpus)) =
```
bplatz: This treats `(Some(base_corpus), None)` as comparable. If a producer stops emitting its sidecar, or the sidecar write/read fails, the current run bypasses the corpus-identity gate and can exit 0. Once the baseline declares a corpus, missing current metadata should be a refusal. The same asymmetry exists in `accumulate` around line 626, where `fresh.clone()` can also erase the prior identity while pooling the sample. Could you add compare and accumulate regression tests for the `Some`/`None` case?

fluree-bench-support/src/baseline.rs
```diff
@@ -310,32 +936,225 @@ fn classify(baseline: f64, observed: f64, budget_pct: f64) -> (f64, CompareStatu
...
+    report: &mut CompareReport,
+    id: &str,
+    side: &str,
+    phases: &BTreeMap<String, PhaseTiming>,
+) {
+    if phases.is_empty() {
+        return;
+    }
+    let sum: f64 = phases.values().map(|p| p.share_pct).sum();
+    if (sum - 100.0).abs() > SHARE_SUM_TOLERANCE_PP {
```
bplatz: I don't think invalid shares are cosmetic here. On a host mismatch, `phase_share` is the only enforcing signal; a malformed current sidecar with all shares at 0 makes every phase look improved while this only emits an anomaly and the command exits 0. Since the file is explicitly treated as untrusted input, please refuse non-finite/negative values and a materially invalid sum (or recompute shares from `ns`) before gating.

fluree-bench-support/src/baseline.rs
```diff
@@ -181,12 +575,96 @@ pub fn capture(
...
+/// accumulated file takes `fresh`'s label, sha, and host.
+///
+/// So both guards live here, in the library, rather than in the CLI that
+/// happens to call it — a bench harness or script calling `accumulate` directly
+/// gets the same protection:
+///
+/// - **host class** must match between `prev` and `fresh`.
+/// - **corpus identity** must match for every scenario present in both,
+///   resolved entry-then-file exactly as [`compare`] resolves it.
+pub fn accumulate(prev: &BaselineFile, fresh: &BaselineFile) -> Result<BaselineFile, String> {
```
bplatz: `compare` refuses a schema newer than this binary understands, but `accumulate` has no equivalent guard. An older binary can deserialize an additive schema-3 file, clone a fresh schema-2 capture, and save it back while silently dropping fields it did not understand. Please reject `prev.schema_version > BASELINE_SCHEMA_VERSION` before merging.

fluree-bench-support/src/baseline.rs
```diff
@@ -181,12 +575,96 @@ pub fn capture(
...
+            if let Some(detail) = prev_corpus.identity_mismatch(fresh_corpus) {
+                return Err(format!(
+                    "refusing to accumulate across corpora for {id}: {detail}. Samples over \
+                     different inputs cannot be pooled into one median/MAD."
+                ));
+            }
+        }
+    }
+
+    let mut out = fresh.clone();
```
bplatz, line 786: This replaces `git_sha` with the fresh SHA while pooling the previous samples without checking revision identity. Running `--accumulate` after a code change mixes pre/post-change populations, relabels all samples as the new revision, and can widen MAD enough to hide the change being measured. I think differing non-empty SHAs should refuse accumulation unless there is an explicit override.

fluree-bench-support/src/baseline.rs
```diff
@@ -22,17 +24,297 @@
...
+        deviations.sort_by(|a, b| a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal));
+        Self {
+            n: samples.len() as u32,
+            median_ns,
+            mad_ns: median_of_sorted(&deviations),
+            samples_ns: samples,
+        }
+    }
+
+    fn is_usable(&self) -> bool {
```
bplatz, line 333: `n`, `median_ns`, and `mad_ns` are trusted independently of `samples_ns`. A stale or hand-edited baseline can claim `n >= 5` with an arbitrary median/MAD and thereby move the comparison baseline or widen the budget. Since the raw samples are stored, load/compare should recompute these fields or reject an inconsistent/non-positive set.

fluree-bench-support/src/baseline.rs
```diff
@@ -62,42 +391,102 @@ pub struct BaselineFile {
...
+    scale: String,
+    /// Schema-1 spelling of `host.class`.
+    #[serde(default)]
+    runner_class: Option<String>,
+    #[serde(default)]
+    host: Option<HostField>,
+    #[serde(default)]
+    corpus: Option<CorpusInfo>,
+    #[serde(default)]
+    entries: BTreeMap<String, BaselineEntry>,
```
bplatz, line 474: Because all top-level fields default, `{}` deserializes as a valid schema-1 baseline. With `--allow-host-mismatch`, every current scenario is merely reported as new, no metric breaches, and compare exits 0. Please validate that a loaded baseline has the required structure and at least one scenario rather than accepting an empty document.

fluree-bench-support/src/meta.rs
```diff
@@ -0,0 +1,502 @@
...
+        use std::fmt::Write;
+        let _ = write!(out, "{b:02x}");
+    }
+    out
+}
+
+/// First 12 hex chars, for report lines. Never used for comparison.
+pub(crate) fn short_hash(hash: &str) -> &str {
+    let end = hash.len().min(12);
+    &hash[..end]
```
bplatz: This byte slice can panic when an untrusted `sha256` contains Unicode and byte 12 is not a UTF-8 boundary, turning a documented refusal into exit 101. Validating the field as 64 lowercase ASCII hex characters would also protect identity semantics; otherwise truncate by characters here.

.github/workflows/bench.yml
```diff
@@ -123,15 +135,27 @@ jobs:
...
+      # same file with --accumulate, which is what turns N runs into one median +
+      # MAD. Passes must NOT be parallelised and the sidecars are cleared each
+      # time: a sample has to be an independent measurement of the same thing.
       - name: Capture cheap subset (CI-class)
         run: |
-          rm -rf target/fluree-bench-mem
-          for scale in tiny small; do
-            FLUREE_BENCH_SCALE=$scale cargo bench -p fluree-db-api --bench query_overlay_matrix
-            FLUREE_BENCH_SCALE=$scale cargo bench -p fluree-db-api --bench query_hot_bsbm
+          samples="${{ github.event.inputs.capture_samples || '1' }}"
```
bplatz, line 145: `capture_samples=0` makes `seq 1 0` run no passes. The baseline is never created, and `upload-artifact` only warns on a missing file by default, so the capture job can succeed without an artifact. Please validate a positive integer and set `if-no-files-found: error` on the upload step.

BENCHMARKING.md
```diff
@@ -197,21 +299,29 @@ portable of the two signals in the blocking position.
...
+A note on `host_class` values. The derived default is `{os}-{arch}`, which is
+deliberately coarse and deliberately not a promise: an M1 and an M4 both derive
+`macos-aarch64` and their absolute numbers are not interchangeable. A class is a
+claim *a human makes* that two machines' numbers may be compared, so any host
+whose numbers are meant to gate should set `FLUREE_BENCH_HOST_CLASS` explicitly
+(`ci-ubuntu-latest`, `bench-m8gd`, …) rather than inherit the default.
 
 Two ways to reach phase 2:
 
 1. **Committed CI-class baseline (implemented).** Run `bench.yml`'s
```
bplatz: The capture job writes `guardrails-pre-ci.json`, while both CI compare jobs read `guardrails-pre.json`. Following 'download the artifact, commit it, and remove `--allow-host-mismatch`' literally leaves CI reading the old local baseline and refusing every run. Please explicitly say to replace/rename `guardrails-pre.json`, or update the compare path as part of phase 2.

fluree-bench-support/src/baseline.rs
aaj3f: Fixed in `ecba51446`. You're right that absence of evidence was being read as evidence of the same input. `(Some, None)` is now a refusal; the reverse (`None, Some`) stays permitted, since a schema-1 baseline predates corpus blocks and refusing there would make every old baseline unusable — that asymmetry is now explicit in a `corpus_refusal` helper that both `compare` and `accumulate` call, so the two can't drift apart again. `accumulate` had the same hole plus the `fresh.clone()` erasure you spotted. Regression tests both places: `baseline_corpus_with_no_current_corpus_refuses` and `accumulate_refuses_when_fresh_drops_the_corpus`.

fluree-bench-support/src/baseline.rs
aaj3f: Agreed, and this was the worse of the two — thank you. Treating shares as cosmetic is exactly wrong in the case where they're the only thing enforcing. Fixed in `ecba51446`: shares are validated before anything gates on them, and non-finite, negative, or a sum more than 1pp off 100 is now a refusal (`MismatchKind::PhaseShares`) rather than an annotation. I went with refusing over recomputing from `ns` to match the rest of the gate — a file that disagrees with itself gets "I won't guess" — and the remedy string names `ScenarioMeta::with_phase_ns` as the way to regenerate. The `anomalies` channel is gone since this was its only user. Your exact scenario is pinned by `all_zero_current_shares_refuse_instead_of_gating_as_improved`, which asserts no bogus `Improved` row is emitted.

fluree-bench-support/src/baseline.rs
aaj3f: Fixed in `ecba51446` — `accumulate` now rejects `prev.schema_version > BASELINE_SCHEMA_VERSION` before merging, matching `compare`. The silent-field-drop path you describe is the real hazard: it writes the file back, so the loss is permanent rather than just a bad comparison. Test: `accumulate_refuses_a_newer_schema`.

fluree-bench-support/src/baseline.rs
aaj3f: Fixed in `ecba51446`. Differing non-empty `git_sha`s now refuse, with `--allow-revision-drift` (and `AccumulateOptions::allow_revision_drift` for library callers) as the explicit override. The MAD-widening consequence you name is the one that worried me most — it doesn't just mislabel the samples, it actively raises the threshold that would have caught the change. Test: `accumulate_refuses_across_revisions_unless_overridden`, plus `accumulate_allows_a_matching_or_absent_revision` for the normal loop and the no-git case.

fluree-bench-support/src/baseline.rs
aaj3f: Fixed in `ecba51446`. Since the raw samples are stored, as you say, they're now the authority: `n`/`median_ns`/`mad_ns` are recomputed from `samples_ns` at load. A set that can't be recomputed — fewer than `MIN_NOISE_SAMPLES`, or any non-finite/non-positive sample — is returned unusable rather than trusted, which drops the floor for that entry. That's the safe direction, since the floor only ever widens a budget. Tests: `noise_stats_are_recomputed_from_their_own_samples_at_load`, `fabricated_noise_without_samples_is_unusable`, `non_positive_samples_are_rejected_as_a_floor`.

fluree-bench-support/src/baseline.rs
aaj3f: Fixed in `ecba51446`. `load_baseline` now requires `schema_version >= 1` and at least one scenario. The failure mode you describe is the one I'd least want to ship — a gate that passes *because* it had no reference points looks identical to a gate that passed on merit. Test: `empty_document_is_rejected_at_load` covers both `{}` and an explicit empty `entries`.

fluree-bench-support/src/meta.rs
aaj3f: Fixed in `55cc517e7`, both ways you suggested. `short_hash` now slices on a character boundary so no input can panic it, and `CorpusInfo::validate` rejects anything that isn't exactly 64 lowercase ASCII hex digits. Doing only the truncation would have left the identity semantics resting on an unvalidated field, and doing only the validation would have left the panic reachable from the very error path that reports it — the crash landed precisely when the tool was trying to explain its refusal. Tests: `short_hash_never_panics_on_multibyte_input` and `validate_rejects_malformed_hashes`.

.github/workflows/bench.yml
aaj3f: Fixed in `fb2f60b41`. Validates a positive integer before the loop and sets `if-no-files-found: error` on the upload. A capture job that goes green having captured nothing is the worst outcome available here, because the next person reasonably assumes the artifact exists.

BENCHMARKING.md
aaj3f: Fixed in `517a03c5a`. You're right that following it literally lands phase 2 as a red gate — the flag comes off and the compare then refuses against the stale local-class file. The doc now spells out the `mv guardrails-pre-ci.json bench-baselines/guardrails-pre.json` replace step explicitly, and names the alternative (keep the `-ci` filename and repoint every `--baseline` path) so the choice is visible rather than implied.

aaj3f:
@bplatz -- ready for re-review when you have time

fluree-bench-support/src/baseline.rs
```diff
@@ -181,15 +639,143 @@ pub fn capture(
...
+            });
+        match out.entries.get_mut(id) {
+            Some(entry) if entry.mean_ns > 0.0 => {
+                let mut samples = prior;
+                samples.push(entry.mean_ns);
+                entry.noise = Some(NoiseStats::from_samples(samples));
+            }
+            // Not rerun this pass: carry the previous entry forward whole.
+            _ => {
+                out.entries.insert(id.clone(), prev_entry.clone());
```
bplatz: The corpus guard above skips ids absent from `fresh` (the `continue` at the top of that loop), but this arm re-inserts exactly those entries into `out` — whose file-level `corpus` is `fresh`'s. An entry that declared no entry-level corpus and was inheriting `prev.corpus` therefore resolves against a different corpus after the merge, with its samples unchanged.

Reproduced with `prev` = {q1, q2} under a file-level corpus A, and a pass that reran only q3 under corpus B:

```
prev   q2 corpus = ("A", "61be55a8")
merged q2 corpus = ("B", "81cc5b17")   # same samples, relabelled
```

The overlapping-id case is correctly refused, so this only bites when a pass doesn't rerun every scenario — `--only` subsetting, or a bench that emits no criterion output that pass. The CI loop reruns the same subset each pass, and nothing produces corpus blocks yet, so it isn't reachable today. It is the same silent identity rewrite as the `(Some, None)` case though, just on the carry-forward path instead of the pooling path, so I'd rather it not sit latent: either stamp the resolved `prev.corpus_for(id)` onto the carried-forward entry, or run the identity check over the union of `prev.entries` and `fresh.entries` rather than the intersection.

fluree-bench-support/src/baseline.rs
```diff
@@ -181,15 +639,143 @@ pub fn capture(
...
+            "refusing to accumulate across host classes: existing baseline is '{}' ({}), this run \
+             is '{}' ({}). Samples from different machines cannot be pooled into one median/MAD.",
+            prev.host.class,
+            prev.host.describe(),
+            fresh.host.class,
+            fresh.host.describe(),
+        ));
+    }
+    if !opts.allow_revision_drift {
+        if let (Some(prev_sha), Some(fresh_sha)) = (&prev.git_sha, &fresh.git_sha) {
```
bplatz: This requires both sides `Some`, and the merge below takes `fresh`'s `git_sha` — so a single pass with no revision resets the accumulator and reopens the gate for every pass after it:

```
after a no-git pass, git_sha = None
next pass at a different revision -> ACCEPTED
  pooled samples   = [1000.0, 1010.0, 5000.0]
  labelled git_sha = Some("ccccccc")
```

That's the MAD-widening case this guard exists for, reached in two steps rather than one. The trigger is `git_short_sha()` returning `None`, so it's a local/container concern rather than a CI one — CI always has git. Carrying `prev.git_sha` forward when `fresh` has none, or treating `None` vs `Some` as drift, would close it, and either way `--allow-revision-drift` stays the explicit path.

bplatz (approved):
Re-reviewed. All nine items are addressed, and I checked each against the code rather than the replies — the two fail-open paths I called merge-blocking are genuinely closed:

- `corpus_refusal` is a single helper called from both `compare` and `accumulate`, so the asymmetry can't drift apart again, and the `(Some, None)` direction refuses.
- Share validation runs ahead of any gating decision, and `MismatchKind::PhaseShares` refuses rather than annotates. The all-zero-shares case emits no bogus `Improved` row.
- Noise stats are recomputed from `samples_ns` at the serde boundary, so every reader sees derived numbers; an unrecomputable set drops the floor, which is the widening-only direction.
- `{}` and an explicit empty `entries` are both rejected at load.
- `short_hash` is char-boundary safe *and* the hash is validated — agreed that doing only one would have left either the panic or the identity semantics exposed.
- The capture job fails a non-positive `capture_samples` and errors on an empty upload.
- The doc names the `mv` explicitly, so phase 2 doesn't land as a red gate.

Verified locally: 139/139 tests (17 new, one per item), clippy `--all-targets` clean, `fmt --check` clean, and the committed schema-1 `guardrails-pre.json` still refuses at exit 2 and compares advisory under `--allow-host-mismatch`.

Approving. Two residuals inline — both non-blocking and both latent today. They're the same shape as items 1 and 4 on the sibling code path, so worth a look before merge rather than a reason to hold it.

fluree-bench-support/src/baseline.rs
aaj3f: Fixed in `e06187439`, and I did both — they turn out to be complementary rather than alternatives, which I didn't expect going in.

The union walk alone still lets the case where prev resolved to NO corpus and fresh introduces a file-level one slip through: the carried entry inherits an identity its samples never had. The stamp alone has nothing to stamp in exactly that case. So: carried-forward entries get `prev.corpus_for(id)` stamped when it resolves to something, and where it resolves to nothing while fresh declares a corpus, the merge refuses — "explicitly no corpus" isn't expressible through inheritance, so there's no honest value to write.

Resolution stays single-sourced as you asked: every decision on both sides goes through `corpus_for`, and the intersection case still calls the same `corpus_refusal` helper `compare` uses.

Your repro is the regression test verbatim (`carried_forward_entries_keep_their_own_corpus`): prev {q1,q2} under file-level A, a pass rerunning only q3 under B, asserting q2 still resolves to A with its samples unchanged and that q3 keeps B. Plus `carrying_forward_into_a_newly_declared_corpus_refuses` for the unstampable case. Both confirmed failing before the fix.

fluree-bench-support/src/baseline.rs
aaj3f: Fixed in `e06187439`. I went with **treating None vs Some as drift** rather than carrying `prev.git_sha` forward.

Carrying it forward would assert that the fresh samples were taken at a revision nothing actually observed — fabricating provenance in order to preserve a label, which is the same shape as the identity rewrites in R1 and item 1. Treating absence as part of revision identity says the honest thing instead: a sample whose revision is unknowable is precisely the one that can't be pooled safely. Both-absent still accumulates fine, so the no-git tarball/container workflow is unaffected, and `--allow-revision-drift` remains the explicit path through.

Your two-step repro is the test (`a_no_git_pass_cannot_launder_a_revision_change`), written to accept either outcome — the no-git pass itself refusing, or it being accepted and the *next* pass at a different revision refusing — so it pins the hole rather than my particular choice of where to close it. Confirmed failing before the fix. One earlier test flipped with the contract: one-sided `None` was asserted as allowed and is now asserted as drift in both directions.

PR #1551, opened 2026-07-29
fix(import): document-scoped blank node identity, and skolem ids SPARQL can write

aaj3f:
# Scope blank-node labels to the document, and make the ids writable

Two bugs with one root. Bulk import derived a blank node's identity from *where the chunker happened to cut*, and rendered that identity in a form no SPARQL parser accepts.

## What was wrong

**Labels split across chunk boundaries.** Import skolemized labeled blank nodes from the per-chunk commit id, so `_:x` in chunk 1 and `_:x` in chunk 7 of one Turtle file became two different subjects. RDF scopes a blank-node label to the document containing it; one `.ttl` file is one document. This is the chunked-Turtle sibling of the TriG bug fixed in #1432, which established the rule that a label unifies across one document and stays distinct across documents — #1432 fixed the named-graph half, the shipping parallel Turtle path still split labels the other way. Nothing surfaced it: the labels come out as opaque `_:fdb-…` ids, so the data simply arrived silently un-joined.

**The ids could not be written back.** The skolem key was `{ledger_id}-{scope}`, and a ledger id contains `/` and `:`. So a minted id looked like `_:fdb-lubm/bench:main-d3-genid10`, and SPARQL's `BLANK_NODE_LABEL` production admits neither character. Those ids came out of a query and could never go back in — a bulk-imported blank-node structure was the one kind Fluree could not edit in place from SPARQL, and `docs/transactions/update-where-delete-insert.md` said so out loud as a standing caveat.

**And the identity moved when the import did.** The document scope was the file's *position* in the sorted directory listing, and anonymous nodes carried the global chunk ordinal on top of that. Dropping one more file into the directory renumbered every id behind it, for data whose bytes had not changed.

## The format

```
fdb-d<13 lowercase base36 digits>-<label>
    ^ ^                          ^
    | |                          `- the label as written in the source
    | `- xxh64(namespace ‖ 0x00 ‖ document key), zero-padded
    `- the document-scope marker
```

`fluree_db_core::skolem` is the single definition. `[0-9a-z-]` only, so the ids are writable in SPARQL, Turtle and JSON-LD with no escaping — which is the headline, and the thing the PR exists to deliver.

Three properties are load-bearing:

**The namespace replaces the ledger id, it does not drop it.** Two ledgers loaded from one source tree still mint disjoint blank nodes, because the ledger id salts the hash. It is the *normalized* id, so `my/ledger` and `my/ledger:main` — one ledger, two spellings — agree. `--skolem-namespace <id>` overrides it for the cases where identical ids across ledgers are the point: a rebuild-and-diff, a sharded load of one logical dataset, a staging/production pair.

**The document key is relative.** A path relative to the import root; for remote sources, the address relative to the listing prefix. So the same tree mints the same ids whether it is mounted at `/tmp/x` or `/data/x`. `RemoteSource::OrderedObjects` supplies no prefix, so its keys are whole addresses — and since `RemoteObject::address` is backend-opaque by contract, hashing it pins those ids to that backend's address encoding. That is called out at the code.

**The width is fixed.** 13 base36 digits hold any `u64` (`36^13 > 2^64`), so the scope is always exactly 14 bytes and splits off a minted id cleanly — even from a label that contains any number of `-` itself. That is what makes the manifest (below) usable, and it is what gives structural disjointness from the other two `fdb-` minters: import's first hyphen falls at offset 14, a staged transaction's at 16 (nanoseconds-since-epoch is 16 hex digits for every clock between 2004 and 2154), `BNODE()`'s at 8 (RFC-4122 text). Pinned by a unit test.

**Anonymous nodes** now carry the sub-chunk index within their document rather than the global chunk ordinal, so they stop depending on the rest of the file set too. `(document, sub-chunk, counter)` is jointly unique: the base per document, the sub-chunk per chunk within a document, the counter per node within a chunk. Being statement-local means nothing can *reference* an anonymous node from elsewhere; it does not stop two of them from colliding on an id, which is why the mint needs all three.

## The ndjson arm

Previously chunk-scoped, so labels merged across lines within a chunk and split across chunks — a boundary set by `chunk_size_mb`, not by anything in the data. An ndjson document is now one **context segment**: the run of lines sharing one `@context`.

Not the file, and not the line. Three facts in our own code fix it there. `ndjson_splitter`'s module doc says each line is one complete JSON-LD *node*, not document. A line cannot be read standalone anyway — it resolves against a `@context` declared on a different line. And the emitter wraps a run of nodes as members of one `@graph`, which is what co-document means. The NDJSON convention that each line is an independent JSON *value* is satisfied by that and says nothing about the JSON-LD document boundary.

The boundary is the `@context` switch: a mid-stream lone `{"@context": …}` line replaces the shared context, and the reader seals the outgoing chunk before rebuilding the prefix, so nodes on either side become different `@graph` documents under different contexts. That is the `cat a.jsonl b.jsonl > combined.jsonl` seam — a first-class NDJSON workflow, and the case where per-file scoping would have silently merged `a.jsonl`'s `_:x` with `b.jsonl`'s.

The key gains the segment ordinal: `xxh64(namespace ‖ 0 ‖ doc_key ‖ 0 ‖ segment)`, segment 0 everywhere else. It is an **ordinal** — how many switches precede the segment — deliberately not the chunk index the segment starts at, which would move with `chunk_size_mb` and put back exactly the import-shape dependence this change removes. The reader gets its own `NdjsonChunkPayload` rather than widening the shared `ChunkPayload`, which the structured JSON-LD splitter also uses and which has no notion of segments.

## Fail-loud on duplicate documents

`DocIds::new` hashes the whole document set up front and refuses an import whose documents do not hash apart, naming both. A duplicate is nearly always a duplicate *key* — `RemoteSource::OrderedObjects` takes a caller-supplied list, and nothing stops it naming one address twice — and much more rarely an `xxh64` collision between two different keys. Both are fatal the same way: two documents sharing a scope silently merge their `_:x` nodes. Refusing up front costs nothing; discovering it after a multi-hour import costs the import.

The local arms cannot produce a duplicate (`read_dir` yields each entry once), so the guard is unit-tested directly and the reachable arm is covered end to end.

## Provenance

A hash carries no provenance, so the import now writes the mapping into the `txn-meta` graph it already builds: one `db:importSource` triple per source document, subject = the scope's own blank node, object = the document key. Records ride the existing meta chunk — no extra pass, no extra file. The ndjson arm discovers documents as it reads (a `@context` switch opens one), so its producer logs the segments it actually opens and the meta chunk adds a row for each against the same file name — every minted scope resolves, even though the manifest does not say *which* segment of the file it was.

```sparql
SELECT ?source WHERE {
  GRAPH <urn:fluree:my/ledger:main#txn-meta> {
    _:fdb-d1t3k9x0abcdef <https://ns.flur.ee/db#importSource> ?source
  }
}
```

`skolem::split_doc_scope` takes a minted id to its scope; the manifest takes the scope to the file.

## Migration

There is no mixed-format ledger to migrate. Import requires a fresh ledger (`import.rs`, the `already has commits` guard), so a ledger's import ids are all of one format. Nothing parses a minted label apart, either — the only structural reader is `stable_blank_node_local`, which just checks the `fdb-` prefix.

Existing ledgers are untouched; this changes only what a *new* import mints. Two consequences worth stating plainly:

- **Re-importing a source produces different ids than it did before.** Anything that recorded a pre-4.1.5 `_:fdb-…` id externally and expects a rebuild to reproduce it will not get it back.
- **ndjson blank-node scope changes.** Labels that used to merge within a chunk and split across chunks now merge across a whole context segment and split at `@context` switches.

`ns_encoding::canonical_split` keeps its `_:` special case: pre-4.1.5 ledgers hold ids with colons in them, and a user-authored label can carry one on any surface that admits it. The comment now says that is what it is for.

Two serialization paths are pinned down rather than changed. `validate`'s Turtle renderer keeps its lossy sanitizer and now says why that is right for a human-facing report and wrong on the export path. `export` says explicitly that it writes labels verbatim to preserve identity, and that a pre-4.1.5 ledger is the one case where that emits Turtle a strict parser rejects — sanitizing there would merge two distinct nodes onto one id, which is worse than invalid syntax.

## Cost

No cross-worker coordination and no throughput cost. `blank_node_sid` is a pure function of the key, so two threads parsing chunk 1 and chunk 7 independently derive the same Sid for the same label — no shared map, no lock, no ordering dependency. The chunk channel stays allocation-free: `DocChunk` is a `Copy` 16-byte struct, and the skolem base is rendered once per chunk in the worker (the same one `format!` per chunk the old code paid). Hashing happens once per *document*, at chunk-source construction.

## Tests

Failing before, passing after:

| test | before |
|---|---|
| `imported_blank_node_id_is_writable_in_sparql` | the `DELETE` template does not parse: `_:fdb-test/skolem-sparql:main-d0-shared` |
| `minted_ids_do_not_depend_on_the_rest_of_the_directory` | one extra file ahead of the document changes both its labeled and its anonymous ids |
| `ndjson_blank_label_is_scoped_to_its_source_not_its_chunk` | the two occurrences land in different chunks and become two subjects |
| `import_ttl_blank_label_scoped_across_chunks` | `_:shared` in chunk 0 and chunk 1 of one file are two subjects |
| `import_ttl_anonymous_nodes_distinct_across_chunks` | both anonymous nodes resolve to one id |
| `ndjson_context_switch_starts_a_new_blank_node_scope` | both `_:shared` occurrences resolve to `_:fdb-d2dleqty4209so-shared`, one node |
| `import_manifest_covers_ndjson_context_segments` | one manifest row for a file that minted two scopes |
| `trig_fast_path_delegation_keeps_the_document_key` | (mutation-verified) rebuilding the key inside `import_commit` splits the delegating and direct TriG paths |

Invariant pins that hold on both sides, kept because the new code re-derives what they assert: `ledger_id_spelling_does_not_change_minted_ids`, `different_ledgers_mint_different_ids`, `labels_stay_distinct_between_documents`.

Plus: `explicit_skolem_namespace_shares_ids_across_ledgers`, `import_manifest_maps_blank_node_scopes_back_to_documents` (the full id → scope → document round trip), `duplicate_remote_addresses_abort_the_import`, `rechunk_scopes_survive_split_and_coalesced_files_interleaved` (serial vs. 4-worker parity plus a no-reused-`(doc, sub-chunk)`-pair check), the `fluree_db_core::skolem` unit tests, and a CLI integration test that drives the real binary — the only thing that catches a flag being defined but not plumbed.

## Named residuals

- **Federated same-name servers still unify.** The salt is a ledger id, not a server identity, so two Fluree servers hosting `my/ledger:main` and importing the same tree mint the same blank-node ids. Under federation those look like one node. `--skolem-namespace` is the lever if that matters.
- **ndjson written as one independent document per line still unifies.** An inline `@context` on *every* line makes those lines ordinary nodes in one `@graph` — one document — so a `_:x` reused across them resolves to one node. Correct JSON-LD, but not what the NDJSON-of-independent-documents convention suggests. Making each inline-context line its own segment would vary the document key *within* a chunk, and the pipeline is one sink and one skolem base per chunk. Documented in `docs/cli/create.md`; exposure needs explicit `_:` labels reused across lines of a per-line-context file.
- **Anonymous ids remain chunk-config dependent.** They key on the sub-chunk index, so re-importing the same tree at a different `chunk_size_mb` renumbers them (labeled ids are unaffected). They also move if coalescing pulls a file into a neighbour's payload — `coalesce_unsafe` refuses files with `_:` labels, which protects labeled nodes but not anonymous ones. Nothing can *reference* an anonymous node, so this is an identity-stability limit, not a correctness one.
- **`OrderedObjects` keys are backend-flavored.** No listing prefix to strip, and `RemoteObject::address` is opaque by contract, so re-importing the same objects through a different backend mints different ids.
- **The server's `POST /import` route has no `skolem_namespace`.** It exposes no import tuning knobs at all today (no memory budget, no parallelism), so adding one there is its own change.
- **`StreamingTurtleReader` does not terminate on a sub-KB chunk size.** Found while sizing the rechunk fixture. Unreachable from the import config, which sizes chunks in megabytes; noted at the fixture and not fixed here.

### DIFF
--- a/fluree-db-api/src/import.rs
+++ b/fluree-db-api/src/import.rs
@@ -809,10 +809,14 @@ type RemoteChunk = (usize, DocChunk, Vec<u8>);
 ///
 #[derive(Debug)]
 pub(crate) struct DocIds {
+    /// Salt for every id in this import — the normalized ledger id unless the
+    /// caller overrode it. Retained because the ndjson arm discovers extra
+    /// documents (context segments) mid-read, after this table is built.
+    namespace: String,
     /// Document keys in source order — paths relative to the import root, or
     /// remote addresses relative to the listing prefix.
     keys: Vec<String>,
-    /// `skolem::doc_id` of each key, same order.
+    /// `skolem::doc_id` of each key at segment 0, same order.
     ids: Vec<u64>,
 }
 
@@ -831,7 +835,7 @@ impl DocIds {
         let mut ids = Vec::with_capacity(keys.len());
         let mut seen: HashMap<u64, usize> = HashMap::with_capacity(keys.len());
         for (idx, key) in keys.iter().enumerate() {
-            let id = fluree_db_core::skolem::doc_id(namespace, key);
+            let id = fluree_db_core::skolem::doc_id(namespace, key, 0);
             if let Some(&first) = seen.get(&id) {
                 return Err(ImportError::Source(format!(
                     "two import documents share the blank-node scope \
@@ -846,7 +850,11 @@ impl DocIds {
             seen.insert(id, idx);
             ids.push(id);
         }
-        Ok(Self { keys, ids })
+        Ok(Self {
+            namespace: namespace.to_string(),
+            keys,
+            ids,
+        })
     }
 
     /// Hashed id of document `idx`.
@@ -863,6 +871,12 @@ impl DocIds {
         self.ids.clone()
     }
 
+    /// Document keys in source order, for the ndjson producers, which hash
+    /// their own per-segment ids.
+    fn keys(&self) -> Vec<String> {
+        self.keys.clone()
+    }
+
     /// The import manifest: `(blank-node local name of the scope, document
     /// key)` for every document.
     ///
@@ -870,21 +884,41 @@ impl DocIds {
     /// every id minted from that document shares — so a reader holding an
     /// opaque `_:fdb-d…-label` recovers the source with
     /// `fluree_db_core::skolem::split_doc_scope` and one lookup.
-    fn manifest(&self) -> Vec<(String, String)> {
-        self.ids
+    ///
+    /// `extra_segments` names `(document index, segment ordinal)` pairs beyond
+    /// segment 0 — the ndjson arm cuts one file into several documents at its
+    /// `@context` switches, and which ones exist is only known once the file
+    /// has been read. Each gets its own row against the same document key, so
+    /// every minted scope resolves even though the manifest does not say which
+    /// segment of the file it was.
+    fn manifest(&self, extra_segments: &[(usize, u32)]) -> Vec<(String, String)> {
+        let scope_row = |id: u64, key: &str| {
+            (
+                format!(
+                    "{}{}",
+                    fluree_db_core::ns_encoding::STABLE_BLANK_NODE_LABEL_PREFIX,
+                    fluree_db_core::skolem::doc_scope(id)
+                ),
+                key.to_string(),
+            )
+        };
+        let mut rows: Vec<(String, String)> = self
+            .ids
             .iter()
             .zip(&self.keys)
-            .map(|(&id, key)| {
-                (
-                    format!(
-                        "{}{}",
-                        fluree_db_core::ns_encoding::STABLE_BLANK_NODE_LABEL_PREFIX,
-                        fluree_db_core::skolem::doc_scope(id)
-                    ),
-                    key.clone(),
-                )
-            })
-            .collect()
+            .map(|(&id, key)| scope_row(id, key))
+            .collect();
+        for &(idx, segment) in extra_segments {
+            if segment == 0 {
+                continue; // already covered above
+            }
+            let Some(key) = self.keys.get(idx) else {
+                continue;
+            };
+            let id = fluree_db_core::skolem::doc_id(&self.namespace, key, segment);
+            rows.push(scope_row(id, key));
+        }
+        rows
     }
 }
 
@@ -918,6 +952,13 @@ type RemoteChunkRx = Arc<std::sync::Mutex<std::sync::mpsc::Receiver<RemoteChunk>
 /// exit to distinguish clean completion from producer failure.
 pub struct RemoteChunkProducer {
     pub(crate) rx: RemoteChunkRx,
+    /// `(document index, segment ordinal)` for every context segment an ndjson
+    /// source actually opened, filled by the producer as it reads. Empty for
+    /// whole-object sources, which are one document each.
+    ///
+    /// Read after the producer thread is joined; that is the only point at
+    /// which it is known to be complete.
+    pub(crate) segments: Arc<std::sync::Mutex<Vec<(usize, u32)>>>,
     /// Take-once handles, accessible through `&self` via Mutex<Option<_>>.
     /// `error_rx` carries `Some(err)` on producer failure, `None` on success.
     pub(crate) error_rx:
@@ -1343,7 +1384,8 @@ fn resolve_chunk_source(
             let channel_capacity = config.effective_max_inflight();
             let producer = spawn_local_ndjson_producer(
                 files,
-                doc_ids.ids(),
+                doc_ids.keys(),
+                skolem_namespace.to_string(),
                 config.ndjson_first_line_context,
                 chunk_size_bytes,
                 channel_capacity,
@@ -1412,7 +1454,8 @@ fn resolve_chunk_source(
         let doc_ids = single_file_docs(path)?;
         let producer = spawn_local_ndjson_producer(
             vec![path.to_path_buf()],
-            doc_ids.ids(),
+            doc_ids.keys(),
+            skolem_namespace.to_string(),
             config.ndjson_first_line_context,
             chunk_size_bytes,
             channel_capacity,
@@ -1701,6 +1744,8 @@ fn spawn_remote_producer(
 
     RemoteChunkProducer {
         rx: Arc::new(std::sync::Mutex::new(std_rx)),
+        // One whole object is one document; nothing segments it.
+        segments: Arc::default(),
         error_rx: std::sync::Mutex::new(Some(error_rx)),
         bridge_handle: std::sync::Mutex::new(Some(bridge_handle)),
         estimated_count,
@@ -1813,21 +1858,37 @@ fn estimate_ndjson_chunks(total_bytes: u64, chunk_size_bytes: u64) -> usize {
     total_bytes.div_ceil(chunk_size_bytes.max(1)).max(1) as usize
 }
 
-/// Spawn a producer that streams ndjson from a sequence of `(label, factory)`
-/// byte sources, chaining one [`NdjsonReader`] per source and forwarding their
-/// JSON-LD chunks into a shared channel with a global chunk index. The readers
-/// are independent, so each source's own leading `@context` is honored.
+/// One ndjson byte source for [`spawn_chained_ndjson_producer`].
+struct NdjsonSource {
+    /// Human-readable name — a path or remote address — for error text.
+    label: String,
+    /// Position of this source in the import's document list.
+    doc_idx: usize,
+    /// The key this source's documents hash under (see [`DocIds`]). Each of
+    /// its context segments is a document in its own right, so the id is
+    /// derived per chunk rather than precomputed.
+    doc_key: String,
+    factory: NdjsonSourceFactory,
+}
+
+/// Spawn a producer that streams ndjson from a sequence of byte sources,
+/// chaining one [`NdjsonReader`] per source and forwarding their JSON-LD chunks
+/// into a shared channel with a global chunk index. The readers are
+/// independent, so each source's own leading `@context` is honored.
 /// Reuses [`RemoteChunkProducer`]'s channel + `error_rx` machinery;
 /// `per_chunk_format` is empty because the chunks ride the
 /// [`ChunkSource::JsonLdStream`] path, which always parses them as JSON-LD.
 fn spawn_chained_ndjson_producer(
-    sources: Vec<(String, u64, NdjsonSourceFactory)>,
+    sources: Vec<NdjsonSource>,
+    skolem_namespace: String,
     policy: FirstLineContextPolicy,
     chunk_size_bytes: u64,
     in_flight: usize,
     estimated_count: usize,
 ) -> RemoteChunkProducer {
     let in_flight = in_flight.max(1);
+    let segments: Arc<std::sync::Mutex<Vec<(usize, u32)>>> = Arc::default();
+    let segment_log = Arc::clone(&segments);
     let (std_tx, std_rx) = std::sync::mpsc::sync_channel::<RemoteChunk>(in_flight);
     let (error_tx, error_rx) = tokio::sync::oneshot::channel::<Option<ImportError>>();
 
@@ -1838,7 +1899,14 @@ fn spawn_chained_ndjson_producer(
         .name("ndjson-producer".into())
         .spawn(move || {
             let mut global_idx = 0usize;
-            for (label, doc_id, factory) in sources {
+            for source in sources {
+                let NdjsonSource {
+                    label,
+                    doc_idx,
+                    doc_key,
+                    factory,
+                } = source;
+                let mut last_segment: Option<u32> = None;
                 let byte_source = match factory() {
                     Ok(r) => r,
                     Err(e) => {
@@ -1864,16 +1932,32 @@ fn spawn_chained_ndjson_producer(
                     };
                 loop {
                     match reader.recv_chunk() {
-                        Ok(Some((local_idx, bytes))) => {
-                            // One ndjson SOURCE is one document. `NdjsonReader`
-                            // applies the source's leading `@context` to every
-                            // chunk it cuts, so the file — not the chunk, and
-                            // not the line — is the unit Fluree already treats
-                            // as one JSON-LD document. `local_idx` is the
-                            // reader's own 0-based counter, so it is the
-                            // sub-chunk index this document needs.
+                        Ok(Some((local_idx, segment, bytes))) => {
+                            // One CONTEXT SEGMENT is one document. Within a
+                            // segment the nodes share an `@context` and are
+                            // emitted as members of one `@graph`, which is
+                            // what makes them co-document; a `@context` switch
+                            // seals that document and opens the next, so
+                            // `cat a.jsonl b.jsonl` keeps the two files' `_:x`
+                            // labels apart. Neither the chunk nor the line is
+                            // the unit: a chunk is a cut of a segment, and a
+                            // line is a node.
+                            //
+                            // `local_idx` is the reader's own chunk counter,
+                            // monotonic across the whole file and never reset
+                            // at a switch, so it is distinct within every
+                            // segment — which is all the sub-chunk index has
+                            // to be.
+                            if last_segment != Some(segment) {
+                                last_segment = Some(segment);
+                                segment_log.lock().unwrap().push((doc_idx, segment));
+                            }
                             let doc = DocChunk {
-                                doc: doc_id,
+                                doc: fluree_db_core::skolem::doc_id(
+                                    &skolem_namespace,
+                                    &doc_key,
+                                    segment,
+                                ),
                                 sub_chunk: u32::try_from(local_idx).unwrap_or(u32::MAX),
                             };
                             if std_tx.send((global_idx, doc, bytes)).is_err() {
@@ -1907,6 +1991,7 @@ fn spawn_chained_ndjson_producer(
 
     RemoteChunkProducer {
         rx: Arc::new(std::sync::Mutex::new(std_rx)),
+        segments,
         error_rx: std::sync::Mutex::new(Some(error_rx)),
         bridge_handle: std::sync::Mutex::new(Some(producer_handle)),
         estimated_count,
@@ -1925,7 +2010,8 @@ fn spawn_chained_ndjson_producer(
 fn spawn_remote_ndjson_producer(
     storage: Arc<dyn StorageRead>,
     objects: Vec<RemoteObject>,
-    doc_ids: Vec<u64>,
+    doc_keys: Vec<String>,
+    skolem_namespace: String,
     policy: FirstLineContextPolicy,
     chunk_size_bytes: u64,
     in_flight: usize,
@@ -1934,11 +2020,11 @@ fn spawn_remote_ndjson_producer(
     let ranged = storage.supports_ranged_reads();
     let total_bytes: u64 = objects.iter().map(|o| o.size_bytes).sum();
     let estimated_count = estimate_ndjson_chunks(total_bytes, chunk_size_bytes);
-    let sources: Vec<(String, u64, NdjsonSourceFactory)> = objects
+    let sources: Vec<NdjsonSource> = objects
         .into_iter()
         .enumerate()
         .map(|(idx, obj)| {
-            let doc_id = doc_ids.get(idx).copied().unwrap_or(0);
+            let doc_key = doc_keys.get(idx).cloned().unwrap_or_default();
             let storage = Arc::clone(&storage);
             let handle = handle.clone();
             let label = obj.address.clone();
@@ -1969,11 +2055,17 @@ fn spawn_remote_ndjson_producer(
                     decode_buffered(comp, Box::new(std::io::Cursor::new(bytes)))
                 })
             };
-            (label, doc_id, factory)
+            NdjsonSource {
+                label,
+                doc_idx: idx,
+                doc_key,
+                factory,
+            }
         })
         .collect();
     spawn_chained_ndjson_producer(
         sources,
+        skolem_namespace,
         policy,
         chunk_size_bytes,
         in_flight,
@@ -1985,7 +2077,8 @@ fn spawn_remote_ndjson_producer(
 /// `open_decoded` handles `.gz`/`.zst` transparently.
 fn spawn_local_ndjson_producer(
     files: Vec<PathBuf>,
-    doc_ids: Vec<u64>,
+    doc_keys: Vec<String>,
+    skolem_namespace: String,
     policy: FirstLineContextPolicy,
     chunk_size_bytes: u64,
     in_flight: usize,
@@ -1998,18 +2091,24 @@ fn spawn_local_ndjson_producer(
         .map(|m| m.len())
         .sum();
     let estimated_count = estimate_ndjson_chunks(total_bytes, chunk_size_bytes);
-    let sources: Vec<(String, u64, NdjsonSourceFactory)> = files
+    let sources: Vec<NdjsonSource> = files
         .into_iter()
         .enumerate()
         .map(|(idx, f)| {
-            let doc_id = doc_ids.get(idx).copied().unwrap_or(0);
+            let doc_key = doc_keys.get(idx).cloned().unwrap_or_default();
             let label = f.display().to_string();
             let factory: NdjsonSourceFactory = Box::new(move || open_decoded(&f));
-            (label, doc_id, factory)
+            NdjsonSource {
+                label,
+                doc_idx: idx,
+                doc_key,
+                factory,
+            }
         })
         .collect();
     spawn_chained_ndjson_producer(
         sources,
+        skolem_namespace,
         policy,
         chunk_size_bytes,
         in_flight,
@@ -3327,7 +3426,8 @@ where
                     let producer = spawn_remote_ndjson_producer(
                         Arc::clone(storage),
                         objects,
-                        doc_ids.ids(),
+                        doc_ids.keys(),
+                        skolem_namespace.clone(),
                         config.ndjson_first_line_context,
                         chunk_size_bytes,
                         in_flight,
@@ -5166,7 +5266,16 @@ where
         let p_import_source = spool_config
             .predicate_alloc
             .get_or_insert_parts(fluree::DB, db::IMPORT_SOURCE);
-        let manifest = doc_ids.manifest();
+        // The ndjson producer discovers documents as it reads (a `@context`
+        // switch opens a new one), and it has been joined by now, so its log
+        // is complete.
+        let observed_segments = match &**chunk_source {
+            ChunkSource::Remote(producer) | ChunkSource::JsonLdStream(producer) => {
+                producer.segments.lock().unwrap().clone()
+            }
+            _ => Vec::new(),
+        };
+        let manifest = doc_ids.manifest(&observed_segments);
 
         // txn-meta is always pre-seeded as dict_id=0 in the graph allocator
         // (via SharedDictAllocator::new_graph), so g_id = dict_id + 1 = 1.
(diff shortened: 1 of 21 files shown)


fluree-db-api/src/import.rs
```diff
@@ -657,8 +657,32 @@ pub(crate) enum RemoteFormat {
...
+/// the parser workers.
+///
+/// `doc_index` names the **source document** the bytes were cut from. RDF
+/// scopes blank-node labels to a document, and only the producer knows how
+/// chunks map onto input files — a large file is sub-split into many chunks,
+/// several small files can share one chunk. Workers turn it into the parser's
+/// `doc_scope` so `_:x` unifies across every chunk of one file and stays
+/// distinct between files. Producers that emit exactly one whole document per
+/// chunk set it equal to `chunk_index`.
+type RemoteChunk = (usize, usize, Vec<u8>);
```
bplatz: This is the type that would change under the proposal in the review body:

```rust
#[derive(Clone, Copy)]
struct DocId(u64);
type RemoteChunk = (usize, DocId, Vec<u8>);
```

Still one `Copy` value per chunk, so the channel stays allocation-free — no per-chunk `String`. The point of computing the hash at the producer rather than at the parse site is that the producer is the only layer that knows what a document IS for its arm (a path, an object address, a label), which is the same argument this PR already makes for `doc_index` itself.

fluree-db-api/src/import.rs
```diff
@@ -657,8 +657,32 @@ pub(crate) enum RemoteFormat {
...
+/// [`ChunkSource::Streaming`] always reads exactly one local file, so every
+/// chunk it emits belongs to the same document.
+const STREAMING_DOC_IDX: usize = 0;
+
+/// Render a document index as a parser `doc_scope` key.
+///
+/// The `d` prefix keeps these keys disjoint from the commit-ordinal keys used
+/// by the serial TriG/JSON-LD paths, so no two documents in one import can
+/// collide on a skolemization base.
+fn doc_scope(doc_idx: usize) -> String {
```
bplatz: Two things about a positional index as the scope key.

**It is not reproducible.** Add one file earlier in the sorted listing and every later document's index shifts, so every blank node in every later file changes identity. Same data, one extra file, completely different node ids.

**It is not traceable.** `d47` tells an operator nothing about which file a node came from, which is a shame given the producer had the path in hand and threw it away.

Hashing the relative path fixes both and costs nothing at this width — see the review body for the base36/collision detail. The `d` prefix disjointness argument in this doc comment still applies unchanged; an 8-char base36 hash is equally disjoint from the bare commit ordinals.

fluree-db-api/src/import.rs
```diff
@@ -1375,7 +1399,8 @@ fn spawn_remote_producer(
                     "remote object size differs from listing metadata"
                 );
             }
-            if tokio_tx.send((idx, bytes)).await.is_err() {
+            // One remote object == one whole document, so doc_index == idx.
+            if tokio_tx.send((idx, idx, bytes)).await.is_err() {
```
bplatz: `obj.address` is right here and is the best document identity in the whole import path — a stable, meaningful, globally-unique key. Hashing it (relative to the listing prefix, so the same bucket layout under a different prefix still matches) is strictly better than `idx`, which reshuffles whenever the listing changes.

fluree-db-transact/src/import_sink.rs
```diff
@@ -754,10 +761,20 @@ mod inner {
...
+                    // while `skolem_base` is deliberately shared by all
+                    // chunks of a document, so `-b{N}` alone would make
+                    // chunk 0's Nth anonymous node and chunk 1's Nth the
+                    // same subject. Being statement-local means nothing can
+                    // *reference* an anonymous node from elsewhere; it does
+                    // not stop two of them from colliding on an id. `t` is
+                    // unique per chunk, which restores that.
                     self.blank_counter += 1;
-                    let label = format!("-b{}", self.blank_counter);
+                    let label = format!("-b{}-{}", self.t, self.blank_counter);
```
bplatz: The reasoning here is right and the comment is a good one. Worth recording the asymmetry it exposes, though: an anonymous node needs CHUNK uniqueness, which `t` now supplies, while `skolem_base` supplies DOCUMENT uniqueness — so the full name `fdb-{ledger}-d{doc}--b{t}-{N}` carries two scopes where one would do. In the `Files` arm they are near-duplicates (`doc == idx`, `t == idx + 1`); in the streaming arm `d0` is pure overhead.

Anonymous nodes are the majority in typical Turtle (every `[ … ]`, every collection cell), so under today's naming that redundancy is worth removing via a separate chunk-scoped base on the sink. Under the 8-char hashed base proposed in the review body it shrinks to a few bytes and I would leave it alone — noting it mainly so the choice is deliberate either way.

fluree-db-transact/src/import.rs
```diff
@@ -834,13 +843,33 @@ mod inner {
...
+    /// chunks — each its own commit — so the commit `t` is the wrong key: it
+    /// splits one document's labels across chunk boundaries.
+    ///
+    /// `doc_scope` is therefore supplied by the caller, which is the only layer
+    /// that knows which chunks came from which source file. It must be stable
+    /// across every chunk of a document and distinct between documents. No
+    /// cross-worker coordination is needed: `blank_node_sid` is a pure function
+    /// of the key, so two threads parsing chunk 1 and chunk 7 independently
+    /// derive the same Sid for the same label.
+    fn skolem_base(ledger_id: &str, doc_scope: &str) -> String {
```
bplatz: This helper is where the contract is documented, but it is only called from `parse_chunk` and `parse_chunk_with_prelude`. The three serial paths (`import.rs:143`, `:305`, `:463`) and `parse_jsonld_chunk` (`:1062`) each build `format!("{ledger_id}-{t}")` inline instead.

Four spellings of one concept, and the readers most likely to get the scoping wrong are exactly the ones not looking at this doc comment. Routing all five through the helper (the serial paths passing their commit ordinal as the `doc_scope`, which is what they already mean) would put the contract where every caller sees it.

bplatz (commented):
The fix is correct and the test coverage is the right four cases (cross-chunk unification, anonymous distinctness, cross-file distinctness on the split path and on the coalesce path). Locally: `cargo clippy -p fluree-db-api -p fluree-db-transact --all-features --all-targets -- -D warnings` clean, 3559 tests pass, and the four new tests plus the pre-existing TriG one are green. Two details I checked specifically and they hold: `coalesce_unsafe` really does reject any payload containing `_:` anywhere, so "a coalesced payload reports its first file" is sound; and the `d` prefix really is disjoint from the bare commit ordinals the serial paths use.

What follows is not a defect list. It is a proposal to change one design decision — **what the document scope is** — which fixes a reproducibility gap this PR leaves open AND makes every skolemized blank node materially shorter. Raising it now because changing it later is a data-format change, and doing it in the same PR that establishes document scoping is much cheaper than doing it after.

## The gap: a positional index is not a document identity

`doc_scope` is `d{file_idx}` — the file's *position* in the sorted listing, not the file. Add one file earlier in the sort order and every subsequent document's index shifts, so every blank node in every later file gets a different id. Re-import the same logical dataset with one new file and nothing lines up. The ids also carry no traceability: `d47` tells an operator nothing about which file a node came from.

Second, and this is where the compaction comes in: the full stored local name today is

```
fdb-{ledger_id}-{doc_scope}-{label}
```

e.g. `fdb-test/blanks:main-d3-shared` (30 chars) and `fdb-test/blanks:main-d3--b4-7` for an anonymous node. The `{ledger_id}` component is provably redundant *within* a ledger (`doc_scope` already separates documents, `t` already separates commits) and it goes stale the moment a ledger is cloned or renamed — the name still records the ORIGINAL ledger. It is also the reason `canonical_split` (`fluree-db-core/src/ns_encoding.rs:168`) carries a blank-node special case: the branch `:` in `my/ledger:main` would otherwise make the opaque-IRI splitter cut at the embedded colon and re-encode the node into the EMPTY namespace.

The one thing the ledger id does buy is cross-ledger distinctness. That looks like the wrong place to solve it: a cross-ledger query that wants to act on a node has to name the graph anyway in order to target the right ledger, so the blank-node identifier is not carrying that weight.

## Proposal: hash a real document key, drop the ledger id

```
fdb-{8-char-base36-hash(doc_key)}-{label}
```

`fdb-3f9a1c2d-shared`. Fixed width regardless of ledger name length, reproducible across machines and across runs with different file sets, no embedded colon (so new data stops needing the `canonical_split` special case — it must stay for existing ledgers, worth a comment there saying why).

**Every producer already holds a stable document identity at the exact line where it currently sends the positional index.** No new plumbing is needed to obtain it; only the payload type changes.

| Arm | Site | Identity already in scope |
|---|---|---|
| `Files` | `import.rs:4413`, `:4515` | `ChunkSource::Files(files)` → `files[idx]`; `is_trig(index)` already indexes the same Vec the same way |
| `LocalRechunk` serial | `local_rechunk_loop`, `emit!(fi, …)` | `files: &[PathBuf]` + `fi` |
| `LocalRechunk` parallel | `local_rechunk_loop_parallel` | `files` + `file_idx` |
| `Remote` | `spawn_remote_producer` | `obj.address` — the strongest identity of the set |
| `Streaming` | `import.rs:3895` | one local file; `ImportSource::Local(PathBuf)` at the pipeline level |
| ndjson | `spawn_chained_ndjson_producer` | `label` per source, already threaded for error messages |

**Key to hash:** the path **relative to the import root**, not absolute — otherwise the same directory imported from `/mnt/a` and `/mnt/b` yields different ids. `obj.address` relative to the listing prefix for remote; basename for the single-file streaming arm.

**Hash:** `xxhash-rust` (`xxh64`) is **already a `fluree-db-api` dependency** (`fluree-db-api/Cargo.toml:89`), so no new dep. Please do NOT reach for `FxHasher` here despite it being everywhere in this repo — it is deterministic in practice but carries no stability contract, and this value lands in persisted data.

**Render base36, not hex.** Same eight characters, but 36^8 ≈ 2.8e12 (~41 bits) versus 16^8 = 4.3e9 (32 bits). Probability-1% collision threshold: ~237k documents for base36, ~9.3k for hex. A 100k-small-file directory import is not exotic, so hex-8 is genuinely too narrow; base36-8 is comfortable.

**Fail loud on collision.** A collision silently merges two documents' blank nodes — exactly the bug this PR exists to fix — so it should not be left to probability. Every arm knows its full document set up front (`Files(Vec<PathBuf>)`, `objects`, `files`), so a single uniqueness check at `ChunkSource` construction covers all of them in one place, matching the fail-closed posture the rest of the import path takes.

### Consequence to accept knowingly

With the ledger id gone, two ledgers that import the SAME file mint the SAME blank-node ids, so in a cross-ledger dataset query those nodes compare equal. For genuinely-the-same-document that is arguably better semantics than today's accident, but `_:fdb-<hash>-x` stops being ledger-unique — worth a grep for anything that assumes a blank node names exactly one ledger's node before committing.

## Smaller notes, independent of the above

- `skolem_base()` carries the contract documentation, but the three serial paths and `parse_jsonld_chunk` each build `format!("{ledger_id}-{t}")` inline instead of calling it. Four spellings of one concept, and the readers most likely to get it wrong are the ones not looking at the helper.
- The title says "document-scoped blank node identity across chunks", which holds for Turtle/N-Triples but not for JSON-LD ndjson (correctly, and the code comment explains why). Worth saying so in the body so nobody reads the title as broader than it is.
- Re-running an import after this change yields different blank-node ids than before it. Inherent to the fix, but one line in the body would save someone a confused bug report.

Leaving this as a comment rather than an approval only so the scope question gets settled first — the fix as written is correct and I would approve it as-is if you would rather land it and do the hashed key as a follow-up.

aaj3f:
This review moved the design, and most of it survives contact with the code — the shipped result is your hash with one structural change. We went deep on this (adversarial design study over the full candidate space, then an implementation pass with its own review), so this reply is long because the evidence is, not because we disagree with much.

**Conceded, with stronger evidence than the review had.** The ledger id had to come out of the visible name — but the decisive reason turns out to be worse than redundancy: the embedded colon makes every bulk-imported blank node **unwritable in SPARQL** (BLANK_NODE_LABEL forbids `:`), which broke the documented stable-`_:fdb-` editing feature for imported data — our own docs admitted it at `update-where-delete-insert.md`. It is also why `canonical_split` carries the blank-node special case you found, why `validate` needs a lossy sanitizer, and why the Turtle exporter emitted syntactically invalid output for imported ledgers. Your positional-index critique also held: not reproducible under file-set changes, and the robustness gain from a hash is real (an index resolves only against an unchanged listing; a hash resolves against any candidate set). Your collision math checked out exactly. And the construction-time uniqueness check is implementable on every arm — each materializes its document set before the producer starts — so it's in.

**Countered on one point, with the code: dropping the ledger id entirely is a safety regression, not a neutral trade.** Cross-ledger blank-node unification is real and deliberate in the query layer: multi-ledger queries re-stamp bindings and compare by decoded IRI string with `ledger_alias` explicitly excluded (`binding.rs` — "the key for correct cross-ledger joins"), and blank nodes are doubly exposed because namespace code 10 is a fixed built-in, so their raw Sids are byte-identical across ledgers. Today's format unifies only when two ledgers share a *name* — impossible on one server. The unsalted hash unifies whenever two ledgers share a *relative path* (`data.ttl`, `part-00000.ttl`, the same dump imported into staging and prod) — same server, deterministic, silent, and the key captures no content identity, so "the same file" is really "the same path with any bytes behind it." The per-import uniqueness check structurally cannot see it: the second ledger is imported later. On semantics, RDF 1.1 §3.5 wants a globally unique skolem IRI *per blank node* — two separately-imported graphs hold two blank nodes even when the bytes match — so unification-by-default is a deviation rather than an upgrade. One correction on our side of the ledger-id argument: rename/clone staleness has no mechanism today (neither operation exists), but `branch` is real and gives you that point legitimately — a branched ledger carries ids salted with its source's name.

**What shipped** (on the branch now, full body updated):

- `fdb-d{base36₁₃(xxh64(normalized_ledger_id ‖ 0x00 ‖ doc_key))}-{label}` — your hash, salted with the normalized ledger id. Compact and fixed-width, no colon (the SPARQL round-trip is now a test: import, take a minted id, edit it via `_:fdb-…` in DELETE/INSERT), per-ledger reproducible under file-set changes, cross-ledger distinctness preserved, and the `d` marker kept so disjointness from the other `_:fdb-` minters stays structural rather than a length coincidence.
- **13 hash chars, not 8 — and not because bigger is safer.** 36⁸ (~2⁴¹) and any width below 13 truncate the u64, which means the fail-loud check can abort a legitimate import on a truncation collision that the hash itself never produced (at 8 chars, ~1-in-560 for a 100k-document import). 36¹³ > 2⁶⁴ is the principled stop: lossless, so the guard fires only on a genuine duplicate key or a true xxh64 collision.
- **Anonymous nodes got your reproducibility fix too.** They're the majority in real Turtle, and their `-b{t}-{N}` carried the *global* chunk ordinal — so file-set changes shifted them even under a perfect document key. The ordinal is now the sub-chunk index within the document (pinned by a test driving the same input at 1MB and 64-byte chunk sizes and asserting identical ids).
- **The ndjson arm is fixed rather than exempted.** Its labels were chunk-scoped (merging across lines within a chunk, splitting across chunks — wrong under every reading). Scoping is now per *context segment*: our reader defines lines as nodes sharing a document-level `@context`, so per-line would break intra-file `_:` references, but a mid-stream context switch starts a new document — which also keeps `cat a.jsonl b.jsonl` from merging the two files' `_:x`. The one residual (a file with an inline `@context` on every line still lands in one `@graph`, so *explicitly reused* labels unify) is documented with the workaround rather than silent.
- **An import manifest answers traceability completely**: `doc_key → relative path` recorded in txn-meta, one row per document and per ndjson segment. `d47` and a hash are equally opaque to a human; the manifest makes both resolvable, and it fixes the branch-salt invisibility above.
- **Your cross-ledger unification benefit exists as a deliberate choice**: `--skolem-namespace <id>` overrides the salt, so two ledgers built from one source can be made to mint identical ids on purpose — opt-in rather than accident.
- Your smaller notes all landed: every skolem base (the three serial paths and the JSON-LD arm included) now goes through the one documented helper; the body carries the ndjson scope story and the re-import-ids-differ note.

Two things considered and rejected, so they don't come back: **content hashing** (a byte-identical duplicate file — `a.ttl` plus its `.bak` — would silently merge two documents' blank nodes, and the streaming arms would need a full extra read pass before parsing could start), and **per-inline-context ndjson segmentation** (the key would vary within a chunk against a one-sink-one-base architecture; documented instead). One implementation note that affects your table: the Remote arm hashes `obj.address` relative to the listing prefix with a full-address fallback for `OrderedObjects` (which has no prefix), and `address` is documented backend-opaque — so re-importing the same objects through a different storage backend mints different ids; the caveat is at the resolution site.

Migration is clean for a reason worth stating: nothing in the codebase parses a minted label apart (the only structural reader strips `_:` and the `fdb-` prefix), and bulk import requires a fresh ledger, so no ledger can hold both formats. The `canonical_split` special case stays, with its comment updated to say it exists for pre-change ledgers and user-authored labels rather than for anything the import now mints.

bplatz (approved):
Approving. The scope question is settled, and settled better than I proposed it.

## The counter-argument lands

I'll take the correction on dropping the ledger id. The `binding.rs` evidence is decisive and I did not check it: the cross-ledger join key excludes `ledger_alias` by design, and blank nodes sit on the fixed built-in namespace code 10, so their raw Sids are byte-identical across ledgers. My format unifies only when two ledgers share a *name*, which cannot happen on one server; an unsalted hash unifies whenever two ledgers share a *relative path* — `data.ttl`, `part-00000.ttl`, the same dump into staging and prod. Same server, deterministic, silent, and structurally invisible to the construction-time check because the second import happens later. Keeping the ledger id as the hash namespace and making the unification opt-in via `--skolem-namespace` is the right shape: it preserves the RDF 1.1 §3.5 default and turns the thing I wanted into a lever rather than an accident.

Same on width. I proposed base36-8 reasoning only about hash collisions and missed that anything under 13 truncates the u64 — so the fail-loud check could abort a legitimate import on a *truncation* collision the hash never produced. 36^13 > 2^64 is the principled stop and I should have gotten there.

And the SPARQL-unwritability finding is a better argument than the one I made. I flagged the embedded colon as the reason `canonical_split` carries a special case; it was also silently breaking the documented stable-`_:fdb-` editing path for every imported ledger, which our own docs claimed worked.

## Verified locally

- `cargo test -p fluree-db-core --lib skolem::` — 11/11.
- `cargo test -p fluree-db-api --test grp_import --all-features` — **99 passed, 0 failed**, covering all 15 new `it_import_skolem` tests and the full pre-existing `it_import_ndjson` suite, which matters because the ndjson arm is where the scoping semantics actually changed.
- Read the diff end to end. `skolem.rs` is the sole minter (`DocChunk::skolem_base` delegates to it, ids are only produced at `DocIds::new`, the segment lookup, and the ndjson producer). `DocIds::new`'s uniqueness check is release-active and its message names both documents. The `canonical_split` comment now says it exists for pre-4.1.5 ledgers and user-authored labels. The `export`/`validate` split — verbatim on the round-tripping path, lossy only on the human-facing report, each stating why the other differs — is exactly right.

The test set maps onto the review point for point, which made it easy to check: `minted_ids_do_not_depend_on_the_rest_of_the_directory` is the reproducibility gap and it asserts on the *anonymous* id too, `duplicate_remote_addresses_abort_the_import` is the fail-loud check, `import_manifest_maps_blank_node_scopes_back_to_documents` is the traceability note, and `imported_blank_node_id_is_writable_in_sparql` is the round-trip that was quietly broken.

Three things you fixed that I did not ask for and should have: the anonymous-node global chunk ordinal (same reproducibility defect I raised for labeled nodes, one layer down), the ndjson arm being chunk-scoped rather than exempted, and — after the review — the `DocIds::id` out-of-range fallback, which was the one remaining path to a silent multi-document merge through the back door.

## Non-blocking

1. `SubChunkCounter`'s producer-contiguity guard is `#[cfg(debug_assertions)]`, so `seen` does not exist in release. An interleaving producer would re-issue sub-chunk 0 and silently merge two documents' anonymous nodes — the failure class this PR exists to close — while `DocIds::new` guards the analogous case unconditionally. I think debug-only is defensible for an internal producer invariant that no external input can perturb, but the two guards having different strength is worth one line saying why.

2. `skolem.rs` module-doc drift: the format block gives the hash as `xxh64(namespace ‖ 0x00 ‖ doc_key)`, omitting `‖ 0 ‖ segment`. The `NAMESPACE_SEPARATOR` doc a few lines below has it right, and that one is load-bearing (it is the re-cutting argument), so this is just the summary being stale.

3. Your reply describes the anonymous-node fix as "pinned by a test driving the same input at 1MB and 64-byte chunk sizes and asserting identical ids." What is actually pinned is serial-vs-parallel parity at one 64 KB size, and the PR body's residual — anonymous ids remain `chunk_size_mb`-dependent — is the accurate statement. The body is the record and it is honest, so nothing to change in the code; noting it only so the stronger claim does not get quoted later.

The residuals are all documented up front rather than left to be discovered, including the two I would have gone looking for (federated same-name servers, and `OrderedObjects` addresses being backend-opaque). Merge conflicts against main are yours to resolve; note that no CI has run on this branch yet, so the rebase is the first real gate.

aaj3f:
Rebased onto main at 94595c7d3 — the first real CI gate you called out is running now. The conflict story, since this branch merges under your approval and the resolutions deserve daylight:

**Two textual conflicts, both append-vs-append with zero shared semantics**: `ImportConfig` gained both mains' `virtual_source` and this branch's `skolem_namespace` (neither reads the other), and the CLI integration test file gained both the bm25 and skolem sections — with both halves verified to *execute*, not merely compile, since a mis-stitch there would have silently swallowed a section.

**The #1552 protocol seam needed no manual reconciliation, verified rather than assumed**: this branch edits the intern side of `ImportSink` (skolem base, sub-chunk, `term_blank`) while #1552 changed the emit side (fallible `emit_*`, `finish` → `into_parts`) — disjoint regions, auto-merged, with the deferred-error contract confirmed intact at the new head (stash-first-failure at the emit sites, surfaced in `into_parts`).

**Two judgment calls on main-side code you never saw, disclosed**: main's materialize path calls the widened `ImportSink` constructors, so those sites pass `sub_chunk = 0` — behavior-preserving because materialize's skolem base already embeds `t` (per-chunk unique) and materialize only ever mints *labeled* blank nodes (`term_blank(Some(..))` is its sole call site), so the anonymous path this parameter separates never fires there. And main's new `ImportSource::Virtual` arm gets an empty `DocIds` — a virtual source scans a mapped table, not documents, so zero manifest rows is the honest value, and the `DocIds::id` debug-assert from the review rider now polices that path too. Both covered by the 22 materialize tests plus your 99/99 `grp_import`, which holds at the new head (full api suite 3037/3037).

**Your three notes, all landed**: the sub-chunk guard now states *why* it is debug-only while `DocIds::new` is unconditional (caller-supplied input vs internal producer invariant — your framing); the `skolem.rs` format summary now carries `‖ 0x00 ‖ segment`, synced with the load-bearing `NAMESPACE_SEPARATOR` doc; and on your third note — you're right, and the correction is accepted as stated: my reply overclaimed the anonymous-node pinning. What the chunk-size sweep test pins is the *segment ordinal*; anonymous-id coverage is serial-vs-parallel parity at one size, and the PR body's residual ("anonymous ids remain `chunk_size_mb`-dependent") is the accurate record. Thanks for flagging it before the stronger claim got quoted anywhere else.

One rider disclosed since it rode the rebase push: the `ImportSink` module-doc example was one argument short of the widened constructor (pre-existing on this branch, inside an `ignore` fence so nothing caught it) — fixed as a one-token edit, signature-verified.

Issue #1489, opened 2026-07-14
flaky: binary-index reload race in `values_bound_type_after_delete_incremental_index_timeline` under parallel grp_misc (`dict branch: too small for header`)


aaj3f:
## Symptom

`it_values_type_deleted_repro::values_bound_type_after_delete_incremental_index_timeline` (bundled into `fluree-db-api/tests/grp_misc.rs` via `#[path]`, source at `fluree-db-api/tests/it_values_type_deleted_repro.rs:202`) fails roughly **1 in 6** thread-parallel `cargo test --test grp_misc` runs with:

```
Internal("failed to load binary index: dict branch: too small for header")
```

It passes reliably single-threaded (`--test-threads=1`) and was not observed failing under nextest's process-per-test execution during the same gate batteries.

## Shape of the failing timeline

The test drives the production-like background-indexer sequence: insert across two commits → `build_and_publish_index` → reload ledger → wipe → incremental `build_and_publish_index` folding the retractions → reload → re-insert → index → reload, asserting VALUES/type shapes agree at each step. The failure is on a **reload** step: the freshly published binary index's dict branch fails the header-size check when read back, which points at a reload race (reader observing a partially-written or stale dict-branch artifact) rather than anything in the test's own assertions — i.e. a `fluree-db-core` binary-index load/publish race surfaced by CPU contention, not a query bug.

## Discovery context

Surfaced during the perf-series gate batteries (PR #1486 body records it as an explicitly out-of-scope, pre-existing flake — that PR only moves an unrelated test between binaries and touches no engine code). The failure rate was measured while `grp_misc`'s ~240 bundled tests ran thread-parallel in one process, so timing pressure from sibling tests appears to be the trigger.

## Suggested next steps for the binary-index owner

- Reproduce with `cargo test -p fluree-db-api --test grp_misc values_bound_type_after_delete_incremental_index_timeline` looped under parallel sibling load (or the full `grp_misc` repeated ~10×).
- Audit the publish→reload boundary in the binary-index dict-branch writer/loader for a missing ordering guarantee (rename-before-flush, non-atomic publish of the branch file, or a cached header length read against a superseded artifact).

aaj3f:
Sweep note (claude-assisted): I could not reproduce this at `47e96b7b6` — 25× thread-parallel `grp_misc` runs (262 tests each) on a deliberately loaded box, zero failures, against the filed ~1-in-6 rate (p≈0.01 that we'd see zero if the rate still held).

[...]

PR #1552, opened 2026-07-29
feat(graph-ir): fallible GraphSink protocol, quad capability, statement lifecycle + Turtle ParserOptions conformance

aaj3f:
# PR-0: GraphSink protocol + Turtle ParserOptions

Foundation PR for the `fluree rdf` (riot-analog) effort. It lands nothing
user-visible: it changes the shared parser→sink abstraction so the writers,
converters and diagnostics of the later PRs are expressible at all, and it
makes the Turtle parser able to produce conformant RDF as well as the ingest
shape. Blast radius is every ingest path, so the gate is the full transact +
`fluree-db-api` suites, not the light crates alone.

Implements the locked decisions H-1 (PR-0 split), H-2 (ParserOptions on the
shared parser, ingest pinned to current behavior), and §1.2 of
`riot-analog-pr-plan.md` (fallible emission, `finish()`, quad capability
probe, term lifetime).

## A. GraphSink protocol

**Fallible emission.** `emit_triple` / `emit_list_item` / `emit_reified_triple`
now return `Result<(), SinkError>`, and the new `emit_quad` does too. `term_*`
stay infallible — they intern a term and hand back an id; a writer only
touches its output on emit.

The point is early termination. Today a sink cannot report that its downstream
died, so `convert big.ttl | head -5` would parse to EOF against a closed pipe.
`SinkError` carries an `io::Error` or a rejection message and answers
`is_broken_pipe()`, so a CLI can exit quietly on the expected shutdown and
loudly on a real failure. The Turtle parser stops at the first refusal and
preserves the sink's error as `TurtleError::Sink` rather than flattening it
into a generic parse error; the JSON-LD adapter does the same via
`AdapterError::Sink`.

**Quad capability.** `supports_quads()` + `emit_quad()`, mirroring the existing
`supports_reified_triples` probe. No producer emits quads in this PR — this is
the protocol landing, so M2's TriG/N-Quads readers have something to refuse
against.

The default `emit_quad` body **errors** rather than no-ops. The asymmetry with
`emit_reified_triple` (whose default stays a no-op) is deliberate: every
producer of reified-triple events already refuses the input up-front on the
capability probe, so that default body is unreachable, not a fallback. Nothing
guards quads yet, so the default has to be the guard. Silently folding named
graphs into the default graph is the exact defect the JSON-LD adapter has
today; this is what stops it being repeated.

**Statement lifecycle.** `end_statement()`, called by the Turtle parser at each
statement terminator and only for statements that parsed and emitted cleanly.
It serves two later features that need the same boundary: term lifetime (below)
and the statement-scoped output buffering `--continue-on-error` requires, since
triples are emitted during descent, before the terminating `.` proves the
statement valid. A failed statement never reaches the call, so it contributes
nothing.

**`finish()` and the naming split.** The trait gains
`finish(&mut self) -> Result<(), SinkError>` — flush and finalize, called by
whoever owns the sink, never by a producer (one sink can be fed by many
producer calls, a chunk each on the bulk-import path).

Three sinks already had an inherent `finish()` that *consumed* the sink and
returned its product. Leaving those named `finish()` would have type-checked
while letting a concrete-typed caller silently skip the protocol finish —
survivable for a graph collector, not for a writer. They are renamed to
consuming `into_*` methods, so `finish()` always means "flush" and never "give
me your contents":

| sink | was | now |
| --- | --- | --- |
| `GraphCollectorSink` | `finish() -> Graph` | `into_graph()` |
| `FlakeSink` | `finish() -> Result<Vec<Flake>, _>` | `into_flakes()` |
| `ImportSink` | `finish() -> Result<(writer, prefixes, spool), _>` | `into_parts()` |

Deferred-error semantics are unchanged: `FlakeSink`/`ImportSink` `emit_*` still
record the first invariant violation and keep going, and the consuming method
is still where it becomes a hard error.

**Term lifetime.** `TermId`s now have a documented split. IRI and blank ids are
session-scoped (producers cache them across statements); literal ids are
statement-scoped, because literals are minted per occurrence and never
deduplicated — which is why the term table grew by one live entry per literal
*occurrence* for the length of the document.

`GraphCollectorSink` acts on it: `end_statement` retires the statement's
literal slots for reuse, so the high-water mark is the widest single statement
rather than the whole document (1000 statements × 2 literals leaves 4 slots,
not 2002). Soundness rests on two facts, both pinned by tests: `emit_*` clones
into the graph before returning, so overwriting a slot cannot reach an
already-collected triple; and the parser caches only IRI and prefixed-name ids
across statements.

**This is groundwork, not a shipped win — no production path benefits today.**
`FlakeSink` and `ImportSink`, the sinks on the ingest hot path, ignore
`end_statement` entirely and are deliberately untouched. `GraphCollectorSink`'s
users are the W3C/SHACL testsuite harnesses, the R2RML loader, and
`parse_to_json` — none of them memory-bound. Producers that never delimit
statements, such as the JSON-LD adapter, never recycle at all and keep today's
allocation behavior exactly. The reason to land it here is that the memory
profile of a *writer* sink is the differentiator the effort is being built on,
and the protocol hook plus a proven-sound reference implementation is what a
writer needs to exist. Applying it to the ingest sinks is a separate,
separately-benchmarked change.

**Statement rollback.** `abort_statement()` is the counterpart to
`end_statement()`, and lands now so `--continue-on-error` does not force a
second breaking trait change later. The parser emits during descent — property
lists and collections push triples before the terminating `.` proves the
statement well-formed — so a rejected statement used to leave partial triples
in the sink; riot's "a rejected statement contributes nothing" held only in the
error return, not in the data. The parser calls it only when a statement fails
having already emitted, and exactly one of `end_statement`/`abort_statement`
ends any statement. `GraphCollectorSink` rewinds the graph to a mark taken at
the statement boundary.

Writing that test found a real bug in the first cut: the parser keeps one token
of lookahead, so consuming a statement's `.` lexes the first token of the
*next* statement, and a lexical error there was reported while the finished
statement was still in flight — the rollback then discarded a statement that
was complete and valid. Committing at the terminator, before the lookahead
advance, fixes it; the regression is pinned and mutation-verified.

**Blank-node namespace.** `GraphCollectorSink` minted anonymous blanks as
`b{N}`, the exact shape a document writes by hand, so a file containing `_:b1`
plus any anonymous node produced one node where RDF says two — silently merging
user data into a parser-minted node. Pre-existing via `[ … ]` and reifiers, but
`CollectionStyle::Spine` makes it systematic, since every collection mints
spine nodes. Fixed the way `FlakeSink`/`ImportSink` already do it: mint `-b{N}`,
which cannot lex as a `BLANK_NODE_LABEL` and so is unreachable from any input.
The tests count distinct identities rather than checking isomorphism, because
merged nodes are isomorphic to themselves and an isomorphism check
structurally cannot see this class of bug.

## B. Turtle ParserOptions

The parser canonicalizes in three places that are right for ingest and wrong
for conversion:

- an object-position collection emits flat indexed list items, so the three
  triples W3C defines for `( ex:o )` arrive as one event;
- an object-position `()` emits **nothing** — a silently lost triple, since
  `()` denotes `rdf:nil`;
- `Integer`/`Double` tokens are parsed to native values, so `+1`, `01`, `1e0`,
  `1.0E0` are unrepresentable (`Decimal` and i64-overflowing integers already
  keep their lexical form).

Meanwhile subject-position collections have always emitted a proper spine, so
the two positions disagreed with each other.

`ParserOptions { collections, numerics }` gets both shapes out of one grammar —
one conformance surface, no fork. `ParserOptions::default()` is today's
behavior in every field, so every existing caller is unchanged;
`ParserOptions::conformant()` is the faithful-RDF preset. New entry points
`parse_with_options` and `parse_with_prefixes_base_options` (the latter is the
full one — every other `parse*` delegates to it with something defaulted, so
the chunked reader paths can opt in too).

`CollectionStyle::Spine` needed no new collection code: it drops the
object-list special cases, so `( … )` and `()` fall through to the ordinary
object path, which already builds a spine and already resolves `()` to
`rdf:nil`. Both positions now share one code path.

One consequence worth naming for review: a spine collection has a single object
term (the head), so the RDF 1.2 annotation-on-collection deferral — which
exists *only* because indexed items leave nothing to reify — does not apply in
this mode. The deferral is unchanged in the default mode, and both halves are
pinned by a test.

`NumericStyle::PreserveLexical` routes numeric tokens through the typed-string
lane `Decimal` and `IntegerOverflow` already use. **Deviation from the brief,
flagged for review:** the brief anticipated a new `LiteralValue`/`Datatype`
variant carrying the lexical string. That turned out to be unnecessary — the
existing typed-string lane already represents "this datatype, this exact
lexical", and is the pattern the brief also said to extend. So
`fluree-graph-ir` gains nothing for this feature, the ingest path hits the
identical branch it hits today, and there is no second numeric representation
to keep consistent. Downstream conversion is unaffected:
`convert_string_literal` already parses `"42"^^xsd:integer` back to a native
value (pinned by the pre-existing `test_typed_string_integer`).

## Testing

New tests, beyond the existing suites staying green:

- **Early termination** — a sink that refuses after N triples: the parse stops
  there (emits exactly N), the sink's error survives as `TurtleError::Sink`
  with `is_broken_pipe()` intact, and the failed statement is never terminated.
- **Lifecycle** — `end_statement` fires once per completed statement,
  directives included; a multi-triple predicate-object list is one statement;
  the parser never calls the protocol `finish()`, and two parses can feed one
  sink.
- **Quad refusal** — triple-only sinks report `supports_quads() == false`, and
  the default `emit_quad` body returns an error instead of dropping the graph
  name.
- **Recycling** — literal slots are reused after a statement ends, IRI and
  blank ids are not, an already-emitted triple is unaffected by the reuse, and
  the term table tracks the widest statement rather than the document; plus an
  end-to-end Turtle guard that a later statement reusing an earlier slot cannot
  corrupt the earlier values.
- **Collections** — W3C-shaped spine assertions that *walk* the
  `rdf:first`/`rdf:rest` chain rather than counting triples; one-item,
  two-item, nested and nested-empty collections; `()` in both positions;
  subject and object positions agreeing; annotations admitted in spine mode and
  still refused in the default.
- **Numerics** — seven lexical forms preserved with correct datatypes;
  canonicalization still the default; decimals and big integers identical under
  both modes; non-numeric literals untouched by the knob.
- **Default-mode pinning** — one document covering every construct the options
  touch, asserted triple-for-triple in the default mode. This is the regression
  guard for "existing callers are unchanged". The pre-existing
  `test_collection`, `test_empty_collection` and `test_integer_literal`
  assertions are unmodified and still pass.

### The recycling differential

Slot recycling is the one change here whose failure mode is *aliasing*: a wrong
answer is a perfectly well-formed graph with one term silently substituted for
another. Per-triple expectations written by the author of the recycling are
unlikely to anticipate that, so `fluree-graph-turtle/tests/adversarial_recycling.rs`
asserts against an oracle instead — the same document parsed into a recycling
sink and into one whose `end_statement` is swallowed by a forwarding wrapper,
i.e. today's behavior. Any `TermId` outliving its statement while still
referenced shows up as a divergence.

Two things make that evidence rather than ceremony:

- **Mutation-checked, with the limits stated.** Recycling IRI slots — which the
  parser caches across statements — is detected by **2 of the 9 tests in the
  file through output comparison alone, and by 9 of 9 once the retirement
  assert is present**; the 7 additional detections come from the assert
  tripping on the polluted slot list, not from any graph diverging. Both
  figures were measured by applying the mutation with the assert enabled and
  disabled.
- A *second* mutation, always reusing literal slot 0, escapes this file
  entirely and is caught only by the `fluree-graph-ir` unit tests. The reason
  is worth knowing: no literal is ever minted while another literal id is live
  (the parser emits each literal immediately after minting it), so
  slot-0-collapse is observationally equivalent here. The differential's power
  comes from the aliasing dimension, not from statement width. Both facts are
  recorded in the module docs so nobody trims the corpus believing the cases
  are redundant.
- **Corpus-scale.** Run against the real W3C `rdf-tests` Turtle corpus —
  **1148 files, 718 parsed, 430 rejected, across all three option modes
  (default, conformant, spine-only) — zero output divergence from the
  no-recycle oracle and zero acceptance divergence**, with two temporary
  probes live in `end_statement`: a poison canary overwriting every retired
  slot with a sentinel, and an assertion that every retiring slot still holds a
  `Literal`. Neither probe ships; both are review instrumentation.

The corpus harness itself is **not** in this PR: it hardcodes an absolute path
into the `rdf-tests` submodule. Its home is the `testsuite-rdf` harness of the
plan's §2, which is where this differential should be re-landed so the result
keeps being enforced rather than being a one-off measurement.

## Gates

- `cargo nextest run -p fluree-graph-ir -p fluree-graph-turtle -p fluree-graph-json-ld -p fluree-graph-format -p fluree-db-transact` — 730 passed, 0 failed.
- `cargo nextest run -p fluree-db-api` (full, grouped `grp_*` bins, no per-file
  filters) — 2919 passed, 0 failed, 11 skipped.
- `cargo clippy --all-targets --no-deps` on every touched crate — clean
  (workspace deny-lints, incl. `uninlined_format_args`).
- `cargo fmt --all --check` — clean.
- `cargo check --all-targets` on the excluded `testsuite-sparql` and
  `testsuite-shacl` workspaces — clean. They call the renamed
  `GraphCollectorSink::finish` and would otherwise break in their own CI job,
  which the workspace gates do not cover.

## Review follow-ups deferred to M1

Raised in review, agreed as non-blocking, recorded here so they are not lost:

- **Record that the two lifecycle guards do not overlap.** The two mutations
  on `committed_current` are caught by disjoint sets of tests: dropping the
  *guard* (aborting a statement that already committed) is caught only by
  `lexical_error_in_lookahead_does_not_abort_the_committed_statement`, while
  dropping the *per-statement reset* is caught only by
  `failed_statement_contributes_nothing` and the two
  `abort_still_fires_after_a_committed_statement_*` cases. Neither side
  subsumes the other, and every one of them looks redundant beside its
  neighbours until you actually run the mutations. The test file currently
  explains the latch hazard but not this mapping; it wants one comment naming
  which test covers which mutation, so nobody trims either side.
- **Upstream the rest of the reviewer's probe.** The two
  after-a-committed-statement rollback cases are already shipped, with
  different fixtures — the reviewer reaches the error-at-dot path with a
  trailing `,` where this PR uses a wrong token in the terminator position.
  What is *not* covered here is their
  `exactly_one_terminator_per_statement_everywhere` sweep, which asserts the
  invariant across bracket, collection, bare-bnode, subject-collection,
  SPARQL-prefix and nested-list shapes in a single pass. That sweep is the
  valuable remainder.
- **Promote the sharp invariant phrasing into the parser.** The precise form
  is "no literal is minted while another literal id is live"; the comment at
  `parse_annotation_tail` still leads with the looser "at most one literal id
  is live at a time", which is not literally true (`parse_reified_triple`
  holds a literal object while minting a reifier — the reifier grammar just
  guarantees that term is never a literal). The sharp form is already in the
  test module docs; the parser comment should lead with it too.

## Not in this PR

No quad producer, no writer sink, no CLI surface, no `--continue-on-error` —
those are the protocol's first consumers, in M1/M2. `FlakeSink`/`ImportSink` do
not recycle literal slots and do not implement `abort_statement`; both are
hot-path changes that want their own benchmarking. The JSON-LD adapter has no
statement boundaries, so it neither recycles nor rolls back, and it still drops
`@graph` contents — its module docs now say so accurately (they previously
claimed "flattened into the default graph", which is not what the code does),
and fixing it needs the quad protocol this PR lands.

### DIFF
--- a/fluree-db-api/src/inline_ontology.rs
+++ b/fluree-db-api/src/inline_ontology.rs
@@ -55,7 +55,7 @@ pub(crate) fn parse_inline_ontology_to_bundle(
     })?;
 
     let flakes: Vec<Flake> = sink
-        .finish()
+        .into_flakes()
         .map_err(|e| crate::ApiError::config(format!("inline ontology flake build failed: {e}")))?;
     if flakes.is_empty() {
         return Ok(None);
--- a/fluree-db-api/src/inline_shapes.rs
+++ b/fluree-db-api/src/inline_shapes.rs
@@ -79,7 +79,7 @@ pub(crate) fn parse_inline_shapes_turtle_to_bundle(
 }
 
 fn bundle_from_sink(sink: FlakeSink<'_>) -> Result<Option<Arc<SchemaBundleFlakes>>, TransactError> {
-    let flakes: Vec<Flake> = sink.finish()?;
+    let flakes: Vec<Flake> = sink.into_flakes()?;
     if flakes.is_empty() {
         return Ok(None);
     }
--- a/fluree-db-api/src/tx.rs
+++ b/fluree-db-api/src/tx.rs
@@ -3021,7 +3021,7 @@ impl crate::Fluree {
             let _g = parse_span.enter();
             let mut sink = FlakeSink::new(&mut ns_registry, new_t, txn_id);
             fluree_graph_turtle::parse(turtle, &mut sink)?;
-            sink.finish().map_err(ApiError::from)?
+            sink.into_flakes().map_err(ApiError::from)?
         };
         tracing::info!(flake_count = flakes.len(), "turtle parsed to flakes");
 
--- a/fluree-db-api/src/validate.rs
+++ b/fluree-db-api/src/validate.rs
@@ -703,7 +703,7 @@ mod tests {
         let turtle = sample_report().to_turtle();
         let mut sink = fluree_graph_ir::GraphCollectorSink::new();
         fluree_graph_turtle::parse(&turtle, &mut sink).expect("report Turtle must parse");
-        let graph = sink.finish();
+        let graph = sink.into_graph();
         // report node (type + conforms + result) and the result node's fields
         assert!(graph.len() >= 8, "expected full report triples: {turtle}");
         assert!(turtle.contains("sh:MinCountConstraintComponent"));
--- a/fluree-db-r2rml/src/loader/extractor.rs
+++ b/fluree-db-r2rml/src/loader/extractor.rs
@@ -500,7 +500,7 @@ mod tests {
     fn parse_r2rml(turtle: &str) -> Graph {
         let mut sink = GraphCollectorSink::new();
         parse_turtle(turtle, &mut sink).unwrap();
-        sink.finish()
+        sink.into_graph()
     }
 
     #[test]
--- a/fluree-db-r2rml/src/loader/mod.rs
+++ b/fluree-db-r2rml/src/loader/mod.rs
@@ -52,7 +52,7 @@ impl R2rmlLoader {
     pub fn from_turtle(content: &str) -> R2rmlResult<Self> {
         let mut sink = GraphCollectorSink::new();
         parse_turtle(content, &mut sink).map_err(|e| R2rmlError::Parse(e.to_string()))?;
-        let graph = sink.finish();
+        let graph = sink.into_graph();
         Ok(Self { graph })
     }
 
--- a/fluree-db-transact/src/flake_sink.rs
+++ b/fluree-db-transact/src/flake_sink.rs
@@ -11,7 +11,7 @@ use crate::value_convert::{convert_native_literal, convert_string_literal};
 use fluree_db_core::edge::EdgeKey;
 use fluree_db_core::DatatypeConstraint;
 use fluree_db_core::{Flake, FlakeMeta, FlakeValue, Sid};
-use fluree_graph_ir::{Datatype, GraphSink, LiteralValue, TermId};
+use fluree_graph_ir::{Datatype, GraphSink, LiteralValue, SinkResult, TermId};
 use std::collections::HashMap;
 use std::sync::Arc;
 
@@ -48,7 +48,7 @@ enum ResolvedTerm {
 /// let mut ns = NamespaceRegistry::from_db(&db);
 /// let mut sink = FlakeSink::new(&mut ns, new_t, txn_id);
 /// fluree_graph_turtle::parse(ttl, &mut sink)?;
-/// let flakes = sink.finish().expect("no invariant violation");
+/// let flakes = sink.into_flakes().expect("no invariant violation");
 /// ```
 pub struct FlakeSink<'a> {
     /// Resolved terms indexed by TermId
@@ -93,9 +93,15 @@ impl<'a> FlakeSink<'a> {
 
     /// Consume the sink and return the accumulated flakes, or surface the
     /// first storage-invariant violation observed during parsing as a hard
-    /// error. Mirrors [`ImportSink::finish`](crate::import_sink::ImportSink)
+    /// error. Mirrors
+    /// [`ImportSink::into_parts`](crate::import_sink::ImportSink::into_parts)
     /// — bad input must fail the transaction, not silently omit a triple.
-    pub fn finish(self) -> Result<Vec<Flake>, TransactError> {
+    ///
+    /// Distinct from the protocol's [`GraphSink::finish`] (flush/finalize):
+    /// this is the sink's *product*. Deferred-error semantics are unchanged
+    /// by the fallible protocol — `emit_*` still records the first violation
+    /// and keeps going, and this method is where it becomes a hard error.
+    pub fn into_flakes(self) -> Result<Vec<Flake>, TransactError> {
         if let Some(err) = self.invariant_error {
             return Err(err);
         }
@@ -260,16 +266,24 @@ impl GraphSink for FlakeSink<'_> {
         })
     }
 
-    fn emit_triple(&mut self, subject: TermId, predicate: TermId, object: TermId) {
+    fn emit_triple(&mut self, subject: TermId, predicate: TermId, object: TermId) -> SinkResult {
         if let Some(flake) = self.build_flake(subject, predicate, object, None) {
             self.flakes.push(flake);
         }
+        Ok(())
     }
 
-    fn emit_list_item(&mut self, subject: TermId, predicate: TermId, object: TermId, index: i32) {
+    fn emit_list_item(
+        &mut self,
+        subject: TermId,
+        predicate: TermId,
+        object: TermId,
+        index: i32,
+    ) -> SinkResult {
         if let Some(flake) = self.build_flake(subject, predicate, object, Some(index)) {
             self.flakes.push(flake);
         }
+        Ok(())
     }
 
     fn supports_reified_triples(&self) -> bool {
@@ -287,18 +301,18 @@ impl GraphSink for FlakeSink<'_> {
         predicate: TermId,
         object: TermId,
         reifier: TermId,
-    ) {
+    ) -> SinkResult {
         let Some(s) = self.resolve_sid(subject) else {
-            return;
+            return Ok(());
         };
         let Some(p) = self.resolve_sid(predicate) else {
-            return;
+            return Ok(());
         };
         let Some((o, dtc)) = self.resolve_object(object) else {
-            return;
+            return Ok(());
         };
         let Some(ann) = self.resolve_sid(reifier) else {
-            return;
+            return Ok(());
         };
 
         match crate::generate::flakes::reified_triple_bundle(s, p, o, &dtc, &ann, self.t) {
@@ -310,6 +324,7 @@ impl GraphSink for FlakeSink<'_> {
                 }
             }
         }
+        Ok(())
     }
 }
 
@@ -337,9 +352,9 @@ mod tests {
         let s = sink.term_iri("http://example.org/alice");
         let p = sink.term_iri("http://example.org/name");
         let o = sink.term_literal("Alice", Datatype::xsd_string(), None);
-        sink.emit_triple(s, p, o);
+        sink.emit_triple(s, p, o).unwrap();
 
-        let flakes = sink.finish().expect("no invariant violation");
+        let flakes = sink.into_flakes().expect("no invariant violation");
         assert_eq!(flakes.len(), 1);
         let f = &flakes[0];
         assert!(f.op); // assertion
@@ -356,9 +371,9 @@ mod tests {
         let s = sink.term_iri("http://example.org/alice");
         let p = sink.term_iri("http://example.org/age");
         let o = sink.term_literal_value(LiteralValue::Integer(30), Datatype::xsd_integer());
-        sink.emit_triple(s, p, o);
+        sink.emit_triple(s, p, o).unwrap();
 
-        let flakes = sink.finish().expect("no invariant violation");
+        let flakes = sink.into_flakes().expect("no invariant violation");
         assert_eq!(flakes.len(), 1);
         assert!(matches!(&flakes[0].o, FlakeValue::Long(30)));
     }
@@ -371,9 +386,9 @@ mod tests {
         let s = sink.term_iri("http://example.org/x");
         let p = sink.term_iri("http://example.org/val");
         let o = sink.term_literal_value(LiteralValue::Double(3.13), Datatype::xsd_double());
-        sink.emit_triple(s, p, o);
+        sink.emit_triple(s, p, o).unwrap();
 
-        let flakes = sink.finish().expect("no invariant violation");
+        let flakes = sink.into_flakes().expect("no invariant violation");
         assert_eq!(flakes.len(), 1);
         assert!(matches!(&flakes[0].o, FlakeValue::Double(d) if (*d - 3.13).abs() < f64::EPSILON));
     }
@@ -386,9 +401,9 @@ mod tests {
         let s = sink.term_iri("http://example.org/x");
         let p = sink.term_iri("http://example.org/active");
         let o = sink.term_literal_value(LiteralValue::Boolean(true), Datatype::xsd_boolean());
-        sink.emit_triple(s, p, o);
+        sink.emit_triple(s, p, o).unwrap();
 
-        let flakes = sink.finish().expect("no invariant violation");
+        let flakes = sink.into_flakes().expect("no invariant violation");
         assert_eq!(flakes.len(), 1);
         assert!(matches!(&flakes[0].o, FlakeValue::Boolean(true)));
     }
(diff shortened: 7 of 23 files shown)


aaj3f:
CI note: the `test` job failure is the known flake #1489 (binary-index reload race in `values_bound_type_after_delete_incremental_index_timeline` under parallel `grp_misc` — filed 2026-07-13, pre-dates this branch). The test passes in 2.6s on a local merge of this head (869703009) with current main (5598ffd6a). Rerunning the job.

bplatz (approved):
Approving as-is, no changes requested — the protocol design is sound and the adversarial recycling suite is unusually good.

One thing to settle before M1, not in this PR: I think the tool should be **`fluree convert`**, not `fluree rdf`.

- `fluree export` already writes Turtle / N-Triples / N-Quads / TriG / JSON-LD, so "RDF" is not what distinguishes the new tool. The real axis is that it is file → file with no ledger in the loop — which `rdf` does not say, and which `export` will keep shadowing in a user's head.
- Our data-flow commands are verbs (`export`, `load`, `insert`, `query`, `validate`); the nouns are reserved for management groups with sub-verbs under them (`graph`, `branch`, `remote`). A format converter is a data-flow operation, so a top-level `rdf` noun sits against that grain.
- CSV already reaches this same `GraphSink` path via `csv_import.rs` → JSON-LD, and we import Cypher dumps too. A reader for either lands next to this machinery sooner or later, and `fluree rdf convert --from csv` argues with itself.
- `convert` costs nothing if the scope stays RDF-only and needs no rename if it does not. A `--help` line pointing at riot / rapper / rdfpipe keeps the legibility either way.

Nothing here blocks this PR — it ships no CLI surface, which is exactly why now is the cheapest moment to settle the name.

aaj3f:
Thanks for this — and agreed that now, pre-merge, is the right moment to settle it. Two of your points land cleanly: `export` already writes all of these formats, so "RDF" is genuinely not the differentiator (file→file with no ledger in the loop is), and the `--help` pointer at riot/rapper/rdfpipe is a good idea we'll take regardless of the name.

The complexity that I think changes the calculus: `convert` is the flagship verb but not the scope. The locked plan's later milestone adds `check`, `count`, `canon`, `infer`, and a file-mode `query` as siblings — a riot/rapper-class toolkit, not a single converter. Under the top-level-verb grammar those become `fluree check` (directly beside the *existing* `fluree validate`, which is SHACL-against-ledger — two adjacent top-level verbs both meaning "tell me if this is OK" with different targets), `fluree count`, `fluree canon`. The naming problem doesn't go away with the rename; it comes back one milestone later, worse.

Which is also why I'd read your own grammar observation as supporting the group form: the nouns in our CLI are groups-with-sub-verbs (`branch`, `graph`, `remote`), and `rdf` is exactly that — a tool group whose sub-verbs are data-flow operations on files. The scope line also answers the CSV case: `rdf` scopes to RDF *syntaxes*, and CSV→RDF is a mapping decision (header→predicate, R2RML territory), not a syntax conversion — so it lands next to `import`/the mapping machinery, and `fluree rdf convert --from csv` never needs to exist.

One correction to the cost model, in fairness to both sides: the CLI surface does already exist in the stacked PRs above this one (#1555/#1569/#1578 ship `fluree rdf convert` — command registration, ~90 test invocations, docs, and the bench harness's recorded invocations). Nothing is merged, so a rename is still a mechanical sweep rather than a breaking change — but it's "one focused sweep across three open PRs," not "costs nothing."

The trade-offs as I see them:

- **Keep `fluree rdf <verb>`** — one canonical name, the M5 verbs have a home, and `fluree rdf` reads as "the RDF toolkit" to the exact audience we're courting away from `riot`. Cost: `convert` alone sits noun-side of your grammar.
- **Rename to `fluree convert`** — best verb grammar for the flagship. Cost: the sweep, plus re-solving `check`/`canon`/`count` placement next milestone (top-level collides with `validate`; a `convert`-anchored group has no name for them).
- **Middle path: keep the group, add top-level `fluree convert` as a visible alias for `rdf convert`** — your discoverability point conceded for the one verb users will actually reach for first, at ~30 lines and no test churn. Cost: two spellings of one command in docs and support.

Our lean: keep `fluree rdf <verb>` as canonical for the toolkit-scope and collision reasons above — and if you think the verb-grammar/discoverability case for the flagship is strong enough, we'd happily add the top-level `convert` alias at M1 as the settlement. Genuinely open on that last part; it's cheap and it's your point that motivates it.

aaj3f:
Following up on the naming thread with a decision, because you asked the right question ("if there is some precedence that is a 'norm' here, that is an argument") and we went and looked. We put a fresh-eyes agent on it with no stake in the existing design, a survey mandate, and both of our argument sets as inputs to weigh rather than positions to defend.

**The precedent is on your side, and it isn't close.** No RDF tool anywhere — library or database — puts file→file conversion behind a grouping noun. Oxigraph, the closest analog to our situation, ships `convert` as a bare top-level verb sitting next to `load`, `query`, and `serve`; RDF4J's console does the same; Jena, rapper, serd, and rdfpipe are standalone flat binaries. The only RDF database that made a group at all is Stardog, and facing this exact choice they named it `file`, not `rdf` — your point, made by a vendor. On the "same verb, file mode" pattern, git is the strong precedent: `git diff --no-index <path> <path>` is the documented way to diff two filesystem paths with no repo, same verb, no group.

**Two facts in our own CLI settled the rest.** First, `fluree validate` already takes a file — `docs/cli/validate.md` documents "Two modes," and file mode runs with no ledger at all. So the top-level-verb-that-reads-either shape is something we've already shipped, and `check` had to be renamed under *any* grouping, because `rdf check data.ttl` beside `validate data.ttl` differs on nothing a user can see from the names. Second, every noun group we have — `branch`, `graph`, `remote`, `prefix`, `docs` — names something you can list. `rdf` names none, which is your "RDF what?" objection stated structurally.

**And a correction in your favor on the one point we pushed back on:** we claimed CSV→RDF was mapping territory that would never land next to this machinery. Wrong on the facts — `fluree create --from` already ingests `.csv` (neo4j-admin header convention) and `.cypher` scripts, no R2RML in the path. The reader you predicted is already there, so `fluree convert data.csv --to nquads` costs nearly nothing conceptually, and `fluree rdf convert --from csv` really would have argued with itself.

**What we adopted** (landed at the top of the stack, so the merged result is flat):

- `fluree convert` — the literal Oxigraph name. One help line each way distinguishes it from `export` (convert reads a file; export writes a ledger out).
- `fluree parse` — was `check`, renamed for the `validate` collision above; differentiates on the axis that actually differs (grammar vs shapes), and both words are terms of art the riot audience already has. `check` survives as a hidden alias. We specifically rejected `lint` (connotes style-on-top-of-well-formedness, over-promises for a pass/fail syntax check).
- `fluree count` — as is.
- Future verbs stay flat: `canon` and `infer` collide with nothing (and `infer` stays open to a natural `fluree infer <ledger>` growth a group would have walled off). File-query becomes **`fluree query --from <FILE>`** — the git `--no-index` pattern, using the flag this CLI already means "source data" by (`create --from`, `load --from`), which also hands file-query the entire existing output-format/explain/bench machinery instead of a parallel implementation.
- The `validate` alias is deleted from the plan — it already exists, file mode included. Net scope reduction.
- The "RDF toolkit" framing moves to a docs section ("RDF File Tools" in `docs/cli/README.md`) where grouping belongs and costs nothing.

So: your instinct held up against the survey, the middle you guessed at exists (`file`, per Stardog) but the flat surface beat it on the evidence, and the stack now ships what a riot or Oxigraph user would type on the first try.

PR #1553, opened 2026-07-29
perf: batched read lanes for Cypher relationship-property queries and large UNWINDs

bplatz:
Follow-up to #1549 (stacked on that branch). Addresses the second round of KB-on-Fluree field reports: Cypher relationship-property reads (`r1.p`, `r2.p`, `p.p`) made queries *slower* or OOM'd, and the ES-reindex `UNWIND` shape killed the container at 21k URIs.

## Problem

Reading a relationship property switches the rel var from the value-only OPTIONAL lane (batched in 4.1.3) to the **required** `f:reifies*` chain, which had no batched lane at all. On bulk-imported ledgers (`annotation_index: None`) five separate defects stacked on it, and a sixth planner defect capped how large an `UNWIND` could get before the whole plan inverted:

1. **Sidecar demotion mis-ordered untyped chains.** For `-[p]->` the chain's `f:reifiesPredicate` joins on the base predicate *variable*, so after the base edge is placed the connected pool can be nothing but the chain itself — and the IC5-era sidecar demotion then crowned `reifiesPredicate`, whose bound-object fan is ~(sidecar / #relationship-types) per driving row. This is the repeated `BinaryScanOperator::open` on `reifiesSubject` in the server log and the memory climb behind the OOM.
2. **`EncodedPid` predicates crashed both probe lanes** (untyped rels on arena-backed ledgers were a hard internal error).
3. **The required chain evaluated per driving row.** New `HashAnnotationEdgeProbeOperator`: drain the three `f:reifies*` predicates once into hash maps, sweep the base pattern once (planned scans — policy and novelty semantics preserved), hash-join each driving row on every child-bound position. Gated on driving size, sweep size (stats-bounded), current-state, single-ledger.
4. **No-new-vars filter triples ran as per-row scan re-opens.** New `MembershipJoinOperator` (build-once hash set + exact-join fallback for unbound rows); the `(s-bound, const p, const o)` subset stays on the existing batched-EXISTS lane.
5. **`ValuesOperator` was quadratic** — pairwise `is_compatible` over input × value rows (21k UNWIND × a broad child ≈ the entire residual). Value rows now hash-index on normalized overlap keys; a hash hit still passes `is_compatible`, wildcard/Iri rows keep the scan path, so results are byte-identical.
6. **Large VALUES lost the planner seed to a class scan.** A `ClassNotInStats` fallback estimate (~20k on a 72k-node ledger) out-ranked the *exact* VALUES cardinality, deferring the VALUES — the join keys — to a final filter over ~21M-row intermediates (1.3s at 18k URIs → ~712s at 21k; pre-branch exceeded 25 minutes). VALUES vars now join the seed-race anchor set so the existing disconnected-class-anchor demotion fires at seed time.

Two supporting fixes surfaced along the way: the `DefaultGraphSource` single-graph delegate now re-projects batches to its declared schema (positional consumers could silently misread reordered inner-chain layouts), and both probe lanes now **unify** a pre-bound reifier var instead of overwriting it — caught by W3C `sparql12 eval-triple-terms` pattern-8 during validation.

## Results

Repro at customer scale (`fluree-db-api/examples/cypher_kb_pageload_probe.rs`: bulk-imported n10s-style ledger, 72k `INDEXED` nodes, ~190k reified edges each carrying a `p` annotation, ~50 classes and relationship types):

| Query | Before | After (1k URIs) | After (21k URIs) |
|---|---|---|---|
| `getObjectsByUris` + `p.p` (the OOM query) | 84.5 s @1k, OOM @21k | **0.82 s** | **1.65 s** |
| same, value-only | 0.78 s @1k, >25 min @21k | 0.75 s | **1.44 s** |
| `getObjectAssociations` + `r1.p, r2.p` | 68 s | **2.1 s** | 2.1 s |
| `getObjectAssociations` stock | 1.4 s | 1.3 s | 1.3 s |

Row counts identical before/after in every case.

## Tests

- New planner unit tests (untyped-chain ordering, object-pinned wildcard estimate, VALUES seed vs class anchor) and Cypher integration pins for the rel-prop lanes (required-lane semantics, arena `EncodedPid` decode, 300-row UNWIND through the hash lane).
- Full W3C SPARQL suite green — including four `eval-triple-terms` tests that were previously in the skip register (`basic-4`, `pattern-5`, `pattern-7`, `pattern-9`) and now pass; their stale skip entries are removed.
- `docs/design/edge-annotations.md` documents the three read lanes for the expanded chain and when each engages.

## Blast radius beyond Cypher

One hunk here changes behavior for **every** query language, not just Cypher: in `estimate_triple_row_count`, a wildcard triple whose object is bound or constant (`?s ?p <o>`) moves from `FULL_SCAN` (1e12) to `DEFAULT_BOUND_OBJECT_SELECTIVITY`. SPARQL and JSON-LD compile to the same IR and share this estimator, so their plans move too.

The rationale is that an object-pinned wildcard is a reverse OPST probe bounded by the node's in-degree, not a world scan — right in kind, but in-degree is unbounded in practice, so for a hub object the new estimate *under*-estimates by orders of magnitude where the old one over-estimated. Neither is correct; without an OPST-cardinality statistic the new one is better on average and worse on hubs. Flagging it so that if a nightly `query_hot_bsbm` or `query_hot_property_path` number moves, this is the first place to look.

## Review follow-ups

- **Membership lane driving-side gate** (`3934e108c`) — the lane weighed only the build side, so a few driving rows could drain up to 250k build rows where the nested loop would have done a few probes (`{ ?a :knows ?b . ?b :knows ?a }` is the shape that slips past the batched-EXISTS carve-out). It now takes `left.estimated_rows()` and declines below 256, mirroring `HASH_ANNOTATION_MIN_DRIVING_ROWS` including its treatment of `None` as large. Four unit tests; the small-driving one goes red with the gate removed.
- **Membership lane multi-graph cardinality** (`e4b048486`) — keep/drop is only join-equivalent when a ground triple matches at most once, which is false across a multi-graph dataset (union does not dedupe). Measured: `{ ?a :knows ?b . ?b :knows ?a }` over a graph unioned with itself returned 4 rows through the hash set against the nested loop's 8. The operator now checks the active graph count at `open()` and routes every row through its existing exact per-row fallback when more than one graph is in scope, so the lane is join-equivalent unconditionally. New dataset integration test.
- **Sweep ceiling documentation** (`a166086a0`) — `BASE_SWEEP_MAX_ROWS = 20_000_000` is documented in `docs/design/edge-annotations.md` as a backstop ~100× the validated scale rather than a tuned optimum, with what to measure to replace it; the same section now states that both annotation probe lanes are pipeline breakers (so a small `LIMIT` over a large driving side pays the whole probe) and records the `keep_all` cliff.
- **Upsert bench** — tracked as a follow-up rather than here: an `upsert`-shaped bench beside `transact_commit` with new-subject and persisted-subject variants plus a `regression-budget.json` entry.


fluree-db-query/src/execute/where_plan.rs
```diff
@@ -2658,6 +2678,67 @@ pub(crate) fn build_scan_or_join(
...
+const MEMBERSHIP_JOIN_MAX_BUILD: f64 = 250_000.0;
+
+/// Recognize the membership-join shape for `build_scan_or_join`: every
+/// variable of `tp` (≥ 1) already appears in the left schema, the query is
+/// current-state (in history mode a ground triple matches once per
+/// `(t, op)` version, so keep/drop is not join-equivalent), and stats
+/// bound the standalone build within [`MEMBERSHIP_JOIN_MAX_BUILD`].
+/// Returns the key vars in left-schema order (load-bearing for column
+/// lookup at open), or `None` to keep the nested-loop join.
+fn membership_join_key_vars(
```
aaj3f, line 2710: **Blocking — CRITICAL (performance regression).** `membership_join_key_vars` gates the new lane on the *build* side only. It has no driving-side gate, so a query with a handful of driving rows now drains up to 250,000 build rows where it previously did a handful of scan probes.

Mechanism: the function receives `left_schema` but not `left`, so `left.estimated_rows()` — available at the call site on line 2612, and the exact signal the sibling lane uses — never enters the decision. `MembershipJoinOperator::build_key_set` then performs one unseeded full drain of the right triple on the **first fully-bound row**, bounded only by `MEMBERSHIP_JOIN_MAX_BUILD = 250_000` (`:2685`).

Concrete failing input: `{ ?a :knows ?b . ?b :knows ?a }` (a mutual-relationship / triangle shape, common in both Cypher and SPARQL) where the first triple is selective and yields three rows. The `(s-bound, const p, const o)` carve-out at `:2703-2712` does not catch this — the object is `Term::Var`, so the guard's condition is false and the lane is admitted. The estimate for a const predicate with both ends unbound is the property's flake count, so any `:knows` extension under 250k rows qualifies. Result: a full drain of the entire `:knows` extension to answer three rows, against roughly three per-row scan re-opens before this PR.

This is the asymmetry that makes it look like an oversight rather than a judgement call: the annotation hash lane added in the same PR gates on precisely this signal at `default_graph_source.rs:203-205` (`child.estimated_rows().is_none_or(|n| n >= HASH_ANNOTATION_MIN_DRIVING_ROWS)`), and `optional.rs:120-126` in this same diff names "defeating an early LIMIT" as a reason not to buffer speculatively. The membership lane should inherit the same discipline.

Suggested fix — thread the driving side into the gate and mirror the sibling's threshold:

```rust
// call site, ~:2601
if bounds.is_none() && inline_ops.is_empty() {
    if let Some(key_vars) =
        membership_join_key_vars(&left_schema, tp, planning, hash_planner, left.estimated_rows())
    { … }
}

// in membership_join_key_vars, alongside the history check
// A small driving side is cheaper on the nested loop's per-row probes than
// on a one-shot drain of the whole build side.
if driving_rows.is_some_and(|n| n < MEMBERSHIP_JOIN_MIN_DRIVING) {
    return None;
}
```

An unknown estimate should take the hash lane, matching `HASH_ANNOTATION_MIN_DRIVING_ROWS`'s treatment of `None` — an unbounded stream is exactly what the drain protects against.

fluree-db-query/src/membership_join.rs
```diff
@@ -0,0 +1,279 @@
+//! MembershipJoinOperator — hash-membership evaluation of a join whose
+//! right-side triple binds NOTHING new.
+//!
+//! When every variable of a join's right triple already appears in the left
+//! schema, the "join" is a pure existence filter: in current-state mode a
+//! fully-ground triple matches at most once, so each left row is kept or
```
aaj3f: *Optional (correctness question, not an asserted bug).* The module's founding premise is that "in current-state mode a fully-ground triple matches at most once, so each left row is kept or dropped." I could not confirm that holds for a multi-graph dataset, where the same ground triple can be present in more than one named graph and the nested-loop join would emit one row per matching graph while the hash set collapses them to one.

What makes this worth asking rather than dismissing: the sibling hash lane explicitly gates on `!ctx.is_multi_ledger()` (`default_graph_source.rs:266`), and `MembershipJoinOperator` has no equivalent gate — nor could it easily, since it is selected in `build_scan_or_join` at plan time where the execution context is not in hand. If duplicate-per-graph rows are the existing behavior, this lane silently changes cardinality for those queries; if they are not, a one-line comment recording why would save the next reader the same trip.

fluree-db-query/src/default_graph_source.rs
```diff
@@ -182,6 +255,38 @@ impl DefaultGraphSourceOperator {
...
+    /// Is the probe's one-pass base-edge sweep bounded? A typed
+    /// relationship sweeps one predicate partition (its stats count); an
+    /// untyped one sweeps the whole default graph, bounded by the summed
+    /// property counts. Stats absent → the ledger has no built index
+    /// (novelty-scale) — trivially bounded.
+    fn base_sweep_bounded(
+        &self,
+        shape: &crate::annotation_edge_probe::AnnotationEdgeShape,
+    ) -> bool {
+        const BASE_SWEEP_MAX_ROWS: u64 = 20_000_000;
```
aaj3f, line 282: *Optional (performance, inherited pattern).* `BASE_SWEEP_MAX_ROWS = 20_000_000` is a bare constant with no stated derivation, and it is roughly 100× the scale the PR validated against (~190k reified edges over 72k nodes).

Mechanism: the hash lane is a pipeline breaker — `next_batch` runs `probe_all` to completion before emitting anything (`annotation_edge_probe.rs:1221-1233`) — so a query with a small `LIMIT` over a large driving side pays the entire sweep. The buffering itself matches the pre-existing arena `AnnotationEdgeProbeOperator`, so it is a consistent extension rather than a new anti-pattern, but the hash lane adds the base-edge sweep *on top*, and that sweep is what the 20M ceiling governs. `PlanningContext` (`temporal_mode.rs:59-73`) carries no limit, so a LIMIT-aware gate would need real plumbing — this is a design note, not a one-liner.

There is also a sharp edge inside the sweep: `keep_all` (`annotation_edge_probe.rs:931, 957`) disables the driving-subject filter entirely if **any** single driving row leaves the subject unbound, so one such row flips the sweep from "filtered to the driving subjects" to "every base edge in the graph." Correct, but a cliff a reader would not expect from the gate names.

Suggested: record the derivation of 20M in `docs/design/edge-annotations.md` next to the lane description (it is the one number a future tuner will want and cannot recover), and say explicitly there that both annotation lanes are pipeline breakers so the LIMIT interaction is documented rather than discovered.

fluree-db-query/src/planner.rs
```diff
@@ -263,7 +263,22 @@ pub(crate) fn estimate_triple_row_count(
             DEFAULT_PROPERTY_SCAN_SELECTIVITY
         }
 
-        PatternType::FullScan => FULL_SCAN,
+        PatternType::FullScan => {
```
aaj3f, line 266: *Optional (blast radius worth naming).* This is the only hunk in the PR that changes behavior for **every** query language rather than just Cypher, and the PR description frames the whole change set as Cypher work.

Mechanism: `?s ?p <o>` (or `?o` bound by an earlier pattern) moves from `FULL_SCAN` (1e12) to `DEFAULT_BOUND_OBJECT_SELECTIVITY`. SPARQL and JSON-LD compile to the same IR and share this estimator, so their plans move too. The argument in the comment — a reverse OPST probe bounded by the node's in-degree — is right in kind, but in-degree is unbounded in practice: for a hub object the new estimate underestimates by orders of magnitude where the old one overestimated. Neither is correct; the new one is better on average and worse on hubs.

I am not asking for a change — without an OPST-cardinality statistic there is nothing better available, and the W3C suite and the 1,300 `fluree-db-query` unit tests are green. It is worth a sentence in the PR description so that if a nightly `query_hot_bsbm` or `query_hot_property_path` number moves, the next person knows where to look.

aaj3f (approved):
This is strong, well-evidenced work that fixes two availability failures (an OOM and a >25-minute query) at their causes rather than papering over them, and the lanes extend the codebase's existing batched-read shapes rather than inventing parallel ones. The measurements include the control that matters — the unmodified `getObjectAssociations` query stays flat at 1.3 s — and the row-count-identical claim is backed by 231 Cypher integration tests, 251 `grp_misc` tests, and 1,300 `fluree-db-query` unit tests, all green locally with every new test confirmed by name. 

The one possibly blocking item is narrow and mechanical: `MembershipJoinOperator` is admitted with no regard for how many driving rows it will serve, so a three-row driving side can now drain up to 250k build rows where it previously did three scan probes — while the annotation hash lane added in this same PR gates on exactly that signal, and `optional.rs` in this same diff cites "defeating an early LIMIT" as the reason not to buffer speculatively. Thread `left.estimated_rows()` into `membership_join_key_vars` and this is would be resolved.

- **Performance (speed first, memory second):** ⚠️ Large validated wins on the target shapes, but one possibly-critical regression window: the membership lane has no driving-side gate (blocking, above). A second, non-blocking window is inherited — both annotation lanes are pipeline breakers, and the new one adds a base sweep of up to 20M rows that the arena lane did not have, so a small LIMIT over a large driving side over-reads.

fluree-db-query/src/execute/where_plan.rs
bplatz: Fixed in 3934e108c. `membership_join_key_vars` now takes `left.estimated_rows()` and declines below `MEMBERSHIP_JOIN_MIN_DRIVING = 256`, mirroring `HASH_ANNOTATION_MIN_DRIVING_ROWS` including its treatment of `None` as large. Your `{ ?a :knows ?b . ?b :knows ?a }` case is now a unit test (`membership_lane_declines_small_driving_side`) alongside admit-on-large, admit-on-unknown, and the build ceiling; the small-driving test goes red with the gate removed.

fluree-db-query/src/membership_join.rs
bplatz: Good catch — this is a real cardinality bug, not just an unproven premise, and it is fixed in e4b048486.

Measured it rather than reasoned about it: `{ ?a :knows ?b . ?b :knows ?a }` over a graph unioned with itself returns **4** rows through the hash set against the nested loop's **8**. Union does not dedupe (as `dataset_multiple_default_graphs_no_dedup` already documents), so the same ground triple matches once per graph and keep/drop collapses the second match.

Rather than gate on `is_multi_ledger()` — which would still miss multiple named graphs of a *single* ledger in the active scope — the operator now checks the active graph count at `open()` via `ctx.active_graphs()` and routes every row through its existing exact per-row fallback when more than one graph is in scope. The execution context is in hand there, so no plan-time plumbing was needed. The lane is join-equivalent unconditionally now, and the hash fast path engages only where the premise provably holds. New test: `dataset_membership_shape_keeps_per_graph_multiplicity`.

fluree-db-query/src/default_graph_source.rs
bplatz: Documented in a166086a0. `docs/design/edge-annotations.md` gains a "Buffering and the sweep ceiling" subsection recording that 20M is a backstop ~100x the validated scale rather than a tuned optimum, what to benchmark to replace it (sweep vs nested loop at 1M/5M/20M), that both annotation probe lanes are pipeline breakers so a small LIMIT over a large driving side pays the whole probe, and the `keep_all` cliff — one driving row with an unbound subject flips the sweep from subject-filtered to every base edge in the graph. The constant now points at that section.

fluree-db-query/src/planner.rs
bplatz: Added to the PR description under "Blast radius beyond Cypher" — that the estimator is shared with SPARQL and JSON-LD, that the new estimate under-estimates on hub objects where the old one over-estimated, and that `query_hot_bsbm` / `query_hot_property_path` are the first place to look if a nightly number moves. No code change, agreed there is nothing better available without an OPST-cardinality statistic.

bplatz:
@aaj3f All four items addressed — thanks especially for the membership-lane pair, both turned out to be real.

- **Driving-side gate** (`3934e108c`): threaded `left.estimated_rows()` into `membership_join_key_vars`, declining below 256 to mirror `HASH_ANNOTATION_MIN_DRIVING_ROWS` (unknown estimate treated as large). Your mutual-`:knows` case is a unit test now.
- **Multi-graph cardinality** (`e4b048486`): your question had a bug behind it. Measured — that same shape over a graph unioned with itself returns 4 rows through the hash set against the nested loop's 8. The operator checks the active graph count at `open()` (`ctx.active_graphs()`, which also catches multiple named graphs of one ledger — broader than `is_multi_ledger()`) and routes every row through its exact per-row fallback when more than one graph is in scope, so the lane is join-equivalent unconditionally.
- **Sweep ceiling** (`a166086a0`): 20M documented as a backstop ~100x the validated scale, with what to measure to replace it, plus the pipeline-breaker note and the `keep_all` cliff.
- **Estimator blast radius**: new "Blast radius beyond Cypher" section in the PR description naming the shared estimator and the hub-object trade-off.

Validation: `fluree-db-query` 1,304 unit tests, `grp_query` 385, `grp_query_sparql` 296, `grp_misc` 251, `it_query_cypher` 231, `grp_transact` 183; W3C SPARQL 36/36 suites (246 passed, 0 failed); workspace clippy `--all-features --all-targets` clean. The upsert bench stays a follow-up.

PR #1555, opened 2026-07-29
feat(rdf): RDF toolkit core — quad IR, IR-level writers, W3C conformance harness, fluree rdf check/count/convert

aaj3f:
# RDF toolkit core: quad IR, IR-level writers, W3C conformance, `fluree rdf`

Stacks on **#1552** (GraphSink protocol). Base: `feat/graphsink-protocol` @ `869703009`.

Five workstreams of the `fluree rdf` effort, integrated into one branch. They ship together because they are only useful together —
#1552 defined a `GraphSink` protocol that nothing consumed into bytes, the IR
could not represent a quad, no harness could say whether the parser was
correct, and there was no command to point at a file.

After this branch:

```console
$ fluree rdf check dump.ttl                    # parse, report diagnostics with carets
$ fluree rdf count dump.nt.gz --profile        # tallies, plus where the time went
$ fluree rdf convert dump.ttl -o dump.jsonld   # read one syntax, write another
```

and `fluree-graph-format` exposes five streaming `GraphSink` writers
(N-Triples, N-Quads, Turtle, TriG, JSON-LD), so a parser can emit straight
into an output syntax with nothing materialized in between — which `convert`
now does end to end.

**W3C RDF 1.1 Turtle conformance: 309/313 — 98.7%**, gated in CI, with the
failure register policed in both directions so it can only shrink.

## Three bugs in shipped behavior, fixed here

Each was found by a workstream below, and each was live before this branch.

**A relative `@base` corrupted every IRI in the document.** RFC 3986 §5.1.1
says a base declared relative resolves against the base in scope; the parser
installed the reference verbatim, so `@base <foo/>` made the document base the
string `"foo/"` and every subsequent IRI resolved against a scheme-less base
into `<:foo/a>`. Nothing failed while this happened — an import or conversion
wrote a document full of wrong IRIs and reported success. This is on the
shared ingest parser, not just the new verbs.

**Exported IRIs were percent-encoded, which is not injective.** Given the IRI
`http://ex/a b`, `fluree export` emitted `http://ex/a%20b` — a *different*
IRI, and exactly what a caller separately holding the IRI whose text really is
`http://ex/a%20b` would also emit. Two distinct resources came out as one,
silently. The `IRIREF` grammar provides `UCHAR` for precisely this, so we now
write `<http://ex/a b>`, which reads back as the IRI we were handed. **This
moves the export goldens by four lines**, all IRIs; the JSON-LD golden is
untouched because that path escapes for JSON, not `IRIREF`.

**Prefixed names could be emitted that no parser accepts.** `-` is in
`PN_CHARS` but not `PN_CHARS_U`, so it cannot *start* a local name, and prefix
names were never validated at all — a JSON-LD `@context` key like `1st` or
`has space` became `1st:thing` on every IRI it matched.

A fourth defect was IR-level rather than export-level: `true` became
`LiteralValue::Boolean` while `"true"^^xsd:boolean` — the same RDF term
written longhand — became `LiteralValue::String`. `Term`'s equality and hash
are variant-sensitive, so a graph stating a fact both ways held **two**
triples where RDF says one. Dedup and set semantics were wrong, entirely
independent of any test suite.

---

# 1. Quad-capable IR and diagnostics (6 commits)

**`Quad` / `Dataset`** (`fluree-graph-ir/src/quad.rs`, `dataset.rs`). #1552
added `supports_quads()` and an `emit_quad` that *refuses* by default —
deliberately, because silently folding named graphs into the default graph is
data loss. Nothing could answer `true` to that probe until now. `Quad` wraps a
whole `Triple` plus `Option<Term>`, so every existing triple consumer keeps
working and `None` means the default graph — a real graph, not "no graph".
Ordering is graph-major, which is the grouping TriG and readable N-Quads both
want. An empty named graph (TriG's `<g> {}`) is representable and
distinguishable from an absent one. A literal graph name is refused rather
than stored.

**`DatasetCollectorSink`** (`dataset_sink.rs`). Term interning, blank-node
identity, and statement-scoped literal recycling moved into one `TermTable`
shared by both sinks rather than two copies that can drift on term lifetime —
the invariant recycling's soundness rests on. `GraphCollectorSink`'s behavior
is unchanged by that move (same id order, same `-b{N}` disjoint mint
namespace, same slot reuse). Rollback spans graphs: a statement can write to
the default graph and several named ones before failing, so `abort_statement`
rewinds each and *removes* a named graph only the failed statement created.

**`Diagnostic` + `Severity` + `LineIndex`** (`diagnostic.rs`,
`line_index.rs`). Parse errors carried byte offsets, not positions — not a
value a `--continue-on-error` run can collect, count, or filter. `LineIndex`
builds its line-start table on the *first* lookup through a `OnceCell`, so
constructing one is free and a clean parse never scans for newlines. Columns
are 1-based **character** counts (what makes a caret land under the glyph it
blames). Line 0 marks "no position", because some conditions are facts about a
document rather than about an offset, and inventing offset 0 for them points
the caret at a lie.

Two behaviors worth knowing before building on this:

- **Minted anonymous blank labels are deliberately non-serializable.** `-b{N}`
  buys disjointness from user-written labels through illegality: `_:-b1` is
  rejected by this repo's own lexer. A writer must relabel every anonymous
  mint — it cannot pass one through the way it passes a user-written label
  through.
- **`abort_statement` rolls back statements, not directives.** A
  `@base`/`@prefix` accepted before a statement fails stays accepted;
  `on_base`/`on_prefix` are infallible and a producer cannot un-resolve IRIs
  it already expanded.

**Known protocol gap, deferred to M2.** `emit_list_item` has no quad form, so
a Fluree-style *indexed* collection inside a named graph cannot be expressed
and lands in the default graph. That is a gap in #1552's protocol, not in this
sink, and nothing reaches it until the TriG/N-Quads parsers land.
`PROTOCOL_QUAD_EVENTS` counts the trait's quad-shaped events with a test whose
failure names the follow-up work — a forced decision point, not a detector,
since Rust cannot enumerate trait items and a default-bodied addition compiles
against every impl.

# 2. IR-level streaming RDF writers (15 commits)

Five `GraphSink` implementations in `fluree-graph-format`.

| writer | syntax | quads | memory |
|---|---|---|---|
| `NTriplesWriter` | N-Triples | no | O(1) |
| `NQuadsWriter` | N-Quads | yes | O(1) |
| `TurtleWriter` | Turtle, "blocks" tier | no | O(1) |
| `TrigWriter` | TriG, "blocks" tier | yes | O(1) |
| `JsonLdWriter` | JSON-LD | no | O(document), by construction |

The two export bugs above were found by lifting `export.rs`'s formatting
primitives into a place where they could be tested properly.

### On `<http://ex/a b>` looking alarming

The obvious review question: does an escaped space inside `<…>` mean we are
emitting something a tool will reject? No. `IRIREF` is
`'<' ([^#x00-#x20<>"{}|^`\] | UCHAR)* '>'`, and ` ` is a `UCHAR` — the
production exists precisely so a forbidden character can be *spelled*. Our own
parser reads it back as the IRI we were given, which the round-trip tests
assert.

riot may still warn on such an IRI, but the warning is about the IRI **value**
— a space is not permitted in an RFC 3987 IRI — and that is a property of the
data we were handed, not of how we spelled it. The old percent-encoding made
that warning disappear by *changing the IRI to a different one*, which is the
bug: it silenced a complaint about the input by corrupting it. Emitting the
IRI faithfully and letting a validator object to it is the correct division of
labor; `fluree rdf check` is where that objection belongs.

### Deliberately NOT changed: `:` in local names

`PN_LOCAL` admits `:` in every position, so `ex:has:colon` is a legal prefixed
name that this writer declines to emit. Accepting it would be correct by the
grammar and would change what `fluree export` produces for every IRI with a
colon in its local part — a change with no correctness behind it.

The two fixes above changed shipped output because the previous behavior was
*wrong*. This one would only make it *different*. It is left as its own
decision rather than arriving as a side effect: `is_emittable_local` in
`fluree-graph-format/src/prefix.rs` is where it lives, and dropping its `:`
clause is the whole change. **Open question — decide separately.**

### Rollback semantics, and the one case that cannot roll back

`abort_statement` is honest about what a streaming writer can do. Unbuffered —
the default — bytes already handed to the underlying writer stay written, and
the docs say so rather than implying otherwise. `WriterConfig::buffer_statements`
holds each statement back until the producer commits it, making rollback real
at O(widest statement); that is the switch `--continue-on-error` turns on. The
Turtle run state (`;`/`,` folding) rolls back with it, so a rejected statement
leaves no half-open subject.

One residual case, inherent rather than an oversight: **a directive arriving
mid-statement force-commits that statement**, so a later `abort_statement`
cannot retract it even in buffered mode. A `@prefix` is a top-level thing and
must flush whatever precedes it — it cannot be written inside an unterminated
statement, or inside a TriG `GRAPH` block — so the statement is terminated and
released before the directive goes out. Producers do not do this (Turtle
directives sit between statements), but the protocol permits it and the public
API allows it, so it is stated. Output is valid in both modes; the only thing
lost is the ability to un-emit that one statement.

### Fidelity tiers and blank-node policy

Turtle and TriG implement the **blocks** tier: consecutive same-subject runs
fold with `;`, same-subject-and-predicate with `,`, document order otherwise.
No cross-document regrouping, no `[ ]` reconstruction, no `( )` re-collapse —
all three need the whole document in memory, which is what a streaming writer
exists not to need. riot draws the same line. A buffered `--pretty` tier is
the follow-up. TriG writes quads where they arrive, reopening a `GRAPH` block
when the input returns to a graph it left — valid TriG, and O(1) against
interleaved input where buffering per graph would be O(dataset).

Blank nodes relabel by default: a fresh `b{N}` per input label, with
`_:fdb-…` passing through verbatim for #1432's addressability contract.
Disjointness is by construction rather than by check.
`--preserve-bnode-labels` emits *user-written* labels verbatim; a label that
is not a `BLANK_NODE_LABEL` was never written by a user, so it is an internal
mint or an externally-sourced identifier and is relabelled into a reserved
`fdbw-` namespace instead. The label check is the whole production, not an
approximation: `_:ab.` does not fail to parse — it lexes as `_:ab` followed by
the statement terminator, so the node would be silently renamed.

# 3. W3C conformance harness and parser burn-down (11 commits)

A new excluded workspace, `testsuite-rdf`, plus a CI job. It reads the
existing `testsuite-sparql/rdf-tests` submodule through a relative path, so
one checkout serves both harnesses at one commit.

The RDF 1.1 Turtle and N-Triples suites **gate**: every test must pass or be
listed in `tests/registers/mod.rs`, policed in **both** directions — an
unexpected failure and a stale register entry each fail the suite — so the
register can only shrink. The RDF 1.2 suites and the ingest-default mode run
in the same invocation and are reported as informational, never gated.

The harness landed at 297/313 (94.9%) with 16 Turtle failures under six
causes. This branch clears five:

| # | Cause | Tests | Rate after |
|---|---|---:|---:|
| — | baseline | — | 94.9% |
| A1 | repeated `;` as empty predicateObjectList items | 4 | 96.2% |
| A2 | PN_LOCAL interior dot runs | 1 | 96.5% |
| A3 | relative `@base` resolution | 1 | 96.8% |
| B | directive keyword case, both directions | 3 | 97.8% |
| D | boolean keyword vs longhand IR duality | 3 | **98.7%** |

**RDF 1.1 Turtle 297/313 → 309/313 (94.9% → 98.7%).** N-Triples is unchanged
at 55/70 — every one of its remaining failures needs a strict N-Triples
reader, which is M1 scope, not a parser bug. The 4 Turtle tests still
registered are all **cause C** (IRI and language-tag validation), which is the
H-8 "IRI validation ships in the light crates" workstream and was deliberately
left alone. Register entries were removed in the same commit as each fix,
which the both-way policing requires.

Mode delta, since the harness asserts conformant mode never scores below the
ingest default: ingest-default Turtle 272 → **281** (86.9% → 89.8%),
conformant Turtle 297 → **309**, and `ParserOptions::conformant` is now worth
**28** tests, up from 25. **No test regressed in either mode** — every delta
is a test moving from fail to pass.

### The two that are more than a test count

**A3 was silent data corruption**, described in the lede. One detail worth
recording: the observed `<:foo/a3>` has a leading colon that nothing
prepends. With a scheme-less base, `fluree_vocab::iri::parse_iri_components`
finds no `:`, reports an EMPTY scheme, and treats the whole string as the
path; RFC 3986 §5.3 recomposition then unconditionally emits `scheme ":"`,
which prints as a bare `:` when the scheme is empty. Verified directly rather
than reasoned about — the shared resolver is behaving as specified for an
input that cannot occur once the base is guaranteed absolute, so there is no
second bug and the resolver needed no change.

The same corruption was also reachable through the API without any syntax at
all: `parse_with_prefixes_base[_options]` installed its `base` **argument**
verbatim, and the CLI's `--base` flag feeds that exact entry point. A relative
parameter is now **refused, not resolved** — the directive has a document base
in scope to resolve against, but a caller-supplied base has no context, so
resolving would mean inventing one.

**D was an IR defect, not a comparison artifact** — the boolean duality in the
lede. Fixed exactly the way integers and doubles already were
(`NumericStyle::PreserveLexical` routes the keywords through the typed-string
lane), so `Term`'s equality and hash are untouched. The ingest default is
unchanged: booleans stay native `LiteralValue` values on the storage path,
pinned by a test.

### Harness review fixes (the four leading commits)

- **Exact suite totals.** A pass RATE cannot notice its own denominator
  shrinking: drop half a manifest and the survivors still score 100%. Each
  suite now declares `expected_total` (313 / 70 / 96 / 70) and a mismatch
  fails with a message naming the re-count step, which doubles as the
  submodule-drift tripwire.
- **Register problems are not test failures.** A dangling register entry was
  landing in `failed` and breaking `total = passed + failed + ignored`
  precisely in runs that were already failing.
- **Dirty trees are visible in the freshness contract.** `git_sha()` appends
  `-dirty` and `Conformance` carries `dirty: bool` — the flag serves
  structured consumers, the suffix means a comparator that knows nothing about
  the flag still refuses the run.
- **Language tags fold case in the checker.** RDF 1.1 §3.3 makes a language
  tag case-insensitive for term equality. Folded during normalization in the
  CHECKER ONLY — the IR is untouched, since the parser's job is to preserve
  what the document said.

### Judgment calls a reviewer should check

- **`NumericStyle` covers booleans without being renamed.** Booleans are not
  numerics, so the name is now slightly narrow for what it governs. Renaming
  is a breaking change to a type on the #1552 base that other workstreams
  already build against; the documentation was widened instead.
- **`GRAPH` was left case-sensitive.** It is TriG's keyword, no test here
  exercises its case, and widening it unverified would be a guess.
- **PN_PREFIX keeps its single-dot lookahead.** The A2 fix covers PN_LOCAL,
  which is what the grammar production and the failing test are about. No W3C
  test exercises a dot run in the prefix, so it was left rather than changed
  unverified — flagged because the two productions share a rule.

# 4. `fluree rdf` — check, count, profile (5 commits)

RDF syntax tooling that reads files rather than ledgers: no `.fluree/` lookup,
no ledger, no connection. `check` parses and reports diagnostics; `count`
reports statement and term tallies; `convert` is registered as a named stub
that refuses with a message naming what is missing rather than "not
implemented".

**One syntax table.** `RdfSyntax` names all eight target syntaxes, not just
the readable two, so unsupported input is refused by name with what it waits
on, instead of being sniffed into Turtle and failing thirty lines in.
Resolution is explicit flag > extension > content sniff, with the compression
suffix stripped first and resolved independently. `detect.rs` is untouched —
it still serves insert/upsert/validate and the server; unifying them is a
follow-up.

**Exit codes split "the document is bad" (1) from "the invocation is bad"
(2)**, which is what makes these verbs wrappable. Transparent `.gz`/`.zst` on
both files and pipes; gzip is decoded multi-member so `pigz`/`bgzip` output
does not silently truncate, and the stream is opened exactly once — a pipe
cannot be reopened at byte 0. Input is capped at 4 GiB, the parser's `u32`
token-span limit, with chunked reading deferred to the parallel pipeline.

### The profiler is honest about its own instrument

`--profile` (human) and `--profile=json` (schema `fluree.rdf.profile.v1`) both
go to stderr so piped output stays clean. This part went through three
correction rounds under review and the result is worth reading before trusting
any number it prints.

**The sink usually cannot be measured, and says so.** Timing every sink call
costs two clock reads per event, and on these verbs a clock read costs *more
than the work it measures* — a bracketed discard-sink call is about 20 ns, of
which 19 ns is the clock. Review found the first implementation's sink phase
was ~98% instrument artifact and stamped "trusted". Now: events are counted
exactly and timed on a jittered sample of statements, the clock's own cost is
subtracted from the sample **before** it is scaled up, and if what remains
does not clear 3× the extrapolated artifact the tool reports "below the
measurement floor" rather than a number — on the documented host, **~96 ns per
call**, which is where the clock's own ~32 ns/pair stops being separable from
the work. The JSON says `null`, never `0`, so it cannot be read as "free".

**Sampling is jittered, because a fixed stride is a residue class.** A fixed
127-stride reaches only `P/gcd(127,P)` of a periodic corpus's shapes; on a
one-fat-statement-per-127 corpus that was a 425× over-report when the fat
shape was sampled and a total miss when it was not. The gap is now drawn from
`1..2×SAMPLE_STRIDE`, seeded from the corpus, so a given input still samples
the same statements across runs.

**But the estimate assumes a cooperating corpus, and the docs now say so.**
The schedule is a deterministic function of the input — that is what makes two
profiles comparable — and both `corpus_seed` and the generator are public, so
anyone choosing the input can replay the schedule and park the expensive work
in the gaps. A reviewer did exactly that and hid 315 ms. Reproducible and
unpredictable are incompatible, and reproducible is the one a baseline needs,
so the guarantee is scoped to accidental structure and stated as never
covering an untrusted corpus.

**Two trust verdicts, not one.** `phases_trusted` is driven by the clock reads
actually taken; `sink_trusted` by the extrapolated artifact. They are separate
because the second is false on essentially every `count` — a discard sink's
artifact is a large share of a fast parse — and one combined flag that is
always false is a flag nobody reads. **A regression gate should key on
`phases_trusted`**, and the docs say so in the field table.

**`git_sha` names the build, not the shell.** It shelled to the working
directory, so a binary built from one commit and run from another checkout
reported the wrong SHA — worse than having no field, since a baseline would be
attributed to code that never produced it. It now runs `git -C <dir of
current_exe()>` and honestly says `"unknown"` for a binary installed outside
any checkout.

Also: a costly `finish()` is measured exactly rather than passed through the
sample scaling (routed through it, a 50 ms writer flush was reported as
6.4 s); phase shares are of wall clock, not of the phase sum, so unclaimed
time shows as `unattributed` instead of being redistributed; and the worked
example in the docs is real output from a **release** build, since a debug
binary calibrates a reader roughly 20× low. Peak-RSS figures ship with their
fixtures (`scripts/rdf-rss-fixture.py`) because the ratio depends on corpus
shape — 6.7× with every subject distinct, 1.9× with a hundred reused — the
excess being the parser's IRI cache, which grows with distinct IRIs rather
than bytes.

---

## `fluree rdf convert` — the serial conversion path

The verb the effort exists for. Input (file or stdin, `.gz`/`.zst`) → Turtle
parser in conformant mode → a writer chosen by output syntax → file or stdout.

```console
$ fluree rdf convert dump.ttl --to nt > dump.nt
$ fluree rdf convert dump.ttl -o dump.jsonld
$ fluree rdf convert dump.nt.gz --to turtle --prefixes ctx.json
$ cat dump.ttl | fluree rdf convert --syntax turtle --to nquads
```

Format pairs: `{turtle, ntriples}` in → `{turtle, ntriples, nquads, trig,
jsonld}` out. Quad *input* waits on the M2 parsers. Output syntax resolves
`--to` > `-o` extension > N-Quads, matching riot's documented default.
`--bnode-policy relabel|preserve` and `--prefixes <JSON|PATH>` reach the
writers; the latter accepts a JSON-LD `@context` document unchanged and means
one thing across syntaxes — Turtle and TriG compact with it, JSON-LD takes it
as its context.

**Conformant parser options are mandated, not preferred.** Under the ingest
default a collection arrives as indexed list items, which every writer refuses
because an indexed list item is a Fluree storage shape with no RDF
serialization. A test pins the observable consequence: the
`rdf:first`/`rdf:rest` spine, and `rdf:nil` for `()`.

### Two properties a user has to know before pointing this at real data

**JSON-LD buffers the whole document, and it costs about 10× what the
streaming syntaxes do.** Its writer is document-at-once by construction — the
format has no streaming form short of NDJSON, which is deferred — so the whole
graph is held and nothing reaches the output until the input is exhausted.
Measured on the shipped `scripts/rdf-rss-fixture.py` corpus, release build:
peak RSS 2100 MiB for JSON-LD against 197 MiB for N-Triples on the same 22 MiB
input, ~95× against ~9×. There is also no early pipe exit: `--to jsonld |
head -5` parses everything first, so the trick that makes `--to nt | head -5`
cheap does not apply. This is the same call riot makes and the same one
RDF/XML will make; it is documented, with a reproduction recipe, not fixed.

**A mid-stream failure truncates `-o`.** Refusals that need no input — an
output syntax with no writer, a malformed `--prefixes` namespace, `--pretty` —
are raised before the destination is opened and leave the file untouched.
Anything discovered mid-stream, a parse error or a writer refusing an event,
happens after `-o` was created and written to, so it now holds a prefix of the
conversion. A streaming converter cannot honestly promise otherwise; writing
to a temporary and renaming would trade that for double the disk and no
streaming to the final path. The error says which case you got.

### Telemetry: how `serialize` and `write` became measurable

The review had already ruled out a per-emit clock (at ~96 ns/call the
instrument costs more than the work). `write` is instead measured at the
destination, behind a 64 KiB buffer — writer → `BufWriter` → `TimedWriter` →
file — so the writer's small writes cost no clock and the phase is real I/O
time at a few clock reads per document. `serialize` is then the sink estimate
minus `write`, and is absent entirely when that estimate does not clear its
measurement floor rather than being invented from a number that is not there.
A decomposition supersedes the `sink` row instead of sitting beside it, so
nobody adds a total to its own parts.

Two disclosures that belong with those numbers:

- **`serialize` carries about 17%.** Differential truth is ~0.83× of the
  reported figure — bracketed timing on a superscalar core measures something
  slightly different from marginal wall time. It is presented as a
  measurement, so the uncertainty travels with it.
- **`TimedWriter` is `!Send`.** It shares its clock through
  `Rc<Cell<Duration>>`, chosen because the serial driver is single-threaded
  and an `Arc<Mutex>` would put a lock on the write path. The parallel bucket
  needs either a per-worker clock aggregated at join or an atomic swap;
  recorded here so it is not discovered at thread-spawn time.

### Round-trip fidelity, scoped to what is actually true

An earlier version of this branch asserted byte-identical N-Triples across a
Turtle round trip as a general invariant, "deliberately stronger than
isomorphism". It is not general, and review caught it. Blank-node labels
survive only when the parser mints them in the same order the writer emits
them, which holds for flat shapes and fails the moment they nest — the parser
mints outermost-first, the writer emits deepest-first because an inner node
must exist before the triple referencing it. The fixture carrying the claim
happened to be flat.

The suite now splits: flat shapes keep byte equality with the reason written
down, and nested shapes (collection-of-collections, triple nesting, nested
anonymous nodes — the reviewer's fixtures) assert isomorphism *and* that
relabelling really occurs, so the stronger claim cannot be quietly reinstated.
The isomorphism checker is lifted from the review's rdflib oracle and has its
own test proving it can distinguish graphs; without that, every round-trip
assertion above it would have been vacuous.

### Error semantics

`0` converted, and also a closed downstream pipe — every layer reports `EPIPE`
on `| head` and the right answer to all of them is to stop quietly, which is
riot's behaviour. `1` the input did not parse. `2` the invocation or the
destination was wrong. Broken-pipe detection is on `io::ErrorKind`, never on
the message: glibc translates `strerror` under `LC_MESSAGES`, so a substring
check passes for whoever wrote it and fails for everyone else.

`--continue-on-error` is **not** here. It needs statement-scoped output
buffering wired to a flag — `WriterConfig::buffer_statements` already makes
`abort_statement` a true rollback — plus a decision on the exit code when
statements were skipped.

---

## Review provenance

Each of the four lines passed multi-round adversarial review with independent
verification before integration, and the fixes are folded in here rather than
listed as future work. Concretely, review changed the shipped result in ways
worth naming: the sink estimator was rebuilt after a reviewer proved it was
measuring the clock; a reviewer's hand-built corpus defeated the sampling and
the docs were rescoped rather than the claim defended; the `@base` fix was
found incomplete through the API and extended; two probe files authored by
review are adopted verbatim and credited in-tree
(`fluree-graph-ir/tests/quad_rollback_adversarial.rs`,
`fluree-graph-turtle/tests/dataset_parser_parity.rs`); and the byte-for-byte
pin on the lexer's messages was written and shown green against the
*unmodified* lexer first, so it proves the output is unchanged rather than
documenting whatever it became.

## CI does not fire on this PR

The base is `feat/graphsink-protocol` (#1552's branch), not `main`, and this
repo's CI triggers only on `main`-based PRs. Nothing here has been exercised
by GitHub Actions. The full local battery below stands in for it, and CI will
fire for real when the stack rebases onto `main`.

| Gate | Result |
|---|---|
| `nextest` — graph-ir, graph-format, graph-turtle, graph-json-ld, fluree-vocab, db-transact | **969 passed**, 0 failed |
| `nextest -p fluree-db-api` (full — the shared-abstraction blast radius) | **2923 passed**, 0 failed, 11 skipped |
| `nextest -p fluree-db-cli -p fluree-db-docs` | **450 passed**, 0 failed (91 of them `rdf::`) |
| `cargo test --doc` (graph-ir, graph-turtle, graph-format) | 14 passed |
| `testsuite-rdf` (`cargo test`) | **32 unit + 5 integration**, 0 failed |
| W3C RDF 1.1 Turtle conformance | **309/313 — 98.7%**, register at 4 |
| W3C RDF 1.1 N-Triples conformance | 55/70 — 78.6%, register at 15 |
| `cargo check --workspace --all-targets` | clean |
| `cargo check --all-targets` — excluded `testsuite-sparql` | clean |
| `cargo check --all-targets` — excluded `testsuite-shacl` | clean |
| `cargo clippy --all-targets --no-deps` — graph-ir, graph-turtle, graph-format, db-cli, db-api, bench-support | clean |

## Follow-ups recorded, not silently dropped

- `--pretty` buffered fidelity tier for Turtle/TriG (H-8). `convert --pretty`
  refuses today rather than quietly emitting blocks-tier output under a flag
  that promised more.
- `--continue-on-error` for `convert`: the writers already support the
  statement buffering it needs; the driver wiring and the skipped-statement
  exit code are what is left.
- Compressed *output* (`-o out.nt.gz`), which is refused today rather than
  writing plain text into a file whose name promises gzip.
- The parallel conversion pipeline, which needs a `Send` write clock (above).
- NDJSON streaming mode for JSON-LD; the document mode here is O(document).
  `JsonLdWriter` is triple-only, because `format_jsonld` cannot express a
  dataset — refusing beats reproducing the read-side adapter's silent `@graph`
  drop.
- Canonical N-Triples output mode (`ECHAR` table) belongs with the `canon`
  verb in M5.
- `emit_list_item` has no quad form (M2, above).
- Cause C — IRI and language-tag validation — is the remaining 4 Turtle
  registrations, and belongs to the H-8 light-crate IRI validation workstream.
- Strict N-Triples reader (M1) is what the 15 N-Triples registrations need.
- `detect.rs` and `RdfSyntax` should become one syntax table.
- Chunked/streaming reading above the 4 GiB cap lands with the parallel
  conversion pipeline.


aaj3f:
Surface note: the `fluree rdf {check,count,convert}` CLI group this PR introduces is dissolved into flat top-level verbs (`parse` / `count` / `convert`) at the top of the stack — #1578's final five commits — per the naming discussion on #1552. This diff still shows the grouped spelling; the merged result is flat. Rename landed once at the stack top to avoid rebasing ~80 stacked commits through the same test file.

PR #1565, opened 2026-07-29
fix(turtle): a redeclared prefix must change what later names mean

aaj3f:
# Two correctness fixes against main

Both found while building the RDF toolkit, both independent of it, both small enough that one review slot beats two.

1. **`fix(turtle)`** — a redeclared prefix silently didn't change what later prefixed names meant. Data corruption, no error.
2. **`fix(api)`** — a literal NUL byte in an `admin.rs` comment made the whole file binary to every text tool, so plain `grep` reported "no match" for patterns that occur 23 times in it.

---

## 1. A redeclared prefix must change what later names mean

Silent data corruption in the Turtle parser. Turtle allows a prefix to be rebound part-way through a document; the parser ignored the rebinding for any prefixed name it had already seen.

```turtle
@prefix e: <http://a/> .   e:x <http://p/> "1" .
@prefix e: <http://b/> .   e:x <http://p/> "2" .
```

Both statements land on `<http://a/x>`. The second should be `<http://b/x>`. No error, no warning — the document says one thing and the graph holds another.

## Cause

`prefixed_term_cache` is keyed by span text (`e:x`). That text identifies an IRI only *relative to the bindings in force* when the entry was made, and nothing invalidated it when a binding changed.

Two neighbours were **not** affected, and the PR pins both so the scope is demonstrated rather than asserted:

- `iri_term_cache` keys on the **resolved** IRI, so a different base produces a different key and never collided.
- `@base` redeclaration was always correct — a prefix namespace is resolved once, when its directive is read, and prefixed names never consult the base at lookup time.

## Fix

On a prefix directive, clear the expansion cache **if the prefix rebinds to a different namespace**. Only if.

Redeclaring the *same* binding is the common case rather than a curiosity: chunked import prepends the file's prefix block to every chunk over a pre-seeded prefix map, so the parser sees each prefix declared twice by construction. Every cached expansion stays correct through that, and clearing there would cost hit rate across the entire import path for no correctness gain.

**The per-term path is untouched.** The diff adds no instruction to `resolve_prefixed_term` or `sink_term_iri` — the only executable change is inside the prefix-directive handler. Added work is one hash lookup per prefix *directive*, of which a document has a handful, against a cache consulted once per prefixed *name*, of which it has millions. That is a structural property of the diff, not a benchmark result; a wall-clock A/B on this host was noise-dominated (a 4× swing between runs of identical code under concurrent load) and is not offered as evidence.

Clearing is coarse on purpose: a rebind invalidates only that prefix's entries, but finding them means scanning every key, so we drop them all. A per-lookup epoch check would make rebinding O(1) at the cost of a comparison on every cache hit — the wrong trade when the overwhelming majority of documents rebind nothing.

**The comparison is on meaning, not spelling**, and both directions matter. `parse_prefix_directive` resolves the namespace before storing it, so `prefixes` holds resolved IRIs. The same spelling under two different bases (`@prefix e: <a/>` under `<http://x/>` then under `<http://y/>`) names two namespaces and *must* clear; two spellings that resolve alike (`<http://x/a/>` and `<a/>` under `<http://x/>`) name one and must *not*. Comparing raw text would get each backwards.

### The "only when" needs its own guard

Clearing on a same-binding redeclaration produces *identical triples* — just colder — because `iri_term_cache` absorbs the repeated work before it reaches the sink. So a mutation deleting the condition passed every gate in the first version of this PR; confirmed by running it, with all eleven triple-level tests still green.

The cache hit/miss counters are the only place the behaviour is observable, so a test-only `parse_reporting_cache` exposes them and two tests read them: same-binding redeclaration must leave the cache warm (1 hit, 1 miss), a real rebinding must drop it (0 hits, 2 misses). The same mutation now fails the first.

## Blast radius — honest version

**This is not serial-only.** Bulk import is affected, and this fix is necessary but *not sufficient* there.

The `PrefixAfterData` guard lives in `compute_chunk_boundaries` (the whole-file pre-scan behind `TurtleChunkReader`). Import does not use that reader. `fluree-db-api/src/import.rs` uses `StreamingTurtleReader` exclusively, which has no such guard — verified empirically: a 4000-statement file with a mid-file redefinition is **accepted**, its extracted prefix block is only the head binding `@prefix e: <http://a/> .`, and the redefinition is delivered in a chunk like any other data.

So a mid-document redefinition reaching import hits two independent defects, and `prefix_redefinition_blast_radius.rs` *executes* the split rather than asserting it. The corpus is 4000 statements of `e:s{i} <http://p/> e:tag .` with a redefinition half way through, parsed the way import parses it (prelude seeded, prefix block prepended).

The `e:tag` object is the load-bearing part. It repeats on every line, so it is cached under the first binding and requested again after the rebinding — which is the only shape in which this bug can appear at all. Counting inside the **carrier chunk** (the one that actually contains the redefinition), post-redefinition triples whose object still resolves to `http://a/tag`:

1. **Carrier chunk — 379 stale without the fix, 0 with it.** This is the cache bug, and this assertion is what distinguishes the fix; verified by reverting the parser to the merge base. **Fixed here.**
2. **Later chunks — 1621 subjects mis-resolved.** The head-extracted prefix block is prepended to each chunk, so they parse with the *original* binding and never see the redefinition. **Not fixed here**; that is the §1.4 chunker work (mid-file directive detection → serial fallback), and it needs the splitter, not the parser.

An earlier version of this test used a literal object and asserted only that no *subject* was mis-resolved in the carrier chunk. That was vacuous — every subject is a distinct span, so no prefixed name was ever looked up twice and the cache was never consulted for a span it held. It passed against a parser with the fix entirely removed. Caught in review; the corpus and the assertion above are the remediation.

The later-chunk defect is asserted live (`> 0`) rather than `#[ignore]`d: an ignored test runs nowhere and rots, while this one fails the moment §1.4 lands and tells whoever closed it to come here. The 1621 is a printed diagnostic, not part of the assertion, so ordinary chunking changes cannot make it flap.

Also noted, not addressed: `PrefixCheck` only matches a directive keyword at line start, so even the guarded pre-scan path would miss `… . @prefix e: <…> .` written mid-line. Same §1.4 bucket.

## Discovery

Found while wiring term validation for H-8 (burn-down cause C), when a probe of the prefix path returned two statements on one subject. Nothing about it is specific to validation — it reproduces on plain `parse`, on `main` at `5598ffd6a`, with no feature flags.

## Tests

Ten, in `fluree-graph-turtle/tests/prefix_redefinition.rs`. **Eight fail without the fix** (verified by stashing the parser change and re-running). The two that pass either way are `@base` and same-binding redeclaration — the pair that shows the fix is scoped correctly rather than accidentally.

Covered: the verbatim repro; SPARQL-style `PREFIX`; other prefixes left alone; rebinding back and forth; the default (empty) prefix; namespace-only names (`e:`); same-binding redeclaration as a no-op; the pre-seeded-map shape chunked import uses; `@base` unaffected; and a 1000-statement case where a partial invalidation would show as a partial failure.

---

## 2. Name U+0000 in a comment instead of embedding one

`fluree-db-api/src/admin.rs` line 283 carried a literal NUL byte inside a comment that meant to *name* the character — "callers can otherwise sneak a `<NUL>` past". The lines immediately above spell the same class correctly (`U+0000..=U+001F`, `U+007F`); this is the one that got away.

**The cost is not cosmetic.** A single NUL makes the whole file binary to every text tool:

- `file(1)` reported it as `data`, not source.
- Search behaviour then **splits by grep implementation**, and naming which one matters:
  - **ugrep 7.5.0** — what `grep` resolves to in the dev environment this was found in — exits 1 and prints nothing for a pattern occurring 23 times in the file. Not "binary file matches", not a warning: a clean, confident "not found".
  - **BSD grep 2.6.0-FreeBSD** (`/usr/bin/grep`, GNU-compatible; what CI and most contributors get) prints `Binary file … matches` and exits 0. It withholds the matching lines but does not lie about the match.

So the severe reading holds for anyone on ugrep and the milder one for everyone else — both worse than a file that simply works. After: `file` reports UTF-8 text and both greps find all 23. (An earlier draft of this claim asserted the silent-exit-1 behaviour universally; it was measured in one shell. Corrected on review.)

Comment-only and byte-precise — one line, the NUL replaced by the text `U+0000`. 670 `fluree-db-api` lib tests unchanged, full suite re-run at the tip.

Found by the control-byte guard built for the RDF-toolkit work, which flagged this as its second literal NUL; the first was in a Turtle test fixture and is fixed on that branch. Riding here rather than standing alone because a one-byte PR against main is not worth its own review slot.

---

## Gates

| Gate | Result |
| --- | --- |
| `nextest` turtle + graph-ir + json-ld + format + transact + vocab | 720 passed, 0 failed |
| `nextest -p fluree-db-api` (full) | 2929 passed, 11 skipped |
| `clippy --all-targets --no-deps` | 0 warnings |
| `cargo check --all-targets` on excluded `testsuite-sparql` / `testsuite-shacl` | clean |

### DIFF
--- a/fluree-graph-turtle/src/parser.rs
+++ b/fluree-graph-turtle/src/parser.rs
@@ -596,6 +596,25 @@ impl<'a, 'input, S: GraphSink> Parser<'a, 'input, S> {
     /// The cost on the path that matters is nil: one extra hash lookup per
     /// prefix *directive*, of which a document has a handful, against a cache
     /// consulted once per prefixed *name*, of which it has millions.
+    ///
+    /// # The comparison is on meaning, not spelling
+    ///
+    /// `namespace` arrives already resolved — `parse_prefix_directive` runs
+    /// `resolve_iri` before calling this — so `prefixes` holds resolved IRIs
+    /// and this compares what the two declarations MEAN. Both directions of
+    /// that are load-bearing, and neither is obvious:
+    ///
+    /// - Same spelling, different meaning. `@prefix e: <a/>` under
+    ///   `@base <http://x/>` and again under `@base <http://y/>` is the same
+    ///   text naming two different namespaces. Comparing raw text would find
+    ///   them equal and skip a clear that is required.
+    /// - Different spelling, same meaning. `<http://x/a/>` and `<a/>` under
+    ///   `@base <http://x/>` name one namespace. Comparing raw text would
+    ///   clear a cache that was entirely correct.
+    ///
+    /// So storing the resolved form is not an incidental convenience of the
+    /// directive parser; it is what makes this comparison the right one. Both
+    /// cases are pinned in `tests/prefix_redefinition.rs`.
     fn bind_prefix(&mut self, prefix: String, namespace: String) {
         let rebinds = self
             .prefixes
@@ -603,12 +622,28 @@ impl<'a, 'input, S: GraphSink> Parser<'a, 'input, S> {
             .is_some_and(|current| *current != namespace);
         self.prefixes.insert(prefix, namespace);
         if rebinds {
-            // Coarse on purpose. A rebind could strand only the entries for
-            // that one prefix, but finding them means scanning every key, and
-            // a rebind is rare enough that paying O(cache) once beats making
-            // the common path cleverer.
+            // Coarse on purpose: a rebind invalidates only that prefix's
+            // entries, but finding them means scanning every key, so we drop
+            // them all. A rebind is rare enough that paying O(cache) once
+            // beats making the common path cleverer.
             self.prefixed_term_cache.clear();
         }
+    }
+
+    /// Test-only: parse, then report `(prefixed_cache_hits, prefixed_cache_misses)`.
+    ///
+    /// The "only if it rebinds" in [`Self::bind_prefix`] is invisible in parser
+    /// output — clearing on a same-binding redeclaration yields identical
+    /// triples, just colder, because `iri_term_cache` absorbs the repeated
+    /// work before it can reach the sink. Without a view of the counters,
+    /// nothing stops a later simplification from clearing unconditionally and
+    /// quietly costing hit rate on every chunk of every import.
+    #[cfg(test)]
+    fn parse_reporting_cache(mut self) -> Result<(u64, u64)> {
+        while !self.is_at_end() {
+            self.parse_statement()?;
+        }
+        Ok((self.prefixed_cache_hits, self.prefixed_cache_misses))
     }
 
     /// Parse @base or BASE directive.
@@ -1511,6 +1546,48 @@ pub fn parse_with_prefixes_base_options<S: GraphSink>(
 mod tests {
     use super::*;
     use fluree_graph_ir::{Graph, GraphCollectorSink, Term};
+
+    /// Counts for a document whose only prefixed names are two `e:x`.
+    fn prefixed_cache_counts(doc: &str) -> (u64, u64) {
+        let mut sink = GraphCollectorSink::new();
+        Parser::new(doc, &mut sink)
+            .unwrap()
+            .parse_reporting_cache()
+            .unwrap()
+    }
+
+    /// Redeclaring the SAME binding must leave the expansion cache warm.
+    ///
+    /// This is the shape chunked import produces on every chunk, and it is the
+    /// reason `bind_prefix` clears only on a real rebinding. No output pins it:
+    /// the triples are identical either way.
+    #[test]
+    fn a_same_binding_redeclaration_keeps_the_cache_warm() {
+        let doc = "@prefix e: <http://a/> .\n\
+                   e:x <http://p/> \"1\" .\n\
+                   @prefix e: <http://a/> .\n\
+                   e:x <http://p/> \"2\" .\n";
+        assert_eq!(
+            prefixed_cache_counts(doc),
+            (1, 1),
+            "same-binding redeclaration must NOT clear the expansion cache"
+        );
+    }
+
+    /// A real rebinding must drop the cached expansion — the correctness half,
+    /// pinned at the cache rather than only through the emitted triples.
+    #[test]
+    fn a_rebinding_drops_the_cached_expansion() {
+        let doc = "@prefix e: <http://a/> .\n\
+                   e:x <http://p/> \"1\" .\n\
+                   @prefix e: <http://b/> .\n\
+                   e:x <http://p/> \"2\" .\n";
+        assert_eq!(
+            prefixed_cache_counts(doc),
+            (0, 2),
+            "a rebinding MUST clear the expansion cache"
+        );
+    }
 
     fn parse_to_graph(input: &str) -> Result<Graph> {
         let mut sink = GraphCollectorSink::new();
--- a/fluree-graph-turtle/src/parser.rs
+++ b/fluree-graph-turtle/src/parser.rs
@@ -614,7 +614,8 @@ impl<'a, 'input, S: GraphSink> Parser<'a, 'input, S> {
     ///
     /// So storing the resolved form is not an incidental convenience of the
     /// directive parser; it is what makes this comparison the right one. Both
-    /// cases are pinned in `tests/prefix_redefinition.rs`.
+    /// cases are pinned in `tests/prefix_redefinition_adversarial.rs` (A1 and
+    /// A2) — they need a moving `@base`, which the other file never varies.
     fn bind_prefix(&mut self, prefix: String, namespace: String) {
         let rebinds = self
             .prefixes
@@ -626,6 +627,15 @@ impl<'a, 'input, S: GraphSink> Parser<'a, 'input, S> {
             // entries, but finding them means scanning every key, so we drop
             // them all. A rebind is rare enough that paying O(cache) once
             // beats making the common path cleverer.
+            //
+            // To be exact about the trade, since "scanning every key" could be
+            // read as the reason: `retain` scans every key too, so the cost is
+            // the same order either way. What a `retain` would buy is keeping
+            // OTHER prefixes warm after a rebind; what it costs is a predicate
+            // that has to be right — the keys are span texts like `ex:name`,
+            // so the obvious `k.starts_with(prefix)` evicts `ex:name` when `e`
+            // rebinds, and the correct form has to compare up to the colon.
+            // Simplicity over post-rebind hit rate, not cost over cost.
             self.prefixed_term_cache.clear();
         }
     }
--- a/fluree-graph-turtle/src/parser.rs
+++ b/fluree-graph-turtle/src/parser.rs
@@ -37,9 +37,12 @@ pub struct Parser<'a, 'input, S> {
     iri_term_cache: FxHashMap<Arc<str>, TermId>,
     /// Cache of prefixed name span text -> TermId.
     ///
-    /// Keyed by the raw span text (e.g., `"ex:name"` or `"ex:"`), which uniquely
-    /// identifies the expanded IRI for a given prefix mapping. Handles both
-    /// PrefixedName and PrefixedNameNs tokens in one cache.
+    /// Keyed by the raw span text (e.g., `"ex:name"` or `"ex:"`) — which
+    /// identifies the expanded IRI only WHILE the prefix mapping is stable.
+    /// A `@prefix` redefinition changes what a cached span means, so
+    /// [`Parser::bind_prefix`] clears this cache on any rebinding; every
+    /// prefix write must go through it. Handles both PrefixedName and
+    /// PrefixedNameNs tokens in one cache.
     prefixed_term_cache: FxHashMap<Arc<str>, TermId>,
     /// Cache hit/miss counters (recorded on `turtle_parse_events` span).
     iri_cache_hits: u64,
@@ -1546,7 +1549,15 @@ pub fn parse_with_prefixes_base_options<S: GraphSink>(
     if !prefixes.is_empty() {
         parser.prefixes.reserve(prefixes.len());
         for (prefix, namespace) in prefixes {
-            parser.prefixes.insert(prefix.clone(), namespace.clone());
+            // Through bind_prefix, not a raw insert: on an empty cache the
+            // invalidation is a no-op, but this is the entry point chunked
+            // import uses, and any future change that seeds AFTER names have
+            // been cached ("re-seed after the prelude", "top up between
+            // chunks") would reintroduce the stale-cache bug this parser
+            // guards against — with no test able to see it, because seeding
+            // normally happens before anything is cached. Routing it here
+            // makes the invariant structural instead of conventional.
+            parser.bind_prefix(prefix.clone(), namespace.clone());
         }
     }
     parser.parse()
--- a/fluree-graph-turtle/src/parser.rs
+++ b/fluree-graph-turtle/src/parser.rs
@@ -367,8 +367,10 @@ impl<'a, 'input, S: GraphSink> Parser<'a, 'input, S> {
 
     /// Look up a prefixed name (PrefixedName or PrefixedNameNs) by span text.
     ///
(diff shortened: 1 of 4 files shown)


fluree-graph-turtle/tests/prefix_redefinition_blast_radius.rs
```diff
@@ -0,0 +1,221 @@
...
+         in carrier chunk = {wrong_in_carrier}, in later chunks = {wrong_after_carrier}"
+    );
+    println!("post-redefinition objects still resolving to http://a/tag inside the carrier chunk = {stale_objects_in_carrier}");
+
+    // THE assertion that distinguishes the fix. Without it, `e:tag` after the
+    // rebinding is served from the cache under the OLD namespace: this count
+    // is greater than zero (verified by reverting the parser to merge-base).
+    // With it, every one of them resolves to the new namespace.
+    assert_eq!(
+        stale_objects_in_carrier, 0,
```
bplatz, line 169: **This is the one thing I would like fixed before merge: the assertion is vacuous below a 16 KB chunk size.**

It depends on the carrier chunk actually containing post-redefinition statements, and nothing asserts that it does. Measured across chunk sizes on this exact corpus:

```
chunk_size=   1024  chunks=3930  carrier=2000  post-redefinition triples IN carrier = 0
chunk_size=   2048  chunks=3857  carrier=2000  post-redefinition triples IN carrier = 0
chunk_size=   4096  chunks=3711  carrier=2000  post-redefinition triples IN carrier = 0
chunk_size=   8192  chunks=3418  carrier=1795  post-redefinition triples IN carrier = 0
chunk_size=  16384  chunks=2833  carrier=1795  post-redefinition triples IN carrier = 379   <- the value used here
chunk_size=  65536  chunks=   2  carrier=   0  post-redefinition triples IN carrier = 379
chunk_size=1048576  chunks=   1  carrier=   0  post-redefinition triples IN carrier = 2000
```

The definitive check — parser mutated to `if false && rebinds` (fix removed entirely), line 69 changed from `16 * 1024` to `4096`:

```
post-redefinition subjects still on the OLD namespace: in carrier chunk = 0, in later chunks = 2000
post-redefinition objects still resolving to http://a/tag inside the carrier chunk = 0
test result: ok. 2 passed; 0 failed
```

The test passes against a parser with no fix in it. `wrong_in_carrier` two assertions down has the same problem.

This is the failure the module doc describes remediating — "it passed against a parser with the fix entirely removed" — closed along the object axis and left open along the chunk-boundary axis. It also narrows one claim in the PR body: "ordinary chunking changes cannot make it flap" is true of the `wrong_after_carrier > 0` diagnostic, but these two do not flap, they go silently vacuous, which is the worse failure mode.

Suggested fix, one counter: tally post-redefinition triples in the carrier chunk regardless of which namespace they resolved to, and `assert!(… > 0)` before asserting that the resolutions are correct. Non-vacuity then holds structurally instead of being a property of the `16 * 1024` constant.

fluree-graph-turtle/tests/prefix_redefinition_blast_radius.rs
```diff
@@ -0,0 +1,221 @@
...
+    );
+
+    // The fix repairs the carrier chunk. Later chunks remain broken — the PR
+    // says so explicitly and does not claim otherwise.
+    assert_eq!(
+        wrong_in_carrier, 0,
+        "within the carrier chunk the fix must hold"
+    );
+    assert!(
+        wrong_after_carrier > 0,
```
bplatz, line 180: Optional. The technique is right — an `#[ignore]` would rot, and this fails loudly the moment §1.4 lands, which is the signal you want.

But the message is written for the state that holds today, so it prints backwards at the only moment it will ever be read. Whoever closes §1.4 gets a failing assertion reading "later chunks are still mis-resolved" precisely *because* they no longer are. Phrasing it for the failure direction — "if this fails, §1.4 has landed: flip to `== 0` and delete this note" — puts the instruction where the CI output actually shows it. The module doc has exactly that; the assertion message doesn't.

fluree-graph-turtle/tests/prefix_redefinition_blast_radius.rs
```diff
@@ -0,0 +1,221 @@
...
+            f,
+            "e:s <http://p/> \"v\" . @prefix e: <http://b/> . e:t <http://p/> \"v\" ."
+        )
+        .unwrap();
+    }
+    let (_, ds2) = extract_prefix_block(&mid_line).unwrap();
+    let res = compute_chunk_boundaries(&mid_line, ds2, 1024);
+    println!("mid-line directive (data_start={ds2}) => {res:?}");
+    assert!(
+        res.is_ok(),
```
bplatz, line 236: Optional, same shape as the `wrong_after_carrier` note above: when someone teaches `PrefixCheck` to match a mid-line directive, this fails with a message asserting that it doesn't. Worth phrasing for the failure direction.

fluree-graph-turtle/src/parser.rs
```diff
@@ -456,6 +458,93 @@ impl<'a, 'input, S: GraphSink> Parser<'a, 'input, S> {
...
+    ///   them equal and skip a clear that is required.
+    /// - Different spelling, same meaning. `<http://x/a/>` and `<a/>` under
+    ///   `@base <http://x/>` name one namespace. Comparing raw text would
+    ///   clear a cache that was entirely correct.
+    ///
+    /// So storing the resolved form is not an incidental convenience of the
+    /// directive parser; it is what makes this comparison the right one. Both
+    /// cases are pinned in `tests/prefix_redefinition_adversarial.rs` (A1 and
+    /// A2) — they need a moving `@base`, which the other file never varies.
+    fn bind_prefix(&mut self, prefix: String, namespace: String) {
```
bplatz, line 622: Optional. `parse_with_prefixes_base` seeds the map with a direct `parser.prefixes.insert(...)` rather than coming through here, so the invariant this function establishes is a convention rather than a structure.

Safe today — that loop runs at construction against an empty cache, so there is nothing to invalidate. But it is the entry point chunked import uses, and it is the one place a future change ("re-seed after the prelude", "top up bindings between chunks") reintroduces precisely this bug with no test able to catch it, because the seeding happens before any name is cached. Routing that loop through `bind_prefix` costs nothing on an empty cache and makes the invariant self-enforcing.

fluree-graph-turtle/src/parser.rs
```diff
@@ -280,8 +280,10 @@ impl<'a, 'input, S: GraphSink> Parser<'a, 'input, S> {
 
     /// Look up a prefixed name (PrefixedName or PrefixedNameNs) by span text.
     ///
-    /// The span text (e.g., `"ex:name"` or `"ex:"`) uniquely identifies the
-    /// expanded IRI for the current prefix mappings, so it serves as the cache key.
+    /// The span text (e.g., `"ex:name"` or `"ex:"`) identifies the expanded
+    /// IRI **for the prefix bindings in force**, so it serves as the cache key
+    /// only as long as those bindings hold. [`Self::bind_prefix`] is what
```
bplatz, line 375: Optional nit. This is the right correction, but its twin on the `prefixed_term_cache` field declaration (~line 38) still reads "which uniquely identifies the expanded IRI for a given prefix mapping", with no mention of invalidation and no pointer to `bind_prefix`.

Both sentences said the same thing before this PR and only one was fixed — and the field doc is the one a reader scanning the struct hits first, which is roughly how this got written in the first place.

bplatz (approved):
Approving — the fix is correct, minimal, and the reasoning behind it is written down rather than assumed. One ask before merge (inline on the blast-radius test), plus three optional notes.

## Verified by mutation rather than by reading

- **Removing the fix** (`if false && rebinds`, i.e. merge-base behaviour): lib `a_rebinding_drops_the_cached_expansion` fails, `prefix_redefinition.rs` fails **8 of 10**, adversarial fails 6 of 9, blast-radius fails 1 of 2. The body's claim — eight fail, and the two survivors are `@base` and same-binding redeclaration — is exactly right, down to which two survive.
- **Clearing unconditionally** (`if true`): caught **only** by `a_same_binding_redeclaration_keeps_the_cache_warm`. All 21 triple-level tests stay green. That is the whole case for `parse_reporting_cache`, and I would have pushed back on adding a test-only parse path without seeing it.
- Confirmed in code, not just in the prose, that `resolve_iri_term` resolves before calling `sink_term_iri` — so `iri_term_cache` really does key on the resolved IRI and the `@base` immunity claim holds. The sinks are safe too: `PreludeSink` appends in document order (last wins on seed), `FlakeSink` allocates a code per distinct namespace (both are genuinely present in the document), `GraphCollectorSink` only records.

The "only when it rebinds" condition, the resolved-vs-raw comparison argued in both directions, and the coarse whole-cache clear are all the right calls.

## The ask

The two load-bearing assertions in `prefix_redefinition_blast_radius.rs` go vacuous below a 16 KB chunk size — measurements inline. Short version: with the parser fix removed **and** the chunk size changed to 4096, the test passes. It sits one power of two above a cliff, and it is the same vacuity the earlier review already caught once, reproduced along a different axis.

## Optional, worth a quick look

Three inline: the `parse_with_prefixes_base` seeding loop bypasses `bind_prefix`; the two deliberate reverse-assertions print messages that read backwards at the only moment they ever appear; and the `prefixed_term_cache` *field* doc still carries the pre-fix framing that the method doc was correctly moved away from.

## The `admin.rs` half

Right, and the severe reading of it reproduces here. With ugrep 7.5.0 resolving `grep`, on `main`:

```
grep -c "ledger" fluree-db-api/src/admin.rs   ->  (nothing), exit 1
/usr/bin/grep -c "ledger" (same file)         ->  222, exit 0
file(1)                                       ->  "data"
```

A clean, confident "not found" for a pattern that occurs 222 times. It is also the only `.rs` file in the repo carrying a NUL. Worth its ride-along.

aaj3f:
All four landed in 2a97f0c42, the ask verified by your own scenario.

**The ask — structural non-vacuity.** The carrier loop now counts post-redefinition triples in the carrier regardless of how they resolved, and asserts `> 0` before either correctness assertion can claim anything. Verified with your exact reproduction: parser mutated to `if false && rebinds`, chunk size dropped to 4096 — the test now FAILS on the witness ("VACUOUS: the carrier chunk holds no post-redefinition triples…") where it previously passed clean. Non-vacuity is now a property of the test rather than of the `16 * 1024` constant, which was your framing and is the right one. Your point about the PR body also holds: these two assertions didn't flap, they went silently vacuous, and the witness converts that failure mode into a loud one.

**The reverse-assertions** both print for the failure direction now — the §1.4 tripwire says "if this fails, the chunker fix has landed: flip to `== 0` and delete the tripwire note," and the mid-line-directive one says "if this fails, PrefixCheck has learned mid-line directives — flip to `is_err()`." You're right that the only moment those messages are ever read is the moment the old phrasing was backwards.

**The seeding loop** routes through `bind_prefix` with a comment carrying your argument: a no-op on the empty cache today, but this is the entry point chunked import uses, and any future re-seed-after-caching would reintroduce exactly this bug with no test positioned to see it. Structural beats conventional.

**The field doc** now states the invalidation contract and points at `bind_prefix` — and your diagnosis of how the stale sentence survived (the field doc is what a struct-scanning reader hits first, which is roughly how the bug got written) is the part worth keeping.

Suite green at the new head (129/129), fmt/clippy clean. The `admin.rs` NUL rider stands as shipped — your live ugrep reproduction (a confident exit-1 "not found" for a pattern occurring 222 times) is a better demonstration of the severity than anything in the PR body.

aaj3f:
CI red was the #1529 × #1552 seam class recurring here: this branch predates #1552's merge, whose GraphSink protocol turns `finish()` into a trait lifecycle method — so the tests' `sink.finish()` collector calls compiled standalone against the old base while the PR's merge result did not (local suites green, CI red, same mechanism as the materialize break). Fixed by rebasing onto main (zero conflicts; the admin.rs NUL rider auto-dropped as already-upstream) and moving all seven call sites to `into_graph()`. Suite 163/163 at the new head a2071066a, fmt/clippy clean. The vacuity-witness commit from the previous push rides atop unchanged.

PR #1569, opened 2026-07-30
feat(rdf): M1 completion — dataset formats, validation, parallel convert, competitor harness

aaj3f:
# M1 completion: dataset formats, validation, parallel convert, competitor harness

Stacks on **#1555** (RDF toolkit core). Base: `feat/rdf-toolkit-m0` @ `f6f6857de`.

Four workstreams plus the parser fix from **#1565**, integrated into one
branch: 40 commits, 68 files. Together they finish M1.

**All four W3C RDF 1.1 suites are now at 100%, and every known-failure register
is empty.**

| Suite | Before this branch | After |
|---|---|---|
| RDF 1.1 Turtle | 309/313 (98.7%), 4 registered | **313/313 (100%)**, 0 registered |
| RDF 1.1 N-Triples | 55/70 (78.6%), 15 registered | **70/70 (100%)**, 0 registered |
| RDF 1.1 N-Quads | not run | **87/87 (100%)**, 0 registered |
| RDF 1.1 TriG | not run | **356/356 (100%)**, 0 registered |

The register is policed in both directions, so those zeros are load-bearing: a
single regression either re-populates a register or fails the suite.

After this branch `fluree rdf convert` reads Turtle, TriG, N-Triples and
N-Quads, writes all five text syntaxes, validates the terms it reads, recovers
from bad statements, and parses on as many threads as the input allows.

---

# 1. Dataset formats — TriG and the strict line reader (6 commits)

**TriG is a MODE of the Turtle parser, not a fork.** One lexer, one set of term
and statement productions, `ParserOptions::dialect` selecting the grammar — the
same argument that made `CollectionStyle` an option rather than a second
parser. All three block forms are supported (`GRAPH label { … }`, bare
`label { … }`, and a bare default-graph `{ … }`).

The routing is one field. Every emission already funnelled through
`sink_emit_triple`/`sink_emit_list_item`, so a `current_graph: Option<TermId>`
set at the block boundary is what makes *all* the Turtle productions — object
lists, collections, property lists, reification — land in the right graph. It
is saved and restored around each block including on error, so a failure inside
a block cannot leak graph scope into the rest of the document.

The capability probe fires once per named block rather than per triple: a
triple-only sink cannot represent a named graph at all, and discovering that on
the thousandth statement would mean the first 999 were already misplaced.

**The strict N-Triples/N-Quads reader is deliberately NOT the Turtle parser in
a narrower mood.** The character of these formats is what they *refuse* —
directives, prefixed names, relative IRIs, `,`/`;` lists, `'''`/`"""` strings,
bare numeric and boolean literals, collections — and every one of those is
valid Turtle. A reader built by restricting the Turtle parser therefore cannot
reject them, which is exactly why 14 W3C negative-syntax tests were
unenforceable and registered as burn-down cause E. One scanner serves both
grammars, with `LineDialect` deciding whether a fourth term is allowed;
parameterizing is honest there because N-Triples genuinely *is* N-Quads without
the graph label.

This is a **deliberate behavior change for `.nt` input**: a `.nt` file
containing a directive, a prefixed name, a bare number or a long-quoted string
used to be accepted (all of it is valid Turtle) and is now rejected (none of it
is valid N-Triples). That is riot parity, and accepting it would make the verb
agree with no other tool in the field.

`--base` is not threaded into the line formats. They have no base and no
relative IRIs, so a `--base` passed alongside an `.nq` input is inert rather
than silently mis-applied.

**A crash fixed on the way.** `unicode_escape` sliced the 4/8-byte hex window
by *byte* index after checking only that the bytes existed, so a multi-byte
character straddling that window panicked (`fluree rdf check` on a one-line
file, exit 101). The same scan also inherited `u32::from_str_radix`'s
acceptance of a leading `+`, so `\u+041` decoded as `A`. One idea retires both:
require exactly `width` ASCII hex digits *before* slicing or converting — hex
digits are ASCII, so proving the window is all hex digits also proves it ends
on a char boundary. The gate went into the shared function so neither reader
can drift back. The regression tests assert an exact exit code rather than
`!success()`, because a panic also fails `success()` and the weaker assertion
would have passed against the bug it exists to catch.

Also shared, for the same reason: the `ECHAR` escape table, which the line
reader had transcribed independently from the Turtle lexer.

# 2. Term validation (H-8, 7 commits)

Both grammars constrain a document's *source text*. Neither constrains what a
`\uXXXX` escape expands to, nor what base resolution produces — and RDF
requires the *result* to be a term. That is precisely the
`TestTurtleNegativeEval` class: the document lexes, and the term it denotes is
not a term. Raw forbidden characters were already rejected by the lexer; only
escaped ones slipped through, which is why this needed a post-expansion check
rather than a tighter grammar.

`fluree-vocab` gets two pure predicates beside `resolve_iri`:
`iri::iri_violation` applies the `IRIREF` exclusion set and then absoluteness,
and `lang::language_tag_violation` applies the `LANGTAG` production. Both are
allocation-free on the passing path.

Two deliberate limits, both to avoid over-rejection — the standing hazard for a
conformance parser, invisible to negative tests and caught only by the positive
suites. **U+007F is not excluded**, because the `IRIREF` production does not
exclude it. **No BCP 47 length rules and no registry lookup**, because the RDF
grammars impose neither; the function is named for well-formedness, not
validity.

**`ParserOptions::validate` is off by default and on in `conformant()`, and off
is the load-bearing half.** Bulk import runs the default on the
2M-statements/second path over documents this database itself wrote, so it must
not pay for a scan of every resolved IRI. It is pinned as *behavior* — the
ingest test parses all five bad documents successfully — not merely as a flag
assertion.

The error offset blames the **token**, not the offending character: expansion
changes the string's length, so an index into the resolved value maps back to
nothing in the source. The message carries the finer index instead, and a test
asserts the two genuinely differ.

**Language tags are validated in the parser, not the lexer.** `@BASE` must keep
lexing as a `LangTag` so the parser can reject it in *directive* position —
Turtle's `@`-directives are case-sensitive, which is burn-down cause B, already
fixed. A lexer-level check would silently undo it.

## BOM tolerance — an ungated behavior change, disclosed

A leading U+FEFF is **consumed, not stripped**: token spans are absolute offsets
into the original source, so removing the three bytes would shift every one of
them. Pinned by asserting a diagnostic still lands on the right line and column
with a BOM in front. Only at the start; elsewhere U+FEFF is a zero-width
no-break space and a real lexical error.

**This one is NOT behind `ParserOptions`, so it changes bulk import too.** A
BOM-prefixed Turtle file that import previously rejected with "unexpected
character `\u{feff}` at line 1, column 1" now loads. That is almost certainly a
fix — Windows editors emit BOMs routinely, riot and every other production
reader tolerate a leading one, and the old failure told a user nothing useful
about their file. But it is a change in what the shipping import path accepts,
it is not opt-in, and it is not what H-8 was chartered to do, so it is called
out here rather than left for someone to find in a changelog. Gating it would
mean threading `ParserOptions` into the lexer, which nothing else needs; the
alternative shape, if the team prefers strictness, is to reject in import and
accept in the `rdf` verbs.

## `--nocheck`, and why the profile reports it

`--nocheck` turns off term validation and nothing else, pinned by a test that
feeds it a syntax error and still expects exit 1. All three verbs share one
`verb_options()` so they cannot describe different parses.

`--profile` carries `validated`, because a `--nocheck` run and a validating run
are not comparable measurements. Every other RDF tool worth benchmarking
against validates by default, so an unlabelled `--nocheck` number would be a
faster answer to an easier question.

# 3. Parallel convert (10 commits)

**Workers produce bytes.** Each worker runs its own writer into a byte buffer
and the driver concatenates in chunk order. There is no shared relabeller and
therefore no serial replay, which was roughly half the wall clock of the
collect-then-replay design this replaces.

What makes that sound is a label scheme that is a pure function of (label,
chunk), so workers need no coordination: a user's `_:L` becomes `u{L}`
(prefixing is injective, so distinct labels stay distinct and the same label in
two chunks maps to one output label, which is what document scoping requires);
an anonymous node in chunk `c` becomes `g{c}_{n}`; and `_:fdb-…` passes through
verbatim, keeping #1432's addressability contract.

**The chunker now runs the real byte-level scanner.** The streaming reader was
line-based with an ad-hoc triple-quote toggle: it split mid-statement on `#`
inside a line-final comment, mis-toggled on `'''` inside short strings, and did
not detect mid-file directives at all. It now drives the same `BoundaryScanner`
the pre-scan path uses, and detects a directive at **any token start** — the
earlier guard was case-sensitive, single-candidate and whitespace-anchored.

A mid-file directive makes a document unchunkable, because only the first chunk
would carry the redefinition. Per plan §1.4 that is a **fallback to serial**,
not a refusal: the document is legal Turtle and must still convert.

**Two bugs found by review, both fixed rather than documented.** Parallel
convert *deadlocked* on a dead destination — the writer stopped draining on
EPIPE while workers stayed blocked on a full bounded channel, and
`thread::scope` joined them forever at 0% CPU, so any input over the parallel
threshold piped to `head` hung. And `--bnode-policy preserve` silently
relabelled under parallelism; it now forces the serial path and reports why.

`--continue-on-error` is serial by construction: resync needs the document as
one sequence of statements, and a chunk boundary is not a place a skipped
statement can be reasoned about.

## G1b-2 measurement, with its caveats attached

Dev-host numbers at `48444c026`, workers-produce-bytes, **lower bounds** — read
them with every caveat below, not as a claim.

- K=1: 6.54 s = 918K triples/s native end-to-end. K=16: 3.38 s = **1.94×**
  (up from 1.36× under the replaced architecture), reproduced next morning
  within 2%.
- **The real finding is not the 1.94×.** Pool width is 14.80× at K=16 —
  **92.5% of ideal** — so plan §1.3 works. The cap is the **single-threaded
  chunk scan**: 2.85 s, K-independent, 19.6 ms/MB dead-linear (~52 MB/s, about
  10× slow for a byte DFA). The model `wall(K) ≈ 2.85 + 6.49/width` fits the
  measurements.
- **Asymptotic ceiling on this corpus is 2.29×**, so more threads are now worth
  nothing. The only lever is the pre-pass.
- Caveats that bound all of it: dev host, not the m8gd publication locus; an
  idle floor of ~5.7 is disclosed rather than subtracted; the scan was 43% of
  wall and was charged to `unattributed`, which is why a `Phase::Chunk` lane is
  queued.

**Nothing here is a publishable number.** Per plan §5/R5, nothing publishes
before the official m8gd run.

# 4. Tier-2 competitor harness (6 commits)

Out-of-band bench tooling in `benchmarks/conversion/`, an excluded workspace so
it never lengthens `cargo check --workspace` for anyone else. Its review pass is
worth reading as methodology, because every fix bites the publication run
specifically.

**The instrument was measuring itself.** Timestamps were taken with
`python3 -c` on each side of the interval, putting two interpreter startups
*inside* every measured region — 26–40 ms for the pair, measured. On a corpus
where serdi does ~10 ms of real work that is not a perturbation, it is most of
the number, and being near-constant it compressed every ratio toward whichever
tool was slowest. Replaced with bash's `EPOCHREALTIME` builtin, which forks
nothing.

**A locale could zero every cell silently.** Timings were derived by stripping
`EPOCHREALTIME`'s decimal separator, so under a comma-decimal locale the strip
is a no-op, awk parses `1785358343,127180` as `1785358343`, and every cell
reports 0.000000 s **with status ok**. Reproduced before fixing: a 300 ms sleep
measures as `0,000000 s` under `LC_ALL=de_DE.UTF-8`. The harness now forces
`LC_ALL=C`, and asserts the dot-decimal property the arithmetic relies on
rather than trusting the export.

**Architecture is recorded per tool, per cell.** The earlier translated-shell
finding was corrected: a translated shell does *not* translate its children, so
"the shell is emulated" does not license "these numbers are emulated". The real
hazard is a competitor shipping x86_64-only, because then one column is emulated
and another is not and the matrix reports the emulator as a performance
difference. Two of the five classes are admissions rather than answers —
`translated-x86_64` and `universal-indeterminate` — and riot lands in the
second, being a shell script over a universal JVM. Measured on this host, serdi,
rapper and oxigraph are all single-slice native arm64.

**Ground truth is recounted from the output file** rather than trusted from the
runner's own `out_statements` field, with a disagreement between the two flagged
as a bookkeeping mismatch distinct from a wrong count — a producer's claim about
its own output is not evidence about that output.

Load per core is recorded per **cell** at measurement time, so a build that
starts midway through a matrix contaminates only the cells after it instead of
being averaged away.

# 5. #1565's parser fix, and the two tests it flips

**A redeclared prefix must change what later names mean.** The parser caches
expanded prefixed names by their span text (`ex:name`), which identifies an IRI
only relative to the bindings in force — so a stale cache turned a rebinding
into a silent no-op. Fixed on `main` as #1565; the effort branches predate it,
and landing it here completes §1.4's serial-fallback story.

It was **built to flip two tests**, both of which pinned defects rather than
properties so they would fail the moment the chunker work landed and send
whoever closed the gap to them. Both fired, and both are now flipped to the
closed behavior:

- `the_guard_misses_a_mid_line_directive` → `..._catches_...`: detection now
  fires at any token start.
- The blast-radius test asserted the streaming reader *accepts* a mid-file
  redefinition and delivers it as chunk data. It now refuses it.
- `a_mid_file_directive_falls_back_to_serial_rather_than_refusing` asserted
  that exactly one term after the rebinding denotes the new namespace. It is
  **two**, and the second one is the fix: the trailing statement is
  `ex:z ex:p "z" .`, and before #1565 only the fresh subject moved while the
  predicate — expanded 80,000 times already — came back from the cache under
  the old binding. Asserting one was asserting the bug.

**One thing worth knowing about that refusal**, because the old test made it
look alarming. When the reader thread raises `PrefixAfterData` it stops, and
`recv_chunk` reports the closed channel as a benign `Ok(None)` — so a consumer
that only drains sees a short read and no error (1797 of 4000 statements on the
test corpus). The error lives on the reader thread and only `join()` surfaces
it. **Every live import call site joins**, and `import.rs` carries a comment
saying why, so there is no data-loss path in the product; the old test simply
never joined. The replacement asserts drain-then-join returns
`Err(PrefixAfterData)`, which is the assertion that tells "refused" apart from
"silently truncated".

`ImportSource::recv_next` returns `Ok(None)` without joining. It has no callers
today; a future one would inherit the silent-short-read shape. Noted, not
fixed.

**#1565's `admin.rs` rider is included** (a comment-only one-line change
replacing a literal NUL with the text `U+0000`). It is not optional here: the
bench line ships a control-byte guard over every git-tracked text file, and the
plan records that guard as RED until both NUL fixes land. The other fix is on
the h8 line. Cherry-picking a change that also sits on a branch targeting `main`
is harmless — an identical patch is dropped on rebase — and the alternative was
a red gate.

---

# Integration work beyond the merge

The four lines were built in parallel and had never met. These are the commits
that only exist because they did.

**The TriG register emptied where two workstreams met.** Its last four entries
were `TestTrigNegativeEval` tests, so they read as eval-machinery failures and
were registered under "no N-Quads reader". They were nothing of the kind: three
ill-formed IRIs behind `\uXXXX` escapes and one bad language tag — the exact
four-test shape the Turtle suite carried under cause C. The strict line reader
closed the eval cause and left them behind under the wrong one; term validation
then closed them, reaching TriG for free because the TriG parser is the Turtle
grammar with a dialect flag. That is the argument for one grammar with a
dialect knob rather than a second parser, and it is the evidence for it.

**Both readers now call the same term predicates** (a plan MUST-DO). The Turtle
parser called `fluree-vocab`'s predicates; the strict reader carried its own
absoluteness check and its own hand-rolled `LANGTAG` loop. The IRI half was a
real hole rather than duplication: the reader checked the *source* bytes for the
exclusion set, but `<http://example.org/a b>` is legal source that expands
to a space, so the byte scan cannot object and absoluteness has no opinion. The
language-tag half is now scanner-delimits, predicate-decides. Both are pinned by
tests verified to fail when the wiring is reverted.

**The fourth IRIREF transcription is bound.** A full-Unicode differential
already tied the reader's accept-set, the validator's reject-set and the
writer's escape-set together, and recorded that a fourth copy in
`fluree-db-sparql` was deliberately uncovered because a light crate cannot
depend on the SPARQL engine. It is now bound from `fluree-db-cli` — the smallest
crate already depending on both sides — as a **dev**-dependency, so it costs no
production edge and the copy can stay where it is. Verified by mutation.

**A documented flag-interaction table**, with a test per row in *both* the
serial and the parallel path. Reviewers found three separate cases of one flag
silently overriding another, and three of a kind is a pattern: interactions were
being discovered rather than specified. The standing rule is in `convert.md` — a
new flag adds a row, and if it interacts with nothing it says so, because
"nothing" is a claim a reader can check and an absent row is not.

**The harness's own negative self-test needed re-anchoring.** It proved the
register is enforced by running Turtle in conformant mode against an empty
register and requiring an error — which worked only while the parser had
conformance failures. At 313/313 it returned `Ok` and the check that guards
every other suite was itself red. It now drives `ParseMode::IngestDefault`,
whose failures are a design property nobody is trying to remove. It surfaced
only because the gate battery runs plain `cargo test`; `--test w3c_rdf` never
compiles the `--lib` tests where it lives.

Four smaller fold-ins recorded at the wave closes are applied: the dead-reader
deadlock comment (either mechanism alone suffices — keep both; the mutation that
proves it removes both), a test name that encoded a retired reason, a
byte-slicing caveat on `unicode_escape_value`, and binding the CLI tests' exit
code to the real contract instead of a literal.

## Review provenance

Each of the four lines went through build → adversarial review → fix → verify
with independent mutation and probe verification before integration, and the
fixes are folded in rather than listed as future work. Review changed the
shipped result repeatedly and concretely: the bench harness's clock was
rebuilt after a reviewer showed it was timing two Python startups; the parallel
path's deadlock and its silent `--bnode-policy` override were both found by
review, not by the suite; the `\u` panic was reproduced from a one-line file
before being fixed; and the `@base` fix from the previous wave was found
incomplete through the API and extended.

## CI does not fire on this PR

The base is `feat/rdf-toolkit-m0` (#1555's branch), not `main`, and this repo's
CI triggers only on `main`-based PRs — verified in `ci.yml`, which is
`branches: [main]` on both `push` and `pull_request`. Nothing here has been
exercised by GitHub Actions. The local battery below stands in for it.

| Gate | Result |
|---|---|
| `nextest` — graph-ir, turtle, json-ld, format, vocab, transact | **1061 passed**, 0 failed |
| `nextest -p fluree-db-api` (full — shared-parser blast radius) | **2923 passed**, 0 failed, 11 skipped |
| `nextest -p fluree-db-cli -p fluree-db-docs` | **508 passed**, 0 failed |
| `nextest -p fluree-db-sparql` | **625 passed**, 0 failed |
| `cargo test --doc` (ir, turtle, format, vocab) | 18 passed |
| `testsuite-rdf` — full `cargo test` (lib + suites) | **32 + 7 passed**, 0 failed |
| W3C RDF 1.1 Turtle / N-Triples / N-Quads / TriG | **313/313 · 70/70 · 87/87 · 356/356**, all registers empty |
| RDF 1.1 Turtle, ingest default (informational) | 281/313 — `conformant` is worth 32 tests |
| `cargo check --workspace --all-targets` | clean |
| excluded `testsuite-sparql` / `testsuite-shacl` / bench harness — `check --all-targets` | clean / clean / clean |
| `testsuite-rdf` — `clippy --all-targets -- -D warnings`, `fmt --check` | clean / clean |
| `clippy --all-targets --no-deps` × 8 crates (ir, turtle, format, vocab, cli, api, sparql, bench-support) | **0 warnings** |

**One pre-existing flaky test, disclosed.**
`fluree-graph-ir::timing::the_sampling_error_bound_is_reported_and_grows_with_dispersion`
compares measured dispersion between a uniform and an erratic sink, and under
full-suite parallelism the uniform sink's busy-wait can be preempted enough to
invert the comparison. Measured at **1/12 full `-p fluree-graph-ir` runs on the
base `f6f6857de`** and 2/12 at this tip — the same rate within noise, and this
branch does not touch the sampling code (the only change to `timing.rs` here
adds two `Phase` variants). It belongs to #1555's line, not this PR, and is
recorded rather than papered over.

## Follow-ups recorded, not silently dropped

- **Wave-4 lever:** the single-threaded chunk scan is the parallel ceiling.
  A 500 MB/s scan puts K=16 at roughly 8.9×, and the scan is resync-able so it
  can itself be parallelized at K offsets.
- `Phase::Chunk` lane — the scan is 43% of wall currently charged to
  `unattributed`. One line.
- `ImportSource::recv_next` joins nothing (no callers today).
- A truly quiet measurement box, or the m8gd locus, before any published cell.
- `--pretty` buffered Turtle tier (H-8) is still unimplemented and refuses
  honestly.
- RDF/XML, RDF/JSON and Jelly are M3/M4; `canon`, `infer` and `query` are M5.


aaj3f:
Surface note: the `fluree rdf <verb>` spelling shown in this diff is dissolved into flat top-level verbs (`parse` / `count` / `convert`) at the top of the stack — #1578's final five commits — per the naming discussion on #1552. The merged result is flat.

PR #1571, opened 2026-07-30
CLI agent-surface truth + the manifest rail

aaj3f:
## Why

Fluree AI (fluree/solo) already hands its users' AI assistants off to this CLI — a generated build prompt, LLM-targeted served docs — and the embedded `docs/` corpus is what agents consume as version-exact truth via `fluree docs` and the MCP `docs` toolset. This PR fixes what that corpus gets wrong, closes the CLI-side credential leak, and builds the rail that lets every downstream teacher of CLI grammar (fluree/solo's UI and docs, the upcoming `fluree-cli` Claude Code plugin) validate what it teaches against what actually ships.

## What

**Doc truth in the embedded corpus** (`2dc6ffb`, part of `4dfc…docs/ai` below)
- All five policy examples that put the `?$this` correlation clause under a **nonexistent `$where` key** — silently dropped by the parser, so every documented "restriction" policy allowed everything — rewritten to the single correlated `where`-array form the passing tests use, with a callout on the silent-allow failure mode and `--track-policy` verification.
- `--memory-budget-mb` auto is documented as **80%** of RAM (code truth; docs said 60%) with an explicit co-residency OOM warning.

**Credential hygiene** (`c497bbe`)
- `fluree config list` redacts credentials as `[redacted]` by default (it dumps `.fluree/config.toml`, which carries live access **and refresh** tokens after `auth login`) — same `redact_json_secrets` policy and convention the graph-source `info` path already uses. `--reveal` is the documented raw escape hatch; `config get <key>` stays raw; `auth token` remains the scripting path. `refresh_token` joins `SECRET_CONFIG_KEYS`.

**The manifest rail** (`438bd51`, `071e0ed`)
- Hidden **`fluree manifest`** emits a machine-readable JSON of the CLI surface — command paths, flags, value enums, positionals, compiled features — generated by walking the clap tree so it cannot drift from the binary. Structure only, deliberately no help text. cargo-dist `extra-artifacts` publishes it as **`fluree-cli-manifest.json`** on each release for consumer CI (fluree/solo's `cli-snippets` test, the plugin's contract check).
- It was born `__manifest` and renamed the same day: the `completions_bash` integration test caught that clap_complete's bash generator uses `__` as its command-path separator and **panics** on a command with that literal name. The name is what consumers script against forever, so the collision-free name lands before anything consumes it.
- `fluree remote add` reads the optional `cli.min_version` from `/.well-known/fluree.json` (served by the paired fluree/solo PR) and warns when this binary is older than what the stack's docs teach; lenient semver compare, malformed advertisements never warn. The discovery-failure fallback that silently guessed `<url>/fluree` now says so and names the fix.

**Docs-coverage gate** (`221d01a`)
- `tests/docs_coverage.rs` walks `Cli::command()` and fails when a command lacks a `docs/cli` page, a page outlives its command, a nested action is never mentioned in its parent page, the README index skips a command, or `SUMMARY.md` orphans a page (orphans are still *embedded* — served to agents while invisible on the website). Adding a subcommand previously required zero doc changes to pass CI. The gate immediately caught real gaps on `main`, all fixed here: the missing `fluree branch revert` section, **7 of 43** commands unlinked from the README index, and `load.md` + `server-integration.md` orphaned from the book.

**Agent guide** (last commit)
- `docs/ai/README.md` un-staled (the CLI MCP bullet omitted the entire `docs` toolset; the vector claim now states the HNSW index is behind the non-default `vector` feature while inline similarity functions are in every build).
- New **`docs/ai/claude-code.md`**: the versioned operating guide for agents driving this CLI — probe-first doctrine, machine-output map, remote/device-flow choreography, credential hygiene, destructive-op confirmation rules, shared-machine budgets, policy correlation. It ships in the embedded corpus precisely so agent-side packaging defers to it instead of copying it.

## Cross-repo pairing

- **fluree/solo#893** (`feat/cli-ai-surface`) is the other half: serves `cli.min_version` in `.well-known/fluree.json`, fixes the platform-side handoff defects, centralizes its UI's CLI grammar into `ui/lib/cli-snippets.ts`, and carries an env-gated test that validates those snippets against this PR's manifest once a release ships the asset. Either PR can land first; the handshake and the manifest check both degrade silently until both sides exist. (Post-review: the manifest now emits a root `path: []` entry with `--version` exposed — the producer-side resolution both reviews preferred — so solo’s declared root entry validates as-is.)
- **fluree/claude-plugins**: a `fluree-cli` Claude Code plugin (PR to follow) whose contract file validates against the same release-asset manifest and whose skill defers to this PR's `docs/ai/claude-code.md`.

## Validation

`cargo fmt` clean; `clippy -p fluree-db-cli --all-targets --no-deps` **0 warnings**; full crate suite green (165 lib + 120 integration + 5 docs-coverage); `fluree-db-api` redaction tests green. Honest caveat: the dist `extra-artifacts` hook is config-only here and gets its first real exercise on the next release CI run.

## Follow-ups (deliberately not here)

- `--json` for `list`, `info`, and `show` (the last human-only outputs an agent needs).
- Convert plain-string `--format` flags to clap value enums so their choices surface in `--help` and the manifest.
- A `repository_dispatch` from release CI to fluree/solo + fluree/claude-plugins so a breaking CLI change surfaces in consumer CI the day it ships.
- An integration test pinning the exact cookbook policy strings (the corrected form is test-proven; the doc-string pin is polish).


aaj3f:
First live use of the manifest rail, during plugin-contract authoring: the checker caught that `fluree list` has no `--json` (only `--remote`) — a fact I had wrong from a mis-scoped grep. So the machine-output follow-up here should read: `--json` for `list`, `info`, and `show`. The contract + skill in the plugin PR ship corrected.

fluree-db-api/src/ledger_info.rs
```diff
@@ -2381,6 +2385,36 @@ impl<'a> LedgerInfoBuilder<'a> {
...
     use super::*;
 
+    /// Pins redaction of every secret-bearing key the CLI writes into
+    /// `.fluree/config.toml` (`RemoteAuth` in fluree-db-nameservice-sync) —
+    /// that struct and `SECRET_CONFIG_KEYS` live in different crates with
+    /// nothing else tying them together, so this test is the tie: a new
+    /// credential field on `RemoteAuth` must be added to the list AND here,
+    /// or `fluree config list` leaks it on the next release.
+    #[test]
+    fn redacts_every_cli_config_credential_key() {
```
bplatz, line 2552: The doc says this is the tie between `RemoteAuth` and `SECRET_CONFIG_KEYS` — *"a new credential field on `RemoteAuth` must be added to the list AND here, or `fluree config list` leaks it on the next release."* But it's a hand-written `json!` literal, so adding `id_token: Option<String>` to `RemoteAuth` leaves this green and ships the leak. The intent is right; the mechanism doesn't achieve it.

A real tie belongs in fluree-db-cli, which has both types in scope: construct a fully-populated `RemoteAuth`, serialize it, and assert that no field whose name contains `token`/`secret`/`password` survives redaction. That fails automatically on the next credential field.

Today's coverage is correct — `RemoteAuth`'s current fields are `auth_type`, `token`, `issuer`, `client_id`, `exchange_url`, `refresh_token`, `scopes`, `redirect_port`, `login_flow`, and the two secrets are both caught. This is purely about the next one.

fluree-db-api/src/ledger_info.rs
```diff
@@ -988,6 +988,10 @@ const SECRET_CONFIG_KEYS: &[&str] = &[
     "secret_access_key",
     "session_token",
     "default_val",
+    // OAuth refresh tokens (the CLI's `.fluree/config.toml` stores one per
+    // remote under `remotes.auth.refresh_token`; `fluree config list` redacts
+    // through this same list).
+    "refresh_token",
```
bplatz, line 1017: Context for why the test above matters: `is_secret_config_key` is an **exact** lowercased match, not a substring one — despite the doc comment above this list saying keys match "case-insensitively anywhere in a stored graph-source config JSON tree." That's exactly why `refresh_token` needed its own entry rather than being caught by `token`.

So `id_token`, `access_token`, `api_token`, and `bearer_token` would all print in the clear today. Either tighten the comment to say exact-match, or switch the predicate to substring for the `*_token`/`*_secret` family — but if you do the latter, check it doesn't start redacting the identifying fields the note deliberately preserves.

fluree-db-cli/src/commands/config_cmd.rs
```diff
@@ -87,10 +87,25 @@ fn list_toml(config_path: &Path) -> CliResult<()> {
...
+    let doc = if reveal { doc } else { redact_toml(doc)? };
     print_toml_flat("", &doc);
     Ok(())
 }
 
+/// Redact credential values in a parsed config for display, reusing the same
+/// key policy as graph-source info (`redact_json_secrets`). Round-trips
+/// through JSON: the only lossy case is a TOML datetime becoming a string,
+/// which prints identically in the flat listing.
+fn redact_toml(doc: toml::Value) -> CliResult<toml::Value> {
```
bplatz: The comment says *"the only lossy case is a TOML datetime becoming a string."* It doesn't — the round-trip is lossless. I checked:

```
JSON = {"a":{"$__toml_private_datetime":"1979-05-27T07:32:00Z"},"b":"x"}
BACK OK = Table({"a": Datetime(...), "b": String("x")})
```

toml's magic-key representation survives the JSON hop and deserializes back to a real `Datetime`. Behavior is better than documented — just fix the comment.

Tiny style note while you're here: this does `eprintln!` inside a function named `redact_toml`, which makes a pure-looking helper unusable non-interactively. Returning a "did redact" bool and printing at the call site would keep it reusable.

dist-workspace.toml
```diff
@@ -26,6 +26,32 @@ publish-jobs = ["homebrew"]
...
+#
+# Deliberately a DEBUG build: extra-artifacts run in the global-artifacts job
+# on the default 4-core/16GB hosted runner with a cold cache, and this
+# workspace's release profile is fat-LTO + codegen-units=1 — the exact
+# configuration the runner note below records as OOMing at link time. The
+# manifest is pure clap-tree introspection; optimization buys nothing.
+# It describes the shipped surface because neither this config nor release.yml
+# passes --features (both get the default set); if a features key is ever
+# added to the dist config, this build must get the same one.
+[[dist.extra-artifacts]]
```
bplatz, line 42: The invariant this comment names is the right one, and I confirmed there's no `features` key in this config today, so the debug manifest build and the shipped release binary do agree on the default feature set.

The residual risk is that the invariant is enforced by a comment in a TOML file, and the failure mode is silent — the manifest would confidently describe a surface the binary doesn't have. Since the manifest already emits its `features` array, having consumer CI cross-check that against the release binary would turn the comment into an assertion. (No platform concern: there's no `cfg(unix)`/`cfg(windows)` gating in the CLI surface, so one manifest legitimately covers all four targets.)

fluree-db-cli/tests/docs_coverage.rs
```diff
@@ -0,0 +1,165 @@
...
+    );
+}
+
+#[test]
+fn readme_indexes_every_command() {
+    let readme = read(&docs_dir().join("cli/README.md"));
+    let missing: Vec<String> = visible_top_level()
+        .iter()
+        .map(|c| c.get_name().to_string())
+        .filter(|name| !readme.contains(&format!("({name}.md)")))
```
bplatz: Matching the literal `({name}.md)` means a link written `(./query.md)` or `(query.md#section)` silently passes the gate while the index is actually broken. Fine as a first cut given the README's current style — worth a one-line comment noting the assumption, since the whole point of this suite is that the gate is trustworthy.

Also on the module header: it credits `--all-features` for feature-gated commands being present. True in CI, but the real floor is that `server`, `shacl`, `iceberg`, and `aws` are *default* features for this crate — so a plain `cargo test -p fluree-db-cli` works too, and only `--no-default-features` breaks it. Worth saying, so nobody "fixes" it by adding a feature guard that isn't needed.

fluree-db-cli/src/cli.rs
```diff
@@ -2175,7 +2190,14 @@ pub enum ConfigAction {
...
 
     /// List all configuration values
-    List,
+    ///
+    /// Credential values (tokens, refresh tokens, client secrets) print as
+    /// `[redacted]`; pass `--reveal` to print them in the clear.
+    List {
+        /// Print credential values in the clear instead of `[redacted]`
+        #[arg(long)]
+        reveal: bool,
```
bplatz, line 2445: Note that `ConfigAction::List` → `List { reveal }` is a breaking change to a `pub enum` in a `pub` module — anything matching on it outside this crate stops compiling. Almost certainly fine given who consumes `fluree_db_cli::cli`, just calling it out so it's a decision rather than a surprise.

The default-redacts-with-an-opt-out shape itself is the right call, and routing it through the same `redact_json_secrets` policy as graph-source `info` is better than a second key list.

bplatz (approved):
Approving. Genuinely good work.

The `__manifest` → `manifest` rename story is exactly the kind of thing you want caught before consumers script against a name forever — clap_complete's bash generator using `__` as its path separator is not something you'd find by reading. The manifest module's "Also deliberately NOT a complete surface description" paragraph — naming `default_value`, `num_args`, `requires`/`conflicts_with` as omitted, and warning that `required` isn't the whole truth — is unusually honest and will save a consumer real debugging. `cli_min_version_warning` is pure, correctly numeric (4.10.0 > 4.9.0, not string-compared), and tested on malformed input. The `remote add` guess warning fixes a real silent-404 failure mode.

A few things flagged inline — all minor, none blocking. Worth a look to decide which you want to take now vs. fold into the follow-ups you already listed.

fluree-db-api/src/ledger_info.rs
aaj3f: You're right — the intent was the tie, but a hand-written `json!` literal can't be one; a new credential field would sail past it. Took your mechanism in `33d10e66e`: fluree-db-cli (the one crate with both types in scope) now builds an **exhaustive** `RemoteAuth` struct literal — deliberately no `..Default::default()` — serializes it, redacts, and requires every credential-family key to come out `[redacted]`. So the next field on `RemoteAuth` fails to *compile* in that test, which routes its author through the redaction question rather than past it. Kept the api-side literal test too, but demoted to what it actually is: a wire-shape pin.

fluree-db-api/src/ledger_info.rs
aaj3f: Good catch on the comment/behavior mismatch — 'anywhere in the tree' was meant as depth, but I agree it reads as substring, and the family gap was real either way (`id_token`/`access_token`/`bearer_token`/`api_secret` all printed in the clear). Went with your second option in `33d10e66e`: exact list **plus** `*_token`/`*_secret`/`*_password` suffix matching, so the families redact by default. Checked it against the deliberately-preserved identifying fields — none carry the suffixes, and endpoint-ish keys like `token_url` stay visible (pinned by a new test alongside the leak cases). Comment rewritten to say exact-plus-suffix.

fluree-db-cli/src/commands/config_cmd.rs
aaj3f: Thanks for actually running the round-trip — behavior better than documented is a nice failure mode to have, but the comment was still wrong. Fixed in `33d10e66e`: it now describes the lossless magic-key form. And took the style note: `redact_toml` returns `(doc, redacted)` and the `eprintln!` lives at the call site, so the helper is honest about being pure.

dist-workspace.toml
aaj3f: Agreed on where this belongs — the comment names the invariant but can't enforce it, and the failure mode is exactly the confident-wrong-manifest you describe. The consumer-CI cross-check (assert the manifest's `features` array matches probes against the actual release binary — `validate`/`cluster` presence is enough to distinguish the gated set) is now tracked on fluree/claude-plugins#2 alongside the other flip-day items, since that repo's CI is the one that already downloads the release artifacts. Good to have it confirmed there's no platform gating to worry about — one manifest per release stays legitimate.

fluree-db-cli/tests/docs_coverage.rs
aaj3f: Both taken in `33d10e66e`. The link check now accepts `(name.md)`, `(./name.md)`, and `(name.md#anchor)` with a comment naming the assumption — you're right that the gate's whole value is being trustworthy, and 'linked in one house style' wasn't the claim it should make. And the module header now credits the real floor: `server`/`shacl`/`iceberg`/`aws` are default features, so plain `cargo test -p fluree-db-cli` sees the full surface and only `--no-default-features` breaks it — said explicitly so nobody adds an unneeded feature guard.

fluree-db-cli/src/cli.rs
aaj3f: Fair flag — making it a decision rather than a surprise: it's deliberate. I checked the one external consumer of our git-deps, fluree-quorum, and it doesn't depend on `fluree-db-cli` (it pins the api/nameservice crates), so the break lands on nobody. If the cli crate ever grows real library consumers we should probably stop treating its enums as wire surface anyway — but that's a conversation for when one exists.

PR #1572, opened 2026-07-30
chore(memory): memory-store baseline audit + checkpoint audit tooling

aaj3f:
This started as the overdue housekeeping pass on `.fluree-memory/repo.ttl` and grew (deliberately, one PR — no stacking) into both halves of the problem: reconciling the store we have, and shipping the tooling so it never needs a heroic cleanup again.

## Part 1 — the audited baseline

The committed store is what every contributor's agent loads for context-grounding, and it had drifted a long way from "durable repo knowledge": of 265 repo-scoped memories (244 in my working set + 21 added on main since), a large fraction were effort diaries — PR-stack checkpoints, "SHIPPED as #NNNN" notes, a deployed-stack snapshot with account IDs, session hand-off pointers, refs into scratch files and other people's home directories, four memories with literal serialized tool-call fragments corrupting their bodies. Roughly a quarter carried at least one `artifactRef` that no longer resolves at HEAD.

The part that made this urgent rather than cosmetic: recall is BM25 + re-rank, so every stale checkpoint dilutes ranking for memories that matter — and a confident memory asserting something the code no longer does is worse than no memory. The verification pass proved it: of ~60 nontrivial claims re-checked against the tree (agent-assisted, five parallel passes with exact-symbol evidence), more than twenty were materially wrong at HEAD — three "known gap" memories described gaps since closed (`f:shapesSource`, `targetSubjectsOf` indexing, `f:schemaSource`/`f:rulesSource`), the decimal-corruption memory predated the exactness fixes, two remembered env switches no longer exist, one flagged a CLI secret leak that #1534 already fixed.

Result: 265 → **214** memories. 147 kept, 59 rewritten in place to current-at-HEAD truth (same IDs, per the supersession model), 59 retired (11 folded into stronger survivors), 8 baseline memories added where the gap pass found uncaptured durable knowledge (`grp_*.rs` test aggregation, the binary-store gate on fast paths and why W3C-green never proves one fired, blank-node namespace conventions incl. the zero-arg `BNODE()` minting into the reserved `fdb-` namespace, the four parser depth guards, CI topology, policy fail-closed semantics and the real footgun — lenient IRI resolution silently narrowing enforcement scope — the `fluree-db-consensus` role, and the async-mutual-recursion trap). Everything retired stays recoverable via `git log -p`. The standard itself lives in `docs/memory/guides/hygiene-and-auditing.md`: five tests (durable-at-HEAD, team-audience, non-derivable, actionable, well-formed) + the repeatable procedure, written to be reusable verbatim on any repo using `fluree memory`.

## Part 2 — checkpoint audit tooling, so the cleanup becomes routine

The insight from doing Part 1 by hand: the mechanical 80% of an audit (extract, flag narration/portability markers, check refs against HEAD, diff coverage) is exactly what the tool should own, while the judgment (the five tests) stays with the session. So:

- **`memory_audit` MCP tool + `fluree memory audit` CLI** (`fluree-db-memory/src/audit.rs`, shared renderer in `format.rs`): scopes to the current branch's memories by default (branch provenance + refs intersecting the branch diff), flags narration/status language, over-cap content, absolute paths, missing refs, and **churned refs** (file modified since the memory was last verified, via git), plus a coverage section — files changed on the branch that no memory references, framed as "capture only what code/docs can't show." Read-only; degrades gracefully off-git; entries sort flagged-first so the display cap never hides findings.
- **`mem:updatedAt` re-verify stamps**: every `memory update` (including a no-field-change call) stamps `updatedAt` — so "re-verified at HEAD" is one cheap call, recall's recency bonus uses last-verified rather than first-written, and churn signals measure something real. `SCHEMA_VERSION_SALT` bumped so existing ledgers rebuild to pick up the predicate. The reconciled baseline ships pre-stamped, since every memory in it was verified against origin/main today — without that, the first audit run flags ~93% of the store on churn alone.
- **Stale-ref markers in recall**: results whose refs no longer exist at HEAD render with an explicit marker (text/json/context forms) — marked, not down-ranked, so the reading session is prompted to verify-and-update instead of the memory silently rotting.
- **The checkpoint habit, taught by the tool itself**: the MCP server instructions and the shipped rules file now say to run `memory_audit` at PR-ready checkpoints — effort notes become authoritative or get retired *there*, not later. Since this ships in the crate, any repo adopting `fluree memory` gets the same hygiene loop for free.
- **CI guardrail**: `fluree-db-memory/tests/repo_memory_lint.rs` lints the store itself in the ordinary workspace test job — parseability via the crate's own Turtle path, the 750-char cap, lowercase tags, repo-relative refs that must resolve at HEAD, no control bytes or tool-call debris, valid kinds/severity, and now the `updatedAt` shape.

## Notes for review

Gates at HEAD of branch: fmt clean, clippy zero warnings on the touched crates, 96 tests green across `fluree-db-memory`/`fluree-db-mcp` (plus the CLI builds and was smoke-tested against this repo's real store — branch scoping, JSON output, and bad-base degradation all exercised). The `repo.ttl` diff is larger than the content change because the file regenerates in canonical `write_memory_file` order — best read per-block. One operational note: working trees that still have the old store will conflict on next pull; the MCP build-hash detection rebuilds cleanly from the merged file. Happy to walk through any block-level disposition — the retired set is all in git history if anyone wants to spot-check my judgment calls.


fluree-db-memory/tests/repo_memory_lint.rs
```diff
@@ -0,0 +1,279 @@
...
+fn capitalize(s: &str) -> String {
+    let mut c = s.chars();
+    match c.next() {
+        Some(f) => f.to_uppercase().collect::<String>() + c.as_str(),
+        None => String::new(),
+    }
+}
+
+#[test]
+fn repo_memory_refs_are_repo_relative_and_resolve() {
```
bplatz, line 255: **Worth deciding deliberately: this makes the memory store a hard merge gate on the whole file tree.**

It's a plain `#[test]` in `fluree-db-memory/tests/`, so `cargo nextest run --workspace` runs it for everyone, and it asserts every `artifactRef` resolves at HEAD. The store currently references **342 distinct files** — 82 in fluree-db-api, 58 in fluree-db-query, 32 in docs, spread across the workspace.

So:

- Any PR that renames or deletes one of those 342 files fails the workspace suite, on a PR with nothing to do with memory. The failure message says to run `fluree memory update` — editing `.fluree-memory/repo.ttl` becomes a prerequisite for merging an unrelated refactor.
- Merge-order hazard: PR A renames `foo.rs` (green), PR B adds a memory referencing `foo.rs` (green), both merge, main is red. Neither author did anything wrong.
- Given the refactoring velocity in this repo, this will fire.

It also sits at a different severity than the rest of this PR's philosophy. Stale refs in recall are "marked, not down-ranked, so the reading session is prompted to verify-and-update instead of the memory silently rotting," and `audit.rs` treats missing refs as an advisory flag. This is the same signal as a hard failure, in the place with the largest blast radius.

Options, in my order of preference: (a) print instead of assert; (b) keep it failing but only when `.fluree-memory/repo.ttl` is itself in the diff; (c) scope it to refs on memories captured on the current branch.

The other three lints in this file (parse, cap, debris) are genuinely structural and should stay hard failures — this is specifically about the resolve check. All four pass today; the question is what happens to everyone else's PRs.

fluree-db-cli/src/commands/memory.rs
```diff
@@ -412,6 +428,46 @@ async fn run_status(dirs: &FlureeDir) -> CliResult<()> {
...
 
+async fn run_audit(all: bool, base: &str, format: &str, dirs: &FlureeDir) -> CliResult<()> {
+    let store = build_synced_store(dirs).await?;
+
+    // Refs are repo-relative, so they need the project root. A global-mode store
+    // has no `.fluree-memory/` to sit under; fall back to where the user ran us.
+    let repo_root = match store.repo_root() {
+        Some(root) => root.to_path_buf(),
+        None => std::env::current_dir()
+            .map_err(|e| CliError::Input(format!("cannot determine the current directory: {e}")))?,
```
bplatz: In global mode this resolves repo-relative refs against wherever the user happens to be standing, so essentially every ref reports `MISSING` — a confident wrong answer rather than a graceful degradation.

`format.rs::stale_refs` already takes the other position for the same question (`None` root → no stale refs reported, on the grounds that refs can't be resolved). Suggest matching it: skip the ref checks in global mode and say so in the report header.

fluree-db-memory/src/audit.rs
```diff
@@ -0,0 +1,654 @@
...
+    let range = format!("{merge_base}...HEAD");
+    let out = git(repo_root, &["diff", "--name-only", &range])?;
+    Some(BranchDiff {
+        merge_base,
+        files: out.lines().map(str::to_string).collect(),
+    })
+}
+
+/// Commit date of the most recent commit touching each distinct existing ref.
+fn last_commit_dates(repo_root: &Path, memories: &[&Memory]) -> BTreeMap<String, String> {
```
bplatz, line 475: One `git log` subprocess per distinct ref — ~342 forks on this store under `--scope all`. The comment frames it as an optimization over per-(memory, ref), which it is, but it's still O(refs) processes.

A single `git log --format=… --name-only` walk building the map in one pass would collapse it. Not urgent for branch scope (small by construction), but `fluree memory audit --all` is the documented cleanup path, which is exactly where it bites.

fluree-db-memory/src/audit.rs
```diff
@@ -0,0 +1,654 @@
...
+    }
+    Some(String::from_utf8_lossy(&out.stdout).trim().to_string())
+}
+
+/// Files this branch changed relative to its merge base with `base_ref`.
+///
+/// `None` on any git failure — a missing repo or an unknown base must degrade
+/// the audit to "no coverage section", never fail it.
+fn branch_diff(repo_root: &Path, base_ref: &str) -> Option<BranchDiff> {
+    let merge_base = git(repo_root, &["merge-base", base_ref, "HEAD"])?;
```
bplatz, line 457: `base_ref` comes from the user (`--base`) and goes to `git merge-base <base_ref> HEAD` unguarded — a value like `--upload-pack=…` is parsed as an option rather than a ref. It fails closed to `None` today so this is hygiene, not a hole; noting it because the `artifactRef` paths below are correctly passed after `--`, so the convention is already in the file.

Separately: `merge_base...HEAD` is `merge-base(merge_base, HEAD)`, which is just `merge_base` — the three-dot form is redundant here since you already resolved it. Harmless, but `merge_base..HEAD` says what you mean.

fluree-db-memory/src/audit.rs
```diff
@@ -0,0 +1,654 @@
...
+            ("WIP", Regex::new(r"\bWIP\b").unwrap()),
+            ("landed", Regex::new(r"(?i)\blanded\b").unwrap()),
+            ("merged-as", Regex::new(r"(?i)\bmerged as\b").unwrap()),
+            ("awaiting", Regex::new(r"(?i)\bawaiting\b").unwrap()),
+            ("pending", Regex::new(r"(?i)\bpending\b").unwrap()),
+            ("next-step", Regex::new(r"(?i)\bnext steps?\b").unwrap()),
+            ("hand-off", Regex::new(r"(?i)\bhand-?off\b").unwrap()),
+            ("resume", Regex::new(r"(?i)\bresume").unwrap()),
+            ("pr-number", Regex::new(r"(?i)\bPR ?#?\d+|#\d{4,}").unwrap()),
+            ("iso-date", Regex::new(r"\b20\d\d-\d\d-\d\d\b").unwrap()),
```
bplatz, line 359: This will flag durable memories that legitimately cite a date ("the spec froze 2025-03-01", "deprecated in the 2026-01 release"). Advisory-only, so acceptable — just expect noise on a store that records version history. If it proves annoying, requiring adjacency to another narration marker would tighten it a lot.

fluree-db-memory/src/audit.rs
```diff
@@ -0,0 +1,654 @@
...
+
+struct Markers {
+    narration: Vec<(&'static str, Regex)>,
+    portability: Vec<(&'static str, Regex)>,
+    handle: Regex,
+}
+
+/// `@`-prefixed words that are JSON-LD/RDF keywords rather than people. Without
+/// this list every memory about `@context` or `@fulltext` reads as personal.
+const NON_HANDLE_ATS: [&str; 20] = [
```
bplatz: `prefix` is missing from this list, so a memory mentioning `@prefix` reads as a personal handle — and this PR adds `docs/memory/reference/ttl-format.md`, which makes such memories likely. Worth adding alongside `base`.

bplatz (approved):
Approving. The two-part framing is right, and Part 1's method — re-verify claims against HEAD, retire what's wrong, keep it recoverable in git — is the correct instinct. "A confident memory asserting something the code no longer does is worse than no memory" is the whole argument in one line.

`audit.rs` is clean: read-only, degrades to no-coverage-section on any git failure, one `git log` per distinct ref rather than per pair, flagged-first sort so the display cap can't hide findings. The `NON_HANDLE_ATS` list for `@context`-style false positives is a nice touch. The `updated_at` plumbing is complete — the flat-row index and the named-binding path in `get()` don't conflict, and the `.ttl` round-trip test is the right guard.

Flagged a handful of things inline; the first one (the refs lint as a hard workspace-wide merge gate) is a design call rather than a defect, so have a look and decide the path you want. Nothing there blocks the merge.

aaj3f:
Thank you for the careful pass, @bplatz — every one of these was right, and they're all addressed on the branch now (`086d36027`, plus a merge of main in `cbb72097d` since GH Actions was having a day yesterday and we were 178 commits back anyway). Taking them in your order:

**The refs lint as a merge gate** — you're right, and I went with your option (a). The resolve-at-HEAD check now *prints* an advisory (pointing at `fluree memory audit`) and never fails; the structural checks (parse, cap, debris, id/scope/severity shape, repo-relative-ness — none of which can rot via an unrelated rename) stay hard. Your framing convinced me it wasn't just a blast-radius problem: it genuinely sat at a different severity than the PR's own "marked, not down-ranked" position, and the two-green-PRs-red-main hazard would absolutely have fired at this repo's rename velocity. The reasoning is now a comment in the test so the next person doesn't "fix" it back into an assert.

**Global-mode refs** — fixed the way you suggested: `audit_memories` takes `Option<&Path>` now, and a store with no project root skips the ref/churn checks and coverage and says so in the report header, matching the `stale_refs` position in `format.rs`. There's a regression test pinning skip-not-MISSING (`no_root_skips_ref_checks_instead_of_reporting_missing`).

**O(refs) forks** — collapsed to a single `git log --since=<oldest effective timestamp> --format=%x01%cI --name-only` walk, newest-first, first-seen-path wins. The `--since` bound is the nice part: because the baseline ships re-verified, the walk covers days of history rather than all of it. Smoke on the real store: one subprocess where `--scope all` previously forked ~342.

**`base_ref` hygiene** — both taken: a leading-`-` base ref is rejected before it reaches `merge-base`, and the range is now `merge_base..HEAD` since the merge base is already resolved.

**iso-date noise** — I took your tightening now rather than waiting for it to annoy someone: a bare date (spec-froze / deprecated-in-release provenance) no longer flags; it only flags adjacent to another narration marker. That matches how the manual audit actually treated dates — I'd kept several date-carrying external-fact memories deliberately. Full-store flag count dropped from the churn-storm to 22/214.

**`prefix`** — added, with a small blush given this PR is the one that shipped `ttl-format.md`.

One bit of full-circle evidence from the merge itself, fwiw: pulling main onto the branch brought `rust-toolchain.toml` and workspace `clippy -D warnings` with it — which made the CI-topology memory this PR *adds* wrong in two clauses. The checkpoint-audit habit caught it, the memory got updated and re-stamped, and the cap lint then rejected my first rewrite at 757 chars. The guardrails eating their own dogfood before the PR even merges is about the best validation of them I could have asked for.

CI is re-running on the merged head; I'll let it settle before merging.

PR #1580, opened 2026-07-31
fix(policy): key the class cache on the graph, not just the subject

christophediprima:
`PolicyContext::class_cache` is a `HashMap<Sid, Vec<Sid>>`. The same subject IRI can carry different `rdf:type` values in different named graphs, so two graphs sharing a subject collapse into one cache entry and whichever populated it first decides the classes for both.

Every `f:onClass` restriction evaluated for that subject afterwards uses the wrong class set — allowing or denying based on population order rather than on the data. That is order-dependent behaviour in an authorization path, not a stale-cache annoyance.

This is not a hypothetical shape. Named graphs are a general feature — any `from-named` query spans them — and `docs/security/policy-in-queries.md:203-205` explicitly recommends `f:onClass` as the tool to reach for when "different graphs need different policy regimes". That is precisely the restriction this bug corrupts, so the documented workaround for graph-differentiated policy is the one that misbehaves on multi-graph data.

It has gone unnoticed because it needs the *same subject IRI* present in two graphs with different types, which is rarer than multi-graph data generally. I came across it while auditing how far the graph identifier travels through the read path; R2RML `rr:graphMap` routing is the workload that produces the shape per row #1581, which is why I went looking. Nothing about the bug or the fix is specific to R2RML.

**What changes**

The graph was already available at both ends and simply wasn't carried across:

- `populate_class_cache` (`fluree-db-policy/src/class_lookup.rs`) receives a `GraphDbRef`, whose `g_id` is a public field, so the insert side now keys on the graph it actually read.
- The two lookups live in `filter_flakes_for_graph` / `allow_flake_for_graph` (`fluree-db-query/src/policy/enforcer.rs`) — the one boundary where the graph was being dropped. Both now take a `g_id: GraphId`.
- `class_cache` becomes `HashMap<(GraphId, Sid), Vec<Sid>>`; `cache_subject_classes` and `get_cached_subject_classes` take the graph.

**One design note worth a reviewer's eye.** Callers pass the graph of the `GraphDbRef` they populated the cache from, rather than a literal. Populate and lookup therefore agree by construction rather than by convention. Two callers — `block_fetch.rs` and `graph_commit_builder.rs` — build that ref with a hardcoded graph `0`, so their behaviour is byte-identical after this change. That hardcoding is a separate defect and is deliberately left alone here, but because the lookup now follows the ref, fixing it later corrects both halves in one edit instead of two.

**Tests** — two units in `evaluate.rs`: the same subject cached with different classes in graphs 3 and 4 keeps both entries; a graph with no entry misses rather than borrowing another graph's classes (a miss degrades to "no classes", which is the conservative direction).

**Be aware of what those tests do not do.** They cannot be run against the unfixed code, because the accessors did not take a graph at all — so they pin the new keying without demonstrating the old wrong decision. Showing the authorization outcome end to end needs an enforcer-level integration test over a real two-graph ledger. That belongs with the `rr:graphMap` work, where the data shape arises naturally, and I would rather say so than let the test count imply more coverage than it has.

**Why it is worth having on its own:** it is a correctness bug in an authorization path, the fix is contained, and it stands independent of anything else I have open — it needs no other change to be worth landing, and nothing else needs to land first.

### DIFF
--- a/fluree-db-policy/src/evaluate.rs
+++ b/fluree-db-policy/src/evaluate.rs
@@ -14,7 +14,7 @@ use crate::types::{
     TargetMode, WriteFlakeInfo,
 };
 use crate::Result;
-use fluree_db_core::{FlakeValue, Sid, Tracker};
+use fluree_db_core::{FlakeValue, GraphId, Sid, Tracker};
 use std::collections::HashSet;
 use std::sync::{Arc, RwLock};
 use uuid::Uuid;
@@ -66,8 +66,13 @@ pub struct PolicyContext {
     pub wrapper: PolicyWrapper,
     /// The grounded identity (always has a value, even if random)
     pub identity: Sid,
-    /// Cache of subject -> classes for runtime class membership checks
-    class_cache: Arc<RwLock<std::collections::HashMap<Sid, Vec<Sid>>>>,
+    /// Cache of (graph, subject) -> classes for runtime class membership checks.
+    ///
+    /// Keyed on the graph as well as the subject: the same subject IRI can carry
+    /// different `rdf:type` values in different named graphs, and an `f:onClass`
+    /// decision made against another graph's classes is simply wrong. Keying on
+    /// `Sid` alone made the result depend on which graph populated the entry first.
+    class_cache: Arc<RwLock<std::collections::HashMap<(GraphId, Sid), Vec<Sid>>>>,
 }
 
 impl PolicyContext {
@@ -908,19 +913,19 @@ impl PolicyContext {
         Ok((false, evaluated))
     }
 
-    /// Cache subject classes for repeated lookups
-    pub fn cache_subject_classes(&self, subject: Sid, classes: Vec<Sid>) {
+    /// Cache subject classes for repeated lookups, within one graph.
+    pub fn cache_subject_classes(&self, g_id: GraphId, subject: Sid, classes: Vec<Sid>) {
         if let Ok(mut cache) = self.class_cache.write() {
-            cache.insert(subject, classes);
+            cache.insert((g_id, subject), classes);
         }
     }
 
-    /// Get cached subject classes
-    pub fn get_cached_subject_classes(&self, subject: &Sid) -> Option<Vec<Sid>> {
+    /// Get cached subject classes for a subject in a specific graph.
+    pub fn get_cached_subject_classes(&self, g_id: GraphId, subject: &Sid) -> Option<Vec<Sid>> {
         self.class_cache
             .read()
             .ok()
-            .and_then(|cache| cache.get(subject).cloned())
+            .and_then(|cache| cache.get(&(g_id, subject.clone())).cloned())
     }
 }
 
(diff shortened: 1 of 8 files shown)


fluree-db-query/src/policy/enforcer.rs
```diff
@@ -127,7 +130,7 @@ impl QueryPolicyEnforcer {
             // Get subject classes from cache
```
aaj3f, line 130: **`fluree-db-query/src/policy/enforcer.rs:130`** — *optional, performance.* `get_cached_subject_classes(g_id, &flake.s)` now clones the `Sid` per flake to build the owned tuple key (`cache.get(&(g_id, subject.clone()))`). Since `Sid` is `u16` + `Arc<str>` this is a refcount bump, not an allocation, and the pre-existing per-flake costs on this path (the `RwLock` read, the `Vec<Sid>` clone per hit) dominate it — so I'm not blocking on it. If policy-filtered scans ever show this in a profile, a nested `HashMap<GraphId, HashMap<Sid, Vec<Sid>>>` restores borrowed lookups and gives per-graph invalidation for free; we may want that shape anyway if graph-scoped cache clearing becomes a thing.

fluree-db-query/src/policy/enforcer.rs
```diff
@@ -163,6 +166,7 @@ impl QueryPolicyEnforcer {
     pub async fn allow_flake_for_graph(
```
aaj3f, line 166: **`fluree-db-query/src/policy/enforcer.rs:166`** — *observation, no action.* `allow_flake_for_graph` has no callers anywhere in this repo, and none in fluree/solo either (checked against solo's `src/rust` for the pin-bump question). The signature change is therefore free — noting it so nobody burns time hunting for a missed call site.

fluree-db-policy/src/evaluate.rs
```diff
@@ -1190,6 +1195,46 @@ mod tests {
         );
     }
 
+    #[test]
```
aaj3f, line 1198: **`fluree-db-policy/src/evaluate.rs:1198-1236`** — *praise.* The tests pin the two properties that define the fix (per-graph entries, miss-not-borrow) and the comments explain the production shape each guards against. Declining to fake an end-to-end test the current shapes can't exercise — and saying where it should land instead — is the right call.

aaj3f (approved):
Thanks for this @christophediprima -- order-dependent authorization is the worst kind of latent bug — correct in every test that populates one graph, wrong the day two graphs share a subject — and this fixes it at the right altitude: the cache key names the graph, the two accessors take the graph, and the five call sites pass the graph of the ref they already hold, so populate and lookup agree by construction rather than convention. The sequencing logic is also right: `rr:graphMap` (#1581) is exactly the feature that makes this shape common, and landing the policy fix first means the graph-routing work arrives onto a correct cache. I verified rather than trusted: cold package builds in an isolated worktree — `fluree-db-policy` 39/39 with both new tests present by name, `fluree-db-query --lib` 1304/0, clippy clean on both crates — and confirmed the signature changes strand no caller (zero in-repo callers of `allow_flake_for_graph`; zero uses of any changed symbol in fluree/solo's Rust, so the next pin bump is unaffected).

- **Correctness:** ✔ The keying fix is complete at the boundary it names; the hardcoded graph-0 refs in `block_fetch`/`graph_commit_builder` are honestly scoped out with behavior unchanged and automatic flow-through once fixed.
- **Performance (speed first, memory second):** ✔ Effectively neutral — the new per-flake `Sid` clone is an `Arc` refcount bump dominated by the pre-existing `RwLock` read and `Vec` clone on the same path; no allocation, lock, or collect added to any loop. Optional nested-map note above if it ever profiles.
- **Shared abstractions:** ✔ Extends the existing `filter_flakes_for_graph` per-graph parameter shape (snapshot/overlay/t were already per-graph; `g_id` joins them) rather than inventing a parallel mechanism.
- **Testing:** ✔ Both new tests run and pin the fix's contract; the deferred enforcer-level integration test is named with its natural home (#1581) rather than skipped silently.
- **Conventions:** ✔ Conventional multi-line commit with real root-cause narration; doc comments record the *why* on the key type; no AI-attribution trailers; clippy/fmt clean.

Verified locally at branch HEAD `18ad842c4`: see the summary's verification block.

Recommend merge, ideally before #1581 so the graph-routing work lands on the corrected cache.

PR #1581, opened 2026-07-31
feat(r2rml): refuse the graph-map constructs the materializer cannot honor

christophediprima:
**Rescoped — and I should have caught this sooner.** This PR was opened on 2026-07-31 to add `rr:graphMap` routing. That routing had already landed in `main` nine days earlier, in `2daa9a1e2`, which is my own commit merged as part of #1422. The PR still showed +494/−6 only because its merge-base predates that landing, so the diff replayed work that was already upstream. Apologies for the noise — if you looked at this in the last two weeks, that is why it did not match the code.

What is genuinely absent from `main` is the part that decides what happens when a mapping asks for something the materializer cannot do. This PR is now only that.

## The problem

`extract_graph_map` (`loader/extractor.rs`) returns `Option<GraphMap>` and yields `None` for anything it cannot parse. So a mapping that asks for named-graph routing and gets none is indistinguishable, downstream, from one that never asked.

That distinction has data consequences rather than merely cosmetic ones. The graph is resolved once per row and the accumulator keys on it, so a dropped graph map leaves every row in the default graph — and two partitions holding the same subject IRI then collapse onto one key and overwrite each other per predicate. The failure is silent, it lands in the data rather than the logs, and the mapping author gets no signal at any point.

## The change

`extract_graph_map` returns `R2rmlResult<Option<GraphMap>>`, and five shapes outside the implemented subset are refused by name:

| construct | why it cannot be honored |
|---|---|
| more than one `rr:graph` / `rr:graphMap` on a term map | R2RML treats these as **cumulative** — the triple goes into every named graph. One graph is implemented, so all but the first would be discarded. |
| `rr:graph rr:defaultGraph` | Parsed as an ordinary constant it mints a named graph literally called `http://www.w3.org/ns/r2rml#defaultGraph` — the opposite of what was asked for. |
| `rr:graphMap [ rr:constant rr:defaultGraph ]` | The long form of the same mistake. |
| `rr:graphMap` with no `rr:template`, `rr:column` or `rr:constant` | It names no graph, so it is malformed rather than absent. Treating it as absent is exactly how routing degrades to the default graph unseen. |
| `rr:graph` / `rr:graphMap` on a **predicate-object map** | Valid R2RML that scopes just that map's triples. `PredicateObjectMap` has no field for it and the graph is resolved from the subject map, so it would be read and then ignored. |

`R2RML::DEFAULT_GRAPH` is added to the vocabulary for the sole purpose of being recognised and rejected.

`Ok(None)` still means "no graph map", which remains the common case, and default-graph behavior is unchanged. Nothing that parses today stops parsing — every one of these five previously produced `None`.

## On the POM case

That one is a real R2RML feature rather than a malformed mapping, so refusing it is a scoping decision rather than a correctness fix. I refused rather than implemented it because the materializer resolves the graph per row from the subject map, and threading a per-POM graph through that is a larger change than this PR should carry. The error names the construct and says to move the graph map to the subject map, so implementing it later can relax the check without changing what the message means. Happy to take the other route if you would rather see it supported now.

## Tests

Six. **Five assert the error message, not `is_err()`** — deliberately, because these guards are adjacent and an untargeted assertion lets one refusal stand in for another. Removing the `selectOne`-style guard in a neighbouring PR still produced *a* decline, and a bare `is_err()` stayed green; that is the failure mode this avoids.

The sixth, `a_single_subject_graph_map_is_still_accepted`, pins that the refusals do not over-fire on the supported shape.

Each of the five guards was mutation-tested by deleting it: each deletion is caught by exactly one test, and the accepted-shape test plus both pre-existing graph-map tests stay green under all five.

`fmt` and `clippy -p fluree-db-r2rml --all-features --all-targets -- -D warnings` clean; `fluree-db-r2rml` 147 pass.


fluree-db-r2rml/src/loader/extractor.rs
```diff
@@ -213,26 +213,73 @@ impl<'a> MappingExtractor<'a> {
...
+                graph_shortcuts.len(),
+                graph_maps.len()
+            )));
+        }
+
         // rr:graph <iri> — constant shortcut.
-        if let Some(graph_obj) = self.find_object_optional(triples, R2RML::GRAPH) {
-            if let Some(iri) = self.term_to_iri(&graph_obj) {
-                return Some(GraphMap::constant(iri));
+        if let Some(graph_obj) = graph_shortcuts.first() {
```
aaj3f, line 281: **`fluree-db-r2rml/src/loader/extractor.rs:254`** — *blocking.* **`rr:graph` with a non-IRI object silently degrades to no graph map — the exact silent-drop this PR exists to close, inside the function that closes it.** The shortcut branch is two nested `if let`s with no `else` on either: when `term_to_iri` returns `None` (a literal object, a blank node), control falls out of both, `graph_maps` is empty, and `:270-272` returns `Ok(None)`. Concretely, `rr:graph "http://example.org/g1"` and `rr:graph [ ]` both compile today with `graph_map = None` (probe-confirmed at this head) — the author wrote a graph map, and their triples silently land in the default graph, which is precisely the failure mode your five refusals were built to make impossible. The fix is the idiom you already use for the same shape two branches up: mirror the `rr:subject` `InvalidValue` treatment at `:162-167` with an error naming the offending term.

fluree-db-r2rml/src/loader/extractor.rs
```diff
@@ -213,26 +213,73 @@ impl<'a> MappingExtractor<'a> {
...
+        if let Some(constant_obj) = self.find_object_optional(&gm_triples, R2RML::CONSTANT) {
+            if self.term_to_iri(&constant_obj).as_deref() == Some(R2RML::DEFAULT_GRAPH) {
+                return Err(R2rmlError::Unsupported(
+                    "rr:graphMap [ rr:constant rr:defaultGraph ] is not supported. Omit the \
+                     graph map entirely to target the default graph."
+                        .to_string(),
+                ));
+            }
+        }
         let mut graph_map = GraphMap::default();
```
aaj3f: **`fluree-db-r2rml/src/loader/extractor.rs:283`** — *optional, sharper than it used to be.* `rr:graphMap [ rr:template "…{a}" ; rr:column "c" ; rr:constant ex:gc ]` compiles with all three value sources populated — `:283-300` sets them independently and only `is_empty()` is checked at `:305` — and now that the materializer consumes graph maps, `materialize_graph_from_batch` picks `constant` first (`term.rs:835-842`) and the other two are silently discarded, so the wrong graph actually gets minted rather than hypothetically. An exclusivity check here is a few lines and matches the spec's exactly-one rule. Minor and non-blocking — but if you agree it's right, I'd rather see it folded in now than lost in the backlog.

fluree-db-r2rml/src/loader/extractor.rs
```diff
@@ -213,26 +213,73 @@ impl<'a> MappingExtractor<'a> {
...
+            )));
+        }
+
         // rr:graph <iri> — constant shortcut.
-        if let Some(graph_obj) = self.find_object_optional(triples, R2RML::GRAPH) {
-            if let Some(iri) = self.term_to_iri(&graph_obj) {
-                return Some(GraphMap::constant(iri));
+        if let Some(graph_obj) = graph_shortcuts.first() {
+            if let Some(iri) = self.term_to_iri(graph_obj) {
+                if iri == R2RML::DEFAULT_GRAPH {
```
aaj3f: **`fluree-db-r2rml/src/loader/extractor.rs:256`** — *optional, a design question rather than a defect now.* This is more of a question than a suggestion: since a constant graph map of `rr:defaultGraph` *is* the default graph per spec, would recognizing it and returning `Ok(None)` be strictly better than refusing? The refusal now guards against a real wrong behavior (the `…#defaultGraph` minting your commit describes — I verified that claim is true at this base), so this is no longer blocking anything — but `Ok(None)` would make the one spec-blessed spelling work instead of erroring, and the error text could then be reserved for the shapes that genuinely can't be honored.

fluree-db-r2rml/src/loader/extractor.rs
```diff
@@ -252,12 +299,17 @@ impl<'a> MappingExtractor<'a> {
             }
         }
 
-        // A graphMap that parsed no usable value source is treated as absent.
+        // A graphMap that parsed no usable value source is a malformed mapping,
+        // not an absent one. Treating it as absent is exactly how routing
+        // silently degrades to the default graph.
         if graph_map.is_empty() {
```
aaj3f: **`fluree-db-r2rml/src/loader/extractor.rs:305`** — *low.* `rr:graphMap [ rr:constant "http://example.org/g" ]` (literal, not IRI) is rightly rejected but with "has no `rr:template`, `rr:column` or `rr:constant`" — it has one; it's just not an IRI (`term_to_iri` dropped it at `:296-300`). A "constant must be an IRI" arm would tell the author what to actually fix.

fluree-db-r2rml/src/loader/extractor.rs
```diff
@@ -213,26 +213,73 @@ impl<'a> MappingExtractor<'a> {
         }
 
         // Extract rr:graph / rr:graphMap (subject-map-level named-graph routing).
-        subject_map.graph_map = self.extract_graph_map(&sm_triples);
+        subject_map.graph_map = self.extract_graph_map(&sm_triples)?;
```
aaj3f, line 240: **`fluree-db-r2rml/src/loader/extractor.rs:161`** — *low.* The `rr:subject` shortcut returns at `:161-168`, before `extract_graph_map` at `:216` — so `rr:subject ex:s1` beside `rr:subjectMap [ … ; rr:graph ex:g1 ]` compiles with the graph map silently dropped (probe-confirmed). Same silent-drop family as the blocker, lower stakes.

(Commenting at the `extract_graph_map` call site because `:161` isn't in this diff.)

fluree-db-r2rml/src/loader/extractor.rs
```diff
@@ -279,6 +331,25 @@ impl<'a> MappingExtractor<'a> {
...
+        // that map's triples. `PredicateObjectMap` has no field for one and the
+        // materializer resolves the graph once per row from the SUBJECT map, so a
+        // POM-level graph map would be read and then ignored — the silent
+        // cross-partition collapse described on `extract_graph_map`. Refuse it by
+        // name; implementing it later can relax this without changing what the
+        // error means today.
+        if !self.find_objects(&pom_triples, R2RML::GRAPH).is_empty()
+            || !self.find_objects(&pom_triples, R2RML::GRAPH_MAP).is_empty()
+        {
+            return Err(R2rmlError::Unsupported(
```
aaj3f, line 395: **`fluree-db-r2rml/src/loader/extractor.rs:344`** — *praise, and a note on the rescope.* The POM refusal's "Move the graph map to the subject map" now routes users to a path that genuinely works — your `2daa9a1e2` on main is what made that true, and rescoping this PR down to the refusal layer once the consumer landed was exactly the right call. The five message-asserting tests plus the accepted-shape pin are real guards: I re-ran two of the claimed mutations and each killed exactly its one named test.

fluree-db-r2rml/src/loader/extractor.rs
```diff
@@ -279,6 +331,25 @@ impl<'a> MappingExtractor<'a> {
...
+                "rr:graph / rr:graphMap on a predicate-object map is not supported. Only \
+                 subject-map-level graph maps are honored; a POM-level graph map would be \
+                 ignored and its triples would land in the subject's graph. Move the graph \
+                 map to the subject map."
+                    .to_string(),
```
aaj3f, line 400: **`docs/graph-sources/r2rml.md:351` (not in this diff — noting from the nearest anchor).** — *optional.* Limitations still lists four items and says nothing about graph maps, while this PR makes five constructs hard-fail mapping load. One paragraph there turns a support ticket into a doc read.

(Commenting here because `docs/graph-sources/r2rml.md` is not in this diff.)

aaj3f (changes requested):
The rescope of this PR improved it significantly, and I'm grateful for being so thoughtful about the rewrite, @christophediprima. I re-ran two of your guard mutations and each killed exactly its one test, as claimed. Thank you for continuing to push on this: the graph-map story is materially more complete than it was a month ago.

The one thing I'd ask before approving: `rr:graph` with a literal or blank-node object still compiles silently with no graph map (`extractor.rs:254-267`, no `else` on either `if let`). I probe-confirmed both shapes at this head. That's the same silent degradation your refusals exist to prevent, in the same function, so I don't think this PR can land while it's open. The fix is the `InvalidValue` idiom you already use for `rr:subject` at `:162-167`. 

The rest is take-or-leave: the exclusivity check and the defaultGraph design question inline, two low-stakes ergonomics items, the docs paragraph, and one inaccurate compatibility sentence in the body.

Related, connecting rather than asking: the virtual-path half of this story is already tracked in #1607 (@bplatz filed it off #1422's review, and it names this PR as the vocabulary half). My read at this merge-base confirms its premise: zero `graph_map` references under `fluree-db-query/`, so a mapping with an accepted graph map queried virtually returns everything in the default graph while its materialized twin scopes per graph. 

Nothing to do in this PR, other than one coordination note worth having in writing: #1607's scope says the full placement rules should coordinate with this vocabulary work, which suggests the POM-level refusal you're adding here is the first one that gets relaxed when it lands.

One body nit to fold into the same editing pass: "every one of these five previously produced `None`" is false for three of the five — multiple `rr:graph` previously produced `Some(first)` with the rest discarded, and both `rr:defaultGraph` forms produced the minted constant, which your own table says (that's the defect). The refusals are still right; the closing compatibility line just overstates.

Verified at branch HEAD (`4f4973502`), agent-run and lead-re-verified: full `-p fluree-db-r2rml --all-features` suite 147/0; two guard mutations re-run red-then-restored; five still-live probes reproduced (both blocker shapes, three-value-source, literal-constant message, subject-shortcut drop); merge-base `a85e03682` consumer path confirmed at `r2rml_materialize.rs:2074`.

Happy to re-review as soon as the literal/blank-node arm is in — it's a small change and everything else here is ready.

bplatz:
Finishing this up on the branch, on top of a merge with current `main`. Three commits:

- `fix(r2rml): refuse malformed graph maps and honor rr:defaultGraph`
- `fix(r2rml): count the subject column and graph-map columns among a map's referenced columns`
- `fix(materialize): refuse a graph-mapped source in the twin builder`

From the review:

- **Blocker** — `rr:graph` with a literal or blank-node object is now `InvalidValue` naming the term, no longer a silent "no graph map".
- **Exclusivity** — a `rr:graphMap` with more than one of `rr:template` / `rr:column` / `rr:constant` is refused; so is one with none, and an `rr:termType` other than `rr:IRI`.
- **Non-IRI constant** — `rr:graphMap [ rr:constant "…" ]` now says the constant must be an IRI rather than "no value source".
- **`rr:subject` beside `rr:subjectMap`** — refused (exactly one subject map), so a graph map on the subject map can no longer be dropped by the shortcut.
- **`rr:defaultGraph`** — took the design question the other way: it is honored as the default graph, in the extractor for constants and per row in the materializer.
- **Docs** — Limitations paragraph in `docs/graph-sources/r2rml.md` on the supported subset and what is refused.

Found on the way and included here: `referenced_columns()` now covers graph-map columns and an `rr:column` subject, so `validate_r2rml` flags a graph-template column the table lacks; the twin builder (`fluree materialize`) refuses a graph-mapped mapping instead of flattening every triple into the default graph, which its parity gate could not detect; and `mark_sql_source` is gated on the `sql` feature so `clippy --features iceberg` is clean.

Every guard was deleted in turn to confirm its own tests catch it. Kept out and filed separately: #1798 (constant literal objects lose their datatype) and #1799 (term maps accept several value sources). Query-path parity stays with #1607.

PR #1582, opened 2026-07-31
perf(r2rml): vectorized fused-aggregate fold + late-rendered decode seam (Gap-4 P1)

aaj3f:
## What & why

Gap-4 program, **P1 — the fused-tail fix** (the first gated PR family from the adjudicated slate). The external-engine A/B measured the *fused* R2RML/Iceberg aggregate family 6–52× behind DuckDB at scale, and a co-resident **CPU profile spike (P0)** settled *where* that time goes. On the genuinely-fused pairs (cq014, cq027) RDF **term-minting is 0.0%** of on-CPU wall — the fused fold never mints — refuting "late materialization is the cure." The tail is two costs, co-primary:

| pair (fused) | term-mint | fold mechanics | decode-copy | arrow/IO |
|---|---|---|---|---|
| cq014 | **0.0%** | 39.6% | 32.9% | 13.7% |
| cq027 (1M-row) | **0.0%** | 45.9% | **46.2%** | 4.5% |

(For contrast, the shapes that *decline* the fuse — cq038, crt_highcard — are 58–83% minting; that is an admission/coverage problem, out of scope here.)

This PR attacks exactly those two costs, corridor-only (r2rml/iceberg), each behind a default-on kill switch whose OFF path is byte-identical to today.

## Mechanisms

**N1 — vectorized fused-aggregate fold** (`fluree-db-query/src/r2rml/fused_aggregate.rs`). The GROUP BY fold allocated a fresh `Vec<GKey>` **and cloned each string key** every row, then hashed a `Vec<GKey>` through SipHash into a `HashMap<Vec<GKey>, Vec<Acc>>` — on cq027 that is 2M `String` clones + 1M SipHashes to map 1M rows onto **18** groups (`GroupCol::key_at` 14.4% + SipHash 7.5%, P0). New path: build the composite key **borrowed** (`GKeyRef`, no per-row `String` clone / `Vec` alloc) into a reused scratch, hash with `FxHasher`, and intern it to a dense id in a `hashbrown::HashTable<(Vec<GKey>, u32)>`; the owned key is materialized **only when a new group is inserted**. Typed accumulators live in `group_accs[id]`.
`HashTable` (not `HashMap` + `insert_hashed_nocheck`) is deliberate: the borrowed probe and the resize-rehash go through the SAME hash function (`hash_key_refs` / `gkeys_hash`, kept in lockstep), so a table grow can never re-bucket a key away from its probe. (An earlier `HashMap`+`insert_hashed_nocheck` draft did split groups on resize — the live A/B differential caught it before commit; the grouping test is now hardened to force grows.)
Preserved exactly: the exact-arithmetic domain (`Dec{i128,scale}` → `BigDecimal` on overflow, unchanged `accumulate_*`/`Acc`), every fold semantic, the memory-budget group-growth checkpoint, and the whole decline/fallback surface (admission untouched). f64 `SUM` stays bit-identical because each group's rows fold in the same row order under both paths.

**N2 — direct Arrow→Column decode** (`fluree-db-iceberg/src/io/{arrow_reader,parquet}.rs`). Decode was a **two-hop copy**: Arrow array → `Vec<Option<ColumnValue>>` (`arrow_column_to_values`) → `Column`'s `Vec<Option<T>>` (`build_columns_from_values`) — a full intermediate allocation per column, and **two `String` allocations per string cell**. New `arrow_column_to_column` builds the `Column` directly from the Arrow array (one hop; one string alloc). Its fast arms cover only the case where the Arrow physical type matches the field type — each provably value-identical to the old composition — and every other (cross-type coercion, timestamp unit-scaling, unhandled type, schema-evolution-absent column) falls back to the exact old mapping via the extracted `column_from_values`. So behavior is unchanged everywhere; only the hot common path is shortened.

## Switch inventory (both default ON; OFF = byte-identical old path)

- `FLUREE_FUSED_VECTOR_FOLD` — N1. Off restores the owned-key `HashMap<Vec<GKey>, Vec<Acc>>` fold.
- `FLUREE_ARROW_DIRECT_DECODE` — N2. Off restores the two-hop `arrow_column_to_values` + `build_columns_from_values` path.

## Gates

- **Value-identical differentials (hermetic):**
  - `direct_decode_matches_two_hop` — asserts `arrow_column_to_column` == the two-hop path (Debug-equal) across every fast type + nulls + NaN + cross-type fallbacks.
  - `key_ref_at_matches_key_at` — the borrowed key read equals the owned read for every Column kind, null, out-of-bounds, wrong-type, and the NUMBER(n,0) physical-Decimal integer coercion (`None` ⟺ `GKey::Null`).
  - `vector_fold_grouping_matches_owned_map` — the dense dict partitions + counts a key sequence identically to the owned-key map.
- **fmt** diff-clean (`fluree-db-query`, `fluree-db-iceberg`). **clippy** `--all-targets --no-deps` diff-clean on both crates (no new warnings introduced by this change).
- **Crate tests:** `fluree-db-query`, `fluree-db-iceberg --all-features`, `fluree-db-api` (featureless + iceberg) — all green.

## Live yardstick (Substrate B, SF0.1, warm; same binary, old switches OFF vs new ON)

Median of 7 warm reps, fresh process/rep, local MinIO+REST fixture (macOS, co-resident host — absolute ms are noisier than the EC2 Wave-C rows and not host-comparable to them; the OFF→ON ratio on one binary is the acceptance evidence). DuckDB column is the EC2 Wave-C SF0.1 number for context only.

| pair | old (OFF) | new (ON) | speedup | result | DuckDB (EC2 WaveC) |
|---|---|---|---|---|---|
| cq014 (fused single-table COUNT+SUM) | 26.0 ms | 14.1 ms | **1.84×** | IDENTICAL | 56 ms |
| cq027 (fused 1M-row grouped COUNT) | 199.1 ms | 117.9 ms | **1.69×** | IDENTICAL | 77 ms |
| cq008 (fused fact→dim→dim rollup) | 203.4 ms | 208.5 ms | 0.98× | IDENTICAL | 128 ms |
| crt_minmax (MIN/MAX — separate fast path) | 126.1 ms | 126.5 ms | 1.00× | IDENTICAL | 59 ms |
| crt_highcard (declines → generic path) | 1440 ms | 1470 ms | 0.98× | IDENTICAL | 238 ms |

**Results are byte-identical (sorted) between OFF and ON on every pair — kill-criterion-1 satisfied.** The movers are the single-table fused GROUP BYs the P0 profile fingered (cq014, cq027). cq008 is join-resolution-dominated (its fold/decode are a small share, so ~flat); crt_minmax uses a different fast path; crt_highcard declines to the generic path (no movement by design — see below). An earlier draft that split groups on table resize was caught by this very differential and fixed before commit.

## Why crt_highcard does NOT move here (expected, by design)

`crt_highcard` (`GROUP BY ?c` where `?c = edw:customer`) **declines the fuse** — its group key is a `RefObjectMap`/IRI, and `scalar_column_for_var` admits only `ObjectMap::Column` — so it runs the **generic** per-row `materialize_batch` path. The P0 profile confirms it is a *minting* shape (term-mint 58.2%, via `expand_template_from_batch` + regex), not a fold/decode shape, so P1 correctly shows ~no movement on it (its small decode-seam share, 3.6%, is all N2 can touch). It becomes a mover only after **admission widening (P2 / N3)**: admit a `RefObjectMap` group key by folding on the FK **column value** and **minting the group-key IRI once per group at output** — i.e. this PR's dict + late-materialize-only-on-emit substrate extended to IRI keys. That is the designed follow-up; it is deliberately **not** built here.

## Scope / non-goals

Corridor-only: the engine-wide `hash_join.rs` / `group_aggregate.rs` (native, Binding-based) are untouched, so native legs are byte-identical for the mechanical reason they never enter this path. Minting on the *un-fused* shapes (cq038/crt_highcard) is the next program step (admission widening, N3), not this PR. MIN/MAX is a separate fast path, not this fold.

## 2026-08-03 addendum — scope expansion: post-merge #1529 review fixes + main merge-up

This branch merged current `main` (`96370dd10`, post-#1554/#1529 — conflict-free; the P1 value-identity differentials re-proven at the merged head since family-C's widened admission now feeds the vectorized fold). It then became the fix vehicle for a high-rigor post-merge review of #1529 (the findings arrived after that PR merged; per the maintainers' consolidation preference they land here rather than in a new PR). Every finding was independently re-verified against the merged tree before fixing — all eight substantive findings CONFIRMED:

| Finding | Fix |
|---|---|
| CRITICAL-1 — gate-failure drop was ledger-wide while the freshness guard was branch-scoped (`--into existing:branch` + failed gate purged every branch) | up-front refusal when the twin's ledger name has any live non-target branch (`6a5403ff6`) |
| CRITICAL-2 — worker-pool death hung the build forever (driver-held receiver made the disconnect path unreachable; mutex poison cascaded) | receiver dropped post-spawn, poison-tolerant locks, unconditional close+join on error paths; errors-not-hangs test under a tripwire (`fade7a0ae`) |
| CRITICAL-3 (minimal) — `--verify full` false-rejected every twin with an `xsd:float` column (export writer hardcoded `xsd:double`) | float→double value-canonicalization fold + export arm resolves the declared type + multiset caveat in count-check failures (`c83c67bc6`); durable single-renderer restructure tracked |
| MAJOR-1 — composite-FK join keys transposed on the live virtual path (parent-sorted index vs child-declared probe; twin was correct → real 1:1 divergence) | `canonical_join` shared from fluree-db-r2rml into both operator sites + a disagreeing-name-order composite test (`757a54868`) |
| MAJOR-2 — Direct-path watermark effectively unpinned (2s metadata TTL) | loud-fail: any second distinct `metadata_location` observed for a table fails the build; cache-off refuses (`2d1a1d407`) |
| MAJOR-3 — warehouse-root auto-detect misrouted single-table dirs not named after their table | fallback to single-table when the listing carries a `metadata/` child (`2d1a1d407`) |
| MAJOR-4 — two per-row/per-edge alloc sites in the enumerator hot loop | per-batch fold for ref-edge stats; hoisted map resolution in `index_batch` (`b6387a6a6`) |
| MAJOR-5 — three verification gaps (budget-guard failure arm, live-path keep-min, multi-chunk assert) | negative/engine-level tests added (`b6387a6a6`) |

Cheap minors folded in the same commit (dupParentKeys malformed-payload warn, verify-spool dir suffix, corrected drop instructions and `0 = auto` doc line, `ImportConfig` `#[non_exhaustive]`); the remaining review minors are enumerated as explicitly-deferred in the audit record. Phase B of this PR (the named-gap perf items: IRI group-key admission and the bounded multi-fact columnar join) follows on this same branch.
## 2026-08-04 — Phase B COMPLETE (named-gap perf items)

Phase B lands the two named-gap perf items plus a shared output-bounding mechanism on this branch, each gated + differential-verified, corridor-only (r2rml/iceberg). Live numbers are Substrate B (SF0.1, local MinIO + Iceberg REST, macOS co-resident host, `/usr/bin/time -l` peak RSS, fresh process per cold rep, timing boundary = query_exec_serialize); the DuckDB column is the same-host local yardstick (`~/bin/duckdb`, arch -arm64). Load stayed ~5 throughout.

**P2a — high-cardinality RefObjectMap (IRI) GROUP BY fusion — DONE** (`537a18613`). A `GROUP BY ?c` on a referenced parent IRI (crt_highcard) now fuses: the FK→IRI resolver is built by scanning the parent dim once (deterministic keep-min on a duplicate parent key, byte-identical to `build_parent_lookup`), the fold runs through the existing Dim group-key machinery + the N1 dict, and the key emits as an IRI term (`GKind::Iri`). Hermetic tests: value-identity + dangling-FK drop, keep-min on a duplicate parent key, ~100k-group cardinality (now drained across multiple output batches). Gated by `FLUREE_FUSED_R2RML_AGG_JOIN`; off byte-identical.

**P3 — bounded multi-fact branching-star join (crt_join_reorder class) — DONE** (`de0f2b921` scaffolding, `82ff82940` soundness core, `59c4f06b7` const-object fix). A fact root that branches to one GROUP-KEY dim chain AND one SEMI-JOIN dim chain — crt_join_reorder = FactOrderLine → {`edw:order`→FactOrder→`edw:customer`[segment=Enterprise]} + {`edw:product`→DimProduct[isCurrent]→category}, `SUM(quantity) GROUP BY category` — now FUSES instead of declining (the 917×/DNF gap). The GROUP-KEY branch resolves through the shared linear-chain resolver (`resolve_linear_chain_at_open`, extracted from `resolve_join_at_open`, byte-identical on the pure-linear path) with the SEMI-JOIN FK stripped from the fact's required objects. The SEMI-JOIN branch resolves to a KEEP-MIN-THEN-FILTER membership set: each hop resolves to the keep-min parent subject (byte-identical to `parent_key_insert_keep_min`), THEN the kept row is filtered — so on a duplicate intermediate key it tests the keep-min row, matching the generic chained inner join (filter-then-union would wrongly admit a key whose keep-min row fails but a discarded duplicate passes). Constraint gathering is an OR over BOTH rewrite forms — folded `star_constraints` AND the standalone const-object (`predicate_filter` + `object_constant`) the rewrite emits for a subject with no var-object member — declining if a constraint can't resolve to a scalar column. The membership build is budget-charged (`record_alloc` + `checkpoint`) → typed `MemoryBudgetExceeded` (507) before OOM. Gated by `FLUREE_FUSED_R2RML_MULTIFACT` (default on; off = pre-P3 byte-identical decline).

P3 live (Substrate B, SF0.1, same-host before/after, N=5 cold): crt_join_reorder goes from the DNF/decline class to **fused** — wall **~83,000 → ~1,020 ms (~81×)**, peak RSS **~1.4 GB → ~260 MB (~5.4×)**, result EXACT-matches DuckDB (10 rows; DuckDB same-host ~250 ms). "Before" = the current-main baseline binary, which declines and materializes the 4-table join at ~83 s / 1.2–1.4 GB. Hermetic differentials (mock-provider `open()`/`next_batch()`): value-identity (integer + exact-decimal i128 SUM), the dup-ORDER_KEY keep-min-then-filter discriminator plus a subject-swapped control (proving keep-min — not "any duplicate" — decides), built so a filter-then-union implementation FAILS it; budget-exceeded typed fail; switch-off decline; and both constraint forms.

**Output-bounding — DONE** (`5bd3d52dc`). A high-cardinality fused GROUP BY materialized its entire output in one `next_batch` batch; output-bounding drains the finalized rollup in bounded chunks (`OUTPUT_BOUND_ROWS` = 8192 per call, popped from the back so each group's key + accumulators are freed as it is finalized), one mechanism shared by the single-table fold, the fact-dim join, and the P3 branching-star join (they all reach the same emission). Gated by `FLUREE_FUSED_R2RML_OUTPUT_BOUND` (default on; off = single-batch, byte-identical — same rows, dict/map order). Hermetic: a >8192-group rollup emits multiple ≤8192-row batches whose union is the full result (on) vs one batch (off), same total; the 100k-group cardinality test now drains all batches.

**Output-bounding as an attribution experiment — correcting the P2a residual hypothesis.** The earlier P2a yardstick attributed crt_highcard's ~1.1 GB residual to the 259k-row grouped *output* materialization (serialize + `ORDER BY`). The output-bounding ON/OFF flip is the cleanest attribution experiment the program has run, and it REFUTES that hypothesis: a switch that removes the full output materialization moves NOTHING — crt_highcard ON 1139–1432 / OFF 1140–1149 MB, wall ~1410 ms both, 259,418 rows EXACT (DuckDB same-host ~350 ms / 138 MB). So the residual is not output-bound; it lives in the scan + 259k-group accumulator/resolver state — the peak is reached before emission, and a downstream `ORDER BY` Sort re-collects every row regardless of how the operator batches its output. This re-points the next RSS lever at the Gap-3 columnar accumulator-state work (compact / columnar accumulators, SYNTHESIS N1), now filed — a better-targeted follow-up than an output-materialization drop would have been, with an experiment behind it rather than an assumption. Output-bounding is kept on its own merits: correct, gated (off = byte-identical), hermetic chunking proof, no regression on the shared fold+join output path, and it does bound output-heavy shapes that lack a blocking Sort and aid `LIMIT` early-termination — it is simply not the crt_highcard RSS lever.

### Regression / yardstick columns (Substrate B, SF0.1, N=5 cold, fused ON)

| pair | fused (wall / RSS) | DuckDB (wall) | result |
|---|---|---|---|
| crt_join_reorder (P3 branching join) | ~1,020 ms / ~260 MB | ~250 ms | EXACT, 10 rows (baseline ~83,000 ms / 1.4 GB) |
| crt_highcard_groupby (P2a, 259k-group IRI GROUP BY) | ~1,410 ms / ~1,140 MB | ~350 ms | EXACT, 259,418 rows |
| cq027_web_events_by_type_device (fused 1M-row grouped COUNT) | ~378 ms / ~194 MB | ~192 ms | EXACT, 18 rows |
| cq014_channel_mix (fused single-table) | ~225 ms / ~61 MB | ~180 ms | tol-EXACT, 4 rows |
| cq008_revenue_by_region (fused fact→dim rollup) | ~655 ms / ~180 MB | ~238 ms | tol-EXACT, 9 rows |

No regression on the existing fused pairs: cq027/cq014/cq008 are unchanged (P3 only adds the branching-star arm; output-bounding is off-byte-identical and these fit one batch), and crt_join_reorder was re-run on the output-bounding binary (which touches the shared emission) to confirm it stays ~1,020 ms and EXACT.

### Honesty notes (reviewer's non-blocking items + the fixture-fidelity lesson)

- **Fixture-fidelity lesson.** The P3 soundness core initially passed its hermetic differentials yet produced a ~6× SUM over-count live: the tests hand-built the segment constraint as `star_constraints`, but the rewrite emits it as a standalone const-object pattern, so the semi-join applied no filter and admitted every order. The live DuckDB oracle caught it (fused Apparel 325,586 vs 54,606). Fix: the OR-gatherer handles both forms, and the fixtures now use the real const-object representation (would have caught it). Hand-built fixtures must match the rewrite's actual emission shape.
- **v1 shape scope.** The branching star admits exactly one GROUP-KEY + one SEMI-JOIN branch (the crt_join_reorder class); the branch model is a typed list so K-branch is additive; interior-dim group keys, multi-column FKs, and K>2 branches decline (sound, additive later).
- **KeptChainRow budget charge** is conservative (per-level keep-min maps charged, not released across levels) — the fail-loud direction; it never under-charges.
- **Decimal oracle referentiality.** The exact-decimal value-identity test compares against the same `Acc::Exact` finalize the fold uses, pinning the i128 → BigDecimal path bit-for-bit.
- **Test switch convention.** `FLUREE_FUSED_R2RML_MULTIFACT` and `_OUTPUT_BOUND` read uncached under `#[cfg(test)]`; resolver tests call the resolver methods directly (no env), and only the gate tests toggle env.

### Follow-ups (final state)

- **#1583 (RefObjectMap fan-out).** Both P2a's FK→IRI group key and P3's semi-join membership rest on FK single-valuedness (keep-min ⇒ semi-join ≡ inner join); a fan-out RefObjectMap must revisit both sites — cross-referenced in-code at each.
- **Gap-3 columnar accumulator compaction** (SYNTHESIS N1): the crt_highcard ~1.1 GB residual is scan/accumulator-bound, PROVEN so by the output-bounding ON/OFF attribution experiment above (not assumed) — the next RSS lever is compact/columnar accumulator state, now filed, out of Phase-B scope.
- **K-branch / interior-dim / multi-column-FK admissions**: decline-only today, additive on the same typed branch model.


fluree-db-query/src/r2rml/fused_aggregate.rs
```diff
@@ -1887,11 +2399,29 @@ impl FusedR2rmlAggregateOperator {
...
-            };
-            validity_cols.push(col.clone());
-            count_non_null_cols.push(col);
+            match Self::scalar_column_for_var(&pattern, tm, v) {
+                Some((col, _)) => {
+                    validity_cols.push(col.clone());
+                    count_non_null_cols.push(col);
+                }
+                None => {
+                    // P2a: a RefObjectMap object var (e.g. the `?c` group key) has no
```
bplatz, line 2398: **Blocking — dangling FKs are counted when the ref object var is not a GROUP BY key.**

This arm used to `return Ok(None)` for any non-scalar object; it now admits a `RefObjectMap` object var by pushing its FK child columns into `validity_cols`. The comment says *"A present-but-dangling FK is dropped later by the resolver probe"* — but `group_resolver` is only built when that ref var is an admitted **GROUP BY key** (`ref_group_key`). For a bare `COUNT(*)` / `SUM` over `?o :custRef ?c` with no GROUP BY there is no resolver, so only *null* FKs drop and dangling FKs are folded in.

Verified on this branch with a throwaway test using the same data as `ref_iri_group_key_folds_to_parent_iris_and_drops_dangling`, only with `group_by: []` and `CountAll`:

```
PROBE: fused ungrouped COUNT over a ref object var = Some(4) (generic = 3)
```

Order 4's `CFK=99` has no customer row — the generic path yields no triple, the fused path counts it.

Suggested fix: decline unless `v` is the admitted `ref_group_key` var (cheap, and matches the comment's intent), or build the FK resolver for existence-only checking in the ungrouped case.

Secondary: those same columns land in `count_non_null_cols`, so a provider that proves them zero-null lets the manifest `record_count` shortcut fire with dangling FKs included.

Every P2a test currently has a GROUP BY — that's why this slipped through. Worth adding the ungrouped case to the fixtures regardless of which fix you pick.

fluree-db-query/src/r2rml/fused_aggregate.rs
```diff
@@ -943,6 +1122,90 @@ struct GroupKeyResolver {
...
+/// SUBJECT, byte-identical to the generic `build_parent_lookup`, THEN the kept row is
+/// filtered), the probe is byte-identical to the generic chained inner join even on a
+/// duplicate intermediate key.
+///
+/// #1583: semi-join ≡ inner join here rests on each FK being SINGLE-VALUED — one fact
+/// row matches at most one branch row per hop. A fan-out `RefObjectMap` (one child →
+/// many parents) would make a discarded-dup row a legitimate join partner, so a
+/// membership *set* would then under-represent the fan-out; the fan-out follow-on must
+/// revisit this site (mirrors the P2a #1583 caveat on the FK→IRI group key).
+struct SemiJoinSet {
```
bplatz, line 1148: The keep-min-then-filter equivalence argument (and the b / b-control discriminator tests) covers duplicate join keys with **distinct** subjects. It does not cover duplicate keys with the **same** subject: `parent_key_insert_keep_min` treats an equal subject as a benign duplicate and keeps the first row, so if two dim rows share a key *and* mint the same subject but differ in a constraint column or their next-hop FK, this admits/rejects based on whichever row the scan hit first. The generic path effectively asks "does *any* row with that subject satisfy the constraint", and can also fan out to multiple join partners.

Not exotic for the shape being targeted: the fixture's `Product` is `http://ex/prod/{PRODUCT_KEY}` **with an `IS_CURRENT` column** — an SCD-2 dimension where multiple versions per key mint an identical subject is the normal case.

I believe `GroupKeyResolver` has the same modeling gap pre-PR so this isn't newly introduced, but P3 extends it to *constraint evaluation*, where the divergence becomes scan-order-dependent rather than just a fan-out difference. Minimum: document it here next to the #1583 caveat. Better: decline when a level observes a same-key/same-subject pair whose `passes_own` or `next_fk` differ — the key collision is already detected.

fluree-db-query/src/r2rml/fused_aggregate.rs
```diff
@@ -943,6 +1122,90 @@ struct GroupKeyResolver {
...
+}
+
+/// Comparable ordering key for a parent SUBJECT term — a LOCAL mirror of the
+/// materialize crate's private `subject_sort_key` (`graph.rs`). The two MUST agree:
+/// the P3 semi-join keep-min must pick the SAME parent row the generic
+/// `build_parent_lookup` (via `parent_key_insert_keep_min`) picks, so a fused probe is
+/// byte-identical to the generic chained inner join on duplicate intermediate keys.
+/// The dup-ORDER_KEY differential (kept row failing, discarded dup passing) guards any
+/// drift between the two.
+fn semijoin_subject_sort_key(term: &RdfTerm) -> &str {
```
bplatz: This PR makes `canonical_join` `pub` precisely because two copies of one invariant had drifted — and then adds a local mirror of the private `subject_sort_key` (`fluree-db-r2rml/src/materialize/graph.rs:226`), plus a third hand-rolled keep-min (`if iri < *cur`) in the P2a resolver build below. Three copies of the same rule, two of them carrying "these MUST agree" comments.

`parent_key_insert_keep_min` is already `pub` and is exactly this logic; `subject_sort_key` could be exported the same way `canonical_join` just was. Suggest reusing one of them and deleting both copies — a comment is the weakest available enforcement, and this file has already paid for that once.

fluree-db-query/src/r2rml/fused_aggregate.rs
```diff
@@ -4787,4 +6423,717 @@ mod tests {
...
+        );
+    }
+
+    /// OUTPUT-BOUNDING: a high-cardinality GROUP BY (> OUTPUT_BOUND_ROWS groups) emits
+    /// in bounded chunks with the switch ON — multiple batches, each ≤ the bound, whose
+    /// union is the full result — and in a SINGLE batch with the switch OFF (the
+    /// byte-identical pre-bounding path). Same total either way. Single-table ref-IRI
+    /// GROUP BY (one order per customer → one group per customer).
+    #[tokio::test]
+    async fn p3_output_bounding_chunks_high_card_and_off_single_batch() {
```
bplatz, line 7225: `std::env::set_var` from a test, with the switch read uncached under `cfg(test)` specifically to make this work. Rust test threads run in parallel within one process, so `setenv`/`getenv` race here — UB at the libc level, and `set_var` becomes `unsafe` in edition 2024.

Concretely: `ref_iri_group_key_high_cardinality` reads `FLUREE_FUSED_R2RML_OUTPUT_BOUND` while this test has it set to `0`. It passes today only because that test drains all batches.

Also, the comments disagree with each other — `multifact_enabled` says *"Only the gate test toggles this var (under a lock)"*, and `p3_multifact_switch_gates_the_branching_path` says *"so no lock is needed"*. There is no lock either way.

Suggest making the switches operator fields defaulted from env at construction (testable without touching process state), or serializing the toggling tests behind a `static Mutex`.

fluree-db-query/src/r2rml/fused_aggregate.rs
```diff
@@ -1456,18 +1838,66 @@ impl Operator for FusedR2rmlAggregateOperator {
...
+                            hash,
+                            |(k, _)| {
+                                k.len() == scratch.len()
+                                    && scratch.iter().zip(k).all(|(r, o)| r.eq_owned(o))
+                            },
+                            |(k, _)| gkeys_hash(k),
+                        ) {
+                            hashbrown::hash_table::Entry::Occupied(o) => o.get().1,
+                            hashbrown::hash_table::Entry::Vacant(v) => {
+                                let id = group_accs.len() as u32;
```
bplatz: `group_accs.len() as u32` wraps silently past 4B groups and would corrupt accumulators rather than fail. The memory budget normally trips first, but it isn't set on every path — a `debug_assert!` or an explicit check here is free.

fluree-db-query/src/r2rml/fused_aggregate.rs
```diff
@@ -1249,6 +1552,43 @@ impl FusedR2rmlAggregateOperator {
...
+            .as_ref()
+            .ok_or_else(|| QueryError::Internal("fused aggregate not resolved".to_string()))?;
+        let gcols = &resolved.group_cols;
+        let folds = &resolved.folds;
+        let num_cols = gcols.len() + folds.len();
+        let rows = self
+            .pending_rows
+            .as_mut()
+            .ok_or_else(|| QueryError::Internal("no pending fused output".to_string()))?;
+        let take = OUTPUT_BOUND_ROWS.min(rows.len());
```
bplatz, line 1553: Popping from the back reverses chunk order relative to the switch-off single-batch path. That's fine per SPARQL (group order is unspecified), but it means `LIMIT` without `ORDER BY` returns *different rows* with the switch on vs off. The doc comment claims byte-identity only for the OFF path, so this is consistent — just worth one sentence here so the next reader doesn't have to derive it.

fluree-db-api/src/materialize.rs
```diff
@@ -1180,13 +1236,33 @@ pub enum VerifyMode {
...
     Match,
-    /// A count differs between the source and the twin.
-    Mismatch { source: u64, twin: u64 },
+    /// A count differs between the source and the twin. `note` carries a caveat
+    /// (rendered with the failure) about why a count difference can be spurious even
+    /// on a faithful twin — see [`COUNT_MULTISET_NOTE`].
+    Mismatch {
+        source: u64,
+        twin: u64,
+        note: Option<&'static str>,
```
bplatz: This field is set to `Some(COUNT_MULTISET_NOTE)` at both construction sites, unconditionally — it carries no information, but it *is* a breaking change to a public enum, it's in the derived `PartialEq` (which is why `it_graph_source_r2rml.rs` had to move from `assert_eq!` to `matches!` with `..`), and it forces every downstream matcher to spell out an extra field.

Rendering the caveat in the reporting layer for `count:*` checks would get the same operator-facing benefit without the API change.

fluree-db-api/src/materialize.rs
```diff
@@ -3164,4 +3320,98 @@ mod tests {
...
+            let rt = tokio::runtime::Builder::new_multi_thread()
+                .worker_threads(2)
+                .enable_all()
+                .build()
+                .unwrap();
+            let shared_alloc = Arc::new(SharedNamespaceAllocator::from_registry(
+                &NamespaceRegistry::new(),
+            ));
+            let spool_dir =
+                std::env::temp_dir().join(format!("fluree-c2-test-{}", std::process::id()));
```
bplatz: Same collision-on-concurrent-runs problem that this very commit fixed for the verify spool dir with `VERIFY_DIR_SEQ` — two runs in one process share this path. Worth the same atomic suffix for consistency.

fluree-db-query/src/r2rml/provider.rs
```diff
@@ -214,6 +214,22 @@ pub trait R2rmlProvider: Debug + Send + Sync {
...
+
+    /// MAJOR-2 (#1529 review): assert the snapshots read during a materialize build
+    /// are trustworthy — every table stayed on ONE `metadata_location`, and pinning
+    /// is actually in effect. Returns `Err(reason)` (a human-readable string) when a
+    /// build must be refused. Default: `Ok` — a non-Iceberg / test provider has no
+    /// moving-snapshot hazard. A real Iceberg-backed provider overrides this to
+    /// refuse when the loadTable cache is disabled (pinning is then a no-op) or when
+    /// a table yielded a second distinct `metadata_location` mid-build (the source
+    /// committed and the twin's stamped watermark would not describe its contents).
+    fn verify_build_snapshot_integrity(
```
bplatz, line 226: `Result<(), String>` on a public trait method conflicts with the project's stated error convention (`thiserror`, structured per-crate enums, not stringly errors). Both call sites immediately wrap the `String` back into `R2rmlError::Materialization`, so the type isn't buying anything.

Separately: this turns `FLUREE_ICEBERG_LOADTABLE_CACHE=0` into a hard materialize failure. That's a user-visible behavior change and only `docs/cli/materialize.md` got updated — `docs/operations/configuration.md` documents that env var and should mention the new refusal.

fluree-db-cli/src/commands/materialize.rs
```diff
@@ -115,6 +115,32 @@ pub async fn run(dirs: &FlureeDir, params: &MaterializeParams<'_>) -> CliResult<
     let fluree: &'static Fluree = Box::leak(Box::new(context::build_fluree(dirs)?));
     let provider = Arc::new(FlureeR2rmlProvider::new(fluree));
 
+    // CRITICAL-1 (#1529 review): a failed parity gate drops the WHOLE ledger NAME
```
bplatz, line 120: This fixes the symptom rather than the cause. The actual hazard is that a failed parity gate performs a *name-scoped* `DropMode::Hard` across every branch. Refusing to build when the name has a live non-target branch closes it for the CLI, but the destructive drop stays unchanged and unguarded for anyone driving the API directly.

Scoping the drop to the branch the build created would make this guard belt-and-braces instead of the only thing between a user and losing unrelated branches. If that's out of scope here, worth a tracked issue linked from this comment.

fluree-db-r2rml/src/materialize/graph.rs
```diff
@@ -410,16 +415,30 @@ impl ParentIndexSet {
         let mut ambiguous_this_batch: Vec<Vec<String>> = Vec::new();
         {
             let entry = self.index.entry(tm.iri.clone()).or_default();
+            // MAJOR-4 (#1529 review): create each canonical col-set's map ONCE (one
```
bplatz, line 425: Pre-creating every col-set's map means a col-set that never yields a key now leaves an **empty** map where previously there was **no entry**. Worth confirming no downstream `get` / `contains_key` distinguishes "absent" from "present but empty" — the perf win is right, this is just the one behavioral edge it introduces.

fluree-db-iceberg/src/io/arrow_reader.rs
```diff
@@ -332,6 +352,96 @@ fn arrow_column_to_values(
     }
 }
 
+/// N2 kill switch (`FLUREE_ARROW_DIRECT_DECODE`, default **on**). Off
+/// (`0`/`false`/`off`/`no`, trimmed + case-insensitive per the switch family)
+/// restores the byte-identical two-hop `arrow_column_to_values` +
+/// `build_columns_from_values` path. Read once per process (`OnceLock`, the family
+/// idiom): set it at startup, not per query.
+fn direct_decode_enabled() -> bool {
```
bplatz, line 361: This hand-rolls the env parsing that `env_switch_enabled` does for the other three switches. Different crate so there's no direct reuse, but the family is now four switches with one odd one out — a small shared helper (or moving the parser somewhere both crates can see) would keep them from drifting on the next accepted-value tweak.

fluree-db-iceberg/src/io/parquet.rs
```diff
@@ -632,169 +632,177 @@ pub fn build_columns_from_values(
...
+
+    Ok(columns)
+}
+
+/// Map one column's row-collected `ColumnValue`s to the typed [`Column`] for
+/// `field_type`. Extracted verbatim from [`build_columns_from_values`] so the
+/// Arrow direct decoder ([`crate::io::arrow_reader::arrow_column_to_column`]) can
+/// reuse the EXACT same `ColumnValue` -> `Column` mapping for its cross-type
+/// fallback — this keeps the direct decode path byte-identical to the two-hop path.
+pub fn column_from_values(values: &[Option<ColumnValue>], field_type: &FieldType) -> Column {
```
bplatz, line 648: Good extraction — this is the right way to keep the direct path from diverging. Minor: now that all the per-field work lives here and this is infallible, `build_columns_from_values` above still returns `Result` though it can no longer fail. Worth dropping the `Result` while the callers are being touched anyway.

bplatz (commented):
Review of the fused vectorized fold + late-rendered decode seam. Overall this is careful work — the P3 keep-min discriminator pair (b / b-control) is exactly the right test shape, the 2500-key grow test genuinely guards the split-group hazard, the `canonical_join` transposition fix is a real latent-bug fix with a regression test, and CRITICAL-2's `drop(work_rx)` + unconditional join with a timeout-guarded hang test is properly done.

One blocking correctness issue (dangling FKs counted in an ungrouped aggregate over a RefObjectMap object var — verified locally, see inline), one soundness caveat worth documenting or gating, and a handful of smaller cleanups. Details inline.

fluree-db-query/src/r2rml/fused_aggregate.rs
aaj3f: Great catch, and confirmed — this was a real over-count. You're exactly right that `group_resolver` only exists when the ref var is an admitted GROUP BY key (`ref_group_key`), so a bare `COUNT(*)`/`SUM` over `?o :custRef ?c` with no GROUP BY reaches `:2407`, admits on FK-non-null validity, and nothing ever probes the parent — so a dangling FK folds in. I reproduced it on the same data as `ref_iri_group_key_folds_to_parent_iris_and_drops_dangling` but with `group_by: []` + `CountAll`: fused = `Some(4)`, generic = `3` (order 4's `CFK=99` has no customer row). Fixed in `560fb1d4c` by taking the cheap path you suggested — decline unless the ref object var is the admitted `ref_group_key` (`if !self.group_by.contains(&v) { return Ok(None) }`) — which also closes the secondary `count_non_null_cols` / `record_count`-shortcut hole you flagged, since a declined shape never reaches the manifest shortcut. Added the ungrouped case to the P2a fixtures so it can't creep back.

fluree-db-query/src/r2rml/fused_aggregate.rs
aaj3f: Agreed, and you're right that the SCD-2 case (`IS_CURRENT` on `Product`, multiple versions per key minting the same subject) is the normal case for this shape, not exotic — so scan-order-dependent constraint evaluation is a real hazard, not just a fan-out difference. I took the decline rather than only documenting it: `0797bdc2a` declines when a level observes a same-key/same-subject pair whose `passes_own` or `next_fk` differ (the key collision is already detected, so it's cheap). And you're right it's a pre-existing modeling gap in `GroupKeyResolver` — P3 is what extends it to *constraint evaluation*, where the divergence turns scan-order-dependent, so declining is the conservative call for what this PR actually introduces.

fluree-db-query/src/r2rml/fused_aggregate.rs
aaj3f: Fair — and a little embarrassing, given this PR makes `canonical_join` `pub` precisely because two copies had drifted, then adds a third hand-rolled keep-min below. `d8ad558cd` exports `subject_sort_key` the same way `canonical_join` just was and single-sources the comparator across the three sites, so a comment isn't the only thing holding them together anymore. One caveat worth stating: the value types differ (`parent_key_insert_keep_min` stores `RdfTerm`, P2a stores `Vec<GKey>`, the semi-join stores `KeptChainRow`), so it's the sort-key comparator that's shared, not a single insert fn — but that's the part that was drifting, so it's enough.

fluree-db-query/src/r2rml/fused_aggregate.rs
aaj3f: You're right, and this is the load-bearing one — thank you. Test threads share the process, so `setenv`/`getenv` race, and it's UB regardless of how carefully one switch is guarded (and `set_var` goes `unsafe` in edition 2024). `59d71e23a` makes the four switches this PR owns (`FLUREE_FUSED_VECTOR_FOLD`, `_OUTPUT_BOUND`, `_MULTIFACT`, `FLUREE_ARROW_DIRECT_DECODE`) operator fields defaulted from env at construction, so the tests set a field and touch no process state — no lock, no `cfg(test)` uncached read. The same defect shows up in #1589 (its `GEN_ENV_LOCK` band-aid over only `_MULTIFACT_GEN`); that PR converts `_MULTIFACT_GEN` and drops the lock, so the stack lands the fix coherently rather than in half-measures. Also fixed the two comments that contradicted each other on whether a lock existed.

fluree-db-query/src/r2rml/fused_aggregate.rs
aaj3f: Good cheap insurance — `e5a56a51a` adds a checked cast so `group_accs.len()` past `u32::MAX` fails loudly instead of wrapping and corrupting accumulators. Agreed the memory budget usually trips first, but it isn't set on every path, so the explicit check is worth the zero cost.

fluree-db-query/src/r2rml/fused_aggregate.rs
aaj3f: Right — popping from the back reverses chunk order relative to the switch-off single-batch path, so `LIMIT` without `ORDER BY` can return *different rows* on vs off. It's spec-legal (group order is unspecified) and the byte-identity claim is already scoped to the OFF path, so it's consistent — just under-documented. `ad7e8c79d` adds the one sentence so the next reader doesn't have to derive it.

fluree-db-api/src/materialize.rs
aaj3f: Agreed the field carries no information at either construction site, and you're right it's a breaking change to a public enum in the derived `PartialEq` (which is exactly why `it_graph_source_r2rml.rs` had to move from `assert_eq!` to `matches!`). `59bcb5b5f` renders the count-multiset caveat in the reporting layer for `count:*` checks instead — same operator-facing benefit without the API change or forcing every downstream matcher to spell out the extra field.

fluree-db-api/src/materialize.rs
aaj3f: You're right about the pattern, and I applied the same atomic suffix in `96538ef78` for consistency. One honest correction on severity, so the record's straight: `:3385-3386` is inside a `#[cfg(test)]` path (the CRITICAL-2 poison-hang test), building `fluree-c2-test-{process_id}` — unlike the *production* verify spool dir that this same commit gave the `VERIFY_DIR_SEQ` suffix. Only that one test uses the prefix, and it creates+removes its own dir, so the live collision risk is effectively nil (it isn't concurrent with itself). So it's cheap test hygiene, worth doing — I just didn't want to bill it as the production hazard the verify-dir fix addressed; it's a different severity tier.

fluree-db-query/src/r2rml/provider.rs
aaj3f: Agreed on both. `0db38cbec` gives `verify_build_snapshot_integrity` a typed per-crate error instead of `Result<(), String>` — both call sites were just re-wrapping the `String` into `R2rmlError::Materialization`, so the type wasn't buying anything. And you're right that turning `FLUREE_ICEBERG_LOADTABLE_CACHE=0` into a hard materialize failure is a user-visible change that only `docs/cli/materialize.md` had captured — I added the refusal to `docs/operations/configuration.md` in the same commit.

fluree-db-cli/src/commands/materialize.rs
aaj3f: You've put your finger on the real hazard, and I want to be honest that this commit doesn't fully close it. The CLI-level up-front refusal (`blocking_existing_branch`, `cli/materialize.rs:118-142`) closes it for the CLI, but the destructive name-scoped `DropMode::Hard` — dropping *every* branch of the ledger name on a failed parity gate — is unchanged and still unguarded for anyone driving the API directly. Scoping the drop to the branch the build created is the right belt-and-braces fix, but it's more than this review's scope, so I took the tracked-issue option you offered: `e9a1592b2` links #1592 from the comment. The CLI is the only driver today, so it's non-blocking now — but #1592 keeps the API-direct hazard from getting lost.

fluree-db-r2rml/src/materialize/graph.rs
aaj3f: Good edge to flag. I checked the downstream `get`/`contains_key` sites and none distinguish "absent" from "present but empty" — a col-set that never yields a key just leaves an empty map that reads the same as a miss. `88ee9a6a4` notes it's safe-by-construction next to the change so the reasoning is on the record; the perf win stands.

fluree-db-iceberg/src/io/arrow_reader.rs
aaj3f: Agreed — four switches with one odd one out is exactly how they drift on the next accepted-value tweak. Different crate, so there's no direct reuse of `env_switch_enabled`, but `94c097a55` adds a crate-local equivalent in `fluree-db-iceberg` so the arrow-decode switch parses the same way as the other three.

fluree-db-iceberg/src/io/parquet.rs
aaj3f: Agreed, good cleanup while the callers are already being touched. After the extraction `build_columns_from_values` only wraps an infallible `column_from_values` in `Ok(...)`, so `8eeb80126` drops the `Result`. The two in-repo callers used `?`; I confirmed no out-of-crate callers rely on the `Result` before changing the `pub` signature.

aaj3f:
Thanks for the thorough pass, @bplatz — genuinely good review, and two of the notes were real correctness findings rather than nits. Quick summary of what landed:

The blocking one (`:2408`) was real: an ungrouped `COUNT`/`SUM` over a ref object var counted dangling FKs, because the resolver only exists for GROUP BY ref keys. Reproduced (fused=4 / generic=3) and fixed by declining unless the var is the admitted `ref_group_key` (`560fb1d4c`), which also closes the `record_count`-shortcut hole. The SCD-2 same-key/same-subject collision (`:1172`) was the other substantive one — now declines when `passes_own`/`next_fk` differ (`0797bdc2a`).

The load-bearing maintainability item was the `set_var` test UB (`:6846`) — the four switches are now operator fields defaulted from env, so tests touch no process state (`59d71e23a`), and #1589 does the same for `_MULTIFACT_GEN` so the stack is coherent. The rest landed as suggested: single-sourced keep-min (`d8ad558cd`), checked group-id cast (`e5a56a51a`), reporting-layer caveat instead of the vacuous enum field (`59bcb5b5f`), typed provider error (`0db38cbec`), infallible-`Result` drop (`8eeb80126`), crate-local env parse (`94c097a55`), and the two doc notes.

Two honest notes so the record's straight: (1) the tmp-dir collision (`:3386`) is a `#[cfg(test)]`-only path, not the production hazard the verify-dir suffix addressed — I applied the same atomic suffix for consistency, but it's a different severity tier. (2) The `DropMode::Hard` hazard (`:118`) — the CLI refusal closes it for the CLI, but the destructive name-scoped drop is unchanged for API-direct callers, so I filed #1592 rather than claim it's fully fixed; scoping the drop is the follow-up.

Re-requesting review when you have a moment — replies inline on each thread.

bplatz (approved):
Looks good!

PR #1584, opened 2026-08-01
BM25 index management over the server API

zonotope:
Adds server-side management of BM25 full-text indexes: HTTP endpoints for creating and syncing them, an opt-in worker that keeps them current automatically, and CLI support for driving all of it against a remote server. Before this branch a BM25 index could only be built and maintained in-process — through the Rust API or a CLI run on the same filesystem — so a hosted deployment had no way to manage one.

### HTTP endpoints

Two new admin-gated routes, mounted in `v1_admin_protected_writes` so they sit behind `require_admin_token` and, under Raft, leader-forwarding:

- `POST /v1/fluree/bm25/create` — build an index over a source ledger from an indexing query. Accepts `name`, `ledger`, `query`, and optional `branch`, `k1`, `b`. Returns `201` with the resulting alias and index stats.
- `POST /v1/fluree/bm25/sync` — bring an index up to date. `?t=<t>` pins how far to sync; omitted, it syncs through the source's head. A no-op sync returns `200` with zero counts.

Only creation and sync need family-specific routes. An index is a graph source, so listing, inspection, and drop are already served by `GET /ledgers`, `GET /info/<alias>`, and `POST /drop` — the same division the Iceberg family uses.

`GET /ledgers` gained a `dependencies` field on graph-source entries, naming the source ledgers each derives from. That is what lets a client pair an index against its source's `t` — the staleness check behind `fluree bm25 list` — from a single response. The field is omitted for ledger entries rather than serialized as an empty array.

### Automatic sync

A server started with `--bm25-auto-sync` (env `FLUREE_BM25_AUTO_SYNC`, or `indexing.bm25_auto_sync` in the config file) syncs each index whenever its source ledger commits. **Off by default**, because it introduces background storage writes on a server that previously performed none for graph sources.

The existing `Bm25MaintenanceWorker` was written but never instantiated, and was `!Send` — five `Rc<RefCell<…>>` fields plus a borrowed `&'a Fluree` — so it could not be spawned onto a multi-threaded runtime. It now uses `Arc<parking_lot::Mutex<…>>` and `Arc<AtomicBool>`, and owns an `Arc<Fluree>` so it is `'static` and movable into a task. Three compile-time assertions pin `Send`, `Send + 'static`, and the `Send`-ness of `run()`'s future, since these can only regress at compile time.

Gating: skipped in peer mode, where writes are forwarded to the transaction server and the driving commit events are published on that node's bus rather than this one. Under Raft it joins the leader-task set alongside the indexer and evictor, because a sync publishes through `GraphSourcePublisher`, which proposes `PublishGraphSourceIndex` and must originate on the leader. On shutdown the worker is asked to stop rather than aborted, so an in-flight sync finishes its publish.

Startup performs a registration pass over `all_graph_source_records()`; the worker's `auto_register` only picks up indexes created while it is running, so without it an index created before boot would never sync.

### CLI

`fluree bm25 create`, `sync`, `drop`, and `list` all take `--remote <name>` and route through the endpoints above, following the same resolution `fluree iceberg` uses: an explicit remote, otherwise a locally-running server, and `--direct` to force in-process execution. The in-process route is unchanged and still works under `docker exec` against a directory a server is already serving. `sync` also gained `--t` to reach the endpoint's pinned form.

`bm25 list` computes staleness on both routes from a single listing. Against a server too old to send `dependencies`, the source is unknown and the index reports as not-stale — so `list` prints a warning naming the affected indexes rather than letting the listing quietly look current, which would leave a `--stale` maintenance loop doing nothing.

## Bugs fixed along the way

**Dropping a BM25 index leaked every snapshot.** `drop_graph_source` — the path behind `POST /drop` and `fluree drop` — retracted the nameservice record but only deleted Iceberg mapping blobs. A BM25 index dropped that way left all its snapshots on disk with nothing able to reach them. Only `drop_full_text_index` (`fluree bm25 drop`) ever swept them. The sweep is now shared by both paths via `delete_bm25_snapshots`, which takes the manifest rather than loading it so each caller keeps its existing ordering relative to retraction.

**`POST /drop` discarded its own file count.** The graph-source branch hardcoded `files_deleted: None`, so a hard drop that removed artifacts was indistinguishable from one that removed nothing. This predates BM25 — the Iceberg branch already incremented the count and the route dropped it too.

**Remote `bm25 drop` reported success for a nonexistent index.** `POST /drop` answers `200` with `status: "not_found"`, so the CLI printed a success line and exited `0` over a server while erroring under `--direct`. A script checking exit codes got different answers depending on whether a server happened to be reachable.

## Docs

`docs/cli/bm25.md` is new — the `fluree bm25` command reference, which the CLI docs never had. `docs/api/endpoints.md` documents both endpoints in the shape the adjacent Iceberg entry uses, and its `GET /ledgers` response was corrected: it had documented a `{"ledgers": [...]}` object with `ledger_id`, `commit_t`, `created`, and `last_updated` fields, none of which exist. The handler returns a bare array of `{name, branch, type, t}`. Anyone who wrote a client against the documented shape was getting parse errors.

The BM25 guide and graph-source pages had several claims that this branch made false — that index creation had no HTTP route, that the CLI runs in-process rather than calling a server, and that nothing syncs automatically on commit — all corrected.

## Tests

- `fluree-db-server/tests/bm25_http_integration.rs` (new, 15 tests) — create, sync in both forms, the no-op sync, `?t` validation, 404s, the drop sweep, and the `dependencies` field.
- `fluree-db-server/tests/bm25_auto_sync.rs` (new, 4 tests) — a commit advancing the watermark with nobody calling sync, its control showing the watermark provably does not move without a worker, and that the flag defaults off.
- `fluree-db-cli/src/commands/bm25.rs` (14 unit tests) — request-body shaping, remote row building and staleness, the version-skew degradation, and the drop-status mapping.
- Additions to the API graph-source suite, including an on-disk assertion that a hard drop actually removes snapshot blobs.

The snapshot-leak test was checked against the unfixed code and fails there with `files_deleted: 0`, so it regresses the bug rather than merely passing.

## Known limitations and follow-ups

- **`/swagger.json` is a hand-written stub.** It describes 6 paths; the router mounts roughly 67. This branch adds the two BM25 paths to it, which makes BM25 its best-documented area — lopsided, and an artifact of what was touched last rather than what matters. `utoipa` is already a non-optional dependency with no handler annotated. Written up in `issue-openapi-spec-is-a-stub.md`.
- **Some client errors surface as `500`.** Syncing a dropped index and requesting a `t` beyond the source's head are both refused, but with the wrong status. Written up in `issue-server-ledger-notfound-maps-to-500.md`.
- **Vector graph sources leak on drop, and `DropMode::Hard` silently degrades to `Soft` for them.** Not the same defect as the BM25 leak — neither vector path deletes, deliberately, pending a time-travel question that has to be settled before the BM25 fix can be copied across. Latent today: `vector` is a non-default feature neither the server nor CLI enables. Written up in `issue-vector-graph-source-drop-never-reclaims-blobs.md`.
- **`vector_worker.rs` is a near-identical `Rc<RefCell>` twin** of the BM25 worker and would need the same treatment before vector auto-sync were possible.
- **`bm25 list` needs a server new enough to send `dependencies`.** It degrades with a warning rather than an error; an N+1 fallback over `GET /info/<alias>` would remove the requirement if older servers need supporting.


fluree-db-api/src/bm25_worker.rs
```diff
@@ -418,7 +415,7 @@ impl<'a> Bm25MaintenanceWorker<'a> {
 
         loop {
             // Check for stop request
-            if *self.stop_requested.borrow() {
+            if self.stop_requested.load(Ordering::Relaxed) {
                 info!("BM25 maintenance worker stopping");
                 break;
             }
```
aaj3f, line 473: **`fluree-db-api/src/bm25_worker.rs:427-438`** — *optional, but the one I'd most want a second look at.* Nothing stops the same index from being synced twice concurrently. `pending` is a `HashSet`, so it dedupes — but a graph source is removed from `pending` at `:432` and pushed into `in_flight` at `:438` with no record that it is already running, so a commit arriving while a sync is in flight re-inserts it and the next flush starts a second sync of the same index. Concretely: commit at `t=5` flushes a sync, that sync is still running (it re-executes the whole indexing query, so on a real corpus that is seconds), commit at `t=6` arrives, debounce elapses, `in_flight.len()` is 1 which is `< 4`, and now two `sync_bm25_index` calls are walking the same manifest. Both read manifest `M`, both write a snapshot, both blind-publish — so one snapshot is orphaned on disk, the published watermark can go backwards if the higher-`t` sync publishes first, and the work is duplicated. It does self-heal on the next sync, which is why I'm not calling it blocking. The loop is pre-existing and unchanged here, but this PR is what instantiates the worker for the first time, so it's the first release where it's reachable. Minimally an `in_flight_ids: HashSet<String>` checked at `:428` and cleared in the completion arm at `:483` would close it; happy to punt to a follow-up issue if you'd rather ship this and fix it there.

*Commenting on line 421 because `fluree-db-api/src/bm25_worker.rs:427` is not an addressable diff line.*

fluree-db-api/src/graph_source/bm25.rs
```diff
@@ -1033,19 +1045,7 @@ impl crate::Fluree {
         info!(graph_source_id = %graph_source_id, "Starting BM25 full resync");
```
aaj3f, line 1045: **`fluree-db-api/src/graph_source/bm25.rs:975`** — *optional / scalability.* Step 8 re-runs `execute_bm25_indexing_query` over the entire source ledger and *then* filters to the affected subjects, so the "incremental" sync is O(corpus) per call rather than O(delta). That's pre-existing and it was fine when a sync was something you ran from a maintenance job. `--bm25-auto-sync` changes the economics: it now runs once per debounce window of commits, so a write-heavy source ledger with a large corpus is doing a full indexing-query execution every 100ms-ish, and combined with the concurrency note above it can be doing several at once. I don't think this blocks the flag landing — it's off by default and you've documented it that way — but it seems worth a line in `docs/indexing-and-search/bm25.md` saying auto-sync's cost scales with corpus size and not with the size of the commit, so nobody turns it on over a few million documents expecting a cheap delta.

*Commenting on line 1045 because `fluree-db-api/src/graph_source/bm25.rs:975` is not an addressable diff line.*

fluree-db-server/src/lib.rs
```diff
@@ -54,7 +54,49 @@ pub use state::AppState;
...
+use fluree_db_api::{Bm25MaintenanceWorker, Bm25WorkerHandle, Fluree};
 use std::sync::Arc;
+
+/// Build a BM25 maintenance worker seeded with the indexes that already exist.
+///
+/// `auto_register` only picks up indexes created while the worker is running,
+/// so without this pass an index created before startup would never sync.
+/// Failing to enumerate is not fatal: those indexes stay unregistered until
+/// their next config publish, and ones created from here on still register.
+async fn build_bm25_worker(fluree: Arc<Fluree>) -> (Bm25MaintenanceWorker, Bm25WorkerHandle) {
```
aaj3f, line 116: **`fluree-db-server/src/lib.rs:66-92`** — *optional.* `build_bm25_worker` and the gating around it at `:368-393` — flag off, peer-mode skip, Raft-leader ownership — have no test. `bm25_auto_sync.rs:159` (`dropped_indexes_are_not_registered`) reads as coverage for the registration pass, but it re-implements the predicate rather than calling `build_bm25_worker`: it does its own `all_graph_source_records()` and its own `.filter(|gs| gs.is_bm25() && !gs.retracted)`, so if the filter in `build_bm25_worker` changed the test would still pass. The end-to-end behavior *is* well covered — `commit_advances_the_index_without_an_explicit_sync` genuinely drives the worker and asserts the watermark moves, and its control test is a nice touch — it's specifically the server-side wiring that isn't. `build_bm25_worker` being private is what blocks it; if it took the record list as an argument, or if the filter moved into a `pub fn eligible_for_auto_sync(&GraphSourceRecord) -> bool` on the nameservice record, the test could assert against the real thing. Cheap insurance given how much of the feature's correctness lives in that one gate.

fluree-db-server/src/lib.rs
```diff
@@ -419,6 +490,11 @@ impl FlureeServer {
         if let Some(task) = ledger_maintenance_task {
             task.abort();
         }
+        // Ask the BM25 worker to stop rather than aborting it, so an in-flight
+        // sync finishes its publish instead of being cut mid-write.
+        if let Some(handle) = bm25_auto_sync {
```
aaj3f: **`fluree-db-server/src/lib.rs:495-496`** — *optional / a claim I couldn't fully verify.* The comment says the worker is asked to stop "rather than aborting it, so an in-flight sync finishes its publish instead of being cut mid-write" — but `handle.stop()` only flips an `AtomicBool` and nothing joins the spawned task afterward, so once `serve()` returns and the runtime drops, an in-flight sync is dropped at its next await point anyway. It's still strictly better than `abort()`, so this is really about the comment promising more than the mechanism delivers; a `task.await` (or a bounded `timeout` around one) would make it true. Related: under Raft the worker goes into the leader task set at `:943`, and `raft.rs:592` does `abort_and_await` on leadership change — so that path gets exactly the abort semantics this one deliberately avoids. That's arguably correct for a demotion (you don't want an ex-leader finishing a publish), but the two paths differ and only one of them is commented.

fluree-db-api/src/bm25_worker.rs
aaj3f, line 473: **`fluree-db-api/src/bm25_worker.rs:457`** — *nit.* The comment above the `select!` reads "Prefer stop checks + flushing, but still service events promptly", but `biased;` polls the branches in source order and `subscription.receiver.recv()` is first, so the actual preference is the opposite: events win, and the in-flight completion arm and the debounce tick only get polled when no event is queued. Pre-existing wording, just easy to fix while the file is open.

*Commenting on line 421 because `fluree-db-api/src/bm25_worker.rs:457` is not an addressable diff line.*

fluree-db-cli/src/cli.rs
```diff
@@ -402,13 +402,13 @@ pub enum Commands {
...
     /// Create, list, sync, or drop BM25 full-text search indexes (graph sources).
     ///
-    /// BM25 index creation has no HTTP endpoint today — it is a Rust-API-only
-    /// operation (`Bm25CreateConfig` + `create_full_text_index`). These commands
-    /// expose it: they run in-process against local storage (no server
-    /// round-trip), so they work under `docker exec` against a running server's
-    /// data directory. Querying the resulting index is done either through the
-    /// standalone `fluree-search-httpd` service (`POST /v1/search`) or, embedded,
-    /// via an FQL `f:searchText` query.
+    /// These commands run against a server when one is reachable — `--remote
```
aaj3f, line 405: **`fluree-db-cli/src/cli.rs:405-411`** — *optional / compatibility.* Worth calling out explicitly somewhere user-facing that the default route flipped. `fluree bm25 sync` with no flags used to always run in-process; it now auto-routes to a locally-running server whenever `server.meta.json` points at a live pid (`context.rs:738-762`), and that client is constructed with no token. On a server running `admin_auth.mode == None` that's invisible, but on an auth-configured server the exact `docker exec` workflow the old docs recommended now returns 401 instead of working. The docs do tell people about `--direct`, so this is mostly about existing cron jobs and scripts silently changing behavior on upgrade — maybe a line in the CLI docs' migration-shaped prose, or nothing at all if you think the `routing through local server at …` stderr line is loud enough.

docs/api/endpoints.md
```diff
@@ -1617,29 +1617,29 @@ GET /ledgers
```
aaj3f, line 1617: **`docs/api/endpoints.md`** (sync status codes) — *nit.* The status-code list for `POST /bm25/sync` gives 200/400/401/403/404, and then the prose says syncing a dropped index and requesting a `t` beyond the head are "refused" — but per your own follow-up note both of those currently come back as 500s. A reader working from the table will write a client that treats 500 as retryable and retry forever. Since the follow-up issue is already written up, maybe just a parenthetical: "(both currently surface as `500`; see issue-server-ledger-notfound-maps-to-500)".

*Commenting on line 1617 because `docs/api/endpoints.md:1` is not an addressable diff line.*

fluree-db-server/src/routes/bm25.rs
```diff
@@ -0,0 +1,278 @@
...
+        Ok(Json(response))
+    }
+    .instrument(span)
+    .await
+}
+
+/// Read the `t` query parameter, if present.
+///
+/// Deliberately surfaces a deserialization failure as a 400 instead of
+/// following the `.ok().unwrap_or_default()` idiom used elsewhere for optional
```
aaj3f, line 255: **`fluree-db-server/src/routes/bm25.rs:255-278`** — the `sync_target_t` doc comment is the kind I wish more of this codebase had: it names the idiom it's deliberately *not* following (`.ok().unwrap_or_default()`), says why silently discarding `t` would be worse than a 400, and then explains the `t < 1` case separately with the actual consequence (a watermark moved backwards to a value no commit can restore). Someone changing this in a year will know not to "simplify" it.

aaj3f (approved):
This is really nicely done, @zonotope. Particularly in how little new machinery it ends up having to add.

The two bugs fixed along the way are the real value here and I checked both rather than taking the description's word for it: I disabled the BM25 branch of the drop sweep and watched `drop_graph_source_deletes_bm25_snapshots` and `drop_reports_deleted_snapshots` go red, then reverted `sync_bm25_index_to` to its old bare lookup and watched both retracted-index tests go red. The snapshot leak was real and the tests genuinely regress it. I also went looking specifically for a dropped `?t=` on the forwarding paths, since that would silently turn a pinned sync into a head sync — both `peer/forward.rs` and the Raft `forward_to_leader` build the target from `path_and_query()`, so it survives.

Nothing here blocks. My two substantive notes are both on the auto-sync path and both survivable because the flag is off by default: the worker has no in-flight guard, so a commit landing during a sync starts a second concurrent sync of the same index (orphaned snapshot, possible non-monotonic watermark, self-heals next sync); and `sync_bm25_index` re-runs the entire indexing query and filters afterward, so auto-sync's cost scales with corpus size rather than with the commit — which I think deserves a sentence in the BM25 doc before someone points it at a few million documents. The third is that `build_bm25_worker` and its gating have no test — `dropped_indexes_are_not_registered` re-implements the filter rather than calling the function, so it would stay green if that filter changed.

- **Patterns/abstractions:** ✔ Extends `v1_admin_protected_writes`, `forward_write_request` / `apply_leader_forward`, `RemoteLedgerClient`, `ListEntry`, and the shared `POST /drop`; invents no parallel mechanism. Mirrors the Iceberg family's create-route-plus-generic-fallbacks division deliberately and says so in the code.
- **Performance (speed first, memory second):** ✔ No query-engine, operator-loop, or per-flake path is touched — routes, CLI, a background worker, and a drop-time sweep. `bench-compare` green. **No hot-path performance-degradation risk.** The one caveat is the opt-in worker's O(corpus)-per-sync cost noted above, which is default-off and pre-existing in `sync_bm25_index`.
- **Testing:** ✔ 4047 tests pass at HEAD with `--all-features`. All new tests confirmed running by name (15 HTTP integration, 4 auto-sync, 14 CLI unit, 7 CLI integration, 2 in `grp_graphsource` — so the `it_*` additions are properly wired into a `grp_*` binary). Two mutation checks confirmed the fixes are genuinely regressed. ⚠️ only for the untested `build_bm25_worker` gating.

Approving so you can merge when ready — but I do think the in-flight guard and the auto-sync cost note are worth either doing here or landing as a tracked follow-up before anyone turns `--bm25-auto-sync` on in anger.

---

**Notes without a diff anchor:**

**`fluree-db-api/src/admin.rs:1328`** — *optional.* The `?` on `load_or_create_bm25_manifest` propagates out of `drop_graph_source` before the retract at `:1339`. That means a BM25 index whose manifest blob is unreadable — storage error, or a CAS blob already removed — can no longer be dropped at all through `POST /drop`, where before this change the drop always retracted. Drop is usually the recovery action for a wedged graph source, so failing it closed is the awkward direction. This matches what `drop_full_text_index` already did, so it isn't a new pattern, but the engine's general habit is to degrade to the correct slower path rather than hard-fail on an edge case — pushing the error into `report.warnings` and continuing to the retract would keep the sweep best-effort (which `delete_bm25_snapshots` already is, per its own doc comment) and keep drop always droppable. I don't feel strongly, and it's entirely reasonable to make an issue and follow up.

**Commit messages** — *nit, and this one belongs in the body rather than on a line.* The 15 commits are single-line subjects with no bodies and no conventional-commit scope: `cleanup`, `update docs`, `wire up remote server back bm25 management in the cli`. The repo norm is `feat(server):` / `fix(query):` style with multi-line bodies carrying the rationale. The rationale plainly exists — the PR description is one of the more thorough ones I've read, and the code comments are genuinely excellent — it just never made it into the commits, so `git log` won't carry any of it. Not worth a rebase on my account, but flagging since PR titles feed the auto-generated release notes and the commit trail is what survives.

PR #1586, opened 2026-08-03
fix(api): materialize sink call sites — finish() → into_parts() (#1529 × #1552 merge skew)

aaj3f:
Main currently fails `cargo check -p fluree-db-api --all-features --all-targets` with two E0308s in `fluree-db-api/src/materialize.rs` (:327, :869). Cause: cross-PR semantic skew — #1529 (merged 18:12) added two `sink.finish()` call sites destructuring the spool tuple; #1552 (merged 19:40) renamed that tuple-returning method to `into_parts()` and adapted every caller visible on its branch, which predated #1529's merge. Each PR was CI-green against pre-other main; no textual conflict existed to flag it.

This is the same two-line mechanical rename #1552 applied to its own callers, nothing more. The two similarly-named `writer.finish()` calls at :1442/:1757 are a different type (the verify spool writer) and are deliberately untouched.

Verified at head: `cargo check -p fluree-db-api --all-features --all-targets` clean (the reproducing command), materialize lib tests 20/20.

**Update (third skew site):** `#1552` also made `emit_triple` fallible; #1529's `InterningObserver` discarded the new `Result` (an encode error would have vanished silently — and it fails main's `-D warnings` clippy). `2910309a7` propagates it as an `R2rmlError` through the observer's existing `Result` return, matching the file's error-wrap idiom. Note main's own push CI at the #1552 merge is red across `test`/`clippy`/`testsuite-sparql` from these same skew sites — this PR should restore it.

### DIFF
--- a/fluree-db-api/src/materialize.rs
+++ b/fluree-db-api/src/materialize.rs
@@ -866,8 +866,9 @@ fn spawn_produce_workers(
 
                         // Finish + ship this chunk (always non-empty: it holds
                         // `first`). Mirror build_stamp_chunk's finalize.
-                        let (writer, prefix_map, spool_ctx) =
-                            sink.into_parts().map_err(|e| format!("flake encode: {e}"))?;
+                        let (writer, prefix_map, spool_ctx) = sink
+                            .into_parts()
+                            .map_err(|e| format!("flake encode: {e}"))?;
                         let op_count = writer.op_count();
                         let new_codes = worker_cache.into_new_codes();
                         let spool_result = spool_ctx.map(SpoolContext::finish_buffered);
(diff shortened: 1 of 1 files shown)


aaj3f:
Independent corroboration: a second session diagnosed the same break in parallel and derived the same three call sites before discovering this PR (see closed #1587). Root cause verified from the CI history — parent dcfd7838c green, merge commit eadc31e11 red — and from a clean worktree at origin/main with no other changes: #1552 renamed the inherent ImportSink::finish() → into_parts() to free the name for GraphSink::finish() (returns unit), #1529 landed materialize.rs calling sink.finish() during that review, and each PR's CI ran green against its own pre-merge base — the merge commit is the first tree where the two coexist, so no per-PR CI could have caught it. Equivalent edits verified: cargo check --workspace --all-targets clean, clippy -p fluree-db-api --all-features --all-targets -- -D warnings clean, materialize-scoped tests 22/22. The propagation form on the emit_triple site is the right call — ImportSink defers encode errors (emit_* stashes the first failure; into_parts() surfaces it), so the ? cannot fire today and stays sound if that contract ever changes. Every open PR in the repo inherits main's red until this lands.

zonotope (approved):
🪶

PR #1587, opened 2026-08-03
fix(api): repair the #1529 x #1552 merge seam — main is red

aaj3f:
main is red at eadc31e11 (the #1552 merge): two E0308s in `fluree-db-api/src/materialize.rs`, plus an `unused_must_use` behind CI's `-D warnings`. This is a three-line repair.

## Why neither PR's CI could see it

This is a semantic seam between two PRs that were each green on their own base:

- #1552 renamed the inherent `ImportSink::finish()` → `into_parts()` — deliberately, to free the `finish` name for the new `GraphSink::finish()` trait method, which returns `()`. Its doc comment says so.
- #1529 (materialize-builder) landed on main while #1552 was in review, adding `materialize.rs` with two `sink.finish()` calls that destructure the 3-tuple.

Merged together, `sink.finish()` resolves to the trait method returning `()`, and both destructurings fail to typecheck. Each PR's CI ran against its own pre-merge base; the merge commit is the first place the two changes coexist, and main's CI went red exactly there (parent `dcfd7838c` = success, `eadc31e11` = failure).

## The fix

- Both `sink.finish()` call sites → `sink.into_parts()` (the renamed inherent method; same 3-tuple, same error).
- `self.sink.emit_triple(s, p, o)` → `let _ = self.sink.emit_triple(s, p, o)` with a comment stating the contract: `ImportSink` uses deferred-error semantics — `emit_*` stashes the first encode failure and `into_parts()` surfaces it as a hard error — so the discard is deliberate, not lossy. If the materialize owner prefers a per-triple failure site, propagating instead is also sound; happy to follow up.

Verified: `cargo check --workspace --all-targets` clean at this branch, `cargo clippy -p fluree-db-api --all-features --all-targets -- -D warnings` clean, `fluree-db-api` test suite green (counts in the checks below).

Every open PR is currently inheriting this red; nothing else in the repo touches these lines, so this should be safe to land quickly.

### DIFF
--- a/fluree-db-api/src/materialize.rs
+++ b/fluree-db-api/src/materialize.rs
@@ -246,7 +246,11 @@ impl TripleObserver for InterningObserver<'_, '_> {
         let s = intern_term(self.sink, subject);
         let p = self.sink.term_iri(predicate);
         let o = intern_term(self.sink, object);
-        self.sink.emit_triple(s, p, o);
+        // ImportSink uses deferred-error semantics: emit_* stashes the first
+        // encode failure and `into_parts()` surfaces it as a hard error, so
+        // discarding this Result is deliberate, not lossy. Propagating here
+        // instead would also be sound if a per-triple failure site is wanted.
+        let _ = self.sink.emit_triple(s, p, o);
         *self.bytes += triple_weight(subject, predicate, object);
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


aaj3f:
Closing as a duplicate of #1586, which was opened first and takes the better form on the emit_triple site (real propagation rather than a documented discard — behaviorally identical today since ImportSink's emit_triple unconditionally returns Ok and defers encode errors to into_parts(), but propagation is future-proof if that contract ever changes). Two sessions independently diagnosed the same three sites, which is corroboration for #1586, not a conflict. The seam analysis in this PR's commit message and body applies to #1586 verbatim.

PR #1588, opened 2026-08-03
[Fixes Main] Follow the fallible GraphSink protocol

zonotope:
bf7c0a6f7 made `GraphSink` fallible and renamed `ImportSink::finish` to `into_parts`, updating every caller inside import_sink.rs. materialize.rs landed on a parallel branch and kept the old calls, so main has not compiled since the two merged. This patch updates materialize.rs to follow he new fallible `GraphSinc` protocol.


aaj3f:
I think this may duplicate #1586 -- sorry @zonotope I meant to make a comment in Slack about the PR I already opened for this

PR #1593, opened 2026-08-05
fix: reject SPARQL-style and constant JSON-LD filter expressions (fail-open no-op)

bplatz:
## Problem

A JSON-LD `["filter", ...]` expression spelled the way every SPARQL user tries first —

```json
["filter", "contains(lcase(?name), \"zzz_no_match\")"]
```

— was accepted without error and filtered nothing. `parse_s_expression`'s atom fallback turned the whole string into a plain-string constant, and expression eval treats non-boolean constants as truthy, so the filter matched every row. A filter that silently becomes a no-op is fail-open: callers using it to narrow results (including authorization-adjacent narrowing) get strictly more data than they asked for, with no signal. Bare-word or numeric constant filters (`["filter", "gustavo"]`) had the same trap.

The parenthesized variant `(contains(lcase(?name), ...))` mis-tokenized into `?name` in operator position and failed later with the misleading `Unknown function: ?name`.

## Fix (fail closed)

- `parse_s_expression` rejects SPARQL-style `ident(...)` spellings with an error pointing at the S-expression syntax: `'contains(...)' looks like SPARQL function-call syntax ... use S-expression syntax instead, e.g. (contains (lcase ?name) \"fred\")`. Quoted strings never match, so BIND can still express call-shaped string literals.
- A variable in operator position now errors with a prefix-syntax diagnosis instead of `Unknown function: ?name`.
- FILTER and HAVING sites reject expressions that parse to a bare non-boolean constant ("constant and would match every row"). Boolean constants, bare variables (`["filter", "?flag"]`), and composite constant expressions like `(= 1 1)` (W3C #1439 semantics) are unchanged; BIND/UNWIND are value contexts and keep accepting constants.

## Scope notes

- `datalog_rules.rs` has its own strict mini-parser and already rejected these spellings; SPARQL surface unaffected.
- The reporter's adjacent observation (bare-variable `select` returning empty values over HTTP) does not reproduce at the API layer — tracked separately if it resurfaces.

## Testing

- Parser unit tests in `filter_sexpr.rs` for the rejected spellings and the quoted-literal escape hatch.
- End-to-end regression `jsonld_sparql_style_filter_spelling_errors` in `it_query_expression_semantics.rs` covering all three reported variants plus the working S-expression control.
- Docs: filter syntax rule added to `docs/query/jsonld-query.md`.


fluree-db-query/src/parse/filter_sexpr.rs
```diff
@@ -104,6 +121,27 @@ pub fn parse_s_expression(s: &str) -> Result<UnresolvedExpression> {
...
+            .all(|c| c.is_alphanumeric() || matches!(c, '_' | ':' | '.'));
+    is_ident.then_some(callee)
+}
+
+fn sparql_call_syntax_error(callee: &str) -> ParseError {
+    ParseError::InvalidFilter(format!(
+        "'{callee}(...)' looks like SPARQL function-call syntax, which is not supported \
+         in JSON-LD expressions; use S-expression syntax instead, \
+         e.g. (contains (lcase ?name) \"fred\")"
+    ))
```
zonotope, line 149: I think we should add language to the error message about requiring that literal strings are quoted. `sparql_style_callee` might also be evaluated in the context of a `bind` form where the user actually meant a string like "Acme(Inc)". In that case, the `sparql_style_callee` check would still trigger this err and the user would get a confusing message.

fluree-db-query/src/parse/filter_sexpr.rs
```diff
@@ -104,6 +121,27 @@ pub fn parse_s_expression(s: &str) -> Result<UnresolvedExpression> {
...
+/// If `s` is spelled like a SPARQL function call — an identifier immediately
+/// followed by `(` — return the identifier. Quoted strings never match, so
+/// `"\"contains(x)\""` remains expressible as a string literal.
+fn sparql_style_callee(s: &str) -> Option<&str> {
+    let open = s.find('(')?;
+    let callee = &s[..open];
+    let is_ident = !callee.is_empty()
+        && callee
+            .chars()
+            .all(|c| c.is_alphanumeric() || matches!(c, '_' | ':' | '.'));
```
zonotope: this is missing '-', so names like `is-blank`, `is-iri`, ..., from `lower_function_name` (lower.rs#1658-1777) slip past.

fluree-db-query/src/parse/options.rs
zonotope: This function is missing the guard and public. It doesn't really matter because it's dead code, but I think that suggests that we should just delete it along with its unit tests instead of adding the guard here too.

zonotope (approved):
🚗

fluree-db-query/src/parse/filter_sexpr.rs
bplatz: Addressed in 14c3b58d46be0574c0211b39eb0a4b63a637ea2e

fluree-db-query/src/parse/filter_sexpr.rs
bplatz: Addressed in 14c3b58d46be0574c0211b39eb0a4b63a637ea2e

fluree-db-query/src/parse/options.rs
bplatz: Addressed in 14c3b58d46be0574c0211b39eb0a4b63a637ea2e

PR #1597, opened 2026-08-06
fix: stop hiding user f:-namespace data from variable-predicate scans

bplatz:
> Stacked on #1593 — contains that branch's commit; diff shown here is just this change.

## Problem

A `?s ?p ?o` scan of a stored `f:AccessPolicy` node returned only its `rdf:type` triple — `f:action`, `f:allow`, `f:exMessage` were invisible, while explicit-predicate lookups of the same triples returned them. Variable-predicate scans hid the **entire** `https://ns.flur.ee/db#` namespace in the default graph.

That hide is a fossil from before named graphs existed, when commit metadata was stored in the main graph. Today commit metadata routes to the txn-meta graph (`g_id=1`) on every path — novelty-side stamping (`stamp_graph_on_commit_flakes`, all three call sites) and both indexer emitters hard-code `g_id=1` — and an empirical dump of a current ledger's default graph with `includeSystemFacts: true` contains zero system-written `f:` triples. The only data the hide still caught was user-authored `f:` vocabulary, policies being the prime case: a wildcard inspection of a policy write looks like the write silently dropped everything but `@type`, and an export-by-crawl silently loses policy definitions.

Worse, the hide only ran on the indexed scan path — the novelty path never applied it — so the same query returned all triples right after the write and dropped them once the background indexer caught up.

## Fix

Narrow the filter to the seven `f:reifies*` predicates, hidden in **every** graph as before. Those stay hidden deliberately: they are the system-written storage encoding of edge annotations (user transactions cannot assert them, so surfacing them would break export/re-import round-trips), they are redundant with the edge and annotation content already visible in results, and they live in the same graph as the edge they reify so there is no system graph to banish them to. `opts.includeSystemFacts: true` still reveals them for inspection.

Applied at all three mirrors so scan and fast-path semantics agree:
- `BinaryScanOperator::is_internal_predicate` (`binary_scan.rs`)
- the whole-graph fold's overlay-flake walk (`fast_whole_graph_agg.rs`)
- the fold's exactness preflight `graph_has_scan_hidden_predicates` (`fast_whole_graph_agg.rs`)

Other `FLUREE_DB` checks in the query crate (planner selectivity, annotation probes, wildcard hydration, fulltext datatype) were audited and are reifies-specific or unrelated — untouched.

## Compatibility note

A ledger whose index predates txn-meta routing and has never been rebuilt could re-expose old commit metadata in default-graph wildcard dumps; reindexing re-emits commit metadata into the txn-meta graph.

## Testing

- New `it_query_wildcard_system_facts.rs`: user `f:` data visible to wildcards **pre-index and post-index** (pinning the novelty/indexed flip), and a full default-graph dump stays free of commit-metadata predicates.
- Existing `it_edge_annotations.rs` reifies-hide and `includeSystemFacts` tests pass unchanged.
- Full `fluree-db-api` suite green (30 targets, 3,041 tests); `fluree-db-query` full tests green; clippy `--all-features` clean on both crates.
- Docs: variable-predicate visibility rule, the `f:reifies*` exception, and `includeSystemFacts` documented in `docs/query/jsonld-query.md` (the flag was previously undocumented).

### DIFF
--- a/.fluree-memory/repo.ttl
+++ b/.fluree-memory/repo.ttl
@@ -2360,6 +2360,21 @@ mem:constraint-01kyke812e1np5c9ag45ac7580 a mem:Constraint ;
     mem:createdAt "2026-07-28T04:03:09.006298+00:00"^^xsd:dateTime ;
     mem:rationale "Using required_aggregate_vars (rather than checking the projection directly) folds in projection, HAVING, post-binds and ORDER BY in one check — a post-bind whose EXPRESSION reads a non-key var still reads a grouped list, which a bind-output whitelist would miss. Regression tests need a fan-in fixture; the pinned grouped-list test has none and cannot catch dedup." .
 
+mem:fact-01kz9zn33sjc3rh7r9s2zr825w a mem:Fact ;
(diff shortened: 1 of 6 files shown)


zonotope (approved):
🖲️

PR #1598, opened 2026-08-06
fix: stop mapping one VMA per dictionary pack (ENOMEM with memory free)

christophediprima:
## Summary

A process is capped at `vm.max_map_count` memory mappings — **65,530** by default on Linux. Past that `mmap` returns `ENOMEM` **regardless of how much memory is free**, which surfaces as:

```
failed to load binary index: Cannot allocate memory (os error 12)
```

`ForwardPackReader::from_pack_refs` maps **every pack in a dictionary's routing table**, eagerly, one mapping each, held for the reader's lifetime, with no bound and no eviction. Dict packs are per-ID-range and never compacted, so the routing table grows without limit; loading many ledgers in one process multiplies it.

The fix is a size split: `read()` packs at or below 64 KiB onto the heap, keep mapping the larger ones. Behaviour is otherwise unchanged.

## Why this is worth your time

Measured on a deployment running 23 ledgers under continuous ingest:

```
dict packs on disk                       103,441
live mappings held by the server          47,336   (72% of the 65,530 cap, climbing)
  of which @shared/dicts/*.dict           47,336   (99.8% of all mappings)
largest single ledger                     19,629   packs in one routing table
process memory at the time of ENOMEM     1.6 GiB of a 24 GiB limit, memory.events:oom_kill = 0
```

**Failing to allocate at 6% of the memory limit is the part that makes this expensive to diagnose.** Every instinct says "memory pressure", and every memory metric says you have plenty. We raised the container limit from 16 to 24 GiB, which of course changed nothing, and spent a day on the wrong resource. The error string contains the word "memory" and the cause has nothing to do with memory.

## Why a size split is the right shape

Pack sizes are extremely skewed. Across those 103,441 packs:

| bucket | packs | share |
|---|---|---|
| < 4 KiB | 94,415 | **91.3%** |
| < 64 KiB | 96,637 | 93.4% |
| ≥ 1 MiB | 5,728 | 5.5% (holding 12.1 of the 12.5 GiB) |

The median pack is **113 bytes**. Of the packs actually mapped at the time of measurement, **99.8% were under 64 KiB and held 30 MB between them**.

Mapping a 113-byte file spends a VMA and a page of address space to expose less than a cache line of useful data, and buys nothing a `read()` does not: demand paging cannot help a file smaller than one page, and the kernel-side cost of the mapping exceeds the data. So the trade is heavily one-sided.

On that deployment's data, the worst case — every pack on disk loaded at once — becomes:

```
before   103,441 mappings
after      6,793 mappings  +  47 MiB of heap
```

15x headroom against the cap, for 47 MiB out of gigabytes. The large packs that genuinely justify demand paging are still mapped.

## What changed

- **`load_pack_backing`** replaces `mmap_file` as the single place that opens a pack, and decides read-vs-mmap by size. `LoadedBacking::InMemory` already existed for the in-memory constructor, so there is no new representation.
- **`LazyLoaded` now holds a `LoadedBacking`** rather than a bare `Mmap`, so the lazy path gets the same treatment as the eager one. Previously only the eager path could have been fixed in isolation.
- **The remote-fetch path still writes to cache and re-opens** rather than keeping the `Vec` it already has. That looks like a wasted read, and it is deliberate: it keeps `load_pack_backing` the only place the threshold is applied, so a large pack still avoids heap duplication and a small one still lands on the heap, without a second copy of the policy that can drift.
- **A failed `metadata()` falls through to mmap.** The threshold is an optimisation; losing the size must not lose the read.
- **`FLUREE_DICT_PACK_MMAP_MIN_BYTES`** overrides the threshold. `0` restores the previous always-mmap behaviour exactly, which makes this trivially revertible in a deployment without a rollback.

## Tests

Three tests, and they assert on the **backing variant**, not on lookups:

| test | pins |
|---|---|
| `small_packs_are_read_not_mapped` | a pack under the threshold yields `InMemory`, and the bytes survive |
| `large_packs_are_still_mapped` | over the threshold still yields `Mmap` — this is a split, not a move to heap reads |
| `threshold_boundary_is_inclusive` | exactly-at-threshold is read, so an off-by-one cannot silently move the most common pack size across the boundary |

Asserting on the variant rather than on lookup results is the whole point. **Lookups pass either way** — which is precisely why a mapping-per-pack survived from the v4 baseline without anyone noticing. The variant is the only assertion that fails if someone reverts to always-mmap.

`fluree-db-binary-index`: **345 passed**. `fmt` clean; `clippy --all-features --all-targets` clean.

**Heads-up on an unrelated failure that is already on `main`:** `read::binary_index_store::tests::open_leaf_handle_caches_remote_metadata` fails on **pristine `upstream/main`** with no changes from this PR present — verified by reverting `pack_reader.rs` to the `upstream/main` version on this very branch and re-running (342 tests filtered, so none of the three tests above were even compiled in). It is in a file this PR does not touch. Flagging it rather than letting it look like fallout from this change; happy to open a separate issue if useful.

## Notes for review

- **64 KiB is a judgement call from one deployment's distribution.** We have no attachment to it. The argument for "at least a few pages" is that below one page a mapping cannot pay for itself; the argument against a much larger value is heap growth. If you prefer a different default, or prefer the threshold opt-in rather than on by default, both are one-line changes.
- **This bounds mappings, not the pack count.** The underlying growth — dict packs are per-ID-range and never compacted — is untouched, and a deployment large enough will eventually reach 65,530 packs of ≥64 KiB. Compaction of dictionary packs would be the real fix for that and is out of scope here.
- **A bounded mmap cache with eviction would be the more general answer** (the leaflet cache already does exactly this via `try_get_or_load_leaf_mmap`). We went with the size split because it is smaller, has no eviction policy to tune, and removes 99.8% of the mappings on real data. If you would rather see the packs go through a bounded cache instead, say so and we will do that version.

### DIFF
--- a/fluree-db-binary-index/src/dict/pack_reader.rs
+++ b/fluree-db-binary-index/src/dict/pack_reader.rs
@@ -1,4 +1,4 @@
-//! Multi-pack reader with mmap support for forward dictionary packs.
+//! Multi-pack reader for forward dictionary packs.
 //!
 //! `ForwardPackReader` manages one or more `FPK1` packs and routes lookups
 //! to the correct pack via binary search on ID ranges.
@@ -56,7 +56,7 @@ enum LoadedBacking {
 
 struct LazyLoaded {
     meta: ParsedPackMeta,
-    mmap: memmap2::Mmap,
+    backing: LoadedBacking,
 }
 
 impl LoadedBacking {
@@ -70,7 +70,7 @@ impl LoadedBacking {
 
 impl PackHandle {
     /// Get pre-parsed metadata and raw bytes. For lazy packs, triggers
-    /// fetch + cache + mmap + parse on first call (subsequent calls return cached).
+    /// fetch + cache + load + parse on first call (subsequent calls return cached).
     fn ensure_loaded(&self, ctx: Option<&LoadContext>) -> io::Result<(&ParsedPackMeta, &[u8])> {
         match &self.inner {
             PackInner::Loaded { meta, backing } => Ok((meta, backing.bytes())),
@@ -83,7 +83,7 @@ impl PackHandle {
                 let lazy = loaded.get_or_try_init(|| {
                     fetch_and_load(self.first_id, self.last_id, pack_cid, cache_path, ctx)
                 })?;
-                Ok((&lazy.meta, lazy.mmap.as_ref()))
+                Ok((&lazy.meta, lazy.backing.bytes()))
             }
         }
     }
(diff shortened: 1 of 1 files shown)


bplatz (approved):
Great, thanks for addition, referenced fwd pack compaction doc which we'll need to do to help manage this going forward.

PR #1599, opened 2026-08-06
fix(storage): treat a zero-length blob as absent, not as empty content

christophediprima:
**Rescoped after #1612.** This PR originally carried two fixes: atomic blob writes, and treating a zero-length blob as absent. #1612 landed the first one, and landed it better than this did — staging plus rename for `write_bytes`, `hard_link` for the CAS path, a configurable `Durability` with an observable fsync counter, staging cleanup on every error path, and `list_prefix` skipping staging artifacts. I have dropped that half entirely and rebuilt the branch on current `main`. What remains is the read side.

## What is left, and why it still matters after #1612

`FileStorage` decides a blob is present by asking whether the path exists. Three read paths do that, and none of them look at the size:

- `read_bytes` maps only `NotFound`, so an empty file returns `Ok(vec![])`;
- `resolve_local_path` returns any path where `path.exists()` — and callers mmap or parse that path directly;
- `exists` answers from `metadata` regardless of length.

**A content-addressed store can prove an empty blob is not content.** The address commits to a digest, and no real artifact hashes to empty. So a zero-length file at such an address is not data — it is debris from a write that created the file and never filled it.

I want to be precise about what #1612 already fixes here, because it changes the argument. `stage_bytes` writes to a temp file and removes it on failure, so an ENOSPC during a write no longer leaves a zero-length file at the final path. Going forward this state is much harder to create through this API. What #1612 does not do is anything about:

- **debris already on disk.** One deployment of ours accumulated **~4,050 zero-length blobs** under ENOSPC, 3,981 of them commit blobs. Upgrading to #1612 does not remove them, and every read of one still fails.
- **emptiness introduced from outside this process** — a restore from a bad backup, an external tool, a filesystem that surfaces a truncated file after an unclean shutdown.

So this is no longer "the fix for the ENOSPC bug" — #1612 is. It is the read side declining to infer validity from presence, which is an assumption worth removing on its own.

## Why "absent" rather than "empty"

The two answers are not equally recoverable, and that asymmetry is the whole point:

- **absent** → the caller re-fetches or rebuilds, and the deployment heals itself;
- **empty content** → a parse failure surfaces at a distant call site that no caller knows how to repair.

Concretely, those 4,050 files produced `pack header: need 40 bytes, got 0` and kept a ledger broken for days after the disk itself had been fixed. Finding the bad file by hand meant decoding a pack header and matching its length against 55,007 candidate blobs. That should not be the recovery procedure for a byte count.

## The change

All three read paths agree that zero length means absent:

- `read_bytes` returns `NotFound` and `warn!`s with the address and path, naming it as failed-write debris and saying it will be re-fetched.
- `resolve_local_path` returns `None`, so nothing hands an empty file to an mmap or a parser.
- `exists` returns `false`.

**`exists` is in the list deliberately.** Leaving it out would be worse than doing nothing: a caller that checks before reading would see the blob present and then have the read refuse it. Answering `false` also lets a writer replace the debris rather than skip it as already-present, which is how the bad file finally leaves the disk.

The guard is `len() > 0`, not a content check — cheap, and it cannot reject anything a content-addressed store could legitimately hold.

## Tests

Four, all against a zero-length file planted directly on disk, which is the only way the state arises now that #1612 made writes atomic — and which is exactly the pre-existing-debris case this targets:

- `zero_length_blob_reads_as_absent` — `read_bytes` returns `NotFound`, not `Ok(vec![])`.
- `resolve_local_path_rejects_a_zero_length_blob` — presence is not validity.
- `exists_agrees_with_read_bytes_on_a_zero_length_blob` — the consistency property, plus proof that a normal write over the debris restores service.
- `a_one_byte_blob_is_still_present` — the guard does not fire on legitimate content; one byte is the smallest thing genuinely there.

### DIFF
--- a/fluree-db-core/src/storage/file.rs
+++ b/fluree-db-core/src/storage/file.rs
@@ -406,6 +406,28 @@ impl StorageRead for FileStorage {
                     crate::error::Error::io(format!("Failed to open {}: {}", path.display(), e))
                 }
             })?;
+            // The fourth read path, held to the same rule as the other three:
+            // an empty file at a content address is debris, not content. A
+            // ranged read would otherwise stop at EOF and hand back an empty
+            // buffer — the "empty content" answer this whole change exists to
+            // replace with "absent". This arm is not hypothetical: once
+            // `resolve_local_path` refuses the debris, the leaflet reader
+            // falls through to `ContentStore::get_range`, which lands here for
+            // the very same file. Read off the open handle, so there is no
+            // extra stat and no window between the check and the read.
+            if file.metadata().map(|m| m.len() == 0).unwrap_or(false) {
+                tracing::warn!(
+                    address,
+                    path = %path.display(),
+                    "zero-length blob treated as absent on a ranged read (failed write \
+                     debris); it will be re-fetched or rebuilt. Delete it to reclaim the inode."
+                );
+                return Err(crate::error::Error::not_found(format!(
+                    "{}: {} (zero-length blob, treated as absent)",
+                    address,
+                    path.display()
+                )));
+            }
             #[cfg(unix)]
             {
                 use std::os::unix::fs::FileExt;
@@ -1211,6 +1233,40 @@ mod tests {
         assert_eq!(storage.read_bytes("z/empty.dict").await.unwrap(), b"real");
     }
 
+    /// A ranged read must agree with `read_bytes` on the same blob. Without the
+    /// guard the read stops at EOF and returns `Ok([])` — empty content, the
+    /// answer no caller can heal — where `read_bytes` says absent.
+    ///
+    /// Reached in practice *because of* `resolve_local_path`: the leaflet
+    /// reader tries the local path first, that guard refuses the debris, and it
+    /// falls through to `ContentStore::get_range`, which reads the same file
+    /// through here.
+    #[tokio::test]
+    async fn zero_length_blob_reads_as_absent_through_a_ranged_read() {
+        let (_dir, storage) = storage();
+        let address = "z/ranged.dict";
+        plant_zero_length(&storage, address);
+
+        let err = storage
+            .read_byte_range(address, 0..40)
+            .await
+            .expect_err("a zero-length blob must be absent on a ranged read, not empty content");
+        assert!(
+            matches!(err, crate::error::Error::NotFound(_)),
+            "expected NotFound, got {err:?}"
+        );
+
+        // The two read surfaces must not disagree about the same blob.
+        assert!(storage.read_bytes(address).await.is_err());
+
+        // And a real write over the debris restores both.
+        storage.write_bytes(address, b"real").await.unwrap();
+        assert_eq!(
+            storage.read_byte_range(address, 0..4).await.unwrap(),
+            b"real"
+        );
+    }
+
     /// The guard must not fire on legitimate content. A one-byte blob is the
     /// smallest thing that is genuinely there.
     #[tokio::test]
@@ -1221,5 +1277,10 @@ mod tests {
         assert!(storage.exists("z/tiny.dict").await.unwrap());
         assert!(storage.resolve_local_path("z/tiny.dict").is_some());
         assert_eq!(storage.read_bytes("z/tiny.dict").await.unwrap(), b"x");
+        // The ranged path agrees that one byte is present.
+        assert_eq!(
+            storage.read_byte_range("z/tiny.dict", 0..1).await.unwrap(),
+            b"x"
+        );
     }
 }
--- a/fluree-db-core/src/storage/file.rs
+++ b/fluree-db-core/src/storage/file.rs
@@ -333,21 +333,59 @@ impl FileStorage {
 impl StorageRead for FileStorage {
     async fn read_bytes(&self, address: &str) -> Result<Vec<u8>> {
         let path = self.resolve_path(address)?;
-        tokio::fs::read(&path).await.map_err(|e| {
+        let bytes = tokio::fs::read(&path).await.map_err(|e| {
             if e.kind() == std::io::ErrorKind::NotFound {
                 crate::error::Error::not_found(format!("{}: {}", address, path.display()))
             } else {
                 crate::error::Error::io(format!("Failed to read {}: {}", path.display(), e))
             }
-        })
+        })?;
+        // A ZERO-LENGTH blob is not content — it is debris, and reporting it as
+        // absent is strictly better than returning it.
+        //
+        // Blobs here are content-addressed, so the address commits to a digest and
+        // no real artifact hashes to empty. An empty file at such an address can
+        // therefore only be a failed write (create succeeded, write did not — the
+        // classic ENOSPC shape, which left ~4,000 of these on one deployment).
+        //
+        // The distinction matters because the two outcomes are not equally
+        // recoverable: "absent" makes callers re-fetch or rebuild, while empty
+        // content propagates as a parse failure at some distant call site
+        // ("pack header: need 40 bytes, got 0") that no caller knows how to repair.
+        if bytes.is_empty() {
+            tracing::warn!(
+                address,
+                path = %path.display(),
+                "zero-length blob treated as absent (failed write debris); it will be \
+                 re-fetched or rebuilt. Delete it to reclaim the inode."
+            );
+            return Err(crate::error::Error::not_found(format!(
+                "{}: {} (zero-length blob, treated as absent)",
+                address,
+                path.display()
+            )));
+        }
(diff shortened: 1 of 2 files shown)


fluree-db-core/src/storage/file.rs
```diff
@@ -425,7 +463,13 @@ impl StorageRead for FileStorage {
...
         let path = self.resolve_path(address)?;
         match tokio::fs::metadata(&path).await {
-            Ok(_) => Ok(true),
+            // Zero length is absent here too, and the consistency is the point:
+            // reporting `true` for a blob `read_bytes` then refuses to return is a
+            // worse contract than either answer alone — a caller that checks before
+            // reading would see the blob appear and then vanish. Answering `false`
+            // also lets a writer replace the debris instead of skipping it as
+            // already-present, which is how the bad file finally leaves the disk.
+            Ok(m) => Ok(m.len() > 0),
```
aaj3f, line 494: **`fluree-db-core/src/storage/file.rs:472`** — *praise.* `exists()` answering `m.len() > 0` closes the check-then-read divergence that was the sharpest edge here — `has()` calling debris "present" while `read_bytes` calls it absent would have made the self-healing story false — and the comment gives both reasons (consistency, and `false` letting a writer replace the debris). I verified the commit's "each guard killed by exactly one test" claim for all three guards: reverting this arm fails exactly `exists_agrees_with_read_bytes_on_a_zero_length_blob`, deleting the `read_bytes` guard fails exactly `zero_length_blob_reads_as_absent`, loosening `resolve_local_path` fails exactly its test. That's an unusually honest test-to-guard mapping.

fluree-db-core/src/storage/file.rs
```diff
@@ -333,21 +333,59 @@ impl FileStorage {
...
-            Some(path)
-        } else {
-            None
+        // PRESENCE IS NOT VALIDITY. This returned any path that merely `exists()`,
+        // and callers then mmap or parse it directly — so a zero-length blob became
+        // an unrecoverable reader error rather than a miss the caller could heal.
+        // Excluding empty files here is what converts that poison back into a fetch.
+        // See `read_bytes` for why empty can never be legitimate content.
+        match std::fs::metadata(&path) {
+            Ok(m) if m.len() > 0 => Some(path),
```
aaj3f, line 379: **`fluree-db-core/src/storage/file.rs:379`** — *optional, carried over.* The "(zero-length blob, treated as absent)" explanation still doesn't survive the content-addressing bridge — `StorageContentStore::get` swallows the primary `NotFound` and ends in a generic `not_found(address)` at `storage.rs:685`. The new `warn!` at the detection point mitigates it well (the diagnosis lands in logs with the path and the reclaim hint), so this is genuinely minor now. If you agree it's worth carrying the reason through the bridge error, I'd rather see it folded in here than lost in the backlog; if you think the warn is the right stopping point, no argument — commenting here because `storage.rs` isn't in this diff.

fluree-db-core/src/storage/file.rs
```diff
@@ -333,21 +333,59 @@ impl FileStorage {
 impl StorageRead for FileStorage {
```
aaj3f, line 333: **`fluree-db-core/src/storage/file.rs:330`** — *nit, carried over.* `write_bytes(addr, b"")` still lands a zero-length file the read side immediately calls absent. The system is now at least self-consistent about it (a later write replaces the debris), and nothing in the tree writes an empty blob, so this stays a nit — a cheap `debug_assert!` or early error on empty input would make the invariant explicit at the write door.

(Anchored a few lines below the real spot — `:330` isn't in this diff's hunks.)

aaj3f (approved):
This rescope is very good and I appreciate it, @christophediprima. And the remaining claim is real: the ~4,050 pre-existing zero-length blobs on a live deployment don't care how well future writes are staged. Something has to stop inferring validity from presence at read time, and that something is these three guards.

I verified rather than assumed: 26/26 storage file tests green at head (they need `--features native`, worth knowing, since a plain `--lib` run silently skips the module), and each of the three guards is killed by exactly one named test, which I proved by reverting each guard individually. The write-half mutation from my earlier read of this branch now correctly falls to **main's** #1612 tests rather than yours--the vacuity concern I had against the old revision has no referent in this one.

- **Performance (speed first, memory second):** ✔ length checks on already-fetched results; no hot-path cost.
- **Patterns/abstractions:** ✔ read-side rule lives in the file backend beside its peers; no parallel mechanism invented, and the rescope actively removed a parallel one.
- **Testing:** ✔ four new tests, each pinned to its guard by mutation; suite green at head.
- **Conventions:** ✔ the rewritten single-commit message is a model of the form — provenance, the war story (`pack header: need 40 bytes, got 0`, 55,007 candidates), and a mutation claim that survives checking.

Verified locally at branch HEAD (`8c220db3e`): `cargo test -p fluree-db-core --features native --lib file::tests` → 26/26; three per-guard mutations → three specific reds, restored green; diff scope confirmed as `file.rs` only (+128/−7).

Approving for whenever you're ready. Nothing here needs to move before merge, the two carried-over notes are yours to take or leave.

fluree-db-core/src/storage/file.rs
bplatz: Folded in — both bridges, not just `get`. `get_range` had the identical swallow a few lines down (`storage.rs:748`), so the reason was being discarded on the ranged path too.

Each now binds the primary miss's message and carries it into the final error instead of rebuilding a bare `not_found(address)`. That is strictly better even for an ordinary miss: the reconstructed error was dropping the resolved path that `read_bytes` had already put in its message.

fluree-db-core/src/storage/file.rs
bplatz: Left as a nit, deliberately — I probed it and the current behavior is already self-consistent:

```
write_bytes(b"")  -> Ok(())
  then read_bytes -> absent
  then exists     -> false
```

So an empty write is a no-op you cannot read back, rather than debris that reads as content, and a later real write replaces it. Nothing in the tree writes an empty blob. A `debug_assert!` would document the invariant without changing behavior, which did not feel worth widening the diff for — happy to add it if you disagree.

bplatz:
Pushed two follow-ups to this branch (`9eee00179`) — @aaj3f's two carried-over notes, plus one thing neither of us had spotted. @christophediprima, shout if you'd rather own any of it differently; the maintainer-edit flag made it easiest to just put it on the branch.

### `read_byte_range` was a fourth read path

The PR says all three read paths agree that zero length means absent. There is a fourth, and it did not:

```
read_byte_range(0..40) on a zero-length blob  ->  Ok([])          <- empty content
read_byte_range        on a missing file      ->  Err(NotFound)   <- absent
```

Same blob, and `read_bytes` calls it absent while the ranged read hands back an empty buffer — the exact answer this PR exists to replace.

What makes it worth fixing here rather than filing: **it is reachable because of this PR's own `resolve_local_path` guard.** In `binary_index_store.rs:2890-2920` the leaflet reader tries the local path first, the new guard correctly refuses the debris, and it falls through to `ContentStore::get_range` — which reads the very same file through `read_byte_range`. The guard deflects the poison into the one arm that still serves it. So the "presence is not validity" property is not actually established until this arm agrees.

The check reads length off the already-open handle, so no extra stat and no window between check and read. New test `zero_length_blob_reads_as_absent_through_a_ranged_read`, mutation-checked against your one-test-per-guard mapping: removing this guard fails exactly that test and leaves your three green. Also added a ranged assertion to `a_one_byte_blob_is_still_present`, so the fourth guard cannot overshoot either.

### Both bridges keep the miss reason

`storage.rs:670` and `:748` — `get_range` had the same swallow as `get`, so the reason was being discarded on the ranged path too. Both now carry the primary message through rather than rebuilding `not_found(address)`.

### Not folded in

The `write_bytes(addr, b"")` nit — probed, and the behavior is already self-consistent (write is `Ok`, read is absent, `exists` is `false`, a later real write replaces it). Reasoning in the thread.

### Filed separately: #1671

While probing the ranged path I hit a pre-existing panic — `FileStorage::read_byte_range(addr, 0..u64::MAX)` blows up on `vec![0u8; len]` with `capacity overflow`, while `ProxyStorage` supports that exact convention and clamps (`proxy_integration.rs:1823` asserts it). Latent, since no production caller passes an open-ended range to `FileStorage`, but it is a real divergence between two implementations of the same trait method. Unrelated to this PR's subject, so it is #1671 rather than more scope here — though the fix is two lines given the `metadata()` call this PR now makes in that same function.

Local: `fluree-db-core --features native --lib` 772 passed (27 in `file::tests`), `--all-features` 788, `grp_query` 422, `grp_misc` 262, clippy `-D warnings` and fmt clean.

PR #1601, opened 2026-08-06
fix: bound retained index versions by count, not only by age

christophediprima:
## Summary

`clean_garbage` applies `gc_max_old_indexes` and `gc_min_time_garbage_mins` as an **AND**, so the slower of the two always wins. Under any sustained publish rate that is always the age guard, which makes the count target a bound on nothing at all: real retention becomes *"however many versions fit inside the age window"*, which is unbounded in bytes because it scales with both publish rate and per-version size.

A time window cannot bound bytes. That is the whole defect.

This adds `hard_max_old_indexes`: past that many versions, collection proceeds **regardless of age**. It defaults to `4 × max_old_indexes`, so with the shipped default of 5 the ceiling is 21 versions (`1 + 5×4`) and nothing changes for any deployment that was already inside its target.

## Measured

On a deployment running ~17 ledgers under continuous ingest, with the shipped defaults (`max_old_indexes = 5`, `min_time_garbage_mins = 30`):

```
retained index versions            79      against a target of 5
objects/history                    14 GiB  for a ledger whose live index was 260 MiB
whole commit log                   417 MiB
                                   -> history was ~34x the data it derives from
```

The mechanism is visible in the collector's own debug output: every pass walked the chain, found the oldest versions past the count target, and `break`d because their garbage records were inside the 30-minute guard. Nothing was wrong with GC — it was doing exactly what the config said, and the config could not express "no more than N".

After deploying the ceiling, on the same cluster:

```
Past hard retention ceiling, collecting regardless of age
  t=7794 chain_position=21 hard_keep=21 age_mins=16
```

`hard_keep=21` is the derived default, and it collected a version **16 minutes old** — inside the 30-minute guard, which is the point. Over 40 minutes the ceiling fired **90 times**.

## What this does *not* fix, stated plainly

**This bounds the chain, not the disk.** We learned that the expensive way: after deploying it, GC was provably working — `chain_len` pinned at 22, ceiling firing 90 times in 40 minutes — and the volume kept climbing, because the same ledgers held 370, 169 and 125 root files on disk. Artifacts that have fallen off the chain are outside anything a retention setting can reach.

Part of that gap was a separate bug — full rebuilds published roots with no `prev_index`, orphaning the history in one stroke, fixed in `fix/rebuild-gc-chain-link`.

**But the more useful correction is that the rest was never garbage at all, and this is the number that makes the case for a COUNT bound.** With the chain fixed, a later measurement on the same deployment found two ledgers holding 131 GiB and 43 GiB of `objects/history` with only 17 and 13 retained index roots — and the orphan sweep correctly reported `candidates=8, deleted=0`, because every one of those artifacts was reachable from a retained root. Nothing was collectable. The retention policy was simply asking to keep more than the disk held:

```
17 roots,  91 history sidecars @ 1932 MiB  ->  131 GiB   (~7.7 GiB per retained version)
13 roots,  56 history sidecars @  830 MiB  ->   43 GiB   (~3.3 GiB per retained version)
```

That is the argument for this PR, sharper than the original framing. **A time-based guard cannot price a retained version, because the price is per-version bytes and varies by three orders of magnitude across ledgers.** Only a count bound lets an operator reason about disk at all. Setting `gc_max_old_indexes = 2` (ceiling `1 + 2×4 = 9`) took the first ledger from 17 retained versions to 6 and freed **~100 GiB** — and it freed it because the count ceiling overrode an age guard that was otherwise holding every version.

One caveat worth stating for reviewers: **the count target is not sufficient on its own either.** GC must write in order to free anything, so a volume already at 0 bytes cannot collect no matter how tight the target is. On that deployment we had to free ~3 GiB of regenerable cache by hand before GC could run at all — after which it reclaimed 100 GiB unaided. A future improvement would be reserving a small write budget so GC can always make progress; that is out of scope here.

## The commits

**1. `fix(indexer): bound retained index versions by count, not only by age`** — the ceiling. Past the hard ceiling, collect regardless of age; otherwise break on a too-recent record as before. The retention target itself is never at risk, because the loop only ever visits indexes at or beyond `keep_count = 1 + max_old_indexes`. *(As originally posted this paragraph also claimed `hard_max_old_indexes` was clamped up to `max_old_indexes` to protect the target. That clamp turned out to be unobservable and is removed in commit 4 — see the follow-up comment.)*

**2. `feat(server): expose GC retention settings via config and env`** — a `feat` in a fix branch, which we would normally split, and here is the argument for keeping it: **none of the GC settings were reachable from a deployed server.** An operator watching a volume fill had no lever at all — not to tighten retention, not to loosen it, not to turn the new ceiling off. Shipping a retention fix that cannot be tuned in the environment where retention matters seemed worse than a slightly mixed branch. Happy to split it out if you would rather.

**3. `test(indexer): pin that repeated publish-then-GC bounds root objects on disk`** — see below.

**4. `fix(indexer): drop an unobservable clamp, and the vacuous test covering it`** — added 2026-08-13 with the rebase. Two self-corrections, both found by mutation-testing rather than by review. Detailed in the follow-up comment on the thread.

## Tests

The two reproduction tests are the ones worth reviewing, because they fail on `main`:

| test | pins |
|---|---|
| `repeated_publish_then_gc_keeps_root_count_bounded` | 40 generations, no age guard — root objects stay bounded |
| `repeated_publish_then_gc_bounded_with_a_live_age_guard` | 40 generations **with** a live 30-minute guard and fresh garbage records — the path a real cluster takes, and previously untested |

The second is the one that matters: it is the regime the defect lives in, and no existing test exercised it. On `main` the root count climbs monotonically toward 190; with the ceiling it stays at roughly `1 + max_old × 4`.

| test | pins |
|---|---|
| `test_hard_ceiling_collects_despite_recent_garbage` | the ceiling overrides the age guard |
| `test_hard_ceiling_never_collects_inside_retention_target` | the retention promise outranks the ceiling — **rewritten in the 4th commit; the original was vacuous, see the follow-up** |

**The pre-existing `test_clean_garbage_respects_time_threshold` passes unchanged**, which is the proof that the default ceiling does not weaken the age guard: the same 3-version chain with a 5-minute-old record is still left alone.

`fluree-db-indexer`: **387 passed**, 2 ignored *(343 at the time of the original posting; the increase is upstream's own tests arriving with the rebase)*. `fmt` clean; `clippy --all-features --all-targets` clean on the changed crates.

## Notes for review

- **`DEFAULT_HARD_MAX_MULTIPLE = 4` is a judgement call from one deployment.** We have no attachment to the number. If you prefer a different default, or prefer the ceiling opt-in rather than derived from `max_old_indexes`, both are one-line changes.
- **The `debug!`-level give-up paths in the collector are left as they are.** Promoting the genuinely anomalous ones (unparseable or unloadable garbage record) to `warn!` would make a wedged GC visible at default log level — during our incident we had to raise the log level to learn that GC was healthy. That is a logging-policy call rather than part of this fix, so we would send it separately if wanted.
- **Severity context:** the end state of unbounded index history is unrecoverable, not merely large. GC must write to free anything, so at 0 bytes free every index attempt fails and GC never runs again. Ours was dead ~34 hours while reporting healthy.


christophediprima:
Rebased onto current `main` (`a85e03682`), and while doing it I found two defects in my own work. Flagging both here rather than letting them sit, because one of them is a test that was passing for the wrong reason.

**The rebase itself.** Only `gc/collector.rs` conflicted. #1614 moved the age check out of the retention loop and into `release_manifest_nodes`, which signals "stop the whole walk" by returning `None` — and it does that for **two** distinct reasons: the record is too recent, *or* the manifest is unreadable/unparseable.

That distinction decides how the ceiling has to be expressed. The obvious re-derivation — skip the call entirely when past the ceiling — would override both reasons, so a corrupt manifest would quietly stop halting the walk and the release would proceed against nodes nothing had read. Instead the ceiling passes a **zero age floor** into the helper:

```rust
let age_floor_ms = if i >= hard_keep { 0 } else { min_age_ms };
```

Only the age reason is overridden; an unreadable manifest still breaks the loop exactly as #1614 intends. There is a comment on it in the code, because it looks like a stylistic choice and is not.

**Defect 1: the `hard_max_old_indexes` clamp was dead code.** I clamped the configured ceiling up to `max_old_indexes`, and documented it as preventing a misconfigured ceiling from collecting inside the retention target. It cannot do that. The retention loop starts at `keep_count = 1 + max_old_indexes`, so every index it visits is already outside the retention promise. In every case where the clamp changed `hard_keep` it moved it between two values both at or below `keep_count`, which leaves `i >= hard_keep` true either way.

I only noticed because I mutation-tested it: deleting the clamp broke no test at all. It is removed, and the comment now points at `keep_count` as where the guarantee actually lives. A guard that reads as load-bearing and is not seems worse than no guard.

**Defect 2: the test that claimed to cover it was vacuous.** `test_hard_ceiling_never_collects_inside_retention_target` used `max_old_indexes: Some(5)` against a **3-entry chain**. `clean_garbage` returns early on `index_chain.len() <= keep_count`, so the retention loop never ran. The test asserted `indexes_cleaned == 0` and passed because nothing was ever attempted — not because anything was protected. It stayed green with the entire ceiling feature deleted.

Rewritten with a chain **longer** than `keep_count` so the loop actually executes, and it now asserts both halves of the property at the most hostile setting (`hard_max_old_indexes: Some(0)`, i.e. override the age guard everywhere):

- the two gc-eligible versions **are** collected despite garbage records well inside the 30-minute age guard — the ceiling doing its job;
- the two retained versions **survive** — the retention promise holding regardless of what the ceiling says.

It now fails under two independent mutations (starting the loop at `0`, and removing the ceiling override) where before it survived both.

Both corrections are in a separate fourth commit rather than squashed into the original, so the diff of what changed and why stays readable.

**Verification on the new base:** `fmt` clean, `clippy -p fluree-db-indexer -p fluree-db-server -p fluree-db-api --all-features --all-targets -- -D warnings` clean, `fluree-db-indexer` 387 passed / 0 failed / 2 ignored — including #1614's own new tests (`manifest_gap_strands_only_its_own_step`, `gc_bounds_retained_chain_despite_manifest_gap`, `manifest_without_timestamp_does_not_stop_gc`), which was the check I most wanted to see pass given this lands on top of that rewrite.

fluree-db-indexer/src/gc/collector.rs
```diff
@@ -795,6 +836,401 @@ mod tests {
...
+                    max_old_indexes: Some(MAX_OLD),
+                    min_time_garbage_mins: Some(0),
+                    ..Default::default()
+                },
+            )
+            .await
+            .unwrap();
+
+            let roots = storage
+                .list_prefix(&format!("{LEDGER}/main/index/roots/"))
```
aaj3f: **`fluree-db-indexer/src/gc/collector.rs:992`** (and `:1093`) — *blocking.* **Both `repeated_publish_then_gc_*` harness tests count zero root objects in every generation, so their assertions pass no matter what GC does — including with GC deleted outright.** The prefix they list with is `"{LEDGER}/main/index/roots/"` with `LEDGER = "test:main"` (`:525`) → `"test:main/main/index/roots/"`, while the roots actually live at `fluree:memory://test/main/index/roots/…` (`content_path` maps ledger id `test:main` to path `test/main`, `fluree-db-core/src/storage.rs:1109-1114`; `MemoryStorage::list_prefix` is a raw `starts_with`, `memory.rs:85-92`). The prefix can never match.

  I proved it rather than inferring it: replacing `clean_garbage` with an immediate `return Ok(default())` reds 11 honest gc tests while **both harness tests stay green**, and adding `assert!(final_roots > 0)` fails both with zero roots observed — every assertion reduces to `0 < bound`.

  This also carries a description problem that's now an internal contradiction: the PR body still says these two tests are "the ones worth reviewing, because they **fail on main**" with "the root count climbs monotonically toward 190" — numbers a zero-observation harness cannot have produced — while your own commit `652eebaf9` says "**Both pass** … a narrowing result rather than a fix." One of those is right, and the fixture bug says it's the commit.

  The fix is small, and your own rewritten ceiling test already shows the way: count via `store.has(cid)` (or list the real `test/main/index/roots/` prefix), add the `final_roots > 0` guard so the harness can never go quiet again, re-run against main, and make the body match whichever result comes back.

fluree-db-indexer/src/gc/collector.rs
```diff
@@ -237,6 +237,20 @@ pub async fn clean_garbage(
     let min_age_mins = config
         .min_time_garbage_mins
         .unwrap_or(DEFAULT_MIN_TIME_GARBAGE_MINS);
+    // Past this many retained old versions the age guard is overridden — see
+    // `CleanGarbageConfig::hard_max_old_indexes`.
+    //
+    // Deliberately NOT clamped up to `max_old_indexes`. An earlier revision did
```
aaj3f: **`fluree-db-indexer/src/gc/collector.rs:243`** — *praise, and a retraction of mine.* Dropping the clamp was correct, and I want to be explicit that the pending version of my review had this one backwards — it praised `.max(max_old_indexes)` for a safety property ("misconfigured ceiling degrades to no-override rather than deleting inside the retention target") that never existed. Your commit's analysis is right: `hard_keep` is only ever consulted as `i >= hard_keep` inside `for i in (keep_count..len).rev()`, the clamp moved it between two values both ≤ `keep_count`, and the retention promise lives in the loop bound. I verified all three claims by mutation: re-adding the clamp changes nothing (38/38 green), removing the ceiling override reds exactly the two ceiling tests, and lowering the loop start into the retention target reds seven. Hunting down and deleting a behaviorally-unobservable guard *plus* the vacuous test that appeared to pin it is exactly the discipline this codebase needs — which is also why the finding above stings: the same hunt, pointed at the two headline harness tests, catches them the same way.

fluree-db-indexer/src/gc/collector.rs
```diff
@@ -294,6 +312,29 @@ pub async fn clean_garbage(
...
+        // `max_old_indexes` without limit and their history dominates the disk.
+        // Past `hard_keep` the disk wins — losing a concurrent read is
+        // recoverable by retry, running out of space is not, because GC itself
+        // has to write in order to free anything.
+        //
+        // Expressed as a zero age floor rather than as a skip of the check, so
+        // that the OTHER reasons `release_manifest_nodes` stops the walk — an
+        // unreadable or unparseable manifest — still stop it. Only the age
+        // reason is overridden.
+        let age_floor_ms = if i >= hard_keep {
```
aaj3f: **`fluree-db-indexer/src/gc/collector.rs:326`** — *praise.* Expressing the ceiling as a zero age floor into `release_manifest_nodes` — so only the *age* reason is overridden while an unreadable or unparseable garbage record still stops the walk, with the orphaning rationale documented at `:302-305` — is a careful preservation of #1614's fail-safe semantics through a nontrivial rebase conflict. The residual (a genuinely corrupt tail record still defeats the ceiling permanently) is a deliberate, documented inheritance from that design, and I'm fine with it as-is; noting it here only so it's a known property rather than a surprise.

fluree-db-api/src/lib.rs
```diff
@@ -1919,6 +1919,43 @@ impl FlureeBuilder {
         self
```
aaj3f, line 2110: **`fluree-db-api/src/lib.rs:1917`** — *optional, carried over.* The `FLUREE_GC_HARD_MAX_OLD_INDEXES` → `ServerConfig` → `with_gc_settings` → `IndexerConfig` → `CleanGarbageConfig` path is four hops with no test anywhere pinning it (only hits outside definitions are `config.rs:449-450` and `state.rs:512-515`). One test that sets the env var and asserts the value lands in `CleanGarbageConfig` is cheap insurance for the knob this whole PR exists to add. Minor and non-blocking — but if you agree it's right, I'd rather see it folded in now than lost in the backlog.

fluree-db-indexer/src/gc/mod.rs
```diff
@@ -43,7 +43,26 @@
...
+//! age-guarded garbage.
+//!
+//! ## Hard Ceiling
+//!
+//! So there is a third threshold that can override the age guard:
+//! - `hard_max_old_indexes`: past this many old versions, collect regardless of age
+//!
+//! A time window cannot bound bytes; only a count can. The guard still applies up
+//! to the ceiling, so the common case is unchanged and only genuinely runaway
+//! chains trade delay for disk.
```
aaj3f: **`docs/indexing-and-search/reindex.md:103` (not in this diff — noting from the nearest changed file).** — *optional, carried over.* Both GC docs still enumerate the retention knobs without mentioning the ceiling, so an operator reading them sees two settings and the old semantics. A short paragraph on `gc-hard-max-old-indexes` (and the ~7.7 GiB-per-retained-version arithmetic from your module docs, which are good) closes the loop.

(Commenting from the module docs because neither GC docs file is in this diff.)

aaj3f (changes requested):
@christophediprima, the rewrite improved this, so thank you for that! 

The one thing blocking approval is that the two headline harness tests: `repeated_publish_then_gc_keeps_root_count_bounded` and its age-guard sibling observe zero roots in every generation (`"test:main/main/index/roots/"` can never match the stored `test/main/index/roots/…` paths), so they pass with GC deleted outright; I ran that mutation and 11 honest tests went red while both stayed green. The body still presents them as "fail on main" with a "climbs toward 190" trajectory they cannot have measured, while your own commit message says both pass. Your rewritten ceiling test already demonstrates the fix: count via the store/CIDs, add a `final_roots > 0` guard, re-run against main, and let the body say whatever that run actually shows. Once the harness observes something real, this approves quickly: the ceiling itself is correctly implemented and genuinely pinned.

- **Performance (speed first, memory second):** ✔ GC planning path only; an integer comparison per eligible entry. No regression risk.
- **Patterns/abstractions:** ✔ the zero-age-floor expression extends #1614's release machinery instead of bypassing it.
- **Testing:** ⚠️ the ceiling tests are real and mutation-verified; the two headline harness tests are vacuous (the blocker), and the env/config plumbing is unpinned (optional).

Verified at branch HEAD (`5a96971f9`), agent-run and lead-re-verified: GC-disable mutation (11 red / harness green — the vacuity proof); `final_roots > 0` probe red on both; clamp re-add no-op (38/38); ceiling-override removal → exactly the two ceiling tests red; loop-start mutation → 7 red; `merge-tree` clean vs `origin/main` fetched today.

Happy to re-review the moment the harness counts something. The fix is a few lines and everything else here is ready.

bplatz:
Pushed a merge of current `main` plus four commits addressing the review:

- **Harness fix (blocking):** the two `repeated_publish_then_gc_*` tests now derive the roots prefix from a real address and refuse a zero count. Mutation-checked: GC deleted → both bounded tests red; override removed (main's behaviour) → exactly the ceiling tests red. Note for the description: with a real harness only the ceiling test fails on `main`; the no-age-guard baseline passes there.
- **Ceiling is now opt-in.** Overriding the age guard can release artifacts a query still reading an older root needs, so no default ceiling is derived; unset means the guard always holds. Each override is counted in `CleanGarbageResult::age_guard_overridden` and the completion log line. A new test pins that no default is derived.
- **Docs/comments** no longer describe the ceiling as bounding disk: it bounds versions, not bytes.
- Optional items folded in: flag → `IndexerConfig` path pinned at both hops, the three keys accepted in `[server.indexing]`, and the GC docs updated.

PR #1604, opened 2026-08-07
perf(bm25): scope the incremental indexing query to the subjects that changed

christophediprima:
## Summary

An incremental BM25 sync already knows, from the commit log, exactly which subjects changed. It then runs the **whole indexing query over the whole ledger** and applies that set afterwards as a filter over the results. The step's own comment says so:

```rust
// 8. Re-run indexing query and filter to affected subjects
let results = self.execute_bm25_indexing_query(&ledger, &query).await?;
```

So a one-document delta costs O(corpus). That is the dominant cost once an index is *maintained* rather than built once.

This binds the query's subject variable to the affected IRIs with a `values` clause **before** executing it, so the narrowing reaches the scan instead of the result set.

A correctness fix comes first, because the performance fix depends on it: the two sync paths did not canonicalise the document `@id`, so the same document could be stored under two spellings.

## Commit 1 — canonicalise the document `@id` in both sync paths

`create_full_text_index` expands prefixed `@id` values to full IRIs via `expand_ids_in_results` before building the index. Neither `sync_bm25_index` nor `resync_bm25_index` did, so **the document key depended on which code path last wrote it**.

The effect is a silent split. With an `@context` mapping `ex:` to `http://example.org/`, the initial build stores `http://example.org/doc1`; the first sync touching that document re-inserts it as `ex:doc1` **without removing the original**. The index then holds both spellings, `@id` lookups by full IRI miss anything re-synced, and document counts drift upward.

Expanding in the sync paths too makes the key the same value whichever path wrote it.

**This deletes existing code, so it is worth being explicit about.** The sync path currently compensates for the split by expanding the affected set to hold both spellings (`affected_iris_expanded`) so that a match succeeds whichever way the key was written. Once both sides of the match are canonical that expansion has nothing left to do, so it goes away and `apply_update` takes `affected_iris` directly. The workaround was treating the symptom; this fixes the cause. If it is load-bearing for a case I have not found, say so and I will keep it.

**For existing indexes:** the first sync after this change rewrites prefixed keys back to canonical form, so a split index heals itself rather than needing a manual rebuild.

## Commit 2 — scope the query

Measured with tracked fuel, above the flat 1.000 query-entry floor, one changed subject out of a corpus:

| corpus | full | scoped |
|---|---|---|
| 100 | 2.110 | **0.031** |
| 400 | 8.410 | **0.031** |
| 1600 | 33.610 | **0.031** |

Scoped cost is flat in corpus size; unscoped is linear. That is the whole claim.

### Two limits, stated plainly

**The saving is over the indexed portion only.** A bound subject seeks into leaflets there. Novelty is a linear structure with no subject index, so rows committed since the last index build are still walked — over that portion alone the win measures **~1.5×, not ~270×**. This is therefore **complementary to the reindex thresholds, not a substitute**: `reindex_min_bytes` is what bounds how much novelty a sync must walk.

**It declines any shape it cannot narrow safely.** `scope_indexing_query_to_subjects` returns `None` for:

- a list-form `select` (`["?x", "?title"]`), which does not identify which variable is the document;
- a query already carrying a top-level `values`, which would have to be merged rather than replaced;
- an empty affected set, which the caller already treats as a full resync.

Every such case falls back to the existing full-scan-and-filter path, which stays correct. The subject variable is taken as the single key of an object-form `select` — the shape `PropertyDeps::from_indexing_query` already assumes and the BM25 docs specify.

## Tests

**`bm25_incremental_sync_matches_full_resync`** — the differential test, and the one that matters. It mutates a strict subset of documents, drives one index through incremental sync, rebuilds another from scratch, and requires the two to agree on document set, per-document lengths and vocabulary size. It **fails before commit 1** with the two spellings visible in the diff. Nothing in it asserts *how* the query was narrowed — only that narrowing changed no outcome, which is the property that must hold even if the scoping heuristic changes later.

**`scoped_indexing_query_narrows_the_indexed_scan`** — pins the fuel claim. A `values` clause that degraded into a post-filter would still return the right rows and would still fail this test, which is the point: it measures the mechanism, not the result.

## Verification

Rebased onto `upstream/main` (0 behind) and verified at the tip:

- `fluree-db-api --lib --all-features` → **971 passed**
- `--test grp_graphsource` → **152 passed**; the 21 bm25 tests by name → **21 passed**
- `cargo fmt --all -- --check` clean
- `cargo clippy -p fluree-db-api --all-features --all-targets` clean

The rebase needed one real resolution, in `it_graph_source_bm25.rs`: the conflict cut *through* a test function body, with both sides sharing a `let query = json!(…)` block and a single closing brace. Taking both halves would have fused two different tests into one. Resolved by rebuilding the whole conflicted region twice — once per side — so each test kept its own body.

## Notes for review

- **The two commits are separable but ordered.** Commit 2 is safe without commit 1 only because the affected-set expansion papers over the split; with the expansion removed, the canonicalisation is what makes the match sound. I would land them together, but happy to split into two PRs if you would rather review the correctness fix on its own.
- **I have not measured an end-to-end sync latency improvement**, only query fuel. The end-to-end number depends on the novelty/indexed ratio at the moment of sync, which is a deployment property rather than a property of this change — hence the two limits above rather than a single headline figure.
- **Why this matters beyond a micro-optimisation, in one line:** at fixed data size the cost of maintaining an index grows with corpus, so this is the term that gets worse as a deployment succeeds. It is the only one of the scaling terms we have measured that worsens with *data* at a *fixed* number of indexes.


fluree-db-api/src/graph_source/bm25.rs
```diff
@@ -20,7 +20,81 @@ use fluree_db_query::{execute, ContextConfig, ExecutableQuery, QueryOutput, VarR
...
+/// structure with no subject index, so rows committed since the last index
+/// build are still walked and the win over that portion is only ~1.5x. That
+/// makes this complementary to the reindex thresholds rather than a
+/// substitute — `reindex_min_bytes` is what bounds how much novelty a sync
+/// has to walk.
+fn scope_indexing_query_to_subjects(
+    query: &JsonValue,
+    affected_iris: &HashSet<Arc<str>>,
+) -> Option<JsonValue> {
+    if affected_iris.is_empty() || query.get("values").is_some() {
```
aaj3f: *Blocking (correctness).* **The decline list does not cover `limit` / `offset`, and with either of them present the scoped query indexes documents a full rebuild would exclude.** The guard checks for a top-level `values` and for the select shape, and nothing else, so a query carrying `limit` is narrowed and executed. Nothing upstream stops that reaching here — `Bm25CreateConfig::validate` (`fluree-db-api/src/graph_source/config.rs:114-160`) requires only that a `select` or `selectOne` exists.

The mechanism is `apply_update`'s two halves (`fluree-db-query/src/bm25/builder.rs:419-443`). Unscoped, `seen_iris` is whatever the truncated result set contained, so an affected subject outside the top-N lands in `affected \ seen` and is passed to `remove_document`. Scoped, that same subject is the only thing the query returns, so it is upserted instead. Same input, opposite outcome.

Concretely, with `"limit": 2` over a four-document corpus: the initial build indexes `doc1, doc2`. A commit then touches `doc4`. Today the sync runs the full query, gets `doc1, doc2` back, finds `doc4` unseen, and no-ops. With this change the sync runs `values ?x [{"@id": doc4}] … limit 2`, gets `doc4` back, and upserts it — leaving a three-document index where `resync_bm25_index` would produce two. That is exactly the invariant `bm25_incremental_sync_matches_full_resync` exists to protect, and it wouldn't catch it because its query has no `limit`.

`selectOne` at `:69` is the same class and I'd argue it's strictly safer to decline it than to accept it "for symmetry with the parser": a `selectOne` query returns one row, so scoping changes *which* row, not just how it's found.

The fix is small, and I think it belongs in the early return rather than anywhere cleverer:

```rust
if affected_iris.is_empty()
    || query.get("values").is_some()
    || query.get("limit").is_some()
    || query.get("offset").is_some()
    || query.get("groupBy").is_some()
{
    return None;
}
```

…and dropping the `.or_else(|| query.get("selectOne"))`. Every one of those falls back to the path that is correct today, so this costs nothing but the optimization on shapes we can't reason about — which is the trade the function's own doc comment already commits to.

fluree-db-api/src/graph_source/bm25.rs
```diff
@@ -969,30 +1050,45 @@ impl crate::Fluree {
...
-                    affected_iris_expanded.insert(Arc::from(prefixed));
-                }
-            }
-        }
+        let results = expand_ids_in_results(results, &prefix_map);
 
         // 9. Apply incremental update
         let mut updater = IncrementalUpdater::new(source_ledger_alias.as_str(), &mut index);
-        let update_result = updater.apply_update(&results, &affected_iris_expanded, ledger_t);
+        let update_result = updater.apply_update(&results, &affected_iris, ledger_t);
```
aaj3f, line 1197: *Blocking (the migration claim in the PR description is wrong).* The description says "the first sync after this change rewrites prefixed keys back to canonical form, so a split index heals itself rather than needing a manual rebuild." It doesn't. `apply_update`'s removal loop iterates `affected_iris` (`fluree-db-query/src/bm25/builder.rs:437`), which after commit 1 contains only full IRIs, so it never visits the orphaned prefixed `DocKey`. The upsert writes the canonical key and the prefixed one stays.

I drove this directly rather than inferring it — two `apply_update` calls against one index, first with a prefixed `@id` and prefixed affected set, then with the full-IRI pair — and `num_docs()` comes back **2, not 1**.

What makes this more than a doc correction is that commit 1 *removes* the thing that used to clean up. At merge-base, `affected_iris_expanded` put both spellings in the affected set, so the removal loop took the full-IRI twin out; you can see it in the revert run, where every document ends up with exactly one spelling. After this change the migrating sync leaves both. So for one sync the fix produces the duplicate-key state the old code avoided, and document counts really do drift up — the thing the description attributes to the *old* behavior.

Only `apply_full_sync` heals, because it enumerates `existing_keys` for the ledger first (`builder.rs:475-482`) — i.e. `resync_bm25_index`, not `sync_bm25_index`. I don't think that needs new code; it needs the description and the release note to say "existing BM25 indexes need one `resync` after upgrading", so operators don't skip a rebuild they actually need. (Your commit message for `43b508c22` already describes the non-removal mechanism accurately — it's just the PR body that reads the other way.)

Commenting here because `fluree-db-query/src/bm25/builder.rs` is not in this diff; this is the line that swaps `affected_iris_expanded` for `affected_iris` and so makes the removal loop stop visiting the prefixed twin.

fluree-db-api/tests/it_graph_source_bm25.rs
```diff
@@ -1497,3 +1497,221 @@ async fn bm25_sync_refuses_a_retracted_index() {
...
+/// the point of scoping the query.
+///
+/// Measured against a corpus large enough that a full scan cannot be mistaken
+/// for a narrow one, on an **indexed** ledger because that is what the sync
+/// path meets in practice: novelty holds only commits since the last index
+/// build, and everything older is in leaflets that a bound subject can seek
+/// into. Fuel on the indexed corpus is flat in corpus size when scoped and
+/// linear when not.
+#[tokio::test]
+async fn scoped_indexing_query_narrows_the_indexed_scan() {
```
aaj3f, line 1656: *Blocking (coverage).* **Nothing in the tree fails if the scoping is never applied.** I set `scoped = None` at `bm25.rs:1058` so the full query always runs, and all 5 unit tests plus all 21 integration tests passed. `scoped_indexing_query_narrows_the_indexed_scan` also passes with the *entire* PR source reverted to merge-base — it hand-builds its own `values` clause, so it measures whether the engine narrows a scan, which was already true before this PR.

That's a genuinely useful characterization test and I'd keep it — it would catch a future engine change that quietly turned `values` into a post-filter. But it is not a guard on this PR, and as things stand the performance claim that is the entire point of commit 2 could be reverted by accident and CI would stay green. The cheap version is to build the measured query with `scope_indexing_query_to_subjects` instead of by hand, which at least ties the fuel number to the function. The better version drives `sync_bm25_index` itself and asserts on fuel or on a call counter, so the *wiring* is what's pinned.

fluree-db-api/src/graph_source/bm25.rs
```diff
@@ -969,30 +1050,45 @@ impl crate::Fluree {
...
-        // 8. Re-run indexing query and filter to affected subjects
-        let results = self.execute_bm25_indexing_query(&ledger, &query).await?;
+        // 8. Re-run the indexing query, scoped to the affected subjects where we
+        //    can. `scope_indexing_query_to_subjects` returns None when the query
+        //    shape is not one we can safely narrow, in which case this falls back
+        //    to the full scan and the filtering in `apply_update` below does the
+        //    work — same results either way, just more of them computed.
+        let scoped = scope_indexing_query_to_subjects(&query, &affected_iris);
+        let scoped_query = scoped.as_ref().unwrap_or(&query);
+        if scoped.is_some() {
```
aaj3f: *Optional.* Since the fallback is silent and correct, an index whose query shape can never be narrowed gets no speedup and no signal — this `debug!` fires only on the success path. Might be worth a one-time `debug!` on the `None` path naming which condition declined it — cheap, and it turns "why is my sync still slow" into a log line. Happy to punt to a follow-up.

fluree-db-api/tests/it_graph_source_bm25.rs
```diff
@@ -1497,3 +1497,221 @@ async fn bm25_sync_refuses_a_retracted_index() {
...
+/// O(corpus) per sync, but it is only a safe optimisation if the narrowed query
+/// yields exactly what the wide one did.
+///
+/// So: build two identical indexes over one ledger, mutate the ledger, then
+/// take one down the incremental path and rebuild the other from scratch, and
+/// require the resulting structures to match. Nothing here asserts *how* the
+/// query was narrowed — only that narrowing changed no outcome, which is the
+/// property that has to hold even if the scoping heuristic changes later.
+#[tokio::test]
+async fn bm25_incremental_sync_matches_full_resync() {
```
aaj3f, line 1516: *Praise.* `bm25_incremental_sync_matches_full_resync` is the right test and it is not vacuous. I reverted the source to merge-base and it fails with the split printed straight into the assertion message: `left: [("ex:doc2", 5), ("ex:doc4", 2), ("http://example.org/doc1", 3), ("http://example.org/doc3", 3)]`. Asserting the *property* (incremental ≡ rebuild) rather than the mechanism is exactly right, and the `!sync.was_full_resync` guard against the test silently degrading into a resync is a nice touch.

fluree-db-api/src/graph_source/bm25.rs
```diff
@@ -20,7 +20,81 @@ use fluree_db_query::{execute, ContextConfig, ExecutableQuery, QueryOutput, VarR
...
+/// The subject variable is the single key of an object-form `select` —
+/// `{"select": {"?x": ["@id", "ex:title"]}}` — which is the shape
+/// `PropertyDeps::from_indexing_query` already assumes and the shape the BM25
+/// documentation specifies. A list-form select (`["?x", "?title"]`) does not
+/// identify which variable is the document, so it is left alone.
+///
+/// The saving comes from the *indexed* portion of the ledger, where a bound
+/// subject seeks into the leaflets instead of scanning them: fuel is flat in
+/// corpus size when scoped against linear when not, ~270x at 400 documents
+/// (`scoped_indexing_query_narrows_the_indexed_scan`). Novelty is a linear
```
aaj3f, line 128: *Praise.* The two limits stated here and in the description — the novelty portion only winning ~1.5×, and this being complementary to `reindex_min_bytes` rather than a substitute — are the kind of thing that usually gets left out of a perf PR. Stating them makes the ~270× credible rather than suspicious.

aaj3f (changes requested):
**🔄 Request changes.**

I'm genuinely glad someone is looking at this, @christophediprima — the framing that this is the scaling term that gets *worse as a deployment succeeds* is right, and it's the sort of cost that hides because every individual sync still finishes. Using a `values` binding rather than inventing a subject-filter parameter is the right instinct too; it reuses what the engine already optimizes. And `bm25_incremental_sync_matches_full_resync` is a properly good test — I reverted the source to merge-base and it fails with the two spellings printed right in the assertion.

Three things I'd like resolved first, all of them fixable without rethinking the design.

The one that actually worries me is at `bm25.rs:63`: the decline list covers `values` and the select shape but not `limit` or `offset`, and with a `limit` present the scoped query upserts documents that the initial build and a full resync both exclude — I walked it through `apply_update` and the unscoped path *removes* that subject where the scoped path *adds* it. Nothing upstream rejects a `limit` in an indexing query, and the differential test wouldn't catch it because its query doesn't have one. A few more keys in the early return closes it.

Second, the "split index heals itself" line in the description isn't what the code does. `apply_update`'s removal loop only walks `affected_iris`, so the orphaned prefixed key is never removed — I drove `apply_update` directly and got `num_docs() == 2`. And because commit 1 removes `affected_iris_expanded`, which is what used to clean the twin up, the migrating sync briefly *creates* the duplicate state the description attributes to the old behavior. I don't think that needs code, but existing deployments need to be told to run one `resync` after upgrading.

Third, nothing fails if the scoping never runs — I forced the fallback and all 26 tests stayed green, and the fuel test passes with the whole PR reverted. Worth tying it to `scope_indexing_query_to_subjects` at minimum, so the win has something defending it.

- **Patterns/abstractions:** ✔ Reuses the `values` clause and the existing fallback-to-correct-slow-path convention instead of adding a parallel narrowing mechanism.
- **Performance (speed first, memory second):** ⚠️ Real and well-characterized win on the indexed portion, honestly bounded in the description — but unguarded by any test, and the `limit`/`offset` gap means the narrowing can change results, which under this repo's rubric outranks the speed win.
- **Testing:** ⚠️ Both new tests are wired into `grp_graphsource` and run in CI, and the differential one has teeth; but the perf commit's integration has no failing-on-revert coverage at all.

Verified locally at branch HEAD (`e77f6433c`): `cargo nextest run -p fluree-db-api --lib -E 'test(scope_tests)'` → 5 passed; `--test grp_graphsource -E 'test(bm25)'` → 21 passed with both new tests by name; fmt and clippy clean. Mutations: source reverted to merge-base → `bm25_incremental_sync_matches_full_resync` red (good); `scoped = None` forced at `:1058` → 0 red across all 26 tests; `apply_update` probe → `num_docs() == 2`; `scope_indexing_query_to_subjects` with `limit` → returns `Some`.

None of this is a redesign — the `limit`/`offset` guard is a few lines, the migration note is a paragraph, and the coverage fix is rewiring one test to call the function it's meant to be measuring. Happy to look again as soon as those land, and if you'd rather split the coverage work into a follow-up issue and just get the guard in, that's fine by me too.

christophediprima:
All three blocking items were real and all three are fixed. Two of them changed my mind about something rather than just my code, so I want to be specific about what.

**The `limit` divergence.** Accepted exactly as described, and the worked example is what made it obvious — I had been reasoning about scoping as "the same query, found faster", which is true right up until a clause makes the result set depend on how many rows the scan produced. The guard now declines every clause of that class before it looks at anything else.

While implementing it I found the suggested snippet is one spelling short. `parse/options.rs:272` reads `groupBy` **or** `group-by`, and `parse/mod.rs:273` reads `selectOne` **or** `select-one`, so a guard naming only the camelCase forms leaves the divergence reachable through the alias. I also added `having`, which filters groups and truncates for the same reason. `limit`, `offset` and `having` have no alias. `orderBy` is deliberately still allowed: ordering alone does not change which rows come back, and every clause that turns an order into a truncation is now declined. Clause and reason live together in one table so a future spelling cannot inherit the wrong explanation.

`selectOne` is dropped from the accepted shapes, for the reason given — symmetry with the parser is not a safety argument.

**The migration claim.** You were right and I had the mechanism backwards. `apply_update`'s removal loop walks `affected_iris`, which after commit 1 holds only full IRIs, so it never reaches the orphaned prefixed key. Worse, removing `affected_iris_expanded` removes the thing that used to take the twin out, so for exactly one sync this PR *creates* the duplicate state my description blamed on the old behaviour.

I have not added code to heal it. Making the incremental path enumerate existing keys would give every sync the O(corpus) cost this PR exists to remove, and `resync_bm25_index` already does it correctly. So it is an upgrade note instead — and I put it in `docs/indexing-and-search/bm25.md` under **Syncing**, not only in the PR description, so an operator meets it where they look. That section also documented the old "sync cost scales with corpus size" behaviour as fact, which this PR falsifies; it now describes the narrowing, lists every declined shape, and says the decline is logged.

**The coverage gap.** Confirmed by your experiment, and it is the criticism I am most glad you made — the optimisation was genuinely unguarded. New test: `incremental_sync_actually_scopes_the_indexing_query`. It drives `sync_bm25_index` end to end and asserts on the branch the sync reports taking, so forcing `scoped = None`, deleting the call, or making the guard reject this shape all fail it.

I did not take the cheap version — building the measured query with `scope_indexing_query_to_subjects` — because the function is private and the fuel test is an integration test, so it cannot reach it without widening the crate's API for a test. The wiring test supersedes it anyway: it pins the real call site rather than the shape of one argument.

`scoped_indexing_query_narrows_the_indexed_scan` stays, and its doc comment now says plainly that it is a characterization test of the query engine, that it passes with this PR reverted, and that it is not part of this PR's guard set. It is there to catch a future engine change that turns a `values` binding into a post-filter.

**Optional, taken.** The decline is no longer silent. `scope_indexing_query_to_subjects` returns `Result<JsonValue, &'static str>` where the error is the reason, and the caller logs it at `debug` on the fallback path. That also made the unit tests stronger: each one now asserts *which* condition fired, so deleting one guard cannot be covered by another still returning "no".

fluree-db-api/src/graph_source/bm25.rs
christophediprima: Accepted, with two additions to the list you suggested.

The early return now declines `limit`, `offset`, `groupBy`, `group-by`, `having`, `selectOne` and `select-one`, and `.or_else(|| query.get("selectOne"))` is gone.

`group-by` and `select-one` matter because the parser accepts both spellings — `parse/options.rs:272` and `parse/mod.rs:273` — so the camelCase-only guard would have left exactly the divergence you described reachable through the alias. `having` filters groups and truncates for the same reason. `orderBy` is still allowed on purpose: ordering does not change which rows come back, and everything that converts an order into a truncation is declined.

Your `limit` example is now the worked example in the function's doc comment, and `declines_every_result_shaping_clause_in_both_spellings` covers each clause in each spelling.

fluree-db-api/src/graph_source/bm25.rs
christophediprima: You are right, and thank you for driving it rather than inferring it — `num_docs() == 2` is not something I would have argued with, and my description was asserting the opposite.

The description is rewritten to say a `resync` is required, to name why the incremental path cannot heal it (`apply_update` walks `affected_iris`; only `apply_full_sync` enumerates `existing_keys`), and to say that this change makes it briefly worse by removing `affected_iris_expanded`. The upgrade note is in `docs/indexing-and-search/bm25.md` under **Syncing** so it is not description-only.

No code change: making the incremental path enumerate existing keys would restore the O(corpus) cost the PR removes, and a full resync already repairs it correctly.

fluree-db-api/tests/it_graph_source_bm25.rs
christophediprima: Confirmed and fixed. `incremental_sync_actually_scopes_the_indexing_query` drives `sync_bm25_index` over a three-document corpus with one document changed and asserts the sync reports the scoped branch and not the decline. Your `scoped = None` mutation fails it; so does deleting the call site.

I went with the "wiring" version rather than the cheap one: `scope_indexing_query_to_subjects` is private, so an integration test cannot build the measured query from it without widening the crate API for a test's benefit.

`scoped_indexing_query_narrows_the_indexed_scan` stays as you suggested, with a doc comment that now states what it is — an engine characterization test that passes with this PR reverted, kept to catch a `values` binding degrading into a post-filter, and explicitly not part of this PR's guard set.

fluree-db-api/src/graph_source/bm25.rs
christophediprima: Taken rather than punted. The function returns `Result<JsonValue, &'static str>`; the error is the reason, and the caller emits it at `debug` on the fallback path alongside `graph_source_id` and `affected_count`.

The unit tests improved as a side effect — they now assert the reason string, so each guard is pinned individually instead of any decline satisfying any test.

docs/indexing-and-search/bm25.md
```diff
@@ -246,7 +246,13 @@ println!("Upserted: {}, Removed: {}", sync_result.upserted, sync_result.removed)
...
+
+> **Upgrading an existing index.** Indexes built before document identifiers were canonicalised can hold two keys for one document — the full IRI written at build time and a prefixed form written by a later sync. **Run one `resync_bm25_index` after upgrading.** An incremental sync will not repair it: the removal pass only visits subjects the commit affected, so it never reaches the orphaned key, and document counts stay inflated until a full resync enumerates the existing keys and rewrites them.
```
aaj3f, line 255: *Blocking (prose only, relocated).* **The live PR body still says a split index "heals itself rather than needing a manual rebuild" — the claim we established is false, and this docs line is the truth it needs to match.** The body still says "the first sync after this change rewrites prefixed keys back to canonical form, so a split index heals itself rather than needing a manual rebuild" — the exact claim we established is false, and the one thing your (otherwise excellent) docs commit didn't reach. Your reply says the description was rewritten, but as of today the live body still carries the old sentence, and the decline-list section below it is also stale (three cases / "returns `None`" vs. the shipped seven-reason `Result`). `docs/indexing-and-search/bm25.md:255` now tells the true story — one `resync_bm25_index` required, incremental can't heal — so this is just bringing the body up to the standard the docs already meet. Two paragraphs, no code.

(Anchoring on the docs passage because the PR body itself has no diff line to comment on.)

fluree-db-api/src/graph_source/bm25.rs
```diff
@@ -20,7 +20,177 @@ use fluree_db_query::{execute, ContextConfig, ExecutableQuery, QueryOutput, VarR
...
+
+/// Clauses that shape the result set rather than the scan, paired with why each
+/// one refuses narrowing, in **both** spellings the query parser accepts
+/// (`parse/options.rs` reads `groupBy` or `group-by`; `parse/mod.rs` reads
+/// `selectOne` or `select-one`). Any of them present means the query cannot be
+/// narrowed safely — see [`scope_indexing_query_to_subjects`].
+///
+/// Reason and clause live together so adding a spelling cannot silently inherit
+/// the wrong explanation.
+const RESULT_SHAPING_CLAUSES: &[(&str, &str)] = &[
```
aaj3f, line 41: **`fluree-db-api/src/graph_source/bm25.rs:41`** — *praise.* The clause list with a per-clause reason string is better than the bare decline I asked for — and catching the parser's kebab-case aliases (`group-by`, `select-one` at `parse/options.rs:272` / `parse/mod.rs:273`) closed a divergence the camelCase-only version of this fix would have left reachable. I re-ran the sweep for anything else result-shaping: `selectDistinct`/`construct`/`ask` fall to the no-`select`-clause arm, boolean `distinct` is safe because rows are keyed by `@id`, and your `orderBy`-stays reasoning holds. Nothing left over.

fluree-db-api/tests/it_bm25_sync_scoping.rs
```diff
@@ -0,0 +1,121 @@
...
+use fluree_db_api::{Bm25CreateConfig, FlureeBuilder};
+use serde_json::json;
+
+/// Message emitted by `sync_bm25_index` when it narrows the indexing query.
+const SCOPED: &str = "Scoped indexing query to affected subjects";
+/// Message emitted when it declines and falls back to the full scan.
+const DECLINED: &str = "Indexing query not narrowed; falling back to the full scan";
+
+#[tokio::test(flavor = "current_thread")]
+async fn incremental_sync_actually_scopes_the_indexing_query() {
```
aaj3f, line 37: **`fluree-db-api/tests/it_bm25_sync_scoping.rs:37`** — *praise.* This is the wiring test the round-1 finding asked for, and you chose the harder, more honest version — driving `sync_bm25_index` itself rather than widening the crate API so a test could reach the private helper. I verified both mutations red: forcing the scope to `Err` fails it, and deleting the call site fails it. The relabeling of `scoped_indexing_query_narrows_the_indexed_scan` as an engine characterization test outside the guard set is exactly the right disposition for a test that passes with the PR reverted.

aaj3f (approved):
This all looks very good, @christophediprima and I appreciate the feedback responses! All three blockers were engaged on the merits, and the two code-side ones are not just fixed but pinned: I re-ran your mutations rather than hand-waving them, and removing `offset` from the clause list fails `declines_every_result_shaping_clause_in_both_spellings` on its specific reason string, while both gutting mutations (scope forced to `Err`, call site deleted) fail `incremental_sync_actually_scopes_the_indexing_query`. The kebab-case alias catch (`group-by`, `select-one`) made the guard better than the one I asked for, and taking the decline-observability suggestion as a typed `Result` with per-reason unit assertions instead of punting it was exactly the fold-it-in-now spirit.

The one thing I'd ask before this merges is a body edit, not code: the PR body still says a split index "heals itself rather than needing a manual rebuild". The claim we established is false, and which your own `docs/indexing-and-search/bm25.md:255` now correctly contradicts (one `resync_bm25_index` required; incremental can't heal because `apply_update` only visits affected subjects). The body's decline-list section is also a generation behind the shipped code. Your thread reply says the description was rewritten--I think that edit just never made it from the docs/commit-message pass into the body itself, but since the body is what a future reader (or a squash-merge) consults (and also is used during our release note generation), those two paragraphs should tell the same story the docs do. I'm approving now rather than holding the PR hostage to prose.

Verified at branch HEAD (`da75fbaea`), agent-run and lead-re-verified: scope unit suite green + clause-removal mutation red; scoping wiring test green + both gutting mutations red; full bm25 integration set green; live body text re-fetched 8/19 to confirm the stale claim; parser alias lines confirmed at `parse/options.rs:272` / `parse/mod.rs:273`.

PR #1612, opened 2026-08-08
feat(storage): atomic write configuration for file storage

bplatz:
## What

Adds a configurable durability mode for the local file backend, so a deployment can choose when a write is acknowledged: after the bytes reach the device, or after they reach the OS page cache. Set it per storage node in JSON-LD config, or per process with `FLUREE_STORAGE_FSYNC`.

Applies to the local file backend. S3 acknowledges after replication and the Raft log flushes on its own path; neither is affected.

## The two settings

| Mode | Acknowledged when | Protects against |
|---|---|---|
| `sync` (default) | the bytes and the directory entry naming them are flushed to the device | process crash, `SIGKILL`, kernel panic, power loss, hard reset |
| `page-cache` | the bytes reach the OS page cache | process crash, `SIGKILL` |

Under `page-cache` the kernel owns the bytes, so anything that kills only the process leaves them intact. Losing the kernel or the power can drop writes that were already acknowledged.

Both settings write atomically: bytes are staged alongside the destination and moved into place, so a reader at an address sees either the previous contents or the complete new contents, and an interrupted write leaves nothing at the final name. That holds in either mode — the setting governs power-loss durability, not visibility.

## Choosing one

`sync` suits a ledger that is the system of record on a single node, or any deployment where an acknowledged commit has to survive the machine going down uncleanly.

`page-cache` is the better trade when the data can be reconstructed or the acknowledgement is not load-bearing:

- **Bulk imports**, which are restartable from the source data
- **CI and local test loops**, where per-commit flushes buy nothing
- **Benchmarks**, when you want the storage layer out of the measurement
- **Replicated deployments**, where a node that loses power is rebuilt from a peer rather than from its own disk

## What it costs, per platform

Measured on the storage primitive directly:

| | per write |
|---|---|
| no flush | 0.059 ms |
| `fsync` — Linux `sync_all` | 0.081 ms |
| `F_FULLFSYNC` — macOS `sync_all` | 4.024 ms |

The platform gap is the important part. On Linux, `fsync` reaches the device, so `sync` costs roughly 0.02 ms over `page-cache` — small enough that most deployments should leave it on. On macOS, `fsync` does not flush the drive's write cache; only `F_FULLFSYNC` does, and that is what Rust's `sync_all` issues there. The macOS number buys a real guarantee, but it is ~50× the Linux cost and should not be read as representative: a transaction-heavy suite runs about 4× slower on macOS under `sync` and shows no such gap on Linux.

The practical upshot is that macOS development machines are the main place `FLUREE_STORAGE_FSYNC=0` earns its keep.

## Scope of the flush

Content that can be rebuilt from the commit chain — index nodes, dictionaries, sketches, annotation arenas — is written page-cache in either mode. It is written at far higher volume than commits and is reproducible, so keeping it off the flush path holds indexing throughput steady regardless of the setting.

The setting governs content that is the system of record: commits, transactions, ledger config, graph-source mappings, and nameservice records. The classification lives in `ContentKind::is_derived()` behind an exhaustive match, so a new content kind has to declare which side it is on.

## Configuring it

Per storage node:

```json
{
  "@id": "fileStorage",
  "@type": "Storage",
  "filePath": "/var/lib/fluree",
  "durability": "sync"
}
```

Read by both the JSON-LD and flat storage parsers. An unrecognized value is rejected at config load rather than falling back to a default, so a typo cannot quietly select a durability nobody asked for.

Per process, `FLUREE_STORAGE_FSYNC` accepts the usual falsy spellings (`0`, `false`, `off`, `no`) and takes precedence over the config property, matching the documented ladder where environment outranks a config file. That way a one-off run — an import, a benchmark — needs no config edit. It is read once per storage construction and defaults an operator field, so tests set the field directly and no test touches process environment.

Default is `sync`. Deployments that prefer the throughput profile of the page-cache path can select it at either surface.

## Notes for review

- `StorageConfig` gains a `durability` field, so struct-literal construction needs the extra field (five in-tree sites updated).
- `create_file_storage` now takes `&StorageConfig` so it can honour the setting. The prior signature had no callers beyond its own test, so it was folded in rather than adding a second factory alongside it.
- `Durability` lives in `storage` rather than the `native`-gated `file` module, since `StorageConfig` carries it on every target.
- Staging paths are unique per process and per call, because `write_bytes` takes no lock and two writers may target one address concurrently.
- Directory flushes are Unix-only, matching `raft/storage/fs.rs`; Windows has no equivalent and takes the weaker guarantee rather than failing the write.

## Docs

Durability section in the storage operations guide covering both modes, the platform timings, and when to pick each. The property in the JSON-LD config reference. The environment variable and its precedence in the configuration table. The per-backend write visibility and durability contract in the storage-traits design doc.

Two adjacent doc updates while in those files: the `StorageCas` block in `storage-traits.md` now matches the trait's current `insert` / `compare_and_swap` signatures, and the JSON-LD reference notes that tiered `commitStorage`/`indexStorage` routing is constructed on the S3 path, so a file connection builds a single backend.

`design/performance.md` now states the failure model and storage mode behind the Pokec write comparison.

## Testing

Storage-level coverage: staging cleanup on success and on both create-if-absent outcomes, `list_prefix` excluding staging files, whole-file replacement on overwrite, and atomicity under `page-cache`. Policy coverage pins the default, the derived-content exclusion, the precedence rule, and the accepted spellings, all against pure functions so none read process environment. Config coverage spans both parsers, an absent property staying unset, and rejection of an unknown mode.

Full workspace run at the default setting.

## Follow-ups

- Wiring `commitStorage` for file storage, which would make durability selectable per storage instance rather than by content kind.
- Re-running the Pokec write comparison under the default setting.


fluree-db-core/src/storage/file.rs
```diff
@@ -10,26 +10,172 @@ use crate::{
...
+    name.ends_with(TMP_SUFFIX)
+}
+
+/// Write `bytes` to a staging sibling of `path`, returning the staging path.
+///
+/// Under [`Durability::Sync`] the contents are flushed before returning, so a
+/// caller that then makes the file visible has its bytes on the device first.
+/// The staging file is removed if any step fails, leaving nothing behind for
+/// `list_prefix` or a later reader to find.
+fn stage_bytes(path: &Path, bytes: &[u8], durability: Durability) -> std::io::Result<PathBuf> {
```
aaj3f: **Neither half of this PR has a test that fails when it's removed.** This is the one I'd want addressed before merge, and I verified it two ways rather than reading the tests.

**Mutation A** — I hardcoded `Durability::syncs()` (`storage.rs:33`) to `false`, so nothing anywhere fsyncs. `cargo test -p fluree-db-core --features native storage::file` → **16 passed, 0 failed.**

**Mutation B** — I replaced `write_atomic` entirely with the pre-PR body, `std::fs::write(path, bytes)`: no staging file, no rename, no directory flush. **16 passed, 0 failed** again — including `page_cache_mode_still_writes_atomically` and `compare_and_swap_writes_through_staging`, both of which are named for the property they turn out not to observe.

The mechanism is that every test asserts an *outcome* that an in-place write produces identically. `write_bytes_leaves_no_staging_file` and the three `leftovers.is_empty()` assertions are satisfied by code that never creates a staging file in the first place. `write_bytes_overwrite_replaces_entire_contents` writes 4096 bytes then 5 — but `std::fs::write` truncates, so it passes on the old path too. `list_prefix_skips_staging_files` plants its own `.tmp` file and so genuinely guards the `is_tmp_artifact` skip; `insert_reports_creation_once_and_preserves_the_original` genuinely guards `hard_link` over `rename`. Those two have teeth. The staging/rename and the fsync — the two things the PR exists for — have none.

The practical consequence is that someone can revert either half in six months and CI stays green, and the fsync half is the one preventing acknowledged-then-lost commits. Both are testable without fault injection by asserting the *shape* of the write rather than its outcome. For the rename, inode identity is the cheapest signal:

```rust
/// The destination must never be opened for truncation: a rename replaces
/// the inode, an in-place write reuses it.
#[tokio::test]
async fn write_bytes_lands_via_rename_not_in_place() {
    use std::os::unix::fs::MetadataExt;
    let (_dir, storage) = storage();
    storage.write_bytes("k.json", &vec![b'a'; 4096]).await.unwrap();
    let path = storage.resolve_path("k.json").unwrap();
    let before = std::fs::metadata(&path).unwrap().ino();
    storage.write_bytes("k.json", &vec![b'z'; 4096]).await.unwrap();
    assert_ne!(before, std::fs::metadata(&path).unwrap().ino(),
               "blob was written in place, not staged and renamed");
}
```

For the fsync I don't have as clean a trick — the honest options are a `#[cfg(test)]` counter on `stage_bytes`/`fsync_parent_dir` incremented under `Sync` and asserted zero under `PageCache`, or (nicer, if you want it observable in production too) a metric. Either is fine by me; the requirement is just that *something* goes red when `durability.syncs()` stops being consulted. I'd note this is the same finding I left on #1599 in this same file — which is partly why I think it's worth closing properly here rather than a third time.

fluree-db-core/src/storage/file.rs
```diff
@@ -10,26 +10,172 @@ use crate::{
...
+    Ok(())
+}
+
+/// Stage `bytes` and link them onto `path` only if `path` is absent.
+///
+/// Returns `false` when `path` already exists, leaving it untouched. Uses
+/// `hard_link` rather than `rename` because `rename` would replace an existing
+/// file, and the create-if-absent answer is what callers use to detect a
+/// duplicate ledger.
+fn create_new_atomic(path: &Path, bytes: &[u8], durability: Durability) -> std::io::Result<bool> {
```
aaj3f: *Optional, and more of a question than a suggestion.* `create_new_atomic` swaps `O_CREAT|O_EXCL` for stage + `hard_link`, and the reasoning for `hard_link` over `rename` is right. The thing I'm less sure about is that `O_EXCL` works on essentially anything with a filesystem, while `hard_link` doesn't — some FUSE mounts, exFAT, and a few NFS configurations return `EPERM` or `EOPNOTSUPP`. On those, `insert` (which is how a duplicate ledger gets detected) goes from working to hard-failing. I have no evidence anyone runs Fluree on such a mount, so this may be entirely theoretical — but a fallback to the old `create_new` open when `hard_link` returns `EPERM`/`EOPNOTSUPP` is a few lines and keeps the guarantee everywhere.

fluree-db-core/src/storage/file.rs
```diff
@@ -10,26 +10,172 @@ use crate::{
...
+        return Err(e);
+    }
+    Ok(tmp)
+}
+
+/// Stage `bytes` and rename them onto `path`.
+///
+/// A concurrent reader of `path` observes either the previous contents or the
+/// complete new contents; the final name is never a partially written file.
+fn write_atomic(path: &Path, bytes: &[u8], durability: Durability) -> std::io::Result<()> {
```
aaj3f: *Optional.* A `SIGKILL` between `File::create(&tmp)` and the `rename` still leaves an orphaned `<name>.<pid>.<seq>.tmp`, and nothing in the repo sweeps them. This is a lot less bad than it was on #1599 — `list_prefix` now excludes them so they can't be handed out as a content address, they're inert — but on a crash loop they accumulate on the volume, and a full volume is one of the things that causes the crash loop. A sweep of `*.tmp` older than some threshold at storage construction would close it; `O_TMPFILE` + `linkat` would close it properly on Linux. Entirely reasonable to make this a follow-up issue rather than grow this PR.

fluree-db-core/src/storage/file.rs
```diff
@@ -10,26 +10,172 @@ use crate::{
...
+    let tmp = stage_bytes(path, bytes, durability)?;
+    if let Err(e) = std::fs::rename(&tmp, path) {
```
aaj3f, line 129: *Optional, very minor.* Replacing a file by rename gives it a new inode, so any ownership, mode, ACL or hard link a deployment had applied to a destination path is dropped on the next write. For content-addressed blobs that's meaningless; for the nameservice head refs written through `blocking_locked_write` it's conceivable someone cares. Not worth doing anything about, probably worth a sentence in `docs/operations/storage.md` next to the durability section.

fluree-db-core/src/storage.rs
```diff
@@ -9,6 +9,91 @@
...
+    /// death, but a power loss or kernel panic can lose writes already reported
+    /// as committed.
+    PageCache,
+}
+
+impl Durability {
+    /// Environment variable selecting the default for new `FileStorage`.
+    pub const ENV_VAR: &'static str = "FLUREE_STORAGE_FSYNC";
+
+    pub(crate) fn syncs(&self) -> bool {
```
aaj3f, line 33: *Optional, latent rather than live.* `ContentStore::put_with_id` resolves the address from the `ContentId` and then writes through `write_bytes`, which takes `self.durability` unconditionally — so the `is_derived` classification is bypassed on that path. I traced the production callers (`rebase.rs:595`, `cypher_txn.rs:461`, `commit_worker.rs:533/538/1021`, `cli/sync.rs`) and they are all commit and txn blobs, so today it happens to land on the correct answer. It just isn't *deciding* the correct answer, and the day someone ingests index blobs by CID — a pack-sync path is the obvious candidate — index writes silently start fsyncing. Since the `ContentId` already carries the kind, routing this through `content_write_bytes_with_hash` would make it decide rather than coincide.

Commenting here because `put_with_id` (`fluree-db-core/src/storage.rs:664`) is not in this diff; this is the per-instance setting it consults instead of the per-kind one.

fluree-db-connection/src/storage.rs
```diff
@@ -2,19 +2,22 @@
...
 
-/// Create a file storage instance
+/// Create a file storage instance for `config`.
+///
+/// Durability comes from the config, with `FLUREE_STORAGE_FSYNC` taking
+/// precedence over it — see [`Durability::resolve`].
 #[cfg(all(feature = "native", not(target_arch = "wasm32")))]
-pub fn create_file_storage(base_path: &str) -> FileStorage {
-    FileStorage::new(base_path)
+pub fn create_file_storage(base_path: &str, config: &StorageConfig) -> FileStorage {
```
aaj3f, line 19: *Optional, just flagging the blast radius.* `create_file_storage` gaining a `&StorageConfig` parameter and `StorageConfig` gaining a public field are both source-breaking for anyone constructing them outside this workspace. I checked our own pinned consumer and solo touches neither (one test-only `FileStorage::new`, unchanged), so nothing to do — worth mentioning only so it doesn't surprise us if another embedder turns up.

fluree-db-core/src/content_kind.rs
```diff
@@ -164,6 +164,35 @@ pub enum ContentKind {
...
 
 impl ContentKind {
+    /// Whether this kind is derived from the commit chain and can be rebuilt.
+    ///
+    /// Commits, transactions and the objects describing a ledger are the source
+    /// of truth: losing one loses data. Everything else — index nodes,
+    /// dictionaries, sketches, arenas — is recomputable from those, so storage
+    /// may trade durability for throughput when writing it. See
+    /// `docs/operations/scaling-and-resilience.md`.
+    pub fn is_derived(&self) -> bool {
```
aaj3f, line 174: *Praise.* Putting the split on `ContentKind` behind an exhaustive match, rather than at the call sites, is the thing that makes this feature cheap to keep correct — the indexer and the transactor both get the right durability without knowing the setting exists, and forgetting to classify a new kind is a compile error rather than a silent fsync. Same instinct behind rejecting an unknown `durability` value at config load instead of defaulting it.

fluree-db-core/src/storage/file.rs
```diff
@@ -10,26 +10,172 @@ use crate::{
...
+/// Suffix marking a staging file left by an interrupted atomic write.
+const TMP_SUFFIX: &str = ".tmp";
+
+/// fsync the directory holding `path` so the rename or link that put the file
+/// there survives power loss.
+///
+/// Unix-only: Windows exposes no equivalent, so the call is skipped and the
+/// weaker guarantee accepted rather than failing the write. Mirrors
+/// `fluree-db-consensus/src/raft/storage/fs.rs`.
+fn fsync_parent_dir(path: &Path) -> std::io::Result<()> {
```
aaj3f: *Praise.* `fsync_parent_dir` and the per-process/per-call `tmp_sibling` are precisely the two gaps I left on #1599, closed without being asked. The `with_extension`-mangles-multi-part-extensions note is a good catch too.

aaj3f (approved):
This is carefully built, @bplatz, and I want to say up front what I think is best about it: putting the durable/derived split on `ContentKind` behind an exhaustive match means the indexer and the transactor get the right answer without either of them knowing the setting exists, and a new content kind can't quietly land on the wrong side. The platform timing table is the kind of thing I wish more storage PRs had — the macOS `F_FULLFSYNC` number is exactly why the escape hatch needs to exist, and saying so plainly is better than picking a default and hoping. It also closes both of the gaps I left on #1599 in this same file (the missing directory fsync, and the fixed `.tmp` name that two unsynchronized writers to one address could clobber) without being asked.

What I'd maybe put emphasis on is current coverage, and I checked it rather than inferring it. I ran two mutations against `cargo test -p fluree-db-core --features native storage::file`. First I hardcoded `Durability::syncs()` to `false` so nothing fsyncs anywhere: **all 16 tests pass**. Then I replaced `write_atomic` with the pre-PR `std::fs::write(path, bytes)` — no staging, no rename, no flush: **all 16 pass again**, including `page_cache_mode_still_writes_atomically` and `compare_and_swap_writes_through_staging`. Every assertion is an outcome an in-place write produces identically; only `list_prefix_skips_staging_files` and `insert_reports_creation_once_and_preserves_the_original` have real teeth, and they guard the two smallest pieces. So both halves of this PR can be reverted later with CI green, and the fsync half is the one standing between an acknowledged commit and losing it. I sketched an inode-identity test above for the rename; for the fsync a `#[cfg(test)]` counter is probably the honest option. The specific approach is entirely negotiable — something just has to fail when the flush goes away.

Two smaller things I'd like you to at least think about rather than fix: `create_new_atomic` trades `O_EXCL` — which works on anything — for `hard_link`, which some FUSE/exFAT/NFS mounts refuse, and on those `insert` goes from working to hard-failing; a fallback on `EPERM`/`EOPNOTSUPP` is cheap insurance. And a `SIGKILL` mid-write still leaves an orphan `.tmp` with nothing to sweep it — much less serious now that `list_prefix` excludes them so they can't be served as content addresses, but a crash loop still accumulates them on the volume whose exhaustion causes the crash loop. Happy for that one to be a follow-up issue.

One merge-order note: this rewrites the same three write paths in `fluree-db-core/src/storage/file.rs` that #1599 rewrites, and this version is strictly the better one — it handles the directory fsync, the temp-name collision, and the derived/source-of-truth split, none of which #1599 does. #1599's other half (treating a zero-length blob as absent, and the CAS healing that goes with it) is orthogonal and still worth having. Worth the two of you agreeing which lands first, because a mechanical merge of both will not do the right thing.

- **Patterns/abstractions:** ✔ `Durability` on `StorageConfig` and the classification on `ContentKind`; three separate write paths collapsed onto two shared helpers rather than duplicated; extends the crate's own error types, no `anyhow`, no new `unwrap` on a fallible path.
- **Performance (speed first, memory second):** ✔ Derived content is slightly *faster* (one `spawn_blocking` hop for mkdir+stage+rename instead of two; the `to_vec` is not new, `tokio::fs::write` already `to_owned`s). Source-of-truth adds one file fsync + one dir fsync per write, measured and opt-out-able. No S3 path touched, so no regression surface for the Lambda deployments.
- **Testing:** ✖ 16 new tests all pass with the fsync fully disabled *and* with the stage-and-rename fully reverted. The config/precedence/classification half is properly covered; the storage-mechanism half is not.

Verified locally at branch HEAD (`2b1ed3623`): `cargo test -p fluree-db-core --features native storage::file` → 16/0 with all sixteen confirmed by name; `cargo test -p fluree-db-connection storage_durability` → 5/0; both mutations above green-at-16 and then reverted with `git status` clean; traced `ContentKind` through `EncryptedStorage`, `TieredStorage` and `AddressIdentifierResolverStorage` (all preserve it) and through `put_with_id` (drops it, but every production caller is source-of-truth today); confirmed solo constructs neither `StorageConfig` nor `create_file_storage`.

Everything here should land — I'd just like a test that fails when the flush and the rename go away, and a decision on the `hard_link` fallback, before it does.

PR #1613, opened 2026-08-08
fix(iceberg): warehouse-root resolution for pin resolution and materialize scans

bplatz:
prepare_iceberg_scan's Direct branch read metadata at the raw table_location, so current_snapshot_id and scan_for_materialize_stream failed on catalog-less multi-table warehouse roots (version-hint.text at the root) even though the query path (load_table_context) resolves the per-table directory fine.

This runs the same resolve_direct_table_location step before the metadata read. Single-table Direct locations resolve unchanged, so existing sources are byte-identical; the direct metadata cache is now keyed by the effective per-table location.

Hit in practice by Solo's batch-mastering pin resolution over a catalog-less multi-table copy (three tables under one warehouse root): the pin op failed with 'Failed to read version-hint.text at {root}/metadata/version-hint.text' while SQL queries over the same graph source worked. Verified end-to-end on a deployed stack after this fix.


fluree-db-api/src/graph_source/r2rml.rs
```diff
@@ -883,6 +883,29 @@ impl<'a> FlureeR2rmlProvider<'a> {
...
+                // (`current_snapshot_id`) and materialize scans read
+                // `{root}/metadata/` and fail on any multi-table copy the
+                // query path handles fine. Single-table locations return
+                // unchanged, so existing sources are byte-identical.
+                let lt_key = super::catalog_session::IcebergCatalogSession::load_table_key(
+                    graph_source_id,
+                    &table_id.namespace,
+                    &table_id.table,
+                );
+                let effective_location = self
```
aaj3f, line 911: **The fix ships with no test, and nothing in the repo covers the path it fixes.** I went looking before saying this: there is no test anywhere in `fluree-db-api/tests/` or `src/graph_source/` that drives a warehouse-root location through `prepare_iceberg_scan`. `match_warehouse_table_dir` has good unit tests in `fluree-db-iceberg/src/catalog/direct.rs`, but those cover the string matching, not the call site you're adding — which is the entire bug. So the divergence that produced this bug (one resolution step, two call sites, one of them not calling it) can reappear in exactly the same way and CI won't notice.

The reason I think this is worth insisting on rather than waving through is that this is a *divergence* class of bug, and those recur. The moment somebody adds a third entry point into Direct-mode metadata resolution, we're back here.

The good news is that #1611 makes the test nearly free. Before it, a catalog-less warehouse root needs real S3 (or LocalStack) because `list_dir` only works on `S3IcebergStorage`; after it, `IcebergStorageBackend::File` lists local directories and #1611 commits a real pyiceberg table at `fluree-db-api/tests/fixtures/iceberg/silver/people`. Pointing `table_location` at the *parent* — `file://…/tests/fixtures/iceberg/silver` — with `rr:tableName "silver.people"` makes `leaf_name` (`silver`) mismatch `requested` (`people`), which is precisely the warehouse-root branch, and it exercises both `current_snapshot_id` and `scan_for_materialize_stream`. That's a copy of `it_iceberg_local_fs.rs` with one string changed. Which is really an argument for merge order: land #1611 first, rebase this on top, add that test.

fluree-db-api/src/graph_source/r2rml.rs
```diff
@@ -883,6 +883,29 @@ impl<'a> FlureeR2rmlProvider<'a> {
...
             }
             CatalogConfig::Direct { table_location } => {
+                // Warehouse-root resolution — the same step the query path
+                // (`load_table_context`) performs: when `table_location` is a
+                // catalog-less multi-table ROOT, resolve THIS table's own
+                // directory beneath it. Without this, pin resolution
+                // (`current_snapshot_id`) and materialize scans read
+                // `{root}/metadata/` and fail on any multi-table copy the
+                // query path handles fine. Single-table locations return
+                // unchanged, so existing sources are byte-identical.
```
aaj3f: *Optional, accuracy.* "Single-table Direct locations resolve unchanged, so existing sources are byte-identical" is true under `resolve_direct_table_location`'s own definition of single-table, but that definition is `leaf_name.eq_ignore_ascii_case(requested)` — the location's leaf directory name has to match the mapping's `rr:tableName`. A genuinely single-table location whose directory happens to be named something else (`s3://b/wh/people_v2` against `rr:tableName "silver.people"`) is classified as a warehouse root, LISTs, finds only `data/` and `metadata/`, and hard-errors — where before this change pin resolution worked.

I want to be careful about how much weight that carries, because I chased it down and it mostly evaporates: `load_table_context` has been running this same resolution on `main` all along, so any such deployment's *queries* already fail this way, and there is no configuration that works end-to-end today and breaks after this merge. So I'm not asking for a code change — just for this sentence (and the commit body) to say "locations whose directory is named after the table" rather than "single-table locations", since the current wording is the kind of thing that gets quoted back during an incident.

For what it's worth I checked our own: solo's execution-log source is `s3://{bucket}/system-iceberg/execution_log` against `rr:tableName "system.execution_log"`, so it takes the unchanged branch. No solo regression.

fluree-db-api/src/graph_source/r2rml.rs
```diff
@@ -883,6 +883,29 @@ impl<'a> FlureeR2rmlProvider<'a> {
...
+                    .resolve_direct_table_location(
+                        table_location,
+                        &table_id,
+                        &lt_key,
+                        &iceberg_config,
+                    )
+                    .await?;
+                let table_location = &effective_location;
+
                 let storage: Arc<S3IcebergStorage> = Arc::new(
```
aaj3f: *Optional, very minor.* On the warehouse-root branch, `resolve_direct_table_location` builds a storage client through `direct_session_storage` (to do the LIST), and then this function immediately builds its own `S3IcebergStorage::from_default_chain` here and uses that one — so the session-cached client is constructed and discarded. It only costs anything on a path that used to be a hard error, so it's not worth restructuring for; but if the #1611 rebase has you rewriting this block anyway, pulling the storage out of the session there instead would be free.

fluree-db-api/src/graph_source/r2rml.rs
```diff
@@ -883,6 +883,29 @@ impl<'a> FlureeR2rmlProvider<'a> {
...
+                // `{root}/metadata/` and fail on any multi-table copy the
+                // query path handles fine. Single-table locations return
+                // unchanged, so existing sources are byte-identical.
+                let lt_key = super::catalog_session::IcebergCatalogSession::load_table_key(
+                    graph_source_id,
+                    &table_id.namespace,
+                    &table_id.table,
+                );
+                let effective_location = self
+                    .resolve_direct_table_location(
```
aaj3f, line 912: *Praise.* Calling the existing `self.resolve_direct_table_location` rather than inlining a second copy of the root-detection logic is the right answer to a bug whose whole cause was two call sites disagreeing — and re-keying the `get_direct_metadata_location` cache on the effective location so the two paths share cache entries rather than silently missing each other is a detail that's easy to skip and would have bitten later.

aaj3f (approved):
The diagnosis is good, @bplatz, and I like that the fix is a second call to the *existing* `resolve_direct_table_location` rather than a second copy of it — for a bug whose entire cause was two entry points disagreeing about resolution, adding a third implementation would have been the wrong instinct. Re-keying the direct metadata-location cache on the effective location so the query path and the scan path share entries instead of silently missing each other is a nice catch too. I chased the regression question properly before writing this: `load_table_context` has been running this same resolution on `main` all along, so the only configs that resolve differently after this merge are ones whose queries already fail today, and our own solo Direct source (`s3://{bucket}/system-iceberg/execution_log` against `rr:tableName "system.execution_log"`) takes the unchanged branch. So there's no regression I can construct that's reachable from a working config.

The one big note is that the fix has no test, and I checked that nothing in the repo covers the warehouse-root path through `prepare_iceberg_scan` at all — `match_warehouse_table_dir`'s unit tests cover the string matching, not the call site. That matters more than usual here because this is a divergence bug, and divergence bugs come back: add a third entry point into Direct metadata resolution and we're in the same place. I don't think it's expensive, though, and it's really a merge-order question. Before #1611, a catalog-less warehouse root needs real S3 to LIST; after it, `IcebergStorageBackend::File` lists local directories and there's a committed pyiceberg fixture at `tests/fixtures/iceberg/silver/people`. Point `table_location` at the *parent* directory with `rr:tableName "silver.people"` and you're in the warehouse-root branch — that's `it_iceberg_local_fs.rs` with one string changed, covering both `current_snapshot_id` and `scan_for_materialize_stream`.

Which lines up with the other reason to sequence these: #1611 and this PR conflict textually — `git merge-tree` reports a content conflict in `r2rml.rs`, since you're both rewriting the same `CatalogConfig::Direct` arm of `prepare_iceberg_scan`. One thing to watch when you resolve it: #1611 sets `direct_location = Some(table_location.clone())` from the **raw** location, and after the merge that has to become the **effective** one (which is what `load_table_context` already does at `:2489`). Keep the raw location and #1611's relocation remap will compare the metadata's declared location against the warehouse root instead of the table directory, and fire a bogus remap on every local warehouse root.

- **Patterns/abstractions:** ✔ Extends the single existing resolution method to a second call site rather than duplicating it; matches `load_table_context`'s block exactly, including the cache key.
- **Performance (speed first, memory second):** ✔ One string comparison and an early return on the common single-table path; the LIST is session-cached and only on the path that previously hard-failed. No engine or indexer code touched. No performance-degradation risk. (One nit: the warehouse-root branch builds a session storage client that this function then discards in favour of its own.)
- **Testing:** ✖ No test for the fix, and no existing coverage of the warehouse-root path through `prepare_iceberg_scan`. Cheap to add once #1611's local fixture lands.

fluree-db-api/src/graph_source/r2rml.rs
bplatz: Addressed in 2a652c39cbb019b9dfc6df9e813ed4daefd56f90

fluree-db-api/src/graph_source/r2rml.rs
bplatz: Addressed in 2a652c39cbb019b9dfc6df9e813ed4daefd56f90

fluree-db-api/src/graph_source/r2rml.rs
bplatz: Addressed in 2a652c39cbb019b9dfc6df9e813ed4daefd56f90

PR #1617, opened 2026-08-10
feat: Forward dictionary pack compaction

### LINKED ISSUE #1616: dict: compact forward dictionary packs (16 MiB targets, compact at >8 packs, single-level)
**Origin: #1598.** That PR removes the *symptom* — one mmap per dictionary pack, which hit `vm.max_map_count` and returned `ENOMEM` at 6% of the memory limit with 24 GiB free. It explicitly leaves the growth itself out of scope. This is that growth.

## What grows, and why

Forward dictionary packs are **append-only and never merged**. `update_forward_packs` does `existing_refs.to_vec()` then `extend(new_refs)` (`fluree-db-binary-index/src/dict/incremental.rs:78-79`, `:131-132`); the module doc states it as the design — "All existing packs are reused unchanged" (`:3-7`). A regression test currently locks that in (`fluree-db-indexer/tests/it_incremental_indexing_strategy.rs:162-232`).

Each incremental index build appends **at least one pack per dict stream it touched** — one for strings, one per touched namespace for subjects — because the builder always flushes the trailing partial pack (`if pack_data_est >= target_pack_bytes || is_last`, `fluree-db-binary-index/src/dict/pack_builder.rs:99-101`). A build that adds three new IRIs emits a ~100-byte pack and a permanent routing entry. The current target pack size is 256 MiB; the median pack on a real deployment is **113 bytes**.

Measured on the deployment in #1598 (23 ledgers, continuous ingest): **103,441 packs on disk**, **19,629 in a single ledger's routing table**, 91.3% under 4 KiB. The only thing that ever reduces the count is a **full rebuild**, which re-packs every dictionary from scratch (`fluree-db-indexer/src/build/upload_dicts.rs:312-416`, `:549-611`) — a whole-ledger reindex, and per #1548 a reindex currently breaks the GC chain.

## Cost per pack, beyond the mappings #1598 removed

`build_dictionary_set` runs on **every index-root load** (`fluree-db-binary-index/src/read/binary_index_store.rs:2571`, reached from `apply_index_v2` / `load_and_attach_binary_store`), and `ForwardPackReader::from_pack_refs` walks every pack: `resolve_local_path` + `exists()` + `File::open` + map + full `parse_pack_meta`, O(N) per load (`dict/pack_reader.rs:143-187`). Also:

- Routing entries are inline in the index root (`format/wire_helpers.rs:284-306`), so the root blob grows ~50 B per pack **per root**, and is re-uploaded and re-fetched on every publish.
- Resident `ParsedPackMeta` + `PackHandle` per pack for the life of the store — trivial each, N of them.
- On S3/remote, N packs is up to N separate CAS GETs and N cache files on first load.
- `prewarm` walks packs in order until a byte budget is hit (`pack_reader.rs:254-269`); a long tail of tiny packs spends that budget on per-pack metadata instead of useful pages.
- #1598's size split caps mappings at the count of packs **≥ 64 KiB**. That set also only grows, so a large enough deployment reaches `vm.max_map_count` again — just much later.

## Proposed shape

Single-level tail compaction. **No tiering** — no levels, no fanout, no per-level size ratios.

- **16 MiB pack target**, down from the current 256 MiB (`dict/pack_builder.rs:16`). The point of the smaller target is to bound the work of one compaction event: a merge rewrites at most ~128 MiB, not a multi-GB stream.
- **Compact when a stream accumulates more than 8 un-compacted packs.** The 8 tail packs are merged into 16 MiB-target packs and the routing table is rewritten. Steady state per stream is therefore *(compacted packs, ≈ data size / 16 MiB)* + *(≤ 8 small tail packs)*, which is bounded by data volume rather than by number of builds — the property that is missing today.
- **Metadata to distinguish compacted from tail packs.** `PackBranchEntry` today carries only `{first_id, last_id, pack_cid}` (`format/wire_helpers.rs:33-38`), which is not enough to know what a compaction should target. Adding that marker likely means a **root format bump to FIR7** — TBD, and the main open question below.
- **Keep #1598's mmap thresholds unchanged.** 64 KiB default, `FLUREE_DICT_PACK_MMAP_MIN_BYTES` escape hatch. Compaction does not change the policy, it changes the population: compacted packs are all far above the threshold and get mapped (correctly — demand paging is worth it at 16 MiB), and only the ≤ 8-pack tail per stream lands on the heap.
- **Compacted-away segments must go into GC.** The input packs of a merge become garbage the moment the new root is published, and are the first forward-pack garbage the system has ever produced. They must be written into the garbage manifest and reclaimed by the existing collector (`fluree-db-indexer/src/gc/collector.rs:57-210`), never deleted eagerly — older roots still reference them, and pack CIDs are already enumerated by `IndexRoot::all_cas_ids` (`format/index_root.rs:1132-1143`) and `collect_dict_cids`. This is where the work meets #1548.

### Open question on the trigger

The above reads "8" as *8 un-compacted tail packs*, which is what makes the FIR7 metadata necessary. If it instead means 8 packs total in the stream, then any dictionary past ~128 MiB exceeds the cap by construction and would re-pack the entire stream on every build — unbounded write amplification. Worth settling explicitly before implementation, since it determines whether the metadata bump is needed at all.

## Constraints a compactor has to respect

1. **ID stability.** Merging is a pure re-partition of the same `id → value` map, but only ID-contiguous runs may be merged — `encode_forward_pack` rejects any gap inside a pack (`dict/forward_pack.rs:574-586`), while gaps *between* packs are legal today (see `test_gap_between_packs`, `dict/pack_reader.rs:677-697`). Tail runs are contiguous in the common case, but the merge must verify rather than assume.
2. **A new root.** `DictPackRefs` lives inline in the root, so there is no way to swap packs without publishing a new index root with a rewritten routing table.
3. **Reader invariants:** header range must equal the routing entry, `kind`/`ns_code` must match, ranges strictly increasing and non-overlapping (`pack_reader.rs:334-348`, `:351-385`).
4. Per-pack hard ceiling is 4 GiB (`page_offset` is `u32`, `dict/forward_pack.rs:629-630`) — far above a 16 MiB target, but the merge should not be able to produce a pack that violates it.

## Related defect surfaced while scoping this

`DictPackRefs` is serialized with **`u16` counts and an unchecked cast** (`format/wire_helpers.rs:286`, `:299`):

```rust
buf.extend_from_slice(&(packs.string_fwd_packs.len() as u16).to_le_bytes());
...
buf.extend_from_slice(&(ns_packs.len() as u16).to_le_bytes());
```

At 65,536 packs in one stream the count wraps to 0 and the routing table is **silently truncated on write** — no error, no warning, and the root reads back cleanly, having quietly lost ID ranges. A single observed routing table is already within ~4× of that on a monotonically increasing counter.

Compaction defuses this in practice (16 MiB packs put the cliff around a terabyte per stream) but does not remove it. The cast should become checked regardless; if the FIR7 bump above happens, that is the moment to widen the count to `u32`.

## For reference: the reverse dictionary already has the right shape

Reverse is a CoW B-tree — `update_reverse_tree` routes new entries by key range and **merge-inserts into the affected leaves only**, leaving untouched leaves at their existing CIDs (`dict/incremental.rs:198-228`). Leaf count tracks data size, not build count. (Separately and much smaller: reverse leaves split at `SPLIT_FACTOR = 1.5` and never merge back, so the reverse tree can fragment under heavy deletes.)

## Acceptance

- Pack count per dict stream is bounded by data size, not by number of index builds — a long incremental run converges instead of growing linearly.
- No more than 8 un-compacted packs per stream persist across builds; compacted packs hit the 16 MiB target.
- Superseded input packs land in the garbage manifest and are reclaimed by the existing GC path; no eager deletion while older roots reference them.
- IDs resolve identically across a compaction boundary — a lookup test spanning merged packs, including a range with a pre-existing inter-pack gap.
- The `u16` routing count can no longer silently truncate.
- `it_incremental_indexing_strategy.rs:162-232` is updated to assert the new invariant rather than pure append-only preservation.
- If the format bumps to FIR7: roots written by the previous version still load.

bplatz:
Closes #1616.

## The problem

Every incremental index build appends at least one forward dictionary pack per stream it touched — including a small trailing pack for a handful of new entries — and nothing ever merged them back. The routing table grew once per build, forever.

On the deployment in #1598 that reached **103,441 packs on disk, 19,629 in a single ledger's routing table**, with a median pack of **113 bytes**. #1598 stopped each pack costing an mmap, which cleared the `ENOMEM`-with-memory-free failure. It deliberately left the growth itself out of scope. This is that growth.

## Approach

After appending, the indexer merges qualifying **runs** in each stream: the longest contiguous run that fits a 16 MiB target and whose members are either all small (≤ 64 KiB) or within a 4× size ratio, requiring either 8 packs or a group already at half the target. Merging copies input pages verbatim and rebases their offsets — no dictionary value is decoded, and page-directory bytes are conserved rather than grown.

Four design points worth review attention:

**No compaction state in the root.** FIR6 and `PackBranchEntry` are unchanged. The rule is scale-free — decided from sizes and ID ranges alone — so nothing needs to record which packs were previously compacted. Sizes come from an in-cycle memo, local file metadata, or a 40-byte header range read, never a full fetch.

**Runs, not suffixes.** A merge output is larger than the packs around it. Anchoring candidates at the newest pack means that once a merge lands, every later candidate contains it, it fails the size ratio against its smaller neighbours, and anything behind it is unreachable forever. Simulating the reported 19,629-pack table, suffix-anchored planning stalls after 19 merges with 19,040 packs left; run-based planning drains it to 6.

**Tail first, unconditionally.** New packs land at the tail, so that window is what keeps the table bounded and it must never be starved by backlog repair. Sweeping head-first lets a table longer than the scan budget hide its own tail — with 1,017 mature packs plus appends, the visible window freezes at seven tiny packs and growth becomes unbounded again for exactly the largest dictionaries.

**Two shared budgets.** 64 MiB of merge input and a count of storage operations, both cycle-wide rather than per stream. A byte budget alone does not bound requests: a backlog of tiny packs consumes almost none of it while still costing a probe and a GET per pack, and a per-stream counter would be multiplied by the namespace count. Only reads that actually reach storage are charged, so local deployments are not throttled for network cost they do not pay. Probes and fetches issue concurrently.

Compaction also sweeps streams that received no novelty, rotating coverage by `index_t`, so a namespace that goes quiet — or whose work was cut short when a budget ran out — does not keep its fragmentation while its packs stay loaded with the index.

Packs created *and* consumed inside one cycle (a freshly appended pack that compaction absorbed, or a cascade's own intermediate output) are referenced by no root, so the base-vs-new diff in `set_dict_refs` cannot see them. Those are recorded as garbage explicitly; everything else flows through the existing retention-aware GC.

Also here: the pack target drops from 256 MiB to 16 MiB, which bounds the work of one merge; and the `u16` pack count in the root encoder becomes checked — it was an unchecked cast over an unbounded table, so passing 65,535 packs would have silently truncated the routing table into a root that reads back clean having lost ID ranges.

`FLUREE_DICT_COMPACTION=off` restores the previous append-only behaviour exactly, so this is revertible in a deployment without a rollback.

## Tests

Merge primitive: adjacency, overlap, gap, kind and namespace mismatch, page conservation, header-derived length, and associativity across repeated merges.

Policy: tiny-run collapse, peer comparability, near-target groups merging below full width, oversized groups shrinking, gaps, a pre-seeded 19,629-pack backlog that must drain rather than stall, and convergence over 4,000 cycles (18 packs for 131 MB at 3.7× write amplification).

Driver, over a real content store: tail reachability under a starved scan budget, id survival across a merge, shared-budget charging, and the off-switch as a true no-op.

End to end, over real incremental index cycles: the routing table shrinks, merges consume their inputs, every dictionary object ever written is either referenced by some root or recorded as garbage, every value and subject IRI still resolves, and a namespace that goes quiet stays bounded and readable.

Two of these were verified to fail without their fix — the garbage-accounting test reports two orphaned packs, and the tail test leaves all 50 packs unmerged under head-first ordering.

## For reviewers

**The `u16` guard panics rather than returning `Result`.** `IndexRoot::encode()` is infallible across 18 call sites, and threading a `Result` through `drop.rs`, `gc/collector.rs`, and `root_assembly.rs` would have widened this well past compaction. A panic aborts the in-progress build and leaves the previously published root intact and readable; a truncated routing table is not recoverable. Happy to do the fallible version if preferred.

**Historical backlogs are repaired rather than left to a full reindex.** Normally reindex would be the right answer for pre-existing fragmentation. It currently is not: `rebuild.rs:1137` publishes the rebuilt root with no garbage manifest, so per #1548 a reindex permanently severs the GC chain for that ledger. If #1548 is fixed, this capability stops earning its keep.

**Not validated against real S3 or a real large deployment.** The 16 MiB figure comes from the S3/Lambda measurements in the design discussion, which were not reproduced here; convergence evidence is simulation plus a 12-cycle integration test. The maintenance pass's stream selection and rotation are exercised but not directly asserted, and its repair behaviour (as opposed to its merging, which is covered) would need budget injection the API harness does not expose.

`docs/design/index-format.md` also corrects the forward-dictionary section, which still described the `DLF1` per-leaf model — a format no longer present anywhere in the code.


fluree-db-binary-index/src/format/wire_helpers.rs
```diff
@@ -305,6 +305,23 @@ pub(crate) fn write_dict_pack_refs(buf: &mut Vec<u8>, packs: &DictPackRefs) {
...
 }
 
+/// Pack counts are `u16` on the wire while the routing table itself is
+/// unbounded, so an unchecked cast would wrap and silently drop routing entries
+/// — a root that reads back cleanly having lost ID ranges.
+///
+/// Panicking aborts the in-progress build instead. Roots are immutable and
+/// published last, so the previously published index stays intact and readable;
+/// a failed build is recoverable, a truncated routing table is not.
+fn pack_count_u16(len: usize, what: &str) -> u16 {
```
aaj3f, line 341: **🟠 The `u16` guard is right, but a table that is already over the cap can't get back under it.**

You asked whether the panic should be a `Result` instead. Having traced it, I think the panic-vs-`Result` choice isn't actually the interesting axis — either way the build fails and the previously published root survives, which is the outcome you want. The thing I'd raise instead is that **neither version lets the ledger recover**, and compaction is the mechanism that was supposed to.

`IndexRoot::encode()` runs at the very end of the build (`build/incremental.rs:3979` and `:4032`), *after* compaction has already run and uploaded its merged packs. So for a ledger that arrives at this build with more than 65,535 packs in one stream, the sequence is: compaction runs and shortens the routing table by whatever the cycle budget allowed → `encode()` panics → the root is never published → the next build reads the same `base_root` with the same original table. The merged packs are sitting in the store, but the routing-table progress is discarded on every cycle, forever. And `gc/collector.rs:419` also calls `root.encode()`, so GC — the other way you might dig out — panics on the same table.

The budget is what makes this a real corner rather than a theoretical one. `COMPACTION_MAX_REQUESTS` is 2048 per cycle, shared across all streams, and one merge takes at most `COMPACTION_MAX_WINDOW = 32` packs at ~32 probes + 32 GETs, removing 31 entries. That's a ceiling around 992 packs removed per cycle against a remote store (~1,984 where `resolve_local_path` makes probes free). A table at, say, 70,000 needs five or more cycles to get under the cap, and every one of those cycles panics before it can publish a thing.

I want to be fair about the severity here, because I don't think this blocks: it needs a stream above 65,535 packs, the worst case anyone has actually observed is 19,629, and this PR is precisely what makes reaching 65,535 unlikely. The pre-existing behavior — a silent truncating cast producing a root that reads back clean having lost ID ranges — is clearly worse than a loud failure. So this is strictly an improvement. It's just that the improvement stops one step short of "and here's how you get out of it."

Minimally, this could be solved by letting compaction ignore its budgets for a stream whose routing table is within some margin of the cap — the cost ceiling exists to protect publish latency, and a ledger that cannot publish at all has nothing left to protect:

```rust
// In compact_forward_packs, before the tail loop:
// A table at the u16 wire cap cannot be published at all, so the budget that
// exists to bound publish latency has nothing left to protect. Drain first.
let over_cap = refs.len() > u16::MAX as usize - COMPACTION_CAP_MARGIN;
if over_cap {
    budget.bytes = u64::MAX;
    budget.requests = usize::MAX;
}
```

Entirely reasonable to punt this to a follow-up issue given how far off the trigger is — I'd mostly just want it *tracked* rather than fixed here, since the guard and its escape valve belong together in someone's head.

fluree-db-binary-index/src/dict/incremental.rs
```diff
@@ -144,6 +146,135 @@ pub fn build_incremental_subject_packs_for_ns(
...
+/// `scan_from` skips positions already known not to start a qualifying run.
+/// Replacing a run with its (larger) merge output cannot make an earlier start
+/// qualify — the run would be shorter and its size spread wider — so a caller
+/// merging repeatedly can carry the previous `start` forward and keep the whole
+/// drain linear instead of rescanning the table on every merge.
+pub fn plan_compaction(
+    refs: &[PackBranchEntry],
+    sizes: &[u64],
+    max_output_bytes: u64,
+    scan_from: usize,
```
aaj3f, line 241: **🟡 `scan_from` is never non-zero in production.**

The only production caller, `compact_window` at `build/incremental.rs:4361`, passes `0`. The single non-zero use is the simulation test at `dict/incremental.rs:975`. So the doc comment's argument — "a caller merging repeatedly can carry the previous `start` forward and keep the whole drain linear instead of rescanning the table on every merge" — describes the test rather than the shipped driver, which rescans each window from its start after every merge.

I don't *think* this matters for speed: the windows are `COMPACTION_TAIL_SPAN = 64` and `COMPACTION_HEAD_SPAN = 256`, so the rescan is a few thousand comparisons against a request budget that dominates by orders of magnitude. So this is really an altitude note rather than a perf one. But it seems like it should either be wired up in the driver (carry `plan.start` forward across the `continue` in the head sweep) or the doc comment should say the parameter exists for the drain simulation, so the next person doesn't read a linearity guarantee the system isn't actually getting. `CLAUDE.md`'s guidance on deferred capability points the same way.

While I was in there — I convinced myself the "merging cannot make an earlier start qualify" claim isn't quite universal, fwiw. With sizes `[5 MiB, 1 MiB × 8]`, the run at index 1 merges to 8 MiB, and *then* `[5 MiB, 8 MiB]` qualifies at start 0 (ratio 8/5 is inside 4×, total 13 MiB clears the near-target half of 16 MiB) where it didn't before. Since the driver passes 0 it re-finds this anyway, so nothing is actually lost today — it's only a reason not to wire the optimization up naively later.

fluree-db-binary-index/src/dict/pack_builder.rs
```diff
@@ -13,7 +13,12 @@ use super::forward_pack::{encode_forward_pack, KIND_STRING_FWD, KIND_SUBJECT_FWD
 pub const DEFAULT_TARGET_PAGE_BYTES: usize = 512 * 1024;
 
 /// Default target pack size (bytes). Packs are large immutable CAS objects.
-pub const DEFAULT_TARGET_PACK_BYTES: usize = 256 * 1024 * 1024;
+///
+/// 16 MiB balances object count against the cost of a single cold cache fill:
+/// S3 has no special latency tier here, and a 16 MiB fetch breaks even with
+/// roughly six sequential 512 KiB page misses. It also bounds the work of one
+/// compaction merge, which rewrites at most this many bytes.
+pub const DEFAULT_TARGET_PACK_BYTES: usize = 16 * 1024 * 1024;
```
aaj3f, line 33: **🟡 Worth being explicit that merging raises first-touch fetch cost.**

This is more of a question than a suggestion, and it may well be one you already settled in the design discussion. The read path whole-fetches a pack — `pack_reader.rs:430` calls `cs.get(&cid)`, writes it to the cache, then mmaps it — so there's no ranged page read on lookup. Which means merging a run of 113-byte packs into one 16 MiB pack takes the cost of a *first* cold touch of any single ID in that range from ~113 bytes to as much as 16 MiB.

Total bytes are conserved and round trips go down a lot, so for anything doing real dictionary work this is clearly the better side of the trade, and it's a 16× improvement on the 256 MiB default besides. The case I keep poking at is a sparse one — a cold reader resolving a handful of scattered IDs — where this is straightforwardly more bytes than before. That's the shape Solo has on Lambda, where the disk cache is per-container and evaporates on every cold start.

You're upfront in the PR description that the 16 MiB figure comes from S3/Lambda measurements in the design discussion that weren't reproduced here, so I'm not asking you to reproduce them. I'd just suggest this doc comment name the sparse-lookup direction explicitly alongside the break-even it already cites — right now it reads as though the trade is one-directional, and the next person tuning this constant would benefit from knowing which workload pushes it back down.

fluree-db-binary-index/src/dict/forward_pack.rs
```diff
@@ -705,6 +708,173 @@ fn encode_page_into(entries: &[(u64, &[u8])], buf: &mut Vec<u8>) -> io::Result<(
...
+            .last_id
+            .checked_sub(meta.first_id)
+            .and_then(|span| span.checked_add(1))
+            .ok_or_else(|| {
+                io::Error::new(
+                    io::ErrorKind::InvalidInput,
+                    format!("pack concat: pack {i} has last_id < first_id"),
+                )
+            })?;
+        if covered != declared || meta.directory[0].page_first_id != meta.first_id {
```
aaj3f, line 793: **🔵 This guard has no test.**

I replaced the condition with `if false && (...)` and all five `concat_*` tests still pass. `concat_rejects_gaps_overlaps_and_mismatched_streams` covers the adjacency, kind, ns, ordering and arity rejections, but nothing exercises the `covered != declared || directory[0].page_first_id != meta.first_id` branch.

It only fires on a genuinely corrupt input — I checked, and `encode_forward_pack:577-589` hard-validates contiguity on every write, so no legally written pack can trip it — which is exactly why it's worth keeping and worth pinning. It's the check standing between a header that lies about its span and a merged pack that claims IDs it has no values for, and that's the failure mode that would be hardest to trace back. A hand-mangled header (bump `last_id` by one and re-encode) would cover it in a few lines.

fluree-db-binary-index/src/dict/pack_builder.rs
aaj3f, line 33: **⚪ The off-switch isn't quite an exact revert.**

Small thing: `FLUREE_DICT_COMPACTION=off` disables merging, but `DEFAULT_TARGET_PACK_BYTES` still drops 256 MiB → 16 MiB unconditionally, so an off-switched deployment isn't byte-for-byte the pre-compaction behaviour — a build emitting more than 16 MiB of new dictionary data in one cycle now produces several packs where it previously produced one. Almost certainly irrelevant in practice (pack count is driven by build frequency, not by the target), and arguably the smaller target is the safer default to keep either way. Just noting it since the PR description says "restores the previous append-only behaviour exactly", and the docs entry inherits that claim.

fluree-db-indexer/src/build/incremental.rs
```diff
@@ -3942,6 +4083,404 @@ struct LeafUploadCounts {
...
+            input_bytes = plan.input_bytes,
+            bytes_remaining = budget.bytes,
+            "forward pack compaction deferred: cycle byte budget exhausted"
+        );
+        budget.bytes = 0;
+        return Ok(false);
+    }
+
+    let input_packs = merge_end - merge_start;
+    if budget.take_requests(input_packs) < input_packs {
```
aaj3f: **⚪ The deferral path spends the budget it just declined to use.**

`budget.take_requests(input_packs)` decrements by the granted amount and *then* the `< input_packs` check bails, so the partial grant is consumed without a merge happening. It's harmless — this only occurs when the remaining budget is already smaller than one merge, and both loops break on `exhausted()` right after — but it reads as though it's meant to be a reservation. A `if budget.requests < input_packs { … return Ok(false); }` peek before taking would say what it means.

fluree-db-indexer/tests/it_incremental_indexing_strategy.rs
```diff
@@ -159,6 +159,11 @@ fn incremental_branch_only_fetches_and_rewrites_touched_leaves() {
     assert_eq!(update.branch_cid, expected_branch_cid);
 }
 
+/// The *build* step is append-only: it never rewrites an existing pack.
+///
+/// Bounding the routing table is a separate stage — the indexer compacts the
+/// tail after these packs are uploaded (`compact_forward_pack_tail`), so a
```
aaj3f: **⚪ Stale function name.** The comment you added points at `compact_forward_pack_tail`; the function is `compact_forward_packs`. Trivial, but it's a new comment so it may as well land correct.

fluree-db-indexer/src/build/incremental.rs
```diff
@@ -3942,6 +4083,404 @@ struct LeafUploadCounts {
...
+
+/// Uploads from this cycle that no longer appear in the published routing
+/// tables, and were not inherited from the base root.
+///
+/// `set_dict_refs` records garbage by diffing the base root against the new
+/// one, which cannot see a pack that was created *and* consumed inside a single
+/// cycle — a freshly appended pack that compaction immediately absorbed, or a
+/// cascade's own intermediate output. Those are referenced by no root at all,
+/// so without this they would sit in the store forever.
+fn unreferenced_uploads(
```
aaj3f, line 4409: **👏 `unreferenced_uploads` is the good catch here.**

This is the one I'd have most expected a compaction PR to miss, and it's the one that would have leaked silently forever: a pack created *and* consumed inside a single cycle is in neither the base root nor the published one, so the `set_dict_refs` diff structurally *cannot* see it — referenced by nothing, collectable by nothing. I removed the `add_replaced_cids(transient)` call and the end-to-end test reported exactly two orphans, matching your stated verification. The fact that the integration test asserts the whole-store invariant ("every dictionary object ever written is either referenced by some root or recorded as garbage") rather than just the happy path is what makes it hold up.

fluree-db-binary-index/src/dict/incremental.rs
```diff
@@ -144,6 +146,135 @@ pub fn build_incremental_subject_packs_for_ns(
...
+/// contains it, it fails the size ratio against its smaller neighbours, and any
+/// backlog sitting *behind* it becomes permanently unreachable. Scanning for a
+/// run lets that older fragmentation be repaired in place.
+///
+/// `scan_from` skips positions already known not to start a qualifying run.
+/// Replacing a run with its (larger) merge output cannot make an earlier start
+/// qualify — the run would be shorter and its size spread wider — so a caller
+/// merging repeatedly can carry the previous `start` forward and keep the whole
+/// drain linear instead of rescanning the table on every merge.
+pub fn plan_compaction(
```
aaj3f, line 237: **👏 The runs-not-suffixes reasoning.**

The design note on why anchoring at the newest pack strands everything behind it is genuinely the crux of this feature, and pinning it with a 19,629-pack drain test rather than prose is the right call. Same for the tail-first ordering — I mutated it to head-first and the starved-budget test went red as advertised. Both of those are the kind of thing that would otherwise have quietly regressed in six months.

aaj3f (approved):
## Mutation results (what I ran to check the guards actually have teeth)

| Guard mutated | File | Result |
|---|---|---|
| Adjacency check in `plan_compaction` (`refs[end].first_id == refs[end-1].last_id + 1` → `true`) | `dict/incremental.rs:243` | ✅ **red** — `non_contiguous_packs_are_not_merged` fails |
| Size-comparability in `qualifies_for_merge` (`comparable` → `true`) | `dict/incremental.rs:272` | ✅ **red** — 3 fail, incl. `repeated_merges_converge_on_the_output_target` |
| Tail-first ordering (head sweep moved before tail sweep) | `build/incremental.rs:4286` | ✅ **red** — `tail_is_compacted_even_when_the_scan_budget_cannot_reach_it` fails |
| Transient garbage accounting (`add_replaced_cids(transient)` removed) | `build/incremental.rs:1258` | ✅ **red** — *"2 dictionary objects are referenced by no root and recorded as no garbage"*, exactly matching the author's stated verification |
| Structural `covered != declared` check in `concat_forward_packs` | `dict/forward_pack.rs:793` | ❌ **stays green** — see the note below |

Each mutation was applied, run, and reverted; `git status` confirmed clean afterwards.

**✅ Approve with nits.**

This is carefully built, @bplatz — and the parts I'd have most expected to go wrong are exactly the parts you'd already found. The runs-not-suffixes analysis is the crux of the whole feature and pinning it with an actual 19,629-pack drain rather than prose was the right instinct; `unreferenced_uploads` is a genuinely subtle leak (a pack created *and* consumed inside one cycle is invisible to the `set_dict_refs` diff by construction) and the end-to-end test asserting the whole-store invariant rather than the happy path is what makes it stick. I re-ran your two stated verifications independently and both reproduce: removing the transient accounting reports exactly two orphaned packs, and reordering the sweep head-first leaves the tail unmerged. I also mutated the adjacency check and the size-comparability rule and both go red, so the policy has teeth beyond the ones you called out.

On the question you raised for reviewers about the `u16` guard: I don't think panic-vs-`Result` is the axis that matters, since either way the build fails and the previously published root survives, which is the right outcome. What I'd flag instead is that **neither form lets a ledger that is already over the cap get back under it** — `encode()` runs after compaction, so the panic discards the routing-table progress compaction just made, and the next cycle starts from the same oversized base root. With a 2048-op cycle budget a 70,000-pack table needs several cycles it will never be allowed to finish, and `gc/collector.rs:419` encodes a root too, so GC is closed off as well. I'm explicitly **not** treating that as blocking: it needs a stream past 65,535 where the worst observed is 19,629, this PR is what makes reaching that unlikely, and the behaviour it replaces (a silently truncating cast) was worse. But the guard and an escape valve belong together, so I'd like it tracked — letting compaction ignore its budgets when a table is near the wire cap is a few lines, and a follow-up issue is a perfectly fine home for it.

Two smaller things worth a look before merge: `plan_compaction`'s `scan_from` parameter is only ever passed `0` by the production caller (`build/incremental.rs:4361`), so its documented linearity benefit applies to the drain simulation and not to the shipped driver — worth either wiring up or re-wording. And the `covered != declared` structural check in `concat_forward_packs` is the one guard in this PR with no test behind it; I disabled it and everything stayed green. It can only fire on genuine corruption, which is why I'd keep it, and also why a mangled-header test is cheap insurance.

- **Patterns/abstractions:** ✔ Extends the existing `FPK1` page layout and `PackBranchEntry` routing rather than inventing a parallel structure; superseded packs flow through the existing `set_dict_refs` diff and retention-aware GC. Keeping compaction state *out* of the root via a scale-free rule is what keeps `FIR6` stable, and that's the right call. One small ding for `scan_from` being unused in production.
- **Performance (speed first, memory second):** ✔ No query hot path is touched; the engine only sees a shorter routing table and far fewer mmaps. Write-path cost is bounded cycle-wide by both bytes and request count, and the request budget is the right instrument — a byte budget alone genuinely doesn't bound round trips for 113-byte packs. Memory is bounded by construction at ~32 MiB per merge. The one real trade is that whole-pack fetch makes a sparse cold lookup more expensive after merging; deliberate and documented, worth naming for Lambda.
- **Index-format compatibility:** ✔ `FPK1`, `FIR6`, and `PackBranchEntry` are all unchanged, and `CURRENT_INDEX_VERSION` stays at 2 (`fluree-db-indexer/src/lib.rs:112`, not in the diff). Old readers read compacted artifacts and vice versa. Solo should take this pin **without** bumping `EXPECTED_INDEX_FORMAT_VERSION` — an unnecessary bump would force a fleet-wide reindex for nothing.
- **Testing:** ✔ Unit coverage on both the merge primitive and the policy, driver-level tests over a real content store, and two end-to-end tests over real incremental cycles — correctly wired into `grp_index.rs:6-7`, so they actually run. Four of five central guards go red under mutation. The gap is the structural check noted above, plus the maintenance pass's stream selection and rotation being exercised but not directly asserted, which you already call out.

Verified locally at branch HEAD `7a907aea6`: `cargo nextest run -p fluree-db-binary-index -p fluree-db-indexer --all-features --no-fail-fast` → 729 passed / 2 skipped / 0 failed, with every new test appearing by name; `cargo nextest run -p fluree-db-api --all-features --test grp_index -E 'test(fwd_pack)'` → both end-to-end tests pass by name; `cargo fmt --check` and package-scoped `cargo clippy --all-features --all-targets -- -D warnings` both clean; five mutation checks as tabled above.

Approving so you can merge when ready

fluree-db-binary-index/src/dict/pack_builder.rs
bplatz: Good catch, and it now cuts deeper than when you wrote it — as of 26fd7b7e9eb49fa85607d0b68b05b5afcd5b3dc2 the pack target is **adaptive**, so `off` differs from pre-compaction behaviour in two ways rather than one:

- `DEFAULT_TARGET_PACK_BYTES` is 16 MiB rather than 256 MiB (your original point), and
- the target now doubles for streams past 256 GiB, so a build cuts packs by a size the switch does not control.

Both are build-side, and that is deliberate rather than an oversight: with merging disabled the scaled target is the only thing keeping a very large dictionary's routing table under the u16 cap, so gating it behind the compaction switch would hand `off` a hard wall. The floor means this is invisible below 256 GiB — an ordinary ledger's packs are cut byte-identically either way.

Fixing the PR description accordingly, and saying what `off` does and does not revert instead of dropping the word "exactly".

PR #1618, opened 2026-08-10
fix(bm25): worker fixes

christophediprima:
Five independent fixes to the BM25 maintenance worker. No new behaviour, no new API surface beyond one return type, and nothing here depends on anything else we have open. Each commit stands alone and can be dropped without affecting the others.

Two of them cause an index to silently stop being maintained. The other three are wasted work and misleading logs. Severities below are deliberately specific — I have tried not to overstate any of them.

## 1. Aliases are compared in two different spellings, so some indexes never sync

`5baa3fde8`

A BM25 index created with `ledger: "docs"` never auto-syncs, and nothing says so.

`Fluree::create_full_text_index` stores the source alias exactly as the caller wrote it — `std::slice::from_ref(&config.ledger)` at `fluree-db-api/src/graph_source/bm25.rs:184` — and `Bm25CreateConfig::ledger` is a free-form `String` where a bare `name` is a legal spelling meaning `name:main`. `LedgerCommitPublished` always carries canonical `name:branch`. So the worker files the index under `docs`, every commit event for it says `docs:main`, `graph_sources_for_ledger` misses, and no sync is queued.

Nothing logs, because from the worker's side no index depends on that ledger. A restart does not heal it: the start-up pass seeds the worker from the same raw `dependencies`. What the operator sees is a successful create, a working index, and query results frozen at creation time.

A bare `--ledger` is accepted at both entry points with no validation — the CLI (`cli.rs`, `commands/bm25.rs` passes it straight to `Bm25CreateConfig::new`) and the HTTP route (`routes/bm25.rs`, same) — and the create then *succeeds*, because `FileNameService::lookup` resolves `docs` through `split_ledger_id` to `("docs", "main")`. `Bm25CreateConfig::validate` rejects only an empty ledger; it checks `name` for a `:` but imposes nothing on `ledger`.

Reproduced by lifting upstream's `Bm25WorkerState` map methods verbatim (`bm25_worker.rs:127-206`, dropping only `stats` and the two `debug!` calls, which touch no keys) into a standalone program:

```
watched_ledgers()                     = ["docs"]
graph_sources_for_ledger("docs:main") = []          <-- commit wakes nothing

registered_graph_sources()            = ["search", "search:main"]
graph_sources_for_ledger("docs:main") = ["search", "search:main"]
                                                    <-- one index, two syncs/commit

after unregister_graph_source("search"):
registered_graph_sources()            = ["search:main"]
                                                    <-- untrack by bare name no-ops
```

To be precise about severity: the index is **not** corrupt, and an explicit `fluree bm25 sync` still works — `check_bm25_staleness` resolves the dependency through `nameservice().lookup()`, which normalizes. Only the automatic trigger is dead.

The skew is structural rather than incidental. Every nameservice canonicalizes its own record and lookup keys (`file.rs`, `memory.rs`, `storage_ns.rs`) but stores the `dependencies` payload verbatim — `notifying.rs`, `file.rs`, `memory.rs`, `storage_ns.rs` and the Raft applier in `state_machine.rs` all pass it through unchanged. So the key side is canonical and the payload side is raw, and the worker compares one against the other.

This hazard is already known in the tree. `resolve_source_t` (`fluree-db-cli/src/commands/bm25.rs`, added in 8611fa561) tries the stored alias and then `{alias}:main`, and its doc comment states the reason outright: a bare `name` means `name:main` to Fluree while ledger records are keyed `name:branch`. This applies the same rule at the layer where the mismatch actually drops work.

Normalizing every key also makes two spellings one registration instead of two. That closes a related hazard in the public API: `Bm25WorkerHandle::register_graph_source` files under the caller's string rather than the record's id, and `in_flight_ids` is keyed the same way — so registering both `search` and `search:main` would defeat the one-sync-per-index guard and race two `sync_bm25_index` calls against the same manifest. No in-tree caller does that today; the fix makes it unreachable.

## 2. Every graph-source type gets registered with the BM25 worker

`6edbf0eae`

`registration_skips_other_graph_source_types` (`fluree-db-server/src/lib.rs`) already asserts that Vector, Iceberg and Geo must not be registered with the BM25 worker — but it covers only the start-up path. `vector_worker.rs` applies exactly this filter on the same event, with the comment "Only auto-register Vector graph sources". `bm25_worker`'s `GraphSourceConfigPublished` arm instead destructures with `..`, discarding the `source_type` the event has always carried, and registers whatever was published.

So this is not a new policy — it is the one place in the crate that skips a rule stated twice elsewhere.

What it costs, precisely:

- A **Vector** source lands in the dependency map, and every commit to its source ledger queues a `sync_bm25_index` against it, which fails at the manifest deserialize and repeats for as long as that ledger keeps committing. That path needs the `vector` feature, which is in neither `default` nor `full`.
- **Iceberg** and **R2RML** sources land in the map too, but both publish with no dependencies (`&[]`, `r2rml.rs:507` — "No ledger dependencies for Iceberg graph sources"), so no commit ever queues a sync for them. Their cost is a polluted `registered_graph_sources()` and an inflated `registered_graph_sources` stat.
- **Geo** has no production publish site.

Wasted work and misleading counters, not data loss: the failing sync dies before any write, and the vector index and its record are untouched.

## 3. A lagging event bus discards the events that survived

`d650cd553`

Same fix shape as f8c737132 ("harden raft cache listener lag recovery"), applied to the other consumer of the same bus.

The `recv()` error arm resubscribes on any error. For `RecvError::Lagged` that loses ground: the receiver has already been fast-forwarded to the oldest event still in the ring, so it is positioned to deliver everything that survived the overflow. Resubscribing starts a fresh receiver at the tail instead, discarding that remainder on top of what the channel already dropped — and the `skipped` count, the one number telling an operator the ring is undersized for the commit rate, was folded into a generic message.

The reasoning is f8c737132's own: a dropped event "could be the sole notice a ledger gets, leaving its cached view stale until the next event." Here the equivalent is an index that stays behind until something happens to commit to its source again.

`Closed` was worse: `subscribe()` on a closed bus returns a receiver that yields `Closed` immediately, so the arm spun the loop at full tilt. Since the worker holds an `Arc<Fluree>` and the bus lives on it, `Closed` means teardown in practice — so it exits, draining in-flight syncs first for the same reason the stop path does.

**Deliberately not here:** f8c737132 also added a catch-up sweep on lag, which is the right complement. Sweeping needs a way to enqueue a sync that no commit event asked for, and that primitive belongs with the follow-up rather than in a behaviour-preserving fix. The `skipped` log is what makes the gap visible meanwhile. Happy to pull the sweep forward into this PR if you would rather have it here.

## 4. Re-registering an index leaks reverse-map edges

`b07a3f30b`

`register_graph_source` upserts the forward map but never prunes the reverse one, so re-registering with a changed dependency set leaves every dropped ledger still listing the index. Those ledgers then queue a sync of an index that no longer reads them, and `watched_ledgers()` reports a ledger nothing watches.

This is the weakest of the five and I want to be straight about it: **it is a public-API invariant, not a live in-tree failure.** `register_graph_source` and `register_graph_source_with_deps` are the documented registration entry points — the module doc example uses one, the start-up pass uses the other — and nothing in their names, signatures or docs says "at most once per id". `register` reads as an idempotent upsert, and for the forward map it is. Any embedder or future caller that re-registers with a narrower dependency set silently leaks.

No in-tree caller triggers it today: there is no path that republishes a live BM25 index's config with different dependencies, and drop-then-recreate is clean because the retract arm already prunes. The one non-hypothetical route is a `GraphSourceRetracted` lost to a broadcast overflow — the lossiness this worker's own `biased;` comment describes — after which the recreate's config publish registers without pruning.

The fix also makes the same-dependency case a true no-op and the widened case correct, so it is strictly safer than the status quo for every input. If you would rather not carry an invariant fix with no live trigger, drop this commit; the other four are independent of it.

## 5. The retract log fires for graph sources the worker never held

`3f613b97e`

`GraphSourceRetracted` carries no `source_type`, so the retract arm fires for every retraction in the deployment and logged `Unregistered retracted graph source` at info for all of them. On a deployment with one BM25 index and a dozen other graph sources, nearly every one of those lines is about something that did not happen.

`unregister_graph_source` now returns whether the id had been registered, and the arm logs only a real removal. Returning it from the same lock acquisition that performs the removal is also the only non-racy way for a caller to learn the answer. This matches `MaterializeTrackingWorker::untrack` (`fluree-db-api/src/materialize_worker.rs:178`), which already returns `bool` for exactly this purpose and logs only when a job was removed.

Both methods previously returned `()` and every call site in the tree uses them as statements, so this is source-compatible.

## Testing

`cargo fmt --all -- --check` and `cargo clippy --all --all-features --all-targets -- -D warnings` clean. 11 new unit tests in `bm25_worker.rs`; existing `bm25_auto_sync` integration tests unchanged and passing.

Every new test was mutation-tested: for each one I reverted the specific line it covers and confirmed the test fails. Upstream's existing worker tests spell both map slots `name:branch`, so a branchless dependency had no coverage at all before fix 1.

## Two things found in passing, not in this PR

- **`vector_worker.rs` shares the alias bug.** Its maps are raw-keyed the same way and `create_vector_index` publishes the dependency verbatim (`graph_source/vector.rs:202`). I left it out to keep this diff to one file and out of `#[cfg(feature = "vector")]` CI surface. Happy to send it separately.
- **Under Raft, the worker's graph-source arms may never fire at all.** The Raft state-machine adapter emits only `Ledger*` events (`fluree-db-consensus/src/raft/state_machine_adapter.rs:354-378`); I did not find `GraphSourceConfigPublished` or `GraphSourceRetracted` among them. If that is right, `auto_register` and the retract arm are both dead in Raft mode and an index created after start-up would not auto-sync until a restart. I have not traced this far enough to send a fix and it is orthogonal to the five above — flagging it in case it is news.


fluree-db-api/src/bm25_worker.rs
```diff
@@ -532,13 +583,40 @@ impl Bm25MaintenanceWorker {
...
                             }
                         }
-                        Err(e) => {
-                            // Broadcast channel lagged or closed
-                            warn!(error = %e, "Event channel error, resubscribing");
-                            subscription = self
-                                .fluree
-                                .event_bus()
-                                .subscribe(fluree_db_nameservice::SubscriptionScope::All);
+                        Err(RecvError::Lagged(skipped)) => {
```
aaj3f, line 604: *Blocking (coverage).* **Fix 3 is the only one of the five with no test, and it is also the one that changes the worker's lifecycle.** I reverted each fix in turn and re-ran the module; the other four go red. This one I reverted *completely* — both `RecvError` arms collapsed back to the pre-PR `Err(e) => { warn!(...); resubscribe }` — and all 21 tests in `bm25_worker` stayed green. So nothing in the tree distinguishes the new behavior from the old, and a future refactor could put the resubscribe back without CI noticing.

I don't *think* the `Closed` half needs a test — I traced it and agree with your commit body that it is unreachable in practice (`Fluree::event_bus()` hands back `&Arc<LedgerEventBus>` at `fluree-db-api/src/lib.rs:3310`, the bus owns the `broadcast::Sender` at `fluree-db-nameservice/src/event_bus.rs:25`, and the worker holds the `Arc<Fluree>` for the life of `run()`, so the sender cannot drop underneath it). The `Lagged` half is the one worth pinning, and it *is* reachable: `LedgerEventBus::new` takes a capacity, so a small-capacity bus plus a burst of commits gets you there deterministically. Something along the lines of "overflow the ring, then publish one more commit for a registered index, and assert the worker still queues a sync for it" would fail against the resubscribing version, because that version restarts at the tail and drops the survivors.

Entirely reasonable to push back if you think the `Lagged` path is genuinely untestable at this layer — but then I'd rather see the reasoning in the commit body than have the commit read as covered by the "11 new unit tests" line when none of them touch it.

fluree-db-api/src/bm25_worker.rs
```diff
@@ -145,18 +170,27 @@ impl Bm25WorkerState {
...
+        let graph_source_id = canonical_alias(graph_source_id);
+        let deps_set: HashSet<String> = dependencies.iter().map(|d| canonical_alias(d)).collect();
+
+        // Clear the old reverse edges first. Overwriting the forward entry
+        // alone would leave any ledger dropped from `dependencies` still
+        // pointing here, so its commits would keep queueing a sync of an index
+        // that no longer reads it, and `watched_ledgers` would report a ledger
+        // nobody watches. Reachable through the `auto_register` arm, which
+        // re-registers on every config republish.
+        self.unregister_graph_source(&graph_source_id);
```
aaj3f: *Optional.* Now that `register_graph_source` calls `unregister_graph_source` to prune, a perfectly ordinary config republish emits `debug!("Unregistered graph source from maintenance")` on the way through. It's only `debug`, so this is minor, but if someone is reading logs to work out why an index stopped syncing, "Unregistered" during a re-register is a misleading breadcrumb. Could inline the prune (pull the reverse-edge removal into a small `prune_reverse_edges` helper that doesn't log) if you think it's worth it — happy to punt this to a follow-up.

fluree-db-api/src/bm25_worker.rs
```diff
@@ -685,6 +763,226 @@ mod tests {
         assert_eq!(graph_sources, vec!["gs2:main"]);
     }
 
+    /// `create_bm25_index` stores `Bm25CreateConfig::ledger` verbatim, so a
```
aaj3f, line 784: *Nit, on the PR description's "Testing" section.* It says "11 new unit tests in `bm25_worker.rs`"; the diff adds 10 (8 `#[test]` + 2 `#[tokio::test]`). Not important except that the same paragraph is what says fix 3 is covered.

Commenting here because a PR-description note has no diff line of its own, and this is the block of tests it's about.

fluree-db-api/src/bm25_worker.rs
```diff
@@ -47,6 +51,27 @@ use tracing::{debug, error, info, warn};
 /// executor via `tokio::spawn`.
 type SyncFuture<'a> = Pin<Box<dyn Future<Output = (String, Result<()>)> + Send + 'a>>;
 
+/// Canonicalize a ledger / graph-source alias to `name:branch`.
```
aaj3f, line 54: *Praise.* The `canonical_alias` doc comment is the model for how this kind of fix should be written up: it names both spellings, says where each comes from, and states the consequence. And reusing `normalize_ledger_id` rather than hand-rolling a `{alias}:main` fallback (the way the CLI's `resolve_source_t` does) is the right call — it fixes the skew at the layer where work gets dropped instead of adding a second read-side workaround.

fluree-db-api/src/bm25_worker.rs
```diff
@@ -371,11 +414,15 @@ impl Bm25MaintenanceWorker {
...
-                ..
             } => {
-                // Auto-register graph source if configured
-                if self.config.auto_register {
+                // BM25 only. Vector, R2RML and Iceberg sources have their own
+                // maintenance paths; registering one here would queue a
+                // `sync_bm25_index` against it on every commit to its source
+                // ledger, and every one of those fails. This mirrors the
+                // start-up pass, which already filters on `is_bm25()`.
+                if self.config.auto_register && *source_type == GraphSourceType::Bm25 {
```
aaj3f, line 443: *Praise.* I checked this against both places the rule already lives (`indexes_to_auto_sync` filtering `is_bm25()` at `fluree-db-server/src/lib.rs:106`, and `vector_worker.rs:339-347` doing the mirror image on the same event) and the framing in the description is exactly right: this wasn't new policy, it was the one arm that skipped a rule stated twice elsewhere.

aaj3f (approved):
This is a genuinely good bug report as much as a patch, @christophediprima — the alias-skew writeup (canonical on the key side, verbatim on the payload side, worker comparing one to the other) is the kind of thing that's very easy to look at and not see, and the fix goes in at the right layer rather than adding a second read-side workaround next to the CLI's `resolve_source_t`. I traced all five premises against the tree and they hold: `notifying.rs:353` really does build the event id canonically while passing `dependencies` through verbatim, `bm25.rs:184` really does store `config.ledger` raw, and `Bm25CreateConfig::validate` really does constrain `name` but not `ledger`. I also checked the one thing that worried me — that canonicalizing the *graph-source-id* key would push a spelling into `sync_bm25_index` that can't resolve — and it's fine, because `lookup_graph_source` normalizes through `split_ledger_id` (`fluree-db-nameservice/src/file.rs:1137`), and in practice ids are already canonical since `graph_source_id()` is `format_ledger_id(name, branch)`.

One thing I'd like to request: Fix 3 is the only commit here with no test, and it's also the one that changes the run loop's lifecycle; I reverted each fix in turn and the other four go red, while fix 3 reverted completely leaves all 21 tests green. I'd like to see the `Lagged` arm pinned (a small-capacity bus, a burst that overflows it, then one more commit that must still queue a sync) — the `Closed` arm I'm happy to take on your reasoning, which I verified.

Neither of those is a design objection, so I'm approving rather than blocking — the four tested fixes are correct and I'd rather they land than sit.

- **Patterns/abstractions:** ✔ Extends `normalize_ledger_id`, mirrors `vector_worker`'s type filter and `MaterializeTrackingWorker::untrack`'s `bool` return; invents nothing parallel.
- **Performance (speed first, memory second):** ✔ Event-driven maintenance path, nothing per-flake or per-row. One `String` allocation per `LedgerCommitPublished` in `graph_sources_for_ledger` (`:234`), negligible against the sync it gates; fix 2 strictly reduces work. No performance-degradation risk.
- **Testing:** ⚠️ 10 new unit tests, all wired and all with teeth (four independent mutations, each red) — but fix 3 has none, and reverting it is invisible to CI.

fluree-db-api/src/bm25_worker.rs
bplatz: Folded in (`e5a55fb96`). Extracted `prune_reverse_edges` — the non-logging removal both callers share; the re-register path calls it directly, and `unregister_graph_source` adds the log on top. Map behaviour is unchanged and the three existing re-register tests still cover it.

Pinned by `re_registering_does_not_log_an_unregistration`, which asserts both halves — no "Unregistered" on a republish, **and** a real unregister still logging. A change that just suppressed the message would satisfy only one.

Worth flagging what that test taught me, because it nearly shipped vacuous. It passed under the mutation (`prune_reverse_edges` → `unregister_graph_source`) on a bare parallel `cargo test`, and only bites under `--test-threads=1` or `cargo nextest`: `with_default` is thread-local while tracing's callsite-interest cache is process-global, so a sibling test hitting those callsites with no subscriber installed leaves them cached as uninteresting. CI runs nextest so it is genuinely enforced there, and the trap plus the per-runner mutation results are written into the test's doc comment. Same hazard `it_minmax_fast_path_fired` documents for the integration suite.

fluree-db-api/src/bm25_worker.rs
bplatz: Confirmed rather than taken on trust — I collapsed both `RecvError` arms into a bare no-op log and all 22 tests stayed green, so nothing distinguishes the new lifecycle from the old.

Deferred to #1673 rather than folded in, deliberately. None of the 22 tests drive the run loop — they exercise the state maps directly — so this needs new scaffolding rather than a variation on an existing test, and the four tested fixes include two that stop indexes silently never syncing. Holding those behind test scaffolding seemed the worse trade, which matches your own read in approving.

#1673 carries your test shape verbatim (small-capacity bus, overflow burst, then one more commit that must still queue a sync) plus a second case for the `Closed` arm, since a test that merely asserts the loop exits also catches the spin. It notes the nextest requirement too.

PR #1622, opened 2026-08-10
Add an offline index-version graph diagnostic

zonotope:
Decodes every .fir6 root in a copied roots directory, rebuilds the DAG they form through prev_index, and reports which of three shapes a ledger is in: one broken chain, multiple lineages, or roots never linked in at all. Those have different fixes upstream, which is why the distinction is worth a tool rather than an inspection.

Carried over from #1600, whose fix is superseded by the reclamation work now on main. The diagnostic is not — nothing on main reconstructs the chain offline, and it is what ended the investigation. On a live ledger it reported 226 roots, 15 reachable, 211 orphaned, with fork points 0 and 211 distinct index_t values, which eliminated both the publish-race and built-twice explanations; two roots carried no prev_index at all, at 6739 and 8627, neither of them genesis. Six earlier theories had each been eliminated by measurement before this named the cause directly.

Run against a roots directory copied out of a pod:

    cargo run -p fluree-db-indexer --example root_graph -- <roots-dir> [head-digest]

Diagnostic only — an example target, not built by a plain cargo build, and not shipped behaviour.

Refs #1600, #1548


fluree-db-indexer/examples/root_graph.rs
```diff
@@ -0,0 +1,203 @@
...
+    for (t, d, p) in dangling.iter().take(8) {
+        println!(
+            "    index_t={t} {} -> MISSING {}",
+            &d[..16.min(d.len())],
+            &p[..16.min(p.len())]
+        );
+    }
+
+    // Reachability from the head: what GC can actually see.
+    let head = head_arg.or_else(|| {
```
aaj3f: ### Worth fixing before merge

**`fluree-db-indexer/examples/root_graph.rs:137`** — *blocking-ish (small fix).* The `[head-digest]` argument as the header documents it can't match, and when it doesn't the tool reports the most alarming possible answer with no warning.

The header at `:23-25` tells the operator to pass "the published index head (the nameservice's `index_head_id`)". That value is a `ContentId` (`fluree-db-consensus/src/raft/state_machine.rs:1050`), and everywhere it gets stringified it goes through `Display` — e.g. `fluree-db-cli/src/remote_client.rs:2579` does `cid.to_string()` — which renders **base32-lower CIDv1** (`fluree-db-core/src/content_id.rs:156-161`). But `nodes` is keyed by file stem, i.e. `digest_hex()` (`fluree-db-core/src/storage.rs:1114`). Those two strings never coincide, so `nodes.get(&d)` at `:150` misses on the very first iteration, the walk breaks with `seen` empty, and `:159-165` prints total orphaning.

I ran this against a real single-root `index/roots/` directory to be sure I wasn't inventing it. With no head argument: `reachable from head 23e139c4f5c4503d : 1 of 1 roots (0 UNREACHABLE)`. Same directory, same binary, head passed in CID form: `reachable from head bafkreibd4e44yrh : 0 of 1 roots (1 UNREACHABLE)`. A perfectly healthy ledger reporting 100% orphaned, silently — which is exactly the reading that starts the next week-long investigation.

The cheap fix is to refuse to guess: if the supplied head isn't in `nodes`, say so and stop rather than reporting a walk of length zero.

```rust
if let Some(h) = &head_arg {
    if !nodes.contains_key(h) {
        eprintln!("!! head {h} is not a root in this directory — pass the hex digest (the .fir6 file stem), not the CID form of index_head_id");
        std::process::exit(2);
    }
}
```

Fuller version if we want to be kind to whoever is pasting from `fluree info` at 2am: try `ContentId::from_str(&arg)` first and fall back to its `digest_hex()`, so both forms work.

---

*Addendum, post-drafting:* @bplatz's comment at `:150` asks for the same loud failure from the typo end. The addition this note carries beyond his: it isn't only a mistyped digest that walks into the silent `0 of N` — the value the module header *tells* the operator to pass (`index_head_id`) stringifies as base32 CIDv1 and can structurally never match the hex file stems, so the guard wants either a CID→`digest_hex()` translation on the way in or the header rewritten to say "pass the hex digest file stem." Either closes both his case and mine.

fluree-db-indexer/examples/root_graph.rs
```diff
@@ -0,0 +1,203 @@
...
+            "\nreachable from head {} : {} of {} roots  ({} UNREACHABLE)",
+            &head[..16.min(head.len())],
+            seen.len(),
+            nodes.len(),
+            nodes.len().saturating_sub(seen.len())
+        );
+
+        // How do the unreachable ones distribute over index_t? A contiguous block
+        // means one truncation event; a scatter means repeated forking.
+        let mut unreachable_ts: Vec<i64> = nodes
```
aaj3f: ### Worth fixing before merge

**`fluree-db-indexer/examples/root_graph.rs:169-199`** — *should fix.* The "fork signature" line is computed over unreachable roots only, so it can't actually see the ordinary fork — which matters because #1600's writeup leans on it.

`unreachable_ts` is filtered by `!seen.contains(*d)` at `:171`, and `dupes` at `:187` keeps only `t` values with `n > 1`. In the textbook fork — one parent, two children at the same `index_t`, one published and one abandoned — the winner is reachable and only the loser lands in `unreachable_ts`. That's a count of 1 at that `t`, the `n > 1` filter drops it, and the line prints nothing. So "distinct unreachable index_t: 211 (so 0 share a t with another root)" does **not** rule out competing builds; the thing that actually ruled it out in #1600 was the separate `fork points: 0` line at `:99`, which is sound.

I don't think this changed the #1600 conclusion — `fork points: 0` carried it — but a line labelled "fork signature" that can't fire on a fork is the kind of thing that gets trusted next time. Building the multiplicity map over all of `nodes` rather than the unreachable subset makes the label true, and it's a one-line change to what gets folded.

Two smaller things in the same block, since they're right there: the count in `:189-191` reads `unreachable_ts.len() - distinct.len()`, which is the number of *excess* roots rather than the number of roots sharing a `t` (with three roots at one `t` it says "2 share a t with another root", where the honest answer is 3); and `:88`, `:100`, `:128`, `:187` all `take(n)` without printing an "… and N more", so a truncated preview looks the same as a complete list.

fluree-db-indexer/examples/root_graph.rs
```diff
@@ -0,0 +1,203 @@
...
+    let genesis: Vec<_> = nodes
+        .iter()
+        .filter(|(_, (_, p))| p.is_none())
+        .map(|(d, (t, _))| (*t, d.clone()))
+        .collect();
+    println!(
+        "roots with no prev_index (expect 1 = genesis): {}",
+        genesis.len()
+    );
+    for (t, d) in genesis.iter().take(5) {
```
aaj3f: ### Optional / nits

**`fluree-db-indexer/examples/root_graph.rs:88`** — *optional.* Output isn't reproducible run to run, which is a shame for a tool whose main use is comparing two measurements. `genesis`, `forks`, and `dangling` are all built from `HashMap::iter()`, so *which* five or eight get previewed shuffles between runs; and `:137-142` picks the head with `max_by_key` over the same iterator, so when two roots tie on `index_t` — precisely the fork case the tool exists to detect — the head is chosen arbitrarily. I built a directory of four roots sharing an `index_t` and ran it five times: the head came out `dddd…`, `dddd…`, `bbbb…`, `cccc…`, `cccc…`, and diffing two full runs shows the genesis listing reordered as well. Sorting each preview by `(index_t, digest)` and breaking the head tie on digest would fix both, and it might be worth an explicit note in the output when the max-`t` is tied, since that tie is itself a finding.

fluree-db-indexer/examples/root_graph.rs
```diff
@@ -0,0 +1,203 @@
...
+        }
+        let digest = path
+            .file_stem()
+            .and_then(|s| s.to_str())
+            .unwrap_or_default()
+            .to_string();
+        let bytes = std::fs::read(&path).expect("read root");
+        match IndexRoot::decode(&bytes) {
+            Ok(root) => {
+                let prev = root.prev_index.as_ref().map(|p| p.id.digest_hex());
```
aaj3f: ### Optional / nits

**`fluree-db-indexer/examples/root_graph.rs:55`** — *optional, and probably the highest value-per-line suggestion here.* `BinaryPrevIndexRef` carries `t: i64` next to `id` (`fluree-db-binary-index/src/format/wire_helpers.rs:87-92`), and we throw it away. Two things fall out for free if we keep it: a dangling pointer at `:128-134` could print the *`index_t` of the missing parent* rather than just an opaque digest, which is exactly the "where is the hole" answer shape 1 is asking for; and comparing `prev.t` against the parent's own decoded `index_t` when the parent *is* present is a cheap consistency check on the chain metadata itself — a disagreement there is a fourth shape the tool currently can't see.

fluree-db-indexer/examples/root_graph.rs
```diff
@@ -0,0 +1,203 @@
...
+        eprintln!("usage: root_graph <roots-dir> [head-digest]");
+        std::process::exit(2);
+    });
+    let head_arg = args.next();
+
+    // digest -> (index_t, prev_digest)
+    let mut nodes: HashMap<String, (i64, Option<String>)> = HashMap::new();
+    let mut undecodable = 0usize;
+
+    for entry in std::fs::read_dir(&dir).expect("read roots dir") {
```
aaj3f: ### Optional / nits

**`fluree-db-indexer/examples/root_graph.rs:42-52`** — *optional.* The error handling is asymmetric in the direction I'd expect to bite: a root that fails to decode is counted and survived (`:58-61`), but a root that fails to *read* takes the whole run down via `expect("read root")` at `:52` (same for `read_dir` at `:42` and the entry at `:43`). Given the tool is pointed at a directory copied out of a pod — possibly mid-GC, possibly a partial rsync — one unreadable file killing a 226-root survey seems like the wrong trade. Folding read failures into the same counter (renamed to something like `unreadable`) would keep the run alive and make the count itself diagnostic.

fluree-db-indexer/examples/root_graph.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    // Shape 3: roots with no prev_index. Exactly one (genesis) is expected.
+    let genesis: Vec<_> = nodes
+        .iter()
+        .filter(|(_, (_, p))| p.is_none())
+        .map(|(d, (t, _))| (*t, d.clone()))
+        .collect();
+    println!(
+        "roots with no prev_index (expect 1 = genesis): {}",
```
aaj3f: ### Optional / nits

**`fluree-db-indexer/examples/root_graph.rs:85`** — *nit.* "expect 1 = genesis" isn't right for the healthy steady state. Once GC has legitimately truncated past genesis the correct answer is 0, which is why #1600's writeup had to spend a paragraph explaining that the `0` in the "after" output wasn't a regression. Something like "expect 0 or 1 — 0 once GC has truncated past genesis" would carry that in the output instead of in prose.

fluree-db-indexer/examples/root_graph.rs
```diff
@@ -0,0 +1,203 @@
...
+                eprintln!("  ! undecodable {digest}: {e}");
+            }
+        }
+    }
+
+    println!("roots on disk    : {}", nodes.len());
+    println!("undecodable      : {undecodable}");
+
+    // Which roots are pointed AT by some other root?
+    let mut referenced: HashSet<&String> = HashSet::new();
```
aaj3f: ### Optional / nits

**`fluree-db-indexer/examples/root_graph.rs:69`** — *nit.* `referenced` is inserted into at `:72` and never read. Looks like a leftover from an earlier shape of the analysis.

fluree-db-indexer/examples/root_graph.rs
```diff
@@ -0,0 +1,203 @@
...
+//! Those have completely different fixes, which is why guessing was not good enough.
+//!
+//! Usage — copy a ledger's roots directory out of the pod, then:
+//!     cargo run -p fluree-db-indexer --example root_graph -- <roots-dir> [head-digest]
+//!
+//! `head-digest` is the published index head (the nameservice's `index_head_id`) if
+//! known; supplying it lets the tool report reachability from the real head rather
+//! than guessing the newest by `index_t`.
+
+use fluree_db_binary_index::IndexRoot;
```
aaj3f: ### Optional / nits

**`fluree-db-indexer/examples/root_graph.rs:27`** — *more of a question than a suggestion.* The file's only non-`std` import is `fluree_db_binary_index::IndexRoot` — nothing from `fluree-db-indexer` is used at all. Would `fluree-db-binary-index/examples/` be a better home? The argument for where it is now is thematic (this is a GC/chain question and `gc/` lives here), and that's a real argument. The argument against is that during an incident `cargo run -p fluree-db-indexer --example root_graph` builds core, novelty, spatial, nameservice and the indexer to decode some file headers, possibly on the box that's already unhappy. Entirely reasonable to leave it if you'd rather keep it next to the code it reasons about.

fluree-db-indexer/examples/root_graph.rs
```diff
@@ -0,0 +1,203 @@
+//! Reconstruct a ledger's real index-version graph from its root blobs.
```
aaj3f: ### Praise (worth preserving)

**`fluree-db-indexer/examples/root_graph.rs:1-25`** — the module header is the best part of this PR and the strongest reason to keep the file in the repo. Naming the three shapes *and* what each implies upstream ("GC deletes oldest-first precisely to make this impossible, so finding it means something deletes out of order") is what makes the output actionable six months from now by someone who wasn't in the incident.

fluree-db-indexer/examples/root_graph.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    // Reachability from the head: what GC can actually see.
+    let head = head_arg.or_else(|| {
+        nodes
+            .iter()
+            .max_by_key(|(_, (t, _))| *t)
+            .map(|(d, _)| d.clone())
+    });
+    if let Some(head) = head {
+        // Count only digests that are actually PRESENT on disk. Walking into a
```
aaj3f: ### Praise (worth preserving)

**`fluree-db-indexer/examples/root_graph.rs:144-151`** — counting only digests actually present on disk, with the comment explaining that walking into a missing parent would make `seen` exceed `nodes` and underflow the subtraction, is exactly right, and the cycle guard at `:153-156` is the sort of thing that gets skipped in a throwaway tool and then bites.

fluree-db-indexer/examples/root_graph.rs
```diff
@@ -0,0 +1,203 @@
...
+        );
+    }
+
+    // Shape 1: dangling prev pointers — a root references a parent that is GONE.
+    // Each dangling pointer is a place the chain walk stops.
+    let dangling: Vec<_> = nodes
+        .iter()
+        .filter_map(|(d, (t, p))| {
+            p.as_ref()
+                .filter(|p| !nodes.contains_key(*p))
```
bplatz: Undecodable roots never enter `nodes`, so a corrupt root is reported as `MISSING` here — truncating vs deleting the same mid-chain root gives byte-identical output. That aims shape 1 at GC delete-ordering when the cause is a bad blob; worth tracking undecodable digests separately and gating the shape-1 verdict on `undecodable == 0`.

fluree-db-indexer/examples/root_graph.rs
```diff
@@ -0,0 +1,203 @@
...
+            .map(|(d, _)| d.clone())
+    });
+    if let Some(head) = head {
+        // Count only digests that are actually PRESENT on disk. Walking into a
+        // missing parent (the tail of a GC truncation) and counting it would make
+        // `seen` exceed `nodes` and the "unreachable" subtraction underflow.
+        let mut seen = HashSet::new();
+        let mut cur = Some(head.clone());
+        while let Some(d) = cur {
+            let Some((_, prev)) = nodes.get(&d) else {
```
bplatz: A head digest that isn't on disk breaks on the first iteration and prints `0 of 202 roots (202 UNREACHABLE)` with no warning, so a typo or a CID pasted instead of raw hex is indistinguishable from total corruption. Worth failing loudly when the supplied head isn't in `nodes`.

aaj3f (approved):
Since I first drafted this review, @bplatz's review comments landed, so I double-checked everything below.: 

Brian's `:150` note and my head-digest item below are the same defect approached from opposite ends — his case is the mistyped digest, mine is that the input the header *documents* can never match at all — so please read mine as a +1 with an addendum rather than a second, separate ask. His `:120` point about undecodable roots being reported as `MISSING` is one I hadn't caught and I think he's right, particularly the suggestion to gate the shape-1 verdict on `undecodable == 0`; it composes with my note at `:42` about read failures, since both are cases of the tool folding a distinct cause into its scariest verdict.

To say a bit more about `head-digest` handling at `:137`, which is also Brian's `:150` ask. The header tells the operator to pass the nameservice's `index_head_id`, but that value stringifies as a base32 CIDv1 (`fluree-db-core/src/content_id.rs:156-161`) while `nodes` is keyed by the hex digest file stem (`fluree-db-core/src/storage.rs:1114`), so the documented input misses on the first lookup, the walk ends with `seen` empty, and the tool prints total orphaning. 

I ran it to be sure: same healthy one-root directory reports `1 of 1 roots (0 UNREACHABLE)` with no head argument and `0 of 1 roots (1 UNREACHABLE)` with a CID-form head, silently. A three-line "that head isn't a root in this directory" guard closes it. 

The second note I'd care about is the fork-signature block at `:169-199` — it's computed over unreachable roots only, so an ordinary fork (winner reachable, loser orphaned) contributes exactly one root at that `t` and gets filtered out by `n > 1`; I don't think that changed the #1600 conclusion, since `fork points: 0` is what actually carried it, but a line labelled "fork signature" that can't fire on a fork will get trusted next time. Everything else in the list is small.

One cross-PR note, no action needed here: this tool's end-of-chain rule ("digest not on disk → chain ends") deliberately differs from main's walk, which errors when the *head* can't be read and only ends gracefully on a missing parent after `store.has()` confirms it. Your #1643 is restructuring that walk into `PrevIndexChainWalk` and has an open question about end-of-chain semantics at the head position, so it's probably worth making the two tell the same story once that lands.

- **Patterns/abstractions:** ✔ Follows the repo's existing diagnostic-example convention across four crates; doesn't reinvent the on-main chain walk (which it couldn't call anyway — `pub(crate)`), and doesn't duplicate anything the CLI can do. Only altitude question is whether it belongs in `fluree-db-binary-index/examples/` instead, since that's its sole dependency.
- **Performance (speed first, memory second):** ✔ No performance-degradation risk. Example target only — `cargo build` doesn't build it, nothing links it, no engine or per-flake path is touched, no dependency added. The sole cost is `clippy --all-targets` compiling it in CI.
- **Testing:** ⚠️ No tests, which matches how the repo treats its other diagnostic examples (`concurrent_add_repro.rs` says so in its own header), so I'm not asking for any. Worth knowing that the classification logic is therefore unguarded — if we ever want it pinned, it has to move into a lib module first.
- **Conventions:** ✔ `fmt` and `clippy -D warnings` both clean locally on the example target

Verified locally at branch HEAD: `cargo clippy -p fluree-db-indexer --example root_graph -- -D warnings` → exit 0 with the example target confirmed in the output; `cargo fmt -p fluree-db-indexer -- --check` → exit 0; built the example and ran it against a real `index/roots/` directory (healthy case correct), against the same directory with a CID-form head (false 100%-orphaned reading), and against a synthetic four-root directory with tied `index_t` five times (head selection varied across runs).

zonotope:
@bplatz, @aaj3f  all nine actionable items are in, one commit each, so they're reviewable separately:

| Feedback | Raised at | Commit |
| --- | --- | --- |
| head-digest can't match, and a typo is indistinguishable from corruption | `:137`, `:150` (bplatz) | `9c4957323` |
| undecodable roots reported as `MISSING` | `:120` (bplatz) | `9cfacd6cb` |
| fork signature computed over unreachable roots only | `:169` | `1fc20b26d` |
| output not reproducible run to run | `:88` | `b88e26390` |
| `prev.t` thrown away | `:55` | `e78f8f20c` |
| read failures kill the run | `:42` | `08d6920f3` |
| "expect 1 = genesis" | `:85` | `3235f7a41` |
| write-only `referenced` | `:69` | `a684c5bd6` |
| crate home | `:27` | `5a0cf0af2` |

PR #1624, opened 2026-08-11
preserve fully-retracted history through full index rebuilds

aaj3f:
A full rebuild materializes no row for a predicate whose facts have all been retracted, so leaflet segmentation never opened a partition for it — its transitions got filed under the next predicate's leaflet, where a predicate-keyed reader skips them by `p_const`. The same absence hit fully-retracted subjects in SPOT: their transitions fell outside the leaf's routing keys, so no read ever routed to them. The visible effect was that after a full reindex, time-travel queries over a fully-retracted predicate or subject returned no rows at a historical `t` (or, with `OPTIONAL`, rows with nulls where values existed), while incremental and novelty-only indexes answered the same queries correctly.

The fix emits a zero-row leaflet plus its own history sidecar for each segment that ends without a live row, and grows routing keys to span history that sorts outside a leaflet's live rows. That's the same shape the incremental writer already preserves when a merge empties a leaflet, and the reader already replays it — this just makes the rebuild produce it too.

Those keys are widened on the **leaflet**, not only on the leaf header. `build_leaf_from_group` re-derives the header from the leaflet directory and `update_branch` republishes the branch entry from it, so a bound recorded only on the header is narrowed away by the first incremental that touches the leaf — reintroducing the original symptom for that subject. `merge_and_encode_leaflet` needed the same treatment, since it re-encodes from live rows while carrying history that sorts outside them. Both halves are pinned by round-trip tests through `update_leaf`, and end-to-end by a deleted entity that stays readable at `@t:1` across a later incremental.

On #1603: its `row_count == 0` guard is what lets a later incremental merge into these zero-column-block partitions. It merged to `main` on 2026-08-10, so the dependency is already satisfied here — it binds only for backports into any release line that lacks it. A test pins the seam (reverting the guard turns it red with the exact failure #1603 fixed).

Verification: an 11-shape × 3-index-strategy matrix (the full-rebuild column was wrong in 8 of 11 cells before, all correct after; incremental and novelty pinned as controls), the existing pinned test extended with the plain single-triple variants, history-surface coverage, and a repeat-rebuild test showing an index built by a pre-fix binary is fully repaired by rebuilding again with this fix — the rebuild walks commits from genesis and reuses no leaves, so reindexing is the complete remediation for any already-affected index. Artifact cost measured at +0.69% bytes on a retraction-heavy fixture; rebuild time within noise.

Because nothing repairs an already-published index automatically, the symptom and the reindex remediation are documented for operators at `docs/troubleshooting/historical-results-missing.md`.

### DIFF
--- a/fluree-db-binary-index/src/format/leaf.rs
+++ b/fluree-db-binary-index/src/format/leaf.rs
@@ -82,6 +82,40 @@ pub struct LeafInfo {
 
 // ── Writer ─────────────────────────────────────────────────────────────
 
+/// The value a leaflet holds constant for a given sort order. POST/PSOT
+/// leaflets are predicate-homogeneous and OPST leaflets are type-homogeneous;
+/// SPOT leaflets are unsegmented, so every row shares the one key.
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum SegmentKey {
+    Predicate(u32),
+    OType(u16),
+    Unsegmented,
+}
+
+impl SegmentKey {
+    fn of(order: RunSortOrder, p_id: u32, o_type: u16) -> Self {
+        match order {
+            RunSortOrder::Post | RunSortOrder::Psot => Self::Predicate(p_id),
+            RunSortOrder::Opst => Self::OType(o_type),
+            RunSortOrder::Spot => Self::Unsegmented,
+        }
+    }
+}
+
+/// The routing identity of a history entry, as a record. `t` and `g_id` are
+/// carried for completeness; neither appears in an ordered routing key.
+fn record_from_history(entry: &HistEntryV2) -> RunRecordV2 {
+    RunRecordV2 {
+        s_id: entry.s_id,
+        o_key: entry.o_key,
+        p_id: entry.p_id,
+        t: entry.t,
+        o_i: entry.o_i,
+        o_type: entry.o_type,
+        g_id: 0,
+    }
+}
+
 /// V3 leaf writer — segmentation-aware, columnar, sidecar-producing.
 pub struct LeafWriter {
     order: RunSortOrder,
@@ -179,6 +213,22 @@ impl LeafWriter {
             self.flush_leaf_if_target_reached()?;
         }
 
+        // Anything still pending belongs to a segment that ended without ever
+        // materializing a row — every one of its facts was retracted. Give each
+        // such segment its own zero-row leaflet before this record's leaflet
+        // claims the entries. Must precede the `leaf_first_record` assignment:
+        // those segments sort below this record, so they set the leaf's low
+        // routing bound when they open the leaf.
+        self.flush_history_only_segments(Some(&record))?;
+
+        // This record's transition entries (and those of any preceding
+        // non-materialized facts in the same gap) belong to the leaflet that
+        // holds this record — the current one, post any segment-boundary flush.
+        // Committed before the leaf's routing keys are taken from `record`,
+        // since those entries sort at or below it and so set the leaf's low
+        // bound when they open it.
+        self.commit_all_pending();
+
         // Track first record of the leaf.
         if self.leaf_first_record.is_none() {
             self.leaf_first_record = Some(record);
@@ -194,11 +244,6 @@ impl LeafWriter {
             }
             RunSortOrder::Spot => {}
         }
-
-        // This record's transition entries (and those of any preceding
-        // non-materialized facts in the same gap) belong to the leaflet that
-        // holds this record — the current one, post any segment-boundary flush.
-        self.commit_all_pending();
 
         self.record_buf.push(record);
         self.last_record = Some(record);
@@ -237,6 +282,27 @@ impl LeafWriter {
             self.sidecar_builder.start_leaflet();
             self.current_segment_started = true;
         }
+        // A fact whose final state is absent materializes no row, so nothing
+        // else widens the leaf's routing keys to cover it — and branch-level
+        // leaf selection drops a leaf whose range stops short of the key being
+        // probed, which is how a fully-retracted subject becomes unreadable at
+        // a historical `t`. Entries are committed in build order, so taking the
+        // latest unconditionally keeps the leaf's bounds monotonic.
+        let rec = record_from_history(&entry);
+        debug_assert!(
+            self.last_record.is_none_or(|last| {
+                let (mut prev, mut next) = ([0u8; ORDERED_KEY_V2_SIZE], [0u8; ORDERED_KEY_V2_SIZE]);
+                write_ordered_key_v2(self.order, &last, &mut prev);
+                write_ordered_key_v2(self.order, &rec, &mut next);
+                prev <= next
+            }),
+            "history entries must be committed in build order — an out-of-order \
+             entry would push the leaf's last routing key backwards"
+        );
+        if self.leaf_first_record.is_none() {
+            self.leaf_first_record = Some(rec);
+        }
+        self.last_record = Some(rec);
         self.sidecar_builder.push_entry(entry);
     }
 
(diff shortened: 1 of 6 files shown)


fluree-db-binary-index/src/format/leaf.rs
```diff
@@ -237,9 +281,34 @@ impl LeafWriter {
             self.sidecar_builder.start_leaflet();
             self.current_segment_started = true;
         }
+        self.widen_leaf_bounds(record_from_history(&entry));
```
zonotope: The widened bound here does not survive an incremental update. This is not a regression because this was broken unconditionally before this pr, but this pr only now fixes it for one mechanism and not the other. 

This call grows only the leaf header's keys. When history is committed into an existing leaflet via commit_history_entry — SPOT always, and PSOT/POST when a fully-retracted subject sorts above the last live row — that leaflet's own dir keys still come from encode_leaflet(&records) and do not span the history.

That's fine for the freshly-rebuilt leaf, but the header keys are re-derived on the next incremental touch:

- incremental_leaf.rs:647-654 — build_leaf_from_group takes first_key from group.first() and last_key from group.last(), i.e. from the leaflet dir entries.
- incremental_branch.rs:444-450 — the branch entry is then read back out of that header.
- branch.rs:110 — find_leaves_in_range computes start by partition_point(|e| cmp(&e.last_key, min_key) == Less), so a narrowed last_key skips the leaf again.

So a full rebuild fixes the leaf, and the first update_leaf that touches it silently narrows the bound back and reintroduces the original symptom for that subject. trailing_fully_retracted_subject_is_not_dropped_in_spot asserts leaf.last_key.s_id == SubjectId(2) on the rebuild output only; test_update_into_rebuild_emitted_empty_partition does round-trip through update_leaf, but only for the history-only-leaflet case — whose leaflet keys do span its history, so it's the safe half.

Widening the leaflet's first_key/last_key alongside the leaf's would make the bound re-derivable, and would also make the leaflet_out_of_range prune at binary_cursor.rs:370 sound on its own rather than depending on the needs_history_replay guard.

bplatz (commented):
Traced the writer paths against the format code. Three things I checked and can clear, so nobody re-derives them:

- **`record_from_history`'s `g_id: 0` / `t` are genuinely inert.** `write_ordered_key_v2` writes only `(s_id, p_id, o_type, o_key, o_i)` in every order, and `cmp_v2_*` compares the same five columns. The doc comment is accurate.
- **`last_record` monotonicity holds across the new call paths.** `flush_leaf` resets `leaf_first_record` but not `last_record`, so a leaf made up only of history-only leaflets compares against the *previous* leaf's last record — that's always smaller in build order, so `widen_leaf_bounds` updates correctly. See the inline note.
- **#1603 is on `main` (merged 2026-08-10)**, so the stated dependency is satisfied here; it binds only for backports into any release line lacking it. Worth restating that way in the body.

One thing that isn't in the diff and I think should be: **an operational note.** The failure mode is "your existing indexes silently return empty at historical `t`, and reindexing is the remediation." That's exactly the kind of thing an operator needs to find in `docs/troubleshooting/` or `docs/operations/` rather than in a PR description — especially since nothing triggers the repair automatically. This PR touches no docs at all.

fluree-db-binary-index/src/format/leaf.rs
```diff
@@ -296,18 +365,127 @@ impl LeafWriter {
...
+            self.emit_history_only_leaflet(&group)?;
+        }
+
+        self.pending_history = retained;
+        Ok(())
+    }
+
+    /// Append one zero-row leaflet holding `entries` as its history segment.
+    fn emit_history_only_leaflet(&mut self, entries: &[HistEntryV2]) -> io::Result<()> {
+        debug_assert!(
```
bplatz: This should be a hard `assert!` (or an `io::Error`), not a `debug_assert!`.

The invariant it guards is what keeps the sidecar segment index aligned with the leaflet directory index. If it's ever violated, every leaf written from that point has its history segments off by one — silently, and only visible as wrong history replay, which is precisely the bug class this PR exists to fix. It's also unrepairable except by another full rebuild.

CI runs debug so the assertion fires there, but index artifacts are written by release builds, where this compiles to nothing. The check is two boolean loads on a path that runs once per retracted segment — the cost argument for `debug_assert!` doesn't apply.

fluree-db-binary-index/src/format/leaf.rs
```diff
@@ -237,9 +281,34 @@ impl LeafWriter {
...
+        let cmp = cmp_v2_for_order(self.order);
+        if self
+            .leaf_first_record
+            .is_none_or(|first| cmp(&rec, &first) == std::cmp::Ordering::Less)
+        {
+            self.leaf_first_record = Some(rec);
+        }
+        if self
+            .last_record
+            .is_none_or(|last| cmp(&rec, &last) == std::cmp::Ordering::Greater)
```
bplatz, line 333: Correctness note rather than a defect — I convinced myself this is right, but it took a while, so it's worth making local.

`flush_leaf()` resets `leaf_first_record = None` but leaves `last_record` set. That was previously harmless: a leaf couldn't be flushed without records, and `push_record` overwrites `last_record` unconditionally. Now a leaf *can* consist solely of history-only leaflets, and its `last_key` comes from this widening compared against a value carried over from the previous leaf.

It works because build order is ascending, so the carried-over value is always smaller and the `Greater` branch fires. But that's a non-obvious cross-leaf dependency. Adding `self.last_record = None;` next to the `leaf_first_record = None;` reset in `flush_leaf` would make the invariant local instead of global.

fluree-db-binary-index/src/format/leaf.rs
```diff
@@ -296,18 +365,127 @@ impl LeafWriter {
...
+    /// leaflet restores the same shape the incremental writer preserves when a
+    /// merge empties an existing leaflet (`row_count == 0` with `history_*`
+    /// intact), which the reader already knows how to replay.
+    ///
+    /// `next` is the record about to be buffered, or `None` at `finish`.
+    /// Entries in its segment stay pending so they land in its leaflet.
+    ///
+    /// Callers must have no buffered records and no open sidecar segment, so
+    /// the emitted leaflet and its segment stay index-aligned.
+    fn flush_history_only_segments(&mut self, next: Option<&RunRecordV2>) -> io::Result<()> {
```
zonotope, line 429: This runs on every push_record. In the common case — all pending entries belong to next's segment — it still does mem::take, pushes every entry into a fresh retained, and assigns back (:409-433), then commit_all_pending immediately takes it again. That's a malloc + copy per record added to full-index rebuild.

Entries are sorted and next's segment sorts last, so inspecting pending_history[0] is sufficient for an O(1) early return before the take.

bplatz (approved):

fluree-db-binary-index/src/format/leaf.rs
```diff
@@ -296,18 +365,127 @@ impl LeafWriter {
...
+        // Segment first, so its index matches the leaflet pushed just below.
+        self.sidecar_builder.start_leaflet();
+        for entry in entries {
+            self.sidecar_builder.push_entry(*entry);
+        }
+
+        let homogeneous_o_type = entries
+            .iter()
+            .all(|e| e.o_type == first.o_type)
+            .then_some(first.o_type);
```
zonotope: homogeneous_o_type scans all entries even under OPST, where the match arm discards it.

zonotope (approved):
Optimistically approving, but I've noted some improvements that would be nice to get in too.

fluree-db-binary-index/src/format/leaf.rs
aaj3f: Confirmed, and it was the important one. I traced the chain you gave and it holds exactly: `build_leaf_from_group` takes `first_key`/`last_key` from `group.first()`/`group.last()` dir entries, `update_branch` reads the branch entry back out of that header, and `find_leaves_in_range`'s `partition_point` on `last_key` then skips the leaf. I turned it into a failing test before fixing anything — a SPOT leaf with a trailing fully-retracted subject, round-tripped through `update_leaf`, came back with `last_key.s_id == 100` instead of `200`.

Fixed by widening the leaflet's own keys as you suggested, so the bound is re-derivable. One addition beyond your note: `merge_and_encode_leaflet` drops it the same way on the other side — it re-encodes each chunk from live rows via `encode_leaflet` while attaching history that sorts outside them, so the widening has to be reapplied there too or the very next incremental undoes it again. The `merged.is_empty()` branch already preserved the range via `empty_encoded_leaflet_with_keys`; this makes the still-has-rows branch honour the same contract. Widening stays inside the leaflet's routing interval (carried-forward history was already in it, novelty is sliced to it), so siblings can't overlap.

Coverage now: `test_update_preserves_widened_bound_for_retracted_subject` for the rebuild→`update_leaf` round trip, and `time_travel_fully_retracted_subject_survives_later_incremental` end-to-end — a deleted entity read at `@t:1` after a later incremental. I checked both actually bite by disabling the incremental-side widening: the end-to-end one drops to 0 rows from 2.

fluree-db-binary-index/src/format/leaf.rs
aaj3f: Agreed, promoted to a hard `assert!`. Your framing is the deciding one: CI runs debug so the assertion fires there, but the artifacts that matter are written by release builds where it compiles to nothing, and an off-by-one between the sidecar segment index and the leaflet directory is both silent and unrepairable short of another rebuild — the exact failure mode this PR exists to remove.

I also widened what it guards. The invariant is now `record_buf.is_empty() && !current_segment_started && leaflet_hist_min.is_none()`, since the leaflet-key widening added a third piece of per-leaflet state that has to be clear before a history-only leaflet can be appended. It runs once per retracted segment, so the cost argument doesn't change.

fluree-db-binary-index/src/format/leaf.rs
aaj3f: Taken — `flush_leaf` now resets `last_record = None` next to `leaf_first_record = None`. Your reasoning about why the old code was correct matches mine, and that's precisely the argument I'd rather not have to reconstruct again: it depended on build order being ascending *across* leaves, which is true but is a global property being used to justify a local read.

With the reset the next leaf's bounds derive only from its own contents. A leaf made up solely of history-only leaflets gets its bound from `widen_leaf_bounds` against `None` rather than against a carried-over value, so the `Greater` branch is no longer load-bearing. Comment at the reset says why.

fluree-db-binary-index/src/format/leaf.rs
aaj3f: Fixed. Inspecting `pending_history[0]` is sufficient exactly as you say — entries are sorted and `next`'s segment sorts last, so if the first entry is already in `next`'s segment then every entry is, and there is nothing to emit. That check now returns before the `mem::take`, so the common path on `push_record` costs one comparison instead of a take, a full copy into a fresh `retained`, and an assign-back that `commit_all_pending` immediately undoes.

fluree-db-binary-index/src/format/leaf.rs
aaj3f: Fixed — the scan moved inside the non-OPST match arm, so OPST takes `Some(first.o_type)` directly. It segments by type, so the constant is known without looking at the rest.

aaj3f:
@bplatz — Thanks for clearing those three; I've left them as they are rather than re-deriving. On the third: I've restated the #1603 dependency in the body as satisfied on `main` (merged 2026-08-10) and binding only for backports into a release line that lacks it, rather than as an open constraint on this PR.

On the operational note — agreed, and it was a real gap. Added `docs/troubleshooting/historical-results-missing.md`, linked from the troubleshooting README and `SUMMARY.md`. It covers the symptom (including `OPTIONAL` returning a full result set with unbound values, which is the form least likely to prompt a second look), which data is affected (fully-retracted properties and deleted entities; partial retraction is not), that the commit history is intact and only the index artifact is incomplete, and the reindex remediation with the CLI and admin-API invocations. It states plainly that nothing triggers the repair automatically and that reindexing costs a full index build.

@zonotope — All five are in. The leaflet-keys one turned out to have a second half on the incremental side — details in the reply on `leaf.rs:284`.

PR #1625, opened 2026-08-11
make governance default-allow tri-state so ledger config applies

aaj3f:
`GovernanceOptions.default_allow` was a bare `bool`, so "the caller didn't say" and "the caller said false" were indistinguishable — and `merge_policy_opts` returned the request opts unchanged whenever any policy input was present. The consequence: merely supplying an identity acted as an implicit override back to deny, so an operator who set `f:defaultAllow true` on a ledger never saw it apply to any identity-carrying request.

`default_allow` is now `Option<bool>`: request wins when explicitly set, ledger config fills unset, and the fallback stays fail-closed. Resolution collapses to a concrete bool at one boundary (`effective_default_allow()` in the policy builder) so the `Option` doesn't leak past it.

The same lost-unset flattening existed one scope down in `SourcePolicyOverride`, which was already `Option<bool>` but collapsed to `false` on read. Carrying the tri-state through there was necessary but not sufficient: `SourcePolicyOverride::has_policy()` counts `is_some()` while `has_any_policy_inputs()` counts only `Some(true)`, so a per-source override naming **only** `default_allow: false` was applied as an override but then landed on the config-defaults path, where config's `f:defaultAllow true` overwrote the explicit `false`. `merge_policy_opts` is now precedence-correct rather than path-dependent: config fills only a genuine unset, and an explicit request value always survives. The one thing still entitled to overrule an explicit caller value is override control (`OverrideControl::None` / a non-permitted identity), which continues to force config — the documented truth table is unchanged and still pinned.

One deliberate semantics choice to flag: an explicit `Some(false)` does **not** count as a policy input in `has_any_policy_inputs()`, exactly as the bare bool behaved. Counting it would have flipped anonymous writes against a `f:defaultAllow false` ledger from unenforced to enforced — a much bigger behavior change than this PR intends. The cost is that "anonymous but enforce default-deny" remains inexpressible; happy to revisit that separately if we want it.

The CLI keeps both spellings: `--default-allow` and a new `--no-default-allow` (clap-enforced mutually exclusive), with neither leaving the value unset so the ledger's `f:defaultAllow` governs. `--no-default-allow` is the deny-by-default escape hatch, and the `fluree-default-allow` header now travels for an explicit `false` as well as `true`, so the posture survives over HTTP instead of silently working only embedded.

**Cache-digest residual.** `hash_governance` widens to `None=0 / Some(false)=1 / Some(true)=2`, so it stays injective within a version. Across a rolling upgrade it is not: a pre-upgrade `false` and a post-upgrade `None` both hash to `0`, so a retry spanning the upgrade could hit a stale entry. Not worth a change — the digest's docstring already scopes the contract to "same client, same retry," `GovernanceOptions` reaches no durable ledger or raft state (it lives only on request/queue-envelope types), and the window is one deploy wide.

The two existing fail-closed posture tests pass unchanged, plus new coverage for config reaching identity-only requests, explicit request-level override in both directions, the previously-untestable unset case, the per-source `default_allow: false` shape, CLI tri-state parsing and header emission, and a server-level integration test driving real HTTP with a Bearer identity against an `f:defaultAllow true` ledger. Full workspace suite green.

Note on the `fluree-default-allow` header: truthiness is byte-for-byte the shared convention every boolean Fluree header uses (`true`, `1`, or empty), so a bare `fluree-default-allow:` reads as `Some(true)`; only a present, non-empty, unrecognized value reads as explicit `false`.


fluree-db-cli/src/cli.rs
```diff
@@ -138,7 +138,10 @@ impl PolicyArgs {
             },
             policy: self.resolve_policy()?,
             policy_values: self.resolve_policy_values()?,
-            default_allow: self.default_allow,
+            // `--default-allow` is a presence flag: absent stays unset so the
+            // ledger's `f:defaultAllow` can apply. There is no CLI spelling for
+            // an explicit `false`.
+            default_allow: self.default_allow.then_some(true),
```
bplatz: This changes CLI behavior and removes the escape hatch in the same stroke. Against a ledger with `f:defaultAllow true`, `fluree query --identity did:...` used to return nothing and now returns everything (the fix) — but there's no longer any spelling for the old deny-by-default posture, since `remote_client.rs` also only sends the header when true. Worth adding `--default-allow[=true|false]` or a `--no-default-allow` counterpart.

fluree-db-api/src/dataset.rs
```diff
@@ -318,7 +318,9 @@ impl SourcePolicyOverride {
             policy_class: self.policy_class.clone(),
             policy: self.policy.clone(),
             policy_values: self.policy_values.clone(),
-            default_allow: self.default_allow.unwrap_or(false),
+            // Already tri-state here; it used to collapse to `false` at this
+            // boundary, which is the same lost-unset bug one scope down.
+            default_allow: self.default_allow,
```
bplatz, line 334: The flattening is fixed here, but the asymmetry one level up defeats it for one shape. `SourcePolicyOverride::has_policy()` counts `default_allow.is_some()`, while `GovernanceOptions::has_any_policy_inputs()` deliberately counts only `Some(true)`.

So a per-source override setting **only** `default_allow: false` passes `has_policy()` -> `wrap_policy` -> `merge_policy_opts` -> `has_any_policy_inputs()` is false -> falls to the config-defaults path -> config's `f:defaultAllow true` overwrites the explicit `false`.

Pre-existing (the old `.unwrap_or(false)` reached the same place), but the PR body claims this scope was fixed too. Either align `has_policy()` with `has_any_policy_inputs()`, or make the config path respect an explicit request value.

bplatz (commented):
Read through the whole change; the core is sound. Two things I went looking for and can clear:

- **Cache digest is handled correctly.** `hash_governance` widens to `None=0 / Some(false)=1 / Some(true)=2`, so it stays injective within a version. Residual: old-`false` and new-`None` both hash to `0`, so a retry spanning a rolling upgrade could hit a stale entry. The docstring already scopes the contract to "same client, same retry", so I'd just note it in the PR body rather than change anything.
- **No path silently loosens.** The only widening is config -> unset-request, which is the intended fix.

Four inline comments. Two more that don't anchor to a line:

**No server-level test for the changed seam.** New coverage is API-layer plus header unit tests, but nothing drives an actual HTTP request with `fluree-identity` against an `f:defaultAllow true` ledger. The header field changed shape and `inject_into_opts` was restructured, so that seam is worth one integration test — #1627 adds exactly that harness in `policy_integration.rs`.

**The `f:defaultAllow false` + anonymous hole gets sharper, not softer.** After this lands, half of `f:defaultAllow` works as operators expect (`true` opens identity requests) and half still silently does nothing (`false` doesn't close anonymous ones). Someone who verifies the first will reasonably assume the second. The pinned test documents it and #1627's `setting-groups.md` rewrite explains it — I'm flagging it as a decision to make deliberately, not a change request here.

**Merge note:** this and #1627 both rewrite the same doc comment on `PolicyDefaults::default_allow` (`fluree-db-core/src/ledger_config.rs:73-77`) — a direct textual conflict. They also both touch `extract/headers.rs` and `consensus/caching.rs` at different hunks. Whoever lands second should keep #1627's version of that doc comment; it's strictly more informative and already describes the post-merge semantics.

fluree-db-server/src/extract/headers.rs
```diff
@@ -459,6 +464,15 @@ fn is_header_truthy(headers: &HeaderMap, name: &str) -> bool {
     }
 }
 
+/// Tri-state read of a boolean header: `None` when absent, `Some(truthy)` when
+/// present. Used where "the caller didn't say" must stay distinct from "the
+/// caller said false" — anything present but unrecognized reads as `Some(false)`,
+/// which is the fail-closed direction for a security-relevant header.
+fn header_bool_opt(headers: &HeaderMap, name: &str) -> Option<bool> {
+    get_header_str(headers, name)
+        .map(|v| v.eq_ignore_ascii_case("true") || v == "1" || v.is_empty())
```
bplatz, line 479: The doc comment above contradicts this line: it says "anything present but unrecognized reads as `Some(false)`, which is the fail-closed direction", but `v.is_empty()` is in the truthy disjunction, so a present-but-empty `fluree-default-allow:` yields `Some(true)`.

Behavior is unchanged from `is_header_truthy`, so this is a comment fix rather than a code fix — but it's a comment about a security-relevant header, which is exactly what a future reader will take at face value.

fluree-db-api/src/policy_view.rs
```diff
@@ -542,7 +542,7 @@ pub async fn wrap_identity_policy_view<'a>(
 ) -> Result<PolicyWrappedView<'a>> {
     let opts = GovernanceOptions {
         identity: Some(identity_iri.to_string()),
-        default_allow,
+        default_allow: Some(default_allow),
```
bplatz: Hard-wiring `Some(..)` means this helper can never inherit the ledger's `f:defaultAllow` — the one public entry point that takes an identity is the one that opts out of the fix. Only a test calls it today, but it's `pub` and re-exported from `lib.rs`. Either take `impl Into<Option<bool>>`, or document it as deliberately an explicit-override helper.

bplatz (approved):

fluree-db-cli/src/cli.rs
aaj3f: Good catch, and you're right that this was the worst of the four — the other notes are about precision, this one actually took a capability away. Fixed: there's now a `--no-default-allow` counterpart, clap-enforced mutually exclusive with `--default-allow`, so the three states are `--default-allow` → `Some(true)`, `--no-default-allow` → `Some(false)`, neither → unset and the ledger's `f:defaultAllow` governs. `fluree query --as did:... --no-default-allow` gets the old deny-by-default posture back explicitly.

On the spelling — I went with the `--no-` counterpart rather than `--default-allow[=true|false]` mostly because it's what the CLI already does: `--no-color`, `--no-user`, `--no-conflicts`, `--no-vended-credentials` are all in `cli.rs` today, and the `num_args = 0..=1` shape only ever shows up for optional *values* (paths, aliases), never for a boolean. So `--default-allow=false` would have been a novel arg shape for us. Happy to switch if you'd rather have the single-flag form though — it's a small change and I don't feel strongly.

You were also right that `remote_client.rs` was half the bug and not just a detail: I'd have fixed the flag and still shipped a CLI where the posture silently evaporated over HTTP while working embedded. Both the header builder (`policy_headers`) and the JSON-LD body-opts path (`inject_policy_into_json_opts`) now send the explicit `false` as well as `true`, and there are tests on both so that seam can't quietly rot again.

fluree-db-api/src/dataset.rs
aaj3f: This is the one I'm most glad you found — I'd convinced myself the `SourcePolicyOverride` scope was done and it wasn't, and the PR body said so, which is worse. Your trace is exactly right: `has_policy()` counts `is_some()`, `has_any_policy_inputs()` counts only `Some(true)`, so an override naming **only** `default_allow: false` gets applied as an override and then lands on the config-defaults path, where config's `f:defaultAllow true` overwrites the explicit `false`.

I took your second option — making the config path respect an explicit request value — rather than aligning `has_policy()` with `has_any_policy_inputs()`. Aligning the two predicates would have fixed this shape by making the override stop being an override at all, which trades a wrong value for a silently ignored one, and it would also have pulled `Some(false)` into "counts as a policy input" through the back door, which is the semantics change I deliberately avoided elsewhere in this PR. So `merge_policy_opts` is now precedence-correct instead of path-dependent: config fills only a genuine unset, and an explicit request value always survives. Override control is still the one thing entitled to overrule an explicit caller value, so `OverrideControl::None` and the non-permitted-identity case continue to force config, and the truth table is unchanged and still pinned.

Worth flagging that this made me delete one of my own tests: I'd written `config_overwrites_explicit_false_on_the_no_policy_input_path`, which pinned precisely the behavior you're calling a bug. It's now `explicit_false_survives_the_no_policy_input_path`, plus `per_source_override_of_only_default_allow_false_survives_config` driving the real `SourcePolicyOverride::to_query_connection_options()` shape end-to-end, and `override_control_none_still_forces_config_over_explicit_request` so the exception stays honest. And you're right that it was pre-existing — the old `.unwrap_or(false)` reached the same place — but the PR body claimed the scope was fixed, so I've corrected that claim rather than leave it overstated.

fluree-db-server/src/extract/headers.rs
aaj3f: You're right, and it's a comment fix rather than a code fix — but I want to say why I landed there rather than just asserting it, since "make the empty case fail closed" was genuinely tempting for a security-relevant header.

The empty-is-truthy convention isn't local to this function: `is_header_truthy` treats `true` / `1` / empty identically, and that's how every boolean Fluree header behaves — a bare `fluree-track-meta:` means "on". If I made empty read as `Some(false)` only for `fluree-default-allow`, then the same header value would mean opposite things depending on which header it was attached to, which seems like a worse footgun than the one we're fixing, and it'd be invisible at the call site. So the behavior stays shared and the comment now says what the code actually does: `true` / `1` / empty are truthy, a bare `fluree-default-allow:` is `Some(true)`, and only a present, non-empty, unrecognized value (`0`, `no`, `nonsense`) reads as explicit `false`.

Also worth noting the test was already right where the prose was wrong — `default_allow_header_is_tri_state` asserts `Some("")` → `Some(true)` — so this was only ever a comment that would have misled a reader, which is more or less exactly your point.

fluree-db-api/src/policy_view.rs
aaj3f: Fair, and the framing that got me was "the one public entry point that takes an identity is the one that opts out of the fix" — that's a real inconsistency regardless of how few callers it has today. Took the `impl Into<Option<bool>>` option, so `wrap_identity_policy_view(&ledger, iri, false)` still compiles for the explicit case and `wrap_identity_policy_view(&ledger, iri, None)` now defers to the ledger's `f:defaultAllow`. Doc comment and the example both updated to describe it as tri-state.

aaj3f:
@bplatz — the four review-body notes, in order:

**On no server-level test for the changed seam:** ## Review body — no server-level test for the changed seam

Agreed, and this was the right thing to push on: `FlureeHeaders`, the header→opts injection, and `force_query_auth_opts` all sit between the wire and `merge_policy_opts`, and none of my coverage crossed that gap. Added two tests to `policy_integration.rs`: one creating a ledger with three documents, zero policies, and a config graph declaring `f:defaultAllow true`, then querying over real HTTP with a Bearer identity token and no `default-allow` anywhere in the request — it now sees all three documents; and one asserting the explicit `false` still wins, through body opts *and* through the `fluree-default-allow` header with nothing in the body, which is the `inject_into_opts` restructure you named.

I did check the first one isn't vacuous, since "identity-carrying request sees everything" and "request wasn't enforced at all" would look identical from the outside: with the pre-fix precedence temporarily restored it returns `[]` rather than three rows, which also confirms the identity is genuinely reaching the policy path.

**On the `f:defaultAllow false` + anonymous hole:** ## Review body — the `f:defaultAllow false` + anonymous hole

I think your framing of this is sharper than mine was, and it's the part of this that I'd most like a decision on rather than a patch. "Half of `f:defaultAllow` works as operators expect and half still silently does nothing, and someone who verifies the first will reasonably assume the second" is exactly the failure mode — this PR arguably makes the trap *more* dangerous by making the `true` half trustworthy.

Mechanically it's that an anonymous request never builds a `PolicyContext` at all (`connection.rs:1080` gates on `has_any_policy_inputs`), so nothing in the config is ever consulted, not just `f:defaultAllow`. Leaving it as-is here deliberately: closing it means anonymous reads against any config-bearing ledger start getting enforced, which is a much bigger blast radius than this PR, and I'd rather that be its own change with its own discussion than a rider on a type change. Pinned as current behavior in `anonymous_request_unenforced_despite_config_default_allow_false` so at least it's documented in code rather than folklore, and it's on the list to decide.

**On cache digest residual:** ## Review body — cache digest residual

Thanks for chasing this one down independently — I'd looked at injectivity within a version and hadn't thought about the rolling-upgrade window. Noted in the PR body rather than changed, as you suggested: old-`false` and new-`None` both hash to `0`, so a retry spanning the upgrade could hit a stale entry, but the docstring already scopes the contract to "same client, same retry", the window is one deploy wide, and `GovernanceOptions` never reaches durable ledger or raft state — it only rides request and queue-envelope types, so there's nothing persisted to migrate.

**On merge conflict with #1627:** ## Review body — merge conflict with #1627

Agreed on all counts, including which version wins: whoever lands second should keep #1627's `PolicyDefaults::default_allow` doc comment, since it describes the post-merge semantics and mine was only trying to correct the old "system default (true)" claim in passing. I'll leave mine in place here rather than pre-emptively reverting it (this PR would otherwise carry a comment that contradicts its own behavior if it lands first), and resolve in #1627's favor at merge. Same for `extract/headers.rs` and `consensus/caching.rs` — different hunks, so those should merge clean, but worth a second look on whichever lands second.

aaj3f:
Follow-up on the anonymous + `f:defaultAllow false` note above: tracked as #1640.

One thing that came out of writing it up, since it sharpens your framing rather than just restating it — I'd assumed the write path behaved the same way as reads, and it doesn't. Probing over HTTP: with `f:defaultAllow false` **alone** an anonymous write succeeds, but with `f:defaultAllow false` + `f:policyClass` it's denied 400, while anonymous *reads* return everything under both. The two paths gate in opposite order — reads test the raw request opts before `merge_policy_opts` runs (`query/connection.rs:1080`, merge happens inside `wrap_policy`), writes merge first and gate after (`policy_view.rs:369` → `:439`). So the config shape our docs use in their examples is one where a denied write is exactly the positive control someone would use to convince themselves the config took effect, while every anonymous read stays open.

Which is to say your "gets sharper, not softer" read was right, and then some.

Issue #51, opened 2021-04-11
Publish benchmarks


alexkreidler:
I noticed [this file](https://github.com/fluree/db/blob/main/src/fluree/db/performance.clj) that seems to run some benchmarks.

It would be great if you could publish these; it would help users understand the scale and type of the workloads you run.

I understand the concern that internal benchmarks might not be representative or useful for comparison with other databases. In that case, I would encourage you to run some standardized benchmarks, like the common SPARQL ones: [BSBM](http://wifo5-03.informatik.uni-mannheim.de/bizer/berlinsparqlbenchmark/), [LUBM](http://swat.cse.lehigh.edu/projects/lubm/) and others from the [Linked Data Benchmark Council (LDBC)](http://ldbcouncil.org/).

This is a very cool project, and I think many people from the semantic web community or using existing graph databases would be interested in trying it out if it were more easily compared with other databases.

Thanks!

bplatz:
Hi @alexkreidler thanks for the note. We will consider publishing these benchmarks and create an internal ticket to track to do so. Fluree does well more than a Semantic graph DB, i.e. SmartFunctions and embedded permissions (which would require the addition of an app server for a standard graph db), provable provenance, time travel... so we need to think through this a bit and where we want to draw comparisons.

PR #1626, opened 2026-08-11
seed connected values through type-anchored stars

aaj3f:
`SELECT ?s ?n WHERE { VALUES ?s { <iri> } ?s a :Thing ; :name ?n }` was O(class size) rather than O(VALUES rows): the values-after-triples deferral fired on any VALUES preceding a 2+-triple star, including one whose variable *is* the star's subject. That left `?s` unbound at planning time, the bound-object `rdf:type` triple won the driver race, and the whole class extent was drained before the deferred VALUES discarded all but the matching rows. Measured in-process at 1.5ms/17ms/47ms for N=2K/20K/50K against a flat ~0.02ms for the equivalent inlined-IRI query.

The gate now checks whether the VALUES shares a variable with the star. Disconnected VALUES (the case the deferral was written for — a pinned vector-search query vector, where seeding would produce a cartesian seed) keep the deferred plan unchanged; connected VALUES seed first, and `build_scan_or_join` picks a bound-driven lane from real cardinality estimates. No cardinality threshold was added: an arity sweep to k=32,768 against a 50,000-member class found no crossover where the extent scan wins — the old plan pays the full scan *and* the post-filter, so it never comes out ahead.

Plan-shape regression tests pin both directions (the existing disconnected-VALUES vector test passes unchanged), plus result-correctness tests on indexed ledgers. Sequencing note: this should land after #1624 — the seeded plan routes some historical-`t` reads through lanes that #1624 makes correct on rebuilt indexes.

### DIFF
--- a/fluree-db-query/src/execute/where_plan.rs
+++ b/fluree-db-query/src/execute/where_plan.rs
@@ -1044,18 +1044,26 @@ fn apply_values(
     operator
 }
 
-/// Does any VALUES pattern bind a variable that the triples also mention?
+/// Split a block's VALUES into those that can seed a same-subject star and the
+/// rest, which must stay behind the triples.
 ///
-/// Answers whether the VALUES can seed the triples. A connected VALUES constrains
-/// them; a disconnected one only multiplies rows.
-fn values_feed_triples(values: &[ValuesPattern], triples: &[TriplePattern]) -> bool {
-    let triple_vars: HashSet<VarId> = triples
-        .iter()
-        .flat_map(TriplePattern::referenced_vars)
-        .collect();
+/// A VALUES binding the star's subject constrains every triple in it at once, so
+/// seeding hands the star a bound driving set. Anything else — a VALUES binding
+/// only a leaf object var, or sharing no variable with the star at all — would
+/// seed a cartesian product instead, so it defers.
+///
+/// The caller only reaches this for a star (`is_property_join`), which guarantees
+/// every triple shares the first triple's subject var.
+fn split_values_seeding_star(
+    values: Vec<ValuesPattern>,
+    triples: &[TriplePattern],
+) -> (Vec<ValuesPattern>, Vec<ValuesPattern>) {
+    let Some(subject) = triples.first().and_then(|tp| tp.s.as_var()) else {
+        return (Vec::new(), values);
+    };
     values
-        .iter()
-        .any(|vp| vp.vars.iter().any(|v| triple_vars.contains(v)))
+        .into_iter()
+        .partition(|vp| vp.vars.contains(&subject))
 }
 
 /// Partition filters into those eligible for inline evaluation and those still waiting.
@@ -1946,30 +1954,34 @@ pub fn build_where_operators_seeded_with_needed(
                     // a pure star block. Wrapping VALUES first seeds the schema/operator and
                     // prevents `build_triple_operators()` from taking the property-join path.
                     //
-                    // Only defer a VALUES that is *disconnected* from the star (e.g. a pinned
-                    // vector-search query vector): seeding it would build a cartesian product
-                    // and cost a profitable fusion. A VALUES that shares a variable with the
-                    // star is a guaranteed producer with exact cardinality, and seeding it
-                    // lets `build_scan_or_join()` pick a bound-driven lane — batched-subject
-                    // -exists, MembershipJoin, or HashJoin — instead of draining the whole
-                    // class or predicate extent and filtering afterwards.
-                    let values_after_triples = operator.is_none()
+                    // A VALUES binding the star's SUBJECT is the exception: it constrains
+                    // every triple at once, so seeding it lets `build_scan_or_join()` pick a
+                    // bound-driven lane — batched-subject-exists, MembershipJoin, or HashJoin
+                    // — instead of draining the whole class or predicate extent and filtering
+                    // afterwards. Every other VALUES still defers, including one binding only
+                    // a leaf object var: seeding that costs the fusion for the whole star
+                    // without constraining the driving scan.
+                    //
+                    // The split is per-VALUES because the hazard is. A block may legally hold
+                    // both kinds, and seeding a disconnected VALUES alongside a subject-bound
+                    // one would reintroduce the cartesian seed the deferral exists to prevent.
+                    let (seed_values, deferred_values) = if operator.is_none()
                         && !block.values.is_empty()
                         && block.triples.len() >= 2
                         && is_property_join(&block.triples)
-                        && !values_feed_triples(&block.values, &block.triples);
+                    {
+                        split_values_seeding_star(block.values, &block.triples)
+                    } else {
+                        (block.values, Vec::new())
+                    };
 
                     // The deferred VALUES wrapper joins on its vars AFTER the triple
                     // operators run, so those vars must survive schema pruning even
                     // when the SELECT list doesn't mention them.
-                    if values_after_triples {
-                        if let Some(rwv) = augmented_rwv.as_mut() {
-                            for vp in &block.values {
-                                for v in &vp.vars {
-                                    if !rwv.contains(v) {
-                                        rwv.push(*v);
-                                    }
-                                }
+                    if let Some(rwv) = augmented_rwv.as_mut() {
+                        for v in deferred_values.iter().flat_map(|vp| &vp.vars) {
+                            if !rwv.contains(v) {
+                                rwv.push(*v);
                             }
                         }
                     }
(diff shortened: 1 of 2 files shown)


bplatz (commented):
The diagnosis is right and the fix is in the right place. Two gaps in what's established, both inline.

One more that doesn't anchor to a line: the PR notes this should land after #1624 because "the seeded plan routes some historical-`t` reads through lanes that #1624 makes correct on rebuilt indexes." Those lanes (batched-subject probes) carry their own constraints — they read base leaflets directly and are latest-`t` only, with a novelty-overlay bail gate. If the seeded plan can newly reach them, the bail gates deserve a test at a historical `t` and one with a novelty overlay. Right now the coverage is two plan-shape unit tests plus one indexed result test, all at head with no overlay — none of which would catch a lane that's reachable but not correctly gated.

fluree-db-query/src/execute/where_plan.rs
```diff
@@ -1044,6 +1044,20 @@ fn apply_values(
...
+/// Answers whether the VALUES can seed the triples. A connected VALUES constrains
+/// them; a disconnected one only multiplies rows.
+fn values_feed_triples(values: &[ValuesPattern], triples: &[TriplePattern]) -> bool {
+    let triple_vars: HashSet<VarId> = triples
+        .iter()
+        .flat_map(TriplePattern::referenced_vars)
+        .collect();
+    values
+        .iter()
+        .any(|vp| vp.vars.iter().any(|v| triple_vars.contains(v)))
```
bplatz: Two issues with `any`-over-`any`.

**Mixed VALUES lists defeat the gate.** This returns true if *any* VALUES shares *any* var, but `values_after_triples` is a single decision for the whole `block.values` slice. A group holding both a connected VALUES and a disconnected one — legal SPARQL, `{ VALUES ?a {..} VALUES ?b {..} ?s :p ?a . ?s :q ?r }` — now seeds both, and the disconnected one produces exactly the cartesian seed the deferral was written to prevent. The gate is all-or-nothing where the hazard is per-VALUES.

**Any shared var, not the star's subject.** `VALUES ?n { "x" } . ?s a ex:Thing ; ex:name ?n` binds a leaf *object* var and now seeds, which costs the property-join fusion for the whole star — 8 separate joins instead of one fused scan on a wide star. The arity sweep in the PR body only covers the subject-connected shape, so "no crossover where the extent scan wins" isn't established for this one. Narrowing to a VALUES that binds the star's subject would match both the issue and the evidence.

fluree-db-query/src/execute/where_plan.rs
```diff
@@ -3263,6 +3286,94 @@ mod tests {
...
+                Term::Iri(Arc::from("ex:Thing")),
+            )),
+            Pattern::Triple(make_pattern(VarId(0), "name", VarId(1))),
+        ];
+
+        let op = build_where_operators(&patterns, None).unwrap();
+        let plan = op.describe();
+        let ops = plan_ops(&plan);
+
+        assert_eq!(
```
bplatz, line 3472: Asserting the exact operator vector — including `EmptyOperator` as a leaf artifact of the test harness — makes this brittle against any planner refactor, and it doesn't actually assert the property under test (that the class extent isn't drained); it asserts one tree shape that happens to have that property today.

The companion test right below gets this right with `ops.first()` / `ops.last()`. Same treatment here — "`ValuesOperator` is not at the root" is the durable claim.

bplatz (approved):

fluree-db-query/src/execute/where_plan.rs
aaj3f: Both of these are right, and the second one especially — thank you for catching them. I went and reproduced each against the old gate before changing anything, and they both land exactly as you describe.

On the mixed list: your example is the clean repro. With `{ VALUES ?a {..} VALUES ?b {..} ?s :p ?a . ?s :q ?r }` the old gate saw `?a` in the triple vars, returned `true` for the whole slice, and the plan came out `[NestedLoopJoin, NestedLoopJoin, ValuesOperator, ValuesOperator, EmptyOperator]` — both VALUES seeded at the base, including the disconnected `?b`, which is precisely the cartesian seed the deferral was there to stop. So the gate was not just coarse, it was actively re-introducing the thing it was guarding.

On the subject-vs-any-shared-var point: also confirmed, and I think this was the more important of the two. `VALUES ?n { "x" } . ?s a ex:Thing ; ex:name ?n` under the old gate planned as `[NestedLoopJoin, NestedLoopJoin, ValuesOperator, EmptyOperator]` — seeded, fusion gone. You're right that the arity sweep in the PR body doesn't cover that shape at all; it only ever varied `k` on the subject-connected form, so "no crossover where the extent scan wins" was never established for an object-var VALUES and I shouldn't have let the gate cash a cheque the evidence didn't write.

What I did: narrowed the test to the star's **subject** var, and made the decision per-VALUES rather than per-block. `split_values_seeding_star` partitions `block.values` into the ones binding the subject (seed) and everything else (defer), and the call site seeds the first list and applies the second after the triples. I went with the split rather than the defer-all-if-any-disconnected fallback because it turned out to be cheap — `apply_values` already loops over a `Vec<ValuesPattern>`, so it's a `partition` plus passing two lists instead of one, and it keeps the #51 win in the mixed case instead of giving it up. If you'd rather have the blunter defer-all rule for a while on the theory that mixed blocks are rare and not worth the extra branch, I'm happy to swap it — it's a two-line change either way.

One nice side effect of keying on the subject: `is_property_join` already guarantees every triple in the block shares the first triple's subject var, so the gate no longer has to build a `HashSet` of all triple vars to answer the question.

Tests added, all three of which fail against the old gate: `test_mixed_values_seeds_only_the_subject_binding_one` (subject-bound + disconnected in one block — asserts the disconnected one stays at the root and the subject-bound one seeds from the base), `test_values_sharing_only_object_vars_all_defer` (your exact example — asserts nothing seeds the base), and `test_object_var_values_keeps_property_join` (asserts `PropertyJoinOperator` is **retained**).

fluree-db-query/src/execute/where_plan.rs
aaj3f: Agreed, and the `EmptyOperator` bit in particular is fair — that's a harness artifact I had no business pinning. Rewritten to the durable claims, matching the companion test's `first()`/`last()` shape.

The one thing I kept from the old assert, because I think it's the piece that actually speaks to your "it doesn't assert the property under test" point, is `!ops.contains("PropertyJoinOperator")`. That is the property under test, more or less directly: the whole defect was that the star fused and the driver scan drained the class extent, so "no property join here" is the closest thing to "the extent isn't drained" that a plan-shape assertion can say. So the test now asserts `ValuesOperator` is not at the root, no `PropertyJoinOperator` anywhere, and the VALUES is still applied somewhere — no positional coupling beyond the root.

aaj3f:
This is a good catch and you're right that the coverage was thin in exactly the place it mattered: two plan-shape unit tests plus one result test, all at head, none of which would notice a lane that's reachable but mis-gated. Both tests added, in `it_query_sparql_indexed.rs`.

`indexed_values_seeded_star_replays_at_historical_t` changes a value between `t=1` and `t=2` and asserts the seeded plan returns the `t=1` value at `t=1` and the `t=2` value at head. `indexed_values_seeded_star_sees_novelty_overlay` indexes one subject, commits a second one plus an extra value on the first without re-indexing, and asserts both the novelty-only subject and the straddling subject come back complete. Both assert results rather than plan shape, per your framing — a lane that bails to a slower fallback is fine, one that serves the wrong rows is not.

The bit I want to flag rather than bury: each test **also** asserts the query still plans as seeded (`assert_values_seeded` — `ValuesOperator` present, no `PropertyJoinOperator`) before it asserts results. Without that these would pass just as happily on the old deferred plan, since both plans return correct rows here, and the tests would quietly stop covering the lane they were written for the moment the planner changed its mind. Cheap insurance, and it seemed worth being explicit about given that "reachable but not correctly gated" was the concern.

There's a third test now too — `indexed_values_seeded_star_replays_fully_retracted_predicate` — which is really the one your note was pointing at. #1624 merged while I was working through this, so I rebased onto it and could finally cover the shape that actually matters here: a predicate that a later **full rebuild** leaves with no live rows at all, queried at a historical `t` through the seeded plan. That's the case where the two lanes genuinely disagree, since it's unreachable through PSOT/POST and only the SPOT history sidecars #1624 preserves can answer it. It uses `rebuild_and_publish_index` rather than the incremental helper on purpose — the full rebuild is the path that used to drop the partition — and the "empty at head" assertion doubles as proof the predicate really is fully retracted rather than just quiet.

Which lets me say the merge-order note in the PR body more precisely than I originally did, and it was vaguer than it should have been. The reason this wanted #1624 first isn't tidiness: moving these queries onto the seeded lane takes them **off** `PropertyJoinOperator`, and `driver_score`'s replay branch (`property_join.rs:222-234`) was the only thing keeping the fully-retracted shape correct at a historical `t` — it hard-prefers a bound `rdf:type` driver under replay specifically so every other predicate stays on the sidecar-replaying probe. So for that one shape this PR was trading a plan that worked by construction for one that didn't, until #1624 fixed the storage side. With #1624 in, that's closed, and the new test pins it. Current-state queries were never affected either way.

PR #1628, opened 2026-08-11
tighten indexed fast-path eligibility guards

aaj3f:
Six related cases where an indexed-lane fast path answered a question its metadata can't actually answer, each returning a plausible wrong number (or a spurious error) with HTTP 200. All of the fixes are eligibility narrowings — the declined query falls through to the general pipeline, which computes the correct answer in every case — so no currently-correct query changes behavior.

The EXISTS semijoin keyed on `(subject_var, pred)` and gated on a plan-time `Term::is_bound()`, which is blind to a variable an outer pattern binds at runtime — so `NOT EXISTS { ?s2 :member ?x }` with `?x` correlated was answered as "does `?s2` have any member" on the indexed lane (the W3C `subset-02` negation query returned 25 rows where 11 is correct). The COUNT/COUNT DISTINCT/AVG/MIN-MAX detectors never compared subject to object, so a repeated-variable triple like `{ ?x :rel ?x }` counted the predicate's full cardinality instead of the self-loops. The fused SUM detectors never read the subject position at all, so a constant-subject `SUM(?v)` returned the ledger-wide sum for a per-entity question. Routing those SUM detectors through the shared triple validator also picked up its datatype-constraint decline, which they never had of their own — so a datatype-constrained `SUM` no longer takes the fused scan and sums the predicate's whole object column across every datatype under it. A repeated-variable triple joined to a second same-subject pattern failed outright with `Duplicate VarId` (the nested-loop join widened the same variable into two schema columns; the scan layer already folds it). And the whole-graph `COUNT(DISTINCT ?o)` grouped on the OPST `(o_type, o_key)` lead, which for NumBig objects is a per-predicate arena handle rather than a global identity — two different decimals under two predicates share a lead, so distinct counts collapsed toward the largest single predicate's count (every `xsd:decimal` and any `xsd:integer` beyond i64 is arena-routed). That last gate declines on directory metadata alone; an exact NumBig-aware counter is tracked as #1638 so decimal-heavy graphs eventually get the fast path back.

The regression suite asserts both directions via the span-capture harness: must-not-fire (no `proceed` stamp, correct value) for every broken shape, and must-fire (positive stamp) for the shapes that should keep their fast path — including on decimal-bearing ledgers, so an over-broad gate fails loudly. One note for reviewers: these tests wait for background index application and probe the lane before asserting, since a lost indexer race otherwise downgrades an "indexed" test to novelty and it passes vacuously.

### DIFF
--- a/fluree-db-core/src/o_type.rs
+++ b/fluree-db-core/src/o_type.rs
@@ -215,7 +215,34 @@ impl OType {
     /// the `p_id` that scopes the handle.
     #[inline]
     pub const fn o_key_is_globally_identifying(self) -> bool {
-        self.0 != Self::NUM_BIG_OVERFLOW.0
+        let mut i = 0;
+        while i < Self::NOT_GLOBALLY_IDENTIFYING.len() {
+            if self.0 == Self::NOT_GLOBALLY_IDENTIFYING[i].0 {
+                return false;
+            }
+            i += 1;
+        }
+        true
+    }
+
+    /// Every o_type whose `o_key` is scoped to something narrower than the
+    /// graph. Single source of truth for [`Self::o_key_is_globally_identifying`]
+    /// and [`Self::range_holds_non_globally_identifying`], so the point test and
+    /// the range test cannot drift apart.
+    const NOT_GLOBALLY_IDENTIFYING: [Self; 1] = [Self::NUM_BIG_OVERFLOW];
+
+    /// Whether the inclusive o_type range `[lo, hi]` can hold a row whose
+    /// `o_key` is not a graph-wide identity.
+    ///
+    /// Index keys lead with a big-endian `o_type`, so a leaflet's
+    /// `[first_key, last_key]` pair bounds every o_type it contains. This
+    /// answers the range form of [`Self::o_key_is_globally_identifying`]
+    /// without a caller having to know which o_types are the exceptions.
+    #[inline]
+    pub fn range_holds_non_globally_identifying(lo: Self, hi: Self) -> bool {
+        Self::NOT_GLOBALLY_IDENTIFYING
+            .iter()
+            .any(|t| lo.0 <= t.0 && t.0 <= hi.0)
     }
 
     // ── Constructors ───────────────────────────────────────────────────
@@ -676,6 +703,30 @@ mod tests {
             let ot = OType::from_u16(val);
             assert_eq!(ot.as_u16(), val);
         }
+    }
+
+    /// The point predicate and the range form share one exceptions list; this
+    /// pins that they agree, so a new non-identifying o_type added to the list
+    /// is answered by both.
+    #[test]
+    fn globally_identifying_point_and_range_agree() {
+        for &t in &OType::NOT_GLOBALLY_IDENTIFYING {
+            assert!(!t.o_key_is_globally_identifying(), "{t:?}");
+            assert!(OType::range_holds_non_globally_identifying(t, t));
+        }
+        // A range that straddles an exception reports it; one that stops short
+        // does not.
+        let ex = OType::NUM_BIG_OVERFLOW;
+        let below = OType::from_u16(ex.as_u16() - 1);
+        let above = OType::from_u16(ex.as_u16() + 1);
+        assert!(OType::range_holds_non_globally_identifying(below, above));
+        assert!(!OType::range_holds_non_globally_identifying(
+            OType::XSD_STRING,
+            below
+        ));
+        assert!(!OType::range_holds_non_globally_identifying(above, above));
+        assert!(below.o_key_is_globally_identifying());
+        assert!(above.o_key_is_globally_identifying());
     }
 
     #[test]
--- a/fluree-db-core/src/o_type.rs
+++ b/fluree-db-core/src/o_type.rs
@@ -90,7 +90,10 @@ impl OType {
     pub const XSD_DOUBLE: Self = Self(0x0010);
     /// `xsd:float`
     pub const XSD_FLOAT: Self = Self(0x0011);
-    /// `xsd:decimal` (inline f64 — overflow goes to NumBig arena via `NUM_BIG_OVERFLOW`).
+    /// `xsd:decimal`. Reserved: the indexer routes **every** typed
+    /// `xsd:decimal` to the NumBig arena under [`Self::NUM_BIG_OVERFLOW`]
+    /// (`resolver.rs`, `RawObject::DecimalStr` — there is no inline branch), so
+    /// no stored row carries this o_type.
     pub const XSD_DECIMAL: Self = Self(0x0012);
 
     // -- Temporal (various order-preserving encodings in o_key) --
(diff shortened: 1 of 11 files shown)


fluree-db-core/src/o_type.rs
```diff
@@ -191,6 +194,30 @@ impl OType {
...
+    /// under another are both handle `0`, so the pair is an identity only once
+    /// `p_id` is fixed. Every `xsd:decimal` lands there unconditionally, as does
+    /// any `xsd:integer` too large for `i64`.
+    ///
+    /// Call this before using an `(o_type, o_key)` prefix as an object identity
+    /// in any whole-graph structure — a distinct count, a group key, a dedup
+    /// set. Per-predicate consumers are unaffected: their key already carries
+    /// the `p_id` that scopes the handle.
+    #[inline]
+    pub const fn o_key_is_globally_identifying(self) -> bool {
```
bplatz, line 217: This has no call sites — I grepped the branch and `main`. `leaflet_may_hold_non_identifying_o_key` open-codes `lo <= NUM_BIG_OVERFLOW <= hi` instead of calling it, and the only other reference is the doc-link from that function's comment.

Because it's `pub`, the compiler won't warn, so it lands as silently dead API on a type that's part of the storage format's vocabulary. The repo's dead-code policy wants one of: use it, `#[expect(dead_code)]` with why/when/how, or delete.

Using it is easy and better than the current shape — the `leaflet_may_hold_non_identifying_o_key` interval test can express itself in terms of this predicate rather than repeating the `NUM_BIG_OVERFLOW` constant, which is exactly the coupling the doc comment is trying to establish.

fluree-db-query/src/filter.rs
```diff
@@ -204,10 +223,18 @@ fn build_exists_semijoin_cache(
...
     let mut cache = ExistsSemijoinCache::default();
-    for (sv, pred_ref) in exists_nodes {
+    for (sv, pred_ref, ov) in exists_nodes {
         if !schema_vars.contains(&sv) {
             continue;
         }
+        // The object var is fed by an outer pattern, so this EXISTS asks about
+        // one specific object rather than "any object". The subject-set key
+        // cannot express that, and the per-row guard in
+        // `try_eval_simple_exists_semijoin` would decline every row anyway —
```
bplatz: This justification isn't right, and the two gates don't actually agree.

The per-row guard declines only when `batch.get(row_idx, object_var).is_some_and(Binding::is_matchable)`. A var that's *in the schema* but unbound in a given row — an unmatched `OPTIONAL`, say — is `is_matchable == false`, so the per-row guard lets that row proceed. This build-time gate skips the entry whenever the var appears in the schema at all, bound or not. It's strictly more conservative, not equivalent.

The direction is safe (correctness-preserving, perf-only), and there's a perfectly good reason for it — at cache-build time you can't know per-row bindings, and speculatively paying a PSOT scan for rows that may all decline is the wrong trade. That's the argument the comment should make.

Worth noting the per-row guard's treatment of the unbound case is subtly *correct*, not just permissive: an unbound variable inside `EXISTS` is a fresh free variable per SPARQL, so "any object" is the right answer there. That's non-obvious enough to be worth stating where the guard lives.

bplatz (commented):
All five narrowings check out, and I traced the NumBig gate end to end. Three notes beyond the inline comments.

**The test isolation is the right call and worth saying so.** Separate `[[test]]` binaries for the ones that toggle the process-global kill switch or assert routing via tracing capture, and `wait_for_index_application` as a real barrier rather than sleeps. That's what keeps these from becoming the flaky-under-load tests this pattern usually produces.

**One narrowing isn't in the list of five.** Routing `detect_fused_scan_sum_i64` through `validate_simple_triple` also inherits the `tp.dtc.is_some()` decline, which that detector never had — so a datatype-constrained `SUM` that previously took the fused scan and ignored the constraint now falls back. That's a sixth correctness fix riding along; it deserves a line in the body rather than being discovered later.

**The decimal fast-path loss needs a tracked issue, not a sentence.** Every `xsd:decimal` and any `xsd:integer` past `i64` puts a graph on the general pipeline for whole-graph `COUNT(DISTINCT ?o)`, permanently. "A reasonable perf follow-up" in a PR description isn't a plan — and this area has already been through a disable/fix/reinstate cycle once. Please file it and link it here before merge.

One thing I checked so it doesn't get re-derived: the `row_count == 0 || lead_group_count == 0` skip sits *above* the new `non_identifying` check, so the zero-row leaflets #1624 starts emitting can't trip the gate. The two PRs compose cleanly.

bplatz (approved):

fluree-db-core/src/o_type.rs
aaj3f: You're right, and thanks for grepping `main` as well as the branch — I'd quietly counted the doc-link from `leaflet_may_hold_non_identifying_o_key` as a use, which of course it isn't, and `pub` meant nothing was ever going to tell me otherwise.

Fixed by taking your suggestion rather than reaching for `#[expect(dead_code)]`. Both forms now derive from one list, `OType::NOT_GLOBALLY_IDENTIFYING`: `o_key_is_globally_identifying` answers the point question from it, and a new `range_holds_non_globally_identifying(lo, hi)` answers the interval question, which is the one a leaflet directory entry is actually asking — `[first_key[0..2], last_key[0..2]]` bounds every `o_type` in the leaflet. The gate then calls the point predicate for the type-homogeneous case the indexer actually emits and the range form otherwise, and names no `o_type` constant of its own. So a future non-identifying `o_type` is one list entry with no edit in `fluree-db-query`.

I added a small unit test (`globally_identifying_point_and_range_agree`) that walks the list and asserts the point and range forms agree, plus the adjacent-value cases either side of `NUM_BIG_OVERFLOW`, mostly so the two can't drift if that list ever grows.

fluree-db-query/src/filter.rs
aaj3f: You're right and the comment was straightforwardly wrong — thank you for working the case through rather than just flagging it. Schema presence is strictly more conservative: a var that's in the schema but unbound in a given row (the unmatched `OPTIONAL` case you name) is `is_matchable == false`, so the per-row guard lets that row take the semijoin while the build-time gate has already skipped the whole cache entry. "Would decline every row anyway" was me reasoning about the bound case and writing it as though it covered both.

Rewrote it to make the argument you laid out: at cache-build time there are no per-row bindings to consult, so the choice is between skipping the entry and speculatively paying a PSOT scan for rows that may all decline — and the skip is safe to reason about precisely because it costs only performance, with the per-row guard doing the correctness work.

I also took the second half of your note and put it where the per-row guard lives, since you're right that it's non-obvious: EXISTS substitutes only the bindings the row actually carries, so a variable left unbound by an unmatched `OPTIONAL` is a fresh free variable inside the subpattern, and "does this subject have any object for this predicate" is genuinely the right question for those rows. That reads as an accident in the code otherwise.

aaj3f:
@bplatz — the review-body notes:

**On the sixth narrowing:** Good catch, and you're right that it deserved a line rather than being found later. Routing `detect_fused_scan_sum_i64` through `validate_simple_triple` does inherit that helper's `tp.dtc.is_some()` decline, which the SUM detectors never had of their own, so a datatype-constrained `SUM` used to take the fused scan and sum the predicate's whole object column across every datatype under it. It's in the PR body now.

It also had no test, which I liked less than the missing body line — so there's one now. A datatype constraint alongside a *variable* object has no SPARQL surface (it's the JSON-LD `{"@value": "?v", "@type": …}` form), so it runs on its own small ledger with two datatypes under one predicate: constrained to `xsd:integer` the answer is 10, unconstrained it's 30. Before the change the constrained form returned 30 with a `SUM(?o)` proceed stamp, which is the shape you predicted. The same ledger carries the must-fire control — drop the constraint and `SUM(?o)` has to proceed again — so the decline can't quietly widen into a disable.

One fixture note in case it looks arbitrary: the second datatype is `xsd:positiveInteger` because it survives numeric-datatype normalization. `xsd:long` and friends fold into `xsd:integer`, which would have made the constraint a no-op and the test vacuous.

**On the decimal fast-path loss:** Filed as #1638 and linked from the body. Fair point that "a reasonable perf follow-up" in a description isn't a plan, and the disable/fix/reinstate history in this area is exactly why.

The issue carries the mechanism, the two traps I'd want the next person to not rediscover (arena contents over-approximate live rows once anything has been retracted, so unioning the per-predicate arenas over-counts; and recovering `(p_id, handle)` means payload reads, which is precisely the property that makes the current walk cheap), and the argument that the deliverable is a *middle branch* on this gate rather than a replacement for it — with the threshold measured rather than guessed. It also flags the dormant `feature/inline-decimal-encoding` branch, since if that direction is ever revived most of the issue's value evaporates and that's worth establishing before anyone spends M-sized effort.

**On the composition check and the test-isolation note:** Thank you for the `#1624` composition check specifically — the `row_count == 0 || lead_group_count == 0` skip sitting above the new `non_identifying` check is exactly the kind of thing I'd have had to re-derive under time pressure at merge, and having it written down in the thread means nobody does that twice.

And I appreciate you calling out the test isolation as deliberate rather than incidental. The separate `[[test]]` binaries and `wait_for_index_application` as a real barrier came out of getting burned during this work: an early version of the `#48` test passed green with the guard removed, because under default test parallelism the background indexer lost the race and the "indexed" read was still novelty. That's why every indexed-lane test in the branch now probes the lane before asserting — a lost race should fail loudly rather than silently downgrade the test to a novelty-only one.

PR #1629, opened 2026-08-11
accept legal aggregate and property-path forms below the parser

aaj3f:
Three cases where the parser accepts a legal SPARQL 1.1 form and a lower layer then rejects it with a 400, plus two aggregate-evaluation fixes found along the way.

Property paths with literal objects: lowering forced every path object through `Ref::try_from`, so `?s ex:ofMaker/ex:name "Acme"` — an ordinary, matchable shape — was rejected wholesale. Endpoints now carry a `Term` and each arm narrows only as far as it must: sequence final hops and alternative branches take the literal directly, a simple inverse onto a literal lowers to a statically-empty pattern (RDF has no literal subjects — zero solutions, not an error), and the transitive arms keep a narrow error naming the sub-case, since their IR endpoints are typed `Ref`.

Aggregates over a `GROUP BY` key: `SELECT ?k (COUNT(?k) AS ?n) … GROUP BY ?k` (and the `HAVING`/`COUNT(DISTINCT ?k)` variants) returned 400. The check was guarding a real hazard on the traditional grouping path — the aggregate would have answered with the key term itself — so rather than delete it, each aggregated key is copied to a fresh non-key column before grouping and the aggregate reads the copy. The copy preserves unboundness, so an `OPTIONAL`-bound key keeps `COUNT(?k)` at 0 where `COUNT(*)` counts rows. The traditional path is pinned explicitly via EXPLAIN (the streaming path alone would have passed even with the old hazard).

`COUNT(DISTINCT *)` is now implemented (count of distinct solutions, §18.5.1.1 — the W3C `agg-count-rows-distinct` test now passes and its register entry is removed), with distinctness scoped to user-visible variables so the lowerer's synthetic path-join and blank-node variables don't split solutions the spec considers identical.

And `GROUP_CONCAT` over IRI-valued variables no longer silently drops them (it only extracted strings from literal bindings, so an all-IRI group returned null): IRIs expand through the same namespace table `STR()` uses, matching Jena. The spec states no coercion for non-literal members, so if we'd rather this be a type error than a concatenation, say so — either beats dropping members silently.

### DIFF
--- a/fluree-db-query/src/aggregate.rs
+++ b/fluree-db-query/src/aggregate.rs
@@ -254,6 +254,15 @@ impl Operator for AggregateOperator {
                                 )
                             })
                             .collect(),
+                        // COUNT(DISTINCT *): the number of distinct solutions in
+                        // the group, composed across every child column.
+                        None if matches!(spec.function, AggregateFn::CountDistinctAll) => (0
+                            ..batch.len())
+                            .map(|row_idx| {
+                                let n = distinct_group_rows(&batch, row_idx, self.child_col_count);
+                                Binding::lit(FlakeValue::Long(n), xsd_integer())
+                            })
+                            .collect(),
                         None => {
                             let sizes = group_sizes.get_or_insert_with(|| {
                                 (0..batch.len())
@@ -410,6 +419,18 @@ impl AggregateFn {
         match self {
             Self::Count(_) => agg_count(values),
             Self::CountAll => agg_count_all(values),
+            // Unreachable by construction: `CountDistinctAll` has no input
+            // variable, so it never gets a column and never reaches `apply`.
+            // `AggregateOperator` computes it from every child column instead
+            // (see `distinct_group_rows`). Assert in debug so a future wiring
+            // change is caught rather than silently answering Unbound.
+            Self::CountDistinctAll => {
+                debug_assert!(
+                    false,
+                    "COUNT(DISTINCT *) must be computed over the whole row, not one column"
+                );
+                Binding::Unbound
+            }
             Self::CountDistinct(_) => agg_count_distinct(values),
             Self::Sum { .. } => agg_sum(values),
             Self::Avg { .. } => agg_avg(values),
@@ -733,6 +754,32 @@ fn agg_count_all(values: &[Binding]) -> Binding {
 }
 
 /// COUNT(DISTINCT) - count distinct non-Unbound values
+/// Number of distinct solutions in one group of a `GroupByOperator` output row.
+///
+/// `GroupByOperator` emits one row per group: key columns carry the single key
+/// value, every other column carries `Grouped(Vec<Binding>)` with one entry per
+/// solution in the group. Zipping the grouped columns index-wise reconstructs
+/// the group's solutions. Key columns are constant within a group, so they
+/// cannot change which solutions are distinct — and a group with no grouped
+/// column at all is a set of rows agreeing on every column, i.e. exactly one
+/// distinct solution.
+fn distinct_group_rows(batch: &Batch, row_idx: usize, child_col_count: usize) -> i64 {
+    let grouped: Vec<&[Binding]> = (0..child_col_count)
+        .filter_map(|col_idx| match batch.get_by_col(row_idx, col_idx) {
+            Binding::Grouped(values) => Some(values.as_slice()),
+            _ => None,
+        })
+        .collect();
+    let Some(group_size) = grouped.iter().map(|col| col.len()).min() else {
+        return 1;
+    };
+    let mut seen: HashSet<Vec<&Binding>> = HashSet::with_capacity(group_size);
+    for i in 0..group_size {
+        seen.insert(grouped.iter().map(|col| &col[i]).collect());
+    }
+    seen.len() as i64
+}
+
 fn agg_count_distinct(values: &[Binding]) -> Binding {
     let distinct: HashSet<_> = values
         .iter()
--- a/fluree-db-query/src/aggregate.rs
+++ b/fluree-db-query/src/aggregate.rs
@@ -29,7 +29,7 @@ use bigdecimal::{BigDecimal, ToPrimitive};
 use fluree_db_core::{FlakeValue, Sid};
 use num_bigint::BigInt;
 use num_traits::{Signed, Zero};
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::time::Instant;
 use tracing::Instrument;
@@ -225,6 +225,8 @@ impl Operator for AggregateOperator {
 
             let num_cols = self.in_schema.len();
             let mut output_columns: Vec<Vec<Binding>> = Vec::with_capacity(num_cols);
+            // Namespace table for expanding IRI-valued GROUP_CONCAT members.
+            let namespaces = ctx.active_snapshot.namespaces();
             ctx.check_cancelled()?;
 
             // Process child columns (regular aggregates and pass-through)
@@ -239,7 +241,12 @@ impl Operator for AggregateOperator {
                         Some(agg_idx) => {
                             // This column needs aggregation
                             let spec = &self.aggregates[agg_idx];
-                            apply_aggregate(&spec.function, input_binding, self.graph_view.as_ref())
+                            apply_aggregate(
+                                &spec.function,
+                                input_binding,
+                                self.graph_view.as_ref(),
+                                Some(namespaces),
+                            )
                         }
                         None => {
                             // Pass through unchanged
@@ -268,6 +275,7 @@ impl Operator for AggregateOperator {
                                     &spec.function,
                                     input_binding,
                                     self.graph_view.as_ref(),
+                                    Some(namespaces),
                                 )
                             })
                             .collect(),
@@ -344,10 +352,14 @@ impl Operator for AggregateOperator {
 ///
 /// Count variants and the numeric accumulator handle encoded bindings
 /// natively, so they skip the decode.
+///
+/// GROUP_CONCAT additionally needs IRI-valued bindings expanded to their IRI
+/// string (`namespaces`); see [`group_concat_expand_iris`].
 fn apply_aggregate(
     func: &AggregateFn,
     binding: &Binding,
     gv: Option<&fluree_db_binary_index::BinaryGraphView>,
+    namespaces: Option<&HashMap<u16, String>>,
 ) -> Binding {
     let needs_decoded_values = matches!(
         func,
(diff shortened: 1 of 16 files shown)


fluree-db-query/src/aggregate.rs
```diff
@@ -208,6 +225,8 @@ impl Operator for AggregateOperator {
 
             let num_cols = self.in_schema.len();
             let mut output_columns: Vec<Vec<Binding>> = Vec::with_capacity(num_cols);
+            // Namespace table for expanding IRI-valued GROUP_CONCAT members.
```
bplatz, line 228: `ctx.active_snapshot` is one snapshot. In a dataset / multi-ledger query, is this the right namespace table for every IRI that reaches the aggregate?

If bindings arrive already re-encoded into a union namespace space, this is fine. If any arrive carrying a non-active ledger's namespace codes, they expand against the wrong table — which is the #1259 failure mode, and would show up as `GROUP_CONCAT` emitting a wrong IRI rather than dropping one. Neither the new tests nor the existing ones cover `GROUP_CONCAT` over a multi-ledger dataset, so the answer isn't pinned either way.

fluree-db-query/src/execute/dependency.rs
```diff
@@ -36,6 +36,19 @@ pub struct VariableDeps {
...
+    // the pre-grouping row. It reports no input variable, so the backward walk
+    // below cannot see that dependency — trimming a column would silently
+    // collapse solutions that differ only in it. Disable trimming outright.
+    if query
+        .grouping
+        .iter()
+        .flat_map(Grouping::aggregates)
+        .any(|spec| matches!(spec.function, AggregateFn::CountDistinctAll(_)))
+    {
+        return None;
```
bplatz: This disables dead-column trimming for the *entire query* whenever any `COUNT(DISTINCT *)` appears anywhere in it — including scopes the aggregate never reads, and including the common shape where it sits beside other aggregates over a wide WHERE.

The conservatism is correct; the blast radius is wider than the hazard. Worth either scoping it to the grouping's own input scope, or noting in the comment that it's deliberately query-global so the cost isn't a surprise later.

bplatz (commented):
CI is green now. Two inline comments.

On the `GROUP_CONCAT` question in the description: concatenating is the right call over erroring — Jena parity beats a type error for a shape that silently dropped members before.

Checked and clear, so nobody re-derives: `never_matches` is safe with a constant subject (`apply_values` never elides, and `ValuesOperator` with zero rows yields zero rows regardless of schema); and dropping `dtc` on the plain path surface really does match the hand-written triple, since `term.rs:67` also lowers objects without a constraint.

bplatz (approved):

fluree-db-query/src/aggregate.rs
aaj3f: ## `fluree-db-query/src/aggregate.rs:228` — is `ctx.active_snapshot` the right namespace table?

You were right that nothing pinned this either way, and it was worth asking. The short version: for the dataset path it **is** correct, for a reason worth writing down, and there's now a test that fails if that ever stops being true. But the `SERVICE` path is a different story — see the bottom of this note.

**Why the dataset path is safe.** `DatasetOperator` stamps every cross-ledger binding before it leaves the member. `dataset_operator.rs:388` `stamp_provenance` runs each batch through `stamp_binding` (`:428`), which converts `Binding::Sid` → `Binding::IriMatch` carrying the IRI decoded *in its own ledger* via `ctx.decode_sid_in_ledger(sid, ledger_id)` — and it **errors** rather than falling back to `Binding::Sid`, precisely because multi-ledger equality is defined around `IriMatch`. It's gated on `needs_provenance`, set to `multi_ledger` at `:569`, and `:616-620` disables the binary store for those members so scans yield `Binding::Sid` rather than an `EncodedSid` that couldn't be decoded later. So by the time a batch reaches the aggregate, every cross-ledger IRI is an `IriMatch`, and `group_concat_expand_iris` reads that canonical `iri` field directly and never consults `active_snapshot` for it. The single-ledger case is trivially fine because `active_snapshot` *is* that ledger.

**And it's pinned now** (`5f7abcaab`). The new test seeds two ledgers under different prefixes — `http://alpha.example/` and `http://beta.example/` — that get allocated the *same* namespace code, since each ledger numbers its namespaces independently and in insertion order. That's what makes a wrong-table expansion visible rather than merely possible: expanding beta's SID against alpha's table produces `http://alpha.example/b1`, well-formed and wrong, which is exactly the #1259 shape you flagged.

I didn't want to ship a green test that was green for the wrong reason, so I mutation-checked it: routing the `IriMatch` arm through the namespace table instead of its canonical IRI makes it fail with

```
each subject must expand against its OWN ledger's namespace table,
got "http://alpha.example/a1|http://alpha.example/b1"
```

so the assertion really is load-bearing.

**The part that isn't fine — and I don't think it's this PR's.** While chasing the above I probed `SERVICE` against another ledger, and it does **not** go through `stamp_provenance`: `service.rs:189-194` builds the inner tree against `ctx.with_graph_ref(gref)` and buffers the resulting rows as-is, so a `Binding::Sid` carrying the *target* ledger's namespace codes flows back out into a context whose `active_snapshot` is the parent. Against the same two-ledger fixture:

```sparql
SELECT ?s WHERE { SERVICE <fluree:ledger:svc-beta:main> { ?s ?p "shared" } }
```

returns `http://alpha.example/b1` — the wrong IRI, with **no aggregate anywhere in the query**. `GROUP_CONCAT(?s)` over the same pattern inherits it, but the plain `SELECT` is the proof that this predates the change: the formatter is decoding a foreign SID against the primary snapshot. Two useful contrasts from the same probe — `GROUP_CONCAT(STR(?s))` returns `http://beta.example/b1` correctly, because `eval_str` runs per-row while the ctx is still the SERVICE target's; and the `GRAPH <svc-beta:main>` form is also correct, because that one *does* go through the dataset operator.

I'd rather not fix that here, and not because of appetite — it's that the obvious fix (stamp the SERVICE operator's returned batches the way dataset members are stamped) turns every SERVICE-returned IRI into an `IriMatch`, which changes join, `DISTINCT`, and `GROUP BY` key identity for every SERVICE query, and `stamp_binding` hard-errors on `EncodedSid` while the SERVICE path doesn't disable the binary store. That's a real change with its own test surface, sitting in a different subsystem from anything this PR touches. So my instinct is a follow-up issue with the repro above, and I'm happy to pick it up as the next thing — but if you'd rather it ride along here, say so and I will.

fluree-db-query/src/execute/dependency.rs
aaj3f: ## `fluree-db-query/src/execute/dependency.rs:49` — trimming disabled query-global

Agreed, and thank you — the blast radius really was wider than the hazard, and poking at your either/or turned up a third option that I think is better than either.

The thing I'd missed is that this aggregate *does* have inputs; it carries them. `CountDistinctAll` was given the user-visible variable list in `ee670ad06` (the commit that stopped `?__ppN` and blank-node variables from splitting solutions), so the backward walk doesn't need to be told "give up" — it needs to be told what to trace. `6d7d5b7a3` deletes the early return and extends the aggregate step of the walk instead, so `CountDistinctAll`'s variable list is traced exactly the way every other aggregate's single `input_var` already is:

```rust
match &spec.function {
    AggregateFn::CountDistinctAll(vars) => deps.extend(vars.iter().copied()),
    other => deps.extend(other.input_var()),
}
```

That lands better than either of the options you offered, I think. The pre-grouping stages keep precisely the columns the whole-row read needs and genuinely dead columns stay trimmable; the post-grouping sets are recorded earlier in the walk so they're untouched entirely; and variables the list names from other scopes simply never match a schema and drop out in `compute_trimmed_vars`. Net effect is that the common shape you named — `COUNT(DISTINCT *)` sitting beside other aggregates over a wide WHERE — now keeps its trimming everywhere except the columns that are actually load-bearing.

Same anti-vacuity check as above, since a dependency-set widening is the kind of thing that can look right while doing nothing: dropping just that one match arm fails five of the `COUNT(DISTINCT *)` tests, including `sparql_count_distinct_star_survives_projection_trimming`, which projects neither WHERE variable and still has to count 2. So the trace is doing real work rather than belt-and-braces.

aaj3f:
Thank you, @bplatz — and thanks especially for chasing down `never_matches` and the `dtc` drop and writing the answers into the review. Both were things I'd reasoned about but not proved, and having `apply_values` never eliding and `term.rs:67` also lowering objects without a constraint stated plainly means nobody has to re-derive either one. That's exactly the kind of note that saves a future reader an hour.

Good to have the `GROUP_CONCAT` call confirmed too. Jena parity over a type error was the reading I leaned toward but it genuinely was a coin-flip on the spec text (§18.5.1.7 doesn't say what coercion applies to a non-literal member), so I'd rather have it be a decision we made than a default I picked. I've left the spec ambiguity noted in the commit message so the release note can say we chose it.

Both inline notes are addressed below. The `active_snapshot` one turned into the more interesting of the two — the answer for the dataset path is "correct, and now pinned", but chasing it surfaced a real pre-existing hole on the `SERVICE` path that I think wants its own issue.

**Verification at `6d7d5b7a3`** — ## Verification at `6d7d5b7a3`

Workspace-wide this time rather than the touched crates — that was the lesson from the two follow-up commits it took to get CI green here, and `err_expect` would have been caught by it. `cargo clippy --workspace --all-targets` clean on every file this branch touches; `cargo fmt --all` clean. `cargo test -p fluree-db-query` (all targets, not just `--lib`) 1381 + 6 target suites green; `-p fluree-db-sparql` 625 + 6 doctests; `-p fluree-db-cypher` green. `fluree-db-api`: `grp_query_sparql` 330, `grp_query` 414 (this is the target that owns the dataset tests), `grp_misc` 262, `it_query_cypher` 231, `it_query_explain` 13. W3C full suite 36/36; `sparql11_aggregates` 46 total / 45 passed / 0 failed / 1 registered — unchanged, and still the empty-named-graph divergence rather than anything from this branch; `sparql11_grouping` 6/6.

aaj3f:
Following up on the `SERVICE` hole I mentioned in the `aggregate.rs` thread — filed as #1639 and fixed in #1641.

Two things from chasing it that are worth having here, since both bear on what we concluded in this PR.

The scope turned out narrower than I described. Joins across a `SERVICE` boundary were already correct, in both orders: `service.rs` seeds the inner tree from the parent row, so a term crosses by substitution and the target ledger re-encodes it — nothing was ever comparing a foreign SID. So the defect is confined to IRIs a SERVICE block *newly binds* and that then escape to output or aggregation. The plain `SELECT` is still the proof it predates #1629; `GROUP_CONCAT` just made it visible.

And the representation was already mixed rather than uniform, which is the part that made this a small fix instead of a scary one. A cross-ledger `SERVICE` requires a dataset, and any two-ledger dataset already sets `needs_provenance` — so the parent's own bindings were `IriMatch` while SERVICE's were raw SIDs. Stamping the SERVICE rows makes the query uniform rather than introducing a second regime, which is why the join/`DISTINCT`/`GROUP BY` key-identity worry I raised when I flagged this didn't materialize. I pinned it three ways anyway before touching anything.

Thanks again for pushing on the `active_snapshot` question — the dataset path really was fine, but it would not have occurred to me to check `SERVICE` if you hadn't asked whether one snapshot was the right table for every IRI reaching the aggregate.

PR #1631, opened 2026-08-11
align http dataset-clause semantics with the engine

aaj3f:
The HTTP routes carried their own dataset construction for `FROM`/`FROM NAMED`, parallel to (and older than) the engine's within-ledger builder, and the two disagreed on the same query text. Two user-visible consequences. First, the route registered each named source under the ledger id *and* the graph IRI, so `GRAPH ?g` under a `FROM NAMED` clause enumerated the ledger alias as an extra named graph — every solution doubled, and `GRAPH <ledger-alias>` resolved to the wrong graph's triples. Named sources now register under the name the user wrote. Second, the route injected the ledger's default graph whenever the clause list had no `FROM`, where §13.2 (and the embedded engine, which the W3C dataset tests exercise) gives `FROM NAMED`-only queries an empty default graph. The injection is now gated on the body naming no dataset at all — same rule the connection streaming path already had — for SPARQL and for the JSON-LD `from`/`fromNamed` forms on both endpoints, so the two query languages agree.

Because "FROM NAMED-only with non-GRAPH patterns" previously returned data and now correctly returns none, those requests get an `x-fdb-warning` response header naming the change, and the compat doc carries the release note. Also folded in: the `fromNamed` object form read its graph selector only under `@graph` while the single-source form used `graph`, and the wrong spelling was silently ignored (the source resolved to the whole ledger); both forms now accept both spellings via one shared parser.

Route-level regression tests cover the dataset family over actual HTTP for the first time (the W3C dataset tests only reach the embedded path), including cross-language parity assertions, the multi-graph enumeration counts, and negative controls verified by reverting each fix.


fluree-db-server/src/routes/stream_query.rs
```diff
@@ -111,6 +111,9 @@ async fn stream_query_connection_inner(
 
     let fluree = state.fluree.clone();
 
+    // Advisory headers attached to the 200 (empty except for the one dataset
+    // shape whose answer changed under §13.2 semantics).
```
bplatz: This is set in the JSON-LD branch below but never in the SPARQL branch of this function — `build_stream_dataset_for_sparql` builds the dataset and nothing computes the advisory.

So `FROM NAMED` with no `FROM` warns on `/v1/fluree/query` and on the ledger-scoped streaming route (`stream_query_inner`, which does call `dataset_semantics_warning_headers`), but is silent on connection-scoped `/v1/fluree/stream/query`. Three of the four SPARQL surfaces are wired. Given the whole point is that this shape now returns fewer rows with a 200, the streaming surface is the one where a caller is least likely to notice.

fluree-db-server/src/routes/query.rs
```diff
@@ -2421,6 +2429,176 @@ pub(crate) fn ledger_scoped_sparql_dataset_spec(
     Ok(spec)
 }
 
+/// Response header carrying a non-fatal advisory about how a request was
+/// interpreted. The response body is the normal (successful) result.
+pub(crate) const X_FDB_WARNING: &str = "x-fdb-warning";
```
bplatz, line 2444: New permanent public response header with no sunset and no way to turn it off, emitted on a *successful* 200. Two things worth settling before this ships:

- How long does it stay? If it's a migration aid for the §13.2 change, it wants a removal target; if it's a permanent advisory, the name (`warning`, not `deprecation`) and the doc should say so.
- The function doc calls this "the one shape whose answer changed," but on the connection-scoped route §13.2 was already in force via the embedded engine — nothing changed there, yet `sparql_dataset_warning_headers` fires anyway. The header *text* is accurate as general advice, so this is a scope/doc mismatch rather than a wrong warning; worth narrowing one or the other.

bplatz (commented):
Two inline comments. Checked and clear: `needs_default_graph_injection` covers every key spelling `dataset.rs` actually accepts (`from`, `fromNamed`, `from-named`), so no body slips past it.

Merge note: this and #1625 both append at the `mod tests` anchor in `fluree-db-api/src/dataset.rs` (line ~2561) — a guaranteed conflict, mechanical to resolve since the code hunks are disjoint.

bplatz (approved):

fluree-db-server/src/routes/stream_query.rs
aaj3f: You're right, and thank you for catching it — this was a real hole, not a judgment call. `build_stream_dataset_for_sparql` resolves the dataset internally, so nothing on that path ever looked at the dataset clause, and the `warn_headers` I declared at the top of `stream_query_connection_inner` only ever got written in the JSON-LD branch below. Your framing of why it matters is the part I'd underweighted: of the four surfaces this is precisely the one where a short result set is least likely to get noticed, so it being the silent one was backwards.

Fixed in `6717e032f`. The SPARQL branch now computes the advisory off the same body it's about to plan:

```rust
warn_headers = crate::routes::query::sparql_dataset_warning_headers(
    fluree_db_sparql::parse_sparql(&sparql).ast.as_ref(),
);
```

Covered by `connection_sparql_from_named_only_warns_on_stream` in `fluree-db-server/tests/stream_query_integration.rs`, which asserts the header is present on the `FROM NAMED`-with-no-`FROM` shape and absent when everything sits inside `GRAPH { }`. I also un-wired the line and re-ran it to be sure the test actually fails without the fix rather than passing for some incidental reason.

One thing that fell out of writing that test which I think is worth flagging, since it came from your comment rather than from me noticing it independently: the header **text** was wrong on this route. It read "Add a `FROM` clause to query the ledger's default graph", which is fine on the ledger-scoped endpoint but not here — on the connection-scoped route a clause IRI names a *ledger*, so there isn't a "the ledger" to point at. It now reads "…to name a default graph", which is true on both, and the test pins that phrasing so it can't drift back.

Small follow-on: wiring the fourth surface meant both branches of that function assign `warn_headers`, which made the `HeaderMap::new()` initializer dead and tripped clippy. It's now `let warn_headers;` with both branches assigning, so the crate is back to zero warnings.

fluree-db-server/src/routes/query.rs
aaj3f: Both halves are fair, and I'd rather settle them explicitly than leave them ambient, so I've made a call on each and written it down rather than leaving it in the commit history.

**On lifecycle — it's a permanent advisory, and it now says so.** I went back and forth on giving it a removal target, but I think the honest reading of its own content is that it isn't a migration aid. What it describes is a query *shape* — naming only named graphs and then matching outside them — whose §13.2 semantics surprise people, and that stays exactly as surprising in 4.3 as it is during the 4.1.4 upgrade. A sunset would mean someone who writes that shape in a year gets silence where someone writing it this month gets a nudge, which is hard to justify. So: no sunset, no opt-out, and `warning` rather than `deprecation` deliberately, because nothing it points at is going away. That's now on the `X_FDB_WARNING` doc comment itself and in the `docs/reference/compatibility.md` entry, including the bit clients actually need — the status is always the status the request earned on its own, so the header is informational and must not be keyed on for error handling.

**On the scope/doc mismatch — I fixed the doc rather than narrowing the firing.** You've got the diagnosis exactly right: the text is accurate as general advice, the doc comment is what over-claimed. Given the above makes this an advisory about the shape rather than about the 4.1.4 change, firing everywhere is the coherent behavior and the doc was simply describing it wrong. `dataset_semantics_warning_headers` no longer claims "the one shape whose answer changed"; it now says the advisory fires wherever the shape appears, *including* the connection-scoped route where §13.2 was already in force via the embedded engine and nothing changed — and says why that's still worth emitting there, which is that the failure mode is fewer (often zero) rows under a 200 regardless of whether the answer changed. Narrowing the firing instead would have meant the surface with the longest-standing §13.2 behavior stayed the quietest one, which is the same shape of problem as your `stream_query.rs` finding.

If you'd still rather it carried a sunset I'm happy to be argued out of this one — it's a public header and it's easier to add a removal target now than to take one away later.

aaj3f:
Thanks for checking the `needs_default_graph_injection` spelling coverage against what `dataset.rs` actually accepts — that's the check I'd have most wanted a second pair of eyes on, since the whole gate is worthless if a body slips past it under a key I forgot.

On the merge note: agreed, #1625 and this one both append at the `mod tests` anchor in `fluree-db-api/src/dataset.rs` and will conflict. The hunks are disjoint as you say, so it's mechanical — whichever lands second just keeps both test blocks. No preference on ordering from me.

Separately, your ordering note over on #1632 — that a `fromNamed`-only graph-source query lands with an empty `dataset.default` and takes the `is_empty()` early return once both of these are in — is a good catch on a combination neither PR covers alone. That one's being picked up on #1632 rather than here, since the guard and its test both live on that side.

PR #1632, opened 2026-08-11
refuse graph-source patterns the plan cannot route

aaj3f:
Non-lowerable patterns over a graph source (property-path quantifiers, `shortestPath`, subqueries) are refused with a 400 since #1507 — but the guard lived inside the `GRAPH`-scope rewrite, and `maybe_wrap_for_graph_source` skips the auto-wrap for the whole query once it contains any explicit `GRAPH` block. A query mixing a top-level quantifier with a `GRAPH` block therefore left that path reading the graph source's genesis (zero-flake) snapshot and answered HTTP 200 with nothing. The check now also runs after the wrap, over whatever remains outside every `GRAPH` scope, raising the same typed refusal.

The same bail-out strands ordinary top-level triples, which are lowerable in principle but not lowered on that plan — nothing routes them to the provider, so they read the empty genesis index and zero the whole conjunction. Those are now refused too, scoped to conjunctive top-level position where an empty match provably zeroes the result. `OPTIONAL`/`MINUS`/`UNION`/`EXISTS` bodies are deliberately left alone — they degrade rather than zero, and a `UNION` branch can carry a nested `GRAPH` block that routes and contributes real rows, so refusing there would reject queries that answer today; tests pin those exclusions so the guard doesn't widen casually. Dataset queries are guarded only when every default graph is a graph source, since a native member gives a top-level pattern a real index.

Also documents the limitation user-facing for the first time (graph-sources overview + r2rml/iceberg pages, and a scoping note in graph-crawl.md, which promoted property paths for recursive traversal without saying they're native-ledger-only). Tests drive the real query entry point against a graph-source view — the pre-existing rewrite unit test passes whether or not routing reaches it, which is how the bypass survived.

### DIFF
--- a/fluree-db-api/src/view/query.rs
+++ b/fluree-db-api/src/view/query.rs
@@ -56,10 +56,22 @@ pub(crate) fn maybe_wrap_for_graph_source(db: &GraphDb, parsed: &mut fluree_db_q
 /// match only — as HTTP 200. This check makes the refusal independent of the
 /// query's `GRAPH` structure.
 ///
+/// A plain triple stranded out there is refused too, by
+/// [`unroutable_top_level_error`]. It is scoped deliberately to triples in
+/// *conjunctive* top-level position, where an empty match provably zeroes the
+/// whole result and no query can be returning correct rows today. It does not
+/// descend into `OPTIONAL`, `MINUS`, `UNION`, or `EXISTS` bodies: those degrade
+/// rather than zero (a vacuous `OPTIONAL` still emits its left side, a vacuous
+/// `MINUS` excludes nothing), and a `UNION` branch may carry a nested `GRAPH`
+/// block that routes and contributes real rows — so refusing there would reject
+/// queries that answer today. Those shapes stay silently degraded pending the
+/// routing work; see the graph-source docs.
+///
 /// Call it *after* the wrap: everything the wrap moved into the graph scope is
 /// then already covered by the rewrite guard, so a query with no `GRAPH` block
 /// of its own keeps refusing on exactly the route it refuses on today, with the
-/// same message.
+/// same message. It also means a surviving top-level triple is itself proof the
+/// wrap declined — no separate flag needed.
 pub(crate) fn guard_graph_source_patterns(
     db: &GraphDb,
     parsed: &fluree_db_query::ir::Query,
@@ -68,11 +80,43 @@ pub(crate) fn guard_graph_source_patterns(
         return Ok(());
     };
     let unsupported = fluree_db_query::r2rml::unsupported_outside_graph_scopes(&parsed.patterns);
-    if unsupported.is_empty() {
-        return Ok(());
+    if !unsupported.is_empty() {
+        return Err(ApiError::Query(
+            fluree_db_query::r2rml::unsupported_subscope_error(gs_id, &unsupported),
+        ));
     }
-    Err(ApiError::Query(
-        fluree_db_query::r2rml::unsupported_subscope_error(gs_id, &unsupported),
+    let unroutable = parsed
+        .patterns
+        .iter()
+        .filter(|p| matches!(p, Pattern::Triple(_)))
+        .count();
+    if unroutable > 0 {
+        return Err(ApiError::Query(unroutable_top_level_error(
+            gs_id, unroutable,
+        )));
+    }
+    Ok(())
+}
+
+/// The refusal for a triple pattern the wrap left stranded at the top level of a
+/// graph-source query.
+///
+/// Unlike the property-path family this is *lowerable* — it would convert to an
+/// R2RML scan perfectly well — but only inside a `GRAPH` scope, and on this plan
+/// it is not in one. "Lowerable in principle" is no comfort to a user whose
+/// pattern will not be lowered on the plan actually running, so it refuses for
+/// the same reason: it would otherwise read the graph source's empty native
+/// index and return no rows as a success. The wording names the workaround
+/// rather than the internals, because both fixes are entirely in the user's
+/// hands.
+fn unroutable_top_level_error(gs_id: &str, count: usize) -> fluree_db_query::QueryError {
+    fluree_db_query::QueryError::InvalidQuery(format!(
+        "graph source '{gs_id}' cannot evaluate {count} top-level triple pattern(s) in this \
+         query. Patterns reach a graph source only inside a `GRAPH <{gs_id}>` block; Fluree \
+         adds that block automatically, but not to a query that already contains a GRAPH \
+         block of its own — so these patterns would read an empty index and silently return \
+         no rows. Move them inside a `GRAPH <{gs_id}>` block, or drop the explicit GRAPH \
+         block so the whole query is scoped to the graph source."
     ))
 }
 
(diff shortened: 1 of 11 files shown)


fluree-db-api/src/view/query.rs
```diff
@@ -41,6 +42,111 @@ pub(crate) fn maybe_wrap_for_graph_source(db: &GraphDb, parsed: &mut fluree_db_q
...
+/// R2RML scan perfectly well — but only inside a `GRAPH` scope, and on this plan
+/// it is not in one. "Lowerable in principle" is no comfort to a user whose
+/// pattern will not be lowered on the plan actually running, so it refuses for
+/// the same reason: it would otherwise read the graph source's empty native
+/// index and return no rows as a success. The wording names the workaround
+/// rather than the internals, because both fixes are entirely in the user's
+/// hands.
+fn unroutable_top_level_error(gs_id: &str, count: usize) -> fluree_db_query::QueryError {
+    fluree_db_query::QueryError::InvalidQuery(format!(
+        "graph source '{gs_id}' cannot evaluate {count} top-level triple pattern(s) in this \
```
bplatz: `guard_graph_source_patterns` is also called from the three Cypher entry points below (lines ~317, ~338, ~402), so a Cypher query against a graph source can hit this message — which tells the user to write `GRAPH <gs_id> { }` and to "drop the explicit GRAPH block." Neither is Cypher syntax, and a Cypher user has no GRAPH block to drop.

Either branch the wording on the query language, or give the Cypher paths their own message.

fluree-db-api/src/view/query.rs
```diff
@@ -41,6 +42,111 @@ pub(crate) fn maybe_wrap_for_graph_source(db: &GraphDb, parsed: &mut fluree_db_q
...
+    parsed: &fluree_db_query::ir::Query,
+) -> Result<()> {
+    if dataset.default.is_empty()
+        || dataset
+            .default
+            .iter()
+            .any(|view| view.graph_source_id.is_none())
+    {
+        return Ok(());
+    }
```
bplatz, line 236: When the dataset holds several *different* graph sources, the error names only `primary()`'s id — so the suggested fix (`GRAPH <that-one-id>`) routes the patterns to one source and silently excludes the others, which isn't what the user asked for.

The gate is right; it's the message that assumes a single source. Worth either naming all of them or dropping the specific id in the multi-source case.

bplatz (commented):
Two inline comments, both on the refusal message rather than the gate logic.

Note for ordering: this and #1631 both change when a dataset gets a default graph, from opposite directions — #1631 stops injecting the ledger for `fromNamed`-only bodies, and `guard_dataset_graph_source_patterns` here keys on `dataset.default` membership. A `fromNamed`-only graph-source query lands with an empty `default`, which takes the `is_empty()` early return. Worth a test covering that combination once both are in.

bplatz (approved):

fluree-db-api/src/view/query.rs
aaj3f: Good catch, and you're right that the wording was wrong — though it turns out the surface it was wrong *for* isn't the one either of us was looking at. Fixed in `a43973cb4` (branch since rebased onto main; head `9ecd76de2` carries the same change).

On Cypher specifically: the call is there in all three entry points exactly like you say, but I don't think it can actually fire from them. Cypher lowering never emits a `Pattern::Graph` — nothing under `fluree-db-cypher/src/lower/` constructs one — so `has_graph_pattern` in `maybe_wrap_for_graph_source` (`query.rs:29`) is always false for a Cypher query, the wrap always takes the whole query, and there's never a top-level pattern left outside a scope for the guard to find. So on those three paths it's a no-op rather than a wrong-message path.

But the concern is real one surface over, which I'd missed: **JSON-LD**. `["graph", "<id>", …]` (`fluree-db-query/src/parse/where_clause.rs:478`) *does* lower to a top-level `Pattern::Graph`, so a JSON-LD query reaches this message and was being told to go write `GRAPH <id> { }` — which is no more JSON-LD than it is Cypher. So the substance of the comment landed, just on a different caller than the one you flagged.

The fix is what you suggested: the message branches on the surface now (a small `QuerySyntax`, derived from `QueryInput` at the SPARQL/JSON-LD entry points and passed explicitly at the Cypher ones). I kept a Cypher arm rather than dropping it, worded honestly ("Cypher has no graph-scoping syntax … query the graph source on its own"), and pinned the unreachability with `cypher_over_a_graph_source_cannot_reach_the_graph_block_advice`. That way, if Cypher ever grows a scoping construct, the test fails and the arm stops being hypothetical — rather than quietly becoming the wrong advice again.

Two things that fell out of chasing this, both of which I'd otherwise have missed:

`unsupported_subscope_error` had the same defect and I'd written it that way myself — it's raised from `GraphOperator` (`graph.rs:314` / `:530`), which has no idea which surface the query arrived on, and it was suggesting "`p`, `p/p`, …". That one can't branch on anything, so I made it language-neutral instead ("fixed-length steps between nodes").

Separately, and I don't think it's in scope here: none of the three Cypher entry points thread an R2RML provider at all — they all land on `execute_view_internal`'s `NoOpR2rmlProvider`, so `has_r2rml_mapping` is false and a Cypher query against a graph source returns silently empty today regardless of any of this. That seems like its own (larger) issue rather than something to fold in here. Happy to write it up if you agree it's one.

fluree-db-api/src/view/query.rs
aaj3f: Agreed — and worse than cosmetic, since the message is pure advice and following it would have silently dropped the other sources. Fixed in `a43973cb4` (branch since rebased onto main; head `9ecd76de2` carries the same change).

`guard_dataset_graph_source_patterns` now collects every default graph's source id (deduped, insertion-ordered) and hands the whole list to the refusal rather than reaching for `primary()`. Both refusals take the list: `unsupported_subscope_error` now takes a `&[&str]` and pluralizes its subject/verb (`graph.rs`'s two call sites pass a one-element slice, since `GraphOperator` is always refusing inside exactly one scope), and the unroutable-triple message names all of them.

The one thing I did beyond naming them: in the multi-source case the message no longer offers "or drop the explicit block so the whole query is scoped to the graph source" as an escape. That advice is fine with one source and actively wrong with several — dropping the block would scope everything to one of them and silently exclude the rest, which is the same failure you're pointing at. So multi-source gets only the "move each pattern into a block naming the source it belongs to" direction. Pinned by `a_multi_source_dataset_refusal_names_every_source`, which asserts all the ids appear and that the drop-the-block escape does not.

aaj3f:
Agree, and I think the early return is the right answer there rather than a gap — but you're right that it was load-bearing behavior sitting on an unexplained `is_empty()` in a boolean chain, which is not where that reasoning should live.

The way I'd put it: once #1631 stops injecting the ledger as a default graph, a `FROM NAMED`-only body genuinely has an empty default graph per §13.2, so its top-level patterns are *supposed* to match nothing. The empty result is the specified answer, not the silent-wrong-answer this guard exists to catch — the thing the guard is for is an unroutable pattern over a source that does hold data. Refusing on an empty default set would reject a conformant query, so `Ok` is correct, and correct for a different reason than the native-member case it was sharing a condition with.

So in `a43973cb4` (branch since rebased onto main; head `9ecd76de2` carries the same change) I split the two: the empty-default case is its own early return with that reasoning written out above it, and the native-member case is its own arm. Test is `a_named_only_graph_source_dataset_is_not_refused`.

On the ordering: the test constructs the named-only `DataSetDb` directly rather than going through dataset resolution, so it doesn't depend on #1631's code and passes on this branch as-is — it should keep holding when #1631 lands rather than needing to be written then. Worth re-running it at the merge of the two regardless, since the whole point is the interaction. Happy to be the one to check that once both are in, whichever lands second.

PR #1633, opened 2026-08-11
forward the query prologue to remote service sub-queries

aaj3f:
The outgoing `SERVICE` sub-query was built from the body source slice alone — no `PREFIX`, no `BASE`. A body written with prefixed names fails loudly at the remote ("Undefined prefix"), which at least surfaces; a body with relative IRIs under a query-level `BASE` is the quiet one — it parses cleanly at the remote, resolves against nothing, and matches a different predicate than the user wrote, returning wrong or empty results as a success. `SERVICE SILENT` converts the loud mode into the quiet presentation too.

The prologue is now rendered from the lowerer's already-resolved prefix map and base (not re-sliced from source), so body bytes are never touched — a literal like `"ex:p is not a curie"` is safe by construction — relative prefix IRIs arrive base-resolved, and the output is deterministic (sorted). `ServicePattern` carries it and one function builds the shipped string.

Tests pin both failure modes red-to-green (the prefixed body now lowers at the remote to the same predicate refs as the parent's own patterns; the `BASE` body resolves to the same absolute IRIs locally and remotely), plus an executor-level mock asserting the exact sub-query text `service.rs` ships, literal safety, the no-prologue case unchanged, and determinism.

### DIFF
--- a/fluree-db-sparql/src/lower/mod.rs
+++ b/fluree-db-sparql/src/lower/mod.rs
@@ -739,22 +739,72 @@ mod tests {
     // against a different base and matched a different predicate.
     // ------------------------------------------------------------------
 
-    /// Lower `sparql` and return its single SERVICE pattern.
-    fn lower_one_service(sparql: &str) -> fluree_db_query::ir::ServicePattern {
+    /// Collect every SERVICE pattern reachable from `patterns`, at any depth.
+    ///
+    /// Walks the nesting containers, not just the top level: the prologue lives
+    /// on `LoweringContext` so it *should* reach a nested SERVICE identically —
+    /// but "should" is what let the original bug survive, since the body slice
+    /// was captured correctly in every position too. Ordered outermost-first,
+    /// then in pattern order, so a two-SERVICE query has a stable [0]/[1].
+    fn collect_services(patterns: &[Pattern]) -> Vec<fluree_db_query::ir::ServicePattern> {
+        let mut found = Vec::new();
+        for p in patterns {
+            match p {
+                Pattern::Service(s) => {
+                    found.push(s.clone());
+                    found.extend(collect_services(&s.patterns));
+                }
+                Pattern::Optional(inner)
+                | Pattern::Minus(inner)
+                | Pattern::Exists(inner)
+                | Pattern::NotExists(inner)
+                | Pattern::Graph {
+                    patterns: inner, ..
+                } => found.extend(collect_services(inner)),
+                Pattern::Union(branches) => {
+                    for branch in branches {
+                        found.extend(collect_services(branch));
+                    }
+                }
+                Pattern::Subquery(sub) => found.extend(collect_services(&sub.patterns)),
+                _ => {}
+            }
+        }
+        found
+    }
+
+    /// Lower `sparql` and return every SERVICE pattern in it, at any depth.
+    fn lower_services(sparql: &str) -> Vec<fluree_db_query::ir::ServicePattern> {
         let ast = parse_sparql(sparql).ast.expect("parse");
         let mut vars = VarRegistry::new();
         let query = lower_sparql_with_source(&ast, &test_encoder(), &mut vars, Some(sparql))
             .expect("lower");
-        let mut services: Vec<_> = query
-            .patterns
-            .iter()
-            .filter_map(|p| match p {
-                Pattern::Service(s) => Some(s.clone()),
-                _ => None,
-            })
-            .collect();
+        collect_services(&query.patterns)
+    }
+
+    /// Lower `sparql` and return its single SERVICE pattern (at any depth).
+    fn lower_one_service(sparql: &str) -> fluree_db_query::ir::ServicePattern {
+        let mut services = lower_services(sparql);
         assert_eq!(services.len(), 1, "expected exactly one SERVICE pattern");
         services.pop().unwrap()
+    }
+
+    /// Assert a SERVICE ships a sub-query whose body resolves to `expected`
+    /// predicates and that carries the prologue those names need.
+    fn assert_prologue_reaches(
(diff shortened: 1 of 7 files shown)


fluree-db-sparql/src/lower/mod.rs
```diff
@@ -694,6 +729,168 @@ mod tests {
...
+    //
+    // `ServicePattern::source_body` is a verbatim slice of the SERVICE block, so
+    // it stops short of the query's PREFIX / BASE declarations. Shipping it bare
+    // changed what the query meant at the remote — loudly for a prefixed name
+    // ("Undefined prefix"), and silently for a relative IRI, which resolved
+    // against a different base and matched a different predicate.
+    // ------------------------------------------------------------------
+
+    /// Lower `sparql` and return its single SERVICE pattern.
+    fn lower_one_service(sparql: &str) -> fluree_db_query::ir::ServicePattern {
```
bplatz: This helper only finds a SERVICE at the top level of `query.patterns`, so every test here exercises the same position. Two shapes the description names but nothing covers:

- **`SERVICE SILENT`** — called out as the case that converts the loud failure into the quiet one, and it has no test. If `silent` ever affects how the sub-query is assembled, nothing catches it.
- **A nested SERVICE** (inside `OPTIONAL` / `UNION` / a sub-SELECT). The prologue lives on `LoweringContext` so it should reach those identically, but "should" is the reason the original bug survived — the body slice was captured correctly in every position too.

A second SERVICE in one query would be worth a line as well, since both must carry the prologue.

bplatz (commented):
Rendering from the resolved environment rather than re-slicing the source is the right call, and I checked the interpolation: `base` and the prefix IRIs both come from parsed IRIREFs, whose grammar already excludes `<`, `>`, quotes and whitespace, so the unescaped `push_str` into `<...>` can't produce malformed or injected output.

One coverage gap inline.

bplatz (approved):

fluree-db-sparql/src/lower/mod.rs
aaj3f: Guilty — the helper filtered `query.patterns` directly, so all six tests sat in the same position and I'd have called that coverage. Thank you.

Your parenthetical is the part I want to underline, because it's the actual lesson here and not just a nit: *"the prologue lives on `LoweringContext` so it should reach those identically, but 'should' is the reason the original bug survived — the body slice was captured correctly in every position too."* That's exactly right, and it's the same shape as the harness blind spot in #1630 — a thing that was structurally invisible to the tests rather than merely untested. I'd been reasoning from "the prologue is context-scoped, therefore position-independent," which is the same species of argument as "the body slice is span-based, therefore position-independent." That one was true and this one happens to be true, but the tests shouldn't be the place we find out.

So: `collect_services` now walks `Optional` / `Union` / `Minus` / `Exists` / `NotExists` / `Graph` / `Subquery`, and a `Service` nested inside another `Service`. Three tests on top of the existing set:

- `remote_service_silent_query_forwards_prologue` — your first bullet. Worth having precisely because it's the position where the bug is *least* visible: with `SILENT`, a dropped prologue turns the remote's `UndefinedPrefix` into an empty result and the query just quietly returns less.
- `nested_remote_service_query_forwards_prologue` — `OPTIONAL`, `UNION`, sub-`SELECT`, and I added `GRAPH` since it's the same container shape.
- `multiple_remote_services_each_forward_prologue` — two SERVICE clauses, two connections, asserting both sub-queries carry the full prologue.

Each one asserts the shipped sub-query lowers to the same predicate `Ref`s the parent's own inner patterns did, rather than just grepping for `PREFIX`, so it's the resolution that's pinned and not the string.

Two things I checked so this isn't just more green:

- The nested test genuinely exercises nesting — I reverted `collect_services` to the old top-level-only filter and it fails with `expected exactly one SERVICE pattern`, so the lowerer isn't flattening `OPTIONAL`/`UNION` up to the top level and the walk is load-bearing.
- With the prologue forwarding disabled, all six forwarding tests go red (four with `UndefinedPrefix { prefix: "ex", … }`, one with `Sid{ns:0}` vs `Sid{ns:100}` — the silent BASE case) while `does_not_rewrite_literals` and `without_prologue_is_unchanged` stay green, which is the split I'd want.

`fluree-db-sparql` and `fluree-db-query` are green at the new head (2083 tests), as are the five `sparql_service_remote_*` end-to-end tests in `fluree-db-api`. I've also rebased onto `main` now that #1624/#1629/#1631/#1634 have landed — clean, with `cargo check` and `cargo clippy` both `--workspace --all-targets` post-restack, since a zero-conflict rebase can still leave a broken tree.

aaj3f:
@bplatz —

**On the interpolation check:** Thanks for actually chasing that down — that was the one thing I'd reasoned about but not written down anywhere, and it's the reason I was comfortable with `push_str` rather than reaching for an escaper. Worth noting the other half for whoever reads this later: the *body* is never re-serialized at all, it ships as the original byte slice, so a literal whose text happens to look like a CURIE or a relative IRI is untouched. That's also the main reason I went with forwarding declarations over expanding the body to absolute form — expansion would have needed an IR→SPARQL serializer we don't have (`explain.rs`'s `PatternDisplay` is human-readable, not round-trippable), and every gap in one would silently ship a subtly different query to the remote. `remote_service_query_does_not_rewrite_literals` pins that property.

**A separate observation, not fixed here:** While writing the red-state check I noticed `parse_sparql` accepts an undefined prefix with **zero** diagnostics; it's the lowerer that rejects it with `UndefinedPrefix`. So "does this parse?" isn't a safe proxy for "is this a valid query," which is the same parse-versus-lower conflation that has misrouted more than one report lately. Out of scope here and I've left it alone, but I think it's worth its own issue — happy to file it if you agree it's real.

PR #1634, opened 2026-08-11
apply rdf set semantics to construct results

aaj3f:
Two mirror-image defects in CONSTRUCT result handling, one root cause: RDF set semantics were applied in a serializer instead of at the graph layer.

The JSON-LD writer deduplicated object values with one set per *subject*, keyed on the serialized object alone — the predicate never entered the key. So `<s> <p1> "same" . <s> <p2> "same"` serialized as one triple, the first predicate claiming the value and every later same-valued predicate on that subject dropping silently; IRI objects dropped the same way, `DESCRIBE` and the JSON-LD `construct` surface share the path, and the loss scales with fan-out (ten predicates carrying one value collapse to one). Meanwhile RDF/XML had no dedup at all and the construct graph was only sorted, never canonicalized — so genuinely duplicate instantiations (a constant template over multiple solutions) were emitted twice, violating the same §16.2 set-union clause from the other direction.

`instantiate_construct_graph` now canonicalizes (sort + dedupe on full `(s, p, o)` term identity — the existing `Term` equality already compares value, datatype, and language), which fixes the RDF/XML leak and makes both serializers agree.

With uniqueness established at the graph layer, the JSON-LD writer's own dedup pass is gone rather than merely rescoped. Keying it per-predicate fixed the reported bug but left it keyed on the *rendered* JSON, and rendering is lossy: the inferable datatypes all collapse to bare scalars, so `"1"^^xsd:integer` and `"1"^^xsd:long` under one predicate are indistinguishable once rendered and the second was dropped — the same class of loss this PR is fixing, on a narrower input. Post-canonicalization a correctly term-keyed dedup could never fire, so the pass was redundant *and* lossy; `dedupe_values` had exactly one setter (`construct_parity`) and that one production consumer, so the flag is retired outright rather than left as a footgun a future caller could re-enable. Set semantics now live in exactly one place, `Graph::canonicalize()`.

Behavior note: JSON-LD CONSTRUCT/DESCRIBE output now includes previously-dropped triples, and RDF/XML no longer repeats duplicates.

Regression tests cover both serializations: same-value-across-predicates for literals and IRIs, true duplicates still collapsing, equal lexical forms with different datatypes staying distinct (including the sharper case where *both* datatypes are inferable and render identically), the fan-out shape, and the DESCRIBE/FQL surfaces. Zero regressions in the existing corpus — nothing had baked in the old behavior — and a note in the compliance doc records that no W3C CONSTRUCT test covers this shape (isomorphism comparison also can't see a doubled triple), which is how it survived.

### DIFF
--- a/fluree-graph-format/src/jsonld.rs
+++ b/fluree-graph-format/src/jsonld.rs
@@ -353,10 +353,13 @@ impl SubjectData {
         let mut node = Map::new();
         node.insert("@id".to_string(), JsonValue::String(self.id.clone()));
 
-        // Track seen values for deduplication
-        let mut seen_values: HashSet<String> = HashSet::new();
-
         for (pred_iri, triples) in self.predicates {
+            // Track seen objects for deduplication, scoped to THIS predicate.
+            // RDF triple identity is (s, p, o), so the same object reached
+            // through a different predicate is a different triple and must
+            // still be emitted; a subject-wide set would drop it.
+            let mut seen_values: HashSet<String> = HashSet::new();
+
             // Check if this predicate is a list (any triple has list_index)
             let is_list = triples.iter().any(|(idx, _)| idx.is_some());
 
--- a/fluree-graph-format/src/jsonld.rs
+++ b/fluree-graph-format/src/jsonld.rs
@@ -6,7 +6,7 @@ use crate::policy::{BlankNodePolicy, ContextPolicy, TypeHandling};
 use fluree_graph_ir::datatype::iri as dt_iri;
 use fluree_graph_ir::{BlankId, Datatype, Graph, LiteralValue, Term};
 use serde_json::{json, Map, Value as JsonValue};
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{BTreeMap, HashMap};
 use std::sync::Arc;
 
 /// Type alias for IRI compaction functions used in JSON-LD formatting.
@@ -30,12 +30,6 @@ pub struct JsonLdFormatConfig {
     /// When false (default), single values are scalars, multiple values are arrays.
     pub multicardinal_arrays: bool,
 
-    /// Deduplicate identical values within a property (CONSTRUCT parity)
-    ///
-    /// When true, duplicate values for the same predicate are removed.
-    /// When false (default), duplicates are preserved.
-    pub dedupe_values: bool,
-
     /// Optional IRI compaction function for vocab positions (predicates and @type values)
     ///
     /// If None, IRIs are output in expanded form.
@@ -54,7 +48,6 @@ impl std::fmt::Debug for JsonLdFormatConfig {
             .field("blank_node_policy", &self.blank_node_policy)
             .field("type_handling", &self.type_handling)
             .field("multicardinal_arrays", &self.multicardinal_arrays)
-            .field("dedupe_values", &self.dedupe_values)
             .field(
                 "compactor_vocab",
                 &self.compactor_vocab.as_ref().map(|_| "<fn>"),
@@ -94,12 +87,6 @@ impl JsonLdFormatConfig {
         self
     }
 
-    /// Enable value deduplication
-    pub fn with_dedupe_values(mut self, enabled: bool) -> Self {
-        self.dedupe_values = enabled;
-        self
-    }
-
     /// Set the IRI compactor function
     pub fn with_compactor<F>(mut self, compactor: F) -> Self
     where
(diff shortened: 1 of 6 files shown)


fluree-graph-format/src/jsonld.rs
```diff
@@ -353,10 +353,13 @@ impl SubjectData {
...
 
-        // Track seen values for deduplication
-        let mut seen_values: HashSet<String> = HashSet::new();
-
         for (pred_iri, triples) in self.predicates {
+            // Track seen objects for deduplication, scoped to THIS predicate.
+            // RDF triple identity is (s, p, o), so the same object reached
+            // through a different predicate is a different triple and must
+            // still be emitted; a subject-wide set would drop it.
+            let mut seen_values: HashSet<String> = HashSet::new();
```
bplatz: Scoping this per-predicate fixes the reported bug, but the set is still keyed on `obj_val.to_string()` — the *rendered JSON-LD*, not the RDF term. Inferable datatypes render as bare JSON scalars, so `"1"^^xsd:integer` and `"1"^^xsd:long` both stringify to `1` and the second is dropped.

That makes the description's claim inexact: with `canonicalize()` now doing real `(s, p, o)` term dedup upstream, this set isn't a "correct no-op safety net" — it's redundant *and* lossy in exactly the direction the PR is fixing, just on a narrower input.

The existing test for "equal lexical forms with different datatypes stay distinct" would only catch this if the two triples share a predicate and both datatypes are in the inferable family; worth checking it does. Fix is either to key on the `Term` or, since `canonicalize()` already guarantees uniqueness, to skip `dedupe_values` on the construct path.

bplatz (commented):
One inline comment. The `canonicalize()` change itself is right and the NaN caveat is correctly noted.

bplatz (approved):

fluree-graph-format/src/jsonld.rs
aaj3f: Great catch, @bplatz — this is right on both counts, and the second one is the part I'd talked myself out of seeing.

Confirmed the mechanism: `format_literal` (`jsonld.rs:490`) renders the whole inferable family as bare JSON scalars, so `LiteralValue::Integer` goes out as `json!(i)` and the `Datatype` is simply gone by the time we key the set. `"1"^^xsd:integer` and `"1"^^xsd:long` under one predicate both render `1`, and the second gets dropped — output is `"http://ex/p1":[1]`. Distinct RDF terms, one survivor, which is exactly the failure this PR is supposed to be fixing.

And you were right to be suspicious of the existing test. `equal_lexical_form_different_datatype_both_survive` pairs `xsd:integer` with `xsd:string`, and `xsd:string` renders as a quoted `"1"` while the integer renders as bare `1` — so the rendered forms differ, the dedup never fires, and the test passes while proving nothing about the case you're describing. Added `equal_rendering_different_datatype_same_predicate_both_survive` with `xsd:integer` + `xsd:long` (both inferable, both render `1`): red before the fix at `[1]`, green after at `[1, 1]`.

On your two fix shapes — I went with the second, and then a bit further. The thing that decided it: after `canonicalize()`, every triple in the graph is distinct by `(s, p, o)`, so within a single `(subject, predicate)` bucket all the objects are already distinct `Term`s by construction. Which means a *correctly* term-keyed set provably can never fire on the construct path — it'd be dead code that still costs a `HashSet` alloc and a hash per object. So the honest choice isn't between "key it on the `Term`" and "drop it"; it's between dead code and lossy code, and dropping it is the only one of those that's actually load-bearing.

Given that, I retired `dedupe_values` outright rather than leaving it set to `false`. It had exactly one setter (`construct_parity`) and that had exactly one production consumer (`construct::format`), so keeping it would mean shipping a public knob whose only behavior is a rendered-JSON dedup — i.e. leaving the footgun in place for whoever enables it next. Set semantics now live in one place, `Graph::canonicalize()`. The unit tests that exercised the flag moved to `graph.canonicalize()` instead, which I think reads better anyway since they now demonstrate the dedup at the layer that actually owns it.

One consequence worth flagging since it's user-visible: that `xsd:integer` / `xsd:long` pair now renders as `[1, 1]`, which looks odd. It is two distinct RDF triples, and JSON-LD's inferable-datatype rendering is lossy, so we can preserve cardinality or preserve a tidy-looking array but not both — I'd rather not silently drop a triple, so I went with cardinality. If you'd rather we render the datatype explicitly (`{"@value": 1, "@type": "xsd:long"}`) when a subject/predicate bucket holds terms that would otherwise render identically, that's a real option and probably the more correct JSON-LD, but it felt like a bigger change than this PR should carry — happy to do it here if you disagree, or take it as a follow-up.

Also updated the PR description — the "correct no-op safety net" line was wrong for precisely the reason you gave, and it's now describing what the code actually does.

PR #1646, opened 2026-08-12
Name terms and the failed constraint in SHACL violation messages

### LINKED ISSUE #1615: SHACL violations report raw SIDs instead of IRIs, and omit the violated constraint component
Two small things in SHACL violation output that, together, sent us down a wrong diagnostic path for a while on **4.1.4**. Both look cheap to fix and neither changes validation behaviour.

## 1. Focus node and Path are rendered as raw SIDs, not IRIs

A validation failure on a node in a non-default namespace reports:

```json
{"error":"Transaction error: SHACL validation failed with 1 violation(s):\n
  1. ServiceEndpoint requires an address\n
     Focus node: 13ep-v3-svc-graphql-node-gspc-air\n
     Path: 13address\n","status":400}
```

`13` is the internal namespace code, concatenated directly onto the local name. The actual values are:

| reported | actual |
|---|---|
| `13ep-v3-svc-graphql-node-gspc-air` | `https://ixfabric.ai/…/discovery/ep-v3-svc-graphql-node-gspc-air` |
| `13address` | `https://ixfabric.ai/…/discovery/address` |

It reads like a corrupted identifier — a leading integer is not a legal IRI and not part of any name we wrote. It is also not reversible from the message alone: nothing in the response says what namespace `13` maps to, so an operator cannot expand it without going to the source. Everything else in the same error path (the `err:` type, the transaction body echo) uses expanded IRIs, so this is the one place a SID leaks to the wire.

Expanding both to full IRIs, or compacting them against the transaction's `@context`, would fix it.

## 2. The violated constraint component is not reported

The message above is our own `sh:message` on the property shape:

```
sh:path disc:address ; sh:datatype xsd:string ; sh:minCount 1 ; sh:maxCount 1 ;
sh:message "ServiceEndpoint requires an address"
```

Per SHACL, one `sh:message` on a property shape covers every constraint on that property — so this exact string is emitted for `minCount`, `maxCount`, **and** `datatype` alike. The subject in question had `disc:address` **×8** against `sh:maxCount 1`. The report says "requires an address" about a node that has eight of them, and there is nothing in the output to distinguish "you supplied none" from "you supplied too many".

`sh:sourceConstraintComponent` is exactly what disambiguates this, and the validator necessarily knows it at the point it constructs the violation. Including it (alongside the user message, not instead of it) would make the report self-diagnosing:

```
1. ServiceEndpoint requires an address
   Focus node: <full-iri>
   Path: <full-iri>
   Constraint: sh:MaxCountConstraintComponent   <- the missing line
   Value count: 8, expected at most 1
```

We'd also happily take just the constraint name without the counts.

## Why it was expensive

The underlying accumulation was ours to fix, and we found it. But the two above meant the error actively pointed away from it: a "missing address" message plus an identifier that looked malformed reads as a serialization or namespace problem in the writer, not as a cardinality violation on stored data. The first thing we checked was whether the field was being dropped from the payload — it never was. Reading the stored subject was what settled it, and the error could have said so directly.

Happy to open a PR for either if that's useful, and to test against a build.

zonotope:
## Problem

A rejected transaction reported its focus node and property path by concatenating a Sid's namespace code onto its local name — `13address` for `https://ixfabric.ai/ns/address`. Nothing was wrong with the data; the formatter printed two unrelated values with no separator between them. The reporter read it as corruption and spent their time hunting a serialization bug that did not exist.

The same message never said which constraint failed. `ValidationResult` has carried `constraint_component` all along, derived per violation from the specific constraint, but no formatter printed it. One `sh:message` commonly covers several constraints on a property, so the message alone cannot distinguish a value that was absent from one that was repeated or the wrong datatype.

## Scope is narrower than the issue reads, and wider than one call site

`/validate` was never affected. It already resolved identifiers through `resolve_sid` and already emitted `sh:sourceConstraintComponent`; so does the CLI's `fluree validate`, which consumes the same resolved report. The defect was confined to *transaction enforcement*.

But four enforcement paths carried it, not one: the JSON transaction path, Turtle inserts, commit replay in `prepare_push`, and `fluree-db-transact::stage_with_shacl` — a public entry point with no in-workspace callers that external embedders reach. All four are fixed.

## Changes

**Identifiers resolve, and compact where there is a context to compact against.** The api layer renders through `IriCompactor::compact_id_sid` using the transaction's own `@context`, so a violation reads `Focus node: ex:alex` / `Path: schema:name` / `Constraint: sh:MinCountConstraintComponent`. Paths below that layer cannot see a context and report full IRIs, matching `/validate`. All three lines compact by the same rule — explicit prefixes and `@base`, never `@vocab` — so no line can emit a bare term where the others cannot.

**Namespaces an operation introduced now resolve.** A property the operation itself brought in is absent from the snapshot until it commits, so it had no IRI to report. `StagedShaclContext` gained `uncommitted_namespaces: Option<&HashMap<u16, String>>`, which each path fills from what it holds: a staging registry's delta for JSON and Turtle, the pushed commits' `namespace_delta`s for replay. Replay accumulates across the decoded commits rather than reading each in isolation, because a later commit's flakes can reference a namespace an earlier one introduced within the same push.

**One layout, per-caller resolution.** `fluree-db-api` and `fluree-db-transact` had near-identical formatters, and `fluree-db-api → fluree-db-transact` is one-directional, so neither could host the shared code. Both depend on `fluree-db-shacl`, which also owns `ValidationResult` and `FocusNode`, so the layout moved there as `report_text::format_violations`, parameterised by the two things only a caller can know: how to resolve a Sid, and how to render the component IRI. `unresolved_sid` moved with it, so the `[unresolved namespace 13]address` fallback has one definition rather than two string literals to keep in step.

**Dead code removed.** `validate_staged_or_error` had no callers anywhere and formatted violations its own way; with it gone, `ShaclError::ValidationFailed` was constructed nowhere and `FocusNode`'s `Display` — which could only ever render `13address`, since a `Display` impl cannot reach the ledger's namespace map — had no users. All three are removed, so nothing left in the tree can render an unresolved Sid to a reader.

## Breaking changes

Three removals from `fluree-db-shacl`'s public surface. Nothing in this workspace used any of them, but external consumers might:

- `ShaclEngine::validate_staged_or_error`
- `ShaclError::ValidationFailed`
- `impl Display for FocusNode` — code formatting a `FocusNode` with `{}` will stop compiling, which is the intended outcome: it was emitting `13address`. Resolve the Sid against a namespace map instead.

## Testing

- `violation_message_names_terms_and_constraint` — the compacted path end to end: focus node, path and constraint component, plus an assertion that a raw namespace code cannot reappear.
- `violation_message_distinguishes_constraints_sharing_one_message` — the reporter's shape exactly: one property carrying minCount, maxCount and datatype under a single `sh:message`, asserting each violation reports its own component while all three carry the identical message. That last assertion is the point — it proves the message stays ambiguous and the component is what disambiguates.
- `turtle_violation_resolves_a_namespace_the_document_introduced` — a Turtle insert declaring a prefix that appears nowhere in the ledger, asserting the focus node resolves rather than falling back.

Each was verified to fail without its fix: removing the `Constraint:` line fails the first two, and reverting the Turtle wiring reproduces `Focus node: [unresolved namespace 15]alex`.

`cargo fmt --check`, `cargo clippy --workspace --all-targets`, and `cargo check --workspace --all-targets` are clean with the `shacl` feature enabled. Suites: api SHACL 78, transact 304, shacl 59.

## Notes for reviewers

**`cargo check -p fluree-db-api` does not cover this.** `shacl` is off by default in that crate, so the code compiles green while broken. A third `StagedShaclContext` construction site in `commit_transfer.rs` only surfaced once the feature was enabled — enable it, or check the workspace.

**Warn-mode violations now log resolved IRIs.** `tracing::warn!` previously carried the opaque `13address`; it now carries subject IRIs, which can encode customer identifiers. That is the intended improvement, but it puts them into log sinks rather than only into the response to a caller who already had the data.

**Commit replay has no test of its own.** Reaching `prepare_push` needs a follower holding decoded commits with a namespace delta plus a shape that rejects one — considerable scaffolding for the same resolution path the Turtle test already pins. The wiring is one field at one call site.

**`[unresolved namespace 13]address` should be unreachable.** It means a namespace code is absent from both the snapshot and whatever the operation has yet to commit, which is corruption or a bug rather than a missing prefix. It is deliberately not IRI-shaped.

Fixes #1615

### DIFF
--- a/fluree-db-api/src/tx.rs
+++ b/fluree-db-api/src/tx.rs
@@ -3166,7 +3166,10 @@ impl crate::Fluree {
                 // resolve_cross_ledger_shapes_for_tx here when
                 // the use case lands.
                 cross_ledger_shapes: None,
-                staged_ns: None,
+                // Carries the prefixes this document declared, which the
+                // snapshot has not seen yet — without it a violation on a
+                // predicate the document introduces has no IRI to report.
+                staged_ns: Some(&ns_registry),
                 // Turtle carries its prefixes in the document, not as a
                 // JSON-LD context the API sees, so violations name full IRIs.
                 txn_context: None,
--- a/fluree-db-api/src/tx.rs
+++ b/fluree-db-api/src/tx.rs
@@ -1141,7 +1141,7 @@ fn format_violations(
     // and saying so beats emitting something that looks like an IRI.
     let render = |sid: &Sid| {
         compactor.compact_id_sid(sid).unwrap_or_else(|_| {
-            format!("<unresolved namespace {}>{}", sid.namespace_code, sid.name)
+            format!("[unresolved namespace {}]{}", sid.namespace_code, sid.name)
         })
     };
 
--- a/fluree-db-api/src/tx.rs
+++ b/fluree-db-api/src/tx.rs
@@ -572,11 +572,19 @@ pub(crate) struct StagedShaclContext<'a> {
 
     /// Staged `NamespaceRegistry` — D's snapshot namespaces plus
     /// any IRIs the in-flight transaction has registered. Required
-    /// when `cross_ledger_shapes` is `Some`, where it is the term
-    /// context for compiling M's wire-form shapes against D. Also
-    /// resolves identifiers in violation messages, so a property the
-    /// transaction itself introduced still renders as an IRI.
+    /// when `cross_ledger_shapes` is `Some`; consulted as the term
+    /// context for compiling M's wire-form shapes against D.
     pub staged_ns: Option<&'a fluree_db_transact::namespace::NamespaceRegistry>,
+
+    /// Namespace codes this operation introduced, which the snapshot will not
+    /// carry until it commits. Violation messages resolve against these on top
+    /// of the snapshot's, so a term the operation itself brought in still has
+    /// an IRI to report rather than a bare code.
+    ///
+    /// Every path holds these somewhere of its own — a staging registry's
+    /// delta, or the namespace deltas of the commits being replayed — so this
+    /// is the map rather than any one path's carrier for it.
+    pub uncommitted_namespaces: Option<&'a HashMap<u16, String>>,
 
     /// The JSON-LD context the transaction supplied, used to compact
     /// identifiers in violation messages to the terms the author wrote.
@@ -1091,10 +1099,14 @@ fn violation_iri_compactor(
     use crate::format::IriCompactor;
 
     let base = view.base().snapshot.shared_namespaces();
-    let namespace_codes = match ctx.staged_ns.map(NamespaceRegistry::delta) {
-        Some(delta) if !delta.is_empty() => {
+    let namespace_codes = match ctx.uncommitted_namespaces {
+        Some(uncommitted) if !uncommitted.is_empty() => {
             let mut merged = (*base).clone();
-            merged.extend(delta.iter().map(|(code, prefix)| (*code, prefix.clone())));
+            merged.extend(
+                uncommitted
+                    .iter()
+                    .map(|(code, prefix)| (*code, prefix.clone())),
+            );
             std::sync::Arc::new(merged)
         }
         _ => base,
@@ -1345,7 +1357,8 @@ async fn stage_with_config_shacl(
                     crate::cross_ledger::GovernanceArtifact::Shapes(wire) => Some(wire),
                     _ => None,
                 }),
-            staged_ns: Some(&ns_registry),
+            staged_ns: cross_ledger_shapes.as_deref().map(|_| &ns_registry),
+            uncommitted_namespaces: Some(ns_registry.delta()),
             txn_context,
             inline_shape_bundle,
             cross_ledger_schema,
@@ -3171,10 +3184,11 @@ impl crate::Fluree {
                 // resolve_cross_ledger_shapes_for_tx here when
                 // the use case lands.
                 cross_ledger_shapes: None,
-                // Carries the prefixes this document declared, which the
-                // snapshot has not seen yet — without it a violation on a
-                // predicate the document introduces has no IRI to report.
-                staged_ns: Some(&ns_registry),
+                staged_ns: None,
+                // The prefixes this document declared, which the snapshot has
+                // not seen yet — without them a violation on a predicate the
+                // document introduces has no IRI to report.
+                uncommitted_namespaces: Some(ns_registry.delta()),
                 // Turtle carries its prefixes in the document, not as a
                 // JSON-LD context the API sees, so violations name full IRIs.
                 txn_context: None,
--- a/fluree-db-api/src/tx.rs
+++ b/fluree-db-api/src/tx.rs
@@ -572,9 +572,17 @@ pub(crate) struct StagedShaclContext<'a> {
 
     /// Staged `NamespaceRegistry` — D's snapshot namespaces plus
     /// any IRIs the in-flight transaction has registered. Required
-    /// when `cross_ledger_shapes` is `Some`; consulted as the term
-    /// context for compiling M's wire-form shapes against D.
+    /// when `cross_ledger_shapes` is `Some`, where it is the term
+    /// context for compiling M's wire-form shapes against D. Also
+    /// resolves identifiers in violation messages, so a property the
+    /// transaction itself introduced still renders as an IRI.
     pub staged_ns: Option<&'a fluree_db_transact::namespace::NamespaceRegistry>,
+
+    /// The JSON-LD context the transaction supplied, used to compact
+    /// identifiers in violation messages to the terms the author wrote.
+    /// `None` where the request carries no context — Turtle inserts and
+    /// commit replay — and those messages carry full IRIs instead.
+    pub txn_context: Option<&'a serde_json::Value>,
 
     /// Inline shape bundle parsed from `txn.opts.shapes` against the
     /// staged namespace registry. When `Some`, the bundle attaches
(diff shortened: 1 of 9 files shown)


bplatz (approved):
Approving — the consolidation into `fluree-db-shacl::report_text` with caller-supplied resolution is the right shape, and the tests are well-targeted. Verified locally: `cargo check --workspace --all-features --all-targets` is clean, clippy is clean on the three touched crates, and all 93 SHACL-related tests pass.

Two things I'd like looked at more carefully before this goes in, though — the first in particular.

### 1. The policy-enabled write path still drops the authoring context

`fluree-db-api/src/tx.rs:2556` passes `txn_context: None` with the comment "Txn-based entry: no source document, so no authoring context to compact violation messages against." But that call site is `stage_transaction_tracked_with_policy`, which *does* have the source document — `input.txn_json`, which `parse_transaction` consumes about 30 lines earlier. (The identical comment at :2340 is accurate; that one is inside `stage_view_once`, which only ever receives a `Txn`.)

Confirmed empirically — same shape, same transaction, same `@context`, the only difference being whether a policy is attached:

```
no policy:    Focus node: ex:alex
              Path: schema:name
              Constraint: sh:MinCountConstraintComponent

with policy:  Focus node: http://example.org/ns/alex
              Path: http://schema.org/name
              Constraint: http://www.w3.org/ns/shacl#MinCountConstraintComponent
```

The callers are `tx_builder.rs:764` and `graph_transact_builder.rs:281`, i.e. any transaction with a policy attached — which on a deployed server is the normal case, and plausibly the configuration #1615 was reported from. The fix looks like a one-liner (`input.txn_json.get("@context")`), but it's worth a test that goes through the policy path, since none of the three new tests do.

### 2. Docs

`docs/guides/cookbook-shacl.md:609` still reads "The formatted report lists each violation's focus node, property path, and message" — no constraint component, and no mention that identifiers now compact against the transaction's context and fall back to full IRIs when there isn't one.

fluree-db-api/src/tx.rs
```diff
@@ -2458,10 +2547,18 @@ impl crate::Fluree {
...
-            stage_with_config_shacl(ledger, txn, ns_registry, options, &mut resolve_ctx)
-                .await
-                .map_err(|e| TrackedErrorResponse::new(400, e.to_string(), tracker.tally()))?;
+        let (view, ns_registry) = stage_with_config_shacl(
+            ledger,
+            txn,
+            ns_registry,
+            options,
+            &mut resolve_ctx,
+            // Txn-based entry: no source document, so no authoring
```
aaj3f: **Optional (seconding @bplatz, with a wider blast radius).** I traced his finding to confirm it and want to add one data point: it's not just the two builder callers. `stage_transaction_tracked_with_policy` is also what `transact_tracked_with_policy` (`tx.rs:2622`) routes through, and that's the tracked server write entry — so on a deployed server with policy enforcement on, essentially *every* write takes the path where the `@context` gets dropped and violations name full IRIs instead of the author's terms. The mechanism is exactly as Brian described: `parse_transaction(input.txn_json, …)` at `tx.rs:2521` proves the source document is in hand, and the comment at `tx.rs:2554-2555` ("no source document") is simply inaccurate for this entry point. It seems the fix really is the one-liner (`input.txn_json.get("@context")` — `txn_json` is a `&JsonValue` field, still borrowable after the partial move of `input.policy` into `options`), plus the policy-path test Brian asked for, since none of the three new tests attach a policy. To be clear about severity: the #1615 corruption (`13alex`) is fixed on this path too — full IRIs are *correct*, just not compacted — which is why I think this stays a follow-through item rather than a blocker.

aaj3f: **Optional (docs, real location `docs/guides/cookbook-shacl.md:609` — commenting here because that file is not in this diff).** Seconding @bplatz's second note: the cookbook still says the report lists "focus node, property path, and message" — no `Constraint:` line, and no mention that identifiers now compact against the transaction's context with a full-IRI fallback. One sentence there would close it.

fluree-db-shacl/src/report_text.rs
```diff
@@ -0,0 +1,89 @@
...
+
+use crate::compile::Severity;
+use crate::validate::{FocusNode, ValidationResult};
+
+/// Render a `Sid` no namespace map could resolve.
+///
+/// Reached only when a namespace code is absent from both the snapshot and
+/// whatever the operation has yet to commit, which means corruption or a bug
+/// rather than a missing prefix. Says so, rather than emitting something a
+/// reader would mistake for an identifier.
```
aaj3f, line 28: **Praise.** The `[unresolved namespace {code}]{name}` marker is exactly right — the module doc's observation that an unresolvable code means corruption-or-bug rather than a missing prefix, and that the output should *say so* instead of emitting something a reader mistakes for an identifier, is the lesson of #1615 applied all the way down. Same for the caller-supplied-resolution design: layout lives once, and the crate that has no handle on the namespace map doesn't pretend it can resolve.

fluree-db-api/src/tx.rs
```diff
@@ -1044,41 +1060,95 @@ pub(crate) async fn apply_shacl_policy_to_staged_view(
     .await?;
 
     // 6. Apply per-graph mode: warn violations log, reject violations fail.
+    //    The compactor is built only when there is something to report — it
+    //    merges namespace maps and parses the context, neither of which the
+    //    conforming path should pay for.
+    if outcome.conforms() {
```
aaj3f, line 1066: **Praise.** The `conforms()` early-return before building the compactor means the conforming path pays for none of this — namespace-map merge, context parse, all of it sits strictly on the rejection path. That's the right shape for error-path-only formatting work.

fluree-db-shacl/src/validate.rs
```diff
@@ -3045,15 +3003,6 @@ impl FocusNode {
...
-        match self {
-            FocusNode::Node(sid) => write!(f, "{}{}", sid.namespace_code, sid.name),
-            FocusNode::Literal(lit) => write!(f, "{}", lit.value),
-        }
-    }
-}
-
 /// Individual validation result
 #[derive(Debug, Clone)]
 pub struct ValidationResult {
```
aaj3f, line 3008: **Optional, pre-existing.** Not introduced by this PR: the engine's *default* messages still name nested shapes by bare local name (`nested.id.name` at `validate.rs:1261` and `:1391`), so a default message can still contain an unresolved-looking term even though the Focus/Path/Constraint lines now resolve properly. Tiny, and arguably fine since these are shape-authoring diagnostics — happy to leave it, just noting it so nobody re-reports #1615 off a default message.

*Commenting here because `fluree-db-shacl/src/validate.rs:1261` is not in this diff — anchored to `ValidationResult`, the type whose default messages this concerns.*

aaj3f (approved):
I'll second both of @bplatz's notes rather than repeat them, with one addition on the first: the policy-path context drop at `tx.rs:2556` is wider than the two builder callers — `transact_tracked_with_policy` (`tx.rs:2622`), the tracked server write entry, routes through the same `None`, so on a deployed server with policy on, essentially every rejected write reports full IRIs instead of the author's terms. The `13alex` corruption is fixed there too (full IRIs are correct, just un-compacted), which is why I agree this doesn't block — but the comment at `tx.rs:2554` claiming "no source document" is inaccurate for that entry point, the fix looks like the one-liner Brian named, and none of the three new tests attach a policy, so it'd be good to fold in the fix + a policy-path test before or right after merge.

Approving — already unblocked by Brian's approval, and I agree with the verdict; just be sure the `tx.rs:2556` one-liner + a policy-path test land before this ships in a release, since with policy on it's the message quality every deployed server actually sees.

bplatz:
Pushed `9bfe7dbc1` covering both review notes, @zonotope.

**The policy path now compacts** (`tx.rs:2556`). It was the one-liner Brian identified — `input.txn_json.get("@context")` — and the comment claiming "no source document" was inaccurate for this entry point; it reads like a copy of the accurate one at `:2340`, which sits inside `stage_view_once` and really does only receive a `Txn`. Comment corrected alongside.

@aaj3f's blast-radius note checks out: `stage_transaction_tracked_with_policy` has three callers, and `transact_tracked_with_policy` (`tx.rs:2622`) is one of them, so this was the message quality on the tracked server write path — the normal deployed configuration, and plausibly the one #1615 was reported from.

**A policy-path test** (`violation_message_compacts_terms_on_the_policy_path`). The policy is permissive on purpose: it has to admit the write so the transaction reaches SHACL and fails *there*, since a denying policy would reject first and never produce a violation message to inspect. I reverted the one-liner to check the test goes red, so it pins the path rather than the formatter the existing three already cover.

One incidental thing worth knowing, though I've left it alone as out of scope: on this path the violation arrives as a stringified `Http { status: 400 }` rather than the typed `ShaclViolation`, because the tracked entry converts errors through `TrackedErrorResponse`. The text is identical, so the shared extractor accepts either — but an embedder matching on `TransactError::ShaclViolation` won't catch a policy-path violation. Might be worth its own look.

**The cookbook** now shows the real report shape including `Constraint:`, and states the compaction rule — explicit prefixes and `@base`, never `@vocab` — plus the full-IRI fallback where there is no context, so the Turtle and replay behaviour isn't a surprise.

Verified: 79 api SHACL tests, 59 `fluree-db-shacl` tests, fmt clean, clippy clean on the three touched crates with `--all-features --all-targets -D warnings`.

PR #1647, opened 2026-08-13
fix(materialize): skip materialize targets that already applied the window

christophediprima:
A templated job's watermark is shared across every ledger it fans into, and it must not advance while any target is behind — otherwise the laggards skip that window permanently. That is correct, and this PR does not change it.

What it costs today is that the targets which **did** commit earn no credit. The next poll re-reads the same window and re-commits all of them, and it keeps doing that for as long as one target cannot fit.

This is the follow-through on a note I left in `r2rml_materialize.rs` when per-target failure isolation landed in #1422:

> Note the loop already commits per target as it goes, so partial application was ALWAYS the reality. What was missing is recording which targets succeeded.

That PR made partial application survivable; it did not make it *recorded*. This records it. A resolved target that already holds the window is skipped instead of rewritten.

## What it cost us

A 17-source job fanning out per `(tenant, user)` into 23 ledgers, one of which has a window larger than the novelty ceiling:

- `targets_ok=19 targets_deferred=4` on **22 consecutive polls**, byte-identical remainder each time (`items_deferred=421708`)
- the job's watermark was **never written once in 13.7 h**
- so 19 ledgers were re-committed every ~4 minutes for 13.7 h

The re-commits are idempotent but they are still writes. They filled a 100 GiB volume: **479 `No space left on device` failures across 8 target ledgers in 13 minutes**. That state does not self-heal — index GC has to write in order to reclaim anything, so at zero bytes free it cannot run.

After this change, on the same deployment and the same data:

| | before | after |
|---|---|---|
| volume used | 119 GiB, climbing | **79 GiB**, flat |
| CPU | 2359m | **851m** |
| per-poll targets | `ok=19 deferred=4` | `ok=22 deferred=1 skipped=22` |

The deferral count dropping 4 → 1 was not predicted: removing the rewrite churn relieved enough novelty pressure that three of the four stuck targets got through on their own.

## Design

**Additive.** A new `urn:fluree:materialize#appliedSnapshotId` marker per `(source, RESOLVED target, table)`. The shared watermark still chooses the scan window, so the source read is byte-for-byte unchanged. A state ledger written before this simply has no markers, so the first poll behaves exactly as it did and every poll after is cheap.

**Why not per-target scan watermarks**, which is the obvious alternative: taking `min()` across targets as the window start is wrong the moment a target is created. A newly-created `(tenant, user)` ledger has no watermark, so `min()` over the existing ones starts the scan *after* rows that target still needs, and it silently never receives them. Detecting that mid-window and forcing a re-scan is real complexity, and it buys nothing here — per-target watermarks fix the *rewrite*, which this fixes more cheaply, not the *volume*.

**Two orderings, both load-bearing:**

- Markers are written **before** the partial-window early returns. A job whose watermark cannot advance is precisely the job that needs them, and both returns sit between that point and the watermark write. Putting this after them would record progress only for jobs that never needed it.
- Markers are written **after** the target's data commit. A crash in the gap re-applies the window, which is idempotent; a marker written first would skip rows that never landed. Same rule as the watermark, one level down.

**Edges decided rather than left open:**

- A target must be caught up on **every** table in the window, not any. A target that applied one table of a two-table window still owes the other, and skipping it there would drop half a window silently — the only way this change could lose data rather than repeat work.
- An **empty** window marks nothing caught up. `all()` over an empty slice is vacuously true, so without an explicit guard a source with no snapshots yet would skip every target forever and materialize nothing.
- `==`, never an inequality. Both sides are Iceberg **snapshot ids**, which the spec assigns randomly — there is no ordering between two of them, not even between a snapshot and its own parent. An earlier revision of this PR compared them with `>=`, which turns the marker into a one-way ratchet: it climbs the running maximum of a random sequence and after roughly `ln(n)` windows exceeds every later draw permanently. That ran on a 17-source deployment, where every marker ratcheted past every subsequent window — all 17 sources skipped on every poll, each skip counted as an ok target, nothing logged, the shared watermark advancing past data that was never applied, and about 80% of the source entities discarded. Equality is what the question actually asks, and it covers the amplification this PR exists to stop, because a job whose watermark cannot advance re-presents the *same* `to` every poll. If "at or beyond" is ever genuinely needed it requires a quantity that IS ordered — the Iceberg sequence number — stored alongside the snapshot id. Full account in `9820a68ad` and on `target_is_caught_up`.
- Unreadable markers are **non-fatal** and cost only the skip. Re-applying is idempotent, and failing here would turn a state-ledger hiccup into a stalled job.
- `force_full` ignores the markers, for the same reason it ignores the watermark: the caller is asking for a rebuild, not a resume.

## Testing

`fmt` and `clippy --all-features --all-targets -- -D warnings` clean. `fluree-db-api --lib` 1074 pass, 0 fail.

Six tests: five on the predicate and one driving the engine end to end over a re-presented window — same rows, same `to_snapshot_id`, which is exactly what a job whose watermark cannot advance sees on every poll. `FakeSource` drains its batches by design, so that needs an opt-in `repeating()` variant rather than a change to the default.

Four mutations: dropping the non-empty guard, `all` → `any`, `==` → `>=`, and **removing the skip branch entirely**. Each is killed in isolation — the first, second and fourth by exactly one test, and `==` → `>=` by two (the equality predicate test and `a_marker_numerically_above_a_later_snapshot_is_not_caught_up`, which carries five real (marker, later-snapshot) pairs from the incident rather than a sequence-shaped fixture; an ordered fixture would ratify an ordering comparison). Removing the skip branch is the one worth dwelling on — the predicate tests all stayed green against it, because they only exercise a pure function. Only the engine test catches it, and it does so on the target ledger's `t`: `assertion left == right failed: people_acme:main already applied this window; a second commit means the rewrite is back / left: 3 / right: 2`. That mutation was re-run after the rebase onto `a85e03682`, with the same result.

Two assertions were wrong before this landed, and both looked authoritative:

- `subjects_upserted` is `live.len()`, the subjects the accumulator *prepared*, read before the target loop runs. It reads 3 whether or not a single commit happened.
- Reading the target **data** proves nothing either: re-application is idempotent, so it is identical whether the target was skipped or rewritten.

The target ledger's `t` is the one observable that separates them, because it only moves on a real commit. Removing the skip advances it 2 → 3 and fails the test.

## Not in this PR

This stops the livelock **consuming disk**. It does not make an oversized target complete — four targets in our deployment still need ~45 MB in a single pass and still cannot get it, so that table remains without a watermark. Fixing that needs sub-window progress, which is a separate change and a separate discussion: subdividing the window at snapshot boundaries needs no new vocabulary at all, but does nothing when the volume sits in one snapshot, and the general case needs a per-target file cursor. Happy to follow up if you have a view on how progress should be represented.

### DIFF
--- a/fluree-db-api/src/graph_source/r2rml_materialize.rs
+++ b/fluree-db-api/src/graph_source/r2rml_materialize.rs
@@ -740,6 +740,23 @@ impl Fluree {
             deferred: deferred.len(),
             failed: failed.len(),
         };
+        // Report skips on EVERY pass, not only an incomplete one.
+        //
+        // An all-skip pass is `is_complete()`, so it would otherwise be the ONE outcome
+        // that prints nothing whatsoever — and that is precisely the shape a
+        // wrongly-skipped target takes. `all_skipped` is the specific alarm: nothing was
+        // written this pass, yet the watermark is about to advance. That is legitimate
+        // for a genuinely idle window and a data-loss signature otherwise, so it has to
+        // be visible either way rather than inferred by diffing a target ledger's `t`.
+        if skipped_targets > 0 {
+            info!(
+                source = %source_graph_source_id,
+                targets_skipped = skipped_targets,
+                targets_ok = tally.ok,
+                all_skipped = skipped_targets == tally.ok && tally.is_complete(),
+                "materialize: targets already held this window; nothing written for them"
+            );
+        }
         if !tally.is_complete() {
             info!(
                 targets_ok = tally.ok,
@@ -1129,6 +1146,39 @@ fn applied_node(
 /// applied one table of a two-table window still has work, so it must not be
 /// skipped — that is the difference between skipping a no-op and silently dropping
 /// half a window.
+///
+/// # Why this is `==` and must not become an inequality
+///
+/// `applied` and `to` are Iceberg **snapshot ids**, and the spec assigns those
+/// randomly: there is no ordering between two snapshot ids, not even between a
+/// snapshot and its own parent. `a >= to` therefore compares two unrelated random
+/// i64s and calls the answer "has this target already applied the window".
+///
+/// That comparison makes the marker a one-way ratchet. It is only rewritten when a
+/// window's `to` happens to compare greater, so it climbs the running maximum of a
+/// random sequence and then exceeds every later draw permanently — after roughly
+/// `ln(n)` windows a target stops being written to at all. Measured on a 17-source
+/// deployment: every marker had ratcheted to between 8.19e18 and 9.22e18 against an
+/// `i64::MAX` of 9.223e18, so all 17 were skipped on every poll, each skip was
+/// counted as an ok target, and the watermark advanced past data that was never
+/// applied. About 80% of the entities in the source lakehouse were discarded.
+///
+/// Equality is what the question actually asks, and it fully covers the case this
+/// skip exists for: a job whose watermark cannot advance re-presents the SAME `to`
+/// on every poll, so equality matches and the skip fires.
+///
+/// The two cases an inequality was reaching for both survive equality:
+///
+/// - A re-poll of an older snapshot: equality declines to skip, so the target
+///   re-applies. Re-application is idempotent, so the cost is repeated work, never
+///   corruption — strictly safer than skipping data that was never applied.
+/// - A window that moved backwards after a forced full: moot, because `force_full`
+///   discards the markers before the target loop runs.
+///
+/// If "at or beyond" is ever genuinely needed it requires a quantity that IS
+/// ordered — the Iceberg **sequence number**, which increases by one per commit —
+/// which means storing the sequence alongside the snapshot id in the marker. Do not
+/// reintroduce an inequality over snapshot ids.
 fn target_is_caught_up(
     applied: &std::collections::HashMap<(String, String), i64>,
     target: &str,
@@ -1138,7 +1188,7 @@ fn target_is_caught_up(
         && table_watermarks.iter().all(|(table, _from, to)| {
             applied
                 .get(&(target.to_string(), table.clone()))
-                .is_some_and(|a| *a >= *to)
+                .is_some_and(|a| *a == *to)
         })
 }
 
@@ -2948,27 +2998,87 @@ mod tests {
         assert!(!target_is_caught_up(&applied, "silver_acme_u2:main", &tw));
     }
 
-    /// `>=`, not `==`: a marker ahead of this window (a re-poll of an older snapshot,
-    /// or a window that moved backwards after a forced full) still means "has it".
-    /// Equality here would re-apply and rebuild the amplification.
+    /// `==`, and only `==`. A marker that merely DIFFERS from this window's `to` — in
+    /// either numeric direction — has not applied this window.
     #[test]
-    fn an_applied_marker_at_or_beyond_the_window_counts_as_caught_up() {
+    fn only_an_applied_marker_equal_to_the_window_counts_as_caught_up() {
         let tw = vec![("t".to_string(), None, 20i64)];
         assert!(target_is_caught_up(
             &applied_map(&[("x:main", "t", 20)]),
             "x:main",
             &tw
         ));
-        assert!(target_is_caught_up(
-            &applied_map(&[("x:main", "t", 21)]),
-            "x:main",
-            &tw
-        ));
+        assert!(
+            !target_is_caught_up(&applied_map(&[("x:main", "t", 21)]), "x:main", &tw),
+            "a marker numerically above `to` is a DIFFERENT snapshot, not a later one"
+        );
         assert!(!target_is_caught_up(
             &applied_map(&[("x:main", "t", 19)]),
             "x:main",
             &tw
         ));
+    }
+
+    /// Snapshot ids are random, so an applied marker can be numerically far larger
+    /// than a snapshot committed long after it. Skipping on `>=` therefore ratchets:
+    /// the marker climbs the running maximum of a random sequence and then exceeds
+    /// every later draw permanently, and the target is never written to again.
+    ///
+    /// These are real values from a deployment where that happened — all 17 markers
+    /// had reached 8.19e18–9.22e18 against an `i64::MAX` of 9.223e18, every target was
+    /// skipped on every poll, and the watermark advanced past data that was never
+    /// applied. Each `to` below is a genuinely LATER commit than the marker beside it,
+    /// despite being numerically smaller.
+    ///
+    /// Small sequence-shaped fixtures cannot catch this — an ordered fixture will
+    /// ratify an ordering comparison — which is why these are the production numbers.
+    #[test]
+    fn a_marker_numerically_above_a_later_snapshot_is_not_caught_up() {
+        // (table, ratcheted applied marker, the genuinely later snapshot it skipped)
+        let observed = [
+            (
+                "silver.place",
+                9_220_834_252_869_770_488i64,
+                5_644_295_785_472_712_989i64,
+            ),
+            (
+                "silver.observation",
+                9_217_340_116_954_323_563,
+                3_238_671_374_642_079_740,
+            ),
+            (
+                "silver.concept_scheme",
+                9_086_429_568_298_186_029,
+                753_340_878_885_049_461,
+            ),
+            (
+                "silver.concept",
+                9_052_937_455_619_636_065,
+                1_349_575_232_631_351_152,
+            ),
+            (
+                "silver.link",
+                9_196_238_952_972_422_466,
+                9_139_397_570_598_290_786,
+            ),
+        ];
+        for (table, marker, later_to) in observed {
+            assert!(
+                marker > later_to,
+                "{table}: this fixture is only meaningful while the marker is \
+                 numerically larger than the later snapshot — that is the whole trap"
+            );
+            let tw = vec![(table.to_string(), None, later_to)];
+            assert!(
+                !target_is_caught_up(
+                    &applied_map(&[("silver_acme_u1:main", table, marker)]),
+                    "silver_acme_u1:main",
+                    &tw
+                ),
+                "{table}: marker {marker} is a different snapshot from {later_to}, not \
+                 a later one — skipping here is what discarded the window"
+            );
+        }
     }
 
     /// ALL tables, not any. A target that applied one table of a two-table window
--- a/fluree-db-api/src/graph_source/r2rml_materialize.rs
+++ b/fluree-db-api/src/graph_source/r2rml_materialize.rs
@@ -100,6 +100,36 @@ const WATERMARK_TARGET_PRED: &str = "urn:fluree:materialize#target";
 /// Predicate recording which source table the watermark belongs to.
 const WATERMARK_TABLE_PRED: &str = "urn:fluree:materialize#table";
 
+/// Subject-IRI prefix for a per-(source, RESOLVED target, table) applied marker.
+///
+/// Distinct from [`WATERMARK_SUBJECT_PREFIX`] because the two are keyed on
+/// different things and must never collide: the watermark is keyed on the target
+/// SPEC (one per job, `silver_{tenant}_{user}:main`), this on the RESOLVED ledger
+/// (`silver_acme_u1:main`). For a non-templated job those strings are equal, so a
+/// shared prefix would put both on the same subject.
+const APPLIED_SUBJECT_PREFIX: &str = "urn:fluree:materialize-applied:";
+/// Predicate holding the snapshot id a single resolved target has fully applied
+/// (string-encoded, like the watermark, to preserve i64 precision).
+///
+/// WHY THIS EXISTS. The watermark is shared across every target a templated job
+/// fans into, and it may not advance while ANY target is behind — otherwise the
+/// laggards would skip that window permanently. Correct, but it means the targets
+/// that DID commit earn no credit: the next poll re-reads the window and
+/// re-commits all of them, forever, for as long as one target cannot fit.
+///
+/// Measured in production: 19 of 23 targets committing and 4 deferring on every
+/// poll, unchanged across 22 consecutive polls, so 19 targets were rewritten every
+/// ~4 minutes for 13.7 h. The re-commits are idempotent but not free — they are
+/// writes, and they filled a 100 GiB volume to ENOSPC, which is unrecoverable
+/// because GC needs to write in order to reclaim.
+///
+/// This marker records what each resolved target has already applied, so a target
+/// that is caught up is SKIPPED rather than rewritten. It deliberately does not
+/// touch the scan window: the shared watermark still chooses it, so the source read
+/// is unchanged and there is no risk of starting a scan after rows a newly-created
+/// target still needs.
+const APPLIED_SNAPSHOT_PRED: &str = "urn:fluree:materialize#appliedSnapshotId";
+
 /// Subject-IRI prefix for a persisted tracking job. The full subject is
 /// `{PREFIX}{source}:{target-spec}`, matching the worker's own
 /// `(source, target)` job key.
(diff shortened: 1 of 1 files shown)


christophediprima:
Pushed one more commit, and it fixes a data-loss bug in what I originally proposed here. Flagging it prominently rather than quietly amending, because the failure mode is nasty and the reasoning is worth having in the thread.

`target_is_caught_up` compared the applied marker against the window with `applied >= to`. Both are Iceberg snapshot ids — and snapshot ids are randomly assigned, with no ordering between them, not even between a snapshot and its own parent. So the check was comparing two unrelated 64-bit numbers.

That turns the marker into a one-way ratchet. It is only rewritten when a window's `to` happens to compare greater, so it climbs the running maximum of a random sequence and then exceeds every later draw permanently. After roughly `ln(n)` windows the target is skipped forever.

We hit this on a 17-source deployment. Every marker had ratcheted to between 8.19e18 and 9.22e18 against an `i64::MAX` of 9.223e18, so all 17 sources were skipped on every poll — each skip counted as an ok target, nothing logged, and the shared watermark advancing past data that was never applied. 3,505 rows in one source table against 102 in the target. No warning anywhere in three days of logs; the only observable that disagreed was the target ledger's `t`, which never moved.

The fix is `==`. That is what the predicate's own doc comment already claimed — "Already has this whole window" — and it fully covers the amplification this PR exists to stop, because a job whose watermark cannot advance re-presents the *same* `to` on every poll, so equality matches and the skip still fires. The two cases my old comment cited as needing an inequality both survive it: a re-poll of an older snapshot now re-applies, which is idempotent and strictly safer than skipping data that was never applied; and a window moving backwards after a forced full is moot, since `force_full` discards the markers before the target loop runs.

If "at or beyond" is ever genuinely wanted, it needs a quantity that actually is ordered — the sequence number, which increases by one per commit — which means storing that alongside the snapshot id in the marker. I have not done that here; it is a schema change and this PR does not need it.

Two other changes came with it:

- Skips are now logged on **every** pass, not only an incomplete one. An all-skip pass is `is_complete()`, so it was the single outcome that printed nothing at all — which is exactly the shape a wrongly-skipped target takes.
- The test that asserted the old behaviour is **replaced**, not adjusted. It required a marker of `21` to count as caught up against a window ending at `20`. Every fixture in it was sequence-shaped (`10`, `19`, `20`, `21`), which is precisely why `>=` read as correct.

That last point is the part I would most want a reviewer to notice. My earlier description of this PR listed `>=` → `==` as a *killed mutation* — so the suite was actively defending the defect, and the mutation testing gave me false confidence. A mutation suite can only ask whether a test notices an edit; it cannot ask whether the fixtures resemble production values. When a domain has no ordering, ordered fixtures will ratify an ordering comparison. The new regression test therefore carries five real (marker, later-snapshot) pairs from the incident and asserts up front that each marker really is numerically larger than the snapshot it wrongly skipped, so the fixture cannot quietly stop testing the trap.

`fmt` and `clippy --all-features --all-targets -- -D warnings` clean; `fluree-db-api --lib` 1075 pass, 0 fail. Reverting `==` to `>=` fails both new tests.

fluree-db-api/src/graph_source/r2rml_materialize.rs
```diff
@@ -900,6 +1086,112 @@ fn watermark_subject(source_graph_source_id: &str, target_spec: &str, table_name
...
+/// snapshot and its own parent. `a >= to` therefore compares two unrelated random
+/// i64s and calls the answer "has this target already applied the window".
+///
+/// That comparison makes the marker a one-way ratchet. It is only rewritten when a
+/// window's `to` happens to compare greater, so it climbs the running maximum of a
+/// random sequence and then exceeds every later draw permanently — after roughly
+/// `ln(n)` windows a target stops being written to at all. Measured on a 17-source
+/// deployment: every marker had ratcheted to between 8.19e18 and 9.22e18 against an
+/// `i64::MAX` of 9.223e18, so all 17 were skipped on every poll, each skip was
+/// counted as an ok target, and the watermark advanced past data that was never
```
aaj3f: **Optional.** The PR description is now arguing against the code. The body's "Edges decided rather than left open" section still says "`>=`, not `==`: a marker ahead of this window still means the target has it. Equality would re-apply and rebuild the amplification" — but `fc9fdb8d3` reversed exactly that after the production ratchet incident, and the doc comment here now (correctly) forbids reintroducing an inequality. Since PR bodies feed the merge commit and release notes, we should probably refresh that bullet (and the "Testing" section's `>=` → `==` killed-mutation claim, which the head commit shows was defending the defect) before merge so the next reader doesn't take the stale design argument as current.

fluree-db-api/src/graph_source/r2rml_materialize.rs
```diff
@@ -646,15 +676,47 @@ impl Fluree {
...
+                    );
+                    std::collections::HashMap::new()
+                })
+        };
 
         for (target, (live, deletions)) in by_target {
+            // Already has this whole window. Skipping is what stops a job whose
+            // watermark cannot advance from rewriting its healthy targets on every
+            // poll — the amplification that filled a 100 GiB volume.
+            if target_is_caught_up(&applied, &target, &table_watermarks) {
```
aaj3f: **Question, not a suggestion.** The `==` skip only fires while the window re-presents the same `to`, i.e. while the source is quiescent. The module's own no-delta comment at `:578` says the upstream sink bumps snapshot ids 37–72×/hour whether or not any row concerns us — so on an actively-committing source with one stuck target, `to` moves every poll, equality never matches, and the healthy targets are rewritten again (the amplification returns at the source-commit rate rather than the poll rate). I don't think that makes this change wrong — it's strictly better than before, and your own doc comment already names the fix if "at or beyond" is ever needed (store the Iceberg sequence number beside the snapshot id, which *is* ordered). Two thoughts, either of which I'm happy to punt on: (1) it may be worth a sentence in the body scoping the win to stable-`to` windows; (2) interestingly, #1649's per-pass cap makes a backlogged job's `to` depend only on the watermark and the lineage — not the advancing head — which restores exactly the stable-`to` condition this skip needs, so the two PRs compose better than either alone. Might be worth cross-referencing them.

fluree-db-api/src/graph_source/r2rml_materialize.rs
```diff
@@ -3119,6 +3577,91 @@ mod engine_tests {
     /// A templated target fans out: one scan, N target ledgers, each its own commit
```
aaj3f: **Nit.** The new engine test was inserted between `a_templated_target_fans_out_per_row`'s doc comment and that test, so the old doc ("A templated target fans out: one scan, N target ledgers…") is now fused onto the top of the new test's doc block, and the fan-out test at `:3666` lost its doc entirely. Just needs the three old lines moved back above their test.

fluree-db-api/src/graph_source/r2rml_materialize.rs
```diff
@@ -3000,7 +3454,11 @@ mod engine_tests {
             _from: Option<i64>,
```
aaj3f: **Praise.** The `repeating()` opt-in on `FakeSource` (keeping drain-by-default so a fixture pulled twice still gets caught) and the choice of the target ledger's `t` as the assertion — with the doc comment explaining why `subjects_upserted` and the target data are both vacuous observables — is exactly the kind of test that survives contact with a mutation suite. Same for the production `(marker, later-snapshot)` pairs in `a_marker_numerically_above_a_later_snapshot_is_not_caught_up`, with the guard assertion that each fixture really is the trap shape; sequence-shaped fixtures ratifying an ordering comparison is precisely how the original `>=` survived its own mutation testing.

fluree-db-api/src/graph_source/r2rml_materialize.rs
```diff
@@ -900,6 +1086,112 @@ fn watermark_subject(source_graph_source_id: &str, target_spec: &str, table_name
...
+/// reintroduce an inequality over snapshot ids.
+fn target_is_caught_up(
+    applied: &std::collections::HashMap<(String, String), i64>,
+    target: &str,
+    table_watermarks: &[(String, Option<i64>, i64)],
+) -> bool {
+    !table_watermarks.is_empty()
+        && table_watermarks.iter().all(|(table, _from, to)| {
+            applied
+                .get(&(target.to_string(), table.clone()))
```
aaj3f, line 1238: **`fluree-db-api/src/graph_source/r2rml_materialize.rs:1190`** — *blocking (a body edit, not a code change).* The code is right and the description is now arguing against it, in two places. `9820a68ad` reverses the applied-marker comparison to `==` — with the production ratchet incident as the reason, and with `a_marker_numerically_above_a_later_snapshot_is_not_caught_up` pinning it on real (marker, later-snapshot) pairs — but the PR body's "Edges decided rather than left open" section still reads "`>=`, not `==`: a marker ahead of this window still means the target has it. Equality would re-apply and rebuild the amplification", and the body's mutation list still cites "`>=` → `==`" as a killed mutation, which is the suite your own commit describes as defending the defect and replaces. The body is what the next maintainer reads before "fixing" `==` back to `>=`, so those two passages should tell the story the head commit tells — the incident narrative in `9820a68ad` is the best prose in the PR and deserves to be in the body rather than only in the git log.

fluree-db-api/src/graph_source/r2rml_materialize.rs
```diff
@@ -3119,6 +3577,91 @@ mod engine_tests {
...
+        );
+        assert_eq!(
+            second.tally.deferred + second.tally.failed,
+            0,
+            "skipping is not deferral and not failure"
+        );
+    }
+
     #[tokio::test]
     async fn a_templated_target_fans_out_per_row() {
```
aaj3f, line 3714: **`fluree-db-api/src/graph_source/r2rml_materialize.rs:3666`** — *nit.* `a_templated_target_fans_out_per_row`'s doc comment ("A templated target fans out: one scan, N target ledgers, each its own commit domain…") is fused onto the top of the new skip test's doc block at `:3577-3579`, and the fan-out test itself is left with no doc at all. Just the three lines moved back above `:3666`. Minor and non-blocking — but if you agree it's right, I'd rather see it folded in now than lost in the backlog.

fluree-db-api/src/graph_source/r2rml_materialize.rs
```diff
@@ -678,15 +740,65 @@ impl Fluree {
...
             failed: failed.len(),
         };
+        // Report skips on EVERY pass, not only an incomplete one.
+        //
+        // An all-skip pass is `is_complete()`, so it would otherwise be the ONE outcome
+        // that prints nothing whatsoever — and that is precisely the shape a
+        // wrongly-skipped target takes. `all_skipped` is the specific alarm: nothing was
+        // written this pass, yet the watermark is about to advance. That is legitimate
+        // for a genuinely idle window and a data-loss signature otherwise, so it has to
+        // be visible either way rather than inferred by diffing a target ledger's `t`.
```
aaj3f, line 750: **`fluree-db-api/src/graph_source/r2rml_materialize.rs:750`** — *praise.* Reporting skips on **every** pass with the `all_skipped` flag — rather than only on an incomplete one — closes the exact observability hole the incident exposed ("each skip was counted as an ok target and logged nothing"). The one outcome that used to log nothing at all was precisely the shape a wrongly-skipped target takes, and now it can't be silent.

fluree-db-api/src/graph_source/r2rml_materialize.rs
```diff
@@ -2684,6 +2976,154 @@ mod tests {
...
+    /// These are real values from a deployment where that happened — all 17 markers
+    /// had reached 8.19e18–9.22e18 against an `i64::MAX` of 9.223e18, every target was
+    /// skipped on every poll, and the watermark advanced past data that was never
+    /// applied. Each `to` below is a genuinely LATER commit than the marker beside it,
+    /// despite being numerically smaller.
+    ///
+    /// Small sequence-shaped fixtures cannot catch this — an ordered fixture will
+    /// ratify an ordering comparison — which is why these are the production numbers.
+    #[test]
+    fn a_marker_numerically_above_a_later_snapshot_is_not_caught_up() {
```
aaj3f, line 3084: **`fluree-db-api/src/graph_source/r2rml_materialize.rs:3036`** — *praise.* Carrying five real (marker, later-snapshot) pairs from the incident, with the up-front assertion that each marker is numerically larger than the snapshot it wrongly skipped, and the stated reason — "an ordered fixture will ratify an ordering comparison" — is the fixture-representativeness lesson encoded where a mutation suite can't reach it. And the e2e test's choice of the target ledger's `t` as the only honest observable (with the doc block explaining why `subjects_upserted` and the target data are both vacuous) is exactly the kind of test that survives contact with a maintainer.

aaj3f (approved):
I started reviewing this before your more recent updates and the PR has materially approved, @christophediprima. The `==` reversal in `9820a68ad` is the right call for the right reason (two Iceberg snapshot ids simply have no ordering, so `>=` was comparing unrelated 64-bit numbers and ratcheting), and the commit body's incident narrative (17 markers at 8.19e18–9.22e18 against `i64::MAX`, every source silently skipped, ~80% of entities discarded with no warning in three days of logs) is a very good account of a silent-data-loss mechanism. 

I verified the guard rather than trusting it: at head, `a_re_presented_window_does_not_rewrite_targets_that_already_applied_it` passes, and gutting `target_is_caught_up` fails it with exactly the assertion your body quotes (`left: 3 / right: 2` on the target's `t`). The five-real-pairs regression test and the every-pass skip reporting both landed as described.

The one thing I'd ask before this merges is a more about PR body than code itself: the PR description still argues for `>=` in "Edges decided" and still lists `>=` --> `==` as a killed mutation. Both now contradicted by your own head commit. That prose is what the next person reads before deciding the comparison is safe to "fix", so it should tell the incident's story instead. Details inline.

- **Patterns/abstractions:** ✔ extends the module's existing marker/watermark vocabulary; the `repeating()` opt-in on `FakeSource` preserves the drain-by-default trap-catching behavior rather than weakening it.
- **Performance (speed first, memory second):** ✔ comparison-operator change plus one `info!` per pass on the skip path; no hot-path impact.
- **Testing:** ✔ predicate unit tests + the incident-pair regression test + a mutation-verified engine test on the one observable that can't lie. (Note for CI-readers: the module is `iceberg`-feature-gated, so a plain `--lib` run filters these tests out.)

Verified locally at branch HEAD (`9820a68ad`): `cargo test -p fluree-db-api --lib --features iceberg a_re_presented_window` → 1 passed; skip-predicate mutation → red with the body's exact assertion; restored → green; `git merge-tree` against #1663's head → conflict-free, and neither branch contains the other.

Approving so you can merge when ready — just get the body brought up to date with `9820a68ad` first, since right now it's the one thing still defending the bug you fixed.

bplatz:
Pushed four commits to this branch (`maintainerCanModify`), @christophediprima — two housekeeping, two from review. Happy to drop any of them if you disagree, particularly the third.

**`origin/main` merged in** (36 commits, conflict-free). Materialize tests, fmt and clippy re-verified after.

**The fan-out test's doc block moved back to the fan-out test.** Three lines describing `a_templated_target_fans_out_per_row` had been fused onto the head of the skip test's doc, leaving the test they describe undocumented. No code change.

**The applied-marker write is now non-fatal.** This is the one that is a judgement call rather than a tidy-up, so pushing back is reasonable.

The write propagated its error, and it sits *between* the target loop and the failed/deferred returns — so a state-ledger failure surfaced as that error instead of `MaterializePartial { tally, detail }`, discarding the per-target tally and the identity of the target that actually failed. What decided it for me is which condition makes that write fail: at ENOSPC — the incident these markers exist to prevent — every state-ledger write fails, so an operator would get a bare disk error exactly where the useful signal is `"target X failed: No space left on device"`. It also read as asymmetric with the marker *read* 90 lines above, which you had already made deliberately non-fatal on what looks like identical reasoning ("an unreadable marker set costs the skip, not the pass"). An unwritten marker costs one redundant re-apply next poll, and re-application is idempotent. The write moved into a helper so its two `?`s stay inside it.

**A note on what the marker set costs to hold**, in `materialize_applied_markers`. It is unbounded in a way the watermark deliberately is not: the watermark is keyed on the target *spec*, so a templated job keeps one row per source table however wide it fans, while a marker is keyed on the *resolved* ledger — so the set grows with the fan-out, every poll reads all of it, and nothing evicts a marker when a target ledger is retired. Immaterial at the 23 targets this was built for, and not something I think belongs in this PR; recorded where the next person will look.

I also updated the PR description, which was the blocking item in @aaj3f's review: "Edges decided" still argued for `>=` over `==`, and the mutation list still credited `>=` → `==` as killed. Both now tell the ratchet incident instead, and the mutation is recorded in the correct direction and as killed by two tests rather than one — I re-ran it to check that count rather than take it from the review.

Two things I verified rather than trusted while reading: mutating `==` back to `>=` fails exactly two tests (including `a_marker_numerically_above_a_later_snapshot_is_not_caught_up`), and `all` → `any` on the every-table guard fails exactly one. The suite bites where it claims to.

PR #1649, opened 2026-08-13
feat(materialize): advance a backlogged window by a bounded prefix

christophediprima:
## The failure this prevents

The materialize watermark only advances when a whole pass succeeds. A pass that cannot finish its window writes no watermark, so the next pass re-reads a window one poll wider, and so on without bound.

That is not merely slow — it is a one-way door:

```
a pass defers  ->  the window is discarded, no watermark is written
               ->  the pinned snapshot ages out of the source's retention
               ->  the next poll cannot resolve `from_snapshot_id`
               ->  it falls back to a FULL TABLE READ
               ->  a full read is far more expensive, so it defers
               ->  repeat, permanently
```

Each turn makes the next more likely, and there is no exit: the entry condition for the incremental path — a resolvable `from` snapshot — is exactly what gets lost.

**Measured on a 17-table deployment.** One affected table became **13 of 17 in about four hours**. 126 consecutive poll results, every one "partial window, remainder deferred", zero completions. All 13 tables showed an identical `from_snapshot_id` across nine consecutive polls: no watermark moved at all. Resident memory rose 19.5 → 24.7 GiB from the full-table reads alone.

The tables were not individually unhealthy. They were dragged there by a shared property of the watermark, and once there they could not get back.

## The change

`TableMetadata::capped_scan_end(from_id, max_snapshots)` returns where an unpinned incremental consumer should end this pass: the head, or an earlier snapshot when the backlog exceeds the cap. An unpinned materialize scan reads to there instead of to head.

`FLUREE_MATERIALIZE_MAX_SNAPSHOTS_PER_PASS`, default **64**; `0` disables it and restores the previous behaviour exactly.

**64 is deliberately generous.** A source committing 37–72 snapshots an hour — ours does — stays under it at any sane poll interval, so a healthy job never reaches the cap and nothing about its behaviour changes. Only a backlogged job reaches it, and it then drains in bounded steps instead of never.

## Why it is safe

Three properties, each asserted by a test rather than argued:

- **Capping only ever moves `to` earlier**, never later, so no snapshot is skipped. The next pass resumes from the watermark this one wrote.
- **An initial full read (`from = None`) is never capped.** A partial "full" read is worse than an unbounded one — it would look complete while missing rows.
- **Any window error falls through to the head unchanged** — expired ancestor, non-ancestor, rollback. That is the existing full-read fallback's case and it keeps it.

The prefix is the **oldest** snapshots in the window, not the newest. That direction is the one thing here that is silent when wrong: taking the newest would skip everything between the watermark and the chosen end, which is data loss rather than a performance bug. `window_end_capped_takes_the_oldest_prefix` exists to fail if anyone inverts it, and it does.

## On the second commit

The decision originally lived inline in `scan_for_materialize_stream`. Testing it there needs a storage backend and an Iceberg fixture, which in this repo means `aws-testcontainers` — a high price for asserting a branch, and the usual result is that the branch stays unasserted.

The second commit moves it to `TableMetadata::capped_scan_end`, beside the metadata fixtures, so the scan path calls it and has no logic of its own left to get wrong. Every decline path is then covered by an ordinary unit test: no snapshots, cap disabled, initial full read, backlog already fits, window unwalkable.

Three mutations, each killed by exactly one test: taking the newest prefix instead of the oldest, capping an initial full read, and dropping the fall-back-to-head on a window error.

## Scope, stated plainly

**This prevents a table from entering that state. It does not rescue one already in it.** A table whose watermark has already expired takes the full-read path, and this deliberately does not cap a full read — so it stays on that path until one read completes. The fix for the already-stuck case is sub-snapshot progress (a resumable file cursor), which is a larger change to the scan API and the persisted watermark, and belongs in its own PR.

We are happy to bring that too, and would rather agree the shape first — particularly whether a per-`(source, target, table)` cursor is acceptable state growth, and whether it belongs in the state ledger or the target ledger. This PR stands on its own either way: it is the difference between a degradation that spreads and one that does not.


fluree-db-api/src/graph_source/r2rml.rs
```diff
@@ -244,6 +244,38 @@ impl ScanChoice {
...
+/// Observed on a 17-table deployment: one affected table became 13 of 17 in
+/// about four hours, with every table then reading its whole self on every poll.
+///
+/// A cap bounds the work per pass, so the watermark advances every time and
+/// therefore stays inside retention. 64 is deliberately generous — a source
+/// committing 37-72 snapshots an hour stays under it at any sane poll interval,
+/// so a healthy job never notices the cap, while a backlogged one drains in
+/// bounded steps instead of never.
+fn materialize_max_snapshots_per_pass() -> usize {
+    static CACHED: std::sync::OnceLock<usize> = std::sync::OnceLock::new();
```
aaj3f: **Optional.** An unparseable `FLUREE_MATERIALIZE_MAX_SNAPSHOTS_PER_PASS` value silently becomes the default 64 — `.ok()` on the parse swallows, say, `64k` or `-1`, and the operator who set it gets no signal their knob didn't take. Not a big deal since the default is safe, but a one-line `warn!` on the parse-failure arm would be cheap insurance; the `0`-disables contract makes a silently-ignored value slightly more surprising than usual. Happy to punt on this.

fluree-db-api/src/graph_source/r2rml.rs
```diff
@@ -1266,7 +1298,39 @@ impl<'a> FlureeR2rmlProvider<'a> {
                         })?,
                 )
             }
-            None => metadata.current_snapshot(),
+            // Unpinned: read to the source's head, but advance through a long
```
aaj3f: **Question, not a suggestion.** Is it worth cross-referencing #1647 in the body or a comment here? The capped `to` depends only on the watermark and the snapshot lineage — not the advancing head — so a backlogged job re-presents the *same* `to` on every poll even while the source keeps committing. That happens to be exactly the stable-`to` condition #1647's `==` applied-marker skip needs to fire, so the two PRs compose better together than either reads alone (without this cap, an actively-committing source moves `to` every poll and #1647's equality never matches). Might save the next reader from evaluating them independently.

fluree-db-iceberg/src/metadata/table.rs
```diff
@@ -170,6 +170,75 @@ impl TableMetadata {
...
+        &self,
+        from_id: Option<i64>,
+        max_snapshots: usize,
+    ) -> Option<&super::Snapshot> {
+        let head = self.current_snapshot()?;
+        self.window_end_capped(from_id, head.snapshot_id, max_snapshots)
+            .ok()
+            .filter(|id| *id != head.snapshot_id)
+            .and_then(|id| self.snapshot(id))
+            .or(Some(head))
```
aaj3f: **Praise.** The `.ok().filter(≠head).and_then(snapshot).or(Some(head))` shape is the repo's correctness-preserving-fallback pattern applied cleanly: every way this can fail — unwalkable window, missing snapshot, no capping needed — lands on "read to head", which is byte-identical to the pre-PR behavior. And moving the decision out of `scan_for_materialize_stream` so it's testable without an `aws-testcontainers` backend is exactly the right call; the five decline-path tests exist *because* the logic moved to where the fixtures are.

fluree-db-iceberg/src/metadata/table.rs
```diff
@@ -551,6 +620,112 @@ mod tests {
         assert!(m.snapshot_window(Some(1), 42).is_err());
     }
 
+    /// A backlog longer than the cap is advanced in bounded steps, taking the
+    /// OLDEST snapshots first. Taking the newest instead would skip everything
+    /// between `from` and the chosen end — silent data loss, and the reason the
```
aaj3f: **Praise.** Asserting the prefix *direction* with its data-loss rationale spelled out ("taking the newest would skip everything between the watermark and the chosen end") is the one assertion here that's silent when wrong, and it's the first test in the block. Good instinct.

bplatz:
Merged `main` (twice — #1663 and #1581 landed in the meantime) and finished this up. What changed from the original cut:

- **#1663 covered the full-read half**, so the permanent trap is gone. What remained is the incremental half this PR targets: a window over the novelty ceiling still deferred on every poll until the watermark expired into a whole-table full read. That path is now bounded too.
- **Rows, not a snapshot count.** 64 snapshots is disconnected from the constraint that causes the deferral (64 × 10k rows still cannot commit; 200 × 10 rows never needed splitting). `TableMetadata::cap_window_by_rows` sizes the window from each snapshot's `added-records` summary against the same ceiling-derived budget the full read uses, so both paths share one mechanism and one setting. It still decides from metadata with no manifest I/O, stays unit-testable without a backend, keeps the snapshot the budget runs out in whole, and reports typed declines so the log can say why a window went unbounded.
- The cap applies only to an unpinned window `ScanChoice` found incremental-safe; a window with an overwrite full-reads the head anyway, and that read subsumes every prefix.
- `FLUREE_MATERIALIZE_MAX_ROWS_PER_FULL_PASS` → `FLUREE_MATERIALIZE_MAX_ROWS_PER_PASS`, since one budget now covers both paths (unreleased, so no alias).
- New fixture `silver.people_window` and an end-to-end test that advances the source between polls: passes of 1200 / 600 / 600 rows, each incremental, watermark at the capped snapshot rather than the head, every row present once. With the cap mutated to never engage, pass 2 reads all 2,400 rows and defers on novelty.

Docs updated on both pages. `cargo clippy --all-features --all-targets -- -D warnings` clean on both crates; `fluree-db-iceberg` 264, r2rml 100, `it_iceberg_local_fs` 4 pass.

PR #1655, opened 2026-08-16
fix(indexer): stamp the target ledger_id on incrementally built index roots

bplatz:
## Problem

A branch cut from an indexed ledger becomes unwritable, for any process holding it in cache, as soon as the indexer publishes the branch's first own index.

`create_branch` copies the parent's index root by CID, so the branch's base root carries the parent's `ledger_id`. `IncrementalRootBuilder::from_old_root` cloned that root wholesale and nothing in `incremental_index` ever wrote `root.ledger_id`, so the branch's incremental root was published as e.g. `business-model:main` with `index_t = 8`. The load path re-stamps the cached snapshot from the nameservice record, so a cached `LedgerState` for the branch then rejected the new root in `apply_loaded_db`:

```
apply_loaded_db failed: Index ledger_id 'business-model:main' does not match expected 'business-model:pii-hardening'
```

Every cache refresh via `LedgerManager::notify` failed the same way, which a transactor treats as transient and retries — wedging the branch's write queue while reads keep working.

The full-rebuild path and the pack-restore path already stamp the target ledger; incremental was the outlier.

## Fix

`from_old_root(root, ledger_id)` now takes the target ledger id and re-stamps the root, so the stamp can't be forgotten by a caller.

## Tests

- `notify_branch_refresh_after_incremental_index_on_branch` — indexes `main`, branches `dev`, caches `dev` in a peer `LedgerManager`, writes on `dev`, indexes `dev`, then asserts the published root names `dev` and that `notify` refreshes the cached handle. Fails before this change on both counts (root says `:main`, then the exact `apply_loaded_db` error above). Existing branch tests re-open with `fluree.ledger()`, which goes through the re-stamping load path and so never hit this.
- `from_old_root_restamps_foreign_ledger_id` unit test on the builder.

Not included (left for discussion): re-stamping in `LedgerHandle::apply_index_v2` as defence in depth. It would immediately unwedge branches whose bad root is already published, but it makes `LedgerIdMismatch` unreachable from its only live caller — i.e. it removes the check that caught this. Already-published bad roots self-heal after one cold reload (load path re-stamps → next commit → re-index with this fix).


fluree-db-indexer/src/run_index/build/incremental_root.rs
```diff
@@ -24,13 +24,22 @@ pub struct IncrementalRootBuilder {
 }
 
 impl IncrementalRootBuilder {
-    /// Create a builder from an existing root (cloned).
-    pub fn from_old_root(root: IndexRoot) -> Self {
+    /// Create a builder from an existing root (cloned), re-stamped to `ledger_id`.
```
aaj3f, line 27: **Praise.** Making the stamp part of `from_old_root`'s signature (rather than a post-hoc `set_ledger_id` a caller could forget) is the right shape — the next caller of this builder inherits the fix for free, which matters given pack import/clone is the second producer of foreign-id base roots.

aaj3f, line 27: **Optional.** On the defense-in-depth question you left open (re-stamping in `apply_index_v2`): I'd agree with leaving it out — neutering `LedgerIdMismatch`'s only live caller trades away the tripwire that caught this. The one thing we may want somewhere findable (ops notes or the PR body is probably enough) is the remediation for a fleet that already published a bad root before this ships: a cold reload of the affected process self-heals it, but an operator staring at a wedged write queue with healthy reads won't guess that. Happy to punt on this — the body already documents the self-heal path.

aaj3f (approved):
Good find and a clean fix, @bplatz 

Just be sure to merge #1656 first and retarget this, as the CI checks won't have run until then (or be prepared for 1656 to possibly go red for issues related to this PR if you do merge this first)

- **Patterns/abstractions:** ✔ aligns incremental with the load path and pack restore's existing identity-authority treatment; no new construct.
- **Performance (speed first, memory second):** ✔ neutral — one compare per index build, nothing per-flake, query path untouched.
- **Testing:** ✔ end-to-end branch repro (peer-cached manager, notify refresh, post-index write) plus a unit test on the builder; both verified live and passing locally, and the integration file is wired in `grp_index`.
- **Conventions:** ✔ multi-line commit body with the mechanism; fmt clean at the stack tip; the "not included" discussion in the body is model PR hygiene.

Verified locally at branch HEAD: `grp_index notify_branch_refresh_after_incremental_index_on_branch` 1/1, `fluree-db-indexer --lib from_old_root` 1/1, `cargo fmt --all -- --check` clean.

PR #1656, opened 2026-08-16
fix: harden the upsert absent-subject skip against an unattached index

bplatz:
Follow-up to #1645, from a code review of that branch. Two commits: the correctness fix, and an unrelated one-line clippy repair that CI on `main` currently needs.

## 1. `fix(transact)`: keep absence undecidable when the store failed to load

`generate_upsert_deletions` treated a missing range provider as proof that no base index exists, so every subject was decided by novelty alone.

That holds at genesis, but `LedgerManager` treats a binary-store load failure as **non-fatal** (`ledger_manager.rs:1254`, "Failed to load binary store, continuing without"), which leaves an indexed ledger (`t > 0`) with no provider attached. In that state the pre-check reported every subject absent, skipped it, and committed **zero retractions** — old values silently survived alongside the new ones. Before the skip existed the same state failed loudly in `range_with_overlay`.

- Require `t == 0` alongside the missing provider, mirroring `fluree_db_core::range`, which treats a missing provider as an empty index only at genesis and errors otherwise (`binary-only db has no range_provider attached`). A `t > 0` ledger with no provider now falls through to the per-predicate query, which surfaces the load failure instead of silently dropping retractions.
- Drop the `store.sid_to_iri` fallback. The store's namespace table is always a subset of the snapshot's — the index root is a materialized cache at `index_t`, and `ns_helpers::sync_store_and_snapshot_ns` reconciles its codes back into the snapshot on load (`fluree_ext.rs` documents the same invariant on the read-only view path) — so it can never decode a code the snapshot could not. The closure's no-store arm collapses to `false`, which is all `!base_index_absent` could evaluate to under the caller's gate.

### Tests

The existing per-subject-namespace regression test never reached the changed branch: its subject was already indexed, so it hit `find_subject_id_by_parts` on the first probe. It now also covers a subject committed **after** the last index build, where the snapshot decodes the namespace code but the store conclusively misses — the one direction where removing the fail-open could lose a retraction, and the only thing standing in the way is the novelty backstop. Verified by neutralizing that backstop, which fails on exactly the new assertion (`left: ["C1","C2"], right: ["C2"]`) and nothing else.

The span-capture counter assertion moves to its own `it_upsert_skip_fires` binary, per the standalone rule in `docs/contributing/tests.md`: tracing's callsite-interest cache is process-global and the capture is thread-local, so a sibling test in the same process can leave the event undelivered.

### Docs

`namespace-allocation.md` no longer suggests "setting `HostPlusN(n)` on the ledger" as a remedy — there is no user-facing knob, the bulk-import preflight is the only thing that selects it, and the mode is immutable after genesis.

## 2. `chore(policy)`: name the class-cache type

`Arc<RwLock<HashMap<(GraphId, Sid), Vec<Sid>>>>` trips `clippy::type_complexity`, which fails CI's `clippy --all --all-features --all-targets -D warnings`. This arrived with 18ad842c4 (#1580) and `main` is red on it right now. Adds a `ClassCache` alias; no behaviour change. Happy to split this out if it should go back to that PR's author instead.

## Not addressed here

Two findings from the same review are left for a separate perf PR — they are the remaining 30-70s, not correctness:

- `BinaryScanOperator` still widens to a PSOT/SPOT partition walk for a bound subject whose IRI decodes but conclusively has no base rows, while the bound-object `Ref` path already treats `Ok(None)` as a conclusive miss. Only upsert benefits from the caller-side pre-check; SPARQL/JSON-LD/`DELETE WHERE` still pay it.
- `open_overlay_only_fallback` still walks all of novelty per probe instead of using the bounded `Segment::range` seek.

## Validation

- `cargo fmt --all -- --check` clean
- `cargo clippy --all --all-features --all-targets -- -D warnings` clean (was failing on `main` before commit 2)
- `cargo nextest run --workspace --all-features --no-fail-fast`: 11115/11115 passed


christophediprima:
Flagging the second commit here, `fa95d1d64` ("chore(policy): name the class-cache type…"), because it is currently the only thing unblocking clippy on `main`.

`main` at `3ce13529d` fails `cargo clippy --all --all-features --all-targets -- -D warnings`:

```
error: very complex type used. Consider factoring parts into `type` definitions
  --> fluree-db-policy/src/evaluate.rs:75:18
```

Since clippy runs with `-D warnings`, that fails the clippy job on every open PR whose merge includes `main` — not only on changes that touch this file. I ran into it on an unrelated materialize PR where clippy was the one red job and nothing in the diff went near policy.

Would you be open to landing that commit on its own, ahead of the rest of this PR? It is a three-line type alias with no behaviour change, so it could go in immediately and unblock the queue without waiting on review of the upsert hardening here (+246/−134 across 8 files).

I can open that PR if it is useful, but it is your commit and I would rather not duplicate it — happy for you to take it whichever way is least work. Raising it mainly because the type became complex in #1580, which was mine: I had this same alias on my own integration branch and should have carried it into that PR instead of leaving you to write it.

fluree-db-api/tests/it_upsert_skip_fires.rs
```diff
@@ -0,0 +1,125 @@
...
+};
+use fluree_db_api::{FlureeBuilder, IndexConfig};
+use fluree_db_transact::{CommitOpts, TxnOpts};
+use serde_json::json;
+
+fn ctx() -> serde_json::Value {
+    json!({"ex": "http://example.org/ns/"})
+}
+
+#[tokio::test(flavor = "current_thread")]
```
aaj3f: **Praise.** Moving the span-capture assertion to a standalone binary (with the `docs/contributing/tests.md` update naming it) is exactly the right call — the callsite-interest-cache flake class is nasty to debug after the fact, and the module doc here explains the two process-global mechanisms clearly enough that the next person won't re-bundle it.

fluree-db-policy/src/evaluate.rs
```diff
@@ -56,6 +56,9 @@ impl<'a> FlakeEvalParams<'a> {
     }
 }
 
+/// Cache of `(graph, subject) -> classes` shared across concurrent evaluations.
+type ClassCache = Arc<RwLock<std::collections::HashMap<(GraphId, Sid), Vec<Sid>>>>;
```
aaj3f: **Optional.** The `ClassCache` alias is fine and I'm glad `main` clippy is green again. Since this arrived with 18ad842c4 (#1580), we may want a quick heads-up to that PR's thread so the pattern doesn't recur — but bundling the repair here rather than blocking on process was the right pragmatic call, and you flagged it transparently in the body. Nothing to change.

aaj3f (approved):
LGTM, @bplatz. The "name the class-cache type" work, though, did get covered by #1658 -- may worth a fast-foward or merge of main into this branch to ensure it's not now a textual / logic conflict.

- **Patterns/abstractions:** ✔ extends the existing `range`-module absence contract rather than special-casing; the standalone-test convention follows `it_cyclic_bgp_probe` precedent.
- **Performance (speed first, memory second):** ✔ neutral — staging-setup gate only, no hot-path or per-flake change; the #1645 skip's win is preserved and pinned by `it_upsert_skip_fires`.
- **Testing:** ✔ both binaries verified live and passing locally (`it_upsert_skip_fires` 1/1, `grp_transact it_transact_upsert_indexed` 3/3); the new file is correctly wired via `[[test]]` (autotests=false trap avoided).
- **Conventions:** ✔ thorough multi-line commit bodies; clippy/fmt clean (and commit 2 un-reds `main`'s workspace clippy); docs updated where behavior claims changed (`namespace-allocation.md`, `tests.md`).

Verified locally at branch HEAD: `it_upsert_skip_fires` + `it_transact_upsert_indexed` green, `cargo fmt --all -- --check` clean, `clippy -p fluree-db-policy --all-targets --no-deps` clean of `type_complexity`.

PR #1657, opened 2026-08-16
perf: stop widening scans for a bound subject absent from the base index

bplatz:
Two scan-path costs found while reviewing #1645. Both are performance, not correctness — and both were verified against the code and measured before being fixed.

**Stacked on #1655** (`fix/branch-index-root-ledger-id`), which is the base branch. Review that one first.

## 1. An absent bound subject widened into a predicate-partition walk

`BinaryScanOperator::open` keeps the decoded IRI of a bound subject that fails to translate to a persisted `s_id`, switches the scan to PSOT, and re-checks every row of the predicate's partition by resolving its subject IRI. Probing a subject that does not exist cost **O(partition size)** with a dictionary lookup per row; the same probe for a subject that *does* exist is a single dictionary hit.

The root cause is that the IRI was never actually looked up. `build_filter_from_snapshot_sids` resolves `Ref::Sid` through `active_snapshot`, but a pattern SID is encoded against `original_snapshot` — where those namespace tables differ, the filter either used the wrong IRI or never ran. It is also strictly weaker than `subject_ref_to_s_id`, which tries `find_subject_id_by_parts` first and falls back through `store.sid_to_iri`. The in-file comment claiming the bound-object `Ref` arm "mirrors the subject path (`build_filter_from_snapshot_sids`)" is inaccurate — the object arm is stronger, and already treats `Ok(None)` as a conclusive base miss. That asymmetry is the bug.

Now the store's subject dictionary is probed with the IRI the pattern actually means:

| Outcome | Action |
|---|---|
| `Ok(Some)` | Narrow to `s_id`, drop the per-row IRI comparison |
| `Ok(None)` | Conclusive base miss → overlay-only |
| `Err(_)` | Absence undecidable → keep the widened scan |

**This is not an upsert-only path.** `join.rs` and `optional.rs` rebind a correlated subject to `Ref::Iri` once per driving row, so every non-batched join or OPTIONAL probe for a missing subject paid it. The `stage.rs` pre-check from #1645 only short-circuits the upsert caller.

Absent-subject point query, one predicate:

| Rows on predicate | Before | After |
|---|---|---|
| 5,000 | 583 µs | ~11 µs |
| 10,000 | 1.14 ms | ~11 µs |
| 20,000 | 2.20 ms | ~11 µs |
| 40,000 | 4.55 ms | ~11 µs |

Linear before (present-subject probe is a flat ~14 µs throughout); flat after.

## 2. The overlay-only fallback walked all of novelty per probe

`open_overlay_only_fallback` passed `first: None, rhs: None, leftmost: true`, so `Segment::range` returned the whole id slice and `may_overlap` pruned nothing. The `partition_point` seek machinery already existed and was simply unused — `fast_path_common::collect_resolved_overlay_ops` bounds its walk the same way.

Bounds are now derived from whichever bound terms lead the active index order. They are an **optimization only** — the per-flake equality checks stay the correctness backstop — so each pins only components that lead the sort order *and* are fixed by an equality match, letting every trailing component span its full range.

Absent-subject probe against unindexed novelty: 24/29/65/134 µs at 5k/10k/20k/40k flakes → **~9–10 µs, flat**.

The two are coupled: fixing #1 routes *more* probes into this fallback, so #2's per-probe cost matters more than it did. That is why they ship together.

## Correctness notes for review

- **`Ok(None)` as conclusive.** `find_subject_id` answers from the canonical-split namespace plus the EMPTY/OVERFLOW full-IRI fallback. Treating that as conclusive is already load-bearing in the bound-object `Ref` arm and in the #1645 upsert skip, so this is not a new assumption — but it does widen its blast radius from writes to all reads. Flagging it as the one judgment call here.
- **Bound tightness.** `PartialEq` for `FlakeValue` mirrors `Ord` (both route numerics through `numeric_cmp`), so a flake the callback would match cannot sort outside the bracket. `first` is left-exclusive, hence `t: i64::MIN`; `leftmost` must go false whenever it is supplied.

## Testing

`fluree-db-api/tests/it_absent_subject_scan_narrowing.rs`, 5 tests. The cost assertions pin **flatness in fuel** across partition sizes rather than a wall-clock bound, so they are deterministic in CI.

Every test was mutation-checked rather than assumed to have teeth:

- Reverting the narrowing fails both flatness tests (6.01 vs 41.01 fuel).
- The novelty-only test initially asserted only `fuel > 0` and passed under mutation — it was rewritten to assert the returned rows, and now catches a `Ok(None)` arm that reports "no rows anywhere".
- The overlay tests were confirmed to actually reach the fallback with bounds active in all four index orders (Spot/Psot/Post/Opst), and fail when a bound is made too tight on either the subject/predicate or the object axis.

CI parity green: `fmt --check`, `clippy --all --all-features --all-targets -D warnings`, and `nextest --workspace --all-features` at **11122 passed**.

Docs: `docs/design/query-execution.md` gains a section on bound-term resolution, the conclusive-miss vs undecidable split, and the bounds-are-not-correctness contract.


fluree-db-query/src/binary_scan.rs
```diff
@@ -1966,6 +2056,43 @@ impl Operator for BinaryScanOperator {
...
+        //   applies, and the same one `generate_upsert_deletions` relies on.
+        // - `Err(_)` — the dictionary could not answer; absence stays undecidable
+        //   and the widened scan below remains the correct (if slow) answer.
+        //
+        // Skipping this made an absent bound subject cost a full predicate-partition
+        // walk with a dictionary lookup per row — O(partition), not O(1). It is not
+        // an upsert-only path: `join.rs` and `optional.rs` rebind a correlated
+        // subject to `Ref::Iri` per driving row, so every non-batched probe for a
+        // missing subject paid it.
+        if let Some(target_iri) = self.unresolved_bound_subject_iri.clone() {
```
aaj3f, line 2085: **Praise.** The three-way split (`Ok(Some)` narrow / `Ok(None)` overlay-only / `Err` keep the widened scan) honors the repo's degrade-to-correct fallback contract exactly — the `Err` arm keeping the slow-but-correct path is the detail that a hastier version of this fix would have gotten wrong.

fluree-db-query/src/binary_scan.rs
```diff
@@ -1545,6 +1545,79 @@ impl BinaryScanOperator {
...
+            Some(s) => (s.clone(), s.clone()),
+            None => (Sid::min(), Sid::max()),
+        };
+        let p_bounds = || match p_sid.as_ref() {
+            Some(p) => (p.clone(), p.clone()),
+            None => (Sid::min(), Sid::max()),
+        };
+
+        // Only worth bounding when the order's leading component is pinned;
+        // otherwise the range spans everything anyway.
```
aaj3f, line 1596: **Optional, more a question than a suggestion.** `overlay_walk_bounds` returns `None` whenever the index order's leading component is unpinned — so e.g. a Psot walk with a bound subject and object but unbound predicate still walks all of novelty. I *think* that's the right scoping (a range spanning the whole leading component prunes nothing at the segment level anyway, so bounding buys little), but if `may_overlap` prunes per-segment on the full key rather than just the leading component, there might be residual pruning left on the table for the multi-component-bound shapes. Not needed for this PR's target workload — happy to leave it as a note for if a profile ever surfaces it.

fluree-db-api/tests/it_absent_subject_scan_narrowing.rs
```diff
@@ -0,0 +1,355 @@
...
+/// Fuel only — for the cost-shape assertions.
+async fn fuel_for_probe(n: usize, probe_iri: &str) -> f64 {
+    probe(n, probe_iri, None).await.0
+}
+
+/// The cost of probing a missing subject must not grow with the size of the
+/// predicate's partition. Before the narrowing fix this scaled linearly: the
+/// scan walked every `ex:name` row and resolved its subject IRI.
+#[tokio::test(flavor = "current_thread")]
+async fn absent_subject_probe_cost_is_flat_in_partition_size() {
```
aaj3f, line 133: **Praise.** Asserting flatness in fuel rather than a wall-clock bound is the right determinism trade for CI, and `assert_eq!(small, large)` (exact equality, not a tolerance) is a satisfyingly strict pin — a single extra index touch per partition row fails it loudly.

aaj3f (approved):
This looks good to me, @bplatz — both fixes land at the right depth (the subject arm now meets the standard the bound-object arm already set, and the overlay walk reuses the existing `Segment::range` seek machinery instead of growing a parallel mechanism), the PR body names its own judgment call (`Ok(None)` conclusiveness widening from writes to reads) instead of burying it, and the mutation-checking discipline on the tests is exactly what fuel-shaped assertions need to be trustworthy. 

One process point, same as #1655: this PR's base is another PR's branch, so no CI runs here at all — the 11122-test nextest result in the body is local. Just be sure the stack lands bottom-up with a retarget to `main` so the workspace gates get their pass before this merges.

- **Patterns/abstractions:** ✔ aligns the subject arm with the bound-object arm's existing conclusive-miss standard; reuses `Segment::range` seeks per the `predicate_walk_bounds` precedent; below the shared IR, so SPARQL/JSON-LD behave identically.
- **Performance (speed first, memory second):** ✔ strongly positive and measured (linear→flat on two probe classes; correlated joins/OPTIONALs compound the win); present-subject paths untouched; fuel-based regression tests are deterministic guards.
- **Testing:** ✔ 5/5 passing locally at HEAD, wired via `[[test]]`, mutation-checked by the author *and* independently by this review; the novelty-only-subject test pins the one direction where conclusiveness could swallow rows.
- **Conventions:** ✔ thorough multi-line commit bodies, `docs/design/query-execution.md` gains the contract documentation, fmt/clippy clean at HEAD.

Verified locally at branch HEAD: `it_absent_subject_scan_narrowing` 5/5; my `p_max = Sid::min()` mutation → exactly the 2 bounded-walk tests fail (then reverted); `cargo fmt --all -- --check` clean; `clippy -p fluree-db-query --all-targets --no-deps` clean.

PR #1658, opened 2026-08-17
fix(policy): alias the class-cache type (clippy type_complexity, red on main)

aaj3f:
Unbreak-main: #1580's graph-keyed class cache — the fix itself is right — widened `PolicyContext.class_cache`'s type past clippy's `type_complexity` threshold, and with `-D warnings` that turned main's clippy job red (first failing run: 31954057339, at 3ce13529d). Every open PR inherits the red through the merge-ref checkout, which is how we noticed on #1654.

One `ClassCache` type alias — clippy's own suggested shape — no behavior change. `cargo clippy -p fluree-db-policy --all-targets -- -D warnings` clean locally, policy suite 39/39, fmt clean.

### DIFF
--- a/fluree-db-policy/src/evaluate.rs
+++ b/fluree-db-policy/src/evaluate.rs
@@ -56,6 +56,10 @@ impl<'a> FlakeEvalParams<'a> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


bplatz (approved):
LGTM!

PR #1659, opened 2026-08-18
perf: fold a one-row VALUES into its object and cost nested UNION honestly

bplatz:
Two independent planner defects, reported together: a bound-subject claim
lookup ran ~500x slower than the identical query with the IRI written inline,
and UNION-ing three inlined IRIs instead timed out at the gateway.

Both reproduce on a synthetic 200k-claim ledger — see the added
`values_object_scan_probe` example.

## A VALUES binding a leaf object var never reached the scan

`split_values_seeding_star` only seeds a VALUES that binds the **star's
subject**; everything else defers, and `build_property_join_block` defers
unconditionally. So `PropertyJoinOperator` drove off whatever else was bound,
drained that predicate's whole extent, and only then let the one-row VALUES
filter it. Explain showed it plainly: `Project -> Values -> PropertyJoin`.

That deferral was deliberate — seeding a leaf-object VALUES costs the star
fusion. The fix isn't to seed it: a one-row VALUES **is** a constant, so
`inline_singleton_values_objects` folds it into the object position of
same-list triples and keeps the VALUES pattern, leaving the variable bound for
projection and FILTERs.

Gated to one row, reference cells, object position, and a single contiguous
inner-join region. The region boundary is what makes it sound: compound
patterns are not inner-join members, so no region straddles a
MINUS/EXISTS/NOT EXISTS. A folded triple's donor VALUES is therefore always on
the same side of every order-sensitive boundary, which leaves those patterns'
order-preservation sets (built from *preceding* `produced_vars`) unchanged —
and covers `subquery_correlation_vars`, which reads the same thing. Without
that gate a trailing VALUES could shrink the set and let an anti-join float
ahead of the binding that correlates it, degenerating into a disjoint-domain
no-op.

## Chained UNION was costed 1e6x too high

`{A} UNION {B} UNION {C}` parses as `Union([[Union([[A],[B]])],[C]])`, so the
outer branch holding the inner UNION owns no triples of its own.
`estimate_branch_cardinality` scaled every triple-free branch by the
unknown-property-scan default, so a 30-row union costed **20,000,010** against
a 20,000-row triple and was placed behind every real scan. A 4-way union
compounds to 1e12.

That misplacement is superlinear because `UnionOperator` is correlated — it
rebuilds and re-runs each branch once per input row. A branch whose non-triple
members report sized `Source` estimates now costs those; the default still
applies when only Expander/Deferred patterns (OPTIONAL/MINUS/FILTER/BIND) are
present, where nothing is known.

## Measured

200k claims, 20k subjects. Row counts identical before and after in every case.

| Query | Before | After |
|---|---|---|
| `VALUES ?sub` + dimension | 54.3 ms | 0.0 ms |
| inlined constant + dimension (control) | 0.0 ms | 0.0 ms |
| `VALUES ?sub` + 4 more predicates | 239.3 ms | 0.1 ms |
| inlined + 4 more predicates (control) | 0.1 ms | 0.1 ms |
| `VALUES` + 4 preds + 2 OPTIONAL | 147.7 ms | 0.1 ms |
| broad triple, then 3-way UNION | 8,185 ms | 0.1 ms |
| 200k-row triple, then 3-way UNION | 53,016 ms | 0.1 ms |

## Deliberately not covered

- **Multi-row VALUES on an object var** (~92 ms, unchanged). Seeding it is a
  genuine cost decision, and blanket seeding regressed a 21k-row Cypher UNWIND
  previously. It needs a stats-driven gate, not a threshold guess.
- **`UnionOperator` re-planning per input row.** The estimator fix keeps unions
  early, which is where the 53 s came from, but a union that legitimately runs
  second still rebuilds its operator tree per driving row.

## Note for reviewers

Two patterns are classified order-sensitive in `reorder_patterns`
(planner.rs:1398-1402) but *are* inner-join members, so they do not break a
region: a variable-free FILTER and `BIND(BNODE(..))`. Folding can shrink their
preceding-produced-vars set. It is benign today — a variable-free filter is
row-independent, and the retained one-row VALUES removes no rows, so both see
an identical solution set either way — but it is a coupling across two files
worth knowing about if another order-sensitive classification is ever added.

`batched_object_join_merges_novelty` asserted the batched-object lane for a
one-row VALUES, which now folds to a constant scan. It gets a two-row VALUES to
keep exercising the lane, plus the folded form as a third case so the `@list`
retraction fate check covers both plans.

## Validation

`fmt --check` and `clippy --all --all-features --all-targets` clean;
`fluree-db-query` 1501/1501, `grp_query_sparql` 349/349, batched-overlay green.
The full workspace suite (11,140) and the W3C SPARQL suite (36 suites, 0
failures) last ran green immediately *before* the final region rewrite of
`inline_singleton_values_objects` and have not been re-run locally since —
leaving that to CI.

### DIFF
--- a/fluree-db-api/tests/it_join_batched_overlay.rs
+++ b/fluree-db-api/tests/it_join_batched_overlay.rs
@@ -508,30 +508,47 @@ async fn batched_object_join_merges_novelty() {
         .await
         .expect("novelty asserts");
 
-    let queries: &[(&str, &str, usize)] = &[
+    // (name, query, expected rows, must route through the batched-object lane)
+    let queries: &[(&str, &str, usize, bool)] = &[
         (
             "object-probe-mixed",
             r"PREFIX ex: <http://example.org/ns/>
               SELECT ?b ?x WHERE { ex:alice ex:knows ?b . ?x ex:author ?b }
               ORDER BY ?b ?x",
             2, // (bob, b3 injected) + (grace novelty-only, b4 injected);
-               // b2 retracted, b1/d1 authors not known by alice
+            // b2 retracted, b1/d1 authors not known by alice
+            true,
         ),
         (
             "object-probe-list-retract",
+            // Two rows on purpose: a SINGLE-row VALUES is folded into the
+            // object position by the planner (`inline_singleton_values_objects`)
+            // and takes the plain bound-object scan instead of this lane.
+            // `ex:dave` has no `ex:authors` edge, so the answer is unchanged.
             r"PREFIX ex: <http://example.org/ns/>
-              SELECT ?x WHERE { VALUES ?b { ex:bob } ?x ex:authors ?b }
+              SELECT ?x WHERE { VALUES ?b { ex:bob ex:dave } ?x ex:authors ?b }
               ORDER BY ?x",
             1, // the delete's WHERE dedupes the repeated @list value to one
-               // binding, so exactly ONE of the two o_i entries is retracted:
-               // the fate check must discriminate the entries by o_i (a
-               // default-read o_i would either resurrect both or drop both)
+            // binding, so exactly ONE of the two o_i entries is retracted:
+            // the fate check must discriminate the entries by o_i (a
+            // default-read o_i would either resurrect both or drop both)
+            true,
+        ),
+        (
+            // The folded form of the same `@list` retraction, which is what a
+            // one-row VALUES now plans as. Same answer, different lane.
+            "object-const-list-retract",
+            r"PREFIX ex: <http://example.org/ns/>
+              SELECT ?x WHERE { ?x ex:authors ex:bob }
+              ORDER BY ?x",
+            1,
+            false,
         ),
     ];
 
     let view = fluree.db(ledger_id).await.expect("novelty view");
     let mut novelty_results = Vec::new();
-    for (name, query, expected_len) in queries {
+    for (name, query, expected_len, expect_lane) in queries {
         let (spans, guard) = span_capture::init_test_tracing();
         let rows = run_query(&fluree, &view, query).await;
         drop(guard);
@@ -540,15 +557,17 @@ async fn batched_object_join_merges_novelty() {
             *expected_len,
             "{name}: row count under novelty; got {rows:?}"
         );
-        assert!(
-            spans.has_span("join_flush_batched_object_binary"),
-            "{name}: bound-object lane should engage under novelty; spans: {:?}",
-            spans.span_names()
-        );
-        assert!(
-            spans.has_event("join batched object flush merged novelty overlay"),
-            "{name}: object flush should merge the overlay"
-        );
+        if *expect_lane {
+            assert!(
+                spans.has_span("join_flush_batched_object_binary"),
+                "{name}: bound-object lane should engage under novelty; spans: {:?}",
+                spans.span_names()
+            );
+            assert!(
+                spans.has_event("join batched object flush merged novelty overlay"),
+                "{name}: object flush should merge the overlay"
+            );
+        }
         novelty_results.push(rows);
     }
 
@@ -557,7 +576,7 @@ async fn batched_object_join_merges_novelty() {
         .await
         .expect("reindex ground truth");
     let view = fluree.db(ledger_id).await.expect("indexed view");
-    for ((name, query, _), novelty_rows) in queries.iter().zip(&novelty_results) {
+    for ((name, query, _, _), novelty_rows) in queries.iter().zip(&novelty_results) {
         let indexed_rows = run_query(&fluree, &view, query).await;
         assert_eq!(
             &indexed_rows, novelty_rows,
--- a/fluree-db-query/src/execute/where_plan.rs
+++ b/fluree-db-query/src/execute/where_plan.rs
@@ -1805,6 +1805,134 @@ fn elide_redundant_type_filters(
     )
 }
 
+/// The constant term a single-row VALUES cell stands for, when it is a reference.
+///
+/// Only reference bindings convert: a literal cell would have to reproduce the
+/// datatype/language matching rules that `TriplePattern::dtc` and the scan layer
+/// apply to `Term::Value`, and getting that subtly wrong changes results rather
+/// than timings.
+fn values_cell_as_ref_term(binding: &crate::binding::Binding) -> Option<Term> {
+    match binding {
+        crate::binding::Binding::Sid { sid, .. } => Some(Term::Sid(sid.clone())),
+        crate::binding::Binding::IriMatch { iri, .. } => Some(Term::Iri(iri.clone())),
+        _ => None,
+    }
+}
+
+/// Substitute a single-row VALUES' IRI into the object position of the triples
+/// it constrains, keeping the VALUES pattern itself.
+///
+/// `VALUES ?o { <iri> }` *is* the constant `<iri>` — one row, one binding — so a
+/// sibling `?s <p> ?o` is the same pattern as `?s <p> <iri>`. Only the constant
+/// form narrows anything, though: a VALUES binding a leaf object var is deferred
+/// above the star (see [`split_values_seeding_star`] and
+/// [`build_property_join_block`]), so `PropertyJoinOperator` drives off whatever
+/// else is bound, drains that predicate's whole extent, and only then does the
+/// one-row VALUES filter it. Measured on a 200k-claim ledger: 54 ms versus 0.0 ms
+/// for the same query with the IRI written inline, and 240 ms versus 0.1 ms once
+/// the star widens to five predicates.
+///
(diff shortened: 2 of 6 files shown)


fluree-db-query/src/execute/where_plan.rs
```diff
@@ -1805,6 +1805,134 @@ fn elide_redundant_type_filters(
...
+            if let Ref::Var(v) = &tp.s {
+                non_object_vars.insert(*v);
+            }
+            if let Ref::Var(v) = &tp.p {
+                non_object_vars.insert(*v);
+            }
+        }
+    }
+
+    let is_inner_join_member = |p: &Pattern| {
```
aaj3f, line 1878: *Praise.* The region-boundary design is the right shape of gate: deriving soundness from "compound patterns are not inner-join members" means the fold can't straddle a MINUS/EXISTS/subquery by construction, rather than by a list of special cases that would rot. The doc comment explaining why each relaxation is deliberately not taken is exactly what the next person needs.

fluree-db-query/src/execute/where_plan.rs
```diff
@@ -1805,6 +1805,134 @@ fn elide_redundant_type_filters(
...
+        matches!(
+            p,
+            Pattern::Triple(_) | Pattern::Values { .. } | Pattern::Bind { .. } | Pattern::Filter(_)
+        )
+    };
```
aaj3f, line 1883: *Optional.* The "Note for reviewers" coupling you called out (variable-free FILTER and `BIND(BNODE(..))` are order-sensitive in `reorder_patterns` but inner-join members here, so folding can shrink their preceding-produced set) is real, and I agree it's benign today — a variable-free FILTER is row-independent, and the retained one-row VALUES removes no rows.

But nothing pins it: if someone adds a third order-sensitive `Deferred` classification at `planner.rs:1398-1402` that IS an inner-join member, the fold silently starts shrinking its ordering set with no test going red. A cheap insurance move would be a unit test asserting the fold still fires with a variable-free FILTER in the region (documenting the coupling as intended), plus a one-line cross-reference comment at `planner.rs:1398` pointing back at `is_inner_join_member`.

fluree-db-query/src/planner.rs
```diff
@@ -1088,7 +1092,22 @@ pub fn estimate_branch_cardinality(patterns: &[Pattern], stats: Option<&StatsVie
...
+        // multiplying by the unknown-scan default would invent rows that are not
+        // there. This is the common shape of a chained UNION: `{A} UNION {B} UNION
+        // {C}` nests as `Union([[Union([[A],[B]])], [C]])`, so the outer branch
+        // holding the inner UNION has no triples of its own. Scaling it by
+        // DEFAULT_PROPERTY_SCAN_SELECTIVITY put a 30-row union at 20M rows, which
+        // pushed it behind every real scan; the correlated UnionOperator then
+        // rebuilt and re-ran all three branches once per driving row.
+        //
+        // With no sized source (only OPTIONAL/MINUS/FILTER/BIND) nothing is known,
+        // so the unknown-scan default still applies.
```
aaj3f, line 1114: *Optional, more of a question than a suggestion.* With `has_sized_source`, a triple-free branch whose only sized member is a `Service` (or a `Graph` wrapper) is now costed at `DEFAULT_SERVICE_ROW_COUNT` (or the recursive estimate) instead of that × the 20M unknown-scan default — so a SERVICE-bearing union can now be placed much earlier.

I *think* that's the safe direction given `UnionOperator` is correlated (earlier placement means fewer driving rows and fewer branch re-runs), and the estimate is more honest either way, but it did widen the change beyond the nested-UNION shape in the PR title. If you've thought about the SERVICE case and are comfortable, no action needed.

fluree-db-api/tests/it_join_batched_overlay.rs
```diff
@@ -508,30 +508,47 @@ async fn batched_object_join_merges_novelty() {
...
+    let queries: &[(&str, &str, usize, bool)] = &[
         (
             "object-probe-mixed",
             r"PREFIX ex: <http://example.org/ns/>
               SELECT ?b ?x WHERE { ex:alice ex:knows ?b . ?x ex:author ?b }
               ORDER BY ?b ?x",
             2, // (bob, b3 injected) + (grace novelty-only, b4 injected);
-               // b2 retracted, b1/d1 authors not known by alice
+            // b2 retracted, b1/d1 authors not known by alice
+            true,
```
aaj3f, line 520: *Praise.* Reworking `batched_object_join_merges_novelty` to a two-row VALUES so the batched-object lane stays exercised, *and* adding the folded form as a third case so the `@list` retraction fate check covers both plans, is exactly the "don't lose the lane under the optimization" discipline that keeps these tests meaningful.

aaj3f (approved):
This is great, @bplatz

I verified the order-preservation argument end-to-end (`reorder_patterns` builds anti-join `required_vars` from preceding `produced_vars`, and the retained VALUES keeps producing the folded var), ran the full `fluree-db-query` suite (1412 green, all 8 new unit tests present by name), the SPARQL values integration tests (31 green incl. the 9 new ones), the JSON-LD `grp_query` suite for shared-IR parity (422 green), and the probe — VALUES-vs-inlined pairs match in rows and time, and the chained-UNION-after-200k-scan case runs in single-digit ms dev-profile. 

The two inline notes (pinning the FILTER/BNODE coupling with a test, and a question about SERVICE branches under the new sized-source costing) are both optional, but if you agree on them, would rather fold them in here rather than as follow-ups.

- **Patterns/abstractions:** ✔ Extends the existing pre-planning-rewrite mechanism (sits beside `elide_redundant_type_filters`), reuses the constant-object scan path rather than adding a lane; shared-IR, so JSON-LD inherits the win for free.
- **Performance (speed first, memory second):** ✔ Query-planner hot path, strictly positive: plan-build-time O(patterns) pass, no per-row work; measured 54ms/240ms/8.2s/53s → ≤0.1ms with row counts unchanged; multi-row VALUES deliberately untouched.
- **Testing:** ✔ 8 unit + 9 integration tests, all wired into live grp targets and confirmed running by name; batched-overlay lane coverage preserved; CI ran the full gauntlet including the W3C suite (test/clippy/fmt/testsuite-sparql all pass), which covers the PR body's honest caveat about the final region rewrite not being locally re-run against the full workspace.
- **Conventions:** ✔ Thorough multi-line commit body, self-describing title, fmt/clippy clean against the denied-lint set, `docs/design/performance.md` updated with the new "Pre-planning rewrites" section.

Verified locally at branch HEAD (`a8e7d797f`): `cargo test -p fluree-db-query` 1412/1412 · `grp_query_sparql values` 31/31 · `grp_misc batched` 4/4 · `grp_query` 422/422 · probe reproduces all claimed shapes · `fmt --check` + `clippy --all-targets -- -D warnings` clean.

Approving so you can merge when ready — the two optional notes are cheap insurance additions

fluree-db-query/src/execute/where_plan.rs
bplatz: Agreed, and folded in as `219c39388`.

`fold_survives_a_variable_free_filter_in_its_region` asserts the fold still fires with `FILTER(true)` in the region, and — the part that seemed more worth pinning than "it still folds" — that both the VALUES and the FILTER survive the rewrite. The retained VALUES is the actual reason the coupling is benign, since it goes on producing the variable the ordering set is built from, so that is the invariant a future change would break.

Also added the cross-reference you suggested at `planner.rs`, and a matching one on `is_inner_join_member` pointing back, so whichever site someone edits first leads to the other.

fluree-db-query/src/planner.rs
bplatz: Good catch, and digging into it I think the answer is narrower than "is earlier placement safe" — `Service` should not have been satisfying `has_sized_source` at all. Fixed in `219c39388`.

`DEFAULT_SERVICE_ROW_COUNT` is `FULL_SCAN`, and its own doc comment says that value is chosen "high to place them late since they involve network calls to remote endpoints" — so it is a placement sentinel, not an estimate of what the endpoint returns. `has_sized_source` means "this branch knows its own cardinality, do not invent rows on top of it", which is exactly what SERVICE does not know. Counting it read "unknown, and expensive" as "known to be big".

So the widening you spotted was unintended rather than a deliberate call I had thought through. `Pattern::Service` no longer sets the flag, which keeps the documented late placement exact. The nested-UNION shape this PR is actually about is unaffected — those branches are sized by `Union`/`Values`/`Subquery` estimates, which are real recursive counts, and there is now a test for that shape alongside the SERVICE one plus a case showing the exclusion is SERVICE-specific (a sized member next to a SERVICE still sizes the branch).

Both new planner tests were checked against the unfixed code first and do go red, so they are not passing vacuously.

Since this is a behaviour change you raised as a question rather than asked for, worth a look before merge.

PR #1660, opened 2026-08-18
release v4.1.6

bplatz:
Bumps the workspace to 4.1.6.

Six files, version lines only — no dependency drift:

- `Cargo.toml` / `Cargo.lock`
- `testsuite-sparql/Cargo.toml` / `Cargo.lock`
- `testsuite-shacl/Cargo.toml` / `Cargo.lock`

The two testsuite crates are `[workspace] exclude`d, so the workspace version
does not reach them and they carry their own.

Based on `perf/absent-subject-scan-narrowing` (#1659) rather than `main`, so
this PR shows only the bump. Tag `v4.1.6` after both land on `main`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2422,11 +2422,11 @@ checksum = "8bf7cc16383c4b8d58b9905a8509f02926ce3058053c056376248d958c9df1e8"
 
 [[package]]
 name = "fluree-bench-alloc"
-version = "4.1.5"
+version = "4.1.6"
 
 [[package]]
 name = "fluree-bench-support"
-version = "4.1.5"
+version = "4.1.6"
 dependencies = [
  "chrono",
  "rand 0.8.5",
@@ -2442,7 +2442,7 @@ dependencies = [
 
 [[package]]
 name = "fluree-bench-virtual"
-version = "4.1.5"
+version = "4.1.6"
 dependencies = [
  "anyhow",
  "chrono",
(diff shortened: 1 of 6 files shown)


aaj3f (approved):
Clean bump, @bplatz 👍 

I checked the lockfile hunks mechanically and they're all workspace-package version lines, no dependency drift rode along, and the excluded testsuite crates got their own bumps as they need to. The only coordination note is the one your PR body already covers: this is stacked on #1659 so it got no CI (only Graphite's mergeability check ran — `ci.yml` fires on PRs targeting `main` only), and the `v4.1.6` tag should wait until both are on `main`. One small heads-up for the release-notes side: if #1668 (`NameServiceLookup::heads`) is meant to ship in 4.1.6, it needs to land before the tag — as based, this bump covers #1659 only.

- **Patterns/abstractions:** ✔ n/a — version metadata only.
- **Performance (speed first, memory second):** ✔ Neutral, no code change.
- **Testing:** ✔ Nothing to test; the real gate is CI on the retargeted PR plus the tag-triggered `release.yml`.
- **Conventions:** ✔ Dedicated `release/vX.Y.Z` PR per the two-phase flow; synchronized workspace version plus the two excluded testsuite crates; self-describing title.

Verified locally at branch HEAD (`89a759fcc`): diff is exactly 6 files / ±100 lines, all version lines; lock hunk `name =` contexts all workspace packages; no residual `4.1.5` in any non-lock toml.

Approving — just retarget to `main` once #1659 lands, and mind the #1668 ordering if it's meant to make this release.

PR #1663, opened 2026-08-18
fix(materialize): bound the full read a source falls back to when its watermark expires

christophediprima:
## The bug

A materialize source whose stored watermark snapshot has been expired by the source table's snapshot retention cannot prove incremental safety, so it falls back to a full table read. That fallback is unbounded, and being unbounded makes it self-sustaining rather than merely slow:

1. A full read is the most expensive read the table has, so it is the most likely to exhaust the target's novelty and defer.
2. A deferral writes no watermark.
3. So the next poll performs the same full read — and every poll after that, indefinitely. Nothing recovers it.

One deployment had 13 of 17 tables in exactly this state, each reading its whole self on every poll, with no watermark moving for hours.

## The fix

A full read cannot be bounded by snapshot, because there is no `from` to take a prefix after. So bound it by rows and checkpoint at the snapshot that prefix corresponds to.

- `FileScanTask` gains `data_sequence_number`. Iceberg assigns it per commit and never reuses or reorders it, so it is the only value on a task that says when the rows arrived — paths are arbitrary. Both planners already computed it and then dropped it; they now attach it. `Option`, not a sentinel: a task the planner cannot attribute is "unknown", and a consumer must not read that as "oldest".
- Both plans are sorted by `(data_sequence_number, file_path)`. Manifest traversal is sequential and manifest files are immutable, so for a pinned snapshot the order was already stable — but only incidentally, and a resumed scan against a different order either repeats files or skips them. Sequence first with path only as a tiebreak is the point: across a resume boundary, if a key's newer row is processed in one pass and its older row in the next, the older one lands last and wins.
- A row-budgeted prefix of that sorted plan is therefore a prefix in **commit** order. `snapshot_at_or_before_sequence` names the snapshot that prefix leaves the target at, and reporting that as the scan's `to` makes the caller's existing watermark write land there. No new vocabulary, no schema change, and the crash-safety ordering is unchanged — the watermark still follows the data.
- `FLUREE_MATERIALIZE_MAX_ROWS_PER_FULL_PASS`, default `250_000`; `0` restores the previous read-it-all behaviour.

Two rules decide correctness, and both are asserted rather than assumed:

- **The cut extends to the end of its commit.** Stopping inside one leaves the target holding part of a commit, which no snapshot names, and a state that cannot be named cannot be resumed from.
- **The checkpoint rounds DOWN**, to the newest ancestor whose sequence is fully applied. Rounding up would write a watermark claiming rows that were never applied — silent data loss — and the next scan would start after them.

Every reason to decline returns "read it whole", which is today's behaviour: the bound is disabled, the plan already fits, the cut would land on the head anyway, or any task lacks an attributable sequence number. That last one matters — without a sequence there is no safe cut, and an expensive honest read beats a wrong checkpoint.

## The third commit is the one that makes the other two work

`snapshot_at_or_before_sequence` was first written using `snapshot_window(None, to_id)`, which walks the whole history to the root and errors on any missing ancestor.

The tables that need a checkpoint are precisely the ones whose old snapshots have expired — that is *why* they are full-reading. So the helper errored on every one of them, and the caller fell through to the unbounded read it was supposed to prevent. **The bound could never fire on the only tables it was written for, and it failed silently: every unit test passed, because every fixture had an intact chain.**

It now walks parents from `to` and stops at the first qualifying snapshot. The walk is short — the cut lands near the start of the backlog, so the answer is a few hops from `to`, nowhere near the expired region. Reaching the root, or a parent that is gone, means there is no nameable checkpoint: that is `None`, and the caller reads the whole thing, which is correct rather than merely safe.

The regression test uses a deliberately severed chain (5 ← 6 ← 7 retained, 4 expired) because that is the shape the original fixtures could not express.

## Behaviour and compatibility

A bounded pass does not change *what* is read. A full read selects the files live at `to`; applying those at or below the cut is exactly the state at the checkpoint snapshot, and the next pass scans `(checkpoint, to]` for the rest.

The first commit is behaviour-neutral on its own: nothing consumes `data_sequence_number` yet, and the sort only makes the existing order explicit.

## Tests

Three mutations, each killed: cutting inside a commit, rounding the checkpoint up, and letting an unattributed task through. Plus a regression test for the expired-ancestor walk, on a severed snapshot chain.

`cargo clippy --all --all-targets -- -D warnings` and the same with `--all-features` are both clean; `fluree-db-iceberg` 259 pass, `fluree-db-api --lib` 756 pass.


fluree-db-api/src/graph_source/r2rml.rs
```diff
@@ -1358,7 +1651,144 @@ impl<'a> FlureeR2rmlProvider<'a> {
...
+                    to_sequence = ?keep_through,
+                    files_this_pass = kept.len(),
+                    files_total,
+                    budget_rows,
+                    complete = keep_through.is_none(),
+                    "materialize: full read resumed from a sequence cursor"
+                );
+                // `Some` => prefix only, so the caller persists the cursor and
+                // leaves the snapshot watermark alone.
+                to_sequence = keep_through;
```
aaj3f, line 1922: **🔴 Blocking — the cursor route cannot bootstrap: on the expired-history tables this PR names as its motivation, the *first* pass is still unbounded and the livelock survives.**

Commit `d674f0d0d` calls the sequence cursor "the durable fix for the livelock" and argues — correctly — that "no amount of bounding by SNAPSHOT can rescue" a table whose history has expired, because `snapshot_at_or_before_sequence` can only name a retained snapshot. But trace who ever *writes* the first cursor: `to_sequence = Some(...)` is assigned in exactly one place (this line), inside the branch that only runs when `from_sequence` is already `Some` (`cursor_cut = from_sequence.and_then(...)`, `:1683`). `from_sequence` comes from `materialize_sequence_cursor`, which is `None` on first run by design (and its doc-comment is right that absent must never read as 0 — on a v1 table every data sequence IS 0, so a zero cursor would exclude the entire table and advance the watermark over rows never applied).

So on a table whose watermark AND backlog base are expired — the "13 of 17 tables" shape the PR body leads with — the first pass goes: no cursor → checkpoint route → the cut lands in the ancient region, every retained snapshot's sequence is above it, the walk hits the expired parent → `None` (exactly what `snapshot_at_or_before_sequence_survives_an_expired_ancestor`'s `at(5) == None` case pins) → warn + unbounded whole read. If that read defers — the motivating failure — nothing is recorded and the identical read repeats next poll. The mechanism built for these tables can't take its first step on them; it only ever engages once the *checkpoint* route has bootstrapped a cursor, which requires retained history.

The fix looks like one arm: in the unnameable-checkpoint decline (`outcome` match, `:1726-1750`), instead of falling through to `plan.tasks`, keep the `full_read_prefix` prefix and report `to_sequence = Some(cut_seq)` — `watermark_node` already writes cursor XOR snapshot, so the next poll resumes above the cut via the cursor route, and the pass count is exactly what the cursor path would have produced. It's v1-safe for free: an all-zero-sequence table lands in `SingleCommit` inside `full_read_prefix` and still declines. And it's testable with the fixture shape you already built for the walk — severed chain + over-budget plan → first pass must yield a kept prefix and `to_sequence: Some`, not an unbounded read. Notably, `sequence_prefix_cut_resumes_above_the_cursor_without_naming_a_snapshot` already proves the machinery handles a cursor at the very start of history — the wiring just never supplies one.

fluree-db-api/src/graph_source/r2rml.rs
```diff
@@ -244,6 +261,276 @@ impl ScanChoice {
...
+    })
+}
+
+/// The novelty ceiling this deployment commits against, in bytes.
+///
+/// Read from `FLUREE_REINDEX_MAX_BYTES` — the SAME variable that configures
+/// `IndexConfig::reindex_max_bytes`, which `at_max_novelty` compares novelty
+/// against (`novelty.size >= reindex_max_bytes`) — so the budget below and the
+/// wall it has to fit under cannot drift apart.
+fn novelty_ceiling_bytes() -> i64 {
```
aaj3f: **🔴 Blocking — `novelty_ceiling_bytes()` reads the env var, not the configured ceiling, so flag- and config-file-configured deployments derive the budget from the wrong wall — resurrecting the "bound that cannot commit."**

The derivation's whole argument (commit `82317e709`: "read from the same `FLUREE_REINDEX_MAX_BYTES` that configures it so the two cannot drift") only holds for deployments that configure the ceiling through the environment. `--reindex-max-bytes` is a clap flag with `env = "FLUREE_REINDEX_MAX_BYTES"` as *fallback* (`fluree-db-server/src/config.rs:426`), and there's also the `[server.indexing] reindex_max_bytes` config-file key and `FlureeBuilder::with_indexing_thresholds` for embedded callers — in all three of those, `at_max_novelty` enforces the operator's value while `novelty_ceiling_bytes()` sees no env var and silently falls back to `default_reindex_max_bytes()` (a sysinfo RAM-fraction, `server_defaults.rs:48`).

When the real ceiling is smaller than that default — the 8 MiB pin from your own production narrative is exactly this shape — the derived budget is again too big to commit, the pass defers, no watermark or cursor is written, and we're back to the livelock this PR exists to kill, now with a log line claiming the bound is derived "so the two cannot drift." The drift direction can also produce `PlanFits`-forever (the `c5cbdd7bb` failure) when the default exceeds the table.

The provider already holds `&Fluree`, and `Fluree.index_config.reindex_max_bytes` is the resolved value the process actually enforces — thread that in (make `materialize_max_rows_per_full_pass` take the ceiling as a parameter; keep `FLUREE_MATERIALIZE_MAX_ROWS_PER_FULL_PASS` as the explicit override). That also removes the `OnceLock` on the ceiling, which today freezes the first-seen env value for the process lifetime and makes the derivation untestable end-to-end — `rows_for_ceiling` being split out for testability was the right instinct; plumbing the real ceiling finishes the job.

fluree-db-iceberg/src/scan/send_planner.rs
```diff
@@ -153,16 +153,32 @@ impl<'a, S: SendIcebergStorage> SendScanPlanner<'a, S> {
...
                     data_file,
                     projected_field_ids.clone(),
                     self.config.filter.clone(),
                     Arc::clone(&schema_arc),
-                );
+                )
+                // A full read carries it too. This is the read a consumer falls
+                // back to when its watermark expired, so it is exactly the read
+                // that most needs to be resumable.
+                .with_data_sequence_number(eff_seq);
```
aaj3f, line 169: *Accuracy nit.* Commit `4b334e350` says "Both planners already computed it and then dropped it; they now attach it" — the non-Send `ScanPlanner`'s `plan_scan`/`plan_scan_for_snapshot` (`planner.rs:254/:264`) still don't attach a sequence. Harmless (its consumers get `None` and every `None` declines the cut, and materialize uses `SendScanPlanner` exclusively — I checked call sites), but worth a word so nobody trusts ordering from the non-Send planner.

(Commenting here because `planner.rs:254` is not in this diff.)

fluree-db-api/src/graph_source/r2rml.rs
```diff
@@ -244,6 +261,276 @@ impl ScanChoice {
...
+    }
+}
+
+/// The commit sequence to stop a full read at, or why it could not be cut.
+///
+/// Walks `tasks` (already sorted by `(data_sequence_number, path)`) accumulating
+/// rows, and returns the sequence of the commit the budget ran out in. The whole
+/// of that commit is kept: a cut INSIDE a commit leaves the target in a state no
+/// snapshot names, and an unnameable state cannot be checkpointed.
+fn full_read_prefix(tasks: &[fluree_db_iceberg::scan::FileScanTask], max_rows: i64) -> FullReadCut {
```
aaj3f, line 521: *Future-proofing, one comment.* The cross-cut delete-file hazard (a delete file with sequence above the cut targeting data files at/below it) is structurally unreachable today because `mor_guard` fails closed on delete-bearing snapshots before any of this runs — but the day MoR delete support lands, a row-budgeted prefix silently becomes unsound. A one-line comment naming that dependency next to the cut logic would make the future work trip over it instead of past it.

fluree-db-api/tests/it_iceberg_local_fs.rs
```diff
@@ -134,7 +134,7 @@ async fn local_table_end_to_end() {
 
     // Full unpinned streaming read: all 5 rows.
     let scan = provider
-        .scan_for_materialize_stream(gs, "silver.people", &[], None, None)
+        .scan_for_materialize_stream(gs, "silver.people", &[], None, None, None)
```
aaj3f, line 166: *Observation, not a demand.* The unit coverage here is genuinely strong (and I verified all six mutations die by name — the three you claimed plus three more), but both production surprises in this PR's own history (`af8269e64`'s intact-chain fixtures, `c5cbdd7bb`'s individually-correct arithmetic) were *composition* failures no unit test saw. One `it_` test that drains an over-budget local-fs table in multiple passes through the real state ledger would pin the loop the way nothing else can — and the ceiling plumbing (parameterized instead of `OnceLock` env reads) is what makes such a test writable. It does depend on the ceiling plumbing landing first, so it's a real scope call — but given that both production surprises here were composition failures, I'd rather see this PR grow to include it than watch it wait in the backlog.

fluree-db-api/src/graph_source/r2rml_materialize.rs
```diff
@@ -918,10 +998,25 @@ fn watermark_node(
             table_name,
         )),
     );
```
aaj3f, line 1423: *Praise.* The cursor-XOR-snapshot encoding, with the ratchet documented in the field's own doc-comment and pinned by `a_partial_pass_writes_the_cursor_and_never_advances_the_snapshot`, is exactly how an invariant that once discarded 80% of a deployment's entities should be nailed down. Same for the round-down checkpoint walk and its severed-chain test — the "note what is absent from its tests" framing in `d674f0d0d` is the kind of commit-message reviewers dream about.

aaj3f (changes requested):
@christophediprima this is clearly a real issue and I'm grateful for the PR and the engineering narrative established over the six commits. The test discipline also held up under adversarial checking (I ran six mutations, your claimed three included, and every one died by the named test). 

But two wiring gaps undercut the headline before it can deliver on its value. 

First, the cursor route can't bootstrap: `to_sequence` only becomes `Some` inside the branch that requires an existing cursor (`r2rml.rs:1683/:1713`), and the checkpoint route can't name a snapshot on expired-history tables (your own walk test's `None` case). The consequence is that on the "13 of 17 tables" shape this PR is named for, the *first* pass is still the unbounded read, and if it defers the livelock persists exactly as before. 

Second, `novelty_ceiling_bytes()` reads `FLUREE_REINDEX_MAX_BYTES` from the environment while flag-, config-file-, and builder-configured deployments enforce a different ceiling. For them the budget is derived from a sysinfo default and the "bound that cannot commit" (or the `PlanFits`-forever variant you just fixed) comes straight back.

Both fixes are small and local: write the first cursor from the unnameable-checkpoint decline arm (v1-safe via `SingleCommit`), and thread `self.fluree.index_config.reindex_max_bytes` into the budget instead of the env re-read. Details and test shapes inline.

- **Patterns/abstractions:** ✔ attaches data the planners already computed, reuses the existing watermark write by reporting a shorter `to`, keeps the cursor additive beside `lastSnapshotId`--no parallel constructs.
- **Performance (speed first, memory second):** ✔ background materialize path; metadata-only sort; bounded passes cut peak memory and novelty pressure; pre-filter `record_count` budgeting errs safe. No query-hot-path risk — the blockers are availability, not speed.
- **Testing:** ⚠️ excellent unit coverage, all 13 new tests verified running by name and six mutations verified killed — but the two blocking gaps are both composition seams no test crosses, and the PR's own history shows that's where this feature gets hurt.
- **Conventions:** ⚠️ commit bodies are exemplary; fmt/clippy clean (verified package-scoped; CI all-features green); the PR *body* is stale against the branch (it seems to only cover the first few commits) and should be refreshed since it feeds release archaeology.

Verified locally at branch HEAD: `fluree-db-iceberg` 281/281, api `graph_source::r2rml` 88/88 under `--features iceberg,native`, both `it_iceberg_*` targets green, six mutation checks red-then-restored, delete-file/mor_guard and compaction convergence traced through `plan_incremental`.

The bones here are exactly right, just be sure the first-pass cursor bootstrap and the real-ceiling plumbing land before this merges (they're the difference between "fixed the livelock" and "fixed the livelock for env-var deployments whose history hasn't expired yet")

fluree-db-api/src/graph_source/r2rml.rs
bplatz: Addressed in 18b9e223a. The unnameable-checkpoint arm now keeps the `full_read_prefix` prefix and records the cut as the sequence cursor (`to_sequence = Some(cut_seq)`), so the first pass on an expired-history table writes the cursor and the next poll resumes above it via the existing cursor route. The choice is a small pure function, `route_full_read_cut`: a retained ancestor below the head is a checkpoint; `None`, or the head itself, is a cursor. v1 tables and unattributed plans still decline inside `full_read_prefix` (`SingleCommit` / `NoSequence`), as you noted.

End to end, `it_iceberg_local_fs::an_expired_history_table_drains_in_bounded_passes` drains a committed pyiceberg fixture (`silver.people_backlog`: five 600-row appends, first three snapshots expired) in three passes of 1200/1200/600 rows with the cursor at 2, then 4, then retired, and the snapshot watermark written only by the last pass. Reverting this arm fails that test at "cursor after pass 1: left None, right Some(2)".

fluree-db-api/src/graph_source/r2rml.rs
bplatz: Addressed in d9769abc9. `materialize_max_rows_per_full_pass` now takes the ceiling as a parameter and `FlureeR2rmlProvider::full_pass_row_budget()` feeds it `self.fluree.index_config.reindex_max_bytes`, the value `at_max_novelty` actually enforces. `novelty_ceiling_bytes()` and its `OnceLock` are gone; the explicit `FLUREE_MATERIALIZE_MAX_ROWS_PER_FULL_PASS` override is unchanged.

`the_full_pass_budget_tracks_the_configured_ceiling` pins it through `FlureeBuilder::with_novelty_thresholds`, and the end-to-end drain test builds its instance with `with_indexing_thresholds(64 KiB, 500 KiB)` and asserts the derived budget lands between one and two commits before running the passes. Swapping the provider back to `default_reindex_max_bytes()` fails both (the e2e one reports a budget of ~23.8M rows).

fluree-db-iceberg/src/scan/send_planner.rs
bplatz: Fixed rather than footnoted, in ce3affc9b: `ScanPlanner::plan_scan_for_snapshot` now attaches `effective_sequence_number` to every task and sorts by `(data_sequence_number, file_path)`, mirroring `SendScanPlanner`, so the ordering contract is the same from either planner.

fluree-db-api/src/graph_source/r2rml.rs
bplatz: Added in 18b9e223a, on `full_read_prefix`'s doc comment: the prefix is only sound because `mor_guard` refuses merge-on-read delete files; a delete file sequenced above the cut targeting data at or below it would be applied by a later pass against rows already committed as live, so MoR delete support has to keep the deletes that target the prefix, or decline.

fluree-db-api/tests/it_iceberg_local_fs.rs
bplatz: Grown into the PR, in 7a8dec1d2. `silver.people_backlog` is a committed pyiceberg fixture (five 600-row appends, snapshots 1–3 expired via `expire_snapshots`, generator extended in `scripts/local/write_local_iceberg_table.py`). The test drains it through a file-backed instance with a background indexer and a 500 KiB ceiling, retrying a deferred target after `wait_for_idle` the way the worker's poll would, and asserts per pass: rows read (1200/1200/600), cursor (2, 4, none), snapshot watermark (unset until the complete pass), then an empty incremental poll and all 3,000 rows present exactly once including the rows either side of each cut.

One thing the merge with main surfaced, handled in 788db8f29: main's applied markers are keyed on the window's `to`, so a partial pass must not earn one (it holds a prefix of `to`, and a marker would make the next poll skip the target for the rest of the drain), and a complete pass now retires the cursor a partial one left, after the watermark write. An engine-level test drives that round trip through the real state ledger with a fake source.

PR #1678, opened 2026-08-22
fix(consensus): extract fluree-raft-core and add a replicated kv fragment; fix two waiter-map defects

bplatz:
Extracts the application-agnostic Raft substrate out of `fluree-db-consensus` into a new `fluree-raft-core` crate, so other apps can use the raft substrate without forking and copying it.

Stacked on #1677.

## Why

`fluree-db-consensus::raft` mixes two things: generic Raft plumbing (durable log and snapshot storage, node identity, HTTP transport, membership admin, follower→leader forwarding, rendezvous ownership) and nameservice-specific behavior (the command queue, staged receipts, waiter resolution, ledger-cache watermarks). Only the second is about naming ledgers. A second consumer today would have to copy the first.

## What moved

`fluree-raft-core` — no `fluree-db-*` dependencies, by design:

- **storage** — fs + memory backends behind traits over opaque `Vec<u8>` payloads
- **node / group** — `NodeId`, `ClusterNode`, and a validated `GroupId` newtype
- **ownership** — rendezvous (HRW) hashing, generalized off `RefKey`
- **network / admin / forward** — genericized over the type config; routers carry no path prefix, so a host nests them at `/raft` for one group or under a `GroupId` for several
- **config** — `FlureeRaftConfig`, a blanket-implemented profile pinning the six openraft associated types that are genuinely fixed, leaving `D` and `R` to the application
- **state_machine** — the `AppStateMachine` / `StateMachineObserver` seam
- **runtime** — `RaftGroup::bootstrap`, the leader-task lifecycle
- **testing** — an adapter conformance fixture
- **kv** — a replicated key/value fragment (new; see below)

Everything is re-exported at its historical path, so `fluree_db_consensus::raft::storage::…`, `::NodeId`, and `::ClusterNode` keep resolving. `runtime` is re-exported the same way, and the server calls `runtime::default_raft_config` / `runtime::check_election_timeout` rather than keeping its own copies, so the election-timeout livelock invariant has one home. No nameservice behavior, wire format, or route path changes.

`openraft` sits behind a default-off `raft` feature on the new crate, verified absent from the default dependency graph — monolithic builds, which reach this crate only for `http::is_hop_by_hop`, do not compile or link it.

## The state-machine seam

An adapter does two things, and mixing them is where the bugs are. `AppStateMachine` is the pure reduction — no clocks, no RNG, no IO. `StateMachineObserver` captures effects **while the state write lock is held** and publishes them **after it drops**, so a subscriber reading state back cannot re-enter `apply`, and never sees a half-applied batch.

The application owns its snapshot format: the adapter stores whatever `encode_snapshot` returns, verbatim. The versioned `codec` helper is offered, not imposed.

## The `kv` fragment

A key/value fragment an application embeds in its own state machine — not a service, not its own group. A lease fences the work it guards only if both are ordered by the *same* log; a second group reintroduces the two-independent-clocks problem it exists to solve.

- **An entry's version is the Raft log index of the write that created it**, not a per-key counter. Counters reset on delete or expiry, so a token can repeat — and a repeating fencing token is not a fence.
- **Expiry is logical absence**, kept invisible across *partial* sweeps by a monotonic logical-time floor every reclamation raises. Without it, a rolled-back clock sees reclaimed records as absent and same-expiry survivors as live, so what a caller can see depends on how far the last sweep got.
- **Every CAS failure returns the current record** — this saves each consumer a racing follow-up read, and is the recovery path for a lost response: a holder recognizes its own committed renewal in what comes back.
- **TTLs are rejected, never clamped.** Silent truncation turns "why did my entry vanish" into an incident, and clamping `ttl_ms` would not bound `now_ms + ttl` anyway.
- **A snapshot carries records and the floor only.** The expiry index and byte total are rebuilt on decode, so a corrupt or wrongly-migrated snapshot cannot arrive able to underflow the byte total or arm a stale index entry that deletes a live record.
- Tenancy is the application's composition — `BTreeMap<Tenant, KvFragment>` keyed by an **append-only** enum, because postcard is positional: appending a struct field breaks every deployed snapshot, appending an enum variant does not. Discriminants are pinned by golden-bytes tests.

`kv` is gated independently of `raft`: it is pure state plus a pure reduction, so a consumer can hold the semantics without linking openraft. `kv::sweep` (needs both) is the leader-only eviction driver, and owns the two details that otherwise get reimplemented wrong per consumer — re-propose immediately with the **same** cutoff, and propose nothing when nothing has expired.

## Catch-up batch sizing

openraft batches up to `max_payload_entries` (300 by default) per append-entries RPC and, in 0.9, does not shrink a batch the peer refuses. Against `kv`'s default 1 MiB value limit that is ~300 MiB into a 64 MiB body cap: a follower far enough behind gets a 413, the transport reports a network error, and openraft retries the same oversized batch forever. One node silently never rejoins while the cluster looks healthy.

`RaftGroupConfig::max_command_bytes` is a declaration, and bootstrap **caps** the batch to fit rather than warning — the failure is silent, total, and only reachable from an already-degraded node. The nameservice group's commands are uniformly small and leave it unset.

## Bug fixes shipped here

Not a pure refactor — three defects are fixed alongside the extraction, two of them in the waiter path.

**Follower `WaiterMap` leak.** A follower accumulated a waiter entry per applied queue command, forever. The root cause was buffering: it cannot distinguish a late proposer from an absent one. Waiters are now tracked by *local interest* — armed by `request_cid` before proposing, bound to a `queue_id` when this node applies the enqueue. Because the bind happens during apply it strictly precedes any later terminal command, so the race closes structurally and the buffering is deleted rather than bounded. A follower arms nothing and tracks nothing; `WaiterTicket` cleans up on drop, so an abandoned submission leaves nothing behind.

**Displaced `WaiterTicket` evicted the live holder.** Both waiter maps are keyed by something a second submission can collide on — `request_cid` for interests, `queue_id` for bound waiters — and `bind` documents that collision as supported: a duplicate joining an in-flight entry displaces the earlier waiter, whose receiver errors so the caller retries under its idempotency key. But `WaiterTicket::drop` removed by key alone, so a displaced ticket deleted the binding the *current* holder was waiting on — the terminal `ApplyHead` then found no waiter, and the live submission timed out on a commit that had actually applied. The `bound: Arc<OnceLock<u64>>` is already shared between a ticket and its interest, so it serves as an identity; both maps now remove conditionally on `Arc::ptr_eq`. Not reachable through the server as shipped — `CachingCommitter` refuses a concurrent same-key submission with `AlreadyInFlight`, anonymous submissions never get `InFlight`, and a ticket's own retry re-proposes the same `request_cid` so `bind` is a no-op rather than a displacement — so this is a defect in the primitive's contract rather than a live bug, but the primitive is what the workflow and entity-resolution groups build on.

**`NetworkConfig` split.** Per-request settings (timeouts, body caps) are `RaftTransportConfig` and may differ per group; client settings (connect and pool-idle timeouts) are `HttpClientConfig`, because they are baked into the shared `reqwest::Client` and cannot vary per request. Previously a co-hosted deployment would silently get whichever group's connect timeout happened to build the client. The no-redirects SSRF guarantee is now carried by the `RaftHttpClient` newtype, so an injected client cannot quietly reinstate them.

## Verification

The acceptance bar for "a second consumer can do this without forking anything" is two integration tests that depend on `fluree-raft-core` alone, no `fluree-db-*` crate: `multi_node_group.rs` stands up three real HTTP nodes of a counter group, and `kv_group.rs` stands up three nodes whose state machine embeds tenanted `kv` fragments and drives lease acquire / renew / lapse / takeover through `client_write`, with followers answering from their own replicated copy.

Both adapters in the tree — the generic one and the nameservice's bespoke one — now run the same conformance fixture, which is what keeps them from quietly diverging. The persist-before-swap check needs a storage backend whose snapshot writes can be made to fail; without one it silently passes for either ordering.

The rendezvous algorithm is a wire format, so a frozen copy of the pre-extraction implementation is checked against the live one over thousands of key pairs — nodes that disagree can both claim the same key.

Where a test asserts a property, I confirmed it fails when the property is removed: sourcing a kv version from anything but the log index, dropping a renewal's version precondition, removing the time floor, trusting serialized denormalizations, advancing an eviction cutoff mid-drain, removing the idle pre-check, swapping snapshot state before persisting it, dropping the snapshot restore from `open`, returning no response for a membership entry, and reintroducing waiter buffering.

## Not in this PR

Migrating the nameservice's own adapter onto the generic seam. It is feasible — `encode_snapshot` output is stored verbatim, so the snapshot format is preserved, and the seam covers every effect — but it rewrites the path handling every production write, and running both adapters through one fixture already buys most of the anti-drift benefit. It gets its own PR.

## Docs

`docs/design/raft-core.md` is new and indexed in `SUMMARY.md`. The crate map and the Raft operations guide are updated; the latter previously called the per-route body limits hard-coded and described only the oversized-single-request cause of a 413.


fluree-db-consensus/src/raft/waiter.rs
```diff
@@ -57,384 +79,459 @@ pub enum AbortReason {
...
-/// (via `Arc`). Designed to be order-agnostic: a `resolve_*` call that
-/// lands before the matching `register` buffers the outcome, and the
-/// late-arriving `register` picks it up and completes immediately.
-/// Without that buffering, a fast leader (worker proposes `ApplyHead`
-/// before the original `EnqueueCommand` response reaches the
-/// transactor) would resolve to a nonexistent slot and the transactor
-/// would time out on the still-empty receiver — silently re-proposing
-/// under a fresh `request_cid` and producing a duplicate commit on
-/// retry.
+impl Drop for WaiterTicket {
```
aaj3f, line 163: *🟡 Medium (latent in the shipped server, but a real defect in the new primitive's contract).*

**A displaced `WaiterTicket`'s `Drop` evicts the live occupant of the same `queue_id`.** `bind` (`waiter.rs:225-237`) documents that a second interest binding to an occupied `queue_id` displaces the first and the first's receiver errors — but `impl Drop for WaiterTicket` at `:157-166` then removes by key alone: `self.map.waiters.remove(queue_id)` with no check that the entry it is removing is *its own*. So the displaced ticket, on its way out, deletes the binding the current holder is waiting on.

Concrete scenario: two keyed same-body submissions for one `request_cid` alive on the same node; both bind to `queue_id` 5, the first's `wait` returns `Displaced`, the transactor's retry loop burns through `attempts_allowed` (each `wait` now returns `Displaced` immediately because `receiver` is `None`, and each retry is a real `client_write`), returns `stranded_error`, and drops the ticket → `waiters.remove(5)` → the second holder's outcome is dropped on the floor when `ApplyHead(5)` lands, and it times out on a commit that actually applied. I reproduced this by appending a test that mirrors `binding_a_second_interest_to_one_queue_id_displaces_the_first` (`:514`) with a `drop(first)` before `resolve_applied(5, …)`: `assertion failed: second's binding must survive first's drop` (`map.len()` is 0, expected 1). That existing test only passes because it keeps `first` alive.

I want to be fair about reach: in the server as shipped I could not construct the trigger. `CachingCommitter` always wraps `QueuedTransactor` (`fluree-db-server/src/lib.rs:848-855`) and refuses a concurrent same-key submission with `AlreadyInFlight`; anonymous submissions never get `InFlight` (the queue scan at `state_machine.rs:2255` is inside `if let Some(key)`); and a ticket's own retry re-proposes the same `request_cid`, so `bind` is a no-op rather than a displacement. So today this is a footgun rather than a live bug. But the `WaiterMap` is exactly the primitive #1679 and every future consumer will build on, and its own doc comment describes the displacement path as supported behavior.

Fix is small: give `Waiter` (and `Interest`) an identity the ticket can match on and remove conditionally. The `bound: Arc<OnceLock<u64>>` is already shared between `Interest` and the ticket, so the cheapest version is to carry it into `Waiter` too and use `DashMap::remove_if`:

```rust
impl Drop for WaiterTicket {
    fn drop(&mut self) {
        match self.bound.get() {
            Some(queue_id) => {
                self.map
                    .waiters
                    .remove_if(queue_id, |_, w| Arc::ptr_eq(&w.bound, &self.bound));
            }
            None => {
                self.map
                    .interests
                    .remove_if(&self.request_cid, |_, i| Arc::ptr_eq(&i.bound, &self.bound));
            }
        }
    }
}
```

…and add `drop(first)` to `binding_a_second_interest_to_one_queue_id_displaces_the_first` before the `resolve_applied(5, …)` so the property is pinned. Related, and worth a line while you are in there: after `Displaced`, `WaiterTicket::wait` returns `Err(Displaced)` instantly forever (`receiver` is `None`), so the retry loop in `queued_transactor.rs:207-216` spends its remaining attempts on proposals that cannot be waited on — it may be right to treat `Displaced` as terminal rather than retryable.

fluree-db-server/src/raft.rs
```diff
@@ -409,7 +413,7 @@ impl RaftBootstrapConfig {
 /// at runtime so an override of *either* config that breaks it is
 /// caught even when this constructor isn't used.
 fn default_raft_config(network: &NetworkConfig) -> RaftConfig {
```
aaj3f: *optional.* This is more of a question than a suggestion. Commit `b81683c` says `default_raft_config` and the election-timeout check "move across from the server", and they did land in `fluree-raft-core/src/runtime.rs:116` and `:183` — but the server kept its own `default_raft_config(network: &NetworkConfig)` here and its own inline copy of the `election_timeout_min <= rpc_timeout_ms` warn at `:254-262`, now reading `network.transport.rpc_timeout`. So the livelock invariant is spelled out in two places, and the next person to retune one will not know about the other. I *think* the server versions can just become `fluree_raft_core::runtime::default_raft_config(&network.transport)` and `check_election_timeout(&raft_cfg, &config.network_config.transport)` with no behavior change (same `2× rpc, 750 ms floor, 2× spread`). Minor and non-blocking — but if you agree it's right, I'd rather see it folded in now than lost in the backlog.

fluree-raft-core/src/kv.rs
```diff
@@ -0,0 +1,2196 @@
+//! A replicated key/value fragment an application embeds in its own
```
aaj3f, line 1: *comment, for Andrew to rule on rather than a finding.* `kv` and `kv::sweep` (~2.7k lines) are new functionality with no consumer in the workspace — the only state machine that embeds a `KvFragment` is `tests/support/leases.rs`. The design is careful (log-index versions so a fence never repeats, the monotone time floor, denormalizations rebuilt on decode, append-only postcard enums pinned by golden bytes) and I have no correctness finding against it; it also has its own feature flag, so it costs nothing to non-users. It would review more cleanly as its own PR, but bplatz has standing to bundle it, and the commit bodies name the intended consumers (workflow and entity-resolution groups). Flagging the scoping question, not blocking on it.

fluree-raft-core/src/storage.rs
```diff
@@ -26,7 +26,7 @@
 //! log entries must be durable before AppendEntries is acknowledged.
 //! Buffered/async impls that violate this break Raft's safety.
 
-use crate::raft::NodeId;
+use crate::node::NodeId;
```
aaj3f, line 29: `fluree-raft-core/src/storage/fs.rs:1` — *observation.* There is no test that opens a log/snapshot directory *written by the pre-PR binary* through the new crate. I am comfortable that compatibility holds — `storage/fs.rs`, `storage/memory.rs`, and `http.rs` are 100%-similarity renames, the entry payload is still `postcard::to_allocvec(entry)` of an unchanged `Entry<TypeConfig>`, and the ownership digest is pinned against a frozen copy — so this is cheap insurance rather than a gap. If a small golden on-disk fixture is easy to drop into `fluree-raft-core/tests/`, it is the kind of thing that catches the *next* refactor.

Commenting here because `fluree-raft-core/src/storage/fs.rs` is not in this diff.

fluree-raft-core/tests/ownership_algorithm_stability.rs
```diff
@@ -0,0 +1,131 @@
...
+use fluree_raft_core::ownership::{digest_parts, owner_for_digest, RENDEZVOUS_SEED};
+use xxhash_rust::xxh64::xxh64;
+
+/// The seed as it shipped, written out rather than imported.
+///
+/// Importing `RENDEZVOUS_SEED` here would make the reference track the
+/// implementation: changing the production seed would change most of
+/// what this file expects, and the comparison tests would keep passing
+/// while every existing cluster's ownership map silently moved.
+const REFERENCE_SEED: u64 = 0x6661_6566_5246_4252;
```
aaj3f, line 26: the frozen reference with a *literal* seed is exactly right for a wire-format hash. I bumped `RENDEZVOUS_SEED` by one and 7 tests went red across both crates (`published_seed_matches_the_frozen_reference`, `digest_and_owner_match_the_reference_implementation`, `owner_matches_the_reference_across_voter_set_shapes`, `digest_is_stable`, `owner_for_digest_is_stable`, and consensus's two `*_pre_extraction_*` tests).

fluree-db-consensus/tests/it_adapter_conformance.rs
```diff
@@ -0,0 +1,211 @@
...
+use fluree_db_consensus::raft::state_machine::{Command, NameServiceState, NewLedger};
+use fluree_db_consensus::raft::state_machine_adapter::StateMachineAdapter;
+use fluree_db_consensus::raft::TypeConfig;
+use fluree_raft_core::storage::memory::{MemoryRaftLogStore, MemoryRaftSnapshotStore};
+use fluree_raft_core::storage::{
+    RaftSnapshotStore, RaftStorage, SnapshotId, SnapshotMeta, StorageError,
+};
+use fluree_raft_core::testing::{run_all, ConformanceHarness};
+
+/// Memory storage whose snapshot writes can be made to fail.
```
aaj3f, line 37: the fault-injecting `FaultySnapshotStore` is what makes `install_persists_before_it_swaps` (`testing.rs:193`) a real test rather than one that passes for either ordering. And the fixture genuinely runs against the bespoke nameservice adapter, which is the thing that keeps two adapters honest until #1679 lands.

fluree-db-consensus/src/raft/waiter.rs
```diff
@@ -57,384 +79,459 @@ pub enum AbortReason {
...
-    /// If a [`WaiterSlot::Parked`] is already there (a stale
-    /// registration from a former leader) the prior sender is dropped
-    /// — its receiver gets a [`oneshot::error::RecvError`], which the
-    /// caller treats as "the waiter is gone, restart by re-issuing
-    /// the idempotency key."
-    pub fn register(&self, queue_id: u64, ref_key: RefKey) -> oneshot::Receiver<WaiterOutcome> {
+    /// Arming first is what closes the race: by the time this node
+    /// applies the enqueue there is already somewhere to bind, so no
+    /// terminal apply can arrive with nowhere to go.
+    pub fn arm(self: &Arc<Self>, request_cid: ContentId, ref_key: RefKey) -> WaiterTicket {
```
aaj3f, line 211: the arm-before-propose / bind-during-apply design is the correct structural close of the race, and reintroducing buffering in `resolve_with` makes `a_follower_accumulates_nothing` and `dropping_a_ticket_releases_a_bound_waiter` fail, so the leak fix is pinned. The 5-node server suite's empty-map assertion (`raft_multi_node.rs:832-856`) passes.

aaj3f (approved):
This is carefully built, @bplatz. Approving with a few small notes.

The one thing I'd encourage definitely folding in before merge is in `waiter.rs:157`: `WaiterTicket::drop` removes by key with no identity check, so a *displaced* ticket's drop deletes the live occupant's binding for the same `queue_id` (probe-verified; the existing displacement test only passes because it keeps `first` alive). I could not reach it through the shipped server — `CachingCommitter`'s in-flight claim blocks the trigger — so it's a footgun in the primitive rather than a live bug, but the primitive is what #1679 builds on and the fix is a `remove_if` on ticket identity. 

Two smaller notes: the server still carries its own copy of `default_raft_config` and the election-timeout check that `runtime.rs` now owns, and the title/body under-sell that this closes a real follower `WaiterMap` leak (release notes will read it as a pure refactor).

- **Patterns/abstractions:** ✔ Extends the existing raft plumbing into a reusable crate behind a blanket-implemented type-config bound; re-exports every historical path; no parallel construct. ⚠️ minor: the server keeps a duplicate `default_raft_config` / election-timeout check.
- **Performance (speed first, memory second):** ✔ No hot path touched. Nameservice apply path is unchanged except one `DashMap::remove` per applied enqueue that replaces an unbounded per-follower insert; `LeaderView`'s boxed future sits under an HTTP round-trip. No performance-degradation risk.
- **Testing:** ✔ 176/176 (`fluree-raft-core`) and 320/320 (`fluree-db-consensus`) under `--all-features`; 9/9 in the server's `raft_multi_node` suite; all 83 moved tests came along (now 93 + 9 residual); mutations on the ownership seed (7 red) and waiter buffering (2 red) both bite. ⚠️ `binding_a_second_interest…` should also `drop(first)` to pin the Drop-eviction fix.
- **Conventions:** ✔ Multi-line, mechanism-first commit bodies throughout; clippy `-D warnings` clean against the workspace lints; `[lints] workspace = true`, `version.workspace`, `license.workspace` on the new crate; docs (`raft-core.md`, crate map, ops guide) updated and spot-checked against the defaults. ⚠️ Title reads as a pure refactor for a PR that ships a leak fix.

Verified locally at branch HEAD `fa4e130`: `cargo fmt --all -- --check` clean; `cargo clippy -p fluree-raft-core --all-features --all-targets -- -D warnings` and `-p fluree-db-consensus` clean; `cargo nextest run -p fluree-raft-core --all-features` 176 pass; `-p fluree-db-consensus --all-features` 320 pass; `cargo check -p fluree-db-server --all-features --all-targets` clean; `cargo nextest run -p fluree-db-server --all-features --test raft_multi_node` 9 pass.

Approving so you can merge when ready, but I'd really like the `waiter.rs` `remove_if` in first — it's a one-liner and it's the seam #1679 lands on.

fluree-db-consensus/src/raft/waiter.rs
bplatz: Addressed in 3982a2a52f833640ed96d1ef03e25899466865a6 — `remove_if` on `Arc::ptr_eq`, exactly as you laid it out, plus the `drop(first)` assertion in `binding_a_second_interest_to_one_queue_id_displaces_the_first`.

Thank you for being precise about reach rather than just filing it as a bug — the `CachingCommitter` / anonymous-submission / same-`request_cid`-is-a-no-op analysis saved me re-deriving whether this needed a hotfix. It didn't, and it's now described that way in the body.

**The fix went wider than the finding.** The same defect is on the *interests* map: re-arming a `request_cid` before either ticket binds displaces the first, and the first's drop was removing the second's armed interest — so nothing is left to bind when the enqueue applies and the terminal apply has nowhere to go. Your snippet already covered it (you wrote both arms), but the probe was only on the `queue_id` side, so I added `dropping_a_displaced_interest_leaves_the_live_one_armed` to pin it. Removing by key alone fails both tests.

I audited the rest of the map before calling it done: `Waiter` is constructed in exactly one place (inside `bind`, carrying `interest.bound`), so no path can produce an entry whose `Arc` doesn't match its ticket's; and the other three removals — `bind`'s `interests.remove`, `resolve_with`'s, and `drain_all_with`'s — are map-level operations that should take whatever occupies the slot, so they stay unconditional. The only case `remove_if` now declines is a slot owned by a live ticket that will remove it itself, so nothing is stranded; `a_follower_accumulates_nothing` still passes.

On the related `Displaced` point: filed as #1688 rather than folded in, because I think it's a closer call than it looks. The retry re-proposes the same `request_cid` and `SmResponse::IdempotencyHit` is a separate arm that returns successfully — so after a displacement the retries aren't futile, they're a poll for the *other* holder's completion, and one can legitimately succeed. Making `Displaced` terminal loses that. What's actually wrong is narrower: the loop's only pacing is `ticket.wait()`, which returns instantly once displaced, so the budget fires back-to-back. Re-arming isn't a third option — `bind` only fires when this node applies the enqueue, and a retry returning `InFlight` applies nothing. The issue lays out the three options; happy to take your read on it there.

fluree-db-server/src/raft.rs
bplatz: Addressed in 11b9b3a5c6898cadc2dcf95e60e52625cdd191c6 — you were right, and "moved across" was the wrong word in that commit body; they were copied.

Both server copies are gone. `fluree_db_consensus::raft` now re-exports `runtime` alongside `storage`, `admin`, `forward` and `log_adapter`, so the server reaches it the same way it reaches the rest of the moved surface: `runtime::default_raft_config(&network_config.transport)` and `runtime::check_election_timeout(&raft_cfg, &config.network_config.transport)`. Net −52/+9.

I diffed the arithmetic and the warn text before deleting — identical, as you expected, so no behavior change. The core's doc comment already carries the full rationale including the openraft-stock-150–300 ms narrative, so nothing was lost with the server's copy. I kept one short comment at the call site for the only thing the core docs can't say: that this is the place holding *both* configs, so an override of either is caught even when the constructor wasn't used to build it.

Server `raft_multi_node` 9/9 and consensus + raft-core 497/497 after the change.

fluree-raft-core/src/kv.rs
bplatz: Ruling: it stays. There is already a consumer upstream using `kv` in anger, which is the thing that would have made it a separate PR if it hadn't existed — the code isn't speculative, it has a caller today that this crate is being extracted to serve.

Appreciate you flagging the scoping question rather than either blocking on it or letting it pass silently, and for reviewing it on the merits anyway.

fluree-raft-core/src/storage.rs
bplatz: Filed as #1689 rather than folded in — your framing that it's insurance against the *next* refactor rather than a gap in this one is right, and building the fixture properly is more work than it looks.

I carried your compatibility argument into the issue verbatim (100%-similarity renames, unchanged `postcard::to_allocvec(entry)` of `Entry<TypeConfig>`, ownership digest pinned against the frozen copy) so whoever picks it up knows what is already established and doesn't re-litigate it.

fluree-raft-core/tests/ownership_algorithm_stability.rs
bplatz: Thanks for actually perturbing the seed rather than reading the test — seven reds across both crates is the answer I wanted and couldn't have claimed on my own.

fluree-db-consensus/src/raft/waiter.rs
bplatz: Appreciated. Worth noting the seam moved slightly since you reviewed it: `Waiter` now also carries the `bound` `Arc`, so a ticket's `Drop` can tell its own entry from one that displaced it. `resolve_with` is unchanged, so the buffering mutation you ran still bites the same two tests.

fluree-db-consensus/tests/it_adapter_conformance.rs
bplatz: Thanks — the persist-before-swap ordering was the one I was least sure a test could actually catch, so confirming `FaultySnapshotStore` makes it bite rather than pass either way is the check I wanted.

PR #1682, opened 2026-08-24
feat: graph sync — whole-graph replacement committing only the delta

bplatz:
## Problem

Maintaining a graph from an external authority — an ontology in an editor, a reference table regenerated by a pipeline — needs a full-replacement verb: submit the graph's desired contents and get **one commit holding exactly what changed**. Nothing does that today:

- `upsert` derives its retraction set from the payload's `(subject, predicate)` pairs, so it structurally cannot retract what the payload omits.
- `CLEAR` + `INSERT` churns every fact and hides the real delta in history.

## Design

Sync is an Insert-shaped transaction carrying a graph directive (`Txn::sync_graph`). The payload is parsed by the normal JSON-LD insert parser and re-homed onto the caller-named target graph (a payload addressing named graphs itself is rejected — the scope is exactly one graph, never inferred from data). Staging adds a second wave after the upsert wave: every currently-asserted flake of the target graph is pushed as a retraction, and the existing mixed `FlakeAccumulator` nets retract+assert of the same fact to nothing. The staged set is therefore exactly `current − payload` retractions plus `payload − current` assertions; an identical payload stages zero flakes and takes the no-op path (no commit).

Because it rides the normal staging path, modify-policy, SHACL, uniqueness, the reifies-cascade, and novelty backpressure apply unchanged. The current-contents scan follows `CLEAR`'s policy model (not view-filtered — an authoritative replacement must not leave rows the caller cannot see); whole-graph staging keeps `CLEAR`'s memory profile, with chunked staging as the same known follow-up.

Blank nodes skolemize under a deterministic graph-scoped key, so exporters with stable labels resync bnode-rooted structures (OWL restrictions, lists) with zero churn; label-regenerating exporters still churn, and RDFC-style canonicalization is the designed follow-up.

## Surface

- **Rust**: `Fluree::sync_named_graph(ledger, graph, data, {dry_run, allow_empty})` → `{asserted, retracted, committed, t}`; `sync_named_graph_with(…, TxnOpts, Option<PolicyContext>)`; `.sync_graph(iri, json)` on both transact builders (`execute` / `stage` / `build_commit` terminals).
- **HTTP**: `POST /sync?ledger&graph[&dryRun][&allowEmpty]` and `/sync/{ledger}`; JSON-LD; forwarded like every write in peer/Raft modes; dry runs stage under the real run's policy and inline-constraint inputs.
- **CLI**: `fluree sync <ledger> --graph <iri> [-f|-e|stdin] [--dry-run] [--allow-empty] [--json] [--remote]`. Turtle converts to JSON-LD client-side. The source of desired contents is a seam (`SyncSource`) so R2RML-mapped Iceberg/CSV/spreadsheet sources plug in later as one variant, not a new command or endpoint.
- **Consensus**: `TransactionBody::JsonLdGraphSync` wired into both appliers; `RefTransactBuilder::build_commit` now returns `Ok(None)` for a no-change write (the Raft worker republishes the current head, mirroring revert's NoOp) — this also fixes a pre-existing case where a zero-flake update/upsert under Raft would poison the queued request with `EmptyTransaction`.
- `CommitReceipt` gains the assert/retract split so every verb can report both sides (Raft-applied receipts report 0/0; the applied-receipt wire format does not carry the split).

Safety rails: explicit scope (`graph` required), empty payload requires `allowEmpty`, dry-run pre-flight, target validation (absolute IRI, system graphs refused by shape) enforced at staging so every entry point meets it, `BodyKind`/`TransactionBody` variants appended last (postcard-positional in persisted Raft state; ordinals pinned by test).

## Verification

- `it_sync_graph.rs` (12): first population, identical-resync no-op, delta-only counts, dry-run, empty gate, graph scoping, stable-bnode resync, payload/system-graph/malformed-target guards, the policy-gated locked path, and the `build_commit` no-op — the sync wave, deterministic skolem key, and locked-path dispatch each proven load-bearing by mutation.
- Server `/sync` contract test, CLI end-to-end flow through the binary, docs-coverage gate, `body_kind_ordinals_are_append_only`.
- Full sweeps over core/transact/api/consensus/server all-features green (5,676/5,676); CLI suite 342/342; clippy `-D warnings` clean with test targets.
- Docs: `docs/transactions/sync.md`, `docs/cli/sync.md`, endpoint/overview updates, and a Sync Contract section in `docs/cli/server-integration.md` for external servers supporting the CLI.

## Deferred

Chunked staging for whole-graph ops; server-side Turtle payloads; RDFC bnode canonicalization; opt-in default-graph sync; mapped (R2RML) sync sources.


fluree-db-cli/Cargo.toml
```diff
@@ -47,6 +47,7 @@ csv = "1.4"
 fluree-db-novelty = { path = "../fluree-db-novelty" }
 fluree-db-nameservice = { path = "../fluree-db-nameservice" }
 fluree-db-core = { path = "../fluree-db-core" }
+fluree-graph-turtle = { path = "../fluree-graph-turtle" }
```
aaj3f, line 50: *blocking (small).* **`Cargo.lock` was not regenerated for the new `fluree-graph-turtle` dependency.**

The PR adds `fluree-graph-turtle = { path = "../fluree-graph-turtle" }` to `fluree-db-cli`, but the committed `Cargo.lock` has no `"fluree-graph-turtle"` entry under `fluree-db-cli`'s dependency list. The first `cargo` command on a clean checkout of this branch rewrites the lockfile (I saw `M Cargo.lock` after my first build, diff is the single line `+ "fluree-graph-turtle",` at `Cargo.lock:2613`).

Consequence: `ci.yml` doesn't pass `--locked` so CI silently self-heals, but anything that does (`cargo build --locked`, cargo-dist release builds) fails, and every developer who touches the branch gets a dirty tree.

Fix: `cargo update -p fluree-db-cli --offline` (or just commit the one-line `Cargo.lock` delta a build produces).

Commenting here because `Cargo.lock` is not in this diff.

fluree-db-transact/src/stage.rs
```diff
@@ -754,6 +788,45 @@ pub async fn stage(
...
+        // same known follow-up flagged on `scan_graph_flakes`.
+        if let Some((sync_g_id, sync_graph_sid)) = &sync_scan {
+            let mut sync_retractions =
+                scan_graph_flakes(&ledger, *sync_g_id, options.tracker).await?;
+            for f in &mut sync_retractions {
+                // Stamp the graph explicitly: accumulator buckets key on
+                // `flake.g`, and the payload's assertions carry the graph
+                // Sid — both sides must agree for the unchanged-fact
+                // cancellation to fire.
+                f.g = Some(sync_graph_sid.clone());
```
aaj3f: *optional (coverage).* This is more of an observation than a finding. The `f.g = Some(sync_graph_sid.clone())` stamp has a comment saying both sides must agree for cancellation to fire — but when I removed it, all 12 `it_sync_graph` tests stayed green (recompile confirmed). Novelty-resident flakes already carry `g` from `commit_flakes.rs:39`, so the stamp's only job is flakes decoded from the binary index after a reindex, and no sync test reindexes. I couldn't establish in the time whether index-decoded named-graph flakes come back with `g` set; if they don't, this stamp is exactly what keeps an identical resync of an indexed graph from churning, and the suite can't tell the difference. One test that populates, reindexes, then resyncs identically and asserts `committed: false` would pin it. Minor and non-blocking — but if you agree it's right, I'd rather see it folded in now than lost in the backlog.

fluree-db-server/src/routes/transact.rs
```diff
@@ -1154,6 +1166,182 @@ async fn upsert_local(
...
+                "asserted": report.asserted,
+                "retracted": report.retracted,
+                "committed": report.committed,
+                "dryRun": report.dry_run,
+                "t": report.t,
+            }))
+            .into_response());
+        }
+
+        execute_transaction(
```
aaj3f, line 1340: `fluree-db-server/src/routes/transact.rs:1328-1340` — *optional (wire contract; needs your call).* The committed `/sync` response is the standard transact response, so it carries no `asserted` / `retracted` / `committed` — only the dry run returns the report shape. The verb's whole point is "one commit holding exactly what changed", and a pipeline calling the real run can't learn what changed, or whether anything was committed at all, except by comparing `t` to a value it read earlier. The CLI mirrors this: `graph_sync.rs:197-222` prints the raw transact JSON for remote real runs, so the commit message's "`--json` emits the server's dry-run report shape on both paths" holds locally only. I understand the reason (the Raft `AppliedReceipt` doesn't carry the split, so under Raft the counts would read 0/0), but `committed` is derivable on every path (`flake_count > 0 || t advanced`) and the counts are real on the local-committer path. I'd rather settle the shape now than after `fluree sync --remote` scripts depend on the raw one — this is the "needs a decision the author shouldn't make unilaterally" case, so happy to talk it through here rather than defer it.

fluree-db-api/src/admin.rs
```diff
@@ -971,6 +1009,173 @@ impl crate::Fluree {
...
+            // Report the `t` of the state the delta was actually computed
+            // against, not a separately-read head.
+            let staged_against_t = ledger_state.t();
+            let stage_result = self
+                .stage_sync_transaction_tracked(
+                    ledger_state,
+                    graph_iri,
+                    data,
+                    txn_opts,
+                    None,
```
aaj3f, line 1118: `fluree-db-api/src/admin.rs:1113-1121` — *optional.* I don't *think* this matters much, but the dry run stages with `index_config: None`, so the `at_max_novelty` backpressure pre-check (`stage.rs:626-631`) never runs on a dry run while the real run would reject with `NoveltyAtMax`. It slightly weakens the "dry run fails the way the real run would" promise. Passing the ledger's `IndexConfig` through would close it. Fold-in-now if you agree.

fluree-db-transact/src/ir.rs
```diff
@@ -200,6 +200,24 @@ pub struct Txn {
...
+    /// generation (like the upsert wave): every currently-asserted flake in
+    /// the target graph is pushed as a retraction, and the mixed
+    /// [`FlakeAccumulator`](crate::generate::FlakeAccumulator) nets
+    /// retract+assert of the same fact to nothing — so the staged set is
+    /// exactly `A − B` retractions plus `B − A` assertions. An identical
+    /// payload stages zero flakes (no commit).
+    ///
+    /// The transaction's insert templates must all target this graph (the
+    /// parser re-homes them) and the graph must be a user graph — reserved
+    /// system graphs and (for now) the default graph are rejected at
```
aaj3f, line 218: `fluree-db-transact/src/ir.rs:218-219` (and `fluree-db-api/src/admin.rs:1021`) — *nit (doc accuracy).* Both docs say the default graph is "rejected at staging". There's no such check — it's unreachable because registry slot 0 has no IRI (`graph_registry.rs:133`) and nothing aliases it. That's fine today, but the wording implies a guard a future `#default`-style alias would silently bypass. Either an explicit `DEFAULT_GRAPH_ID` refusal in the `sync_scan` match at `stage.rs:697-725` or rewording to "cannot be addressed (no IRI)" would make the doc true.

fluree-db-api/src/tx.rs
```diff
@@ -2167,10 +2173,93 @@ impl crate::Fluree {
...
+        // churn; structural (RDFC-style) canonicalization is the designed
+        // follow-up for those. A caller-supplied id wins.
+        let mut txn_opts = txn_opts;
+        if txn_opts.skolem_txn_id.is_none() {
+            let scope = fluree_db_core::skolem::doc_scope(fluree_db_core::skolem::doc_id(
+                "fluree:graph-sync",
+                graph_iri,
+                0,
+            ));
+            txn_opts.skolem_txn_id = Some(format!("sync{scope}"));
```
aaj3f, line 2221: `fluree-db-core/src/skolem.rs:47-59` — *nit.* The module doc's disjointness table — the thing that exists so `fdb-` minters don't collide — lists three minters. Sync adds a fourth shape via this derivation (`tx.rs:2216-2221`: `syncd<13 base36>-0-<label>`, first `-` after `fdb-` at offset 18). It is disjoint by construction, but the table is the place someone will check next time a minter is added, so it's worth the row.

Commenting here because `fluree-db-core/src/skolem.rs` is not in this diff.

fluree-db-transact/src/commit.rs
```diff
@@ -899,6 +910,7 @@ fn finalize_state_with_base(
     // temporal metadata current so the next commit's event-time guard and
     // dual-stamp decision stay in-memory integer checks.
     let head_temporal = HeadTemporal::from_commit(&commit_record).or(base.head_temporal);
+    let (assert_count, retract_count) = count_ops(&commit_record.flakes);
```
aaj3f, line 924: *nit (perf, non-blocking).* `count_ops` is a new O(n) pass over every commit's flakes, for every verb, to populate the receipt split. It's a single predictable branch per flake and is dwarfed by the CID hash and serialization of the same flakes, so I don't think it moves anything — flagging only because it is a new pass on the commit path and nothing benches it. If there's an existing loop in `finalize_state_with_base` it could piggyback on, that'd be free.

fluree-db-transact/src/stage.rs
```diff
@@ -754,6 +788,45 @@ pub async fn stage(
...
+                f.op = false;
+                f.t = new_t;
+            }
+            tracing::debug!(
+                graph_id = sync_g_id,
+                scanned = sync_retractions.len(),
+                "graph-sync retractions generated"
+            );
+            acc.push_retractions(sync_retractions);
```
aaj3f, line 826: `fluree-db-transact/src/stage.rs:791-828` — the design is the right altitude: a second retraction wave into the *existing* mixed accumulator, so netting is `Flake`'s own `(s, p, o, dt, m)` identity nested per graph. I probed `"Alice"@en` vs `"Alice"` (1/1 both directions, tagged resync no-op) and `@list [x,y,z]` vs `[x,z,y]` (exactly 2 retract + 2 assert, identical list no-op) — both correct with no special-casing.

fluree-db-transact/src/stage.rs
```diff
@@ -754,6 +788,45 @@ pub async fn stage(
...
+        // `finalize()` is exactly `current − payload` retractions plus
+        // `payload − current` assertions — the delta. Scanned flakes carry
+        // correct `m` from storage, so (like the upsert wave) no hydration
+        // is needed.
+        //
+        // Policy model follows CLEAR (roadmap O4): the scan is not
+        // view-policy filtered — sync is an authoritative whole-graph
+        // replacement, and a view-filtered scan would leave rows the caller
+        // cannot see in place, breaking "the graph now equals the payload".
+        // Modify-policy is still enforced on the resulting flakes below.
```
aaj3f, line 803: `fluree-db-transact/src/stage.rs:1074-1090` ordering — cancellation runs before `enforce_modify_policies`, so a non-root caller who re-supplies a protected row unchanged never triggers policy on it, while omitting it fails the whole sync atomically (probed with a `#config`-driven `f:allow false` on `ex:ssn`; dry run fails identically).

fluree-db-transact/src/parse/jsonld.rs
```diff
@@ -163,6 +163,56 @@ pub fn parse_transaction(
...
+    json: &Value,
+    graph_iri: &str,
+    opts: TxnOpts,
+    ns_registry: &mut NamespaceRegistry,
+) -> Result<Txn> {
+    let explicitly_empty = json
+        .get("@graph")
+        .and_then(Value::as_array)
+        .is_some_and(Vec::is_empty);
+    let mut txn = if explicitly_empty {
```
aaj3f, line 196: `fluree-db-transact/src/parse/jsonld.rs:192-199` — the allow-empty rail can only be reached by a literal `"@graph": []`; every degenerate payload I tried (`{"@context":…}`, id-only node, `[]`, `{}`) is rejected by `parse_insert` with the graph untouched.

fluree-db-api/src/tx_builder.rs
```diff
@@ -1224,6 +1375,30 @@ impl Fluree {
             return Ok((stage_result, TxnType::Insert, commit_opts, None));
         }
 
+        // Graph sync: dedicated staging (whole-graph retraction wave). Must
+        // be dispatched before the JSON-like fallthrough, which would
+        // otherwise stage the payload as a plain default-graph insert.
+        if let TransactOperation::SyncGraph { graph_iri, json } = op {
```
aaj3f, line 1381: Commit 2 (`97c1b2f`) — the `stage_under_lock` dispatch fix and the append-last ordinal pin are exactly the two things that would have bitten in production and not in the optimistic-path tests; both are mutation-proven (`build_commit → None` goes red as `EmptyTransaction` when disabled).

aaj3f (approved):
This is a really nice solution for a fairly common scenario we've faced for years, @bplatz, and I like several of the design/implementation choices you did to fold it into the existing code: a second retraction wave into the existing mixed accumulator is a really nice pattern, and it means the delta semantics come out of `Flake`'s own identity rather than anything sync-specific. 

I went at the netting from a few angles that the shipped tests don't cover (language tags, `@list` reordering, a real non-root modify policy, and every degenerate "looks empty" payload I could think of) and it held on all of them; the `stage_under_lock` dispatch fix and the append-last ordinal pin in the second commit are the two things that would actually have bitten under Raft, and both go red under mutation. 

The one thing to maybe get in before merging is mechanical: `Cargo.lock` wasn't regenerated for `fluree-db-cli`'s new `fluree-graph-turtle` dependency, so a clean checkout dirties itself on first build. 

Beyond that my notes are a coverage gap (no sync test runs against index-resident data — the `f.g` stamp at `stage.rs:818` is unpinned), a wire-contract question I'd like your call on (the committed `/sync` response can't tell a pipeline what changed or whether anything committed), and a few doc/nit items. 

Since this is stacked on #1681 it has had no `ci.yml` run — I ran fmt, package-scope clippy `-D warnings`, and the transact/api/consensus suites locally in its place.

- **Patterns/abstractions:** ✔ extends `FlakeAccumulator` + `scan_graph_flakes` + the upsert-wave precedent; models sync as `TxnType::Insert` + directive rather than a parallel pipeline; target validation lives at staging so every entry point meets it.
- **Performance (speed first, memory second):** ✔ ordinary insert/upsert/update pay one per-transaction `Option` check; `stage_graph_mgmt` dispatch untouched; one new O(n) `count_ops` pass per commit (negligible, unbenched); sync's own O(|graph|) staging is the same cliff `CLEAR GRAPH` already exposes and the default graph is unreachable. No performance-degradation risk on the per-flake hot path.
- **Testing:** ⚠️ 12 integration tests wired into `grp_ledger` and green, plus server contract and CLI e2e; skolem key and `build_commit` no-op are mutation-load-bearing; but nothing exercises index-resident data (the `g` stamp mutation stays green) and no non-root policy test (I probed one; it passes).
- **Conventions:** ⚠️ multi-line, mechanism-first commit bodies; fmt + clippy clean; docs added (`docs/transactions/sync.md`, `docs/cli/sync.md`, server-integration contract); `Cargo.lock` not updated.

Verified locally at branch HEAD `ad34872`: `cargo fmt --all -- --check` clean; `cargo clippy -p fluree-db-{transact,api,server,consensus,core,cli} --all-targets -- -D warnings` 0 warnings; `nextest -p fluree-db-transact` 350/350; `nextest -p fluree-db-api --test grp_ledger` 157/157 with all 12 `it_sync_graph::*` by name; consensus `body_kind_ordinals_are_append_only` pass; 4 purpose-written probes green; 3 mutations (skolem key red, `build_commit` no-op red, `g` stamp green).

Approving so you can merge when ready, but do get the `Cargo.lock` line in first, and I'd like to settle the committed-response shape here before `--remote` scripts start depending on it.

Issue #1691, opened 2026-08-26
Streaming SPOT diff for graph sync — memory bounded by the delta, not the graph


bplatz:
Follow-up to #1682, which ships graph sync with the same memory profile as `CLEAR`/`COPY`/`MOVE`: staging materializes every currently-asserted flake of the target graph, plus a hash-map copy in the accumulator, before netting.

## Problem

The delta computation is memory-bounded by the **graph**, not the delta. An identical resync of a large graph materializes the whole graph, nets to zero, and commits nothing — peak memory is ~2 copies of every fact (accumulator key + bucket survivor, ~200 bytes inline per flake plus heap for literal strings) regardless of how little changed. Rough order: 10M triples ≈ 4–6 GB peak; 100M is tens of GB and an OOM kill. Under Raft the branch-owning worker node is the one that dies, mid-stage.

None of the existing guards bound it: `at_max_novelty` runs before the scan but measures only current novelty; `NoveltyWouldExceed` measures only the *surviving delta*, after materialization; fuel charges land after the scan returns its `Vec`.

#1682 adds a pre-flight size cap as a stopgap — a clear resource-limit error instead of an OOM — but the cap refuses large graphs rather than handling them.

## Design sketch: streaming SPOT diff

The pieces are already in place for a two-pointer diff that never materializes the current graph:

1. The whole-graph scan (`scan_graph_flakes`) reads SPOT order — the cursor yields sorted batches natively, and the overlay merge already resolves novelty lifecycles inline.
2. Sort the payload's assertion flakes SPOT once (`O(|payload| log |payload|)` — the payload is already in memory from JSON-LD parsing, so this is the floor).
3. Walk both sides with two pointers on `(s, p, o, dt, m)` identity (the accumulator's fact key, which deliberately excludes `t`/`op`/`g`): equal → skip; scan-only → emit retraction; payload-only → emit assertion.

Memory drops to `O(|payload| + |delta|)`.

## What makes it more than a loop

- **The accumulator is bypassed**, so the netting semantics it provides must be preserved explicitly: per-graph scoping, duplicate collapse within the payload, and the `m` (language tag / list position) component of identity. `finalize()`'s SPOT sort of the delta comes free from the merge order.
- **Modify-policy enforcement** currently sees the full staged set after cancellation; the streaming path must feed it the emitted delta only, which is the same set — but SHACL validation reads the *resulting* state and needs to keep working against a delta produced incrementally.
[...]

Issue #1689, opened 2026-08-25
Add a golden on-disk fixture for pre-extraction Raft log/snapshot directories


bplatz:
Raised in review of #1678.

## Gap

No test opens a Raft log or snapshot directory **written by the pre-extraction binary** through `fluree-raft-core`. The extraction moved `storage/fs.rs`, `storage/memory.rs` and `http.rs` into the new crate, and every existing storage test writes and reads with the same (post-move) code, so on-disk compatibility is asserted by construction rather than by test.

## Why this is insurance rather than a finding

Compatibility does hold today, and the review checked it:

- `storage/fs.rs`, `storage/memory.rs` and `http.rs` are 100%-similarity renames.
- The entry payload is still `postcard::to_allocvec(entry)` of an unchanged `Entry<TypeConfig>`.
- The ownership digest — a wire format, since nodes that disagree can both claim the same key — is pinned against a frozen copy of the pre-extraction implementation over thousands of key pairs.

So nothing is known to be broken. The value is in catching the *next* refactor, where the renames will no longer be 100% similarity and nothing will notice a layout change until a production node fails to reopen its own log.

## Ask

A small golden on-disk fixture under `fluree-raft-core/tests/`: a committed log directory and a snapshot captured from a known-good build, opened through the current `FsRaftStorage` with assertions on the recovered entries, committed index, and snapshot payload. Cheap to add once, and it makes the storage layout an explicit contract rather than an emergent one.

## Origin

Review of #1678 (`extract fluree-raft-core and add a replicated kv fragment; fix two waiter-map defects`).

Issue #1687, opened 2026-08-25
Membership-join lane's answer on list rows is unpinned at or above the driving-size gate


bplatz:
Raised in review of #1674.

## Gap

`it_fastpath_1652_regression`'s `composite join, list rows` case is the only coverage of what a var-object join answers over RDF list elements. After #1674 costs the membership lane against the planner's driving estimate, that fixture sits **below** `MEMBERSHIP_JOIN_MIN_DRIVING`, so the lane declines and the case only exercises the generic pipeline.

Nothing now pins what `MembershipJoinOperator` answers for this shape at or above the gate, and the two disagree:

- generic join — bag join over list elements: s2's two `q-1` entries pair 2×2 = 4, s3 pairs its shared `r-1` once, s1 never matches → **5**
- membership lane — semi-join, one row per driving row → **3**

That divergence was the original 1652 finding. #1674 removed the shape from the lane's reach rather than reconciling it, which is the right call for a regression fix but leaves the behaviour untested.

## Why this is not just "add a fixture above the gate"

Whether list position is part of a literal's identity is an open semantics question — #1676 explicitly deferred it in its "Not in this PR" section. Adding a fixture now would pin one of the two answers as correct before that call is made, which is why #1674 only softened the existing comment rather than adding a case.

So this is really two steps:

1. Settle whether a var-object join over list elements is a bag join (5) or a semi-join / set join (3) — the semantics decision deferred by #1676.
2. Once settled, add a fixture above `MEMBERSHIP_JOIN_MIN_DRIVING` pinning the lane to the same answer as the generic pipeline — and, if they are meant to agree, make the lane decline on list rows rather than return a different answer.

## Origin

Review of #1674 (`fix: two v4.1.6 planner regressions — star-join reorder and membership-join admission`); semantics question deferred by #1676.

aaj3f:
Ran this through the triage sweep's verification protocol (claude-assisted) — verified at `a824bbfd1`: the gate sits at `where_plan.rs:2899`/`:2923`, `membership_join_key_vars` (`:2909-2970`) has no list/`o_i` exclusion so the shape is admitted above 256 driving rows, and `membership_join.rs` is unchanged since `47e96b7b6`, so its semi-join answer stands above the gate.

Three things worth adding:

[...]

Issue #1683, opened 2026-08-25
Retractions of novelty-only predicates are invisible to the crawl read path (deleted data reappears)


bplatz:
Surfaced while reviewing #1675. Pre-existing — **reproduced at that PR's merge base `a65eb1958`**, so it is not introduced by the hydration rewrite. Filing separately because the mechanism is different from anything #1675 touches.

## Symptom

A filtered DELETE of a single `@list` entry commits with `receipt.flake_count = 1`, but the list is unchanged when queried afterwards. No error, no warning.

## Trigger

The list's predicate exists **only in novelty** — no persisted `p_id` — on an otherwise indexed ledger. Narrowing done during review:

- The same probe with the predicate also present in the indexed base **passes**.
- List position does not matter: it fails at position 0, at position 1, and at the last position.
- Staging is not the faulty component. `hydrate_list_index_meta_for_retractions` was instrumented and leaves the retraction with the correct `m = { i: 1 }`; the retraction reaches the commit intact.

That points at the query/overlay side and ephemeral (novelty-assigned) predicate ids rather than at the write path — the retraction appears to be written with a predicate id that does not line up with the one the asserted list row carries.

## Why it matters

Silent data-loss-shaped failure in the opposite direction: the caller is told one flake was retracted and the entry is still readable. Any workload that inserts a list under a brand-new predicate and then deletes from it before the next index build is affected.

## Suggested starting points

- `fluree-db-transact/src/stage.rs:1613` — hydration, confirmed correct; use it as the known-good boundary.
- Predicate-id assignment for novelty-only predicates, and how the overlay resolves them against a persisted dictionary.

## Origin

Review of #1675 (`perf: make filtered-DELETE list-meta hydration O(novelty), not O(retractions × novelty)`).

aaj3f:
Here's the probe as it ran, @bplatz — it's self-contained apart from the three `tests/support` helpers, so it should drop straight into `fluree-db-api/tests/` with the `[[test]]` stanza below. The two tests share one body and differ only in whether the indexed base already carries `ex:items` (as a plain value on the 50 seed rows) — that one boolean is the whole trigger.

```toml
# fluree-db-api/Cargo.toml
[[test]]
name = "it_probe_list_retract_base"
path = "tests/it_probe_list_retract_base.rs"
required-features = ["native"]
```

```rust
// fluree-db-api/tests/it_probe_list_retract_base.rs
#![cfg(feature = "native")]
mod support;
[...]

PR #1692, opened 2026-08-26
docs: make PR↔issue reference direction explicit (Fixes / Follow-up / Partially addresses)

aaj3f:
Small process fix coming out of this week's full-backlog triage: of 94 open issues, 8 turned out to be already-fixed-but-open, and the root cause in every case was the same — the fixing PR referenced the issue without a closing keyword in its body. Meanwhile the *majority* of issue↔PR cross-references in this repo are deferrals (issues filed out of a PR's review, sometimes minutes before merge), which are indistinguishable from fix references without reading each mention's context by hand.

This adds the minimal convention that makes reference direction explicit — `Fixes #N` in the body when a PR resolves an issue, `Follow-up: #N` when it defers one, `Partially addresses #N` with a stated remainder — as a short `CONTRIBUTING.md` plus a comment-only PR template that nudges without imposing structure. It also documents the new triage labels (`P0`–`P3`, `area:*`, `triage:*`) so they don't drift.

Deliberately lean: no process invented beyond what we already do, and CI stays the source of truth for gates. If we'd rather grow a fuller CONTRIBUTING later, this gives it a home.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,7 @@
 <!-- Issue links — keep the reference direction explicit (see CONTRIBUTING.md):
-       Fixes #N               — this PR resolves the issue (keyword must be in this body to auto-close)
+       Fixes #N               — this PR resolves the issue; in THIS body, so the issue links back to the PR
        Follow-up: #N          — work deferred out of this PR (including issues filed during review)
        Partially addresses #N — issue stays open; say below what remains
+                                (write it this way — "partially fixes #N" would close it)
+     Closing keywords only fire when this PR's base is the default branch.
 -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,4 +4,5 @@
        Partially addresses #N — issue stays open; say below what remains
                                 (write it this way — "partially fixes #N" would close it)
      Closing keywords only fire when this PR's base is the default branch.
+     Creating with `gh pr create --body`/`--body-file`? This template is bypassed — add the line yourself.
 -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,5 @@
+<!-- Issue links — keep the reference direction explicit (see CONTRIBUTING.md):
+       Fixes #N               — this PR resolves the issue (keyword must be in this body to auto-close)
+       Follow-up: #N          — work deferred out of this PR (including issues filed during review)
+       Partially addresses #N — issue stays open; say below what remains
+-->
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,19 +1,30 @@
 # Contributing
 
+The full contributor guide — dev setup, code style, testing, PR format — lives in
+[`docs/contributing/README.md`](docs/contributing/README.md). This file covers one thing that
+GitHub surfaces at exactly the right moment: how to link a PR to an issue.
+
 ## Linking issues from PRs
 
 The issue tracker only stays truthful if PR↔issue references carry direction. Three forms, used deliberately:
 
-- **`Fixes #N` / `Closes #N` — in the PR body** — required when the PR resolves the issue. The body specifically, not a comment or a commit message: GitHub only auto-closes the issue on merge when the keyword is in the body (and only when the PR's base is the default branch).
-- **`Follow-up: #N`** — required when the PR *defers* work to an issue, including issues filed during the PR's own review. This marks the reference as a deferral, so it can't be mistaken for a fix.
-- **`Partially addresses #N`** — leaves the issue open by design. Say in the PR body what remains, and rescope the issue after merge so the open remainder is what the issue actually describes.
+- **`Fixes #N` / `Closes #N` — in the PR body** — when the PR resolves the issue. Put it in the body rather than only a comment: a keyword in the body creates the *linked pull request* relationship, so the issue shows what fixed it and the PR shows what it closes. (A closing keyword in a **commit message** also closes the issue when the commit reaches the default branch, but it does not create that linked-PR relationship — so prefer the body, and keep the commit trailer if you like both.) Either way, closing keywords only fire when the PR's base is the default branch.
+- **`Follow-up: #N`** — when the PR *defers* work to an issue, including issues filed during the PR's own review. This marks the reference as a deferral so it can't be mistaken for a fix. Deliberately not a closing keyword.
+- **`Partially addresses #N`** — leaves the issue open by design. Say in the PR body what remains, and rescope the issue after merge so the open remainder is what the issue actually describes. Write it exactly this way — *"partially fixes #N"* still contains a closing keyword and will close the issue on merge.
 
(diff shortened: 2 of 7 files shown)


docs/contributing/README.md
```diff
@@ -131,13 +131,19 @@ More detailed explanation if needed. Wrap at 72 characters.
 Fixes #123
 ```
 
+A `Fixes #123` trailer closes the issue once the commit reaches the default branch, but it does
+not create the *linked pull request* relationship — so put the closing keyword in the **PR body**
+as well (or instead). See [Linking issues from PRs](../../CONTRIBUTING.md#linking-issues-from-prs)
+for the three reference forms and how they behave on stacked PRs.
```
bplatz: **Blocking — this link is dead in the published book.** `docs/` is an mdbook rooted at `docs/` (`book.toml`, `src = "."`), so `../../CONTRIBUTING.md` escapes it. I built it: mdbook emits `href="../../CONTRIBUTING.html"`, no such file exists in the output, and with `site-url = "/db/"` it resolves to `/CONTRIBUTING.html` — outside the site.

It lands on more than the website: `fluree-db-docs/src/embed.rs` embeds `#[folder = "../docs"]`, so the root file isn't in the embedded corpus either — which per `docs_coverage.rs`'s own header backs `fluree docs`, the MCP docs toolset, and the surfaces that "deliberately defer to" it. Nothing catches it (`docs_coverage.rs` gates SUMMARY orphans, not links; `docs.yml` runs bare `mdbook build`).

Minimal fix:

```suggestion
A `Fixes #123` trailer closes the issue once the commit reaches the default branch, but it does
not create the *linked pull request* relationship — so put the closing keyword in the **PR body**
as well (or instead). See [Linking issues from PRs](https://github.com/fluree/db/blob/main/CONTRIBUTING.md#linking-issues-from-prs)
for the three reference forms and how they behave on stacked PRs.
```

Worth knowing there's currently zero precedent for absolute `blob` links in `docs/`, so the structural alternative may be better: put the content at `docs/contributing/issue-linking.md` (in the book, embedded, in `SUMMARY.md`) and make root `CONTRIBUTING.md` a pointer. Root placement is what makes GitHub surface it; the content belongs where the corpus can carry it. Either works — your call.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
+<!-- Issue links — keep the reference direction explicit (see CONTRIBUTING.md):
+       Fixes #N               — this PR resolves the issue; in THIS body, so the issue links back to the PR
+       Follow-up: #N          — work deferred out of this PR (including issues filed during review)
+       Partially addresses #N — issue stays open; say below what remains
+                                (write it this way — "partially fixes #N" would close it)
+     Closing keywords only fire when this PR's base is the default branch.
+-->
```
bplatz, line 8: *suggestion.* This pre-fills only in the web UI, or `gh pr create` **without** `--body`. Anything created with `--body`/`--body-file` bypasses it silently — which is how most PRs here get made, so the nudge half of the design won't reach the workflow it's aimed at. The convention still lands via `CONTRIBUTING.md`; one line closes the gap:

```suggestion
       Partially addresses #N — issue stays open; say below what remains
                                (write it this way — "partially fixes #N" would close it)
     Closing keywords only fire when this PR's base is the default branch.
     Creating with `gh pr create --body`? This template is bypassed — add the line yourself.
-->
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,30 @@
...
+
+- **Retarget to `main` before merging** (what we usually do) — the keyword fires normally. Nothing to do.
+- **Fold a child into its parent** — the child's PR body disappears with it, and any `Fixes #N` it carried is silently lost. Move those keywords into the surviving parent's body, or close the issues by hand with a citation.
+
+Background: a 2026-08 audit of the full open backlog found eight issues that were already fixed but still open, and the cause was the same each time — the fixing PR referenced the issue without a closing keyword, or the reference direction (fix vs deferral) was impossible to tell without reading each mention by hand.
+
+## Gates
+
+CI is the source of truth; see `.github/workflows/ci.yml`. One thing worth knowing locally: `ci.yml` runs **two** fmt/clippy pairs — one over the workspace, and a second with `working-directory: testsuite-sparql`, which is excluded from the workspace. A root `cargo fmt --all` does not reach it, so if you touched `testsuite-sparql/`, format it separately or CI will redden on a file you never opened.
```
bplatz: *suggestion.* This is right and `CLAUDE.md` line 68 is wrong — its "CI runs:" list omits the `testsuite-sparql` pair entirely. Good catch. But shipping the correction here while the stale copy stays put guarantees drift between two sources of truth. Since you already have the correct text, worth fixing line 68 in this PR too, or having one point at the other.

bplatz (changes requested):
Convention is sound and the diagnosis matches what we keep hitting — most of our issue↔PR references are deferrals (this week alone: #1683–#1687, #1689, #1691, all filed out of reviews), and they're indistinguishable from fixes today.

I verified the factual claims before commenting: `ci.yml` really does run two fmt/clippy pairs (lines 29/50 and 112/116), `release.yml` is label-driven with the "Other Changes" fallback, and every `P0`–`P3` / `area:*` / `triage:*` label exists as described. The `"partially fixes #N"` warning is also correct — GitHub matches closing keywords anywhere in the body, not just line-initial — so that phrasing isn't over-cautious.

One blocking item (a dead link in the rendered book) and two suggestions inline.

aaj3f:
Thank you for actually checking the claims rather than taking them on faith, @bplatz — and the dead link is a real blocker I should have caught. All three are addressed in 90700d270.

**The book link.** I took your structural option rather than the absolute `blob` link, for the reason you gave: the content belongs where the corpus can carry it. It now lives at `docs/contributing/issue-linking.md`, in `SUMMARY.md`, so it's in the book, in the embedded corpus behind `fluree docs` and the MCP docs toolset, and in search. Root `CONTRIBUTING.md` keeps a short version — root placement is exactly what makes GitHub surface it on the new-PR page, which is the whole reason it's there — and links to the full page. I built the book to check rather than assume: `contributing/issue-linking.html` renders, `README` emits `href="issue-linking.html"`, and no `../../` href escapes the book anywhere in the output.

**The template bypass.** Correct, and a bit self-indicting: every PR I opened this week was made with `--body-file`, so the nudge would have reached none of them. Added your line verbatim.

**The stale CI list.** This one turned out to be more interesting than a one-line fix. There is no tracked `CLAUDE.md` in this repo at all — it isn't gitignored, it just doesn't exist — so the file you're looking at is your own local copy, and every contributor's drifts independently. That's precisely why yours lost the `testsuite-sparql` pair. So rather than fix one private file I can't see, this PR adds a **tracked** `CLAUDE.md` carrying the conventions that should reach everyone's session — the linking rules, the correct two-pair CI gate description, the label taxonomy, and a few verification habits this codebase keeps rewarding (a skipped check is not a pass; prove a new regression test fails without its fix; the kill-switch oracle isn't infallible — see #1700). `CLAUDE.local.md` is now gitignored as the sanctioned place for personal notes.

One thing to flag before this merges, since it affects you directly: if you have an untracked `CLAUDE.md` at the repo root, git will refuse to check out over it and your next pull will fail until you move it aside. Worth merging your local content into either the tracked file (if it's generally true) or `CLAUDE.local.md` (if it's yours). Happy to fold anything from it in here instead if you'd rather it be shared.

bplatz (approved):
👍

PR #1693, opened 2026-08-26
fix(cli): novelty-aware encoded-binding resolution in the table renderer, and never Debug-print a binding

### LINKED ISSUE #1466: bug: table output prints EncodedSid debug repr for IRI bindings when a join-bound subject's triples are novelty-only
## Observed

`fluree query` with the default `table` format prints the raw Rust debug representation of an internal binding instead of the IRI:

```
| content       | kind                                                       |
+============================================================================+
| probe content | EncodedSid { s_id: 799670408834973892, t: None, op: None } |
```

The engine's result is correct — the **same query** with `--format json`, `--format csv`, or `--format ndjson` decodes (and prefix-compacts) the value properly:

```
"kind": { "type": "uri", "value": "mem:JoinProbe" }   # json
mem:JoinProbe,probe content                            # csv
```

## Repro (narrow shape required)

The bug needs the combination of (a) the SELECTed IRI's triples living **only in novelty** (commit ahead of index) and (b) the subject being bound **through a join** against indexed data. A simple bound-subject lookup over the same novelty triples renders correctly even in table format.

1. On a ledger with an existing indexed dataset (any triples with a string property to join on, e.g. `ex:path`):
   ```
   fluree insert -l <ledger> -e '<urn:m1> a <urn:Probe> ; <urn:ref> "some/indexed/value" ; <urn:content> "probe" .'
   # do NOT run `fluree index`
   ```
2. Query joining through the indexed string into the novelty subject and select its type:
   ```sparql
   SELECT ?kind ?content WHERE {
     ?x <urn:indexed-prop> ?v .          # bound from the base index
     ?m <urn:ref> ?v ; <urn:content> ?content ; a ?kind .   # novelty-only
   }
   ```
3. `--format table` → `EncodedSid { … }`; `--format json|csv|ndjson` → correct IRI. Running `fluree index` also clears it.

## Diagnosis pointer

The table renderer's display path appears to have its own binding-materialization step that, for this operator shape, receives an `EncodedSid` it cannot resolve — plausibly a graph view constructed without the novelty dict routed in (`BinaryGraphView`/`graph_with_novelty` invariant) — and falls through to the `Debug` impl. The serialization formats share a materialization that is novelty-aware, so the fix is likely to route the table renderer through the same path (or at minimum replace the `Debug` fallback with an explicit placeholder so internal structs never reach user output).

Possibly related to the late-materialized `EncodedSid` family in #1443 / #1320, but distinct: results here are semantically correct; only the table presentation is wrong.

Severity: low (interactive default format only, novelty-only window, self-heals on index) — but a raw debug struct in user-facing output warrants a fix or guard.

aaj3f:
Fixes #1466

## Where the reported symptom actually stands

#1466 reported `EncodedSid { s_id: .., t: None, op: None }` turning up in a `--format table` cell. The *query shape* in that report no longer reproduces — I ran it at `47e96b7b6` and the novelty-only subject renders as its IRI every time. For a while I thought that meant the bug had been fixed incidentally somewhere in the engine emit-path work since v4.1.2. It hasn't. The symptom is live on `main`; it just needs a different join shape, and I did reproduce it verbatim.

Indexed ledger, one commit on top that isn't indexed, then a query that joins through a **reference** edge rather than a literal:

```
$ fluree query --format table --sparql \
    'SELECT ?s ?o WHERE { ?s <urn:knows> ?o . ?o a <urn:Target> . }'
+--------+----------------------------------------------------------+
| o      | s                                                        |
+===================================================================+
| urn:x1 | urn:s1                                                   |
|--------+----------------------------------------------------------|
| urn:x1 | EncodedSid { s_id: 3659174697238532, t: None, op: None }  |
+--------+----------------------------------------------------------+
```

The difference is which bindings survive to the formatter. I instrumented `sparql_table_cell` to record the binding variant per output variable across ~20 shapes on an indexed-plus-novelty fixture. Under the reporter's literal join, `?m` and `?kind` arrive as `Binding::Sid` and `?content` as `EncodedLit` — **zero** `EncodedSid`, so that shape would render correctly even with the entire `EncodedSid` arm deleted. Under a reference-edge join against a typed object, `?s` stays late-materialized and reaches the formatter as `EncodedSid`; for a novelty-only subject that's an `s_id` the persisted store cannot resolve. Same for `OPTIONAL { ?s <knows> ?o }`, a two-hop `?a <knows> ?b . ?b <knows> ?c`, and a three-predicate star.

On the `Fixes` line: the issue was rescoped in a comment to *"cli: table renderer should use novelty-aware encoded-binding resolution and never Debug-print bindings"*, which is exactly what this PR does. It also fixes the original symptom under the shape that still produces it, so the closure holds against the issue as it reads now and as it was first reported.

## The gap

`fluree-db-cli/src/output.rs:188` resolved `Binding::EncodedSid` through `gv.store().resolve_subject_iri` — the **base-store** resolver, which consults only the persisted `subject_forward_packs` (`fluree-db-binary-index/src/read/binary_index_store.rs:1302`). A subject first seen after the index snapshot lives in `DictNovelty`, so that lookup **fails**. All six encoded-binding arms then fell back to `format!("{b:?}")` — the internal struct's `Debug` repr — straight into a table cell.

The novelty-aware resolver already exists one level up. `BinaryGraphView::resolve_subject_iri` (`binary_index_store.rs:2436`) tries `dict_novelty` first, then the store, then the novelty subject dict as a fallback. It's what every API formatter reaches through `fluree-db-api/src/format/materialize.rs:52-57`, whose `EncodedSid` arm is explicitly commented *"Novelty-aware: subjects first seen after the index snapshot resolve from DictNovelty"*. The CLI table fast path reads raw batches and bypasses that materialization pass, so it was the one remaining formatter in the tree using the store-only resolver behind a `Debug` fallback.

That's the same class as the newtype-`Display` footgun we've hit before: an internal repr with a plausible-looking string form leaking into something a user reads.

## What changed

**The resolver swap** (`output.rs:227`). `Binding::EncodedSid` now goes through `gv.resolve_subject_iri`, matching the API formatters. Two lines, and it's the whole correctness fix.

**All six `{b:?}` fallbacks become a placeholder that carries the ID** (`output.rs:182`). A failed resolve now renders `(unresolved s_id=3659174697238532)` / `(unresolved p_id=…)` / `(unresolved o_kind=… o_key=… p_id=…)` and emits a `tracing::warn!` naming the ID and whether the failure was a missing graph view or a resolver miss. Even with the resolver fixed, the fallback is the actual guarantee — it's what makes the class impossible rather than merely unlikely.

Two things drove that shape. A bare marker throws away the one datum worth having: every unresolvable binding would collapse to identical text, so two distinct broken rows look alike, and for an arm this cold the day it fires is the day you want the ID. And something like `<unresolved>` in RDF-facing output reads like a relative IRI — a user can reasonably parse that cell as a subject actually named `unresolved`. The parenthesised form can't be mistaken for one. This follows what the rest of the tree already does with the same failure: the API formatters hard-error with the ID in the message (`fluree-db-api/src/format/materialize.rs:63`), and the materializer fabricates `_:unknown_p_{p_id}` alongside a `tracing::warn!` (`fluree-db-query/src/materializer.rs:669-673`). The `EncodedPid` arm had no logging beneath it at all before this — it's an `Option`, and a `None` just vanished. (CLI logs are off by default; `--verbose` surfaces them.)

**The predicate arm deliberately keeps `gv.store().resolve_predicate_iri`**, and `output.rs:232-251` now records why. The reason is *not* that predicates have no novelty layer — they do have one: `EphemeralPredicateMap` (`fluree-db-query/src/binary_scan.rs:2577`), routed by `DictOverlay::resolve_predicate_iri` (`dict_overlay.rs:291`). The reason is reachability. The engine has exactly one construction site for `Binding::EncodedPid` (`binary_scan.rs:1486`), and it's gated on `late_materialize`, which requires `overlay.epoch() == 0` (`binary_scan.rs:1366`) — no novelty overlay at all. Ephemeral p_ids are minted only during overlay translation, i.e. only when an overlay exists, and under an overlay that same scan takes the eager branch (`binary_scan.rs:1488`) and emits an already-resolved `Binding::Sid`. So an ephemeral p_id can't coexist with the path that puts an `EncodedPid` in front of this formatter, and the index root's inline predicate dict is complete for every p_id that can. Every other formatter in the tree resolves predicates the same way (`fluree-db-api/src/format/sparql_xml.rs:302`, `fluree-db-api/src/lib.rs:4580`, `fluree-db-query/src/sort.rs:93`). Only the fallback changed there.

**`commands/branch.rs:1248`** — one more from the audit sweep. `merge preview`'s conflict listing Debug-printed an `Option<Sid>`, so a line read `s=ex:a p=ex:b g=Some("http://ex/g")` while `s` and `p` right beside it rendered plain IRIs. A key with no graph is in the default graph, so it says `<default>` now.

**`commands/create.rs:265`** — same class again. `status["status"].as_str()` is an `Option<&str>`, and the catch-all Debug-printed it, so an unrecognised status rendered as `unexpected remote import status: Some("weird")`. It now prints the value the server actually sent, falling back to the raw JSON when the field is absent or isn't a string (`null`, a number, an object), so the message still names what arrived.

## The audit — what I looked at and what I left alone

I swept the CLI crate for `{:?}` reaching user output. Beyond the eight sites fixed above, everything else is deliberate or harmless, and I'd rather name each than silently leave them:

| Site | Verdict |
|---|---|
| `commands/create.rs:685` | Not stdout — `{phase:?}` goes into a JSON crash-breadcrumb file. Leave. |
| `commands/create.rs:1122,1164` (`EdgePolicy`), `commands/server.rs:1156` (`server_role`), `commands/track.rs:118` (`RemoteEndpoint`), `commands/multi_query.rs:271` (`OutputFormat`), `commands/materialize.rs:149` (`MaterializeVerify`) | Fieldless enums with no `Display` impl. `{:?}` prints a bare variant name — no struct guts, nothing that can leak an ID. Cosmetic casing at worst (`Quick` vs `quick`). Scoped out. |
| `commands/load.rs:266` | `{s:?}` is deliberately quoting user input in an error message so an invisible character is visible. Correct as-is. |
| `commands/materialize.rs:356` | `{:?}` on `CheckOutcome` in the verify parity report — this one *does* print struct guts, but it's a developer-facing diagnostic where the payload *is* the content. I don't think it belongs in this PR, but I'm flagging it in case you disagree. |
| Test-only sites (`remote_client.rs`, `context.rs`, `mcp/ide.rs`, …) | Assertion messages. Fine. |

I also checked the rest of the tree for the same store-vs-view mistake. After this change, `grep` for `store().resolve_subject_iri` outside tests turns up only `fluree-db-query/src/dict_overlay.rs:220,241` — and that one does its **own** watermark routing (`dict_overlay.rs:212-250`), so it's already novelty-aware by construction. `output.rs` was genuinely the only novelty-blind site, and there are none left.

## Tests

Four renderer-level guards in `output.rs`, plus one end-to-end in `tests/integration.rs`:

1. **`novelty_only_encoded_sid_from_a_real_query_renders_its_iri`** — the one that proves the arm is on a live path. It runs the reference-edge join through the real query engine, then, *before* rendering anything, asserts that the engine actually put an `EncodedSid` in the batch **and** that `gv.store().resolve_subject_iri` can't resolve its ID. Those two conditions together are exactly what makes the store-only resolver wrong; asserting them means the guard can't quietly go vacuous if plan selection changes or the fixture stops being novelty-only. It then renders through `format_sparql_table_from_result` and asserts the IRI appears and no `Encoded*` or placeholder does.
2. **`encoded_sid_from_novelty_renders_iri_not_debug_repr`** — the same property at the single-cell level, on a hand-built binding, with the same store-can't-see-this non-vacuity assert.
3. **`unresolvable_encoded_bindings_render_placeholder`** — an unresolvable `EncodedSid` / `EncodedPid` / `EncodedLit`, with and without a graph view, all six paths. Asserts the placeholder shape, that it **carries the ID**, that it contains no `Encoded*`, and that it doesn't start with `<` (i.e. can't be read as an IRI).
4. **`unresolvable_binding_inside_list_renders_placeholder`** — the Cypher list/map arms recurse through the same cell function, so the placeholder and its ID have to hold there too.
5. **`table_output_renders_novelty_only_subject_after_index`** — the repro end to end through the real binary: `insert` → `index` → `insert` → `query --format table`. This used to use the reporter's literal-join shape, which as noted above never reaches the `EncodedSid` arm; it now uses the reference-edge shape that does.

Checked these aren't vacuous by reverting the two-line resolver change and re-running. Both `EncodedSid` tests and the end-to-end test fail, and the rendered table shows the placeholder in place of the subject:

```
+--------+------------------------------------+
| o      | s                                  |
+=============================================+
| urn:x1 | urn:s1                             |
|--------+------------------------------------|
| urn:x1 | (unresolved s_id=3659174697238532) |
+--------+------------------------------------+
```

Reverting the fallback as well as the resolver puts the original `EncodedSid { .. }` back in that cell, which is the repro quoted at the top.

## Gates

- `cargo test -p fluree-db-cli` — 346 passed, 0 failed
- `cargo clippy -p fluree-db-cli --all-targets --no-deps` — clean
- `cargo fmt --all` — clean (run last)

### DIFF
--- a/fluree-db-cli/src/output.rs
+++ b/fluree-db-cli/src/output.rs
@@ -158,10 +158,41 @@ enum SparqlTableFastPath {
     NeedsDisaggregation,
 }
 
-/// Placeholder rendered when an encoded binding cannot be resolved to a
-/// user-facing value (missing graph view, or an ID the dictionaries don't
-/// know). Internal `Debug` representations must never reach user output.
-const UNRESOLVED_CELL: &str = "<unresolved>";
+/// Prefix of the placeholder rendered when an encoded binding cannot be
+/// resolved to a user-facing value. Internal `Debug` representations must never
+/// reach user output.
+const UNRESOLVED_CELL_PREFIX: &str = "(unresolved ";
+
+/// Render — and log — the fallback cell for an encoded binding that did not
+/// resolve, either because there is no graph view to resolve against or because
+/// the dictionaries don't know the ID.
+///
+/// `detail` carries the ID the resolver was handed. Without it every
+/// unresolvable binding collapses to the same text, so two distinct broken rows
+/// look alike and the cell is undiagnosable — which matters more, not less, for
+/// an arm this cold: the day it fires is the day the ID is wanted. This mirrors
+/// what the rest of the tree does with the same failure: the API formatters
+/// hard-error with the ID in the message (`fluree-db-api/src/format/materialize.rs:63`)
+/// and the materializer fabricates `_:unknown_p_{p_id}` alongside a
+/// `tracing::warn!` (`fluree-db-query/src/materializer.rs:669-673`).
+///
+/// The parenthesised shape is deliberate. A bare `<unresolved>` in RDF-facing
+/// output reads like a relative IRI — a user can reasonably parse that cell as a
+/// subject actually named `unresolved`.
+fn unresolved_cell(reason: &str, detail: &str) -> String {
+    tracing::warn!(
+        reason,
+        binding = detail,
+        "table cell: encoded binding did not resolve; rendering placeholder"
+    );
+    format!("{UNRESOLVED_CELL_PREFIX}{detail})")
+}
+
+/// `gv` is `None`: the query result carried no binary graph, so there is
+/// nothing to resolve encoded IDs against.
+const NO_GRAPH_VIEW: &str = "no binary graph view on the query result";
+/// The graph view was present and the resolver still came up empty.
+const RESOLVER_MISS: &str = "graph view could not resolve the ID";
 
 fn strip_question_mark(var_name: &str) -> String {
     var_name.strip_prefix('?').unwrap_or(var_name).to_string()
@@ -188,23 +219,43 @@ fn sparql_table_cell(
 
         Binding::EncodedSid { s_id, .. } => {
             let Some(gv) = gv else {
-                return Ok(UNRESOLVED_CELL.to_string());
+                return Ok(unresolved_cell(NO_GRAPH_VIEW, &format!("s_id={s_id}")));
             };
             // Novelty-aware resolve (parity with the API formatters'
             // materialization): subjects first seen after the index snapshot
             // live in DictNovelty, which the store-level resolver can't see.
             match gv.resolve_subject_iri(*s_id) {
                 Ok(iri) => compact_bnode_strip(compactor.compact_iri_for_display(&iri).ok()),
-                Err(_) => UNRESOLVED_CELL.to_string(),
+                Err(_) => unresolved_cell(RESOLVER_MISS, &format!("s_id={s_id}")),
             }
         }
+        // The store-level predicate resolver is correct here, and deliberate.
+        //
+        // There *is* an ephemeral predicate layer (`EphemeralPredicateMap`,
+        // `fluree-db-query/src/binary_scan.rs:2577`, routed by
+        // `DictOverlay::resolve_predicate_iri` at `dict_overlay.rs:291`), so
+        // "predicates have no novelty layer" would be the wrong reason. The
+        // right one is reachability: the engine has exactly one construction
+        // site for `Binding::EncodedPid` (`binary_scan.rs:1486`), and it is
+        // gated on `late_materialize`, which requires
+        // `overlay.epoch() == 0` (`binary_scan.rs:1366`) — no novelty overlay
+        // at all. Ephemeral p_ids are minted only during overlay translation,
+        // i.e. only when an overlay exists; under an overlay that same scan
+        // takes the eager branch (`binary_scan.rs:1488`) and emits a resolved
+        // `Binding::Sid`. So an ephemeral p_id cannot coexist with the path
+        // that puts an `EncodedPid` in front of this formatter, and the index
+        // root's inline predicate dict is complete for every p_id that can.
+        //
+        // Every other formatter resolves predicates the same way:
+        // `fluree-db-api/src/format/sparql_xml.rs:302`,
+        // `fluree-db-api/src/lib.rs:4580`, `fluree-db-query/src/sort.rs:93`.
         Binding::EncodedPid { p_id } => {
             let Some(gv) = gv else {
-                return Ok(UNRESOLVED_CELL.to_string());
+                return Ok(unresolved_cell(NO_GRAPH_VIEW, &format!("p_id={p_id}")));
             };
             match gv.store().resolve_predicate_iri(*p_id) {
                 Some(iri) => compact_bnode_strip(compactor.compact_iri_for_display(iri).ok()),
-                None => UNRESOLVED_CELL.to_string(),
+                None => unresolved_cell(RESOLVER_MISS, &format!("p_id={p_id}")),
             }
         }
         Binding::EncodedLit {
@@ -216,11 +267,17 @@ fn sparql_table_cell(
             ..
         } => {
             let Some(gv) = gv else {
-                return Ok(UNRESOLVED_CELL.to_string());
+                return Ok(unresolved_cell(
+                    NO_GRAPH_VIEW,
+                    &format!("o_kind={o_kind} o_key={o_key} p_id={p_id}"),
+                ));
             };
             match gv.decode_value_from_kind(*o_kind, *o_key, *p_id, *dt_id, *lang_id) {
                 Ok(v) => flake_value_to_table_cell(&v, compactor),
-                Err(_) => UNRESOLVED_CELL.to_string(),
+                Err(_) => unresolved_cell(
+                    RESOLVER_MISS,
+                    &format!("o_kind={o_kind} o_key={o_key} p_id={p_id}"),
+                ),
             }
         }
 
@@ -521,7 +578,7 @@ mod tests {
     //! placeholder — never a Rust `Debug` representation of an internal struct.
 
     use super::*;
-    use fluree_db_api::LedgerState;
+    use fluree_db_api::{Fluree, LedgerState};
     use fluree_db_api::{FlureeBuilder, ParsedContext, ReindexOptions};
     use fluree_db_binary_index::BinaryGraphView;
     use serde_json::json;
@@ -533,7 +590,14 @@ mod tests {
     /// Mirrors the reported repro's shape: `ex:m9` is committed after the
     /// index snapshot, so its subject ID lives in `DictNovelty` and the
     /// persisted forward packs know nothing about it.
-    async fn indexed_ledger_with_novelty_subject() -> (LedgerState, fluree_db_api::GraphDb) {
+    ///
+    /// `ex:knows` gives both `ex:s1` and `ex:m9` a reference to an indexed
+    /// node. That edge is what lets a query keep the *subject* variable
+    /// late-materialized (see
+    /// `novelty_only_encoded_sid_from_a_real_query_renders_its_iri`); a
+    /// literal-only fixture never produces an `EncodedSid` at all.
+    async fn indexed_ledger_with_novelty_subject() -> (Fluree, LedgerState, fluree_db_api::GraphDb)
+    {
         let fluree = FlureeBuilder::memory().build_memory();
         let ledger = fluree.create_ledger(LEDGER).await.expect("create ledger");
 
@@ -543,8 +607,9 @@ mod tests {
                 &json!({
                     "@context": {"ex": "http://example.org/"},
                     "@graph": [
-                        {"@id": "ex:x1", "ex:indexedProp": "val1"},
-                        {"@id": "ex:s1", "@type": "ex:Probe", "ex:ref": "val1"}
+                        {"@id": "ex:x1", "@type": "ex:Target", "ex:indexedProp": "val1"},
+                        {"@id": "ex:s1", "@type": "ex:Probe", "ex:ref": "val1",
+                         "ex:knows": {"@id": "ex:x1"}}
                     ]
                 }),
             )
@@ -566,7 +631,8 @@ mod tests {
                     "@context": {"ex": "http://example.org/"},
                     "@graph": [
                         {"@id": "ex:x2", "ex:indexedProp": "val2"},
-                        {"@id": "ex:m9", "@type": "ex:Probe", "ex:ref": "val2"}
+                        {"@id": "ex:m9", "@type": "ex:Probe", "ex:ref": "val2",
+                         "ex:knows": {"@id": "ex:x1"}}
                     ]
                 }),
             )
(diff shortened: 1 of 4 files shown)


fluree-db-cli/src/output.rs
```diff
@@ -158,6 +158,11 @@ enum SparqlTableFastPath {
     NeedsDisaggregation,
 }
 
+/// Placeholder rendered when an encoded binding cannot be resolved to a
+/// user-facing value (missing graph view, or an ID the dictionaries don't
+/// know). Internal `Debug` representations must never reach user output.
+const UNRESOLVED_CELL: &str = "<unresolved>";
```
bplatz: The placeholder discards the one datum you'd need if this ever fires.

Every unresolvable binding collapses to identical text, so two distinct broken rows look the same — where the `Debug` repr at least carried the id. And `<unresolved>` in RDF-facing output reads like a relative IRI; a user can reasonably parse that cell as a subject named `unresolved`.

There's precedent for the better shape at `fluree-db-query/src/materializer.rs:666`, which fabricates `_:unknown_p_{p_id}` *and* `tracing::warn!`s. Something like `(unresolved s_id=…)` plus a warn costs nothing and keeps the cell diagnosable. The `EncodedPid` arm is fully silent today — `Option`, no logging anywhere beneath it.

This matters more, not less, given the second comment: if the arm is cold, the day it fires is the day you most want the id.

fluree-db-cli/tests/integration.rs
```diff
@@ -3531,3 +3531,71 @@ fn skolem_namespace_flag_controls_cross_ledger_blank_node_identity() {
...
+/// End-to-end version of the reported repro: an indexed ledger with a subject
+/// whose triples arrived after the index snapshot, joined in through an indexed
+/// literal, rendered as a table.
+///
+/// The bug printed `EncodedSid { s_id: .., t: None, op: None }` in the `kind`
+/// column. The renderer-level guards live in `output.rs`; this pins the whole
+/// pipeline (insert, index, insert, `query --format table`) so a regression
+/// anywhere along it — not just in the resolver — is caught.
+#[test]
+fn table_output_renders_novelty_only_subject_after_index() {
```
bplatz, line 3557: This test doesn't reach the code the PR fixes, and the doc comment above claims it does ("pins the whole pipeline … so a regression anywhere along it — not just in the resolver — is caught").

I reverted the resolver line and ran it: **passes**. Instrumenting `sparql_table_cell` to log its binding variant, this query yields 12 `Sid` + 6 `EncodedLit` cells and **zero `EncodedSid`** — it would pass with the whole `EncodedSid` arm deleted.

I tried six more shapes on the same fixture (`?s ?p ?o`, `a <urn:Probe>`, join-through-literal, `SELECT *`, …) and couldn't produce an `EncodedSid` in any of them; subjects always arrive already materialized as `Sid`. That's probably *why* #1466 stopped reproducing.

Not asking you to drop the test — it's still a net against a future re-route. But either soften the comment to say it's currently non-reddening, or use the shape that does emit `EncodedSid` (you mention confirming the fused path engaged via debug markers — engaging the fused path isn't the same as emitting `EncodedSid` to the formatter; if you have that shape, it belongs here). Worth deciding on `Fixes #1466` too: if nothing live covers it, closing leaves no guard if the symptom returns.

bplatz (approved):
Approving — the resolver swap is right, the audit table is genuinely useful, and the non-vacuity assert inside `encoded_sid_from_novelty_renders_iri_not_debug_repr` is load-bearing (I reverted the resolver line and it reddens with `got "<unresolved>"`).

Two things I'd want looked at before merge, both inline. Neither is a correctness objection to the change itself.

Also worth a one-line wording fix while you're in here: the body says predicates "have no novelty layer at all" as the reason for keeping `store().resolve_predicate_iri`. Keeping it is correct — I inserted a brand-new predicate on a novelty-only subject after an index and it renders fine — but there *is* an ephemeral predicate layer (`EphemeralPredicateMap`, `binary_scan.rs:2577`, routed by `DictOverlay::resolve_predicate_iri` at `dict_overlay.rs:293`). Right decision, wrong reason, and the stated reason is the kind of thing someone builds on later.

aaj3f:
Thanks @bplatz — all three landed, and the third one turned out to be more important than either of us thought.

**Leading with the correction, because it changes the framing of the whole PR.** You're right that the integration test didn't reach the arm — I reverted the resolver and it passed, exactly as you found. But chasing a shape that *does* emit `EncodedSid` turned up that the bug isn't incidentally fixed at all. It reproduces verbatim on `main`:

```
$ fluree query --format table --sparql \
    'SELECT ?s ?o WHERE { ?s <urn:knows> ?o . ?o a <urn:Target> . }'
| urn:x1 | EncodedSid { s_id: 3659174697238532, t: None, op: None } |
```

A **reference-edge join** emits an unresolved `EncodedSid` for the subject; the reporter's *literal*-join shape is the one that stopped reproducing. So "that's probably why #1466 stopped reproducing" is right about the shape and wrong about the bug — I'd written the body around the same wrong assumption before you looked at it, so this is me correcting myself as much as anything. It does mean `Fixes #1466` is sound against the issue as originally reported, not just as rescoped.

I verified by instrumentation rather than debug markers, per your caution — a temporary probe recording the binding variant per output variable across ~20 shapes, plus a per-`EncodedSid` check of whether the base store can resolve it. The reporter's literal join gives `?m` Sid, `?kind` Sid, `?content` EncodedLit and zero `EncodedSid`, matching your measurement exactly. The reference-edge shape gives `?s` as `EncodedSid`, one of them novelty-only. Same for `OPTIONAL`, a two-hop, and a three-predicate star.

The integration test now uses that shape, and there's a new `novelty_only_encoded_sid_from_a_real_query_renders_its_iri` that drives the real engine and asserts — *before* rendering — that the batch contains an `EncodedSid` the base store can't resolve. So it can't go quietly vacuous the way the old one did; if a future re-route stops emitting `EncodedSid` there, it fails and says so. Both red under revert.

**On the placeholder** — agreed, and the `materializer.rs:666` precedent is the right shape. It's now `(unresolved s_id=…)` / `(unresolved p_id=…)` / `(unresolved o_kind=… o_key=… p_id=…)` through one helper that also warns and distinguishes "no graph view" from "resolver miss". Its value showed up immediately: the revert-check failure prints `(unresolved s_id=3659174697238532)` instead of an anonymous marker. Your point that the cold arm is exactly the one that needs the id on the day it fires is the argument that convinced me.

**On the predicate wording** — you're right, and thank you for catching that the decision was right for the wrong reason, which is the version that gets built on. The body claimed predicates have no novelty layer; `EphemeralPredicateMap` at `binary_scan.rs:2577` routed through `DictOverlay::resolve_predicate_iri` plainly is one. The real reason is reachability: `Binding::EncodedPid` has exactly one engine construction site (`binary_scan.rs:1486` — everything else in the tree is a match pattern), gated on `late_materialize`, which requires `overlay.epoch() == 0`. Ephemeral p_ids are minted only during overlay translation, and under an overlay that scan takes the eager branch at `:1488` and emits a resolved `Binding::Sid`. That's now a comment at `output.rs:232-251` naming both, rather than a claim in a PR body nobody reads later.

Also folded in `create.rs:265`, same `Option<&str>` class as the `branch.rs` one, and added `materialize.rs:149` to the audit table's fieldless-enum row.

Since the framing section changed materially, the body is worth a re-skim — it now says the symptom is live rather than incidentally fixed.

PR #1698, opened 2026-08-26
fix(query): novelty-only-predicate retractions were invisible to the crawl read path

### LINKED ISSUE #1683: Retractions of novelty-only predicates are invisible to the crawl read path (deleted data reappears)

### LINKED ISSUE #1684: Add a transact_filtered_delete bench — filtered-DELETE staging has no coverage
Raised in review of #1675.

## Problem

Nothing benches the filtered-DELETE staging path. `fluree-db-api/benches/transact_commit.rs` does not exercise WHERE/DELETE at all, which is how a `retractions × novelty` shape in `hydrate_list_index_meta_for_retractions` sat unnoticed until a 40k-subject delete on a novelty-heavy ledger stopped finishing altogether. #1675 fixed that shape; nothing currently guards the win.

## Ask

A `transact_filtered_delete` bench over a novelty-heavy ledger:

- Enough retracted `(subject, predicate)` groups and enough novelty that a per-group overlay walk is visibly quadratic — the regression this is meant to catch is superlinear, so the fixture has to be large enough for the two curves to separate.
- Cover both branches of the skip guard (`snapshot.has_list_meta == Some(false)` with clean novelty vs. a `@list` present), since they have very different cost profiles: the reference numbers from #1675 were 0.9 s with no lists and 3.1 s with a `@list` present on the same 40.5k-retraction delete.
- Add the matching `regression-budget.json` entry so `workspace_reconcile` stays happy.

## Related

`validate_staged_nodes` in the SHACL path has the same per-subject-range shape and is already named as a follow-up on #1675; worth deciding whether one bench covers both or they want separate fixtures.

## Origin

Review of #1675.

### LINKED ISSUE #1685: Record has_list_meta on bulk-import roots so the hydration skip applies to large ledgers
Raised in review of #1675.

## Problem

#1675 added `IndexRoot.has_list_meta: Option<bool>`; when both it and `Novelty.has_list_meta` say "no list rows", filtered-DELETE staging skips list-meta hydration entirely.

Bulk import leaves the flag `None` (untracked) — see the comment at `fluree-db-api/src/import.rs:6837`. `None` is correct-but-conservative: the write path keeps hydrating until the first full rebuild records the flag exactly. The result is that the skip does not apply to bulk-imported ledgers, which are precisely the ones large enough for the cost to matter.

Correctness is unaffected either way. This is purely about the optimization reaching the ledgers that need it.

## Why it is not a one-liner

`write_record` (`fluree-db-transact/src/import_sink.rs:467`) has the list index in hand, but it runs per chunk. A bool set there would have to ride both `SpoolResult` and `BufferedSpoolResult` and then be OR-folded across chunks at root assembly.

## Suggested approach

Put the sticky bit on the shared `SpoolConfig`, alongside `numbig_pool` / `vector_pool` / `ns_alloc`, rather than threading it through the two result types:

- `saw_list_meta: Arc<AtomicBool>` on `SpoolConfig`.
- A plain `bool` on `SpoolContext`, OR'd in `write_record` only on the `i != LIST_INDEX_NONE` branch — no per-record atomic traffic.
- One relaxed store at `SpoolContext::finish` / `finish_buffered`.
- Read it at root assembly in `import.rs` and set `has_list_meta: Some(saw_list_meta)` instead of `None`.

This mirrors how `has_annotations` is derived at the same site, which reads a merged global structure (the predicate dict) rather than per-chunk state.

## Testing note

The bulk-import path gets no coverage from #1675's tests. Whatever lands here needs its own: an import with no lists asserting the root records `Some(false)`, and an import with a `@list` asserting `Some(true)` — the second matters most, since a wrongly-`false` root would skip hydration and silently drop list retractions.

## Origin

Review of #1675.

### LINKED ISSUE #1703: `resolve_latest_ops_keep_asserts` drops retractions when two flakes share a value but differ in flake metadata
Found during review of #1698 and confirmed by calling the helper directly. This is pre-existing and not introduced by that PR — but it is the same symptom #1698 is fixing (a retraction that silently has no effect), reached through a different door, so it's worth its own issue rather than living in a review thread.

## What happens

`resolve_latest_ops_keep_asserts` (`fluree-db-query/src/binary_range.rs:1580`) is the lifecycle-resolution helper that turns an overlay *log* into current state. It groups flakes by walking adjacent runs and closing a group the moment `same_fact_identity` (`:860`) returns false — and that identity includes `m`, the flake metadata.

But every comparator orders `… t, op, m` — `t` **before** `m` (`fluree-db-core/src/comparator.rs:161-207`; `cmp_spot` is `s, p, object, t, op, meta`). So when two flakes share `(s, p, o, dt)` and differ only in metadata, the later-`t` retraction is separated from its own assert by the *sibling's* assert. The run closes early, the retraction never meets the flake it cancels, and the retracted assert is emitted as live.

## Confirmed

```
@list ["a","b","a"], DELETE the position-0 entry
  -> [("a", i:0), ("a", i:2), ("b", i:1)]     // retraction had no effect

ex:label "hello"@en, "hello"@fr, DELETE "hello"@en
  -> [Some("en"), Some("fr")]                  // retraction had no effect
```

Deleting the *position-2* entry instead passes, which shows the shape of it: it only fails when the retracted entry's metadata sorts below a surviving sibling's. On any list with a repeated value that's roughly a coin flip, not a rare corner.

## Why it matters

The helper is reached from four call sites (`:796`, `:1432`, `:1487`, and — after #1698 — `overlay_only_flakes`), so this affects overlay-only reads generally, not one lane. The user-visible shape is the one we keep coming back to: a delete commits successfully and the data is still there.

This is also a recurrence of a bug class we've hit before. RDF fact identity here is `(s, p, o, dt, m)` with `m = FlakeMeta{lang, i}`, and #1273 was the same story in `range::remove_stale_flakes` — a key that mishandles `m` collapses langString variants sharing a lexical form and `@list` members repeating a value at different indices. Worth noting that the existing regression coverage uses single-occurrence values only, so it passes with this bug present.

## Suggested direction

Group on `(s, p, o, dt)` and resolve per-`m` *inside* the group, rather than relying on run adjacency:

```rust
// inside the group, keep the newest op per distinct `m`
let mut by_meta: HashMap<&Option<FlakeMeta>, usize> = HashMap::new();
// ... then emit the winners that are asserts
```

Reordering the comparators to put `m` ahead of `t` would also work, but that changes index ordering semantics repo-wide and I don't think it's the right lever here.

Whatever lands should be pinned by both shapes above — a repeated `@list` value *and* one lexical form under two language tags — since either alone leaves half the mechanism uncovered.

### LINKED ISSUE #1704: `batched_refs_overlay_only` walks the overlay assert-only, so a revoked `rdf:type` still grants until the first index build
Found during review of #1698. That PR fixes the identical defect in `overlay_only_flakes`; this is the third overlay-only path in the same file and it was not swept.

## What happens

`batched_refs_overlay_only` (`fluree-db-query/src/binary_range.rs:1132`) opens its walk callback with:

```rust
&mut |flake| {
    if !flake.op {
        return;
    }
```

That is `overlay_only_flakes`' pre-#1698 code verbatim, and there is no lifecycle resolution anywhere in the function (`:1110-1157`). The overlay is a log, not a current-state view, so an `rdf:type` asserted and then retracted within the same novelty window comes back asserted.

Its two sibling paths in the same function both resolve correctly — `resolve_overlay_ops` at `:1005` and `apply_raw_overlay_deltas_to_batched_refs` at `:1060` — so this arm is the odd one out in exactly the way `overlay_only_flakes` was.

## Why it matters

It's the no-PSOT-branch fallback of `binary_lookup_subject_predicate_refs_batched_v3` (`:870`, reached at `:953`), and that function's own doc comment at `:868` says it serves policy `rdf:type` lookups and stats refresh.

So on a ledger with no PSOT branch yet, a policy that grants by class membership **keeps granting on a type that was revoked**, until the first index build. That's the fail-open direction.

The same lookup also feeds class attribution for `fluree info`: `LedgerInfoStatsLookup::lookup_subject_classes` (`fluree-db-api/src/ledger_info.rs:466-481`) → `fluree_db_policy::lookup_subject_classes` → `class_lookup.rs:42` → `lookup_subject_predicate_refs_batched` → `binary_range.rs:358` → `binary_lookup_subject_predicate_refs_batched_v3` → `:953` → here. So counts attributed to a class can include a subject whose `rdf:type` was retracted.

## Scope, honestly

The trigger is narrower than #1683's: "no PSOT branch" rather than "bound component has no persisted id", so this is genesis and pre-first-index ledgers rather than any indexed one. But that is precisely the window the overlay-only paths exist to serve, and the policy consequence is the kind that shouldn't wait on an index build.

## Suggested direction

`apply_raw_overlay_deltas_to_batched_refs`, directly above it at `:1060`, is already the correct rule — reuse it rather than writing a third variant. A guard test should assert that a same-window assert+retract of `rdf:type` does **not** report the class, on a ledger with no PSOT branch.

aaj3f:
Fixes the P0 from #1683, folds in the two deferrals that were sequenced behind it, and closes two pre-existing defects of the same class that surfaced while fixing it.

## The bug is bigger and simpler than the issue title

#1683 reported this as a `@list` problem: a filtered DELETE of one list entry commits with `receipt.flake_count = 1` and the entry stays readable. The narrowing in the issue was right that staging is not at fault — `hydrate_list_index_meta_for_retractions` hands the commit a retraction with the correct `m = { i: 1 }`, and I confirmed that again here: novelty at `t=3` holds `s=ex:nov1 p=ex:items o="b" i=Some(1) op=false`, exactly as it should.

The write path was fine. **The read path was dropping the retraction on the floor.**

`fluree-db-query/src/binary_range.rs` — `binary_range_eq_v3` bails to `overlay_only_flakes` whenever a bound component of the pattern has no persisted id (a novelty-only subject, predicate, or object) or the requested order has no branch manifest. `overlay_only_flakes` served that by filtering the walk to `if !flake.op { return; }` — assert-only, no lifecycle resolution at all. But the overlay is a **log**, not a current-state view: the assert at `t=2` and its retraction at `t=3` both sit in novelty. Filtering to asserts resurrects every fact whose assert *and* retract live there.

Consequence: any workload that writes a fact under a brand-new predicate and then deletes it before the next index build is told the flake was retracted and can still read the value. Silent, and shaped like data loss in the opposite direction.

Two things about the shape are worth stating plainly, because both change how the issue reads.

**Nothing about it is list-specific.** I added a plain-valued leg to the repro on a second novelty-only predicate — `ex:tags: ["p", "q"]`, delete `"p"` — and got `["p", "q"]` back at the pin. `@list` is just how the bug was *found*: a brand-new predicate carrying a collection is a very natural way to land a fact wholly in novelty.

**Reload does not heal it.** That was the open remediation question from the sweep, and the answer is no. I ran the reload leg against the unfixed code specifically to answer it: a fresh handle replays the same commit chain into novelty and reads it the same wrong way, so an affected ledger stays wrong across a restart. It heals on the next *index build* (the fact leaves novelty and the cursor path resolves it correctly), not on reload. The regression pin has a reload leg so that stays honest.

That also explains the sweep's puzzle about the CLI probes not reproducing: the BGP path (`select ["?s","?o"]`) resolves the overlay correctly and always did — it takes the cursor path. Only the subject-crawl (`select {"?s": ["ex:items"]}`) goes through `fetch_subject_predicate_pair`, binds both `s` and `p`, fails to resolve the novelty-only `p`, and lands in `overlay_only_flakes`. The maintainer probe in the thread used the crawl; the CLI probes used a BGP. Same ledger, two answers.

## The fix

Keep both ops through the walk, run the matching set through `resolve_latest_ops_keep_asserts` (newest op per fact identity wins, retracted keys drop out), then apply object bounds / offset / limit to the *resolved* state. That is the same rule the cursor path already gets from `resolve_overlay_ops`, and the same thing the bounded twin `overlay_only_flakes_bounded` twenty lines up was already doing — this function was simply the odd one out.

The early exit at `limit` had to go with it: the retraction that cancels an already-collected assert can arrive after the limit is reached. Worth noting it was never buying much — `for_each_overlay_flake` walks the graph's whole overlay either way, so the exit only saved clones of matching flakes. (There's a measured argument about restoring it further down.)

Regression pin lives with its siblings in `fluree-db-api/tests/it_filtered_delete_list_meta.rs` (`novelty_only_predicate_retraction_takes_effect`): novelty-only predicate over an indexed base, live handle across the background-index swap, list leg + plain leg + reload leg, plus a BGP assertion so the two read paths stay in agreement. It fails at the merge base with the reported `["a","b","c"]` vs `["a","c"]`.

## Two more instances of the same defect — #1703 and #1704

Both were already broken before this branch, and the `overlay_only_flakes` fix does not, on its own, close either. Both live in code this branch already holds, and both are the same "the overlay is a log" defect #1683 is about, so they land here rather than as deferred follow-ups. Filed as #1703 and #1704 so the history is honest about them being pre-existing.

### #1703 — the helper I delegate to had the same bug for values that repeat under different metadata

`resolve_latest_ops_keep_asserts` cut resolution groups on full fact identity `(s, p, o, dt, m)` by run adjacency. But every comparator orders `… t, op, m`, with `t` *before* `m` (`fluree-db-core/src/comparator.rs:161-207`), so siblings sharing `(s, p, o, dt)` interleave by `t`: a sibling's assert lands between a fact's own assert and its retraction, the run closes early, and the retracted assert is emitted live. `["a", "b", "a"]` deleting position 0 returned both `"a"` entries; `"hello"@en` + `"hello"@fr` deleting the `@en` returned both. Deleting the *last* position passes, so it only fails when the retracted entry's metadata sorts below a surviving sibling's — about a coin flip on any list that repeats a value.

Runs are now cut on the four fields all four comparators *do* order before `t`, with the winner picked per distinct `m` inside the run. I verified non-vacuity by putting `m` back into the run key and re-running: exactly the two reported outputs come back, and the last-position case stays green — so both directions are pinned.

The per-`m` grouping inside the run needs its own comparator, and this is the part that is easiest to "simplify" wrongly. `FlakeMeta::cmp` (`fluree-db-core/src/flake.rs:110-129`) compares **only** `i` when both sides carry a list index — `lang` is never consulted — while `PartialEq` is derived over both fields. So `{lang: en, i: 0}` and `{lang: fr, i: 0}` compare `Equal` without being equal: `Ord` is not consistent with `Eq` for that type. Sorting by that relation and then cutting groups on `m == m` is sorting by one relation and grouping by a stricter one — structurally the same mistake as #1703 itself, one level down. One list position under two language tags lets the sibling's assert separate a fact from its own retraction, and the retracted value survives. Grouping therefore sorts with a local `meta_group_cmp` that compares `i` then `lang`, so `Equal` means equal in fact and the sort agrees with the `==` the groups are cut on.

**Two things here are load-bearing and shouldn't be simplified away.** Runs are not cut by moving `m` ahead of `t` in the comparators, which would change index ordering semantics repo-wide. And `FlakeMeta::Ord` itself is left alone deliberately — every comparator's `cmp_meta` tiebreak reads it, so repairing the type moves index ordering semantics repo-wide too. That's #1711's job, not something to smuggle in here.

Three tests cover the combination specifically, because the residual lives there and not in either half: tests covering list positions and language tags separately all pass. Both tag orders of the combined shape, plus one that asserts the invariant directly — `meta_group_cmp` returns `Equal` exactly when the metas are equal, across the eight interesting `(lang, i)` combinations, plus the `en@0`/`fr@0` pair that is `Ord`-equal and `Eq`-unequal. Verified non-vacuous by putting `FlakeMeta::cmp` back as the sort key: both new cases fail with the reported output while the other eleven stay green.

**One thing I found trying to write the end-to-end leg, which is worth carrying into #1711.** The combined shape can't currently be built through the write path at all. Asserting a second language tag at an occupied list position commits — the receipt reports one flake at its own `t` — but it never reaches the live view: novelty keys the fact by that same inconsistent ordering and folds it onto the first tag. So the write side collapses the shape before the read side can be asked about it. That means the `Ord`/`Eq` disagreement is silently dropping a legitimately distinct write, not just mis-resolving a read, which I think widens #1711's blast radius beyond how it's currently scoped. I left a note in `it_filtered_delete_list_meta.rs` naming that test as the one to grow an end-to-end leg once the type is fixed, rather than contriving a fixture that can't exist today.

### #1704 — a third assert-only overlay path

`batched_refs_overlay_only` was still filtering to asserts with no resolution, the odd one out among the three paths of `binary_lookup_subject_predicate_refs_batched_v3` — which serves policy `rdf:type` lookups. On a graph with no PSOT branch, a class asserted and revoked inside one novelty window kept granting class-based policy access until the first index build. It now hands the merged set to `apply_raw_overlay_deltas_to_batched_refs`, the same rule its two sibling paths already use, rather than a third variant of it. The guard test drives the real provider entry point on a named graph that lives only in novelty; I confirmed it reaches the intended path, and that restoring the assert-only filter makes it fail with `["Admin", "Person"]`.

Beyond those, the test additions here cover the duplicate-valued list and language-tag shapes end to end, and an offset/limit pin — the fix moved when paging applies relative to resolution, and nothing held that ordering down.

## #1685 — `has_list_meta` on bulk-import roots

`import.rs:6837` left the flag `None`, so the `stage.rs:1623` skip (`snapshot.has_list_meta == Some(false) && !novelty.has_list_meta`) could never fire for bulk-imported ledgers — which are the large ones, i.e. exactly where the O(novelty) walk costs something.

Implemented as the issue proposed, which does match the `numbig_pool` / `vector_pool` / `ns_alloc` precedent: a sticky `Arc<AtomicBool>` on `SpoolConfig`, a plain `bool` on `SpoolContext` OR'd in `write_record`, one relaxed store per chunk at `finish` / `finish_buffered`, read once at root assembly after the workers join. No per-record atomic traffic.

One deliberate detail: the OR gates on the *mapped* `i != LIST_INDEX_NONE`, not on `list_index.is_some()`. An out-of-range index coerces to the sentinel a few lines above and is not a list row in the index, so gating on the option would let a `Some(true)` root claim a list that the index doesn't actually carry. Cheap insurance in the safe direction, but the safe direction here is *not* the obvious one — a wrongly-`false` root is the dangerous case, which is why `import_with_list_records_has_list_meta_true_and_still_retracts` asserts the flag *and* then runs the filtered DELETE it protects.

The `import_sink.rs:253` comment is also corrected: it claimed every `SpoolContext` reaches a finish path, which isn't true (there's a `?` on the parse and on `into_parts()` between construction and finish). The flag is still safe, but the invariant is "a context that doesn't finish means no root", not "every context finishes" — the old wording would let a future change that tolerates per-chunk failures look correct. And `same_fact_key` now carries the note about `g` being absent because every caller scopes to one `g_id`.

## #1684 — `transact_filtered_delete` bench

Two scenarios over a bulk-imported, novelty-heavy ledger, covering both sides of the `stage.rs:1623` gate with the same deleted predicate so the delta is the hydration path and nothing else. The `no_lists` branch is only reachable this cheaply because of the #1685 work above — before it, that branch needed a full rebuild inside the fixture.

Each scenario asserts its gate state during setup rather than trusting the fixture; a fixture that drifts off its branch fails the bench instead of quietly reporting a plausible number for the wrong path.

On the honesty front: at Small both scenarios read ~39 ms, because the WHERE and the commit dominate — the curves only separate from Medium up (69 ms vs 127 ms at 20k groups on my box). I've said that in the module doc so nobody reads a flat Tiny/Small pair as "the gate is free". Either scale still catches what the bench exists for, since a regression back to `retractions × novelty` turns tens of milliseconds into seconds.

Wired into `fluree-db-api/Cargo.toml` and `regression-budget.json`; `workspace_reconcile` passes.

## The read-side bench, and why the `limit` early exit stays gone

`transact_filtered_delete` guards the write side; nothing covered the read side this PR actually changes. `query_overlay_only_range` does, with two scenarios that separate the candidate cost drivers. Quick profile on my box: `narrow_subject` (one matched fact) is 13.6 µs at 2k novelty subjects and 13.8 µs at 10k — flat across a 5× increase in overlay size — while `wide_subject` goes 31.8 µs at 200 matched values to 95.9 µs at 1k, about 80 ns per matched row. So the cost is bounded by the matched fact set, not by novelty.

Restoring the early exit by streaming the resolution is the obvious alternative — `for_each_overlay_flake` is contractually comparator-ordered, so it's available. I don't think it's the right trade here.

The caller trace says the exposure is small. Every caller that sets a limit and can reach this function: `policy_builder.rs:603` and `cypher_write.rs:900` are subject-bound with `flake_limit(1)`; `ledger_view.rs:331` is predicate+object-bound with `flake_limit(16)`; `time_resolve.rs:342` and `ledger_view.rs:235` go through `range_bounded_with_overlay`, i.e. the bounded twin, which already collected everything before this PR. The one that is *not* subject-bound is `probe_timestamp_axis` (`time_resolve.rs:154`), which issues `RangeMatch::predicate(..)` — predicate only, no subject, no object — with `flake_limit(1)`, and adds `object_bounds` when resolving `after: Some`. Its matched set is one flake per commit in the txn-meta graph, so it scales with commits-in-novelty rather than with a bound subject's fact count, and it's doubly exposed because those object bounds now prune after resolution instead of inside the walk. It only takes this path when txn-meta has no POST branch, it reads a system graph, and at ~80 ns per matched row the absolute cost is small — but it is the shape neither bench scenario models, so rather than a third scenario, `query_overlay_only_range`'s doc names the site, its exact bind pattern, the `object_bounds` interaction, and the fact that it's where to measure first if this path ever needs the bound back.

Against that small cost, streaming would trade the helper's internal `sort` — which today self-heals if any overlay provider ever violates the ordering contract — for a hard dependency on that contract, in the fallback path that exists precisely to be defensive. A few hundred nanoseconds of clone on bound patterns isn't worth that. The bench is in the tree, so the day someone does add a limit-setting caller on an unbound pattern, we'll see it. Happy to be talked back into the restructure if the balance reads differently to someone else.

## Gates

- `cargo test -p fluree-db-api` — 3250 passed, 0 failed
- `cargo test -p fluree-db-query` — 1494 passed (1434 lib, incl. 13 resolver unit tests)
- `cargo test -p fluree-db-transact` / `-p fluree-db-policy` / `-p fluree-db-sparql` / `-p fluree-db-cypher` — all green
- `cargo test -p fluree-bench-support --test workspace_reconcile` — green
- `cargo clippy -p fluree-db-query -p fluree-db-transact -p fluree-db-api --all-targets --no-deps` — clean
- `cargo fmt --all` — clean (run last)
- `cargo bench -p fluree-db-api --bench transact_filtered_delete --bench query_overlay_only_range -- --test` — all four scenarios green

Two things I did **not** do, flagging rather than hiding them. The W3C SPARQL suite didn't run — its `rdf-tests` submodule isn't populated in this worktree, and that suite runs on memory ledgers, which don't reach `binary_range_eq_v3` at all, so I don't think it exercises the changed line. And `--all-features` won't build in my environment (`cxx` build script fails), which is pre-existing and unrelated; everything above is default features.

I'd also note for the record that #1683's title and labels understate it now — it isn't an `@list` bug, it's "novelty-only-predicate retractions are invisible to the crawl read path". Happy to retitle the issue if that's useful, or leave it and let this PR body carry the correction.

Fixes #1683
Fixes #1685
Closes #1684
Fixes #1703
Fixes #1704

Follow-up: #1711

### DIFF
--- a/fluree-db-query/src/binary_range.rs
+++ b/fluree-db-query/src/binary_range.rs
@@ -13,9 +13,9 @@ use fluree_db_binary_index::{
 use fluree_db_core::dict_novelty::DictNovelty;
 use fluree_db_core::subject_id::SubjectId;
 use fluree_db_core::{
-    flake_matches_range_eq, range_provider::RangeQuery, Flake, FlakeMeta, FlakeValue, GraphId,
-    IndexType, OType, OverlayProvider, RangeMatch, RangeOptions, RangeProvider, RangeTest,
-    RuntimeSmallDicts, Sid,
+    flake_matches_range_eq, range_provider::RangeQuery, Flake, FlakeValue, GraphId, IndexType,
+    OType, OverlayProvider, RangeMatch, RangeOptions, RangeProvider, RangeTest, RuntimeSmallDicts,
+    Sid,
 };
 
 use crate::binary_scan::{encode_bound_object_prefilter, index_type_to_sort_order};
@@ -793,7 +793,7 @@ fn binary_range_eq_v3(
     // Correctness fallback: merge untranslated overlay flakes (including retracts),
     // resolve per-fact lifecycles (latest-op-wins), then apply RangeOptions.
     flakes.extend(untranslated);
-    let mut resolved = resolve_latest_ops_keep_asserts(flakes, index);
+    let mut resolved = fluree_db_core::range::resolve_current_flakes(flakes, index);
     resolved.retain(|f| flake_matches_range_eq(f, match_val));
 
     if let Some(bounds) = &opts.object_bounds {
@@ -818,151 +818,6 @@ fn binary_range_eq_v3(
 #[inline]
 fn resolve_sid(s_id: u64, view: &BinaryGraphView) -> std::io::Result<Sid> {
     view.resolve_subject_sid(s_id)
-}
-
-/// Resolve fact lifecycles (latest op wins) and drop retracts.
-///
-/// Used as a correctness fallback when some overlay flakes cannot be translated
-/// into V3 `OverlayOp`s (e.g., missing dict novelty), and by the raw-overlay
-/// paths that serve a pattern the persisted index cannot contribute to. The
-/// input should include both cursor output flakes (asserts) and raw overlay
-/// flakes (asserts/retracts).
-///
-/// ## Why the grouping is two-level
-///
-/// A fact's identity is `(s, p, o, dt, m)`: two `@list` entries holding the
-/// same value at different positions, and one lexical form under two language
-/// tags, are DISTINCT facts (the #1273 class). But every comparator orders
-/// `… t, op, m` — `t` BEFORE `m` (`fluree-db-core/src/comparator.rs`). So
-/// flakes that share `(s, p, o, dt)` and differ only in `m` interleave by `t`,
-/// which puts a sibling's assert between a fact's own assert and its
-/// retraction. Cutting runs on full identity therefore closes a group early
-/// and emits the retracted assert as live — the exact resurrection this
-/// helper exists to prevent, on `["a", "b", "a"]` or `"x"@en` + `"x"@fr`.
-///
-/// So runs are cut on [`same_fact_key`] — the four fields all four comparators
-/// DO order before `t`, making those runs contiguous — and the winner is
-/// picked per distinct `m` inside the run.
-fn resolve_latest_ops_keep_asserts(mut flakes: Vec<Flake>, index: IndexType) -> Vec<Flake> {
-    let cmp = index.comparator();
-    flakes.sort_by(cmp);
-
-    if flakes.len() < 2 {
-        return flakes.into_iter().filter(|f| f.op).collect();
-    }
-
-    let mut out: Vec<Flake> = Vec::with_capacity(flakes.len());
-    // Scratch index buffer for the multi-flake run path, reused across runs.
-    let mut idxs: Vec<usize> = Vec::new();
-    let mut start = 0usize;
-    while start < flakes.len() {
-        let mut end = start + 1;
-        while end < flakes.len() && same_fact_key(&flakes[start], &flakes[end]) {
-            end += 1;
-        }
-
-        // Overwhelmingly the common run: one flake, so no grouping at all.
-        if end - start == 1 {
-            if flakes[start].op {
-                out.push(flakes[start].clone());
-            }
-            start = end;
-            continue;
-        }
-
-        // Group the run by `m`. Sorting rather than scanning a winners list
-        // keeps this O(k log k): a list that repeats one value across k
-        // positions puts all k in ONE run with k distinct `m`, and a linear
-        // "have I seen this `m`" probe would make that quadratic — the shape
-        // this whole area of the code exists to have stopped doing.
-        //
-        // The sort keys on `m` ALONE via `meta_group_cmp`, NOT `FlakeMeta`'s
-        // own `Ord` — that one ignores `lang` whenever both sides carry a
-        // list index, so it calls `{en, i: 0}` and `{fr, i: 0}` equal while
-        // `==` calls them distinct. Grouping on `==` after sorting by it
-        // would sort by one relation and cut by a stricter one: the same
-        // mismatch as #1703, one level down. `meta_group_cmp` agrees with
-        // `==` by construction (pinned by
-        // `group_order_agrees_with_metadata_equality`).
-        //
-        // The sort is stable, so the comparator's `t, op` order survives
-        // inside each group and the winner scan below sees candidates in
-        // ascending `t`.
-        idxs.clear();
-        idxs.extend(start..end);
-        idxs.sort_by(|&a, &b| meta_group_cmp(&flakes[a].m, &flakes[b].m));
-
-        let mut g = 0usize;
-        while g < idxs.len() {
-            let mut best = idxs[g];
-            let mut h = g + 1;
-            while h < idxs.len() && flakes[idxs[h]].m == flakes[best].m {
-                let cand = &flakes[idxs[h]];
-                let cur = &flakes[best];
-                // Newest op wins; at equal `t` a retraction beats an assert,
-                // matching `resolve_current_flakes`.
-                if cand.t > cur.t || (cand.t == cur.t && !cand.op && cur.op) {
-                    best = idxs[h];
-                }
-                h += 1;
-            }
-            if flakes[best].op {
-                out.push(flakes[best].clone());
-            }
-            g = h;
-        }
-
-        start = end;
-    }
-
-    // Survivors of a multi-flake run are emitted in `m` order, which is not
-    // the comparator's order for the run (it orders `t` first). Callers take
-    // this as a range result and expect index order, which the
-    // single-winner-per-group predecessor gave them for free. Cheap to
-    // restore: `out` is a subsequence of an already-sorted vec whose only
-    // inversions are inside those runs.
-    out.sort_by(cmp);
-    out
-}
-
-/// Strict total order on flake metadata, for grouping inside a fact run.
-///
-/// [`FlakeMeta`](fluree_db_core::FlakeMeta)'s own `Ord` compares ONLY `i`
-/// when both sides carry a list index — `lang` is never consulted — while its
-/// `PartialEq` is derived over both fields. So `{lang: en, i: 0}` and
-/// `{lang: fr, i: 0}` compare `Equal` without being equal: `Ord` is not
-/// consistent with `Eq` for that type. Sorting by it and then cutting groups
-/// on `==` sorts by one relation and groups by a stricter one, which lets a
-/// sibling's assert separate a fact from its own retraction — the #1703
-/// failure, one level down.
-///
-/// This order is consistent with `==` by construction: it compares `i` and
-/// then `lang`, so `Equal` here means equal in fact.
-///
-/// Deliberately local. Repairing `FlakeMeta::Ord` itself moves index ordering
-/// semantics repo-wide (every comparator's `cmp_meta` tiebreak reads it), so
-/// that belongs in #1711 rather than in a fix for the range path.
-#[inline]
-fn meta_group_cmp(a: &Option<FlakeMeta>, b: &Option<FlakeMeta>) -> std::cmp::Ordering {
-    match (a, b) {
-        (None, None) => std::cmp::Ordering::Equal,
-        (None, Some(_)) => std::cmp::Ordering::Less,
-        (Some(_), None) => std::cmp::Ordering::Greater,
-        (Some(x), Some(y)) => x.i.cmp(&y.i).then_with(|| x.lang.cmp(&y.lang)),
-    }
-}
-
-/// Run key for [`resolve_latest_ops_keep_asserts`]: the four fields every
-/// comparator orders BEFORE `t`, so flakes sharing it land contiguously after
-/// the sort. Deliberately NOT full fact identity — `m` is resolved per-fact
-/// inside the run, because the comparators interleave differing `m` by `t`.
-///
-/// `g` is absent because every caller walks one `g_id` at a time, so a
-/// mixed-graph vector never reaches here. A caller that ever scopes
-/// differently has to add it.
-#[inline]
-fn same_fact_key(a: &Flake, b: &Flake) -> bool {
-    a.s == b.s && a.p == b.p && a.o == b.o && a.dt == b.dt
 }
 
 /// Batched lookup for ref-valued predicate objects across many subjects (V3).
@@ -1188,8 +1043,15 @@ fn apply_raw_overlay_deltas_to_batched_refs(
             None => {
                 subj_entry.insert(class_sid, (flake.t, flake.op));
             }
-            Some(&(t0, _op0)) => {
-                if flake.t > t0 {
+            Some(&(t0, op0)) => {
+                // Newest op wins; at equal `t` the retraction does. Taking
+                // `>` alone let whichever flake the walk happened to yield
+                // first win the tie — and on this lane, which decides
+                // class-based policy grants, the direction it failed was
+                // open: a same-`t` assert+retract of a class left the class
+                // granted. Same rule as `resolve_current_flakes` and
+                // `resolve_overlay_ops`.
+                if flake.t > t0 || (flake.t == t0 && !flake.op && op0) {
                     subj_entry.insert(class_sid, (flake.t, flake.op));
                 }
             }
@@ -1205,6 +1067,14 @@ fn apply_raw_overlay_deltas_to_batched_refs(
             } else {
                 vec.retain(|c| c != class_sid);
             }
+        }
+        // A subject whose overlay flakes are all retractions would otherwise
+        // be left holding an empty vector, which reads as "present with no
+        // classes". `lookup_subject_classes` documents that subjects with no
+        // `rdf:type` assertions are absent from the map, and its per-subject
+        // fallback enforces that — so don't diverge from it here.
+        if vec.is_empty() {
+            out.remove(subj);
         }
     }
 }
@@ -1534,7 +1404,7 @@ fn binary_range_bounded_v3(
 
     // Correctness fallback: merge raw overlay flakes, resolve lifecycles, then apply options.
     flakes.extend(raw_overlay);
-    let mut resolved = resolve_latest_ops_keep_asserts(flakes, IndexType::Spot);
+    let mut resolved = fluree_db_core::range::resolve_current_flakes(flakes, IndexType::Spot);
 
     // Re-apply subject bounds: start_bound.s <= s < end_bound.s.
     resolved.retain(|f| f.s >= start_bound.s && f.s < end_bound.s);
@@ -1589,7 +1459,7 @@ fn overlay_only_flakes_bounded(
     );
 
     // Resolve lifecycles (latest op wins) and drop retracts.
-    let mut resolved = resolve_latest_ops_keep_asserts(flakes, index);
+    let mut resolved = fluree_db_core::range::resolve_current_flakes(flakes, index);
 
     // Apply options after lifecycle resolution.
     if let Some(ref bounds) = opts.object_bounds {
@@ -1610,22 +1480,37 @@ fn overlay_only_flakes_bounded(
 ///
 /// Reached whenever a bound component of `match_val` has no persisted id
 /// (a novelty-only subject, predicate, or object), or no branch manifest
-/// exists for the requested order (genesis / pre-first-index).
+/// exists for `(g_id, order)`.
+///
+/// That second trigger is broader than "genesis / pre-first-index":
+/// `branch_for_order` is keyed per graph AND order, and the bulk-import path
+/// emits SPOT unconditionally while skipping PSOT/POST/OPST when the order
+/// produced no run files (`index_build.rs`'s empty-run-dir early return),
+/// so a graph with zero rows at the indexed `t` lands here for three of the
+/// four orders. Falling back is still correct: the orders that go missing
+/// are exactly the zero-row ones, so there are no persisted rows to drop.
+/// The converse does NOT hold — SPOT can be `Some` with an empty branch —
+/// so nothing here may treat "this order resolved" as "this graph is
+/// indexed".
 ///
 /// The overlay is a **log**, not a current-state view: an assert at `t=2`
 /// and its retraction at `t=3` both sit in novelty. So retractions are kept
-/// through the walk and the whole matching set is lifecycle-resolved
-/// (newest op per fact identity wins, retracted keys drop out) before
-/// options are applied — the same rule the cursor path gets from
-/// `resolve_overlay_ops` and the bounded twin gets from
-/// `resolve_latest_ops_keep_asserts`. Filtering to `op == true` inside the
-/// walk instead silently resurrects every fact whose assert AND retract
-/// both live in novelty.
+/// through the walk and the whole matching set is lifecycle-resolved by
+/// `resolve_current_flakes` — the shared rule, hashing the full
+/// `(s, p, o, dt, m)` identity, which the bounded twin and the untranslated
+/// fallback also use and which the cursor path mirrors in
+/// `resolve_overlay_ops`. Filtering to `op == true` inside the walk instead
+/// silently resurrects every fact whose assert AND retract both live in
+/// novelty.
 ///
-/// This also rules out an early exit at `limit`: the retraction that
-/// cancels an already-collected assert can arrive after the limit is
-/// reached. `for_each_overlay_flake` walks the graph's whole overlay
-/// regardless, so the exit only ever saved clones of matching flakes.
+/// `object_bounds` stays IN the walk: a retraction that can cancel an assert
+/// carries the same `o` by construction (the fact identity requires it), so
+/// the bound is true for both halves or false for both, and pruning early is
+/// equivalent to pruning survivors while collecting and cloning less. The
+/// `offset`/`limit` pair cannot move in with it — they count survivors, and
+/// the retraction that cancels an already-counted assert can arrive after
+/// the count is reached. `for_each_overlay_flake` walks the graph's whole
+/// overlay regardless, so no early exit would shorten the traversal anyway.
 fn overlay_only_flakes(
     store: &Arc<BinaryIndexStore>,
     g_id: GraphId,
@@ -1677,17 +1562,23 @@ fn overlay_only_flakes(
                 }
             }
 
+            // Object bounds are a pure function of `o`, which a fact and its
+            // retraction share, so this prunes whole facts rather than
+            // splitting one — see the note above.
+            if let Some(ref bounds) = opts.object_bounds {
+                if !bounds.matches(&flake.o) {
+                    return;
+                }
+            }
+
             // Keep both asserts and retracts; resolve lifecycles below.
             flakes.push(flake.clone());
         },
     );
 
-    let mut resolved = resolve_latest_ops_keep_asserts(flakes, index);
+    let mut resolved = fluree_db_core::range::resolve_current_flakes(flakes, index);
 
-    // Options apply to the resolved current state, not to the log.
-    if let Some(ref bounds) = opts.object_bounds {
-        resolved.retain(|f| bounds.matches(&f.o));
-    }
+    // Counting options apply to the resolved current state, not to the log.
     if offset > 0 && !resolved.is_empty() {
         let n = offset.min(resolved.len());
         resolved.drain(0..n);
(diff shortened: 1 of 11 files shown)


fluree-db-query/src/binary_range.rs
```diff
@@ -834,31 +852,117 @@ fn resolve_latest_ops_keep_asserts(mut flakes: Vec<Flake>, index: IndexType) ->
...
+
+        let mut g = 0usize;
+        while g < idxs.len() {
+            let mut best = idxs[g];
+            let mut h = g + 1;
+            while h < idxs.len() && flakes[idxs[h]].m == flakes[best].m {
+                let cand = &flakes[idxs[h]];
+                let cur = &flakes[best];
+                // Newest op wins; at equal `t` a retraction beats an assert,
+                // matching `resolve_current_flakes`.
```
bplatz: This says the tie rule matches `resolve_current_flakes`, but it's the opposite. I ran it: `resolve_current_flakes(vec![assert@t2, retract@t2])` returns **1 survivor**, while `retraction_wins_at_equal_t` asserts 0.

Mechanism: `remove_stale_flakes` reverse-scans a comparator-sorted vec, and the comparator orders `op` ascending (`false < true`), so reversing puts the assert first. And `resolve_current_flakes`'s own doc says it exists so callers "get identical results to `range_with_overlay`" — these two are meant to agree.

Either the comment is wrong or one implementation is; right now the divergence reads as deliberate consistency.

fluree-db-query/src/binary_range.rs
```diff
@@ -1129,25 +1244,15 @@ fn batched_refs_overlay_only(
...
-                    .or_default()
-                    .push(class_sid.clone());
+            if flake.p == *predicate && subject_set.contains(&flake.s) {
+                raw_overlay.push(flake.clone());
             }
         },
     );
 
+    let mut out: HashMap<Sid, Vec<Sid>> = HashMap::new();
+    apply_raw_overlay_deltas_to_batched_refs(&mut out, &raw_overlay, predicate, effective_to_t);
```
bplatz, line 1124: Two things about routing through this helper:

**Third tie rule, failing open.** It uses `if flake.t > t0` — strictly greater — so at equal `t` whichever flake the walk yields first wins. Verified: an assert + retract of the same class at the same `t` leaves `["Admin"]`. That's the fail-open direction on the lane that decides class-based policy grants. Pre-existing, but this PR newly routes the overlay-only path through it and adds a unit test pinning its semantics without covering the tie. `flake.t > t0 || (flake.t == t0 && !flake.op)`.

**New behavior diff.** `out.entry(subj).or_default()` now leaves an empty-vec entry for a subject whose only overlay flakes are retractions (verified: `contains_key` true, len 0). `lookup_subject_classes` documents "Subjects with no rdf:type assertions will not be present in the map" and its own fallback enforces that with `if !classes.is_empty()`. Benign today, but one line to fix in the loop below.

fluree-db-query/src/binary_range.rs
```diff
@@ -834,31 +852,117 @@ fn resolve_latest_ops_keep_asserts(mut flakes: Vec<Flake>, index: IndexType) ->
...
+                out.push(flakes[start].clone());
             }
-            i += 1;
+            start = end;
+            continue;
         }
 
-        if flakes[best].op {
-            out.push(flakes[best].clone());
+        // Group the run by `m`. Sorting rather than scanning a winners list
```
bplatz: Was a hash considered here? `fluree_db_core::range::remove_stale_flakes` already solves exactly this — a `HashSet` on the full `(s, p, o, dt, m)` identity, reverse-scanned over the sorted vec. Being a hash, it has no dependency on run adjacency, none on `FlakeMeta::Ord`, no `meta_group_cmp`, and no final `out.sort_by(cmp)`. `FlakeMeta` derives `Hash`/`Eq` consistently and `FlakeValue`'s `Hash` is already documented to agree with its cross-numeric `Eq`.

That's ~110 lines and the whole `Ord`-vs-`Eq` hazard, replaced by a sibling that already exists. There *is* a defensible reason to prefer the sort — hashing `BigDecimal`/`Double` objects isn't free, `hash_decimal_canonical` allocates — but the doc presents the two-level grouping as forced rather than chosen. Worth an answer even if the code stays.

(Separately: `out.sort_by(cmp)` at the end runs unconditionally, including the common all-runs-of-size-1 case where `out` is already sorted. A `bool` set when a run emits >1 survivor skips it.)

fluree-db-query/src/binary_range.rs
```diff
@@ -1568,23 +1677,406 @@ fn overlay_only_flakes(
...
+            // Keep both asserts and retracts; resolve lifecycles below.
             flakes.push(flake.clone());
-            collected += 1;
         },
     );
 
-    Ok(flakes)
+    let mut resolved = resolve_latest_ops_keep_asserts(flakes, index);
+
+    // Options apply to the resolved current state, not to the log.
```
bplatz: `object_bounds` can go back in the walk. A retraction that can cancel an assert has the same `o` by construction — `same_fact_key` requires it — so `bounds.matches` is true for both or false for both, and filtering in the walk is provably equivalent to filtering survivors.

Moving it back shrinks the collected vector, the clone count, and both sorts. For `probe_timestamp_axis`'s `after: Some(target)` leg it's the difference between collecting every commit in novelty and collecting only those past the target — i.e. it retires the "doubly exposed" caveat the bench doc calls out.

Unlike restoring the `limit` early exit — where I agree with the reasoning; trading the internal `sort` for a hard dependency on the walk-order contract is a bad deal in a defensive fallback — this one costs nothing.

fluree-db-query/src/binary_range.rs
```diff
@@ -1501,11 +1606,26 @@ fn overlay_only_flakes_bounded(
     Ok(resolved)
 }
 
-/// Overlay-only results when no branch exists for the requested order.
+/// Overlay-only results when the persisted index cannot contribute a row.
+///
+/// Reached whenever a bound component of `match_val` has no persisted id
+/// (a novelty-only subject, predicate, or object), or no branch manifest
```
bplatz, line 1482: "genesis / pre-first-index" may be understating this. `branch_for_order` is keyed per `(g_id, order)`, so a graph with some orders built and not others lands here too — and in that case this function silently drops persisted rows, which is a bigger hole than the one being fixed.

If all four orders are always built per graph, worth saying so here. If not, worth its own issue rather than a comment that reads as exhaustive.

bplatz (approved):
Approving — the core fix is right, well-diagnosed, and properly pinned. I mutation-checked both halves: restoring the assert-only filter in `overlay_only_flakes` reddens three e2e tests (`["a","b","c"]` vs `["a","c"]`), and restoring it in `batched_refs_overlay_only` reddens the policy test with `["Admin", "Person"]`. The `has_list_meta` work also checks out — root assembly is `base_t: 0` / `prev_index: None`, so there's no prior root a `Some(false)` could clobber, and `import.rs:5412` really is the only non-spool `RunRecord` producer.

Five inline comments, all worth a look before merge. The two I care about most are the equal-`t` tie rules: resolution rules disagreeing across paths that are documented to agree is the same class of defect this PR exists to close.

Minor, not inline: #1685 and #1684 are independent of the read-path fix, so a revert of the fix drags them along. And a footnote for #1711 — `Double(NaN)` has the same `Ord`/`Eq` disagreement one level further down (comparator calls two NaNs equal via `to_bits`, `==` doesn't), so a NaN retraction can't cancel its assert. Unreachable in practice, but it belongs in the same ledger.

aaj3f:
Thanks @bplatz — all five are addressed, and the hash question changed the shape of the PR for the better.

**The tie rule (`:903`) — you were right, and there's a third implementation that settles which one was wrong.** `resolve_overlay_ops` on the cursor path already resolves an equal-`t` assert+retract to *absent*, with a written rationale: it matches the per-commit apply path, and the segment-aware assembly merges runs across segments so the tie has to be deterministic. That makes `remove_stale_flakes` the outlier — and it was contradicting `resolve_current_flakes`'s own "identical results to `range_with_overlay`" promise, exactly as you read it. Fixed there rather than in the comment: the winner is now chosen explicitly instead of falling out of which end a reverse scan happens to reach first. Restoring the original body makes the new tie test fail as you described.

**The third tie rule (`:1254`) — took your fix verbatim**, `flake.t > t0 || (flake.t == t0 && !flake.op)`. And the empty-vec entry was mine, not pre-existing; `lookup_subject_classes` documents absence and its own fallback enforces it, so leaving `or_default()` behind was a real behavior diff. Both now covered by a test that runs assert-first and retract-first orderings plus a retraction-only subject.

**The hash (`:873`) — you're right, and I took it.** `remove_stale_flakes` hashes the full `(s, p, o, dt, m)` identity, which is precisely what the two-level grouping was reinventing less safely. `resolve_latest_ops_keep_asserts`, `same_fact_key` and `meta_group_cmp` are gone (−144 lines), the four call sites now go through `resolve_current_flakes`, and with them go the `Ord`-vs-`Eq` hazard, the unconditional final `out.sort_by(cmp)`, and the superlinearity question you raised in the parenthetical. The resolution tests moved to core beside the implementation, with cases the query-side copy didn't have — net −100 lines and more coverage. The doc no longer presents the grouping as forced, because it isn't there.

On your performance caveat: I couldn't resolve it here and I'd rather say so than quote a number. The box is at load 26 against an idle floor near 5.7, and run-to-run spread on `narrow_subject` was 14.9 vs 20.5 µs against a hash-vs-sort delta around 1 µs — the noise is twenty times the signal. Hashing is O(n) against the sort's O(n log n) so it shouldn't be asymptotically worse, and the CI bench gate against the committed baseline is the honest arbiter rather than anything I measure on this machine.

**`object_bounds` back in the walk (`:1687`) — done.** Your equivalence argument holds: `same_fact_key` requires the same `o`, so a cancelling retraction matches the bounds exactly when its assert does, and filtering in the walk is equivalent to filtering survivors. `offset`/`limit` stay out, since those count survivors rather than candidates. That retires the "doubly exposed" caveat in the bench doc, which was the right thing to chase — thank you.

**`branch_for_order` (`:1612`) — you're right that it wasn't exhaustive, and the investigation says no issue is needed.** Not all four orders are built per graph: bulk import emits SPOT unconditionally, even for zero rows with an empty branch, while PSOT/POST/OPST are skipped when the order produced no run files. But the fallback stays correct, because the missing orders are exactly the zero-row ones — there are no persisted rows to drop. The comment now says that, plus the converse that actually is unsafe: SPOT can be `Some` over an empty branch, so nothing should read "this order resolved" as "this graph is indexed."

**On the NaN footnote — I don't think it holds, and I checked by execution rather than reading:**

```
bits n1=0x7ff8000000000000  n2=0x7ff8000000000001
same-bits   cmp=Equal  eq=true   hash_eq=true
diff-bits   cmp=Less   eq=false  hash_eq=true
```

`Ord` and `Eq` agree in both directions. The reason is that `PartialEq` routes numerics through `numeric_cmp` (`value.rs:401-405`), which carries the *same* `to_bits` fallback the comparator's `same_type_cmp` does (`value.rs:504`) — so identical-bit NaNs are equal under both relations and differing-bit NaNs are unequal under both. So it doesn't belong in #1711's ledger, and I've recorded the negative result there so nobody re-derives it. The `hash_eq=true` on differing bits is just a hash collision, which the contract permits and `Eq` disambiguates — worth noting only because the resolver now hashes that identity.

On the topology note — fair, and a deliberate call rather than an oversight: #1685 and #1684 do ride along with a revert of the read-path fix. Consolidated PRs are the house preference and this one's already reviewed and green, so I'd rather not split it now.

PR #1702, opened 2026-08-26
fix(storage): clamp FileStorage ranged reads to the object, and sweep orphaned .tmp staging files

### LINKED ISSUE #1620: storage: orphaned .tmp staging files are never swept after a crash mid-write
Follow-up from review of #1612, which was happy for this to be a separate issue rather than growing that PR.

## What happens

`FileStorage` writes stage to a sibling `<name>.<pid>.<seq>.tmp` and rename (or hard-link) into place. If the process dies between `File::create(&tmp)` in `stage_bytes` and the rename, the staging file is left behind, and nothing in the repo ever removes it.

`stage_bytes` cleans up on its own error paths, and `write_atomic` / `create_new_atomic` unlink the staging file if the rename or link fails. None of that runs for a `SIGKILL`, an OOM kill, or a power loss.

- `fluree-db-core/src/storage/file.rs` — `stage_bytes` (~L99), `write_atomic` (~L127), `create_new_atomic` (~L158)

## Why it is low severity

`list_prefix` skips anything matching `is_tmp_artifact` (`file.rs` ~L495), so an orphan can never be handed out as a content address or read back as an object. They are inert bytes.

## Why it is still worth closing

They accumulate. A crash loop writes one orphan per attempt, and disk exhaustion is one of the things that produces a crash loop in the first place — so the failure mode is self-reinforcing on exactly the volume that can least afford it. Each orphan is a full copy of the object being written, so an index build or bulk import crashing repeatedly is the bad case, not a commit loop.

## Options

1. **Sweep at storage construction.** In `FileStorage::new` (~L225), walk the base path for `*.tmp` older than some threshold and unlink. Simple and portable; the threshold exists so a sweep never races a concurrent writer's in-flight staging file. Cost is a directory walk at startup, which wants to be bounded or backgrounded on a large volume.
2. **`O_TMPFILE` + `linkat` on Linux.** Closes it properly rather than papering over it: the staging file has no directory entry at all, so a crash cannot leave one. Needs a fallback for macOS and for filesystems without `O_TMPFILE` support, so it does not remove the need for (1) everywhere — it just means Linux never generates orphans to sweep.

Either is fine. (1) alone is probably the right amount of work; (2) is the better answer on the platform that matters most for production.

## Not covered by tests today

Whatever lands should assert that an orphan planted before construction is gone afterwards, and — importantly — that a staging file belonging to a live concurrent write is *not* swept.

### LINKED ISSUE #1671: storage: FileStorage::read_byte_range panics on an open-ended range where ProxyStorage clamps
### Summary

`FileStorage::read_byte_range` panics on an open-ended range. `ProxyStorage` supports that same call and clamps it, so the two implementations of `StorageRead::read_byte_range` disagree on a range shape the trait does not forbid.

```rust
storage.read_byte_range(addr, 0..u64::MAX).await
// -> Err(Io("spawn_blocking failed: task panicked with message \"capacity overflow\""))
```

The panic is in the allocation, before any I/O (`fluree-db-core/src/storage/file.rs`):

```rust
let len = (range.end - range.start) as usize;
...
let mut buf = vec![0u8; len];   // len == usize::MAX for an open-ended range
```

`spawn_blocking` catches it and surfaces it as an `Io` error rather than unwinding the caller, so it degrades to a confusing error rather than a crash — but it is still an unguarded allocation sized directly from caller input.

### Why it matters

Open-ended ranges are an established convention against the same trait. `fluree-db-server/tests/proxy_integration.rs:1823` uses `mid..u64::MAX` for "read to the end" and asserts the behavior explicitly:

```rust
// Range extending past the object is clamped.
let tail = proxy_storage.read_byte_range(&leaf_address, mid..u64::MAX).await
    .expect("tail range read should succeed");
assert_eq!(tail, direct_leaf_bytes[mid as usize..].to_vec());
```

So code written against `StorageRead` and tested on a proxy-backed deployment works, and the identical call fails on a file-backed one. That is exactly the kind of divergence a shared trait is supposed to prevent.

### Reachability

Latent today, not live: no production caller passes an open-ended range to `FileStorage`. Every in-tree `..u64::MAX` is either a proxy test or unrelated. The current `FileStorage` callers compute concrete ranges from leaflet metadata.

Found while reviewing #1599 (probing `read_byte_range` with a zero-length blob); it is unrelated to that PR's subject, hence a separate issue.

### Suggested fix

Clamp the requested length to the file size rather than trusting the range. #1599 now reads `file.metadata()` on the open handle in this same function for its zero-length guard, so the size is already available:

```rust
let file_len = file.metadata().map(|m| m.len()).unwrap_or(0);
let available = file_len.saturating_sub(offset);
let len = (range.end - range.start).min(available) as usize;
let mut buf = vec![0u8; len];
```

The read loop already truncates at EOF, so behavior for in-range requests is unchanged — this only stops the oversized allocation and makes the file backend clamp the way the proxy backend does.

Worth deciding alongside it: whether `StorageRead::read_byte_range`'s doc comment should state that `range.end` past the object is clamped, so it is a contract implementations must meet rather than an accident of each backend.

aaj3f:
Two storage fixes that both live in `fluree-db-core/src/storage/file.rs`, so they're riding together rather than as two one-file PRs.

Fixes #1671
Fixes #1620

## #1671 — ranged reads now clamp, like the trait already said they would

The issue frames this as `FileStorage` and `ProxyStorage` disagreeing. Having gone looking, I think it's actually a bit stronger than that: the trait's own doc comment at `fluree-db-core/src/storage.rs:212-214` already promises the clamp —

> Returns the bytes within the range, which may be shorter than requested if the object is smaller than `range.end`.

— and the default impl (`storage.rs:219-229`), `MemoryStorage` (`storage/memory.rs:94-107`), and the proxy (which has no override, so it inherits the default) all honour it. `FileStorage` was the only `StorageRead` implementation that sized its buffer from `range.end - range.start` rather than from the object, so this is a lone contract violation rather than a sibling divergence. I think that also answers the open question at the end of the issue — there isn't really a doc decision to make, the doc was already right and the code just wasn't meeting it.

Mechanism, and it's worth being precise that this was never only about open-ended ranges: `let len = (range.end - range.start) as usize` ran before any I/O, so the allocation was sized from the *request* for **any** range. `0..40_000_000_000` on a 10-byte file allocated 40 GB of address space and handed the caller a `Vec` with len 10 and capacity 40,000,000,000 — I reproduced that shape directly and got exactly `len=10 capacity=40000000000` at ~1.7 MB max RSS, because `vec![0u8; len]` gets lazy zero pages that never fault in. So oversized ranges "worked" while quietly reserving whatever the caller asked for, and only a width past `Vec`'s capacity limit actually tripped: `mid..u64::MAX` — the established "read to the end" spelling, asserted against the proxy at `fluree-db-server/tests/proxy_integration.rs:1821` — became `vec![0u8; usize::MAX]`, whose capacity-overflow panic `spawn_blocking` caught and relabelled `Io("spawn_blocking failed: task panicked ...")`. A slightly unkind touch: the one backend that couldn't serve the call was also the one that couldn't say why.

The fix is the one suggested in the issue. #1599 already stats the open handle in this same function for its zero-length guard, so the size was sitting right there — I hoisted that single `metadata()` call above both guards and clamped with it:

```rust
let len = requested.min(file_len.saturating_sub(offset)) as usize;
```

The read loops already stopped at EOF, so an in-range request is byte-identical to before.

One deliberate behaviour change worth flagging, since it's the bit I'm least certain about: the zero-length guard used to be `file.metadata().map(|m| m.len() == 0).unwrap_or(false)`, i.e. it tolerated a stat failure and carried on. It now returns `Io` instead. My reasoning is that the buffer size is *derived* from that stat, so without it there's nothing to size the read against, and an error naming the failure beats either guessing a length or falling back to trusting the caller's `u64`. In practice this is `fstat` on a descriptor the thread owns and doesn't fail — but if the permissive form is preferred, I'm happy to put it back.

I checked the other implementations for the same class while I was in there and none of them have it: the default / memory / proxy-filtered paths slice after a full read, `Arc<dyn Storage>` and the two routers in `fluree-db-api/src/lib.rs` are pure delegation, and S3 and proxy-raw hand the range straight to a remote as an HTTP `Range` header, where the server does the clamping per RFC 7233.

One boundary over, though, the backends still don't fully agree, and the trait doc now names **both** known divergences rather than letting one stand as if it were exhaustive — a contract paragraph with an unnamed exception is exactly the shape #1671 grew from. First: at `start >= object length`, `FileStorage` and `MemoryStorage` both return `Ok(vec![])` — this PR's own test asserts it at `99..u64::MAX` — while S3 returns 416. That predates this change, isn't in this diff, and is tracked as #1712; fixing it wants someone who can exercise real S3 rather than my reading of RFC 7233. Second: a zero-length object reads as `NotFound` from `FileStorage` — the #1599 debris guard treats an empty blob as absent on every read surface, this one included — where `MemoryStorage` and the default implementation return `Ok(vec![])`. That one is deliberate and stays deliberate; it's still a divergence in this method's observable behaviour, so the doc says so.

## #1620 — sweeping staging files a crash left behind

Confirmed neither candidate coverer actually covers this. #1612 cleans up on its own error paths in `stage_bytes` / `write_atomic` / `create_new_atomic`, and none of those run for a `SIGKILL`, an OOM kill or a power loss. #1614's sweep plans over `list_prefix`, which `continue`s on `is_tmp_artifact` (`file.rs:561` at the merge base) — so it structurally can never enumerate one of these files, let alone reclaim it. That filter is exactly what makes an orphan harmless and immortal at the same time.

The sweep is `FileStorage::sweep_orphaned_staging()`, an **explicit startup action** — deliberately *not* mounted on the constructor. That placement is load-bearing, not taste. The parser below tolerates one residual (a foreign `<name>.<decimal>.<decimal>.tmp` still parses as a legacy orphan), and the argument for tolerating it leans entirely on the sweep being a deliberate, visible, disable-able startup step. Hung off `new`, that argument collapses, because `FileStorage::new` is reachable from any unit test or tool pointed at a directory the process doesn't own — `fluree-db-connection` has a Debug-formatting test that constructs `FileStorage::new("/tmp/test")`, and with the sweep on the constructor, running that one test unlinks a 90-day-old `mybackup.json.1.2.tmp` an operator left under `/tmp/test`. Each decision is defensible alone; composed, they silently delete operator data. So constructing a storage now touches nothing on disk — which also takes the env reads, the `canonicalize` and the global claim-mutex back out of the constructor — and the layers that actually *know* it's startup make the call:

- `create_sync_connection` / `create_async_connection`'s file arms in `fluree-db-connection`
- `FlureeBuilder::build`, `build_encrypted_internal`, `build_client_file`, and the address-identifier storages in `build_local_storage_from_config` in `fluree-db-api`

That covers every production entry to a file-backed instance — the server and CLI arrive through those builders, and `FileNameService::new` shares the very tree the builder already sweeps, with the once-per-canonicalized-path claim making the layered calls idempotent. `fluree-search-httpd` constructs read-only `FileStorage` handles and deliberately does not sweep: a reader has no business unlinking anything in a writer's tree, and no released build ever swept there.

**The part I'd most like a second opinion on is the concurrency stance**, so it's worth being explicit about what it does and doesn't promise. The tree is shared — between processes in a multi-instance deployment, and between *subsystems* even in one process — and a sweep that guesses wrong pulls a live writer's file out from under it. Three rules:

0. **Never a file this backend's staging writer didn't name.** `.tmp` is a suffix, not a namespace: the indexer's vocab merge, the disk cache, the nameservice tracking file and the Raft log all stage under it, and the nameservice shares this exact tree — `FileNameService::new` builds a `FileStorage` on the path `FlureeBuilder::build` also hands to storage. So a file is only the sweep's business once its name parses as one *this* writer produced, with every segment shape-checked rather than just the dot count. Rules 1 and 2 apply only to names that pass. Predicating on `ends_with(".tmp")` instead — the obvious cheap version — would leave the age heuristic as the only thing standing between `vocab_merge`'s index-build temporaries and an unlink, and a long-running index build is precisely the shape rule 2 is not competent to judge.
1. **Never a staging file bearing this process's token.** `tmp_sibling` embeds a 64-bit token drawn once per process, so a name carrying ours is ours. In-flight and already-leaked are indistinguishable from a directory entry, so both are left alone. This rule is exact — *for the current name format*. A legacy `<name>.<pid>.<seq>.tmp` name carries a pid where the token now sits, and a pid can never equal a 16-hex token, so for legacy names rule 1 is inert and rule 2 stands alone. That case isn't a corner, it's the upgrade path itself: a rolling upgrade has a still-running v4.1.5/v4.1.6 process staging into the shared tree in the old format, protected by the age heuristic and nothing else. Fine in practice — a staging write is open for milliseconds against a one-day threshold — but the rules don't layer there, and the docs now say so plainly instead of presenting 1 and 2 as stacked defenses.
2. **Never a file modified inside 24 hours.** This one is a heuristic and I want to name it as one: it reads another host's clock through ours, and it assumes no legitimate staging write stays open that long.

Anything the sweep can't classify — an unparseable name, an entry it can't stat, an mtime in the future — is kept. Every unknown resolves toward leaving the file alone.

Two notes on that parser, since it's the load-bearing bit. Zero-padded numbers are rejected — nothing here emits one, and it keeps date- and offset-stamped foreign names (`backup.2026.08.tmp`, `wal.000001.000002.tmp`) out of the legacy branch. **One residual remains, deliberately:** a foreign name shaped `<non-empty>.<decimal>.<decimal>.tmp` — `dump.1.2.tmp` — still parses as legacy. I checked all four in-tree `.tmp` producers against the parser and none emits that shape, so it's an operator's own file rather than a collision with anything we ship. Closing it properly means gating legacy reclaim behind an opt-in, which is an upgrade-behavior decision rather than a parser one — and with the sweep now reachable *only* from an explicit startup call an operator can see coming and switch off, rather than from any construction, I think recording it on `staging_token` is the right resting place.

And the legacy branch is load-bearing rather than politeness: staging writes arrived in `85183c9ca`, which **v4.1.5 and v4.1.6 both contain**, so every store written by a currently released build produces orphans in that format. Dropping it would strand them on disk permanently.

On why an age threshold is defensible here when #1635 argues (correctly, I think) that it isn't for the index-build sweep: the objection there is that a reindex replays the whole commit chain, so its duration grows with the ledger and no constant can bound it. A staging file doesn't have that shape — its entire life is one `write_all` of one already-in-memory buffer, followed immediately by a rename. There's no long-running case to get wrong. I believe that's a genuinely different situation rather than the same heuristic reused, but if it reads as the same heuristic wearing a different hat, I'd rather hear that before this ships than after.

**What it explicitly does not guarantee:** it is not a lease, and it does not coordinate with other processes. A foreign staging write that somehow stayed open for over a day would be unlinked. Even then I think the blast radius is small — on POSIX the writer keeps its open descriptor, so its `write_all` and `sync_all` still succeed against the now-unlinked inode and only the final rename fails, surfacing as a write error rather than anything torn or wrong landing under a content address. But it *is* a spurious write error, and I'd rather say that than imply the sweep is airtight. If/when #1635 lands a real lease, this is a natural thing to hang off it.

Practicalities: the walk is handed to the blocking pool when there's a runtime to hand it to (the sweep is called from `create_async_connection` and the API's async client builds, so a recursive `read_dir` on the caller would park a runtime worker — ~1s warm on local APFS for the default budget), and runs inline only for the genuinely synchronous caller (`create_sync_connection`). It runs at most once per base path per process however many startup layers call it, keyed on the canonicalized path so that's a property of the directory rather than of how someone spelled it.

It's also bounded, at 100k directory entries. **Exhausting that budget is not a deferral**, and it would be easy to describe it as one: the walk restarts from the top with no cursor and never removes content, so if the first 100k entries it meets are content, every later start re-walks those same entries and never reaches the orphans past them. That case logs at `warn` (reclaiming logs at `info` — different events), and `FLUREE_STORAGE_TMP_SWEEP_BUDGET` raises the cap, because a warning an operator can't act on is just noise. `FLUREE_STORAGE_TMP_SWEEP=0` turns the sweep off entirely, using the same falsey spellings `FLUREE_STORAGE_FSYNC` already accepts. The budget is a separate variable rather than an overload of that one on purpose — `FLUREE_STORAGE_FSYNC=1` means "on", so reading `FLUREE_STORAGE_TMP_SWEEP=1` as a one-entry budget would be "off" wearing a disguise.

## Tests

For #1671 the guard test fails at the parent commit with the exact error from the issue and passes after. There's also a differential test that runs nine ranges — three of them open-ended — through both `FileStorage` and `MemoryStorage` and asserts they agree, since backend agreement is the property that was actually broken.

For #1620, both directions the issue asked for, plus both directions of the mounting decision — the constructor must not sweep, and no startup path may lose its sweep:

- `construction_is_pure_and_the_explicit_sweep_reclaims_an_orphan` — two pins in one: an aged orphan **survives** `FileStorage::new`, then the explicit sweep reclaims it
- `file_connection_startup_sweeps_orphaned_staging` (`fluree-db-connection`) and `building_a_file_instance_sweeps_orphaned_staging` (`fluree-db-api`) — the flip-side regression guards: opening a file connection and building a file-backed client each still reclaim a planted stale orphan. Both run without a runtime, so the walk is inline and the assertion can't race it.
- `the_sweep_leaves_a_live_looking_staging_file_alone` — a fresh foreign-token file survives
- `our_own_staging_file_is_never_reclaimed_however_old` — rule 1 beats rule 2 even at a zero threshold
- `the_sweep_ignores_tmp_files_other_subsystems_wrote` — the seven real foreign `.tmp` shapes, plus an arity case, all aged past the threshold and swept at a *zero* threshold so age offers no cover and rule 0 is the only thing on trial
- `the_shape_check_still_reclaims_our_own_orphans` — the tightening didn't cost us the orphans the sweep exists for, current format and legacy alike, sitting next to a foreign file that must survive
- `the_walk_is_handed_to_the_blocking_pool_when_a_runtime_exists` / `the_walk_runs_inline_without_a_runtime` — the hand-off both ways; the async one awaits the returned `JoinHandle` rather than polling for the file
- `a_future_mtime_is_treated_as_live`, `only_files_past_the_threshold_are_reclaimed`, `the_sweep_never_touches_content`, `the_walk_is_bounded_by_its_entry_budget`, `a_base_path_is_swept_at_most_once_per_process`

And the concrete shape that motivated the mounting: with a 90-day-old `mybackup.json.1.2.tmp` planted in `/tmp/test`, both hardcoded-path unit tests (`test_connection_handle_file` in `fluree-db-connection/src/lib.rs`, `test_create_file_storage` in `fluree-db-connection/src/storage.rs`) run green and the file survives. The one test that pointed a *real* startup path at `/tmp/test` — `test_fluree_builder_file` — now builds in a tempdir, since a startup path sweeping is exactly its documented behaviour.

I mutation-checked the mounting as well as the rules this time — ten mutants, ten killed, each by the test that owns it: disabling the shape check reddens `the_sweep_ignores_tmp_files_other_subsystems_wrote`; disabling the token rule reddens `our_own_staging_file_is_never_reclaimed_however_old` and only that; disabling the age rule reddens three; admitting zero-padded decimals reddens the parser test with `Some("2026")` for `backup.2026.08.tmp`; a spawned walk that never runs reddens the blocking-pool test; a gutted inline walk reddens two; letting every claim win reddens the once-per-path test; hanging the sweep back on `new` reddens the purity pin; and deleting the wiring call from either `create_sync_connection` or `FlureeBuilder::build` reddens the corresponding startup pin. The assertions check the `reclaimed`/`kept` counts and not just file existence, so a file that was never *examined* can't pass as one that was deliberately kept.

## Gates

- `cargo test -p fluree-db-core --all-features` — 815 lib tests green (46 of them in `storage::file`), plus integration and doc-tests
- `cargo test -p fluree-db-connection --lib` (39), `cargo test -p fluree-db-api --lib` (776), `cargo test -p fluree-db-nameservice --lib` (163) — the crates whose startup paths now own the sweep call
- `cargo check -p fluree-db-connection --features aws` — the async (S3-capable) connection path compiles with its file-arm sweep
- `cargo check -p fluree-db-core --no-default-features --features native --lib` — clean, since `Handle::try_current` needs tokio's `rt` and I didn't want that quietly supplied by a dev-dependency
- `cargo clippy -p fluree-db-core --all-features` / `-p fluree-db-connection` / `-p fluree-db-api` — `--all-targets --no-deps`, clean
- `cargo fmt --all --check` — clean

Docs: the "Staging files left by a crash" section in `docs/operations/storage.md` now says the sweep is an explicit startup step (constructing a storage handle never deletes anything) and carries the legacy-format carve-out in operator terms, next to the durability table.

### DIFF
--- a/fluree-db-core/src/storage.rs
+++ b/fluree-db-core/src/storage.rs
@@ -222,9 +222,19 @@ pub trait StorageRead: Debug + Send + Sync {
     /// gives — absence of bytes is not an error here, and callers distinguish
     /// "no bytes" from "no object" by the `NotFound` error, not by this.
     ///
-    /// Known gap: `S3Storage` returns 416 rather than an empty vec for a start
-    /// past the end. That predates this doc and is tracked in #1712; the
-    /// clamping half of the contract holds on every backend.
+    /// Two known divergences, named so this paragraph stays authoritative
+    /// rather than optimistic — a contract doc that names one exception as if
+    /// it were the only one is how a backend drifts out from under it:
+    ///
+    /// - `S3Storage` returns 416 rather than an empty vec for a start past
+    ///   the end. That predates this doc and is tracked in #1712; the
+    ///   clamping half of the contract holds on every backend.
+    /// - A **zero-length object** reads as `NotFound` from `FileStorage` —
+    ///   the #1599 debris guard treats an empty blob as absent on every read
+    ///   surface, this one included — where `MemoryStorage` and this default
+    ///   implementation return `Ok(vec![])`. Deliberate, and callers relying
+    ///   on the guard want exactly that answer, but it is a divergence in
+    ///   this method's observable behavior all the same.
     ///
     /// The default implementation fetches the full object and slices.
     /// StorageBackends that support native range reads (S3, HTTP) should override
--- a/fluree-db-core/src/storage.rs
+++ b/fluree-db-core/src/storage.rs
@@ -223,7 +223,7 @@ pub trait StorageRead: Debug + Send + Sync {
     /// "no bytes" from "no object" by the `NotFound` error, not by this.
     ///
     /// Known gap: `S3Storage` returns 416 rather than an empty vec for a start
-    /// past the end. That predates this doc and is tracked separately; the
+    /// past the end. That predates this doc and is tracked in #1712; the
     /// clamping half of the contract holds on every backend.
     ///
     /// The default implementation fetches the full object and slices.
--- a/fluree-db-core/src/storage.rs
+++ b/fluree-db-core/src/storage.rs
@@ -213,6 +213,19 @@ pub trait StorageRead: Debug + Send + Sync {
     /// the range, which may be shorter than requested if the object is
     /// smaller than `range.end`.
     ///
+    /// **Clamping is the contract, not an implementation detail.** An
+    /// implementation must size its result from the object, never from the
+    /// width of the range: `mid..u64::MAX` is the established spelling of
+    /// "read to the end" against this trait, so a `range.end` past the object
+    /// is normal input rather than an error. A `range.start` at or past the
+    /// end of the object returns `Ok(vec![])`, the same answer an empty range
+    /// gives — absence of bytes is not an error here, and callers distinguish
+    /// "no bytes" from "no object" by the `NotFound` error, not by this.
+    ///
+    /// Known gap: `S3Storage` returns 416 rather than an empty vec for a start
+    /// past the end. That predates this doc and is tracked separately; the
+    /// clamping half of the contract holds on every backend.
+    ///
     /// The default implementation fetches the full object and slices.
     /// StorageBackends that support native range reads (S3, HTTP) should override
     /// for efficiency.
--- a/fluree-db-core/src/storage/file.rs
+++ b/fluree-db-core/src/storage/file.rs
@@ -112,6 +112,173 @@ fn is_tmp_artifact(name: &str) -> bool {
     name.ends_with(TMP_SUFFIX)
 }
 
+/// The process token embedded in a staging name, for names shaped the way
+/// [`tmp_sibling`] writes them (`<name>.<pid>.<token>.<seq>.tmp`).
+///
+/// Returns `None` for anything else ending in `.tmp` — a file an operator
+/// dropped there, or one written by an older version of this code. The sweep
+/// treats an unparseable name as *not* ours, which is the safe direction: it
+/// still has to clear the age threshold before anything happens to it.
+fn staging_token(name: &str) -> Option<&str> {
+    let rest = name.strip_suffix(TMP_SUFFIX)?;
+    let (head, _seq) = rest.rsplit_once('.')?;
+    let (_prefix, token) = head.rsplit_once('.')?;
+    Some(token)
+}
+
+/// How long a staging file must have gone untouched before a sweep may
+/// reclaim it.
+///
+/// A staging file's entire life is one [`stage_bytes`] call: create, write one
+/// in-memory buffer, optionally fsync, then rename or link immediately. There
+/// is no legitimate case where that takes a long time, which is what makes an
+/// age threshold a sound discriminator here — unlike an index build (#1635),
+/// whose duration grows with the ledger and so can never be bounded by a
+/// constant. A day is orders of magnitude past any real staging write and
+/// swamps NTP-scale clock skew between hosts sharing a mount.
+const STALE_STAGING_AGE: std::time::Duration = std::time::Duration::from_secs(24 * 60 * 60);
+
+/// Directory entries one sweep will look at before giving up.
+///
+/// The walk runs on the construction path, so it must not turn opening a large
+/// volume into a startup stall. Exhausting the budget leaves the rest of the
+/// tree for the next start rather than delaying this one — orphans are inert,
+/// so deferring them costs nothing but the disk they sit on.
+const SWEEP_ENTRY_BUDGET: usize = 100_000;
+
+/// What one sweep did.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+struct StagingSweep {
+    /// Staging files unlinked.
+    reclaimed: usize,
+    /// Staging files deliberately left alone — this process's own, too young,
+    /// or refusing to be removed.
+    kept: usize,
+    /// The entry budget ran out before the walk finished, so part of the tree
+    /// was never looked at.
+    truncated: bool,
+}
+
+/// Unlink staging files left behind by writes that never finished.
+///
+/// A crash between the `File::create` in [`stage_bytes`] and the rename that
+/// follows leaves a full copy of the object on disk under a `.tmp` name.
+/// `list_prefix` filters those out, so one can never be served as content —
+/// which is also precisely why nothing ever removed them. They accumulate, and
+/// the crash loop that produces them is often a disk-exhaustion crash loop, so
+/// the growth lands on the volume that can least afford it.
+///
+/// # What this will not delete
+///
+/// Storage is shared by more than one process in a multi-instance deployment,
+/// so a sweep that guessed wrong would pull a live writer's file out from under
+/// it. Two rules stop that:
+///
+/// 1. **Never this process's own.** The staging name carries a 64-bit token
+///    drawn once per process, so a name bearing our token is ours — in flight
+///    or already leaked, and a directory entry cannot tell those apart. Both
+///    are left alone. This rule is exact, not a heuristic.
+/// 2. **Never a file touched recently.** Anything modified within `older_than`
+///    is left for whoever is writing it. This rule *is* a heuristic: it reads
+///    the writer's clock through ours, and it assumes no legitimate staging
+///    write stays open that long. See [`STALE_STAGING_AGE`] for why that
+///    assumption holds for staging files specifically.
+///
+/// Anything the sweep cannot classify — a name it cannot parse, an entry it
+/// cannot stat, an mtime in the future — is kept. Every unknown resolves
+/// toward leaving the file alone.
+///
+/// If both rules were somehow beaten, the damage is bounded: on POSIX the
+/// writer keeps its open descriptor, so its `write_all` and `sync_all` still
+/// succeed against the now-unlinked inode and only the final rename fails. The
+/// write reports an error; nothing partial is ever published under a content
+/// address.
+fn sweep_orphaned_staging_files(
+    base: &Path,
+    older_than: std::time::Duration,
+    budget: usize,
+) -> StagingSweep {
+    let mut sweep = StagingSweep::default();
+    let own_token = format!("{:016x}", process_token());
+    let now = std::time::SystemTime::now();
+    let mut budget = budget;
+    let mut dirs = vec![base.to_path_buf()];
+
+    while let Some(dir) = dirs.pop() {
+        let Ok(entries) = std::fs::read_dir(&dir) else {
+            continue;
+        };
+        for entry in entries.flatten() {
+            if budget == 0 {
+                sweep.truncated = true;
+                return sweep;
+            }
+            budget -= 1;
+
+            let Ok(file_type) = entry.file_type() else {
+                continue;
+            };
+            if file_type.is_dir() {
+                dirs.push(entry.path());
+                continue;
+            }
+            if !file_type.is_file() {
+                continue;
+            }
+
+            let name = entry.file_name();
+            let name = name.to_string_lossy();
+            if !is_tmp_artifact(&name) {
+                continue;
+            }
+            // Rule 1: ours, whatever its age.
+            if staging_token(&name) == Some(own_token.as_str()) {
+                sweep.kept += 1;
+                continue;
+            }
+            // Rule 2: young enough that someone may still be writing it. An
+            // mtime we cannot read, or one in the future, counts as young —
+            // `duration_since` fails on a future timestamp, and a clock the
+            // sweep does not understand is not grounds for deleting data.
+            let stale = entry
+                .metadata()
+                .and_then(|m| m.modified())
+                .ok()
+                .and_then(|m| now.duration_since(m).ok())
+                .is_some_and(|age| age >= older_than);
+            if !stale {
+                sweep.kept += 1;
+                continue;
+            }
+
+            match std::fs::remove_file(entry.path()) {
+                Ok(()) => sweep.reclaimed += 1,
+                // Someone else got there first, or the platform refuses to
+                // unlink a file another process still holds open (Windows).
+                // Both are fine outcomes for a best-effort reclaim.
+                Err(_) => sweep.kept += 1,
+            }
+        }
+    }
+    sweep
+}
+
+/// Base paths this process has already swept.
+///
+/// `FileStorage::new` runs once per connection, once per nameservice and again
+/// for the API's own handle, frequently on the same directory. Only the first
+/// walk can find anything; the rest would re-walk the tree to look at exactly
+/// the files the first one declined to touch.
+fn claim_sweep(base: &Path) -> bool {
+    static SWEPT: std::sync::OnceLock<std::sync::Mutex<std::collections::HashSet<PathBuf>>> =
+        std::sync::OnceLock::new();
+    SWEPT
+        .get_or_init(Default::default)
+        .lock()
+        .unwrap_or_else(std::sync::PoisonError::into_inner)
+        .insert(base.to_path_buf())
+}
+
 /// Write `bytes` to a staging sibling of `path`, returning the staging path.
 ///
 /// Under [`Durability::Sync`] the contents are flushed before returning, so a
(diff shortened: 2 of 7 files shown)


fluree-db-core/src/storage/file.rs
```diff
@@ -244,14 +500,146 @@ impl FileStorage {
...
+    ///
+    /// Constructing the storage also reclaims staging files orphaned by an
+    /// earlier crash — see [`sweep_orphaned_staging_files`] for what it will
+    /// and will not delete, and [`Self::SWEEP_ENV_VAR`] to tune or disable it.
+    /// The walk is bounded, runs at most once per base path per process, and is
+    /// handed to the blocking pool when there is a runtime to hand it to, so
+    /// constructing a storage never waits on it.
     pub fn new(base_path: impl Into<std::path::PathBuf>) -> Self {
+        let base_path = base_path.into();
+        let _ = Self::sweep_on_construction(&base_path);
```
bplatz: This is the one I'd change before merge: a destructive, environment-mutating sweep hangs off a bare constructor, and it composes badly with the parser residual you documented on `staging_token`.

`fluree-db-connection/src/lib.rs:400` is a unit test whose whole body asserts a Debug string, and it constructs `FileStorage::new("/tmp/test")` — a hardcoded shared path. I created `/tmp/test`, planted `mybackup.json.1.2.tmp` (exactly the `<name>.<decimal>.<decimal>.tmp` shape you describe as "an operator's own file rather than a collision with anything we ship"), stamped it 90 days old, and ran that single test:

```
planted, mtime 90d old: True
test result: ok. 1 passed
ls /tmp/test -> empty
```

The file was gone. Each decision is defensible alone — the residual is tolerable *because* the sweep is a deliberate startup action, and sweeping at startup is reasonable — but together they mean constructing a `FileStorage` silently unlinks operator data.

I'd move it out of `new`: an explicit call from the connection/startup layer, which is the layer that actually knows it's startup, or a builder step next to `with_durability`. That also takes env reads, `canonicalize` and a global mutex back out of a constructor, and makes the `/tmp/test` test harmless again.

If you'd rather keep it here, then I think the residual has to close rather than stay recorded, since `new` being reachable from tests and tooling is what turns it from theoretical into the repro above.

fluree-db-core/src/storage/file.rs
```diff
@@ -112,6 +112,262 @@ fn is_tmp_artifact(name: &str) -> bool {
...
+            // business once its name parses as one *this* writer produced.
+            // Without this, rule 1 protects nothing outside our own naming and
+            // the age heuristic alone stands between every other subsystem's
+            // staging file and an unlink — including the indexer's vocab-merge
+            // temporaries, which are exactly the long-running shape
+            // `STALE_STAGING_AGE` argues an age threshold must not judge.
+            let Some(token) = staging_token(&name) else {
+                continue;
+            };
+            // Rule 1: ours, whatever its age.
```
bplatz, line 331: Worth saying in the doc that rule 1 is exact only for the current format. A legacy name puts a pid in the token slot, so it can never equal a 16-hex `own_token`:

```
staging_token("real.json.<our pid>.0.tmp") -> Some("45116")
own_token                                  -> "3d7b301d75f9a910"   // never matches
```

So for legacy names rule 1 is inert and rule 2 is the only thing standing. The case where that matters is a rolling upgrade — a v4.1.5/v4.1.6 process staging into a shared tree is covered by the heuristic alone. In practice fine, since a staging write won't stay open a day. But the doc presents 1 and 2 as layered, and on the branch that's load-bearing for exactly the deployment shape it was written for.

fluree-db-core/src/storage.rs
```diff
@@ -213,6 +213,19 @@ pub trait StorageRead: Debug + Send + Sync {
...
+    /// **Clamping is the contract, not an implementation detail.** An
+    /// implementation must size its result from the object, never from the
+    /// width of the range: `mid..u64::MAX` is the established spelling of
+    /// "read to the end" against this trait, so a `range.end` past the object
+    /// is normal input rather than an error. A `range.start` at or past the
+    /// end of the object returns `Ok(vec![])`, the same answer an empty range
+    /// gives — absence of bytes is not an error here, and callers distinguish
+    /// "no bytes" from "no object" by the `NotFound` error, not by this.
+    ///
+    /// Known gap: `S3Storage` returns 416 rather than an empty vec for a start
```
bplatz: There are two gaps here, not one. For a zero-length object `FileStorage` returns `NotFound` (the #1599 debris guard) where `MemoryStorage` returns `Ok(vec![])`:

```
ZERO file 0..10 -> Err(NotFound("z/empty: … (zero-length blob, treated as absent)"))
ZERO mem  0..10 -> Ok(0)
```

Deliberate, and I think correct. But this paragraph is here specifically to make the contract authoritative, and naming only S3 reads as exhaustive — which recreates in miniature the thing #1671 was. A clause pointing at the zero-length rule as the other exception would settle it.

bplatz (approved):
Approving. I verified both issues on pristine `main` (`fe3c198c8`) rather than taking them as given, and both are real: `0..u64::MAX` gives `Err(Io("spawn_blocking failed: … capacity overflow"))` on `FileStorage` against `Ok(10)` on `MemoryStorage`, and planted orphans (current and legacy format, aged) survive construction untouched. Your "lone violator" reading holds too — default impl, memory and proxy all clamp.

The clamp fix is clean: on the branch `0..MAX→10`, `5..MAX→5`, `99..MAX→0`, `0..40e9→10`, `3..7→4`, identical to `MemoryStorage` across the board. I also probed the staging parser directly and every claim in the description held — uppercase hex, zero-padded segments and `a.b.c.tmp` all rejected, `dump.1.2.tmp` admitted as the stated residual.

One note for the #1671 write-up: it isn't only open-ended ranges. The allocation was sized from the request for *any* range — `0..40_000_000_000` on a 10-byte file allocated 40 GB of address space and returned a `Vec` with len 10 and capacity 4e10. Lazy zero pages mean no panic and negligible RSS, so only `usize::MAX` actually trips capacity overflow, but the oversized reservation was handed to the caller. Your fix covers all of it; worth a line so it isn't later read as narrowly about `u64::MAX`.

Three things below. The first is a demonstrated defect and I'd want it settled before merge; the other two are a doc sentence each.

On the concurrency stance you asked about: the rules themselves I think are right, and the argument for why an age threshold is sound here but not for #1635 holds up — a staging file's life really is one `write_all` plus a rename, with no long-running case to get wrong. My concern isn't the rules, it's where they're mounted.

aaj3f:
Thanks @bplatz — the `/tmp/test` demonstration settled it, and I took your preferred fix: the sweep moved out of `new`, not the residual closed around it.

`FileStorage::new` is pure again — no filesystem access, no env reads, no claim-mutex. The sweep is an explicit `sweep_orphaned_staging(&self)` keeping the exact `Handle::try_current()`→`spawn_blocking` / inline split, wired at every file-backed startup entry: the connection layer's sync and aws-async file arms, and the four builder paths in `fluree-db-api` (`build`, `build_encrypted_internal`, `build_client_file`, `build_local_storage_from_config`) — the server and CLI arrive through those. Deliberately *not* wired: `FileNameService::new` (it shares the tree the builder already sweeps, and the once-per-canonicalized-path claim dedupes), the bare `create_file_storage` factory, and `fluree-search-httpd`'s read-only handles.

Re-running your repro post-fix: planted `mybackup.json.1.2.tmp` stamped 90 days old in `/tmp/test`, ran that unit test — green, file survives, mtime intact. Chasing the shape further turned up two more of the same class worth knowing about: `test_create_file_storage` also constructs against hardcoded `/tmp/test` (harmless now for the same reason), and `test_fluree_builder_file` in `fluree-db-api` calls the real `build()` path on `/tmp/test` — which means with the sweep wired into `build()` it would have *become* the next deleter. It builds in a tempdir now.

Both regression directions are pinned: `construction_is_pure_and_the_explicit_sweep_reclaims_an_orphan` asserts an aged orphan survives `new` and then dies to the explicit call, and two runtime-free wiring pins (connection and builder) each fail if their sweep call is deleted — so a startup path can't silently lose the sweep, which was the regression I most wanted a guard against after un-mounting it. Full mutation matrix is 10/10, including the two new wiring mutants.

Both doc sentences landed essentially as you framed them: rule 1's legacy carve-out (a pid can never equal a 16-hex token, so rule 2 stands alone for exactly the rolling-upgrade shape), and the contract paragraph now names *two* known divergences — the S3 416 (#1712) and the zero-length `NotFound` from the #1599 guard.

And your allocation finding reproduced exactly — the pre-fix shape on a 10-byte file gives `len=10, capacity=40000000000` at 1.7 MB max RSS, so the reservation was request-sized for any range and only `usize::MAX` widths ever tripped. The body now says so rather than reading as narrowly about `u64::MAX`.

PR #1705, opened 2026-08-26
fix(datalog): IRI comparison in rule FILTERs (fails open today) + a diagnostic when a rule can't derive

### LINKED ISSUE #1556: Datalog rule FILTERs cannot compare against an IRI — the exclusion form fails open silently
## Summary

A datalog rule FILTER that compares a variable against an IRI never matches the IRI. An equality test (`(= ?p ex:knows)`) is **always false** (derives nothing); an inequality test (`(!= ?p ex:password)`) is **always true** (excludes nothing). Both fail with no error and no warning. The inequality case is the dangerous one: an author writing a copy-properties rule who tries to exclude a sensitive predicate gets a silent no-op filter.

Pre-existing on `main` and independent of the property-position-variable work (PR #1541) — but that PR makes it more reachable, because a filter is the natural way to constrain a rule with a variable predicate/object, and this is the reason the obvious "just exclude the sensitive predicate" mitigation isn't available today.

## Mechanism

Three functions in `fluree-db-query/src/datalog_rules.rs` disagree on how an IRI is represented:

1. **Parse** — `parse_filter_term` (`:355`) never expands or resolves IRIs. A colon-bearing token falls through every branch (`?`-var, i64, f64, `true`/`false`) to the final `else`, becoming the **string** `RuleValue::String("ex:knows")`.
2. **Resolve** — `resolve_filter_term` (`:738`) renders a bound `BindingValue::Sid(sid)` as `FlakeValue::String(sid.name.to_string())` — the SID's `name`, i.e. the **local name only**, `"knows"` (the same for `RuleValue::Ref` and `RuleTerm::Sid`).
3. **Compare** — `compare_values` (`:763`) falls back to string comparison, comparing `"knows"` against `"ex:knows"`.

So the two sides never agree:
- `(= ?p ex:knows)` → `"knows" == "ex:knows"` → **false** → rule derives nothing.
- `(!= ?p ex:password)` → `"knows" != "ex:password"` → **true** → nothing is excluded.
- The only operand that matches a bound IRI is its **bare local name**: `(= ?p knows)` → `"knows" == "knows"` → true — but this is namespace-blind, matching `ex:knows`, `foaf:knows`, and anything else named `knows`.

## Not specific to predicate position

`flake_value_to_binding` (`:1062`) maps `FlakeValue::Ref` → `BindingValue::Sid` for **object** position too, so `["filter", "(!= ?o ex:bob)"]` over an ordinary object variable is broken the same way. This is why the bug exists on `main` regardless of PR #1541.

## Failure scenario (the one that hurts)

```json
{
  "where": [
    {"@id": "?s", "ex:sameAs": {"@id": "?other"}},
    {"@id": "?other", "?prop": "?val"},
    ["filter", "(!= ?prop ex:ssn)"]
  ],
  "insert": {"@id": "?s", "?prop": "?val"}
}
```

The author intends to copy every property of `?other` onto `?s` **except** `ex:ssn`. The filter `(!= ?prop ex:ssn)` is always true, so `ex:ssn` is copied anyway — a silent data-exposure footgun with no error or warning.

## Proposed fix

Resolve colon-bearing filter operands the same way node-pattern terms are resolved: through `expand_iri` + `resolve_iri` into `RuleValue::Ref`, and compare `Sid`/`Ref` bindings as `FlakeValue::Ref` (SID equality) rather than stringifying to the local name. `parse_filter_term` has no `@context`/snapshot today, so it would need them threaded in (the node-pattern parsers already receive both).

Minimal alternative: reject a resolvable-IRI filter operand loudly at parse time rather than silently treating it as a string, so an author at least learns the comparison isn't supported instead of getting a wrong answer.

## Acceptance criteria

- `(= ?p ex:knows)` matches a `?p` bound to `ex:knows` and does not match `foaf:knows` (namespace-aware).
- `(!= ?prop ex:ssn)` actually excludes `ex:ssn` in a copy-properties rule.
- The same holds for an object-position IRI filter (`(!= ?o ex:bob)`).
- If full IRI comparison isn't implemented, a resolvable-IRI filter operand produces a clear parse error rather than a silently-wrong string comparison.

## Notes

- Surfaced in review of PR #1541 (`Add regression coverage for sub-SELECT aggregate correlation joins` / datalog property-position variables). Confirmed pre-existing and independent of that PR.
- Related to the "no diagnostics when a rule derives nothing" gap: a silently-empty filter compounds the debugging difficulty, since neither the wrong comparison nor the empty derivation is reported.

### LINKED ISSUE #1558: Datalog JSON-LD rule parser silently mis-classifies bare keys, bare @type values, and JSON-LD keywords
## Summary

The JSON-LD datalog rule parser (`parse_node_pattern` / `parse_term` / `expand_iri` in `fluree-db-query/src/datalog_rules.rs`) has three related classification gaps. All three share the same failure mode — the rule parses "successfully" but matches nothing and derives nothing, with no error or warning — because `LedgerSnapshot::encode_iri` never fails (it falls back to `Sid(EMPTY, iri)` for unrecognized strings), so mis-classified terms become valid-but-bogus Sids instead of loud errors. Found during review of #1531; all three pre-date that fix.

## Gap 1: bare context-mapped keys never expand

`expand_iri` only expands `prefix:local` CURIEs (it requires a `:` in the key). A bare key term-defined directly in the context — normal JSON-LD convention — is returned as-is and encoded as the literal IRI.

```json
{
  "@context": {"name": "http://schema.org/name"},
  "where": {"@id": "?s", "name": "?v"},
  "insert": {"@id": "?s", "ex:hasName": "?v"}
}
```

The predicate resolves to `Sid(EMPTY, "name")` instead of `http://schema.org/name`, so the pattern matches nothing. Existing docs and tests use only prefixed keys, so the gap is latent but real for users following standard JSON-LD style.

## Gap 2: bare @type values parse as string literals

Classification of a bare colon-less string is position-dependent: a key resolves as an IRI, but a value — including `@type` objects — goes through `parse_term`, whose bare-string branch produces a string literal. `{"@type": "Person"}` compiles to `ObjectMatch::String("Person")`, and `flake_object_matches` only pairs Ref-with-Ref, so it can never match `rdf:type` assertions (stored as Refs). The prefixed form `"ex:Person"` contains `:` and works, which is why existing tests pass.

## Gap 3: JSON-LD keywords become bogus predicates

`parse_node_pattern` skips only `@id`, `@context`, and `@type`. Any other keyword key (`@value`, `@graph`, `@list`, `@reverse`, `@language`) falls through to predicate resolution and fabricates a predicate Sid literally named after the keyword. A rule using `@reverse` or `@list` structure parses cleanly and silently matches nothing.

## Proposed fix

`parse_predicate_term` (added in #1531's fix) is now the single chokepoint for key classification, making it the natural place to close all three: consult bare term definitions in the rule's `@context` when expanding keys (and apply the same expansion to `@type` values, or resolve them as IRIs rather than literals), and reject unrecognized `@`-prefixed keys with a parse error instead of resolving them. Where a bare key/value cannot be resolved against the context, error loudly rather than minting `Sid(EMPTY, ...)`.

## Acceptance criteria

- A rule using a bare context-mapped key matches and derives facts identically to the equivalent prefixed form.
- A rule using `@type` with a bare context-mapped class name matches type assertions.
- A rule containing an unsupported JSON-LD keyword key fails to parse with a descriptive error (and is skipped with the existing warn path for stored rules) instead of silently deriving nothing.

### LINKED ISSUE #1560: Datalog rules derive zero facts with no diagnostic when insert-pattern variables don't bind
## Summary

When a datalog rule's `insert` pattern references a variable that is unbound in a binding row — most commonly a `where`/`insert` variable-name typo — or bound to a value unusable in that position (a literal in subject/predicate position), `instantiate_pattern` (`fluree-db-reasoner/src/datalog.rs`) silently returns `None` and the row derives nothing. The rule appears to run, the fixpoint completes, and the only observability is a per-iteration debug-level aggregate counter (`new_facts`) that attributes nothing to any specific rule. Found during review of #1531; the silent-`None` structure pre-dates it for subject variables, and property-position variable support (#1531) extends the same behavior to predicates, where the typo scenario is now newly possible.

## Repro

```json
{
  "where": {"@id": "?s", "ex:relType": {"@id": "?relation"}},
  "insert": {"@id": "?s", "?rel": {"@id": "?s"}}
}
```

`?rel` is never bound (the where clause binds `?relation`), so every binding row is dropped in `instantiate_pattern` via `bindings.get(name)?`. The user sees zero derived facts and has no way to discover why short of reading engine source.

## Notes

Dropping the row is correct semantics — an unbound variable cannot instantiate a fact, and after #1531's follow-up fix a literal-bound subject/predicate variable is likewise a correct empty match. The defect is purely the missing diagnostic.

## Proposed fix

In `execute_datalog_rules_with_query_rules` (`fluree-db-query/src/datalog_rules.rs`), emit a warn-level (or rate-limited) log naming the rule when it produced matching binding rows but zero flakes in an iteration — that condition is precisely "the where clause matched but the insert couldn't instantiate," which is almost always an authoring error. Optionally, a parse-time check that every variable referenced in `insert` patterns also appears in `where` patterns would catch the typo class outright before execution.

## Acceptance criteria

- A rule whose insert references a never-bound variable produces a diagnostic naming the rule (and ideally the variable), while other rules' derivations are unaffected.
- No diagnostic noise for rules that legitimately derive nothing because their where clause matched nothing.

aaj3f:
Two of zonotope's datalog audit findings — they sit on the same plumbing in `fluree-db-query/src/datalog_rules.rs`, so they're here together.

## The one that matters: a datalog exclusion filter fails open, silently

A rule FILTER can't compare a variable against an IRI today. `(= ?p ex:knows)` is always false and `(!= ?prop ex:ssn)` is always **true** — and it's that second direction that makes this a P1 rather than a papercut. An author who writes the obvious copy-properties rule with the obvious mitigation:

```json
"where": [
  {"@id": "?s", "ex:sameAs": {"@id": "?other"}},
  {"@id": "?other", "?prop": "?val"},
  ["filter", "(!= ?prop ex:ssn)"]
],
"insert": {"@id": "?s", "?prop": "?val"}
```

…gets `ex:ssn` copied anyway. No error, no warning, nothing in the diagnostics. The rule that was written specifically to withhold a value is the rule that hands it over. I added that exact repro as a test first and confirmed it fails on `main` — `ex:ssn` and `123-45-6789` both land on `ex:alice`. (Reverting `datalog_rules.rs` to the merge base with the tests kept turns the integration suite red.)

The mechanism is that three functions disagreed about what an IRI even is. `parse_filter_term` (`datalog_rules.rs:355`) had no IRI branch at all, so `ex:ssn` fell past the var/i64/f64/bool arms into the final `else` and became the *string* `"ex:ssn"`. `resolve_filter_term` (`:747`) rendered a bound Sid as `sid.name` — the bare local name `"ssn"`, namespace discarded. Then `compare_values` (`:785`) string-compared `"ssn"` against `"ex:ssn"`. The two sides can never agree, so `=` never fires and `!=` always does. This is pre-existing on `main` and independent of #1541, though that PR makes it much more reachable, since a filter is the natural way to constrain a variable predicate.

Worth noting it isn't predicate-specific either: `flake_value_to_binding` (`:1062`) maps a `Ref` object to a `Sid` binding too, so `["filter", "(!= ?o ex:bob)"]` on an ordinary object variable was broken the same way, in the same fail-open direction. Test included.

### What changed

**Parse** — thread the rule's `@context` and the snapshot into filter parsing (they were never passed in) and resolve an IRI operand exactly the way `parse_term` resolves a node-pattern term. A quoted operand is now unambiguously a string, which gives an explicit escape hatch, and CURIE detection requires an NCName-shaped prefix so colon-bearing *literals* stay literals — no NCName starts with a digit, so `12:30:00` and `2026-08-25T09:30:00Z` don't get swept into IRI classification. (An absolute `http(s)` IRI is itself CURIE-shaped — `http` parses as the prefix — so one classification covers both forms.) A token that survives *none* of these classifications — not a variable, number, boolean, quoted string, or IRI — is a **parse error**, not a string; that's the first behavior change below, and it's what closes the fail-open `!=` direction for good.

**Resolve** — a Sid binding stays a `FlakeValue::Ref` rather than collapsing to its local name, so comparison is namespace-aware. `ex:knows` no longer equals `foaf:knows`, which the old local-name comparison happily conflated.

**Compare** — IRI identity for `=`/`!=` by Sid equality, following RDFterm-equal (SPARQL 1.1 §17.4.1.7): a type error only when *both* operands are literals, so an IRI against a literal is well-defined and simply false. Ordering an IRI is rejected.

**Filter evaluation can't fail open.** Evaluation is three-valued in the SPARQL sense — an unresolvable operand or a literal-vs-literal type mismatch yields `Error`, and a row survives only on an outright `True`. `Error` deliberately does *not* collapse into `False`, because `!(error)` would then be `true` and re-admit a row the engine couldn't actually judge; `RuleFilter::Not` is reachable from SPARQL-lowered rules via `sparql_lang.rs:297`, so that path is live, not theoretical. `And`/`Or` propagate errors the SPARQL way (`false && error` is `false`, `true && error` is `error`).

**Filter *acceptance* couldn't fail open either, and that turned out to matter.** A malformed filter used to return `Ok(None)`, `parse_where_clause` dropped the `None` on the floor, `rule.filters` ended up empty, and `execute_rule_matching` skipped filtering altogether — so the rule derived everything it was written to restrict. Same failure mode, different route, and it slipped past the evaluation-side guarantee entirely. A wrong-length array and a non-string expression are now errors (with no `None` left, the signature simplifies to `Result<RuleFilter>`), and a near-miss on the keyword like `["FILTER", …]` is rejected. Any *other* unrecognized where-clause element keeps its current tolerance but warns instead of vanishing — turning that into a hard error is a language decision I didn't want to make inside this fix.

An IRI operand that resolves to no registered namespace is rejected at parse time rather than demoted to a string, since that demotion *is* the fail-open path — same thing the SPARQL rule lowerer already does at `sparql_lang.rs:203`, so I followed that rather than inventing a second convention.

## ⚠️ Behavior changes for existing rules

Two, and the first is the one most likely to be sitting in a deployed ledger right now.

**1. A bare unquoted filter operand is rejected at parse time.** `(= ?p knows)` — the bare form — was **the only operand form that ever matched a bound IRI** on `main`, so anyone who hit this bug and worked around it wrote exactly this shape. The tempting halfway house is to read it as a string literal and let a diagnostic catch the stale rules, but that fails invisibly in *both* directions: `=` eliminates every row (a string never equals an IRI), which a zero-derivation gate can see — and `!=` keeps **every** row, so `(!= ?prop ssn)` on the copy-properties rule copies `ex:ssn` right back out, and no run-time gate can see it, because every surviving row looks exactly like success. The parser is the one place that still knows the discriminator (quoted vs unquoted), so it refuses to guess: the error names the operand and both rewrites — quote it (`"knows"`) to compare a string, prefix it (`ex:knows`) to compare an IRI. Rejection warn-and-skips the one rule like any other rule parse failure; sibling rules keep deriving. The bare string form was never documented, nothing in the repo's own corpus ever used it legitimately, and the one population known to use it is the pre-fix workaround rules this exists to catch. Bare *numeric* and *boolean* operands (`(> ?age 20)`, `(= ?flag true)`) classify before the rejection can see them and are untouched; both are pinned by tests, as are the `=` and `!=` directions of the rejection itself.

To keep the *quoted* spelling of the same mistake from being a silent loss, a rule whose filters eliminate every binding row while comparing an IRI against a literal — `(= ?p "ex:knows")`, or a string-bound variable against an IRI-bound one — warns with the fix named. That signal needed building deliberately: once IRI-vs-literal became a clean `False` per §17.4.1.7, it stopped being an `Error`, so the existing "could not compare its operands" warn could no longer see it. It stays quiet when such a filter legitimately keeps rows — `(!= ?val "Bob")` over IRI-valued properties is correct usage and stays silent — and there's a test for each direction.

**2. An unresolvable IRI operand now refuses to parse** where it previously ran with a no-op filter. Narrower than it reads: both call sites warn-and-skip only the offending rule rather than failing the rule set (`:91` stored, `:1545` query-time), sibling rules keep deriving, and `encode_iri_strict` fails only on an unregistered *namespace prefix* — not an unused IRI — so `ex:ssn` resolves whenever any `ex:` IRI exists on the ledger. The error names the operand and all three remedies. Still a loss where a rule previously derived useful facts alongside a harmless broken filter, hence the callout.

`docs/query/datalog-rules.md` documents both, plus the operand classification table and the insert-variable requirement below. It also corrects the operator table, which advertised logical, arithmetic and string functions that a JSON-LD rule filter has never actually parsed.

## #1560 — a rule that can't derive anything now says so

Separate and much smaller, but it's the observability lever that makes the above debuggable, so it rides along. A rule whose `insert` references a variable `where` never binds derived nothing and said nothing — `instantiate_pattern` returns `None` per row, the rows get dropped, and the only signal was a per-iteration debug counter that attributes nothing to any particular rule. A where/insert typo looked exactly like a where clause that matched nothing.

Two nets, both using the diagnostic paths this module already has rather than a new one:

- **Parse time** — enforce datalog's range restriction (every `insert` variable must appear in `where`), judged **per insert pattern** to match the runtime semantic: `execute_rule_with_bindings` instantiates each head independently, so a two-head rule with one typo'd head keeps deriving from the good head exactly as it always has — the bad head is skipped with a warning naming the variable and how many heads were kept. Only a rule in which *no* head can ever instantiate — which can't derive anything by construction — is rejected outright, with a message naming the offending variables, flowing through the existing `tracing::warn!(?rule_id, %e, ...)` skip path so one bad rule still can't stop the others. Parser-invented variables get their own wording, since quoting `?__implicit_0` back at an author who never typed it explains nothing; that message points at the missing `@id` instead.
- **Run time** — warn when a rule matched binding rows but instantiated zero flakes. That's precisely "the where clause matched but the insert couldn't instantiate," and it catches what the parse check structurally can't see: SPARQL-lowered rules, and a subject/predicate variable that only turns out to be literal-bound once it binds. Recorded per rule so the fixpoint's repeated rounds don't repeat the warning.

Per the issue's second acceptance criterion, a rule whose where clause simply matched nothing stays silent — that's ordinary and says nothing about soundness. There's a test pinning the no-noise half specifically.

## On #1558 — looked at it, deliberately not included

It does share the "fail loud at parse time instead of minting a bogus `Sid(EMPTY, …)`" seam, and its Gap 3 (JSON-LD keywords becoming fabricated predicates) would genuinely be a few lines on top of what's here. But I don't think it's a small addition, for a reason that only showed up while looking:

Gap 3 done naively would reject the standard typed-literal form `{"@value": 65, "@type": "xsd:integer"}` in a rule pattern. `parse_object_value:679` routes a non-`@id` object map to `parse_node_pattern` as a nested anonymous node, and `parse_node_pattern:584` skips only `@id`/`@context`/`@type` — so `@value` arrives as a *key*, even though `parse_term:717` handles that form directly. Closing Gap 3 correctly therefore also means fixing how node-pattern object values are routed. **That routing bug is a real latent defect that #1558's own text doesn't currently describe**, so it's worth adding to the issue before someone picks it up and scopes only what's written there. Gaps 1 and 2 (bare context-mapped keys, bare `@type` values) are context term-definition expansion, different again.

Three sub-gaps across a different mechanism with its own regression matrix. I'd rather it got its own PR than be smuggled into this one — happy to pick it up next if that's the right call. No `Fixes #1558` here.

## Tests & gates

Sixteen new integration tests in `it_datalog_rules.rs` and nineteen new unit tests in `datalog_rules.rs` (twenty-seven in the module all told). The behavioral ones were confirmed red before their fix — including the `(!= ?prop ssn)` leak repro and the two-head-rule derivation, both red at the previous head of this branch — and the two narrowest arms have their own mutants: reverting just the RDFterm-equal arm turns the IRI-valued-row test red, and restoring the silent `Ok(None)` drop turns the malformed-filter test red.

- `cargo test -p fluree-db-query` — 1,440 lib + all targets green
- `cargo test -p fluree-db-reasoner` — 86 green
- `cargo test -p fluree-db-api` — `grp_reasoning` 98, `grp_query_reason` 100, `grp_policy` 96, `grp_query` 422, all green
- `cargo clippy -p fluree-db-query -p fluree-db-api --all-targets --no-deps` — clean
- `cargo fmt --all --check` — clean

Fixes #1556
Fixes #1560

### DIFF
--- a/docs/query/datalog-rules.md
+++ b/docs/query/datalog-rules.md
@@ -89,6 +89,45 @@ ledger — see the reasoning budget under [Fixpoint evaluation](#fixpoint-evalua
 ]
 ```
 
+Filters can also compare against an **IRI**, which is how you constrain a
+variable-position predicate or an entity-valued object:
+
+```json
+"where": [
+  {"@id": "?s", "ex:sameAs": {"@id": "?other"}},
+  {"@id": "?other", "?prop": "?val"},
+  ["filter", "(!= ?prop ex:ssn)"]
+]
+```
+
+How a filter operand is classified:
+
+| Operand | Read as |
+|---------|---------|
+| `?name` | Variable |
+| `62`, `1.5`, `true` | Number / boolean literal |
+| `ex:ssn`, `http://example.org/ssn` | **IRI** — expanded via the rule's `@context` and resolved against the ledger |
+| `"senior"`, `"John Smith"` | String literal (quote it — quoted operands may contain spaces) |
+| `senior` | String literal (bare, unquoted) |
+
+Three rules worth knowing:
+
+- **IRI comparison is namespace-aware.** `(= ?p ex:knows)` matches `ex:knows`
+  and not `foaf:knows`. Only `=` and `!=` are defined for IRIs; ordering
+  operators against an IRI are rejected.
+- **A bare name is a string, not an IRI.** `(= ?p knows)` compares against the
+  *string* `"knows"` and will never match the IRI `ex:knows`. Write the
+  prefixed or absolute form instead. (Before Fluree resolved IRIs in filters,
+  the bare form was the only one that appeared to work — but it matched
+  namespace-blindly, so `ex:knows`, `foaf:knows` and any other `knows` were
+  treated as equal. A rule still using the bare form now derives nothing and
+  logs a warning naming the operand.)
+- **An unresolvable IRI operand is an error, not a fallback.** If a filter
+  names a prefix the rule's `@context` does not define, or a namespace the
+  ledger has never seen, the rule is rejected and skipped with a logged error
+  rather than quietly comparing the operand as a string. Quote the operand if
+  you did mean a literal.
+
 ### Insert clause
 
 The `insert` clause defines what facts to produce for each set of matching
@@ -105,6 +144,19 @@ variable bindings.
 - Use `{"@id": "?var"}` for IRI/entity values; use `"?var"` directly for
   literal values.
 - Multiple triples can be generated from a single insert pattern.
+- **Every variable used in `insert` must also appear in `where`.** A variable
+  the `where` clause never binds cannot produce a fact, so the rule is rejected
+  at parse time with an error naming the variable rather than running and
+  deriving nothing. A `where`/`insert` typo is the usual cause:
+
+  ```json
+  "where":  {"@id": "?s", "ex:relType": {"@id": "?relation"}},
+  "insert": {"@id": "?s", "?rel": {"@id": "?s"}}
+  ```
+
+  `?rel` is never bound — the `where` clause binds `?relation`.
+- Every node in an `insert` pattern needs an `@id`. An anonymous node has no
+  subject to derive facts about, and is reported the same way.
 
 ## Providing rules
 
@@ -381,21 +433,40 @@ Filters use S-expression syntax within the `where` array:
 
 ### Available operators
 
+A JSON-LD rule filter is a single comparison between two operands:
+
 | Category | Operators |
 |----------|-----------|
-| Comparison | `=`, `!=`, `<`, `>`, `<=`, `>=` |
-| Logical | `and`, `or`, `not` |
-| Arithmetic | `+`, `-`, `*`, `/` |
-| String | `str`, `strlen`, `contains`, `strstarts`, `strends` |
-| Type checking | `isIRI`, `isBlank`, `isLiteral`, `bound` |
+| Comparison | `=`, `!=` (also `not=`), `<`, `>`, `<=`, `>=` |
+
+Logical combinators (`and` / `or` / `not`), arithmetic, and string or
+type-checking functions are **not** available in JSON-LD rule filters — an
+unrecognized operator is a parse error and the rule is skipped. Use several
+`["filter", ...]` entries in the `where` array to require more than one
+condition; they are combined with AND. For richer expressions, write the rule
+in SPARQL (see [SPARQL rules](#sparql-rules)), where `FILTER(… && …)` and
+`FILTER(!(…))` are supported.
+
+Ordering operators (`<`, `<=`, `>`, `>=`) are defined for numbers and strings
+only; using one against an IRI operand is rejected.
 
 ### Examples
 
 ```json
 ["filter", "(> ?age 21)"]
-["filter", "(and (>= ?age 18) (< ?age 65))"]
-["filter", "(contains ?name \"Smith\")"]
 ["filter", "(!= ?person ?other)"]
+["filter", "(!= ?prop ex:ssn)"]
+["filter", "(= ?name \"John Smith\")"]
+```
+
+Two conditions, ANDed:
+
+```json
+"where": [
+  {"@id": "?person", "ex:age": "?age"},
+  ["filter", "(>= ?age 18)"],
+  ["filter", "(< ?age 65)"]
+]
 ```
 
 ## Performance considerations
--- a/docs/query/datalog-rules.md
+++ b/docs/query/datalog-rules.md
@@ -108,20 +108,24 @@ How a filter operand is classified:
 | `62`, `1.5`, `true` | Number / boolean literal |
 | `ex:ssn`, `http://example.org/ssn` | **IRI** — expanded via the rule's `@context` and resolved against the ledger |
 | `"senior"`, `"John Smith"` | String literal (quote it — quoted operands may contain spaces) |
-| `senior` | String literal (bare, unquoted) |
+| `senior` | **Rejected** — a bare unquoted token is ambiguous; quote it for a string, prefix it for an IRI |
 
 Three rules worth knowing:
 
 - **IRI comparison is namespace-aware.** `(= ?p ex:knows)` matches `ex:knows`
   and not `foaf:knows`. Only `=` and `!=` are defined for IRIs; ordering
   operators against an IRI are rejected.
-- **A bare name is a string, not an IRI.** `(= ?p knows)` compares against the
-  *string* `"knows"` and will never match the IRI `ex:knows`. Write the
-  prefixed or absolute form instead. (Before Fluree resolved IRIs in filters,
-  the bare form was the only one that appeared to work — but it matched
-  namespace-blindly, so `ex:knows`, `foaf:knows` and any other `knows` were
-  treated as equal. A rule still using the bare form now derives nothing and
-  logs a warning naming the operand.)
+- **A bare name is rejected, not guessed.** `(= ?p knows)` is a parse error
+  naming the operand: write `"knows"` (quoted) to compare against the string,
+  or `ex:knows` to compare against the IRI. A bare token cannot be safely read
+  as a string, because against an IRI-bound variable a string comparison fails
+  invisibly in both directions — `=` derives nothing, and `!=` keeps every row,
+  so an exclusion filter derives exactly the facts it was written to exclude.
+  (Before Fluree resolved IRIs in filters, the bare form was the only one that
+  appeared to work — but it matched namespace-blindly, so `ex:knows`,
+  `foaf:knows` and any other `knows` were treated as equal. A rule still using
+  the bare form is now rejected and skipped with a logged error naming the
+  operand and both rewrites.)
 - **An unresolvable IRI operand is an error, not a fallback.** If a filter
   names a prefix the rule's `@context` does not define, or a namespace the
   ledger has never seen, the rule is rejected and skipped with a logged error
@@ -145,9 +149,14 @@ variable bindings.
   literal values.
 - Multiple triples can be generated from a single insert pattern.
 - **Every variable used in `insert` must also appear in `where`.** A variable
-  the `where` clause never binds cannot produce a fact, so the rule is rejected
-  at parse time with an error naming the variable rather than running and
-  deriving nothing. A `where`/`insert` typo is the usual cause:
+  the `where` clause never binds cannot produce a fact. Each insert pattern is
+  checked independently: a pattern that references an unbound variable is
+  **skipped with a logged warning naming the variable**, while the rule's other
+  insert patterns keep deriving — so one typo'd head in a multi-head rule does
+  not silence the rest. Only a rule in which **no** insert pattern can ever
+  produce a fact is rejected outright at parse time (with the same
+  variable-naming error), rather than running and deriving nothing. A
+  `where`/`insert` typo is the usual cause:
 
   ```json
   "where":  {"@id": "?s", "ex:relType": {"@id": "?relation"}},
--- a/fluree-db-query/src/datalog_rules.rs
+++ b/fluree-db-query/src/datalog_rules.rs
@@ -283,14 +283,32 @@ fn parse_where_clause(
                     }
                     JsonValue::Array(filter_arr) if is_filter_expression(filter_arr) => {
                         // Filter expression like ["filter", "(>= ?age 62)"]
-                        if let Some(filter) =
-                            parse_filter_expression(filter_arr, context, snapshot)?
-                        {
-                            filters.push(filter);
-                        }
+                        filters.push(parse_filter_expression(filter_arr, context, snapshot)?);
                     }
+                    // A near-miss on the filter keyword — `["FILTER", ...]` —
+                    // is rejected outright. It is unambiguously a filter the
+                    // author meant to apply, and dropping it silently is the
+                    // #1556 failure mode by another route: the filter
+                    // disappears, `rule.filters` is empty, and the rule derives
+                    // everything it was written to restrict.
+                    _ if looks_like_misspelled_filter(item) => {
+                        return Err(QueryError::InvalidQuery(format!(
+                            "Unrecognized element in rule where clause: {item}. \
+                             The filter keyword is lowercase `filter` — this element \
+                             would otherwise be dropped, leaving the rule to derive \
+                             the facts the filter was written to exclude."
+                        )));
+                    }
+                    // Anything else stays tolerated, as it has been. Whether an
+                    // unknown where-clause element should be an outright error
+                    // is a language decision beyond this fix, but it should not
+                    // vanish without a trace.
                     _ => {
-                        // Skip unknown array elements
+                        tracing::warn!(
+                            element = %item,
+                            "unrecognized element in datalog rule where clause — ignored; \
+                             expected a node pattern object or [\"filter\", \"(op ?var value)\"]"
+                        );
                     }
                 }
             }
@@ -311,6 +329,14 @@ fn parse_where_clause(
 /// the head must appear in the body. A rule that breaks it can never derive
 /// anything — `instantiate_pattern` returns `None` for each row — which is the
 /// silent dead end reported in #1560.
+///
+/// Parser-generated names on the where side are deliberately NOT treated as
+/// binding. They are numbered off each clause's own `patterns.len()`, so a
+/// where-side `?__implicit_0` and an unrelated insert-side `?__implicit_0`
+/// would otherwise collide and suppress the diagnostic for exactly the
+/// anonymous-node case it exists to report. An author cannot write one of
+/// these names, so an insert pattern can never legitimately be referring to
+/// the where clause's copy.
 fn unbound_insert_variables(
     where_patterns: &[RuleTriplePattern],
     insert_patterns: &[RuleTriplePattern],
@@ -319,6 +345,7 @@ fn unbound_insert_variables(
     for pattern in where_patterns {
         bind_pattern_vars(pattern, &mut bound);
     }
+    bound.retain(|v| !v.starts_with(IMPLICIT_VAR_PREFIX));
 
     let mut unbound: Vec<Arc<str>> = Vec::new();
     for pattern in insert_patterns {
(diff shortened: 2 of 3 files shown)


fluree-db-query/src/datalog_rules.rs
```diff
@@ -704,37 +1003,163 @@ pub async fn execute_rule_matching(
...
+            // is how a mis-typed filter looks exactly like a rule that
+            // legitimately matched nothing.
+            tracing::warn!(
+                rule = rule.name.as_deref().unwrap_or("<unnamed>"),
+                rule_id = %rule.id,
+                errored_rows = diagnostics.errored_rows,
+                "datalog rule filter could not compare its operands on some binding \
+                 rows; those rows were excluded — check the filter's operand types"
+            );
+        } else if rows_before > 0 && binding_rows.is_empty() && diagnostics.iri_vs_literal > 0 {
```
bplatz, line 1120: This gate can only ever see the `=` direction, which is the harmless one. `!=` against a bare local name keeps every row, so it never reaches `binding_rows.is_empty()`.

Probed on this branch — `["filter", "(!= ?prop ssn)"]` on the copy-properties rule still derives `ex:alice ex:ssn "123-45-6789"`, with zero warnings. That's this PR's headline failure mode, via the operand form the description says people used as the workaround.

The obvious fix (warn on any `iri_vs_literal`) is wrong — `datalog_iri_versus_literal_filter_that_keeps_rows_is_not_flagged` correctly pins `(!= ?val "Bob")` as legitimate. The real discriminator is quoted vs unquoted, which `parse_filter_term` knows and then discards: both land as `RuleValue::String`. Carrying a "came from a bare unquoted token" bit would separate them and cover both directions.

fluree-db-query/src/datalog_rules.rs
```diff
@@ -230,6 +236,13 @@ fn parse_rule_definition(
         .ok_or_else(|| QueryError::InvalidQuery("Rule missing 'insert' clause".to_string()))?;
     let insert_patterns = parse_insert_patterns(insert_json, &context, snapshot)?;
 
+    // Reject an insert pattern that references a variable nothing can bind
+    // (#1560). Left alone it is invisible: every binding row is dropped inside
+    // `instantiate_pattern`, so the rule "runs" and derives nothing.
+    if let Some(message) = unsafe_insert_variables_message(&where_patterns, &insert_patterns) {
```
bplatz: This is whole-rule fatal, but the runtime behaviour it replaces was per-pattern: `execute_rule_with_bindings` instantiates each insert pattern independently, so a `None` dropped only that pattern.

A two-head rule with a typo in one head derived `ex:bob` from the good head on main; on this branch it derives nothing (confirmed both ways). Failing loud may well be the right call, but it's a third behaviour change that isn't in the warning section or the docs — and the loss is warn-only, so a deployed multi-head rule goes dark quietly.

docs/query/datalog-rules.md
```diff
@@ -89,6 +89,45 @@ ledger — see the reasoning budget under [Fixpoint evaluation](#fixpoint-evalua
...
+- **IRI comparison is namespace-aware.** `(= ?p ex:knows)` matches `ex:knows`
+  and not `foaf:knows`. Only `=` and `!=` are defined for IRIs; ordering
+  operators against an IRI are rejected.
+- **A bare name is a string, not an IRI.** `(= ?p knows)` compares against the
+  *string* `"knows"` and will never match the IRI `ex:knows`. Write the
+  prefixed or absolute form instead. (Before Fluree resolved IRIs in filters,
+  the bare form was the only one that appeared to work — but it matched
+  namespace-blindly, so `ex:knows`, `foaf:knows` and any other `knows` were
+  treated as equal. A rule still using the bare form now derives nothing and
+  logs a warning naming the operand.)
```
bplatz: Only true for `=`. With `!=` the bare form derives *everything* and logs nothing — see the comment on the diagnostic gate.

fluree-db-query/src/datalog_rules.rs
```diff
@@ -339,38 +491,185 @@ fn parse_filter_expression(arr: &[JsonValue]) -> Result<Option<RuleFilter>> {
...
+        return Ok(RuleTerm::Value(RuleValue::Double(f)));
+    }
+    if s == "true" {
+        return Ok(RuleTerm::Value(RuleValue::Boolean(true)));
+    }
+    if s == "false" {
+        return Ok(RuleTerm::Value(RuleValue::Boolean(false)));
     }
+
+    let absolute = s.starts_with("http://") || s.starts_with("https://");
```
bplatz: Nit: dead — `curie_prefix("http://example.org/x")` already returns `Some("http")`.

bplatz (changes requested):
Verified the bug is real — reverting `datalog_rules.rs` to the merge base with the tests kept takes `grp_reasoning` from 94 pass to 85 pass / 9 FAILED, headline test included. fmt + clippy clean, CI green. The RDFterm-equal modelling and the three-valued `Error` that refuses to collapse under `Not` are both right.

Two things before merge, one of which leaves the headline scenario reachable. Details inline.

aaj3f:
Thanks @bplatz — both blockers were exactly right, and I reproduced both of your probes verbatim before changing anything: `["filter", "(!= ?prop ssn)"]` derived the SSN onto `ex:alice` with zero relevant warnings at the old head, and your two-head rule derived nothing where main derived from the good head.

**On the `!=` leak — I took your discriminator and went one step further, to parse-time rejection of bare unquoted operands.** Your read was the key: quoted-vs-unquoted is what `parse_filter_term` knows and discards. Before deciding between carrying it as a warning bit versus rejecting outright, I tallied every filter operand in the repo — the only bare alphabetic string operand in the entire corpus was this PR's own stale-workaround test, and the merge-base docs never defined a bare form (the "bare = string" table row was this branch's own addition). Given that, a warning buys nothing over rejection — the rule boundary is already warn-and-skip at both call sites, so the visibility is identical — while warn-only would leave the leak itself intact: the SSN still copies, plus a log line nobody reads. Rejection makes it impossible, covers `=` and `!=` symmetrically, and costs nothing per-row since it's parse-time only. The error names both rewrites (`"knows"` quoted for a string, `ex:knows` for an IRI). Your constraints held: quoted `(!= ?val "Bob")` stays silent — that pin is untouched and green — and bare numerics/booleans (`(> ?age 20)`, `(= ?flag true)`) classify before the rejection, now pinned by their own test. The empty-set gate you probed stays, re-commented for its real remaining audience — quoted CURIE-lookalikes and string-vs-IRI variable pairs — with its own test, since it was about to become untested.

**On the multi-head change — per-pattern restored, exactly as you steered.** `parse_rule_definition` now partitions insert heads by instantiability: a bad head is skipped with a warning naming the variable (plus skipped/kept counts), and only a rule with *no* instantiable head is rejected — that rejection *is* the #1560 fix and keeps its named-diagnostic pin. Your repro is pinned both ways. I looked for an argument that whole-rule-fatal was better and didn't find one; it was never this PR's mandate and its loss was warn-only, which is the worst combination.

The docs row at `:124` now states the actual behavior for both directions, and the `:647` check was confirmed dead rather than wrong-intent — `curie_prefix` already returns `Some("http")` for absolute IRIs and both branches led to the same expand+resolve path — so it's removed with a note that CURIE classification subsumes them.

The full matrix is pinned — bare/quoted × `=`/`!=` × derives/derives-nothing, plus partial multi-head and all-heads-bad — and SPARQL-lowered rules can't produce bare tokens (strings are always quoted there), so the class is closed across surfaces rather than one cell.

bplatz (approved):
👍

PR #1714, opened 2026-08-26
wasm32 groundwork: the engine stack compiles and runs on wasm32-unknown-unknown, native untouched

aaj3f:
This PR makes the whole query/api stack build — and actually run — on `wasm32-unknown-unknown`, while leaving the native build byte-identical everywhere we could manage it and provably-neutral everywhere we couldn't. It's the first of two PRs toward an in-browser Fluree: a "try Fluree in the browser" playground and, longer-term, a browser cache query peer that speaks the peer-proxy wire contract we already ship (CID-verified `GET /storage/objects/{cid}`, which turns the immutability of content-addressed blocks into a cache-forever story on the client). The second PR (browser crate + npm packaging + the server CORS/caching headers) is staged behind this one; nothing browser-facing ships here.

The headline evidence is executable, not a compile check: `wasm-probe/run-node.mjs` builds the probe cdylib with wasm-pack, instantiates it under Node, and drives the real end-to-end path — `FlureeBuilder::memory().build_memory()` → `create_ledger` → JSON-LD `insert` → SPARQL `query` — asserting the future's own `Ok` (exit 0 only on the positive marker, never on "didn't crash"). Dev and release profiles both pass. Size, for the curious: the full engine (r2rml, reasoner, BM25, and Cypher still linked in) is 6.7 MB after `wasm-opt -Oz`, 1.8 MB brotli over the wire.

## What's actually in here

It's over-scoped on purpose — we'd rather review one coherent claim than five stacked rebases. Four layers:

**1. Cargo/target surgery.** The wasm-hostile deps (`tokio` full, `memmap2`, `fluree-db-spatial`'s s2 stack, `fluree-db-indexer`, `dirs`, `sysinfo`) move behind `[target.'cfg(...)']` splits; wasm builds get tokio's wasm-supported feature subset only. The recurring native trap was `default = ["native"]` cascading through dependency defaults — several deps now say `default-features = false` with the native feature re-enabled through `fluree-db-api`'s `native` cascade, so native default builds resolve exactly as before (verified via `cargo tree` diffs where it seemed worth being paranoid, e.g. chrono's `wasmbind`). Two genuinely shippable fixes fell out regardless of wasm: `fluree-db-sparql --no-default-features` compiles again (`fluree-vocab` was optional-but-referenced — burn-down ROADMAP §5 item 9), and `import_sink` now sits behind the `import` feature with `fluree-db-indexer` optional, so non-import consumers of transact stop building the indexer entirely. A handful of dead deps (spatial's tokio, query's memmap2, binary-index's fs2) are gone.

**2. Runtime shims — compiling was not the hard part; not panicking was.** `wasm32-unknown-unknown` panics at runtime on `Instant::now`/`SystemTime::now`, has no threads, and `std::fs` returns `Unsupported`. So: `fluree-db-core::clock` re-exports `std::time` types on native (zero-cost alias, no wrapper) and web-time on wasm; chrono gets `wasmbind` (native-inert, tree-diffed); a `spawn_detached` seam is `#[inline] tokio::spawn` on native and `spawn_local` on wasm; moka — which has no wasm clock and panics at cache *construction* — stays untouched on native and swaps to an lru-backed twin behind the existing wrapper seams on wasm (all in-scope caches are TTL-free, and moka drops out of the wasm artifact entirely); and `disk_cache` treats `Unsupported` reads as a miss rather than an error, so CAS consumers fall through to fetch instead of poisoning the query (`try_read_cached_bytes` previously special-cased only `NotFound`). The full hazard audit with per-site verdicts is in `wasm-tier2/C-runtime-hazards.md` — it doubles as the review map for this layer.

**3. Read-path groundwork ("async residency, sync compute").** The engine's five CAS-bridge chokepoints (`run_sync_on_runtime` and friends in `fluree-db-binary-index`) can't exist on single-threaded wasm, so misses there are now reported through a **store-level miss register** (`fluree_db_core::storage::residency`): the failing arm records `Want { cid, kind }` into the register — deduped, and *multi-want* where the caller knows its routed set (a cursor leaf-open miss registers the scan's whole remaining leaf set; the batched join probes pre-register every non-resident leaf across their chunk ranges) — before surfacing an error. A retry driver (`fetch_wants` / `RetryBudget::after_error`) drains the register, fetches concurrently into a pinning store, and re-enters; for scans the retry happens *in-frame* (the two cursor frames in `binary_scan.rs` re-enter the same cursor in place — no whole-query re-run, no wasted decode), with whole-unit re-run kept only as the backstop for one-shot paths (fast paths, policy sub-queries). Termination is progress-based (the pinned set grows monotonically against a finite want set), with a sanity cap guarding contract violations only. We deliberately did *not* make the error itself the channel: the engine stringifies `io::Error` at dozens of sites (`QueryError::Internal(String)`), so a payload-carrying error is unreliable by construction — a typed `QueryError::NeedFetch` variant exists as belt-and-braces (mirroring the fuel-exceeded precedent, with the fast-path sites routed through `from_io`), but the register is the contract. Residency itself is `StorageRead::resolve_cached_bytes(&ContentId) -> Option<Arc<[u8]>>` — default `None`, forwarded through `StorageContentStore` and `BranchedContentStore`, an `Arc` clone per hit on wasm. The cfg story: production native builds compile the pre-residency read path (the arms are behind `cfg(any(target_arch = "wasm32", feature = "residency"))` and additionally store-driven at runtime) — *semantically* identical, with one inlined delegation, since `open_leaf_handle` now splits into a wrapper over `open_leaf_handle_native`; "byte-identical" was one word too strong for a claim this load-bearing; the dev-only `residency` feature exists so native CI exercises everything — `it_residency_retry` runs four real-query recovery tests (operator scan, fast-path COUNT, dir-walk, policy `f:query` sub-query) differentially against a plain native instance over the same storage, with positive miss markers and three mutation checks. CI's `--all-features` clippy/nextest jobs build and run all of it — no browser required. Along the way we found and fixed two real resume hazards that exist independent of wasm: `BinaryCursor` advanced its leaf index *before* the fallible open (a caught-and-retried error would silently skip a leaf), and the leaflet-level equivalent was worse — the open leaf is `take()`n for the leaflet loop, so an error mid-loop dropped the remaining leaflets of that leaf. Both fixes are mutation-tested: `cursor_retries_failed_cas_reads_without_losing_rows` replays a remote scan against a fail-exactly-call-N store for every N, asserting exactly one surfaced error and zero lost rows; reverting either fix fails the sweep at the expected call position.

**4. Toolchain + CI.** zstd-sys needs a wasm32-capable clang (Apple's has no WebAssembly backend), so `.cargo/config.toml` routes the wasm target through `scripts/wasm-clang.sh`/`wasm-ar.sh` (Homebrew LLVM on macOS, system clang on Linux/CI). A new `wasm32` CI job checks the full engine stack plus the no-default-features parse layer on the pinned toolchain — the door this PR opens stays open. One subtlety worth a reviewer's eye: an environment-level `RUSTFLAGS` overrides target-scoped config rustflags, so the CI job restates the `getrandom_backend="wasm_js"` cfg explicitly rather than trusting the config file.

## The native-neutrality case

This was the prime constraint (speed *and* memory), and it's argued three ways. Structurally: every wasm behavior lives behind `cfg(target_arch = "wasm32")` or a target-split dep; native branches are the previous code, and the shims that do touch shared lines are zero-cost re-exports or `#[inline]` one-liners. The one deliberate shared-code touch is a factored-out `success_or_size_error` helper in the multi-query dispatcher (once-per-subquery, not hot). Empirically: the native suites are green at this HEAD — workspace `check --all-targets`, core (803), binary-index (383, including the new mutation-tested sweep), query (1416+), transact/novelty/nameservice-sync (89 incl. wiremock), api (774 lib + all `grp_*` targets), server. And going forward: the nightly bench baseline gate watches the same paths; we'd suggest a `workflow_dispatch` bench compare on this branch as an advisory pre-merge check. Locally, a criterion A/B on `fluree-db-api`'s `query_hot_bsbm` and `transact_commit` benches (merge-base vs. this branch, run back-to-back so both see the same thermal state; 10 samples each) comes out indistinguishable: `q3` −3.1% (p=0.39), `q5` −0.4% (p=0.30), `q9` −1.0% (p=0.01 — the one statistically significant point, and it's one percent), `transact_commit` fresh −0.4% (p=0.18), populated +2.5% (p=0.13), where negative means the base was faster. Fwiw, a first naive pass right after the full release build showed a uniform "+120%" that vanished on re-run of the identical branch binary — the box, not the code — which is a good reminder that the shared-runner compare being advisory is the right call. Two transparency notes on top of that: this PR touches perf-relevant paths, so the new `bench-compare` CI gate will run and annotate both TIME and PEAK_MEM against the committed baseline — I'd treat its output as part of this review; and I did *not* run a dedicated memory-profile A/B locally (the structural argument for memory neutrality is the same cfg story as speed — native allocates nothing new — but the tracking-allocator numbers from CI/nightly are the honest check, and I'll watch them).

**On the `bench-compare` annotations (please read before reacting to them).** This PR's run posts advisory wall-clock regressions — `query_hot_bsbm/q3` +54.77%, `groupby_overlay` +100.60%, a dozen more in that range. They are not this PR. The committed baseline was captured with `host_class=local` and CI runs `ci-ubuntu-latest`, which `ci.yml`'s own comments call a refusal-by-default comparison — that mismatch is exactly why every one of those lines says `advisory` and the job exits 0. The check: **main's own runs against the same baseline show the same annotations, and larger ones** — main at `8b6a0454a` reports q3 +81.44% / groupby_overlay +117.77%, and at `5d3afe284` q3 +94.77% / groupby_overlay +110.79%, versus this branch's +54.77% / +100.60%. We are *below* main on the same yardstick; the spread is shared-runner noise against a stale local baseline. Meanwhile `peak_mem`, which does compare meaningfully across host classes, is flat to four decimal places on every scenario (−1.85% to +0.08%, most exactly 0.00%) — that is the memory-neutrality evidence, and it is clean. The way to make these lines meaningful again is to capture a `ci-ubuntu-latest` baseline via bench.yml's `workflow_dispatch` and commit it, at which point both metrics start enforcing automatically; that is a repo-level task this PR deliberately doesn't smuggle in.

One rebase-reconciliation note: this branch is rebased over `bf63e54e6` ("gate rand behind native"), which independently fixed a wasm32 build break in `fluree-db-core` and observed that "nothing in CI builds it" for the wasm target — the `wasm32` CI job in this PR is exactly that missing enforcement, so these two changes compose rather than collide (the optional-`rand` gating and the target-scoped JS-randomness unification serve different layers of the same problem).

## Decisions that warrant scrutiny (and the alternatives they beat)

These are the calls I'd want a second opinion on if someone else had made them — each with what we didn't do and why:

- **`PolicyQueryExecutor` is now `Send + Sync` on all targets.** Alternative: keep the pre-existing cfg'd `?Send` wasm twin. Rejected because the engine's operator futures are uniformly Send-bounded, so the `?Send` variant could never actually carry a policy executor through the engine — it predates that design and had become the blocker rather than the enabler. Single-threaded wasm implementors satisfy the bounds trivially; a future JS-backed executor should bridge via channels rather than infect the engine with `?Send`.
- **The miss channel is a store-side register, not an error payload.** Alternative (and our own first implementation): carry `NeedFetch` inside `io::Error` and downcast above the operator boundary. Our internal adversarial review killed it with evidence — `QueryError` has no io-carrying variant and the engine stringifies at ~dozens of sites, so the payload provably never survives a real path. The register is immune to stringification and also enables multi-want reporting (a round learns a scan's whole routed leaf set, not one CID). The typed `QueryError::NeedFetch` variant still exists as belt-and-braces on the frames that do preserve io errors.
- **`disk_cache` compiles on wasm with a zero byte-budget** (writes disabled, reads miss) instead of being cfg'd out. Keeps one code path for CAS consumers; if you'd rather see it excluded wholesale on wasm, happy to talk that through.
- **Clocks: `web-time` re-export shim, not stubbed-out timing.** Alternative: cfg the instrumentation away on wasm. Rejected — the query timing fields are wanted in a browser too (performance.now works there), and on native the shim is `pub use std::time::{Instant, SystemTime}`, a zero-cost alias with no codegen change.
- **moka's wasm stand-in is `lru` + an uncontended `parking_lot` Mutex**, not `quick_cache`. No new workspace dep, and on a single-threaded target an uncontended lock beats sharding. The trade accepted: `get_with` is check-compute-insert, not single-flight — correct single-threaded, and it makes re-entrant use inside a decode closure deadlock-free. Native keeps real moka everywhere; the wasm artifact drops moka entirely.
- **The residency arms are `cfg(any(wasm32, feature = "residency"))` + runtime store-driven**, rather than wasm-only. The extra feature exists purely so native CI can execute the whole machinery (`--all-features` picks it up); production native builds compile the pre-residency path — semantically identical, one inlined delegation through the `open_leaf_handle` wrapper — and even with the feature on, the check is one predictable branch per *blob open* — never per row — engaged only by register-bearing stores.

## Honest caveats

- This PR makes wasm *run*, not *ship*: no JS API, no browser storage, no prefetch loop. A cache miss on a remote ledger under wasm errors with `NeedFetch` by design until PR-2's retry loop lands.
- One mid-assembly mistake is worth naming because it's in the history: the original spike commit replaced `.cargo/config.toml` wholesale and deleted the workspace `RUST_MIN_STACK = 8388608`, which made `proxy_integration`/`grp_graphsource` abort with stack overflows that looked exactly like code regressions. Restored in `fix(build): restore RUST_MIN_STACK…` with the mechanism documented; if you've ever wondered whether that env entry is load-bearing — it is.
- Commit hygiene: two probe support files (`wasm-probe/README.md`, `Cargo.toml` metadata) landed in the `chore: zero out wasm32-target warnings` commit rather than the harness commit due to a concurrent-assembly race; content is correct, attribution is smeared. Flagging rather than rewriting shared history.
- The design memos this work executed against (read-path design, runtime-hazard audit, worker/parallelism research, browser I/O contract) are deliberately kept out of the repo; happy to share them directly with anyone reviewing.

Approving reviewers may want to start with `fluree-db-core/src/clock.rs` + `wasm_cache.rs` (the shim map), then `fluree-db-binary-index/src/read/need_fetch.rs` and the cursor fixes in `binary_cursor.rs` — that's where the substance is; the Cargo diffs are wide but mechanical.

## Pre-review remediation (before undrafting)

Before asking anyone to spend review time here, we ran this PR through an internal blind pre-review (a reviewer with no context on the work, applying the repo's own review rubric, checking out the PR like any colleague's). Verdict was approve-with-notes (no critical or high findings). Every finding was then either fixed or rebutted with evidence — the remediation commits are on the branch, findings referenced in their messages. Highlights: the `MissRegister` now documents its one-draining-query-per-store scope; the wasm32 CI job gates warnings (`-D warnings`) so the zero-warning build can't rot silently; the node smoke asserts the inserted names actually round-trip in the FORMATTED result (the first fix of this fix asserted against Debug output, which renders encoded bindings — the smoke's own honest exit code caught it); `spawn_maintenance`/bm25 worker are native-gated; two stale comments corrected. The pre-review also independently re-applied the cursor mutation to confirm the test sweep catches it, and machine-diffed `cargo tree` to confirm native dependency-resolution neutrality.

One more found by CI rather than by us, worth stating plainly: the first CI run failed clippy on five workspace-denied `manual_assert` sites in the residency test helper. Local per-crate clippy runs had reported clean — a clippy invocation over an already-checked tree can be a fingerprint no-op that prints nothing and exits 0. Fixed, and re-verified by grepping the pattern to zero, forcing a re-check, and running the CI-shaped workspace invocation with `-D warnings`.

## Three corrections from review, and where a regression would show first

Raised in review, verified independently, and worth carrying in the description rather than only in the thread.

**The benches compile the `residency` build, not the production one.** `fluree-db-api`'s dev-dependencies enable `residency`, and under `resolver = "2"` that unifies into every target needing dev-deps — including `cargo bench -p fluree-db-api --bench query_hot_bsbm` (`ci.yml:227`, `bench.yml:101`) and the local A/B above. Measured:

```
cargo tree -p fluree-db-api -e features,no-dev  ->  no residency
cargo tree -p fluree-db-api -e features         ->  fluree-db-binary-index feature "residency"
                                                    fluree-db-query        feature "residency"
```

Two consequences, and both cut in directions worth stating plainly:

1. **The A/B is stronger than claimed, not weaker.** It measured a residency-*enabled* build against a base containing none of that code, and still landed in the noise. That is the conservative direction.
2. **No perf gate measures the production configuration.** "Production native builds compile the pre-residency read path" is true, and it rests on the cfg argument alone — which is sound, because cfg'd-out code is absent rather than predicted-away. But the numbers demonstrate something adjacent to that sentence rather than that sentence itself. Keeping the feature out of bench unification is fiddly and buys little; saying so exactly is better than implying the benches cover what ships.

That second point is also why this PR now caches the residency-tier probe (`residency_mode`) as a bool at construction instead of calling `dyn ContentStore::miss_register()` — six sites, up to ~3 per leaf open. It is a small win on wasm, and it removes dyn dispatch from the build the benches actually compile, narrowing the measured-vs-shipped gap rather than widening it.

**Where to look first if a regression ever appears.** `binary_cursor.rs:417` is the only unconditional native change on a scan path — the success path is unchanged and it is per-leaflet against a µs-scale column decode, so codegen-perturbation risk is low (CLAUDE.md's own note is that fat-bodied loops tolerate this fine). It is named here rather than leaving the diff to read as uniformly risky: that line is the first place to check, and the bench covering it runs the feature-on build.

## Caught up to `main` + adversarial-review remediation (2026-09)

Rebased onto current `main` (49 commits; semantically verified, not just textually — `cargo check --all-targets` after restack plus the wasm and residency suites at each level). A second independent adversarial review of the wasm read-path (the surface the earlier review deliberately set aside) then added, all verified by execution:

- **Bounded residency read-ahead + airtight retry termination.** One leaf-open miss registered the WHOLE remaining routed leaf set, so a `LIMIT`-shaped scan fetched the entire predicate run and — under a byte-budgeted resident tier — livelocked (re-fetching an evicted tail forever, stopped only by the sanity cap). Now records a bounded window, and `RetryBudget::after_error` gates progress on the specific blocking object becoming resident, so termination holds against any tier size. Mutation-checked (bounded-window count; a non-pinning store fails on round 1, not at the cap).
- **`create_dir_all` no longer aborts a wasm index load** (cherry-picked the `ensure_cache_dir` no-op-on-`Unsupported` fix down so this PR is self-consistent; it drops patch-identical when the fix's own commit rebases through).
- **The reasoner's `Instant::now()` — which traps on wasm — routes through `clock::Instant`.**
- **The wasm32 CI job now runs clippy, not just check**, which caught three latent wasm-only lints (a dropped future in the CAS bridge stub, a needless return in a storage arm, an `into_iter` loop in the sequential multi-query twin) that `check` and the host clippy job could never see.

bplatz's four inline items from the first review remain addressed as replied; these are additive on top.


fluree-db-api/Cargo.toml
```diff
@@ -140,6 +135,15 @@ fluree-bench-support = { path = "../fluree-bench-support" }
 fluree-bench-alloc = { path = "../fluree-bench-alloc" }
 rand = { workspace = true }
 fluree-db-nameservice-sync = { path = "../fluree-db-nameservice-sync" }
+# Residency-mode read arms for the wasm read-path recovery tests
+# (it_residency_retry): dev-only feature unification — production native
+# builds keep the feature off and compile the pre-residency read path.
+fluree-db-query = { path = "../fluree-db-query", features = ["residency"] }
```
bplatz, line 141: These dev-deps enable `residency`, and under `resolver = "2"` that unifies into every target needing dev-deps — including `cargo bench -p fluree-db-api --bench query_hot_bsbm` (`ci.yml:227`, `bench.yml:101`) and the local A/B in the description. Verified:

```
cargo tree -p fluree-db-api -e features,no-dev  ->  no residency
cargo tree -p fluree-db-api -e features         ->  fluree-db-binary-index feature "residency"
                                                    fluree-db-query        feature "residency"
```

(`cargo test -p fluree-db-query` alone stays off.)

Two consequences worth stating in the description:

1. The A/B measured the residency-enabled build against a base with none of that code, and still landed in the noise — the conservative direction, so the result is stronger than claimed.
2. No perf gate measures the production configuration. "Production native builds compile the byte-identical pre-residency read path" is true and rests on the cfg argument alone, which is sound — cfg'd-out code is absent, not predicted-away. The numbers demonstrate something adjacent to that sentence rather than that sentence.

Suggest saying both rather than changing the Cargo setup — keeping the feature out of bench unification is fiddly and not worth it.

fluree-db-binary-index/src/read/binary_index_store.rs
```diff
@@ -612,6 +650,52 @@ impl BinaryIndexStore {
...
+        self.get_leaf_bytes_shared_native(leaf_cid)
+    }
+
+    /// True when the content store carries a sync residency tier (it exposes
+    /// a miss register): sync reads are then served exclusively from
+    /// `resolve_cached_bytes` + `NeedFetch` misses, never the filesystem or
+    /// bridge tiers. Always true for browser stores; native stores opt in
+    /// (tests) by exposing a register.
+    #[cfg(any(target_arch = "wasm32", feature = "residency"))]
+    pub(crate) fn residency_mode(&self) -> bool {
```
bplatz, line 686: Suggest caching this as a `bool` field set at construction.

It's a vtable call through `dyn ContentStore`, and it's called at six sites (`:654`, `:733`, `:814`, `:1049`, `arena/vector.rs:706`, `:736`) — up to ~3 per leaf open. The CAS and its miss register are fixed for the store's lifetime, so the result can't change.

Second reason beyond the call itself: it removes dyn dispatch from the feature-on build, which is the build every bench measures per the Cargo comment. Narrows the gap between what's benchmarked and what ships.

fluree-db-binary-index/src/read/binary_index_store.rs
```diff
@@ -715,6 +809,45 @@ impl BinaryIndexStore {
         leaf_cid: &ContentId,
         sidecar_cid: Option<&ContentId>,
         need_replay: bool,
+    ) -> io::Result<Box<dyn super::leaf_access::LeafHandle>> {
+        #[cfg(any(target_arch = "wasm32", feature = "residency"))]
+        if self.residency_mode() {
```
bplatz, line 830: Wording: with the split into `open_leaf_handle` + `open_leaf_handle_native`, "byte-identical" in the description is one word too strong — native is now a delegation through a wrapper, almost certainly inlined but structurally one hop.

No code change suggested. "Semantically identical, one inlined delegation" would be accurate, and this claim carries enough weight in the PR to be worth stating exactly.

fluree-db-binary-index/src/read/binary_cursor.rs
```diff
@@ -380,65 +408,82 @@ impl BinaryCursor {
...
-                                    decode_set,
-                                },
-                            )?
+                    // The leaflet's fallible reads (column load + history
+                    // sidecar), isolated so a failure rewinds: the leaflet
+                    // index steps back and the leaf is restored, leaving the
+                    // cursor re-enterable — the next `next_batch` call retries
+                    // this same leaflet (transient I/O, wasm NeedFetch miss)
+                    // instead of dropping it and the rest of the leaf.
+                    let load_result = (|| -> io::Result<ColumnBatch> {
```
bplatz, line 437: No change requested — flagging as the watch point.

This is the only unconditional native change on a scan path. Success path is unchanged and it's per-leaflet against a µs-scale column decode, so codegen-perturbation risk is low (CLAUDE.md's own note: fat-bodied loops tolerate this fine). But it's where shared code moved inside the hot scan driver, and the bench covering it runs the feature-on build.

Worth naming in the description as the first place to look if a regression ever appears, rather than treating the whole diff as uniformly risky.

bplatz (approved):
Approving. Reviewed for native performance impact specifically — not the wasm design or read-path correctness. Please look at the four inline items before merging; only the Cargo one changes anything you've written.

What I verified, independently rather than from the description:

- `cargo tree -p fluree-db-api -e features,no-dev` — no `residency` in the production graph. `-e features` — present via dev-deps.
- The three files CLAUDE.md names as hot-loop-critical (`count_plan_exec`, `fast_path_common`, `fast_union_star_count_all`) contain only `from_io` inside `.map_err()` on `open_leaf_handle` (error path) and two `Instant::now()` calls outside any loop. No per-row or per-group code added.
- `clock.rs` native is `pub use std::time::{Instant, SystemTime}` — real alias, same function.
- `wasm_compat::SpatialIndexProvider` native is a re-export of the same trait, so the `spatial_provider_map` signature change is an import-path rename, same vtable.
- `leaflet_cache.rs` keeps moka and `in_blocking_region` unchanged; only `arc_io_error` on the loader-error path.
- `leaf_access.rs` additions wholly cfg-gated; no existing handle modified.
- `raw_txn_upload.rs` native code unchanged, attributes only.
- `residency_mode` is itself cfg-gated, so a default-feature native build gets no `dead_code` warning (`--all-features` clippy would not have caught that one).

I did not re-run benches — a criterion A/B here is hours and shared-runner noise is the documented problem. The confidence above is structural.

aaj3f:
Thanks — this is the review I wanted, and the independent verification is appreciated. All four addressed; one changed code, three changed what the description claims.

**The dev-dep feature unification.** Reproduced your `cargo tree` result exactly (0 in the `no-dev` graph, 3 with dev-deps). Taking your suggestion to state it rather than fight the Cargo setup, the description now says both consequences plainly: that the A/B measured a residency-*enabled* build against a base with none of that code and still landed in the noise — the conservative direction, so the result is stronger than it claimed — and that no perf gate measures the production configuration, so "production native builds compile the pre-residency read path" rests on the cfg argument alone. That argument is sound because cfg'd-out code is absent rather than predicted-away, but the numbers demonstrate something adjacent to that sentence rather than that sentence, and saying so exactly is better than implying otherwise.

**The vtable call.** Implemented. `residency_mode` is now a `bool` computed at construction. I checked your premise before relying on it: `cas` is assigned once in each of the two constructors, never reassigned, and the field is private, so the cached answer cannot go stale. The field carries the same cfg as the method, so a production native build has neither; verified building in all three configurations (default native, `--features residency`, wasm32).

Your second reason is the one that persuaded me over the call cost itself — it takes dyn dispatch out of the build the benches actually compile, so it narrows the measured-vs-shipped gap rather than widening it, which given the point above is worth more than the cycles.

**"Byte-identical."** Fair, and corrected in both places it appeared. It now reads "semantically identical, with one inlined delegation," naming the `open_leaf_handle` / `open_leaf_handle_native` split as the reason. A claim doing this much load-bearing work should be stated exactly.

**`binary_cursor.rs:417`.** Adopted as written — the description now names it as the first place to look if a regression ever appears, with your reasoning (only unconditional native change on a scan path, success path unchanged, per-leaflet against a µs-scale column decode, and the bench covering it runs the feature-on build). Better than leaving the diff to read as uniformly risky.

Issue #1670, opened 2026-08-20
Delete semantics: soft drop reserves the alias forever and cannot be undone


bplatz:
### Summary

A soft drop — the **default** for `DELETE /ledger` — permanently reserves the ledger alias, and there is no supported way to reclaim it or to recover the ledger. The only escape is a hard drop, which destroys the data the soft drop existed to preserve.

This surfaced as a report from an application on the DynamoDB nameservice that deletes ledgers and then recreates them under the same name, and gets `Ledger already exists` on every recreate.

### Reproduction

Verified on both the file backend and real DynamoDB (LocalStack):

```
[dynamo soft] init → retract → init  →  Err(LedgerAlreadyExists("soft/db:main"))
[dynamo hard] init → purge   → init  →  Ok(())
```

```
[file soft] drop -> Ok((Dropped, 0 artifacts))
[file soft] RECREATE FAILED: Ledger already exists: app/db:main
[file hard] RECREATE OK
```

### Mechanism

1. `DELETE /ledger` without `"hard": true` maps to `DropMode::Soft` (`fluree-db-server/src/routes/ledger.rs:298`) — the default.
2. Soft mode calls `LedgerLifecycle::retract`, not `drop_branch` (`fluree-db-api/src/admin.rs:657`).
3. `retract` marks the meta row `retracted = true` and leaves every row in place (e.g. `fluree-db-storage-aws/src/dynamodb/mod.rs:1235`).
4. Recreating calls `init`, whose DynamoDB `TransactWriteItems` are conditioned on `attribute_not_exists(pk)` — the surviving rows cancel the transaction, which maps to `LedgerAlreadyExists` (`dynamodb/mod.rs:1225`).

Not backend-specific: the trait contract states it directly — *"Returns an error if a record already exists (including retracted records)"* (`fluree-db-nameservice/src/lib.rs`).

### The actual defect

Soft drop is a one-way door. Nothing in the codebase sets `retracted` back to `false` — there is no un-retract, undrop, or reinstate path anywhere (`restore_ledger` in `commit_transfer.rs` is the unrelated `.flpack` restore). So the mode documented as *"the default and safest option… all data files remain on disk for potential recovery"*:

- permanently burns the alias, and
- provides no recovery mechanism that would justify having reserved it.

`docs/api/endpoints.md:1750` states the reservation but not that it is irreversible.

Either half would be defensible alone. Together they are a trap: the safe-sounding default is the one with no way back, and callers reach it by simply not passing a flag.

[...]

PR #1716, opened 2026-08-26
fix: let a no-op republish complete its raft queue entry

bplatz:
## Problem

Every zero-change write routed through the raft committer hung until the waiter timed out with `submission stranded by leader transition`.

The commit worker's no-change arm completes a zero-change transaction by republishing the branch's **current** head unchanged — same `t`, same commit id. `apply_head`'s monotonicity guard (which exists to refuse stale writers) rejected exactly that shape (`commit_t <= current_t`) and pushed the queue entry back for retry. Re-staging a no-op can only reproduce the same unchanged head, so the submission looped re-stage → reject forever.

## Fix

Equal `t` **and** equal commit id is now recognized as the designed no-op completion:

- the queue entry is consumed and the idempotency record cached
- the ref is deliberately left untouched, so `last_advanced_*` stamps keep pointing at the last real advance

A same-`t` publish with a **different** commit id is still a stale writer and is still refused, so the guard's safety property is unchanged.

## Testing

- New regression test `apply_head_completes_a_no_op_republish_of_the_current_head` pins the completion behavior (entry consumed, ref untouched, stale same-`t`/different-id still rejected)
- `cargo test -p fluree-db-consensus`

### DIFF
--- a/fluree-db-consensus/src/raft/integration.rs
+++ b/fluree-db-consensus/src/raft/integration.rs
@@ -123,6 +123,25 @@ pub struct RaftIntegration {
     /// [`attach_ledger_manager`](Self::attach_ledger_manager) once
     /// `Fluree` exists.
     ledger_manager: Arc<OnceLock<Arc<fluree_db_api::LedgerManager>>>,
+    /// File-registry adoption inputs, when the embedder configured
+    /// them (see [`RaftBootstrapConfig::adopt_file_registry`]):
+    /// where the predecessor registry lives, and the marker file
+    /// under the raft storage root that records a completed replay.
+    adoption: Option<AdoptionPaths>,
+}
+
+/// Where a configured file-registry adoption reads from and how it
+/// remembers completion. Derived in [`RaftIntegration::bootstrap`].
+#[derive(Clone, Debug)]
+struct AdoptionPaths {
+    /// The store root holding the predecessor `ns@v2` registry.
+    registry_root: PathBuf,
+    /// `<raft storage root>/file-registry-adopted` — written after a
+    /// successful replay. Node-local on purpose: only the node that
+    /// initialized the cluster ever adopts, and a crash between the
+    /// (idempotent, conflict-tolerant) replay and this write is
+    /// healed by simply replaying again on the next boot.
+    marker: PathBuf,
 }
 
 /// Shared HTTP transport: a `reqwest::Client` paired with the
@@ -190,6 +209,7 @@ impl RaftIntegration {
             nameservice,
             release_rx: Arc::new(Mutex::new(Some(release_rx))),
             ledger_manager,
+            adoption: None,
         }
     }
 
@@ -224,6 +244,13 @@ impl RaftIntegration {
     /// - `POST /cluster/add-learner` (the existing leader, naming
     ///   this node) — joins an existing cluster.
     pub async fn bootstrap(config: RaftBootstrapConfig) -> Result<Self, RaftBootstrapError> {
+        let adoption = config
+            .adopt_file_registry
+            .as_ref()
+            .map(|root| AdoptionPaths {
+                registry_root: root.clone(),
+                marker: config.storage_path.join("file-registry-adopted"),
+            });
         let storage = Arc::new(FsRaftStorage::open(config.storage_path).await?);
 
         let event_bus = Arc::new(LedgerEventBus::new(config.event_bus_capacity));
@@ -265,7 +292,7 @@ impl RaftIntegration {
 
         let raft = Raft::new(config.node_id, raft_cfg, factory, log, sm).await?;
 
-        Ok(Self::new(
+        let mut integration = Self::new(
             Arc::new(raft),
             config.node_id,
             shared_state,
@@ -280,7 +307,9 @@ impl RaftIntegration {
                 config: network_config,
             },
             release_rx,
-        ))
+        );
+        integration.adoption = adoption;
+        Ok(integration)
     }
 
     /// Inter-node Raft RPC router. Mount under `/raft` on the private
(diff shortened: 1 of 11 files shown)


fluree-db-consensus/src/raft/integration.rs
```diff
@@ -315,6 +344,173 @@ impl RaftIntegration {
...
+                            return Err(FileRegistryAdoptionError::replay(
+                                ledger_id,
+                                "index head",
+                                e,
+                            ))
+                        }
+                    }
+                }
+            }
+            if let Ok(Some(config)) = file_ns.get_config(ledger_id).await {
```
aaj3f: **The config and status replay arms discard a read error, and then the marker gets written anyway — so a one-shot migration can drop a ledger's config permanently and tell nobody.**

`if let Ok(Some(config)) = file_ns.get_config(ledger_id).await` has no `Err` arm. `FileNameService::get_config` returns `Err` on a `split_ledger_id` failure and on any `read_json_from_address` I/O or JSON-parse failure (`fluree-db-nameservice/src/file.rs:1598-1605`), so a truncated or unreadable `ns@v2` record surfaces as `Err` and lands in the discarded arm. The loop then continues, the per-record `tracing::info!("file registry record adopted into the replicated nameservice")` still fires, `adopted` still increments, and `finish(adopted)` writes `<raft storage root>/file-registry-adopted`. From then on the `adoption.marker.is_file()` early return at the top of the function makes every later call `Ok(0)`.

That is unrecoverable from the operator's side by design: adoption is deliberately once-only, and the only way back is deleting the marker by hand — which they have no signal to do, because nothing was returned, nothing was warned, and the info log says the record adopted fine.

This is also the one place in the function that departs from the principle the commit message sets out for itself: "Genuine divergence … refuses loudly rather than guessing which side is authoritative." The two head arms honour that with `FileRegistryAdoptionError::Diverged`. These two swallow.

Failing scenario: transition boot on a store where one ledger's `ns@v2` JSON was truncated by an earlier unclean shutdown. `all_records()` still yields the record, so existence and both heads adopt correctly; `get_config` errors; the ledger arrives in raft with its config silently absent, the marker is written, and the operator's first symptom is a ledger behaving with default config in production.

The fix is small: give both arms explicit `Err` handling and route it through the error type the publisher-side arms immediately below already use — `FileRegistryAdoptionError::replay(ledger_id, "config", e)`. If you'd rather not fail the whole adoption over one unreadable optional record, then at minimum `warn!` it and count it, and have `finish()` refuse to write the marker when the count is non-zero, so a re-run is still possible.

fluree-db-consensus/tests/it_file_registry_adoption.rs
```diff
@@ -0,0 +1,198 @@
+//! The transition boot: a store that lived in the single-node FILE
+//! posture moves into raft, and `adopt_file_registry` carries every
+//! registry record — existence, both heads, config, status — into the
```
aaj3f, line 3: **The test module's own doc claims config and status coverage that isn't there — and those are exactly the two arms with the swallowed error above.**

Line 3 reads "carries every registry record — existence, both heads, config, status — into the replicated nameservice". `seed_file_registry` (lines 37-70) calls `init`, `fast_forward_commit`, and `compare_and_set_ref` only; it never calls `push_config` or `push_status`. Neither test reads config or status back. I checked: `grep -n "config\|status"` over the file returns only this doc line and the name `no_registry_and_no_config_both_answer_zero`, where "config" means the bootstrap config, not a ledger config.

So both replay arms are entirely unexercised. The heads path is well covered — replay fidelity, tombstone skipping, marker once-only, and convergence over a partially adopted machine are all pinned, and the partial-retry case in particular is a nice piece of test design. Config and status just aren't in it, while the doc says they are.

Seeding a config and a status in `seed_file_registry` and asserting both come back through `integration.nameservice()` would cover the arms and make the doc true. It would also give the fix above something to fail against.

---

fluree-db-consensus/src/raft/state_machine.rs
```diff
@@ -2439,10 +2439,26 @@ fn apply_head(state: &mut NameServiceState, log_index: u64, args: StagedHead) ->
...
-        if commit_t <= current_t {
+    //
+    // One deliberate exception: the no-op completion. A worker that
+    // staged a zero-change transaction completes its queue entry by
+    // republishing the branch's CURRENT head unchanged (the
+    // commit worker's no-change arm), which arrives here as
+    // `commit_t == current_t` with `commit_id` equal to the stored
+    // head. That is not a stale writer — it is the designed way an
+    // entry that advances nothing leaves the queue. Rejecting it
+    // (as this guard once did) pushed the entry back and stranded
```
aaj3f, line 2450: `fluree-db-consensus/src/raft/commit_worker.rs:752` — The `process_revert` docstring still asserts the mechanism this PR just disproved — it says `ApplyHead` against the same head "surfaces via `QueueDesync::WrongFront` only if another transactor jumped ahead, which is exactly the race the queue already serializes against." Both halves turned out to be wrong: it surfaced as `HeadNotMonotonic`, not `WrongFront`, and it fired unconditionally rather than only under a jumped-ahead transactor. That belief is arguably why this sat latent since June. The code is fixed now, but the comment sits directly above one of the four arms that depends on the new behaviour, so the next reader gets the old model. Minor and non-blocking — but if you agree it's right, I'd rather see it folded in now than lost in the backlog.

Commenting here because `fluree-db-consensus/src/raft/commit_worker.rs` is not in this diff.

fluree-raft-core/src/forward.rs
```diff
@@ -920,3 +920,119 @@ mod tests {
...
+            Ok(resp) => return Ok(resp.data),
+            Err(openraft::error::RaftError::APIError(
+                openraft::error::ClientWriteError::ForwardToLeader(fwd),
+            )) => {
+                let Some(node) = fwd.leader_node else {
+                    return Err(ProposeError::NoLeader);
+                };
+                match relay_propose::<C>(&node.raft_addr, &cmd).await {
+                    Ok(data) => return Ok(data),
+                    Err(e @ ProposeError::Relay(_)) => {
```
aaj3f, line 973: The relay's retry arm can double-apply. `relay_propose` maps a `resp.bytes()` failure to `ProposeError::Relay(...)` at line 1024, and this arm retries the whole `client_write` on any `Relay` error. By the time `bytes()` fails the leader has already answered a status line, so it has already applied the command — the retry lands it twice. The 503 arm is genuinely safe (the leader refused with `ForwardToLeader`, nothing applied), and the `send()` failure is honestly ambiguous, but the response-read failure is unambiguously post-apply. Since there's no production caller yet — only `multi_node_group.rs:289` — this is a design note rather than a live defect, and I'd normally leave it. I'm raising it because the commit message frames this as the answer for a load balancer spreading writes across a group, so it will get wired to something, and at that point "at least once" becomes a property callers need told about. Either narrow the retry to the 503 and `send()` cases, or say plainly in the `propose_via_leader` doc that delivery is at-least-once and commands should be idempotent.

fluree-db-consensus/src/raft/integration.rs
```diff
@@ -315,6 +344,173 @@ impl RaftIntegration {
...
+        let registry = adoption.registry_root.join("ns@v2");
+        let finish = |adopted: usize| -> Result<usize, FileRegistryAdoptionError> {
+            std::fs::write(
+                &adoption.marker,
+                format!("adopted {adopted} ledger records\n"),
+            )
+            .map_err(FileRegistryAdoptionError::Marker)?;
+            Ok(adopted)
+        };
+        if !registry.is_dir() {
```
aaj3f, line 411: `if !registry.is_dir() { return finish(0); }` writes the marker when the configured root has no `ns@v2` subtree. A typo in the operator's `adopt_file_registry` path therefore burns the one-shot adoption for that raft storage root, silently. Answering `Ok(0)` without writing the marker in this specific case would leave the door open for the corrected config, and you'd lose nothing — the `adoption.is_none()` case above already returns `Ok(0)` without a marker.

fluree-db-consensus/src/raft/integration.rs
```diff
@@ -315,6 +344,173 @@ impl RaftIntegration {
...
+    ///   values VERBATIM. Retracted records are skipped — a tombstone
+    ///   carries no live state.
+    /// - Crash-safe by idempotence, not atomicity: a crash between the
+    ///   replay and the marker write re-runs the replay next boot, and
+    ///   every step tolerates its own prior success (`AlreadyExists`
+    ///   on init; a head CAS conflict where the machine already holds
+    ///   a value at least as new; a config/status conflict).
+    ///
+    /// Returns how many ledger records were carried.
+    pub async fn adopt_file_registry(&self) -> Result<usize, FileRegistryAdoptionError> {
```
aaj3f, line 384: Nothing in `adopt_file_registry` checks that this node is the one that initialized the cluster; the "joining nodes must NOT call this" contract lives only in the doc comment. Given `RaftBootstrapConfig` is the kind of thing that gets templated across every node in a deployment, a cheap `self.raft.current_leader()` check (or an explicit `is_initializer` flag threaded from bootstrap) would make the contract enforceable rather than advisory.

---

aaj3f (approved):
Approving @bplatz but recommending some changes, even if narrowly, and only against the file-registry adoption commit. The titular fix is correct, and I'd approve `fa0a541b4` on its own without hesitation.

The notes I'd want to focus on are largely from `ea1d6972e`. The config and status replay arms at `integration.rs:490` and `:499` discard read errors from `get_config`/`get_status`, and the function then writes its own once-only marker — so an unreadable record silently loses that ledger's config with no error, no warning, an info log claiming success, and no second chance for the operator. It's the only place in that function that doesn't follow the "refuse loudly rather than guess" rule you set for it, and the two arms have no test coverage at all despite the test module's doc claiming they do. Both are small fixes and the design around them is good — the idempotent-not-atomic replay, tolerating prior success on every step, and the partial-retry test are all well judged.

The optional items below the blocking ones are genuinely optional; the propose-relay retry note is about an API with no production caller yet, and I'd not hold anything for it.

### db adherence checklist

| Axis | Assessment |
| --- | --- |
| **Performance (speed-first)** | **Positive.** One `ContentId` equality added per `ApplyHead` — once per committed write, inside a function already doing a map lookup and a `VecDeque::pop_front`. Removes an unbounded re-stage → reject → refresh loop that burned a staging pass plus a raft round-trip per iteration until waiter timeout on every zero-change write. Separately, `b8bb981d3` recovers a raft cluster from running entirely unindexed, which is a large latent read-latency win. No hot-path query code touched. |
| **SPARQL ↔ JSON-LD IR parity** | Not applicable — no query, parse, or lowering code in the diff. |
| **Shared-abstraction adherence** | Good. The fix stays inside the existing guard rather than adding a parallel path; `drop_queue_if_empty`, the idempotency insert, and the `HeadApplied` response all run unchanged. The relay reuses `is_valid_leader_url` and the no-redirect client posture from the existing forwarding middleware instead of inventing a second SSRF story. `adopt_file_registry` replays through the ordinary `NameServiceLookup`/`RefPublisher`/`ConfigPublisher` traits rather than reaching into the state machine. |
| **Tests / CI** | CI ran and is green on all six real jobs (`fmt`, `clippy`, `test` 14m22s, `testsuite-sparql`, `bench-paths`, `plan`); the eight "skipping" checks are release-publishing jobs. The `test` job is `--workspace --all-features`, so the non-default `raft`-gated tests genuinely execute. I re-ran the targeted suites and mutation-checked the new test: reverting the guard exception turns `apply_head_completes_a_no_op_republish_of_the_current_head` red in 0.026s (fails fast, doesn't hang). Gap: the config/status adoption arms are untested — see the second blocking comment. |

---

**Packaging note.** The title and body describe one of four changes. The other three are a new peer-trusted HTTP write endpoint with its own SSRF surface, an indexer lifetime fix, and a 260-line one-shot migration feature with a marker file. The indexer fix in particular deserves better billing than it's getting — a raft cluster running with no background indexer at all, every read walking the commit chain from genesis, is a bigger operational finding than the no-op strand this PR is named for. Splitting `ea1d6972e` out would also let the adoption feature get reviewed as the migration it is.

fluree-db-consensus/src/raft/integration.rs
bplatz: Addressed in 403812cfc0a37812a52cf3e4b3ccfa557ae74d7c.

Your finding was the visible half. Writing the test you asked for in the next comment turned up a second fault in the same two arms: they pushed with `expected: None`, and `apply_versioned_push` (state_machine.rs:2125) reads an absent value as `VersionedValue::absent()` — `ConfigValue::unborn` / `StatusValue::initial` — then requires `expected.as_ref() == Some(&current)`. `None` can never match, so every push returned `Conflict` and the `Ok(_) => {}` swallowed that too. Neither config nor status was ever carried, on any path. The `// A conflict is a retried replay's own prior write` comment was wrong: it was the first attempt failing, every time.

Both arms are now read-then-CAS in the shape of the index-head arm above them — read failure aborts through `Replay`, a real conflict refuses through `Diverged`, and a watermark at or above the registry's absorbs a retried replay's own write.

fluree-db-consensus/tests/it_file_registry_adoption.rs
bplatz: Addressed in 403812cfc0a37812a52cf3e4b3ccfa557ae74d7c. Mutation-checked both ways: restoring `expected: None` fails with `Diverged { machine holds v: 0, registry carries v=4 }`, and restoring the original arm including the swallow fails the readback at `left: 0, right: 4`.

fluree-db-consensus/src/raft/state_machine.rs
bplatz: Addressed in 403812cfc0a37812a52cf3e4b3ccfa557ae74d7c.

fluree-db-consensus/src/raft/integration.rs
bplatz: Addressed in 403812cfc0a37812a52cf3e4b3ccfa557ae74d7c.

fluree-db-consensus/src/raft/integration.rs
bplatz: Not addressed — leaving as-is for now. Agreed the contract should be enforceable rather than advisory, but `current_leader()` doesn't actually separate the initializer from a joining node that later won an election, so the honest version is the `is_initializer` flag threaded from bootstrap. That's a bootstrap-surface change I'd rather not fold into this branch.

fluree-raft-core/src/forward.rs
bplatz: Not addressed — taking your read that this shouldn't hold the branch. The analysis is right: a `bytes()` failure is unambiguously post-apply and the retry double-applies. Since the choice between narrowing the retry and documenting at-least-once depends on what ends up calling it, I'd rather make it with the caller in hand than guess now.

aaj3f:
Heads up @bplatz — this one turned `main` red, and I think it's just a test that encodes the old semantics rather than anything wrong with the change.

`43d758610` makes `ledger_exists` treat a retracted record as absent. `it_ledger_lifecycle::ledger_exists_on_file_storage` asserts the opposite:

```rust
// A soft-dropped ledger still has a (retracted) record: `exists`
// is "is there a record", not "is it live".
assert!(
    fluree.ledger_exists("x:main").await.unwrap(),
    "exists() reports the record, which a soft drop keeps",
);
```

`fluree-db-api/tests/it_ledger_lifecycle.rs:219`. Failing on main since the merge at 01:35 (run 33133302931), and it's inherited by every branch that has merged main since — I hit it on #1701, which touches nothing but the planner.

Your commit message reads like the test is simply now wrong ("exists is a query-path question"), in which case updating the assertion and its comment is the whole fix. I didn't want to push that myself since it's your semantics call and the old assertion was deliberate enough to carry an explanatory comment. Happy to push it if you'd rather — just say which way you want `exists` documented.

One thing worth flagging while you're in here: this settles a sub-question of #1670, which is still open and carries the wider delete-semantics decision (init-reclaims vs reject, whether an un-retract path exists, branch-scoped hard purge). Your commit body already cites the same divergence. Might be worth a line on #1670 recording that the exists half is now decided, so whoever picks up the rest isn't re-deriving it — and so the two don't drift apart, which is the failure mode that issue is about in the first place.

PR #1718, opened 2026-08-27
perf(transact): take the cached ledger state for the commit window — stop per-commit DictNovelty deep-clones for serial writers

### LINKED ISSUE #1707: Cached-handle transact path deep-clones DictNovelty on every commit (O(accumulated novelty) once an index exists)
While profiling sustained continuous-transact workloads (many small commits into one ledger over a long window) I hit the `fluree::cow_probe` warning firing on essentially **every** commit — with zero concurrent readers — and chased it down. I think what we have is a real per-commit perf bug on the server's normal write path, and one that gets worse exactly when we can least afford it (when the indexer falls behind). Writing it up with the mechanism because the fix needs to be made carefully.

**The short version:** the cached-handle transact path (`ledger_cached` → `stage(&handle)…execute()`, which is what every server transact route uses — `fluree-db-server/src/routes/transact.rs:1519`, `:1872`, `:2002`, `:2069`) can **never** satisfy the probe's unique-ownership precondition, so `Arc::make_mut` deep-clones `DictNovelty` (and `RuntimeSmallDicts`) on **every commit**. The regression that 34592f3e2 fixed for the owned/threaded path was never actually fixed for the cached-handle path, and no test guards either probe site.

**Mechanism** (Claude traced and measured this; pasting its findings essentially as-is, fwiw):

- The probe at `fluree-db-transact/src/commit.rs:994-1000` reads `Arc::strong_count` on the dict Arcs moved out of the commit's base state. Baseline `1` is only reachable when the base `LedgerState` is uniquely owned — i.e. the owned path (`stage_owned`), which is what the original fix was measured against.
- On the cached-handle path the LedgerManager's cached `LedgerState` (`fluree-db-api/src/ledger_manager.rs:177`) is **always a second holder**: the optimistic path Arc-clones every field via `snapshot()`/`to_ledger_state()` (`fluree-db-api/src/tx_builder.rs:1627-1630`, `ledger_view.rs:114-119`), and the fast path uses `clone_state()`. Measured `strong_count == 2` from commit #0, index or no index.
- Once an index is attached it's **3**: the commit path itself rebuilds a `BinaryRangeProvider` every commit (`commit.rs:1034-1039`) whose clones live inside the *cached* snapshot, and the detach at `commit.rs:979-981` can't release it — `base.snapshot` is an `Arc<LedgerSnapshot>` *shared* with the cache, so `Arc::make_mut(...).range_provider = None` deep-copies the snapshot and nulls the private copy while the cache's copy keeps its provider. The probe confirms this directly (`had_binary_provider=true` and `dict=3` after the detach has already run).
- `apply_index_v2` itself is clean, for the record — this is not an index-install leak.

**Cost shape:** the clone is ~15ns per accumulated dictionary entry (the expensive part is `NsVecBiDict.reverse`'s one heap allocation per subject), i.e. **O(entries since the last index) per commit → O(commits²) across an unindexed window**. Measured end-to-end (release, fsync off, indexing off to open the window): cached-handle went 0.82 → 2.63 ms/commit over 1,500 commits (total 2.55s vs 1.09s for the owned control); at 2,000×50-node commits the per-commit delta hit 6.8ms, matching the microbench within 15%. With the default `reindex_min_bytes = 100` the dicts stay trimmed and the tax is only ~0.35–0.55 ms/commit — so this mostly hides in steady state, and then **compounds precisely when the indexer lags** (bulk ingest, slow storage, indexer contention), feeding back into ingest slowdown right when the system is already struggling. That coupling is what I think makes this worth fixing rather than shrugging at.

**Fix direction (emulated and verified, not yet implemented):** take-instead-of-clone. Under the write guard in `commit_and_finalize` (`tx_builder.rs:1325`), after the no-op early return, `std::mem::replace` the cached state with a cheap placeholder and drop it — the staged base is provably identical at that point (the optimistic loop verified `t` and `head_commit_id` at `:1644-1648`; the fast path held the lock throughout), and `finalize_commit` already replaces the state on success. An emulation of exactly this via the public API eliminates the growth entirely (probe reads 1 on every commit, latency flat 0.76–0.81 ms/commit on the same workload). Three constraints the emulation surfaced the hard way, for whoever picks this up (probably me):

1. The write guard must be held across the whole commit — a released guard let the cache event listener reconcile against the placeholder and produced a spurious `CommitConflict`.
2. The error path must **evict** (`disconnect`, precedent at `fluree-db-api/src/tx.rs:2710-2719`), never park-and-restore the old Arcs — restoring keeps the count at 2 and defeats the fix.
3. tokio's `RwLock` doesn't poison, so the take-to-replace window needs a drop guard that evicts on unwind rather than leaving a placeholder serving reads.

Two honest caveats: this removes only the *cache's* holder — a concurrent query holding a `LedgerView` across the commit window still pins the Arcs and the COW returns (inherent to the current design); and the durable fix is probably to commit **into** the cached state in place the way `apply_single_commit` already does (`fluree-db-ledger/src/lib.rs:896-898`), which would kill the class of bug instead of this instance. I'd suggest we take the minimal fix now and treat in-place commit as the follow-on design question, since the latter needs a decision beyond a bugfix PR.

I'll put up a PR with the minimal fix plus a regression test asserting unique ownership at the probe sites (there is none today). Happy to talk through any of this — particularly if someone remembers a reason the cached state must survive the commit window that I'm not seeing.

aaj3f:
Closes #1707.

The cached-handle transact path — what every server transact route uses — could never satisfy the commit path's unique-ownership precondition: the LedgerManager's cached `LedgerState` was always a second holder of the dict Arcs (and the snapshot's range provider a third once an index attached), so `Arc::make_mut` deep-cloned `DictNovelty` on **every commit**, O(entries-since-last-index) per commit and O(commits²) across an unindexed window. #1707 has the full mechanism; the fix follows the direction verified there.

**What this does.** `commit_and_finalize` now empties the cache slot for the commit window via a private `DetachedCacheSlot`: it takes the cached state out from under the already-held write guard (leaving a cheap placeholder), which makes the commit's base uniquely owned — and, as a bonus, makes the pre-existing range-provider detach in `commit.rs` effective for the first time, because `base.snapshot` is finally unique. `finalize_commit` is split into `install_committed_state` + `trigger_reindex_if_needed` so the slot stays held across the install and can repair the cache if the install fails; its five other callers keep byte-identical behavior, including releasing the lock before the reindex trigger.

**The placeholder is unobservable — including under cancellation.** The guard is held with zero release points across detach → install → refill, and every state reader routes through that one lock, so no reader can see the placeholder on any normal path. The subtler case is *future cancellation*: the commit runs inside the HTTP request future, and axum drops handler futures on client disconnect — an unshielded version of this change would have let a mid-window cancel release the lock and briefly serve the empty placeholder to parked readers (internal review caught exactly this, with a reproducing test). So the entire detach→refill span — including both failure-repair paths — runs as its own spawned task (`commit_shielded`) whose `JoinHandle` the caller awaits: **a cancelled request abandons the wait, never the commit**, and the cache is always refilled. The spawned future carries the caller's tracing span, so request-id linkage on commit-side logs survives the shield. Cost of the shield: ~0.02–0.03 ms/commit (one spawn + a handful of Arc bumps), re-measured below.

**Error and unwind paths.** On commit failure the primary repair is **reload-in-place** under the still-held guard (reflecting the durable head whether the commit died before or after publishing), with `disconnect` as the fallback when storage itself is unreachable — plain eviction would be wrong as the primary, since callers already holding the `LedgerHandle` would keep reading the placeholder. A drop guard remains as the genuine last resort for panics: releases the lock, logs, and spawns an out-of-band reload (a t=0 placeholder always loses the manager's monotonic-swap guard, so the repair reliably lands). The two consciously-accepted double-failure residuals (storage down during repair; ephemeral handles with caching disabled) are documented in place.

**Both write paths are covered** — `commit_and_finalize` is shared by the optimistic and lock-held (policy/SPARQL/pre-built) paths, and the ownership test includes a SPARQL-UPDATE phase that goes through `clone_state()` to prove the second one.

**Numbers** (release build, fsync off to open the window; Claude ran these and I'm pasting as-is). 1,500 commits × 20 nodes, indexing off — median ms/commit by band:

| band | before | after (shielded) | never-cached control |
|---|---|---|---|
| 0–150 | 0.79 | 0.75 | 0.64 |
| 700–850 | 1.50 | 0.72 | 0.64 |
| 1350–1500 | 2.46 | 0.79 | 0.64 |
| total | 2.41 s | 1.15 s | 1.02 s |

The growth is gone, not reduced — per-commit cost is flat in accumulated novelty, matching the never-cached path (~2.1× at 1,500 commits and still widening with window size before the fix). With background indexing on (1,200 × 50): ownership probe went `{3: 1184, 2: 15, 1: 1}` → `{1: 1200}`, total 3.62 s → 1.88 s. The owned/threaded control path is unregressed.

**Tests.** Three new tests, each verified to fail against a deliberately broken build before being accepted: an ownership harness asserting `dict_novelty` / `runtime_small_dicts` strong-count == 1 at commit time on the cached path (pre-index, post-install, immediately-after-install — the guard that didn't exist, which is how 34592f3e2 quietly covered only the threaded path); a recovery test asserting a failed commit leaves the handle serving the committed head (its ceiling is set one byte above current novelty so the failure fires in the *commit* path — the obvious 1-byte-ceiling version was vacuous, rejected at staging); and a cancellation test that parks a reader behind a mid-window abort and asserts it never observes t=0 and the commit still lands (deterministic: there is no await between lock acquisition and the detach, so `is_locked()` observed from outside implies the window is already open — a property worth knowing before anyone adds an await in that gap).

**Gates.** `cargo test -p fluree-db-api` 3,262 passed / 0 failed; `-p fluree-db-consensus` 53/53 (a `finalize_commit` caller); clippy `-D warnings` clean; fmt clean. One honest caveat: workspace-wide `cargo check --all-targets` was run with `--exclude fluree-search-httpd --exclude fluree-search-service` on my box — those two hit a local C++ toolchain issue (usearch/cxx cannot compile here at all, unrelated to this change); CI's unrestricted run is the authoritative check for them.

**Scope limitation, measured — concurrent writers to the same ledger still see the clone.** Removing the cache's holder makes the base uniquely owned when one transaction is in flight. It does not when several are: the optimistic path takes its base via `ledger.snapshot()` + `to_ledger_state()` *before* acquiring the write guard, so every concurrently-staging transaction holds its own Arc clones of the same dictionaries, and `make_mut` copies again. A controlled sweep on this branch (identical workload, only the submitter count varying) shows it cleanly — 1 submitter: **0 ownership violations** in 3,060 commits; 2 submitters: 3,126 of 3,329 (94%); 4 submitters: 3,640 of 3,720 (98%). So the honest claim for this PR is: **the per-commit clone is eliminated for serial writers to a ledger, and reduced-but-not-eliminated under concurrent ones.** That is still worth having (the serial case is the whole single-node ingest story, and the O(commits²) blowup it removes is real), but nobody should read the benchmark above as "the clone is gone under load." The general answer is the same one the reader case points at — committing *into* the cached state in place, as `apply_single_commit` already does — and it wants its own design discussion rather than being smuggled into this fix.

**Deliberately not in this PR** (each needs the same detach at the top of its own flow, and the failure-path story above should settle once before being replicated): five sibling cached-handle commit flows still clone from the cache — `merge.rs`, `rebase.rs`, `revert.rs`, `cypher_txn.rs`, and the Raft `commit_worker` — with cypher and the Raft worker carrying real write volume. Also unchanged by design: a concurrent reader holding a `LedgerView` across the commit window still pins the Arcs for that window — the durable answer there is the in-place commit shape `apply_single_commit` already uses, which is a design conversation, not a bugfix. If we'd rather fold any of the five in here, happy to talk through the ordering.

---

**Post-review addendum** (commits `35fd8cc3d` … `afcd3cb51`, head `afcd3cb51`):

*Scope grew where the maintainers' fold-it-in preference said it should, and shrank where verification said the premise was wrong.* Of the five deferred sibling flows: **merge and revert are now folded in** (same detach via a shared `apply_staged_detached`, spawn-shielded because both run inside request futures — with one deliberate difference: their failure repair is **evict, not reload**, because both flows roll the nameservice head back after an apply error, and reload-first would cache the very head the rollback is about to withdraw; pinned by `it_branch_op_cow`, mutation-verified). **Rebase turned out not to need it** — its replay base is a direct storage load, never a cache clone; a detach there would be pure risk for zero benefit (comment added saying so). **Cypher's clone has a different mechanism entirely** — per-statement `make_mut`s at *stage* time against state shared with the transaction itself, outside any lock — so a commit-window detach saves nothing; it belongs with the Raft worker's residual (Follow-up: #1742), and relatedly its private-state policy resolution can poison the shared config cache (filed separately as #1745).

*The reviewer's "unexplained third of the gap to the floor" is answered, and it was the whole gap:* `LedgerView::to_ledger_state()` hardcoded `head_temporal: None`, so every optimistic-path commit re-read and re-decoded the **head commit blob from storage inside the locked window** (`ensure_head_temporal` — whose own doc says its purpose is "no per-commit storage reads"). Fixed by carrying the `Copy` field through (`afcd3cb51`). Verified by paired interleaved binaries with an owned-mode negative control, and the delta scales linearly with head-commit size (~0.08 ms/commit at 20-node commits → ~0.55 at 200-node) — the signature of a per-commit blob read, not lock or spawn overhead. Post-fix, cached mode sits ~0.01–0.03 ms/commit above the never-cached control — approximately the spawn shield alone. (Measured on a loaded box: direction and scaling are solid; absolutes are approximate.)

*One corner of the review's own safety argument tightened:* "a poisoned key-0 config entry would self-heal on the next miss" doesn't hold on the panic path — the out-of-band reload never touched `config_cache`, and a config-quiet reloaded state legitimately carries marker 0 too, so the poisoned entry would *hit*. Now `config_cache_clear()` at reload's swap point (`f2b226e45`); needs panic + racing reader + config predating `index_t` to matter, but it's fixed rather than argued.

*Also per review:* `ledger()`'s manager-free invariant now lives in a name at the call site that matters (`pub(crate) load_ledger_uncached`, with `ledger()` as the thin public wrapper — no public API break); the recover-Err comment covers the new-lookup-through-the-manager window as a distinct case; and `docs/transactions/overview.md` now documents the externally-visible change plainly: a client that disconnects mid-transact still has its data committed.

Follow-up: #1742 (Raft commit worker + the cypher stage-time mechanism), #1743 (in-place commit design — the durable end of this class), #1745 (config-cache poisoning, pre-existing, found during this pass).


fluree-db-api/src/ledger/loading.rs
```diff
@@ -57,8 +57,15 @@ impl Fluree {
 
     /// Load a ledger by address (e.g., "mydb:main")
     ///
-    /// This loads the ledger state using the connection-wide cache.
+    /// This loads the ledger state DIRECTLY from the nameservice and storage —
+    /// it does not read, populate, or lock the connection-wide ledger cache.
     /// The ledger state combines the indexed database with any uncommitted novelty transactions.
+    ///
+    /// Keep it manager-free: the cached-handle commit path calls this from
```
bplatz: This property is now load-bearing rather than incidental: `DetachedCacheSlot::recover` calls this while holding the ledger's state write lock, so routing it through `LedgerManager` later would deadlock the transact path — and the failure mode is a hang, not a red test.

The corrected doc is good (the old one said the opposite, which is its own small alarm). But a comment is thin protection for an invariant whose violation hangs the write path. Worth either a `debug_assert`, or a name that carries the constraint — `load_ledger_uncached` or similar — so someone who doesn't read the comment still can't get it wrong.

fluree-db-api/src/tx_builder.rs
```diff
@@ -119,6 +122,171 @@ fn is_retryable_commit_conflict(e: &ApiError) -> bool {
...
+                // *already* holding this handle read the placeholder until they
+                // re-fetch. It takes a double failure to get here (the commit
+                // failed and the reload failed), and with storage down there is
+                // no good state left to install.
+                tracing::warn!(
+                    ledger_id = %self.ledger_id,
+                    %error,
+                    "could not reload the ledger cache after a failed commit; evicting"
+                );
+                drop(guard);
```
bplatz, line 240: Small gap between the documented residual and the code: the comment above covers callers *already holding* this handle, but between this `drop(guard)` and the `disconnect` below, a **new** lookup through the manager gets the cached entry with the placeholder still in it.

Double-failure path and a tiny window, so I wouldn't restructure for it — just widen the comment so the next reader doesn't have to re-derive that the two cases differ.

fluree-db-api/src/tx_builder.rs
```diff
@@ -1578,27 +1778,103 @@ impl Fluree {
...
+    /// Spawned by [`commit_and_finalize`] rather than awaited inline, because
+    /// this span must not be cancellable. `LocalCommitter::transact` awaits the
+    /// commit inside the HTTP request future and axum drops that future when a
+    /// client disconnects (a `tokio::time::timeout` around `execute()` does the
+    /// same). Dropping the caller's future mid-window would run the slot's
+    /// `Drop`, which releases the write lock *before* its out-of-band reload
+    /// lands — handing the genesis placeholder to every reader parked behind
+    /// the commit, i.e. silently empty results at `t = 0` until the reload
+    /// completes. Spawning means a cancelled caller abandons the *wait* for the
+    /// commit, never the commit itself: the slot always runs to completion and
```
bplatz, line 1860: This is the right trade, but it's a real change in externally-visible behaviour that lives only in a code comment: a client that disconnects mid-transact now still has its data written, where before the outcome was undefined.

That's the kind of thing an operator discovers from a support ticket rather than from source. Worth a line wherever transact durability is documented.

bplatz (approved):
Approving. The concurrency reasoning holds up and the tests genuinely pin it — I ran two mutants rather than reading the writeup:

- **Shield removed** (awaited `commit_shielded` inline instead of spawning) → `cancelled_commit_never_exposes_the_empty_cache_slot` FAILS with *"a reader parked behind a cancelled commit observed the empty cache slot at t = 0"*.
- **Detach disabled** (`mem::replace` made a no-op) → `cached_handle_commits_uniquely_own_the_dictionaries` FAILS.

Both fail for exactly the reason they exist. The cancellation test is the strongest thing here — that's a bug that would otherwise only ever show up as intermittent empty reads under client disconnects.

I also checked the two claims the whole design rests on, independently:

- **"Every state reader routes through that one lock"** — holds. Every read path goes through `inner.state.read().await` and `lock_for_write()` takes the same RwLock, so readers park rather than see the placeholder. `config_cache` is the one thing read without the state lock, but its key is `Novelty::config_write_t`, which needs the state lock to obtain — unreachable during the window, and a poisoned key-0 entry would self-heal on the next miss anyway.
- **"A t=0 placeholder always loses the monotonic-swap guard"** — holds, and slightly better than claimed. `ledger_manager.rs:1560` is `new_state.t() >= write_guard.state().t()`. The `>=` rather than `>` means even a reload of a genuinely-empty ledger at t=0 lands, so there's no first-commit-panic edge case.

`grp_transact` 190/190, the three cow tests green, CI green.

Three inline notes below to look at before merging — none of them structural.

Two on the writeup rather than the code:

**The title claims more than ships.** "stop per-commit DictNovelty deep-clones" is true for serial writers only; your own sweep says 2 submitters leaves 94% of commits still cloning, 4 leaves 98%. The body is scrupulous about this, but the title is what lands in the changelog and the merge commit. Worth scoping it — "for serial writers", or similar.

**No `Follow-up: #N` for any of the deferrals.** The body is `Closes #1707` only, but this defers the five sibling commit flows (`merge`, `rebase`, `revert`, `cypher_txn`, Raft `commit_worker` — with cypher and the Raft worker carrying real write volume by your own note), the concurrent-writer case, and the in-place-commit design conversation. Per CLAUDE.md that's exactly what the marker exists for, so a deferral isn't indistinguishable from a fix. Right now this PR body is the only record of all of it.

Minor, on the numbers: the 2.1x is measured with fsync off, which you disclose and which is the right call for isolating the window — but it means the ratio shouldn't be quoted as a production figure, since the absolute ~1.7 ms/commit saving at the tail stays constant while the denominator gets much larger with fsync on. Also unexplained: after-fix still sits ~0.09-0.15 ms/commit above the never-cached control, of which the shield accounts for 0.02-0.03. It's flat across bands so it isn't the quadratic returning, but it's a third of the remaining gap to the floor.

fluree-db-api/src/ledger/loading.rs
aaj3f: Agreed the comment was thin protection for a hangs-not-fails invariant. Renaming the public method would break embedders, so the shape that landed (`35fd8cc3d`): the direct-load body moved to `pub(crate) fn load_ledger_uncached()` — the constraint now lives in the name — with `ledger()` as a thin public wrapper, and `recover()` calls the crate-private name. So the caller that depends on the invariant can't reach the wrong thing without reading past a name that says exactly what it guarantees, and the doc carries the deadlock rationale ("recover() calls this under a ledger's state write lock; routing through LedgerManager would deadlock").

While verifying the surrounding safety story we also tightened one corner of it: the "poisoned key-0 entry self-heals on the next miss" reasoning doesn't hold on the panic path (reload never touched `config_cache`, and a config-quiet reloaded state legitimately carries marker 0 — so the entry would *hit*, not miss). `config_cache_clear()` now runs at reload's swap point (`f2b226e45`). It takes panic + racing reader + config predating `index_t` to matter, but fixed beats argued.

fluree-db-api/src/tx_builder.rs
aaj3f: Widened as suggested (`801fbf8b2` carries it): the comment now covers the new-lookup-through-the-manager window between `drop(guard)` and `disconnect` as its own case, distinct from callers already holding the handle — and the struct-level "never observable" claim is tempered accordingly for the eviction repairs. Agreed it wasn't worth restructuring for; the double-failure framing stands.

fluree-db-api/src/tx_builder.rs
aaj3f: Right — that line deserved better than a code comment. `docs/transactions/overview.md`'s durability section now states it plainly (`d19c40344`): a client that disconnects mid-transact still has its data committed (previously the outcome was undefined), and timed-out clients should re-query or retry idempotently rather than assume the write was lost. That's the operator-facing sentence the support ticket would have needed.

aaj3f:
All of it is in, and your review directly produced two findings beyond its own asks — the second one closes a question you raised.

Title rescoped to "for serial writers" as you suggested, and the deferrals now have their markers: Follow-up: #1742 (Raft worker), #1743 (the in-place-commit design conversation, which your concurrent-writer point is the strongest argument for), plus #1745 (a pre-existing config-cache poisoning found during this pass). On the fold-in front we went further than the markers: **merge and revert now take the detach too** (spawn-shielded, with an evict-not-reload failure repair because both flows roll the head back after an apply error — reload-first would cache the head the rollback is about to withdraw; `it_branch_op_cow`, mutation-verified). Two of the five deferrals turned out to have false premises when verified rather than assumed: rebase's replay base is a direct storage load (never a cache clone — a detach there is pure risk), and cypher's clone happens per-statement at *stage* time outside any lock — a different mechanism that now sits on #1742's list.

Your "unexplained ~0.09–0.15 ms above the never-cached control" — that was the thread worth pulling, and it wasn't partial: `to_ledger_state()` hardcoded `head_temporal: None`, so every optimistic commit re-read and re-decoded the **head commit blob from storage inside the locked window**. Fixed in `afcd3cb51`; verified with paired interleaved binaries + an owned-mode negative control, and the delta scales linearly with head-commit size (~0.08 ms at 20-node commits → ~0.55 at 200) — a per-commit blob read's signature, not lock or spawn overhead. Post-fix, cached sits ~0.01–0.03 ms above the control: approximately the shield alone. Fair point taken on the fsync-off 2.1× as well — the body now scopes it, and the absolute tail saving is the honest production framing.

Head is `afcd3cb51`; full api + consensus suites, clippy, workspace check, fmt all green; CI running.

PR #1719, opened 2026-08-27
fix(server): surface novelty backpressure as 503 + Retry-After with err:db/NoveltyAtMax

### LINKED ISSUE #1708: NoveltyAtMax backpressure surfaces as HTTP 400 InvalidTransaction — retrying clients silently drop the write
Small one, but I think it's a genuine footgun for any client doing sustained writes.

When novelty hits `reindex_max_bytes`, the backpressure check is a hard error (no queue, no wait — `fluree-db-transact/src/stage.rs:626-632`), and by the time it reaches the HTTP surface it's been folded into the catch-all transact error arms (`fluree-db-server/src/error.rs:149`, `:250`): the client sees **HTTP 400 `err:db/InvalidTransaction`**. The retryable carve-outs at `error.rs:144-147` cover only `CommitConflict`/`PublishLostRace`/`NamespaceConflict`, so `NoveltyAtMax` — which is the *most* retryable condition we have, "the indexer just needs a moment" — is excluded.

Consequence: a well-behaved client that retries 5xx and treats 4xx as permanent (i.e. basically every HTTP client) **drops the write**. For an ingest pipeline that's a silent-data-loss shape: the writes that arrive while the indexer is catching up get classified as malformed transactions and discarded by the caller.

The pieces to fix it mostly already exist, which is the mildly frustrating part:

- `err:db/NoveltyAtMax` is defined at `fluree-vocab/src/errors.rs:62` and referenced **nowhere**.
- The docs at `docs/operations/configuration.md:405-412` say reaching `reindex_max_bytes` "blocks commits until reindexed" — it doesn't block, it errors. The comment at `fluree-db-api/src/server_defaults.rs:44` says "block" too. So the docs describe the behavior we probably *wish* we had, which is its own small trap for anyone capacity-planning off them.

Suggested shape (happy to take this myself; it's small):

1. Map `TransactError::NoveltyAtMax` to **503 + `Retry-After`** with the already-defined `err:db/NoveltyAtMax` code, leaving 400 for actually-invalid transactions.
2. Fix the two "blocks" claims in docs/comments to describe the real (erroring) behavior — or, if we'd rather make the docs true instead, that's a bigger conversation about a bounded wait on the commit path, which I don't think we should couple to this fix.
3. A regression test pinning the status code + error code, since the current mapping exists only by falling through catch-alls.

This is more of a bug report than a design proposal — but if anyone feels 503 is wrong for "storage-layer says slow down" (429 is the other candidate), happy to talk it through before the PR.

aaj3f:
Closes #1708.

The issue's sketch was three lines; the fix grew a layer deeper for a reason worth stating up front: **every transact surface (JSON-LD, Turtle, SPARQL UPDATE) routes through the consensus committer, which flattens `ApiError` into `SubmissionError::Execution { status, message }` — the variant's identity dies there**, and status alone can't recover it (the Raft paths emit other, differently-caused 503s). So a mapping-only change at the server edge would have been inert on the paths that matter. The fix threads identity through explicitly, and the final mapping is:

| condition | status | `@type` | `Retry-After` |
|---|---|---|---|
| `NoveltyAtMax` (stage-time: novelty already at ceiling) | **503** | `err:db/NoveltyAtMax` | `3–8` (jittered) |
| `NoveltyWouldExceed`, drainable (`delta < max`) | **503** | `err:db/NoveltyAtMax` | `3–8` (jittered) |
| `NoveltyWouldExceed`, oversized (`delta >= max` — can never succeed by drain, since the commit check is `current + delta >= max` even against empty novelty) | **413** | `err:db/NoveltyDeltaTooLarge` (new vocab code) | none |
| genuinely invalid transaction | 400 | `err:db/InvalidTransaction` | none |

Both novelty variants had to come along: `WouldExceed` is the commit-time predictive check and in a plain HTTP workload it fires *first* — mapping only `AtMax` would leave the exact footgun in place (in-repo precedent: `classify_novelty` in `r2rml_materialize.rs` documents a production incident where handling one variant made a backpressure fix inert). The oversized case is **typed end-to-end** (`SubmissionError::NoveltyDeltaTooLarge` → `ServerError::NoveltyDeltaTooLarge`) rather than a status-map passthrough, deliberately: `read_limited_body` already emits 413 for the HTTP body-size limit, and a status-keyed arm would have mislabeled those. `Retry-After` attaches to **every** 503 this server emits and only to 503s — they are all retryable capacity conditions, RFC 9110 wants the pairing, and the 413 never carries it (pinned by unit assertion). The value is a **conservative jittered constant, uniform in [3, 8] seconds per response** (review feedback: many clients honor `Retry-After` in preference to their own exponential backoff, so the smallest non-zero value is the most aggressive possible choice during exactly the stressed window; jitter de-choruses synchronized retries). The every-503-is-retryable invariant is no longer comment-enforced: an allowlist + `debug_assert` at the attachment point, a corpus test over all six 503-producing shapes, and a no-wildcard `SubmissionError` match mean a future 503-producing variant fails to compile or fails a test until consciously classified.

**Idempotency semantics — the part internal review caught.** The first cut of this branch marked the novelty refusals *unsettled*, intending "retry re-executes after drain." That did the opposite: `record_outcome` skips unsettled errors while the in-flight guard unconditionally disarms its cleanup, so a keyed refusal pinned the slot as InFlight for the 1-hour TTL and a keyed client obeying our own `Retry-After` got **409 AlreadyInFlight on every retry** — worse than pre-PR for exactly the keyed ingest clients this issue is about. Internal review proved it with a failing probe. The refusals are now **settled** — a settled Failed entry is re-attemptable via `try_claim_slot`'s body-hash-matching replace, which is what actually yields retry-after-drain semantics — and the property is pinned three ways: a consensus-level keyed-retry test (scripted committer, executor invoked exactly twice), a real-HTTP keyed-retry integration test whose mutation failure reproduces the 409 symptom verbatim, and the settlement-classification unit pin. No double-apply risk: both novelty raise sites precede commit construction, and Raft never produces the variants (unsettled stays reserved for its maybe-still-commits transitions).

**Callers of the old 400 shape, checked:** CLI uses only generic status handling; peer forwarding relays status + headers so 503/Retry-After passes through; `classify_novelty` matches variants, not codes — and its split-the-chunk answer to `WouldExceed` is exactly right for the oversized case. One residual stated honestly: on the **Raft** path a worker-side novelty failure still surfaces as a generic 422 via poison records (the replicated record carries only a reason enum — threading novelty identity through Raft is real scope beyond this fix; single-node/local, the issue's scenario, is fully covered).

**Docs:** the false "blocks commits until reindexed" claim existed in *four* places, not the two the issue found — `docs/operations/configuration.md`, the `server_defaults.rs` doc comment *and* its generated config-file template, and the `ServerConfig` field doc — all now describe the real erroring behavior and the new semantics.

**Tests** (integration in `fluree-db-server/tests/integration.rs`, running in `cargo test -p fluree-db-server` and CI): stage-time `AtMax` provoked the way production hits it (commit under a permissive threshold, reopen the same storage path with a 1-byte threshold so replay-on-load fills novelty) → 503 + code + header; the oversized case (1-byte ceiling, first insert) → 413 + `NoveltyDeltaTooLarge` + no header; the keyed-retry test above; unit pins for all three error shapes, the drainable/oversized routing at the flattening point, and header presence-on-503/absence-elsewhere. A genuinely-drainable `WouldExceed` over real HTTP isn't cheaply constructible (needs byte-exact flake sizing), so that leg is covered by the AtMax integration test plus the unit-level split pins — the 413 test's doc comment says exactly this.

**Gates** (Claude ran these; pasting the tally): workspace `cargo check --all-targets` green with `--exclude fluree-search-httpd --exclude fluree-search-service` (local C++ toolchain issue in the usearch build script, unrelated — CI is authoritative for those two); clippy `-D warnings` green across fluree-vocab/api/consensus/server/transact; `cargo test -p fluree-db-server` full package green (incl. the new tests); `-p fluree-db-consensus` 56/56; `-p fluree-db-api --lib` 775/775; `-p fluree-vocab` 20/20; fmt clean after the last edit.

---

**Post-review addendum** (commits `e7466dd96`, `f177abf9a`): the docs reconciliation went deeper than the review's two rows — `err:db/PayloadTooLarge` was *also* fictional (the body-limit 413 actually fell through the same catch-all to `err:system/InternalError`, and `docs/api/headers.md` carried a third spelling, `err:http/PayloadTooLarge`). Rather than document that absurdity, this now makes the long-documented code real: `PAYLOAD_TOO_LARGE` added to `fluree-vocab` with a `413 => PAYLOAD_TOO_LARGE` arm for the body-limit path, so the two 413s are genuinely distinguishable by `@type` — which is the point of the split. Also surfaced: there are **four** 503 sources, not three — `SubmissionError::Overloaded` reaches clients as 503 + the internal catch-all `@type`; it *is* retryable capacity so the blanket `Retry-After` stays sound, and it's now consciously pinned in the allowlist and documented honestly (a typed code for it is noted as a residual). `docs/api/errors.md` / `headers.md` / `common-errors.md` now match emitted reality for every status this PR touches.

Follow-up: #1741 (Raft path — worker-side novelty failure still surfaces as a generic 422 via poison records; wire-format change, deliberately not smuggled in here).


fluree-vocab/src/errors.rs
```diff
@@ -61,6 +61,15 @@ pub const LEDGER_EXISTS: &str = "err:db/LedgerExists";
...
 pub const NOVELTY_AT_MAX: &str = "err:db/NoveltyAtMax";
 
+/// A single transaction's delta alone meets or exceeds `reindex_max_bytes`,
+/// so it can never be accepted by waiting for the indexer to drain novelty.
+///
+/// Distinct from [`NOVELTY_AT_MAX`] (retryable backpressure, HTTP 503 +
+/// `Retry-After`): this is a permanent refusal of this transaction at this
+/// configuration, surfaced as HTTP 413 with no `Retry-After` — split the
+/// transaction into smaller pieces or raise `reindex_max_bytes`.
+pub const NOVELTY_DELTA_TOO_LARGE: &str = "err:db/NoveltyDeltaTooLarge";
```
bplatz, line 71: New code with no user-facing documentation. `docs/api/errors.md` is untouched and currently documents different `@type`s for both statuses this PR uses:

| status | errors.md says | this PR emits |
|---|---|---|
| 503 | `err:db/ServiceUnavailable` | `err:db/NoveltyAtMax` |
| 413 | `err:db/PayloadTooLarge` | `err:db/NoveltyDeltaTooLarge` |

`err:db/ServiceUnavailable` isn't defined in `fluree-vocab` at all, so that example is a code the server can't emit. The 413 collision is the sharper half — same status, two `@type`s, only one documented.

Since the point of the PR is that clients branch on `@type`, a developer following the docs writes a handler matching neither new case.

fluree-db-server/src/error.rs
```diff
@@ -386,7 +448,20 @@ impl IntoResponse for ServerError {
...
 
-        (status, [("content-type", "application/json")], json).into_response()
+        let mut response = (status, [("content-type", "application/json")], json).into_response();
+        // Every 503 this server emits is retryable capacity pressure
+        // (novelty at max, materialization deferred, committer overloaded),
+        // and RFC 9110 says a 503 should tell the client when to come back.
+        // Static 1s: the indexer's drain time is not observable at the
+        // error site, and 1s is the shortest interval that stops a tight
+        // retry loop without penalizing recovery.
+        if status == StatusCode::SERVICE_UNAVAILABLE {
```
bplatz, line 470: Worth reconsidering the value. Many HTTP clients honour `Retry-After` in preference to their own exponential backoff, and novelty drain on a large ledger is seconds to minutes — so this pins every blocked client at 1 rps of rejected requests during exactly the window the server is already capacity-stressed. That can be worse than sending no header, which leaves clients on their own backoff.

"Drain time isn't observable at the error site" argues for a conservative constant rather than the smallest non-zero one. A larger default with jitter would be gentler.

Separately: this keys purely on status, so any future 503 inherits the header automatically. The "every 503 here is retryable" invariant is now load-bearing and enforced by a comment.

bplatz (approved):
Approving. Verified the bug and the fix rather than reading the writeup.

**The bug is real, and worse than #1708 describes.** Reverting all six source files with the tests kept, the actual pre-fix response is:

```json
{"error":"Transaction error: Novelty at maximum size, reindexing required",
 "status":400, "@type":"err:system/InternalError"}
```

Not `err:db/InvalidTransaction` — `err:system/InternalError`. The issue analysed the direct `ApiError::Transact` path; the consensus-flattened path every HTTP route actually uses degrades further. That confirms the PR's claim that an edge-only mapping change would have been inert.

**The `delta >= max` boundary checks out** against `commit.rs:643` (`current_bytes + delta_bytes >= max_bytes`), including the `>=` rather than `>`.

**The idempotency fix is genuinely pinned.** Flipping the `is_settled` novelty arm to `false`:

```
[consensus] keyed_retry_after_novelty_refusal_reexecutes ... FAILED
            keyed retry must re-execute and succeed: AlreadyInFlight
[http]      novelty_backpressure_keyed_retry_reexecutes_not_409 ... FAILED
            {"status":409,"@type":"err:db/CommitConflict"}
```

Both levels reproduce the 409 verbatim. Also verified all three 503 sources (`NoveltyDeferred`, deferral-only `MaterializePartial`, novelty backpressure) are genuinely retryable, so the blanket `Retry-After` isn't over-broad. `fluree-db-server` and `fluree-db-consensus` green, CI green.

Two inline notes below, plus one convention item: no `Follow-up: #N` for the Raft residual. It's disclosed honestly in the prose, but a worker-side novelty failure still surfaces as a generic 422 there — the issue's footgun, unfixed, on the clustered path. Per CLAUDE.md that's what the marker is for.

fluree-vocab/src/errors.rs
aaj3f: This was the sharpest catch of the review — and when we pulled the thread it went further than your table. `err:db/PayloadTooLarge` turns out to be fictional too: the body-limit 413 was falling through the same `Http` catch-all to `err:system/InternalError`, and `docs/api/headers.md` carried a *third* spelling, `err:http/PayloadTooLarge`. So before this PR, neither 413 nor 503 ever emitted an `@type` the docs mentioned.

One judgment call to flag for your veto: rather than document the absurdity ("the body-limit 413 carries `err:system/InternalError`"), we made the long-documented code real — added `PAYLOAD_TOO_LARGE` to `fluree-vocab` and a `413 => errors::PAYLOAD_TOO_LARGE` arm in `error_type()`'s `Http` match, same pattern as the existing 401/403/409/422 arms, pinned by `body_limit_413_carries_payload_too_large_not_the_novelty_code`. That's ~10 lines beyond the strict scope here, but it's what makes the two 413s genuinely distinguishable by `@type`, which is the whole point of the split. Easy to revert to docs-only if you'd rather.

`errors.md`, `headers.md`, and `common-errors.md` now match emitted reality for every status this PR touches, and a repo sweep shows zero remaining references to either fictional code. (The broader fiction in `errors.md` — `err:db/BadRequest`, `err:db/Timeout`, `err:db/RateLimited` etc. all diverge from vocab — is real but pre-existing and untouched; flagging rather than folding since it's outside every status this PR changes.) Commits `e7466dd96` + `f177abf9a`.

fluree-db-server/src/error.rs
aaj3f: You're right, and the honor-in-preference-to-backoff point is the part I hadn't weighed — it means the smallest non-zero value is the *most aggressive* possible choice during exactly the stressed window, which is backwards. Now a uniform random integer in [3, 8] per response (`rand` is already a direct dep; consts + rationale comment in `error.rs`), with the unit and integration tests asserting presence + range rather than an exact value.

On the invariant being comment-enforced: also fixed, three layers deep — a `RETRYABLE_503_TYPES` allowlist with a `debug_assert` at the attachment point, a corpus test over every 503-producing shape, and a no-wildcard `match` over `SubmissionError` so a future variant refuses to compile until someone classifies it.

Chasing that invariant surfaced something your enumeration (and ours) missed: there are **four** 503 sources, not three. `SubmissionError::Overloaded` (the committer's in-flight cap) reaches clients as 503 + the `err:system/InternalError` catch-all `@type`. It *is* retryable capacity per its own contract, so the blanket `Retry-After` stays sound — it's now consciously pinned in the allowlist and documented honestly, with a note that it deserves a typed code someday rather than the internal catch-all.

aaj3f:
Thank you for verification-grade review — reverting six files to reproduce the pre-fix response is above and beyond, and it paid off: Claude re-verified your repro against `origin/main` and it's exact, including the message string. The chain: `NoveltyAtMax` → `ApiError::Transact` catch-all 400 (`fluree-db-api/src/error.rs:653`) → `SubmissionError::Execution{status:400}` → reconstructed as `ApiError::http(400, …)` → `error_type()`'s `Http` arm mapped only {401, 403, 409, 422}, so 400 fell to `err:system/InternalError`. Which means #1708 as I filed it *understated* the bug — the `err:db/InvalidTransaction` claim held only on the direct path HTTP routes never take. I've corrected the record on the issue.

Both inline items are addressed (details on each thread), the docs reconciliation went deeper than expected — see the addendum in the PR body, including one small scope judgment (making `err:db/PayloadTooLarge` real rather than documenting the catch-all) I'd like your eyes on — and the Raft residual now has its marker: Follow-up: #1741. Gates re-run green across server/consensus/vocab plus the workspace check; CI is on the new head.

PR #1740, opened 2026-08-28
release v4.2.0

bplatz:
Version bump `4.1.6` → `4.2.0`. Minor rather than patch because this cycle adds new backward-compatible functionality; nothing in it is breaking.

Release notes are generated from merged PR titles at tag time, so the changelog reflects the final contents once this merges and publishes.

## Notes for review

`testsuite-sparql` and `testsuite-shacl` are workspace-*excluded* and carry independent versions and lockfiles, so a root-only `cargo update --workspace` does not reach them — hence six files rather than two.

One line in the diff is not a version bump: `testsuite-sparql/Cargo.lock` gains a `rand 0.8.5` edge under `fluree-db-core`. That dependency was added as an optional `native`-feature dep in bf63e54e6 (#1680); the root lockfile was refreshed then, the excluded one was not. This picks up that correction. No other dependency moved — every remaining hunk across the three lockfiles is a bare `version =` line, and all three resolve under `cargo metadata --locked`.


aaj3f (approved):
🚀

PR #1746, opened 2026-08-28
fix: fail closed on unencodable datatype IRIs; keep the membership-join lane join-equivalent on list rows

### LINKED ISSUE #1686: Unencodable datatype IRI in a triple pattern falls back to xsd:string instead of matching nothing
Raised in review of #1676, not introduced by it.

## Problem

In `fluree-db-sparql/src/lower/term.rs`, the `LiteralValue::Typed` arm resolves the datatype IRI to a SID and falls back to `xsd:string` when the encoder returns `None`:

```rust
let dt_sid = self
    .encoder
    .encode_iri_strict(&dt_iri)
    .unwrap_or_else(|| Sid::new(XSD, xsd_names::STRING));
```

So `?s ex:p "a"^^ex:NoSuchType` matches plain `"a"` rows. The spec answer is no match — no such term exists.

## Why the fallback is not defensible

`encode_iri_strict` (`fluree-db-core/src/db.rs:436`) returns `None` exactly when the IRI's canonical prefix is **not a registered namespace on this ledger**. When the fallback fires, no stored row can carry that datatype, so "no match" is provably correct rather than just spec-nicer.

The comment above the fallback justifies it as "matching the storage-side fallback in `term_to_binding`", but `term_to_binding` has no custom-datatype fallback — its `xsd:string` is the `LiteralValue::Simple` arm. The stated reason does not hold.

## Why it was not fixed in #1676

`lower_literal_with_constraint` returns `Option<DatatypeConstraint>`, where `None` means *no constraint* — fully lenient, the opposite of what is wanted. Expressing "this pattern can never match" needs either:

- a `DatatypeConstraint` variant that matches nothing, or
- an early empty-result path in lowering, so the pattern is recognized as unsatisfiable before it reaches the scan.

Either is a design change rather than a fold-in, so it wants its own PR.

## Note on severity

This is pre-existing and #1676 made it **strictly narrower**, not worse: before that change the pattern matched everything, now it matches plain `"a"` rows. So this is a correctness cleanup, not a regression.

## Origin

Review of #1676 (`fix: match string literals by RDF term identity; normalize language-tag case`).

aaj3f:
Fixes #1686. Partially addresses #1687 — the join-equivalence half; the list-position semantics decision stays open there (details below).

Both of these came out of the same review pass over #1674/#1676 and sit on the same literal-identity seam — what a constant literal's datatype *means* when it meets the scan — so they ship together.

## Problem

**1. Unencodable datatype IRIs failed open (#1686).** Both term-identity sites in SPARQL lowering — the triple-pattern constraint in `lower_literal_with_constraint` and its `term_to_binding` VALUES twin (`fluree-db-sparql/src/lower/term.rs`) — resolved a typed literal's datatype with `encode_iri_strict` and fell back to `xsd:string` when the namespace wasn't registered. So `?s ex:p "a"^^ex:NoSuchType` matched every plain-string `"a"` row, and the VALUES form did the same. `encode_iri_strict` returning `None` means exactly "this ledger has no such namespace", and ingest registers every namespace it stores — so no stored term *on that ledger* can carry that datatype, and for a single-ledger query the empty result isn't just the spec answer, it's provably the only correct one. (In a multi-ledger dataset the constraint is resolved against the primary ledger only and a *non-primary* ledger can carry the datatype — that seam predates this PR and is #1771; see Not in this PR.) The JSON-LD surface already gave it (its `lower_triple_pattern` uses the non-strict encoder), which means the two query surfaces disagreed on the same question — verified live before the fix: SPARQL 1 row, JSON-LD 0.

**2. The membership-join lane silently diverged from the generic join on list rows (#1687).** `MembershipJoinOperator` answers a right triple that binds nothing new as a hash keep/drop, on the premise that a fully-ground triple matches at most once. RDF list rows break that premise inside a single graph — the same `(s, p, o)` recurs at multiple `o_i` positions — and nothing in `membership_join_key_vars` excluded them, so above the 256-row driving gate the lane answered a var-object join over list elements as a semi-join while the generic pipeline answers a bag join. Reproduced above the gate in the new fixture: 276 rows from the lane vs 532 from the generic pipeline on identical data — a scale-dependent silent wrong answer, and a break of the lane's own documented "join-equivalent unconditionally" invariant. On top of that the lane was untestable: no `stamp_fast_path` site, and `FLUREE_DISABLE_QUERY_FAST_PATHS` didn't reach it, so `MustFire`/`MustNotFire` couldn't express engagement at all.

## Changes

**Commit 1 — fail closed on unencodable datatypes (`fluree-db-sparql`)**
- Both sites now resolve through a shared `datatype_sid` helper: non-strict encode, so an unregistered namespace yields the EMPTY-namespace full-IRI Sid — a constraint no stored row carries. Every enforcement lane compares via `dt_compatible` (exact Sid equality outside the XSD numeric families: `range.rs:289` novelty, `binary_scan.rs:846` overlay, `binary_scan.rs:1287` indexed), so the pattern matches nothing in all of them, and the SPARQL answer now equals the JSON-LD one by construction.
- Registered custom datatypes are untouched — their query-time encode resolves to the same canonical Sid ingest stored, pinned by controls.
- A datatype IRI that canonically splits to the EMPTY prefix on *both* sides (a bare-string datatype actually stored that way) still matches by exact equality, as before.
- Unit test pins the lowered constraint form at both sites; integration guards in `it_literal_identity.rs` pin the empty result and the controls on both surfaces, in both lanes (novelty + indexed).

**Commit 2 — keep the membership lane join-equivalent (`fluree-db-query`)**
- Stats carry no list-ness, so the shape is not plan-detectable. Instead the one-shot drain that builds the key set doubles as the detector: a key tuple inserted twice **is** the premise's counterexample (two flakes matched one ground triple). When that happens the operator discards the hash path and routes every row through its existing exact per-row fallback — the same seeded evaluation the generic join runs — so lane and generic agree by construction, whatever the eventual list-semantics ruling says. The detection is exact, not heuristic: keep/drop diverges from the generic join iff some ground probe matches more than one flake, and that's precisely when the drain sees a duplicate key.
- The decline is surgical: on ordinary shapes the added cost is one bool OR per drained key, and the new fixture pins (via routing stamp) that a same-size non-list shape still takes the hash path.
- Testability: `membership_join.rs` now stamps the repo-standard fast-path outcome (site `membership-join` — `Proceed` when the hash path serves, `GateDeclined` on the duplicate-key or multi-graph fallback), and the planner honors `FLUREE_DISABLE_QUERY_FAST_PATHS` for the lane (stamping `KillSwitch`, checked at the choose site so `membership_join_key_vars` stays a pure predicate). The differential harness can finally compare this lane against the generic join.
- New own-binary test `it_membership_join_list_rows` runs both cases *above* the gate (296/280 driving rows) with span-captured stamps, so neither case can silently degrade into a below-gate run and pass vacuously; phase 2 re-answers both under the kill switch and requires identical rows. The stale "the lane's answer at or above the gate is unpinned" comment in `it_fastpath_1652_regression.rs` now points at the new pin.

Both fixes were reverted independently during development to confirm their tests fail for the right reason: the #1686 guards reproduce the 1-row fail-open match, and the #1687 fixture reproduces the 276-vs-532 divergence with the lane engaged.

## Behaviour changes (release notes)
- SPARQL: a typed literal whose datatype IRI has no registered namespace on the ledger now matches nothing (previously: matched plain `xsd:string` rows), in triple patterns and VALUES alike. Registered custom datatypes are unaffected.
- A var-object join over RDF list elements now returns the same rows whether or not the planner picks the membership lane. Below the driving gate nothing changes; above it the lane previously under-counted (one row per driving row instead of one per matching flake).
- `FLUREE_DISABLE_QUERY_FAST_PATHS` now also disables the membership-join lane.

## Not in this PR
- **The list-position semantics decision (#1687's step 1).** Whether a var-object join over list elements *should* be a bag join over positions (532/5-style answers) or a set join over distinct values is still the open product call — this PR only guarantees both pipelines give the same answer, which today is the generic bag join. The two fixtures pin today's numbers with comments saying exactly that, so settling the decision the other way means changing pinned expectations deliberately, not silently. That's why this is "partially addresses": the issue's remaining substance is that decision, and it shouldn't get closed as a side effect of the equivalence fix.
- `term_to_binding` special-cases `fluree:embeddingVector` → `Sid(FLUREE_DB, "vector")` before encoding while the triple-pattern site doesn't; that asymmetry predates this PR and is orthogonal to the fail-open bug, so I left it alone rather than touch vector matching without a vector rig.
- **Multi-ledger datasets: the constraint Sid is resolved against the primary ledger only.** `query_dataset` lowers against the first default graph's snapshot, and per-graph execution re-encodes `s`/`p`/`o` Sids but not `pattern.dtc` (`binary_scan.rs` `build_match_val_for_snapshot`), so a datatype registered only on a *non-primary* ledger constrains as the primary encoded it — confirmed by fixture: this PR turns that from matching the primary's plain-string rows (wrong rows) into matching nothing (silent drop), while registered-on-both datatypes with divergent namespace codes were already dropped before it. The re-encode fix belongs with the pre-existing gap, and this PR's EMPTY-namespace Sid keeps the datatype IRI recoverable for it where the `xsd:string` fallback destroyed it. Follow-up: #1771.

## Gates
- `cargo test -p fluree-db-query` (1,485 lib + integration bins) and `-p fluree-db-sparql` (641): green.
- `-p fluree-db-api --features native`: `it_literal_identity`, `it_fastpath_1652_regression`, `it_membership_join_list_rows`, `it_cyclic_bgp_string_dict`, `it_query_sparql_indexed`: green.
- W3C suite (`testsuite-sparql`, both-way registers): 36/36 groups green.
- `cargo clippy --all-targets --no-deps -- -D warnings` on all three touched crates; `cargo fmt --all --check` clean.

### DIFF
--- a/fluree-db-query/src/membership_join.rs
+++ b/fluree-db-query/src/membership_join.rs
@@ -224,9 +224,11 @@ impl MembershipJoinOperator {
         Ok(())
     }
 
-    /// Exact join fallback for a row that does NOT fully ground the
-    /// pattern: evaluate the triple seeded with the row and emit one
-    /// output row per match, with produced bindings filled in.
+    /// Exact join for a row the keep/drop drain cannot answer: one that
+    /// does NOT fully ground the pattern, or — once the drain declines
+    /// (`exact_only`: duplicate ground rows in the key set) — every row.
+    /// Evaluates the triple seeded with the row and emits one output row
+    /// per match, with produced bindings filled in.
     async fn per_row_join(
         &self,
         ctx: &ExecutionContext<'_>,
@@ -369,7 +371,17 @@ impl Operator for MembershipJoinOperator {
     }
 
     fn estimated_rows(&self) -> Option<usize> {
-        // A filter: at most the child's cardinality (ground rows match ≤ once).
+        // Drain lane (keep/drop): a filter — at most the child's cardinality.
+        // Per-row-join fallback (`exact_only`, entered mid-execution when the
+        // drain finds duplicate ground rows): a bag join whose output can
+        // exceed the child's cardinality (the #1687 fixture emits 532 rows
+        // from 296 driving rows), making this an under-estimate. The regime
+        // isn't known yet when this is read (EXPLAIN `describe()` and lane
+        // gates run at plan/open time, before `build_key_set`), and every
+        // consumer treats a low value conservatively — a lane gate that reads
+        // it (`SubqueryOperator` materialize eligibility, the annotation-edge
+        // hash probe) declines its optimization and stays on the exact generic
+        // path — so the child estimate remains the honest upper-bound hint.
         self.child.estimated_rows()
     }
 }
--- a/fluree-db-query/src/membership_join.rs
+++ b/fluree-db-query/src/membership_join.rs
@@ -21,6 +21,25 @@
 //! scope. So the lane is join-equivalent unconditionally, and the hash fast
 //! path engages only where the premise provably holds.
 //!
+//! **The premise can also fail inside one graph: RDF list rows.** A
+//! list-valued predicate stores one flake per position, so the same
+//! `(s, p, o)` recurs at multiple `o_i` values and the generic join emits
+//! one row per matching flake where keep/drop would emit one per driving
+//! row (#1687). Stats carry no list-ness, so this is not plan-detectable;
+//! instead the one-shot drain that builds the key set doubles as the
+//! detector — a key tuple inserted twice IS the premise's counterexample
+//! (two flakes matched one ground triple). When that happens the hash set
+//! is discarded and every row takes the exact per-row fallback, keeping
+//! the lane join-equivalent without ruling on what a var-object join over
+//! list positions should mean (that semantics call stays with #1687).
+//!
+//! Engagement is observable via [`MEMBERSHIP_JOIN_SITE`] routing stamps
+//! (the repo's `MustFire`/`MustNotFire` oracle): `Proceed` when the hash
+//! path serves probes, `Fallback(GateDeclined)` when the drain disproves
+//! the premise or a multi-graph scope forces exact mode, and
+//! `Fallback(KillSwitch)` from the planner when
+//! `FLUREE_DISABLE_QUERY_FAST_PATHS` suppresses the lane.
+//!
 //! **Rows with an unbound key var keep exact join semantics** — a join
 //! against a partially-ground pattern EXTENDS the row (possibly multiplying
 //! it), which a keep/drop semijoin cannot reproduce. Such rows fall back to
@@ -52,6 +71,14 @@ use async_trait::async_trait;
 use rustc_hash::FxHashSet;
 use std::sync::Arc;
 
+/// Routing-stamp site for this lane (see `fast_path_outcome`). Stamped
+/// `Proceed` when the hash membership path serves probes,
+/// `Fallback(GateDeclined)` when the operator routes rows through the exact
+/// per-row join instead (multi-graph scope, or the build drain disproved
+/// the matches-at-most-once premise), and `Fallback(KillSwitch)` at plan
+/// time when the kill switch keeps the lane out of the plan.
+pub const MEMBERSHIP_JOIN_SITE: &str = "membership-join";
+
 pub struct MembershipJoinOperator {
     child: BoxedOperator,
     /// The right-side triple (all vars ∈ child schema).
@@ -65,9 +92,11 @@ pub struct MembershipJoinOperator {
     /// lazily on the first fully-bound row, so a stream that never grounds
     /// the pattern (or is empty) never pays for the scan.
     key_set: Option<FxHashSet<CompositeGroupKey>>,
-    /// Set at `open()` when more than one graph is in the active scope: the
-    /// hash set is never built and every row takes the exact per-row join
-    /// (see the multi-graph note in the module docs).
+    /// When set, every row takes the exact per-row join instead of the hash
+    /// keep/drop path. Set at `open()` when more than one graph is in the
+    /// active scope, and by `build_key_set` when the drain finds a duplicate
+    /// key tuple — proof that some ground triple matches more than one flake
+    /// (RDF list positions), so keep/drop is not join-equivalent (#1687).
     exact_only: bool,
     planning: PlanningContext,
     norm: Option<EqualityNorm>,
@@ -142,6 +171,7 @@ impl MembershipJoinOperator {
             &self.planning,
         )?;
         let mut set = FxHashSet::default();
+        let mut duplicate_ground_rows = false;
         inner.open(ctx).await?;
         while let Some(batch) = inner.next_batch(ctx).await? {
             ctx.check_cancelled()?;
@@ -155,15 +185,41 @@ impl MembershipJoinOperator {
                         binding_to_group_key_normalized(binding, store, gv)
                     })
                     .collect();
-                set.insert(CompositeGroupKey(key));
+                duplicate_ground_rows |= !set.insert(CompositeGroupKey(key));
             }
         }
         inner.close();
-        tracing::debug!(
-            keys = set.len(),
-            pattern = ?self.pattern,
-            "membership join key set built"
-        );
+        if duplicate_ground_rows {
+            // A key tuple drained twice means one ground triple matched more
+            // than one flake — RDF list positions (`o_i`) are the known case.
+            // The generic join emits one row per matching flake; keep/drop
+            // would collapse them. Discard the hash path and route every row
+            // through the exact per-row fallback so the lane's answer stays
+            // identical to the generic pipeline's (#1687).
+            tracing::debug!(
+                keys = set.len(),
+                pattern = ?self.pattern,
+                "membership join key set has duplicate ground rows; \
+                 falling back to exact per-row joins"
+            );
+            self.exact_only = true;
+            crate::fast_path_outcome::stamp_fast_path(
+                MEMBERSHIP_JOIN_SITE,
+                crate::fast_path_outcome::FastPathOutcome::Fallback(
+                    crate::fast_path_outcome::FastPathFallback::GateDeclined,
+                ),
+            );
+        } else {
+            tracing::debug!(
+                keys = set.len(),
+                pattern = ?self.pattern,
+                "membership join key set built"
+            );
+            crate::fast_path_outcome::stamp_fast_path(
+                MEMBERSHIP_JOIN_SITE,
+                crate::fast_path_outcome::FastPathOutcome::Proceed,
+            );
+        }
         self.key_set = Some(set);
         Ok(())
     }
@@ -226,6 +282,14 @@ impl Operator for MembershipJoinOperator {
             crate::dataset::ActiveGraphs::Single => false,
             crate::dataset::ActiveGraphs::Many(graphs) => graphs.len() > 1,
         };
+        if self.exact_only {
+            crate::fast_path_outcome::stamp_fast_path(
+                MEMBERSHIP_JOIN_SITE,
+                crate::fast_path_outcome::FastPathOutcome::Fallback(
+                    crate::fast_path_outcome::FastPathFallback::GateDeclined,
+                ),
+            );
+        }
         self.child.open(ctx).await?;
         self.key_set = None;
         self.state = OperatorState::Open;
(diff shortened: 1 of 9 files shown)


fluree-db-query/src/membership_join.rs
```diff
@@ -260,28 +324,35 @@ impl Operator for MembershipJoinOperator {
...
+                            }
                         }
+                        continue;
                     }
-                } else {
-                    self.fallback_rows += 1;
-                    self.per_row_join(ctx, &input_batch, row_idx, &mut columns)
-                        .await?;
                 }
+                self.fallback_rows += 1;
```
bplatz, line 355: `estimated_rows()` (:371) still claims "A filter: at most the child's cardinality (ground rows match ≤ once)". This route is exactly where that stops holding — the new fixture returns 532 rows from 296 driving rows.

fluree-db-query/src/membership_join.rs
```diff
@@ -155,15 +185,41 @@ impl MembershipJoinOperator {
...
+                "membership join key set built"
+            );
+            crate::fast_path_outcome::stamp_fast_path(
+                MEMBERSHIP_JOIN_SITE,
+                crate::fast_path_outcome::FastPathOutcome::Proceed,
+            );
+        }
         self.key_set = Some(set);
         Ok(())
     }
```
bplatz, line 225: `per_row_join`'s doc just below still reads "Exact join fallback for a row that does NOT fully ground the pattern" — after this change it's also the path every fully-ground row takes once the drain declines.

bplatz (commented):
Three stale doc comments the new fallback route leaves behind — two inline, plus one outside the diff:

`fast_path_outcome.rs` `stamp_fast_path` says stamps are "called at plan time from the kill-switch gate sites and at `open()` time from `FastPathOperator`". This lane adds a third timing — mid-execution, from `build_key_set` on the first fully-bound row. Worth a clause so the next site author knows that's allowed.

Separately, filed #1765 for the third site on this seam: `ComparableValue::to_binding` (`eval/value.rs`) resolves a `STRDT` datatype with `encode_iri_strict` and raises `Unknown datatype IRI` when the namespace isn't registered — `BIND(STRDT(?v, <http://no-such-ns.example.com/T>) AS ?x)` fails the whole query, while the registered control works. Out of scope here; noting it so the seam is fully mapped.

bplatz (approved):
Couple docs updates noted but also created issue #1765 related to what you fixed here.

fluree-db-sparql/src/lower/term.rs
```diff
@@ -461,6 +454,25 @@ impl<E: IriEncoder> LoweringContext<'_, E> {
...
+    /// canonical prefix is NOT registered on this ledger resolves to the
+    /// EMPTY-namespace full-IRI Sid — the same form the storage side's
+    /// non-strict `encode_iri` produces, and the form the JSON-LD query
+    /// surface already uses (`fluree-db-query/src/parse/lower.rs`). Ingest
+    /// registers every namespace it stores, so no stored term can carry a
+    /// datatype in an unregistered namespace: the EMPTY-namespace Sid
+    /// matches nothing, which is the spec answer for a term that does not
+    /// exist. (The previous `xsd:string` fallback made
+    /// `"a"^^ex:NoSuchType` match every plain-string `"a"` row — #1686.)
+    fn datatype_sid(&self, dt_iri: &str) -> Sid {
```
bplatz, line 470: Does this premise survive a multi-ledger dataset query? "Ingest registers every namespace it stores" is true of *one* ledger, and this Sid is resolved once against the lowering snapshot.

`map_pattern_to_match` (`binary_scan.rs:1796-1830`) re-encodes the subject, predicate and object Sids into the per-graph snapshot, but not `pattern.dtc` — it goes to `dt_compatible` as the lowering snapshot encoded it. So for a datatype registered in ledger B but not in the lowering snapshot A, the constraint is now the EMPTY-ns Sid and matches nothing in B, where before it was `xsd:string` and matched the wrong rows. Both wrong, but the new one silently drops real matches instead of adding false ones.

The underlying gap (dtc never re-encoded per graph) predates this PR — registered custom datatypes are already broken cross-ledger by it — and I read this off the code rather than building a fixture, so it may well not be reachable in practice. Worth confirming either way; if it holds, a follow-up issue for re-encoding `dtc` alongside s/p/o would cover both cases.

bplatz (commented):
One question on the fail-closed choice — inline. Worth looking into before merge, and a follow-up issue is probably the right home for it if it confirms.

fluree-db-query/src/membership_join.rs
aaj3f: Right — verified the numbers against the fixture (`it_membership_join_list_rows` asserts the 532-from-296 shape and is green at head) and rewrote the comment to describe both regimes. I also chased whether the *value* now misleads anything, since doc-patching a live planner input would be the wrong fix: the consumers are `describe()`/EXPLAIN (`operator.rs:149`) and two runtime lane gates that read a child's estimate — `SubqueryOperator`'s materialize-eligibility check (`subquery.rs:237`) and the annotation-edge probe's driving-size gate (`default_graph_source.rs:204`). Both treat a low value conservatively: they decline their optimization and stay on the exact generic lane, so an under-estimate can cost a missed optimization but not correctness. And the operator can't honestly do better — the regime isn't known until `build_key_set` sees the first batch, while `describe()` and both gates read the estimate at plan/open time. So the child estimate stays, now documented as the drain lane's upper bound and the fallback regime's potential under-estimate. Fixed in 14d1e8310.

fluree-db-query/src/membership_join.rs
aaj3f: Fixed in 14d1e8310 — it now reads as the exact join for any row the keep/drop drain can't answer: one that doesn't fully ground the pattern, or every row once `exact_only` is set.

fluree-db-sparql/src/lower/term.rs
aaj3f: You're right, and it's reachable — I built the fixture rather than trust either of our readings. Two memory ledgers through the public `query_dataset` surface: A (primary) holds a plain-string `"a"` row, B holds `ex:s2 ex:p "a"^^ct:T` with `ct:` registered only on B (code 14). `SELECT ?s WHERE { ?s ex:p "a"^^ct:T }` over `[A, B]`:

- at this PR's head: `[]` — the silent drop you predicted (B alone as primary returns `s2`, so the rows are there and the union reaches them);
- with main's `xsd:string` fallback swapped back into `datatype_sid`: `[["ex:s1"]]` — the primary's plain-string row, i.e. wrong rows.

Your pre-existing half confirmed too, with a second case: `ct:` registered on **both** ledgers but under divergent codes (15 vs 14, via a padded namespace table on the primary) drops the non-primary row identically at head and on main. Full fixture + verbatim output are in #1771, filed for the re-encode-`dtc`-alongside-s/p/o fix covering both shapes. I looked at folding it in here: it's three `dt_compatible` sites (`binary_scan.rs:846`, `matches_datatype_constraint` at `:1286`, `range.rs:289`) plus threading a per-open effective dtc through the scan, which felt past the fold-in line for this PR — but happy to talk it through if you'd rather see it land here.

One point in the new behavior's favor beyond fail-closed-vs-wrong-rows: the EMPTY-namespace Sid carries the full datatype IRI in its name, so #1771's per-graph re-encode can recover the graph-local Sid from it directly — the `xsd:string` fallback destroyed that information at lowering time, so no downstream fix could ever have repaired the unregistered-on-primary shape on top of it. Kept the PR's behavior and scoped the body's "provably the only correct one" claim to single-ledger, with the multi-ledger caveat under Not in this PR. The fixture also turned up the decode-side sibling on the same surface (`to_jsonld` against the primary snapshot either errors on B's codes or silently decodes the wrong IRI) — noted in #1771 next to #1639 rather than filed separately.

aaj3f:
Thanks @bplatz — all three doc notes were right; fixed in 14d1e8310. That includes `stamp_fast_path`'s doc, which now names the third timing this PR introduced (mid-execution, from `build_key_set` on the first fully-bound row) and says outright that a new site may stamp at whichever point its gate actually decides — so the next site author isn't left inferring permission from this lane.

On your inline multi-ledger question: confirmed reachable by fixture, both failure modes as you read them — details in the thread, follow-up filed as #1771.

And thanks for filing #1765 — agreed it's the third site on this seam and the only one that fails hard, and the non-strict-encoder direction you sketched there lines up with what the two lowering sites now do. Between this PR (constraint + VALUES lowering), #1765 (expression materialization), and #1771 (per-graph re-encode), I think the seam is now fully mapped.

PR #1747, opened 2026-08-29
Temporal arithmetic per SEP-0002, and make TZ/TIMEZONE reindex-stable

bplatz:
Reported: `?d1 - ?d2` over two `xsd:dateTime`s returns unbound. It did, and
chasing it opened up more than the original report.

## What was wrong

`ArithmeticOp::apply` had no temporal arm, so a pair of dateTimes fell to the
non-numeric catch-all and raised `TypeMismatch`. That is a *dynamic value*
error, and those demote to unbound inside `BIND`/`Extend` — so the query
returned a row with an empty column and no diagnostic, which reads like missing
data rather than a missing operator.

Neither SPARQL 1.1 nor the SPARQL 1.2 editor's draft maps `-` over temporal
operands (I checked both spec texts directly: each contains exactly one
`op:numeric-subtract` and zero `op:subtract-dateTimes`), so the type error was
conformant. [SEP-0002][sep] writes down the XPath F&O semantics for these
operators. This adopts them.

[sep]: https://github.com/w3c/sparql-dev/blob/main/SEP/SEP-0002/sep-0002.md

## What this adds

**Differences** — `dateTime - dateTime`, `date - date`, `time - time` → a
signed `xsd:dayTimeDuration`. (See the update below for how offset-carrying
`date`/`time` operands are read.)

**Shifts** — `± xsd:dayTimeDuration` / `± xsd:yearMonthDuration` over
`dateTime`/`date`/`time`, plus duration ± duration within one family, each
yielding the left operand's kind. Calendar-aware in three ways that each have a
test: month arithmetic clamps to the end of the month (`2026-01-31 + P1M` =
`2026-02-28`; `2028-01-31 + P1M` = `2028-02-29`), `xsd:time` wraps within its
day, and `date ± dayTimeDuration` keeps only the date part.

Combinations XPath does not define stay type errors: mixed temporal kinds, the
two duration families together, `time ± yearMonthDuration`, and every operator
but `+`/`-`.

That completes SEP-0002's operator table. `ADJUST()` is deliberately **not**
supported — its purpose is to set a value's timezone offset, which is exactly
what our UTC normalization discards, so a result would lose its defining
property on write and read back as `Z`. Documented as a decision rather than a
gap.

## Two bugs found along the way

**A duration rendered as `xsd:string`.** Durations have no `ComparableValue`
variant — they ride as a `TypedLiteral` with no datatype constraint, and that
arm of `to_binding` stamped *every* such value `xsd:string`. Pre-existing: any
`BIND` round-tripping a stored `xsd:dayTimeDuration` or `xsd:gYear` came back
re-typed as a string. Confirmed, not inferred.

**`TZ()`/`TIMEZONE()` changed answer after a reindex.** This one needs review
attention — see below.

## The behaviour change worth arguing about

`TZ` now returns `"Z"` and `TIMEZONE` `"PT0S"` for every temporal value, instead
of the source offset.

The offset is not persisted: `ObjKey::encode_datetime` stores epoch micros and
nothing else, and the decoder re-renders with a `Z`. So it survived only while a
value sat in novelty. Reading it back meant the same query over unchanged data
answered `"-08:00"` and then `"Z"` once a background reindex ran — a different
result with no write behind it, which a caller can neither predict nor control.

This deviates from SPARQL 1.1 §17.4.5.8-9, so **`functions/tz-01` and
`functions/timezone-01` are registered as not-supported**, with the trade-off
written up in the register. Per rule 4 of "Managing the Skip List" that entry
needs two reviewers, which is part of why you're both here.

Worth knowing how it hid: those two tests only ever passed because the
conformance harness loads into a fresh in-memory ledger and never reindexes. Run
against the *exact* W3C fixture on an indexed ledger, they already failed 3 of 4
rows apiece. Registering them records a choice that was, in practice, already
made.

Preserving the offset is the only way to satisfy both the spec and
lane-determinism, and it does not fit: an `ObjKey` is a `u64` already spending
59 bits on microsecond-range instants, leaving no room for the 11 bits an offset
needs. An arena handle would destroy the inline key ordering that dateTime range
pushdown depends on; a sidecar consulted only by these two functions is the only
shape that works, and is not warranted for a value the caller can render
themselves. **Ordering, comparison and range queries are untouched** — they use
the normalized instant, which is what is stored.

## Verification

- Every regression test confirmed **non-vacuous** by reverting its fix
  independently and watching it fail — separately for the arithmetic arms, the
  datatype-render fix, the `± duration` arms, and the lane-agreement property.
- Expected values taken from **XPath spec examples** where they exist (`P337D`,
  `PT7H12M`, `PT0S`, the end-of-month clamps), not from what the code produced.
- **Both query surfaces.** SPARQL and JSON-LD lower to the same IR, so each case
  runs both ways with the two asserted equal.
- **Both storage lanes** for the timezone tests, asserted equal as a property in
  its own right — a single-lane test would have passed against the original bug.
- Full workspace green (85 suites) and W3C conformance 36/36. The one failure,
  `fluree-db-docs`'s embed check over two unpublished reasoning design docs, is
  pre-existing — verified against a stashed clean tree.

## Reviewer notes

Three existing tests pinned superseded behaviour and are **updated in place**,
each with a comment saying what it used to assert and why it changed:
`sparql_timezone_positive_offset` (renamed),
`update_where_bind_datetime_functions`, and the register additions. Changed
assertions deserve a second look.

## Follow-ups

No issues filed yet; happy to open them if you agree they're worth tracking:

- The conformance harness never reindexes, so no W3C test covers the
  binary-index read path. That gap is why the `TZ` bug sat behind a green badge,
  and there's no reason to assume it's the only thing hiding there.
- ~~Temporal values render differently across lanes for the *value* itself, not
  just these accessors.~~ **Fixed in this PR** — see the update below. It was
  the same root cause, and it reached further than rendering.

---

## Update after review — offsets are gone entirely (a1dccc2c6)

@aaj3f's two nits turned out to be the visible edge of the representation
itself, so the fix moved down a layer. Both are resolved, plus two cases
neither of us had listed.

**What the nits actually were.** Finding 1's premise did not hold: `time - time`
with a day-crossing offset was *not* lane-stable — `PT0S` in novelty, `PT9H`
once indexed — because the index stores `micros_since_midnight`, the wall clock,
not the UTC-normalized one. And `date - date` was a fourth case nobody had
named. Documenting three deviations would have documented a wrong one.

**Root cause.** `DateTime`/`Date`/`Time` each carried `tz_offset` and `original`
that survived only until the indexer dropped them. Every consumer reading either
was lane-split, not just arithmetic and `TZ`:

| expression | novelty | indexed |
|---|---|---|
| `"2026-01-01+05:00" = "2026-01-01"` | `false` | `true` |
| `HOURS("2010-12-21T15:38:02-08:00")` | `15` | `23` |
| `STR("2010-12-21T15:38:02.5Z")` — *no offset at all* | `…02.5Z` | `…02.500000Z` |
| `"17:00:00-06:00" - "08:00:00+09:00"` | `PT0S` | `PT9H` |
| `"2026-01-01+05:00" - "2026-01-01"` | `-PT5H` | `PT0S` |

Equality flipping is the sharp one — it feeds `FILTER`, joins and `DISTINCT`.

**The fix: Fluree does not support timezone offsets.** Rather than teach each
consumer to ignore a field that only sometimes exists, the field is gone.
`parse()` is the one choke point every construction path goes through, so it
canonicalizes there — `dateTime` applies the offset and becomes a UTC instant;
`date`, `time` and the gYear family validate the designator and **discard** it
rather than applying it, since applying it would move a calendar day and the
index has only ever held the day / the wall clock. No reindex needed; existing
keys already *are* the canonical form.

`TZ`/`TIMEZONE` answering constants is now a consequence of the representation
rather than a special case, which is what makes it defensible.

**Register grew to six.** `functions/hours-01`, `open-world/date-2`, `date-3`
and `date-4` join `tz-01`/`timezone-01` — all four passed only because the
harness never reindexes. Suite stays 36/36. Rule 4's second reviewer now covers
all six.

**Verification.** Workspace clippy + tests clean; W3C 36/36; `testsuite-sparql`
fmt/clippy run separately. `it_temporal_lane_stability` pins 18 cases on both
lanes — rendering, equality, `HOURS`, arithmetic — and against the pre-change
code **13 of the 18 show the lanes disagreeing**.

**Also updated in place** (beyond the three in Reviewer notes): two assertions
this PR had introduced in 11436a543 were novelty-only answers (`PT5H`, `PT0S`)
and are corrected; `datatype_coercion`'s two lexical round-trip assertions; and
~40 `temporal.rs` unit assertions.

**Net −338 lines.** `to_instant`, `to_utc_time`, `timezone()`, `tz_offset_serde`
and `format_tz_suffix` are all deleted; `cmp`/`Hash` are derived.

**Still open, and unchanged by this:** `xsd:date`/`xsd:time` *ordering* went
through the same offset-folding path — that is pre-existing on `main`, not
introduced here, and now moot for stored values since no offset survives
parsing.

Merged `origin/main` (98 commits) and re-ran every gate on the merged tree.


fluree-db-query/src/eval/value.rs
```diff
@@ -296,10 +297,166 @@ impl ArithmeticOp {
...
+            // must — adding two dateTimes is meaningless.
+            (ComparableValue::DateTime(a), ComparableValue::DateTime(b))
+                if self == ArithmeticOp::Sub =>
+            {
+                day_time_duration_between(a.epoch_micros(), b.epoch_micros())
+            }
+            (ComparableValue::Date(a), ComparableValue::Date(b)) if self == ArithmeticOp::Sub => {
+                day_time_duration_between(a.epoch_micros(), b.epoch_micros())
+            }
+            (ComparableValue::Time(a), ComparableValue::Time(b)) if self == ArithmeticOp::Sub => {
```
aaj3f, line 324: Optional, fold-in — `time - time` with a day-crossing offset diverges from XPath and the cited engines.

The difference of two `xsd:time`s is computed here as (UTC time-of-day A − UTC time-of-day B), each reduced mod-24h by `Time::utc_micros_since_midnight` (`temporal.rs`), so the result is always in (−PT24H, PT24H). XPath's `op:subtract-times` instead anchors both operands to a common reference date and subtracts the full instants, which can land on different UTC dates and yield a magnitude ≥ 24h.

Concretely, XPath's own F&O worked example `op:subtract-times("17:00:00-06:00","08:00:00+09:00")` = `P1D`; this code returns `PT0S` (both normalize to a 23:00 time-of-day). The new tests only cover same-day offsets, so the divergence is silent, and it makes the compatibility.md "matches Stardog/GraphDB/RDFox/Jena/Comunica" claim untrue for this input.

It's *lane-stable* (the offset isn't persisted, so both lanes agree on the mod-24h value), so the consistent move is the one you already made for TZ/TIMEZONE: keep the mod-24h answer, add a test pinning the day-crossing case, and document it as a deviation. Anchoring to a common instant to match XPath would only be correct on novelty — once indexed the offset is gone — so it would reintroduce exactly the pre/post-reindex divergence the rest of this PR removes.

fluree-db-core/src/temporal.rs
```diff
@@ -353,6 +410,33 @@ impl Date {
...
+    pub fn epoch_micros(&self) -> i64 {
+        self.to_instant().timestamp_micros()
+    }
+
+    /// Shift by an `xsd:dayTimeDuration`, in microseconds (XPath
+    /// `op:add-dayTimeDuration-to-date`; negative subtracts).
+    ///
+    /// Per XPath the date is taken to midnight, shifted, and the *date part* of
+    /// the result kept — so a sub-day duration can leave the date unchanged.
+    pub fn checked_add_day_time_micros(&self, micros: i64) -> Option<Self> {
```
aaj3f, line 272: Optional, fold-in — `date ± dayTimeDuration` on an offset-carrying date is lane-unstable.

`Date::checked_add_day_time_micros` folds the offset into a UTC instant via `to_instant()`, applies the shift there, and takes `%Y-%m-%d` from the UTC result. On novelty the offset survives, so `2026-01-01-05:00 + PT20H` = `2026-01-02`; once indexed the offset is gone (`2026-01-01` naive), so the same expression = `2026-01-01`. Same query, unchanged data, different answer across a background reindex — the very class this PR fixes for the accessors, now on a write-path the PR itself adds. (XPath keeps the date's own timezone and answers `2026-01-01`, so the indexed lane happens to be the conformant one.)

This is an instance of your own disclosed follow-up ("temporal values render differently across lanes for the *value* itself… a separate decision"), so it isn't undisclosed — a one-line note in the sparql.md arithmetic section that the new shift operators inherit that cross-lane behavior for offset-carrying `date`/`dateTime` operands, plus an optional pinning test in the `it_timezone_accessors` style, closes the loop.

docs/reference/compatibility.md
```diff
@@ -87,6 +87,10 @@ Supported SPARQL features:
...
...e offset — the binary index stores an instant and nothing else — so `TZ` returns `"Z"` and `TIMEZONE` returns `"PT0S"` for every temporal value, rather than the offset the literal was written with. The spec expects the source offset, and the two W3C tests covering it (`functions/tz-01`, `functions/timezone-01`) are registered as not-supported. The reason is determinism across the storage lanes: the offset survives only while a value sits in novelty, so reading it back made these functions answer `"-08:00"` before a background reindex and `"Z"` after — the same query over unchanged data returning a different result, with no write behind it and no way for a caller to predict which they would get. Preserving the offset instead would need either an arena handle (destroying the inline key ordering that dateTime range pushdown depends on) or a sidecar consulted only by these two functions; neither is warranted for a value the caller can supply by rendering in whichever zone they want. **Ordering, comparison and range queries are unaffected** — those use the normalized instant, which is exactly what is stored. Applications that need a wall-clock offset should store it as its own property.
```
aaj3f: Micro, rides with the finding-1 doc sentence: this paragraph (and the sparql.md "Timezone-normalized" bullet) asserts equivalence with Stardog/GraphDB/RDFox/Jena/Comunica without qualification. Given the day-crossing `time - time` case above, scope that claim to same-timezone-origin operands, or add the day-crossing caveat. Purely wording.

aaj3f (approved):
This is great and makes sense @bplatz

Two nits, both at the offset-carrying edge of the new operators, both fold-in and neither is a regression (`main` answered unbound for all of this):

1. `time - time` (and `date ± dayTimeDuration`) with a timezone offset that crosses UTC midnight diverges from XPath and from the engines the docs name. It's lane-stable, so the honest fix is to document it as a deviation the way you did TZ/TIMEZONE and pin it with a test, not to chase an XPath answer the indexed lane can't reproduce.
2. `date ± dayTimeDuration` on an offset date is lane-unstable. An instance of your own disclosed "values render differently across lanes" follow-up, now reachable through arithmetic. A one-line note + a pinning test closes it.

**Adherence checklist**

1. Patterns / abstractions — ✔ extends `ArithmeticOp::apply`, the `Duration` enum is the right small addition, math lives in `fluree-db-core::temporal` beside the types; honors the shared-IR model (both surfaces tested).
2. Performance — ✔ neutral→positive: TZ/TIMEZONE now skip the value decode; new arms guarded off the non-temporal path; no hot scan/encode change.
3. Testing — ✔ unit + both-surface integration + cross-lane reindex, all confirmed non-vacuous by mutation. ⚠️ the offset day-crossing edge (findings 1/2) is untested.
4. Conventions — ✔ self-describing multi-line commits; fmt/clippy clean (default features; `--all-features` via green CI); docs updated across four files; three superseded tests updated in place with change comments.

bplatz:
Thanks — both nits are fixed, but chasing them moved the fix down a layer, so flagging that before you re-look.

**Finding 1's premise didn't hold.** `time - time` wasn't lane-stable: `PT0S` in novelty, `PT9H` once indexed. The index stores `micros_since_midnight` — the wall clock, not the UTC-normalized one. So "keep the mod-24h answer and document it" would have documented a deviation that was itself wrong. **`date - date` was a fourth case** neither of us listed, same split.

Both were symptoms of the representation: `tz_offset` and `original` lived on the structs but only survived until the indexer dropped them, so *every* consumer reading either was lane-split — including `=` (`false` → `true` after a reindex, which feeds FILTER/joins/DISTINCT), `HOURS()` (15 → 23), and `STR()` even on a value with no offset at all (`.5Z` → `.500000Z`).

**So offsets are gone entirely.** `parse()` canonicalizes: `dateTime` applies the offset and becomes a UTC instant; `date`/`time`/gYear validate the designator and discard it (applying it would move a calendar day, and the index only ever held the day / wall clock). No reindex — existing keys already are the canonical form. `TZ`/`TIMEZONE` returning constants is now a consequence of the representation rather than a special case.

Register grew to six: `hours-01` and `open-world/date-2/3/4` joined `tz-01`/`timezone-01`, all of which passed only because the harness never reindexes. Suite still 36/36 — so rule 4's second reviewer now covers six entries rather than two.

`it_temporal_lane_stability` pins 18 cases on both lanes; against the pre-change code 13 of them show the lanes disagreeing. Net −338 lines. Details in the PR body update; `origin/main` merged and all gates re-run on the merged tree.

Your finding 3 (the docs claim) is in — I dropped the engine list rather than scoping it, since we'd rather cite SEP-0002 than the field.

Worth a re-review given the size change.

PR #1748, opened 2026-08-30
feat(graphql): GraphQL endpoint over a schema derived from the ledger

bplatz:
Adds a GraphQL endpoint to every ledger, with the schema derived from what the ledger already contains. No `.graphqls` upload, no resolvers, no build step — point a client at an existing ledger and it introspects, filters, paginates and queries.

## Three tiers

The schema sharpens as the ledger says more about itself. Which tier applies is decided by what the ledger contains, not by config:

| Tier | Trigger | What it adds |
|---|---|---|
| **Inferred** | nothing | Every class a type, every observed property a nullable list field |
| **Shaped** | a `sh:NodeShape` with `sh:targetClass` | Cardinality, datatypes, enums, reverse fields, docs, closed types |
| **Curated** | a `graphql:Schema` instance | What is published, its names, interfaces, and — opt-in — mutations |

The customization layer is **SHACL in the ledger**, plus the `http://datashapes.org/graphql#` vocabulary for the parts SHACL does not cover. So one artifact drives both validation and the API contract: it is versioned with the data and time-travels with `t`.

```graphql
{
  persons(where: { name: { RE: "^A" }, knows: { name: { EQ: "Bob" } } }, limit: 20) {
    id
    name
    knows(orderBy: { name: ASC }, limit: 5) { id name }
  }
}
```

## Surfaces

- `POST/GET /v1/fluree/graphql/<ledger>` and `GET /v1/fluree/graphql-schema/<ledger>`
- `fluree graphql [--schema | --bootstrap | --explain]`
- `Fluree::graphql` (read) and `Fluree::graphql_transact` (write)

Behind a `graphql` feature — default on in server and CLI, in `full` for the api — which implies `shacl`, since tier 2 reads shapes.

## How it works

A request is parsed, its selection tree extracted from the document, and each root field lowered to a **JSON-LD query document**. That document goes through the engine's ordinary read path, so its own parser validates and encodes it and policy, SHACL, time travel and formatting all apply unchanged. The result is reshaped back into the GraphQL response — aliases, fragments, `__typename` and list contracts restored.

`extensions.explain` returns that lowered query, so the JSON-LD a field compiled to is inspectable:

```
?explain=true   |   extensions: {"explain": true}   |   fluree graphql --explain
```

Mutations lower to ordinary transactions through the same write path any other client uses, so a rejected write comes back as a GraphQL error having written nothing.

The new `fluree-db-graphql` crate holds the schema model, naming, datatype mapping, the three builders, lowering and reshaping. It takes plain IRIs; SID resolution, policy pruning and stats assembly stay in `fluree-db-api`, which keeps the crate testable without ledger fixtures.

## What the schema claims

Tier 1 is deliberately weak, because statistics cannot support more. Every field is a nullable list — statistics can say a property has never been seen twice on one subject, not that it never will be — and `orderBy` accepts only `id`, since ordering solutions by a multi-valued key would repeat subjects. No interfaces and no reverse fields: both need someone to say which classes are abstract and what the other direction is called. `xsd:integer` maps to a custom `Long` rather than 32-bit `Int`, and `xsd:decimal` to `Decimal` rather than rounding into `Float`.

Shapes supply what statistics cannot: `sh:maxCount 1` gives a single value, `sh:minCount ≥ 1` gives `!`, `sh:in` gives an `enum`, `sh:inversePath` gives a reverse field, and `sh:closed true` drops observed-but-undeclared properties.

A `graphql:Schema` makes the endpoint a deliberate contract: only listed shapes are published, so it does not grow a type the moment someone writes an instance. `publicShape` gets root fields, `protectedShape` is reachable only through a reference, and `privateShape` degrades references to `Node` — the IRI stays visible without naming a type the caller cannot query.

Policy applies by pruning: a class or property an identity cannot read is absent from introspection rather than present-but-empty.

## Mutations

Off unless a `graphql:Schema` sets `f:graphqlEnableMutations`, and only in tier 3 — a schema derived from whatever a ledger happens to contain should never become a write surface by accident. Each published type then gets `create_`, `update_` and `delete_`.

Writing needs a `LedgerState`, which a read view does not carry, so mutation fields are registered only on the write path: the SDL a read endpoint serves matches what it can answer. `f:graphqlIriBase` is required to mint an IRI, with no default, since a wrong guess writes identifiers that cannot be un-minted.

`update_` lowers to `where`/`delete`/`insert` rather than an upsert: a property set to `null` must be retracted with nothing put back, which an upsert cannot express, and this keeps every property and subject in one atomic transaction. Both `update_` and `delete_` anchor on `@type`, so `delete_Person` on a Company's IRI is a no-op rather than a wipe.

## Also included

Two changes outside the GraphQL layer, both independently useful:

- **JSON-LD queries gain per-level ordering and paging inside a hydration.** A nested selection's value may be an object of `select`/`orderBy`/`limit`/`offset` instead of a bare array, bounding how many values *each subject* shows — a different question from how many rows the query returns, and one the WHERE clause cannot express. Documented in `docs/query/jsonld-query.md`.
- **SHACL compiles `sh:description`, `sh:order` and `sh:defaultValue`** as annotation properties. They constrain nothing but are what a schema generator reads. `sh:defaultValue` is carried and never materialized: a default is a statement about presentation, not about what the graph holds, and asserting the triple would make `sh:minCount 1` self-satisfying.

## Performance

Both the derived model and the registered executable schema are cached per ledger version, keyed so that any write, shape edit or context change invalidates them. Steady state, a small query costs ~84 µs against ~39 µs for the equivalent hand-written JSON-LD query; the difference is document parsing, lowering, execution and reshaping. Tracked by `fluree-db-api/benches/graphql_schema.rs`.

## Testing

71 tests in `fluree-db-graphql`, 40 end-to-end in `fluree-db-api`, 12 over HTTP in `fluree-db-server`, and 6 for the new JSON-LD nested-selection syntax. Full workspace suite green at 11,724.

## Docs

`docs/query/graphql.md` (reference), `docs/cli/graphql.md` (CLI, with the full SHACL and curation tables), a GraphQL vocabulary section in `docs/reference/vocabulary.md`, and the endpoints in `docs/api/endpoints.md`. `GRAPHQL.md` at the root records the design and the deliberate gaps: nested `where`, nested arguments on a reverse field, several `graphql:Schema` instances per ledger, and `::n` intra-mutation references.

Also in this PR: the `values` arity in the JSON-LD query docs was incorrect and is corrected; every broken internal link and anchor in `docs/` is repaired (0 of 280 files now broken, was 20); and the ten workspace crates missing from the crate map are filled in.


fluree-db-server/src/routes/graphql.rs
```diff
@@ -0,0 +1,275 @@
...
+        authorize_read(&state, &ledger, &bearer, &credential)?;
+        let request = parse_request(&params, &credential)?;
+
+        // A GraphQL error is part of the response body, not a transport failure:
+        // the spec has clients read `errors`, and returning 4xx for an unknown
+        // field would break every standard client.
+        let response = if fluree_db_api::graphql::is_mutation(&request) {
+            execute_mutation(&state, &ledger, &headers, &bearer, &credential, &request).await?
+        } else {
+            let view = policy_view(&state, &ledger, &headers, &bearer, &credential).await?;
```
aaj3f: **Blocking — GraphQL reads bypass `query_timeout_ms` and client-disconnect cancellation.**

Every other read surface bounds queries: `routes/query.rs`, `stream_query.rs`, `validate.rs` all run inside `query_control::run_query_task(state.config.query_timeout_ms, …)` / `current_query_execution_options(…)`, which installs a `QueryCancellation` handle plus a timeout task (default `DEFAULT_QUERY_TIMEOUT_MS = 15*60*1000`, `fluree-db-api/src/server_defaults.rs:17`). This route calls `state.fluree.graphql(&view, &request)` directly — not inside `run_query_task` — and `LedgerExecutor::resolve` calls the 2-arg `self.fluree.query(&db, &q)` (`fluree-db-api/src/graphql.rs:884`), which is `query_with_options(…, QueryExecutionOptions::default())` (`view/query.rs:286-289`): no cancellation, no timeout. The timeout lives in a `tokio::task_local!` in `fluree-db-server` that `fluree-db-api` cannot read, so nothing installs it here.

Fuel is not a backstop either: the lowered query carries no `opts.maxFuel`, `tracker_for_limits` (`query/helpers.rs:610`) returns `Tracker::disabled()` when none is present, and the GraphQL surface exposes no way to set one — so the timeout is the only ceiling that would apply, and this is the path that drops it. An expensive GraphQL query runs to completion regardless of the configured (or default 15-min) timeout, where the same query over `/v1/fluree/query` is cancelled; a client that disconnects mid-query does not cancel it. Amplified by aliasing (finding #2): async-graphql resolves root fields concurrently, so one small document with N aliased expensive root fields launches N concurrent, individually-unbounded queries.

Thread `QueryExecutionOptions` into `LedgerExecutor` (a field), have `resolve`/`read_back`/`mutate`'s read-backs call `query_with_options(&db, &q, options.clone())`, and wrap the route body in `crate::query_control::run_query_task(state.config.query_timeout_ms, …)` so a shared cancellation handle covers all of a request's root-field queries. This is the same wiring `routes/query.rs:573` already uses.

fluree-db-graphql/src/runtime.rs
```diff
@@ -0,0 +1,855 @@
...
+            request.field.name
+        )))
+    }
+}
+
+/// Build an executable schema from a model.
+///
+/// `mutations` are the write fields to expose; empty means a read-only schema,
+/// which is every tier below 3 and any tier-3 schema that did not opt in.
+pub fn build_schema(model: &SchemaModel, mutations: &[MutationField]) -> Result<Schema> {
```
aaj3f: **Should-fix (fold in now) — no depth / complexity / aliasing limits on the schema.**

`Schema::build` here is finished with no `.limit_depth(_)` / `.limit_complexity(_)`; async-graphql defaults both to unlimited. Deep nesting recurses `parse_query`, `selection::walk` (`selection.rs:123`), `build_level`, and `reshape`, and the inferred schema has cyclic types (`Person.knows: [Person]`) so depth is attacker-chosen. Note the first two run in `run_graphql` (`api/graphql.rs:1152,1157`) *before* `schema.execute()` (`:1169`), so a schema-level `limit_depth` alone wouldn't cover them.

Either vector is cheap to exploit: `{ a: persons(where:…){…} b: persons … }` × thousands of aliases (a few hundred KB, within the body limit) → thousands of concurrent queries (each unbounded per blocking #1); or one deeply-nested document → deep recursion before any budget applies.

Add `.limit_depth(N).limit_complexity(M)` on the builder in `build_schema`, plus a cheap explicit depth guard in `selection::extract`/`walk` (it already threads state for the cyclic-fragment guard, so a depth counter is a small addition). Fixing #1 bounds each query's wall-clock; this bounds the *count* and the pre-execution recursion. Conservative production defaults, ideally configurable alongside `query_timeout_ms`.

fluree-db-graphql/src/mutate/mod.rs
```diff
@@ -0,0 +1,426 @@
...
+/// A fresh local name for a minted subject.
+///
+/// A process-lifetime counter supplies uniqueness; hashing it under a
+/// per-process random seed keeps the minted IRIs from reading as a visible
+/// sequence. Uniqueness is the property relied on — knowing another subject's
+/// IRI grants nothing on its own, since reads are governed by policy.
+///
+/// Deliberately no clock: `SystemTime::now()` *panics* on
+/// `wasm32-unknown-unknown`, and it would compile fine on the way there.
+fn new_id() -> String {
```
aaj3f, line 425: Optional — minted subject IRIs are 64-bit; consider widening to avoid a silent create-merge.

`new_id` mints a 16-hex (64-bit) local id from a per-process `AtomicU64` counter hashed under a per-process `RandomState`. Across a clustered/replicated write deployment each process seeds independently, so uniqueness is birthday-bound (~2³² mints), and a collision makes `create_<T>` (a `Verb::Insert`, `:178`) silently merge the new subject's facts onto an existing one. Low probability, silent outcome.

A 128-bit id (UUIDv4 or ULID) drops the collision risk to negligible and keeps the no-clock property the last commit was after.

fluree-db-api/src/graphql.rs
```diff
@@ -0,0 +1,1305 @@
...
+        .policy_enforcer()
+        .map(std::convert::AsRef::as_ref)
+        .filter(|p| !p.is_root());
+
+    let mut out = Vec::new();
+    for entry in stats.classes.iter().flatten() {
+        if entry.count == 0 {
+            continue;
+        }
+        if policy.is_some_and(|p| crate::cypher_procedures::class_denied(p, &entry.class_sid)) {
```
aaj3f, line 356: Optional — add a test for policy-pruned introspection and the cache bypass.

The filters that keep a denied class/predicate out of the derived schema (`:300-314`, `:594-643`) and the `cache_key → None` that stops a policy view sharing a cached schema (`:229-231`) are a security boundary with no direct test (no `it_graphql*` references `policy`/`Deny`/`default_allow`).

An end-to-end test — a policy-restricted view's SDL/introspection omits the denied type, and re-deriving under a different identity does not serve a cached (root) schema — should go red if the pruning is removed. Worth folding in now while the mechanism is fresh.

fluree-db-api/src/graphql.rs
```diff
@@ -0,0 +1,1305 @@
...
+    }
+}
+
+/// Whether a request's operation writes.
+///
+/// Decided from the document, not the HTTP method: GraphQL sends everything
+/// over `POST`, so the method says nothing about intent. An unparseable
+/// document reads as a query, so the parse error surfaces from the read path
+/// where it is the only thing wrong.
+pub fn is_mutation(request: &GraphQlRequest) -> bool {
```
aaj3f: Optional — parse the request document once. `is_mutation` (`:91-95`, called from the route) parses the document, then `run_graphql` (`:1152`) parses it again and `selection::extract` walks it. Cheap vs. the query, but the parsed `ExecutableDocument` could be produced once and threaded, deciding read-vs-write from the same parse.

fluree-db-api/src/format/hydration.rs
```diff
@@ -244,6 +248,139 @@ fn reencode_level_for_view(
...
+                    None => (Some(a), Some(b)),
+                };
+                let ordering = compare_sort_values(lhs, rhs);
+                if ordering != std::cmp::Ordering::Equal {
+                    return if *descending {
+                        ordering.reverse()
+                    } else {
+                        ordering
+                    };
+                }
```
aaj3f, line 293: Optional, cosmetic — nested `orderBy` on large integers loses precision. `compare_sort_values` compares via `a.as_f64().partial_cmp(&b.as_f64())`, which misorders `Long`/`Decimal` past 2⁵³. It only orders one subject's already-materialized values, so stakes are low; if easy, compare `serde_json::Number` as i64/decimal before falling back to f64.

aaj3f (approved):
@bplatz this is really neat! And if you'd asked me before this PR, I would not have realized how "easily" (in terms of runtime, not in terms of the work you did) we could derive and provide something like this.

The one thing you may want to consider landing before (or immediately with) merge: it seems GraphQL reads bypass the server's query timeout and client-disconnect cancellation that bound every other read surface. `Fluree::query` here runs with `QueryExecutionOptions::default()` outside any `run_query_task`, so a GraphQL query is unbounded by default where the identical JSON-LD query is cut, for example, at 15 minutes, and aliasing lets one request fan out into many concurrent unbounded queries. Threading `QueryExecutionOptions` through `LedgerExecutor` and wrapping the route in `run_query_task` would close this; adding `limit_depth`/`limit_complexity` (finding `#2`) is the companion hardening a public GraphQL endpoint wants. The minted-IRI width, a policy-pruning test, and the double-parse are all optional thought possibly worth considering now rather than later.

**Adherence checklist**

1. Patterns / abstractions — ✔ Exemplary. Lowers to the shared JSON-LD read path and IR; no parallel evaluator (nested fields are pass-through resolvers, execution is `query_with_options` through the ordinary engine); isolated crate. The 5 shared `fluree-db-query` edits EXTEND the IR (a `modifiers` field on the existing `ForwardItem::Property` + new `NestedModifiers` types + a JSON-LD-only parser object-form), not a GraphQL branch in the shared parse path; SHACL/vocab changes additive. JSON-LD and SPARQL unaffected — verified green at default features (fluree-db-query 1536/1536, grp_query 428/428, grp_query_sparql 369/369).
2. Performance — ⚠️ New work is isolated and the shared hydration/parser changes are perf-neutral (common path untouched, sort-key resolution hoisted, cache key updated). But the endpoint ships **without a query timeout or cancellation** (blocking `#1`) and without depth/complexity limits (`#2`) — a default-active resource-bounding gap on the product's headline axis.
3. Testing — ✔/⚠️ 71 crate + 40 e2e + 12 HTTP + 6 nested-select tests, all wired and green; strong names. Verified locally the shared IR change does not regress the siblings. Gap: the policy-pruning/cache-bypass security path is untested (`#4`).
4. Conventions — ✔ New crate opts into `[lints] workspace=true`, inherits BUSL/version, own `thiserror` error, `async-graphql` pinned at workspace, bench registered in `regression-budget.json`. fmt/clippy clean. Self-describing multi-line commits. Base is `main` (real CI).

bplatz:
Thanks — this was a genuinely useful review, and the blocking finding was correct in every particular, including the detail that fuel is no backstop.

All five findings are addressed. I'd originally planned to defer the three optional ones and you'd have been right to push back on that: "don't churn an approved branch" is an argument about shipped code, and none of this has shipped. The minted-IRI one especially — that's a permanent on-disk identifier, so it's the one item here that genuinely cannot be fixed later.

Branch is rebased on `main` first (two conflicts, both from the wasm-groundwork reshuffle: `fluree-db-spatial` and `pub mod import` moved under `cfg(not(wasm32))`).

### #1 — timeout and cancellation · `b8ae924c4`

`LedgerExecutor` carries `QueryExecutionOptions` and every lowered query goes through `query_with_options`; the route body is wrapped in `run_query_task`, the same wiring `routes/query.rs:573` uses. Root fields resolve concurrently so they share one handle — a timeout or a disconnect cancels the whole fan-out rather than whichever field checks next. New entry points are `graphql_with_options` / `graphql_transact_with_options`; the existing pair delegates.

### #2 — depth / complexity · `b8ae924c4`, `c32bba2de`

`graphql_max_depth` (15) and `graphql_max_complexity` (1000), settable by flag, env var and config file, `0` disabling either the way `query_timeout_ms = 0` does. Your point about the ordering was the load-bearing one: depth is checked in `selection::walk` because extraction runs before `schema.execute()`, and it counts levels exactly as async-graphql's `DepthCalculate` does so the two agree on which documents are refusable. Complexity is left to the schema. Both are in `RegisteredKey` — they're baked into the registered schema, which is cached, so otherwise the first request through would fix the ceiling for everyone after.

**Chasing this turned up something worse than a slow parse.** `async_graphql_parser` caps recursion at 64 — but only while *building the AST*, after pest has already descended the grammar unbounded. Measured: 1,000 levels is refused cleanly, 10,000 levels overflows the stack and aborts the process. A request inside the body limit kills the server, and `SIGABRT` isn't catchable, so a limit that runs after the parse can't help. There's now an O(n) pre-parse scan bounding brace nesting at that same 64 (`limits::guard_nesting`).

### #4 — policy pruning · `b8ae924c4`

Four tests in `it_graphql_policy.rs`: a denied class and a denied property are absent from the SDL rather than present-and-empty, a denied class is unqueryable, and a policy view isn't served a cached root schema (checked in both directions).

### Minted IRIs · `58c919c26`

Now 128 bits, via two independent per-process seeds — collision bound goes from ~2³² mints to ~2⁶⁴, matching UUIDv4. Kept the no-clock property you'd noted from the previous commit: `SystemTime::now()` panics on wasm32, which rules out a ULID or UUIDv7, so the seeds still come from `RandomState`. No new dependency.

### Double-parse · `58c919c26`

It was three parses, not two — `is_mutation`, `run_graphql`, and async-graphql's own. `PreparedRequest` does the parse once, answers `writes()`, and hands the document to `selection::extract` and then to the executor via `Request::set_parsed_query`; `execute` uses it as-is and validation, limits included, still runs on it. `is_mutation` is gone. The bigger win is that this collapses the stack guard to a single call site — it previously had to be repeated at every parse, which is exactly the shape someone forgets.

### f64 sort · `58c919c26`

`compare_sort_values` compares as i64/u64 before falling back to f64.

Worth recording how this one is tested: the obvious end-to-end test **passed with the fix reverted**, because the hydration already delivers those values in ascending order so the tie never bites. The real pin is three unit tests on the comparator itself; the integration test stayed but is labelled for what it actually covers. I also wrote and then removed explicit mixed-sign branches — they behave identically to the f64 fallback, since rounding can collapse neighbours but never flip a sign.

### One thing the merge broke

`main` gained `autotests = false` and the `grp_*` harnesses after this branch forked, so `tests/graphql_http_integration.rs` stopped being auto-discovered — all 12 HTTP tests would have gone dark with CI still green. The crate's own `harness_coverage` guard caught it, which is a nice validation of that guard. Registered in `grp_http` behind `#[cfg(feature = "graphql")]` (`b8b56e470`), plus 3 new HTTP tests for the configured limits.

### Verification

Every new test was checked against a reverted fix — I sabotaged each mechanism in turn (options threading, the walk guard, the schema limits, the cache key, both policy filters, the cache decline, the route's config wiring, the stack guard) and confirmed the specific tests fail. The stack guard's proof is the process abort itself.

Two honest caveats:

- **The single-parse change is not measurable at the bench level.** `graphql_query_overhead` still reads ~83.6 µs. A parse is ~5.4 µs so I expected a few percent, but I don't have a clean before-number on the same machine — treat the perf framing as unverified. It stands on the single-guard-site argument regardless.
- I did not run the full `--workspace --all-features` suite locally (the `target/` dir filled the disk mid-session); CI covers it. Green through `c32bba2de`, with the last commit queued.

Also dropped `GRAPHQL.md` — it was the working design record for this branch and the design is settled. The parts a user needs (the deliberate gaps, and now the limits) moved into `docs/query/graphql.md`, and the two code comments that pointed at it now point there.

PR #1749, opened 2026-08-31
SQL graph sources: R2RML over Trino-protocol endpoints, plus the fluree-sql-bridge sidecar

bplatz:
Adds **SQL graph sources**: R2RML mappings over tables reached through any HTTP endpoint speaking the **Trino client protocol** — Trino / Starburst / PrestoDB directly, or the new `fluree-sql-bridge` sidecar in front of a single Postgres, MySQL or SQLite database. No JDBC and no database drivers in the `fluree` binary; every scan is a stateless `POST /v1/statement` + paged GETs, so this works from a long-running server and from a Lambda alike.

> **#1748 has merged**, so this PR is now based on `main` and CI runs on it. The stacking note that used to live here is resolved: `main` is merged in, the Cargo feature-list overlap is settled, and the `graph_source/cache.rs` cache-type alias main introduced is adopted.

## Design

The engine's R2RML boundary already asks providers for one table at a time (`scan_table(projection, filters, topk)`), so a SQL source renders exactly one `SELECT … WHERE …` per triples map and everything else — joins, OPTIONAL, UNION, aggregation — stays in-engine. No Ontop-style whole-query rewriting.

- **Typed pushdown**: every filter literal is rendered against the column type from a cached `SELECT * … LIMIT 0` probe; an unsafe pairing (string vs `bigint`, tz-mismatched timestamp, NaN) is declined rather than guessed, since a mistyped comparison fails the whole statement in Trino and the in-engine FILTER stays authoritative regardless. Bound subjects push through template reversal; `COUNT` is answered by an exact `SELECT COUNT(*) … IS NOT NULL` (exact where Iceberg can only use manifest stats).
- **Top-k is deliberately not pushed**: a NULL in a key/required column would consume `LIMIT` slots for rows the mapping drops, breaking the superset contract.
- **`rr:sqlQuery`** is supported (SQL sources only): it compiles to a deterministic alias that stands in wherever a table name is expected, and scans as a derived table. Iceberg-backed registration refuses it up front.
- **No snapshots**: SQL sources read live tables. Time travel, materialization and `iceberg track` are refused with an error naming the source (full-rebuild materialization is a follow-up). Build watermarks record `sql://endpoint/table@time`.
- **SSRF posture**: outbound requests follow no redirects and refuse the link-local/metadata range at both the request boundary and DNS resolution — but loopback/private hosts are allowed, because a sidecar next to the database is the primary deployment shape (mirrors the S3 `endpoint` policy, not the catalog policy). The route is admin-token protected like `/iceberg/map`.

## Surface

- `fluree-db-sql` crate (feature `sql`, on by default in server/CLI; stacks on `iceberg` since dispatch lives in `FlureeR2rmlProvider` and it reuses the shared `ConfigValue`/auth/secret-ref machinery)
- `GraphSourceType::Sql` / `f:SqlMapping`; `Fluree::create_sql_graph_source`; `POST /sql/map`; `fluree sql map|list|info|drop` (list/info/drop shared with `fluree iceberg`); hard drop sweeps the CAS mapping blob for all mapped-source families
- `fluree-sql-bridge/`: a standalone workspace (excluded like `testsuite-sparql`, own lockfile and CI job) so sqlx and its three drivers never enter the main dependency tree
- Docs: `graph-sources/sql.md`, `cli/sql.md`, endpoints, vocabulary, crate map — plus a fix to the long-broken GRAPH-join example in `graph-sources/overview.md` (object-with-`"graph"`-key syntax never parsed; the array form + `from-named` is now shown and test-pinned)

## Verification

- **Real Trino 483** (docker, memory connector): registration probe, scans, pushed `WHERE`, `COUNT(*)`, dates, zoned timestamps — all through the public API. Confirmed on the raw wire that `decimal(10, 2)` (with space) and `… UTC` renderings decode.
- **Postgres 16 through the bridge**: live round-trip incl. `NUMERIC`, `TIMESTAMPTZ` (offset normalized to UTC), `TEXT[]`.
- The live test stays in-tree, env-gated (`FLUREE_SQL_BRIDGE_URL`), and **skips loudly** when unset.
- **MySQL 8 and Postgres 16 as CI service containers** on the `sql-bridge` job, covering the injection regression, the `LIMIT 0` probe's Trino type names and value round-trips. A guard test fails if CI ever stops supplying the URLs, so a skipped backend cannot read as a passing one.
- Fake-endpoint e2e asserts the exact SQL sent (projection, typed predicates, derived table). Bridge has in-process protocol tests over a real SQLite file (paging, zero-row probe, errors, auth, cancel).
- Ledger⇄SQL joins tested in both SPARQL (`FROM NAMED` + `GRAPH`) and JSON-LD (`["graph", …]` + `"from-named"`).
- Perf on existing paths: plain ledger queries untouched; the one regression risk found in self-review (a per-scan nameservice lookup added to Iceberg scans for dispatch) is removed by memoizing the decision per query session.

## Known gaps / candidate follow-up issues (none filed yet)

1. Bridge distribution: build-from-source only (no dist target / container image).
2. Full-rebuild materialization for SQL sources.
3. Mapping generation from a SQL schema (`/iceberg/r2rml/generate` is Iceberg-only).
4. Accepting `{"secret_ref": …}` through `POST /sql/map` and `fluree sql map`; both store credentials as literals in the graph-source record today (documented, matches the Iceberg REST catalog).
5. Pre-existing, noticed while testing: `fromNamed` dataset-local aliases don't resolve to graph sources, and the Iceberg path re-runs its nameservice lookup per scan (same memoization opportunity as fixed here for dispatch).

Happy to file any of these if you agree they're worth tracking.

## Review follow-ups (@aaj3f)

- **Blocking — backslash escaping / MySQL injection**: fixed at both ends. The bridge now sets `NO_BACKSLASH_ESCAPES` on every MySQL session, and the renderer *declines* the pushdown for a MySQL string carrying a backslash rather than escaping for a `sql_mode` it cannot observe — `dialect` names the database behind an endpoint that need not be a bridge we configured. Declining is the same valve type mismatches already use, so it costs I/O and not correctness. Regression tests at both layers; the render-side test was verified to fail with the guard reverted.
- **`[lints] workspace = true`**: added. It was not clean underneath — three findings, fixed. The bridge is a separate workspace and cannot inherit, so it mirrors the table with a pointer to the root.
- **Secrets as literals**: documented, and filed above as a follow-up. Correction to the review thread: the CLI stores literals too, not just the route.
- **`row_count` cost**: documented next to the `COUNT` claim.
- **Found while fixing the above**: Postgres had the same exposure through a settable `standard_conforming_strings`; the bridge now pins it per session as well.


fluree-db-sql/src/dialect.rs
```diff
@@ -0,0 +1,580 @@
...
+                .map(|m| render_literal(m, ty, dialect))
+                .collect();
+            rendered.map(|r| format!("{col} IN ({})", r.join(", ")))
+        }
+        (Literal::Set(_), _) | (_, CmpOp::In) => None,
+        (lit, op) => render_literal(lit, ty, dialect).map(|l| format!("{col} {} {l}", op.sql())),
+    }
+}
+
+fn sql_string(s: &str) -> String {
```
aaj3f: **Blocking — `sql_string` does not escape backslashes → SQL injection on MySQL-backed sources.**

String literals are rendered here by wrapping in `'…'` and doubling embedded `'` only; backslash passes through unescaped. That's safe for Trino, Postgres (default `standard_conforming_strings=on`), and SQLite — none process backslash escapes in string literals. MySQL, under its default `sql_mode`, treats `\` as an escape character inside string literals. The bridge runs the engine-rendered SQL verbatim (`fluree-sql-bridge/src/mysql.rs:155`, `sqlx::query(&sql).fetch(&mut *conn)`, no `NO_BACKSLASH_ESCAPES` — grep of the crate confirms none). So for a `dialect=mysql` source, a value like `x\` renders as `'x\'` — MySQL reads `\'` as an escaped quote, the string never closes, and the following SQL is parsed as code. The tainted value is query-author-controlled: `render_literal` pushes `Literal::Str(s)` (a `FILTER` on a String column) and `Literal::TemplateKey(raw)` (a subject-IRI value reversed through a subject template) straight through `sql_string` (`:319`, `:361`). And `dialect=mysql` isn't exotic — it's *required* for MySQL, because it selects backtick identifier quoting.

Scenario: a user with query access to a MySQL-backed SQL graph source issues `… FILTER(?name = "a\' UNION SELECT secret, … FROM other_table -- ")`. The bridge hands MySQL `SELECT … WHERE ` + "`name`" + ` = 'a\' UNION SELECT secret, … -- '`, reading a table the R2RML mapping never exposed. (Stacked `;` statements are likely blocked by sqlx's single-statement execution, but UNION/subquery/blind extraction is not.) The mapping-as-access-boundary is defeated, up to the bridge DB user's privileges.

Preferred fix (one place, canonical): make MySQL obey the standard-SQL string rule the renderer already assumes. In `fluree-sql-bridge/src/mysql.rs`, set `NO_BACKSLASH_ESCAPES` on every pooled connection:

```rust
let pool = MySqlPoolOptions::new()
    .max_connections(max_connections)
    .after_connect(|conn, _meta| Box::pin(async move {
        conn.execute("SET SESSION sql_mode = CONCAT(@@sql_mode, ',NO_BACKSLASH_ESCAPES')").await?;
        Ok(())
    }))
    .connect(url).await…
```

Then `sql_string`'s quote-doubling is correct for all four dialects. Alternative (engine-side defense-in-depth): make `sql_string`/`render_literal` dialect-aware and double `\`→`\\` for `SqlDialect::Mysql` — but that couples the renderer to MySQL's *default* mode and over-escapes if `NO_BACKSLASH_ESCAPES` is ever set, so the bridge-side fix is the clean one. Either way, add a MySQL-backed test that pushes a string filter containing `\` and `'` and asserts the injected value is treated as data — the current bridge tests are SQLite-only, the one backend this can't hit.

(The fix lives in `fluree-sql-bridge/src/mysql.rs`; commenting here on `dialect.rs:287` because this is the render site that produces the unescaped literal.)

fluree-db-sql/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+# `ConfigValue` / `SecretResolver` / `AuthConfig` / `MappingSource` are shared with
+# the Iceberg graph source. The base crate (no `aws` feature) is HTTP + serde only.
+fluree-db-iceberg = { path = "../fluree-db-iceberg", default-features = false }
+
+async-trait.workspace = true
+tokio = { workspace = true, features = ["sync", "time"] }
+futures.workspace = true
+async-stream = "0.3"
+
+reqwest = { version = "0.12", default-features = false, features = ["rustls-tls", "json"] }
```
aaj3f, line 20: Optional, fold-in — add `[lints] workspace = true`.

This new library crate omits the workspace-lints opt-in that siblings `fluree-db-r2rml` and `fluree-db-tabular` both carry, so it silently forgoes the 23 workspace-denied pedantic clippy lints (`uninlined_format_args`, `semicolon_if_nothing_returned`, …). It's clean on the base lints today, but nothing keeps it that way.

Append:

```toml
[lints]
workspace = true
```

and re-run `cargo clippy -p fluree-db-sql --all-targets -- -D warnings`.

fluree-db-server/src/routes/sql.rs
```diff
@@ -0,0 +1,192 @@
...
+            "presto" => WireProtocol::Presto,
+            other => {
+                return Err(ServerError::bad_request(format!(
+                    "unknown protocol '{other}'. Use trino or presto."
+                )))
+            }
+        };
+    }
+
+    if let (Some(url), Some(secret)) = (&req.oauth2_token_url, &req.oauth2_client_secret) {
```
aaj3f, line 174: Optional (decision-needed, so fine to defer with the reason named) — secrets from the HTTP route are persisted as literal `ConfigValue`s.

`build_sql_config` wraps `auth_bearer`/`oauth2_client_secret` in `SqlConfigValue::Literal`, so a secret submitted to `POST /v1/fluree/sql/map` is stored inline in the graph-source record config. The config layer supports `env_var`/`secret_ref` indirection (`config.rs:76`), but the HTTP path can only store literals — the secret lives at rest in the nameservice record. This matches the Iceberg REST catalog's model, so it's not a regression.

Worth a one-line note in `docs/graph-sources/sql.md` that the record carries the credential and should be protected accordingly — or a follow-up to accept a `secret_ref` through the route (a product call on whether the route should take refs).

fluree-db-sql/src/trino.rs
```diff
@@ -0,0 +1,416 @@
...
+        self.inner
+            .schema_cache
+            .lock()
+            .unwrap()
+            .insert(key, (Instant::now(), Arc::clone(&schema)));
+        Ok(schema)
+    }
+
+    /// Exact `COUNT(*)` with the given columns required non-null.
+    pub async fn count(&self, source: &LogicalSource, non_null_cols: &[String]) -> Result<u64> {
```
aaj3f, line 238: Awareness, not a change request — `row_count` runs a real backend `COUNT(*)`. Cardinality is answered with `SELECT COUNT(*) … WHERE c IS NOT NULL AND …` against the endpoint, so on a large SQL table it's a full aggregate scan per cardinality request, where an Iceberg source uses manifest metadata. Fine as an estimate, and the backend may optimize `COUNT(*)` — just a real query cost the planner should know it's paying if cardinality is requested repeatedly.

aaj3f (approved):
@bplatz interested to learn more about the grounded-reality scenario that made you look into this, but that's purely an academic curiosity. Everything here makes sense and the work is good. 

The one thing you may want to block on before merging: `dialect::sql_string` doesn't escape backslashes, which is a pathway for SQL injection into a MySQL-backed bridge (MySQL processes `\` escapes in string literals under the default `sql_mode`, whereas Trino/Postgres/SQLite don't). A query-author-controlled `FILTER` value or subject-IRI key with a backslash breaks out of the string literal and bypasses the mapping's table boundary. It's a small, localized fix in the bridge, plus a MySQL-backed regression test.

**Adherence checklist**

1. Patterns / abstractions ✔ — extends `FlureeR2rmlProvider`/R2RML operator, `LogicalTable::SqlQuery` alias unifies query-vs-table, Iceberg providers refuse SQL aliases; no parallel construct.
2. Performance ✔ — engine flake loops untouched; per-scan (not per-row) rendering; streaming both sides; per-query client/dispatch/schema caches reduce overhead. One note: `row_count` issues a real backend `COUNT(*)`.
3. Testing ⚠️ — strong hermetic coverage for the client (wiremock) and the SQLite bridge; **no MySQL/Postgres bridge tests**, which is exactly where the injection hides.
4. Conventions ⚠️ — clippy-clean on base lints, thiserror error idiom, self-describing commits; but `fluree-db-sql` is missing `[lints] workspace = true`. And: stacked ⇒ no CI ran — retarget to main before merge.

PR #1750, opened 2026-08-31
fix(query): STR() on a double returns the canonical XSD lexical form

### LINKED ISSUE #1695: `STR()` on an `xsd:double` returns a non-canonical lexical form, disagreeing with our own serializer on the same term
Found while verifying the burn-down backlog at `47e96b7b6`. This is the same defect class #1445 closed for the six RDF output sites, but the `STR()` builtin was not one of them — so today two paths in the same engine render the same term two different ways.

## Repro

```turtle
ex:y1 ex:d 1E0 .
```

```
$ fluree query lex --sparql 'SELECT ?o WHERE { ex:y1 ex:d ?o }' --format json
"value": "1.0E0", "datatype": "http://www.w3.org/2001/XMLSchema#double"     ← correct

$ fluree query lex --sparql 'SELECT (STR(?o) AS ?lex) WHERE { ex:y1 ex:d ?o }' --format json
"value": "1"                                                                ← wrong
```

`STR` is defined to return the literal's lexical form, and the canonical lexical form of that value is `1.0E0` — which our serializer already knows, two lines above in the same result set.

## Mechanism

`eval_str` (`fluree-db-query/src/eval/string.rs:93-109`) routes everything that isn't a `Sid` through `ComparableValue::into_string_value`, and that function formats doubles with Rust's `Display`:

```rust
// fluree-db-query/src/eval/value.rs:500
ComparableValue::Double(d) => Some(ComparableValue::String(Arc::from(d.to_string()))),
```

`f64::to_string()` gives `"1"` for `1.0f64` — no mantissa dot, no exponent. The neighbouring `Float` arm at `:501-503` already routes through a helper (`super::cast::format_f32`), so the shape of the fix is established right there.

#1445's fix added `canonical_xsd_double` in `fluree-graph-ir/src/xsd_double.rs:92` and wired it into `export.rs:1256, :1475`, `format/sparql.rs:431, :567`, `format/sparql_xml.rs:370` and `format/delimited.rs:614`. Expression evaluation was simply never on that list — I don't think it was an oversight so much as that the audit that produced #1445 was scoped to output formats.

## Why it matters

`STR()` feeds `FILTER(STR(?o) = "1.0E0")`, `CONCAT`, `IRI(CONCAT(...))`, `REGEX` over stringified numerics, and `BIND(STR(?o) AS ?k)` group keys. All of those silently get the wrong string — no error, just a filter that never matches or a key that collides with a different value's rendering. And a user comparing `STR(?o)` against the very value we handed them in the result set will find they don't match.

`TypedLiteral`'s inner `Double` arm (`value.rs:519+`) looks like it has the same shape and should be checked in the same pass.

## Suggested direction

Route the `Double` arm (and the `TypedLiteral { val: FlakeValue::Double, .. }` arm) through `canonical_xsd_double`, exactly as the six output sites do. One wrinkle: `fluree-db-query` doesn't currently depend on `fluree-graph-ir` (`fluree-db-query/Cargo.toml:10-18`), so this either adds that dependency or moves the helper somewhere both crates already see — `fluree-db-core` would be the natural home, and moving it also stops the next site from re-deriving it.

Worth a regression test on both surfaces per the parity rule, alongside the existing `sparql_double_canonical_lexical_form_across_formats` (`fluree-db-api/tests/it_query_sparql.rs:4645`) — that test pins the serializer and would not have caught this.

aaj3f:
Fixes #1695.

`STR()` on an `xsd:double` was routing through Rust's `Display`, so `STR(?d)` came back `"1"` for the same term the serializer printed as `"1.0E0"` two fields over in the same result set — and `"inf"` for a value that no RDF format spells that way. #1445 fixed this for the six RDF output sites by adding `canonical_xsd_double`; expression evaluation was simply never on that list, so we've had two renderings of one concept since then.

The mechanism is the one the issue traced: `ComparableValue::into_string_value` (`fluree-db-query/src/eval/value.rs`) formatted the `Double` arm with `d.to_string()`. Rather than derive a second formatter next to it, `fluree-db-query` picks up a dependency on `fluree-graph-ir` and calls the serializer's own `canonical_xsd_double` — the same function `export.rs`, `format/sparql.rs`, `format/sparql_xml.rs` and `format/delimited.rs` already call. The `TypedLiteral { val: FlakeValue::Double, .. }` arm the issue flagged takes it too.

Two siblings turned out to be the same defect and are folded in:

- **`STR()` on an `xsd:float`.** There are two layers here. The visible one is the same as the double's: the serializer gave `1.0E0` where `STR()` gave `1`. The subtle one is that a stored `xsd:float` is carried as a *full-precision* `FlakeValue::Double` (ingest never narrows), and expression evaluation narrows the binding to an f32 on the way into `ComparableValue` (`lit_to_comparable`, and the `EncodedLit` re-tag from #1470) — which is right for arithmetic and `datatype()`, and irreversible for lexical purposes. For an f32-inexact value like `"3.14159265358979"^^xsd:float` the serializer prints the stored f64 (`3.14159265358979E0`), so no rendering of the f32 can ever agree with it. `STR()` therefore reads the stored f64 straight off the binding (`stored_float_f64` in `eval.rs`, covering both the `Lit` and the late-materialized `EncodedLit` lanes) and renders `canonical_xsd_double` — exactly the serializer's string, and no widening artifact can arise there because that f64 was parsed from the lexical, never widened from an f32. A float computed *inside* an expression is a genuine f32 and renders `canonical_xsd_float`, which joins its sibling in `fluree-graph-ir/src/xsd_double.rs` — both share the one `finite_canonical` writer behind a small sealed `XsdFloat` trait that keeps the finite-input contract machine-checked (`debug_assert!`) and the `NaN`/`INF`/`-INF` spellings written exactly once, for the cast renderers too.
- **`GROUP_CONCAT`'s numeric coercion** (`aggregate.rs`), which stringified doubles the same way — `GROUP_CONCAT(?d)` produced `1|1000000` where the serializer renders those same two terms `1.0E0` and `1.0E6`.

**One sibling I deliberately did *not* fold in, and I think this is the more interesting part of the PR.** My first pass also routed the `xsd:string()` cast through the canonical form, on the theory that it was the same seam. The W3C suite said otherwise, and it's right: SPARQL §17.5 defers casting to XPath, whose float/double→string rule casts through `xs:decimal` — plain decimal notation — for absolute values in `[1e-6, 1e6)`, and `data-sparql11/cast/cast-string.srx` requires `xsd:string("1E0"^^xsd:double)` to be `"1"`, not `"1.0E0"`. So the cast and `STR()` are answering genuinely different questions and need to stay apart. Since `cast_to_string` was reaching the canonical form only by falling through to `into_string_value`, it now has explicit `Double`/`Float` arms of its own (`eval/cast.rs`) — which also picks up the one part of the cast that was broken under *every* reading of the spec: it was emitting Rust's `"inf"`/`"-inf"` rather than `INF`/`-INF`.

But the XPath range rule is exactly that — a range, and it is **two-sided**: outside `[1e-6, 1e6)` XPath prescribes the canonical lexical representation, which with the canonical writers in hand is a two-line check rather than "implementing XPath's algorithm". So `format_f32`/`format_f64` apply it: `xsd:string(1.0E-7)` is `1.0E-7` (not `0.0000001`), just as `xsd:string(1.0E30)` is `1.0E30` (not a 31-digit integer). Nothing in the W3C corpus moves in either direction — the suite is 36/0 on both sides of the change; the cast columns in the tests pin both bounds.

The invariant that came out of all this, and the one the tests actually assert, is **`STR(?x)` equals what we serialize for `?x`** — checked per-row against the serialized value of the same binding rather than against a hardcoded string, so the two paths can't drift apart again without a red test.

**What stays open, deliberately, in #1776:** the conflation underneath all three symptoms — float renderers keying on the `Double` *variant* rather than the declared datatype. Concretely: whether the serializer should print an `xsd:float` term at f32 precision at all (a user-visible output decision, same shape as #1445's); the group-key `GROUP_CONCAT` lane, whose aggregate input round-trips through `to_binding` and mints a third spelling; `STR(COALESCE(?f))`, which evaluates the inner expression to an f32 before `STR()` can see the binding; and the stored-vs-computed provenance split at `to_binding` (a computed float mints its XPath-cast lexical — each such term is internally consistent across serializer/`STR`/cast, and switching it to the canonical spelling would break `xsd:string()` on the rebound term, so it stays, documented at the site). The unreachable `TypedLiteral { val: Double, .. }` cast arm carries a guarding comment pointing at the same issue.

Follow-up: #1776.

### Tests

- `sparql_str_double_matches_serializer_canonical_form` sits right beside the existing serializer pin `sparql_double_canonical_lexical_form_across_formats`, so the two can't drift. It covers the issue's repro (`1E0`), an integral value, both exponent directions, a negative, `NaN`/`INF`/`-INF` — and pins the cast column alongside on both sides of the XPath range (`1.0E-7`, `1.0E6`, `1.0E30`), precisely so the deliberate relationship between the two functions is stated rather than left for someone to "fix" later.
- `sparql_str_float_matches_serializer_canonical_form` includes f32-INEXACT rows (`3.14159265358979`, `1.23456789012345E-5`) — the discriminating cases: an f32-exact row passes under any implementation, an inexact one only if `STR()` reads the stored f64.
- `sparql_encoded_xsd_float_str_matches_serializer` pins the same agreement on the late-materialized `EncodedLit` lane (reindexed ledger), next to the #1470 re-tag tests it extends.
- `sparql_group_concat_double_canonical_form` for the aggregate sibling.
- `jsonld_bind_str_double_canonical_form` and `jsonld_bind_str_float_canonical_form` are the JSON-LD twins per the parity rule — `(str ?d)` gives the same canonical string while the value itself stays a native JSON number.
- Unit coverage on `into_string_value` and `canonical_xsd_float` in `fluree-graph-ir`, plus `format_f32`/`format_f64` range pins (`format_f64_xpath_range_is_two_sided`) including the inclusive/exclusive bounds and zero.

Non-vacuity checked by revert, one lever at a time: with the `eval_str` fast path disabled, the inexact float rows and both float twins go red (`left: "3.1415927E0"` / `right: "3.14159265358979E0"`) while the f32-exact rows stay green; with only the `EncodedLit` arm of `stored_float_f64` disabled, the encoded-lane test goes red with the same truncation while the `Lit`-lane test stays green — proving that lane actually fires; with the cast range check disabled, the unit and integration cast pins go red (`left: "0.0000001"`). All green on restore.

### Gates

`-p fluree-db-query` (1488 lib, all green), `-p fluree-graph-ir` (51), `-p fluree-db-api` `grp_query_sparql` (372) and `grp_query` (425), the W3C suite (`36 passed; 0 failed`), workspace check (minus the excluded search crates), clippy `-D warnings` clean on all three touched crates, `cargo fmt --all` last and `testsuite-sparql`'s own fmt clean. No register movement in either direction — `testsuite-sparql/tests/registers/mod.rs` is untouched, which is the honest answer here: the cast register never moved because I corrected the scope rather than registering a regression.

### DIFF
--- a/fluree-db-api/tests/it_query_sparql.rs
+++ b/fluree-db-api/tests/it_query_sparql.rs
@@ -4711,11 +4711,12 @@ async fn sparql_double_canonical_lexical_form_across_formats() {
 /// drift apart again) plus the expected canonical spelling.
 ///
 /// The `xsd:string()` cast is pinned alongside precisely because it must NOT
-/// follow: SPARQL §17.5 defers casting to XPath, whose double→string rule is
-/// plain decimal notation in `[1e-6, 1e6)`, and W3C `cast-string` requires
-/// `xsd:string("1E0"^^xsd:double)` to be `"1"`. The two functions answer
-/// different questions; this test states both answers so neither drifts into
-/// the other.
+/// always follow: SPARQL §17.5 defers casting to XPath, whose double→string
+/// rule is plain decimal notation for absolute values in `[1e-6, 1e6)` — W3C
+/// `cast-string` requires `xsd:string("1E0"^^xsd:double)` to be `"1"` — and
+/// the canonical (scientific) lexical form outside that range, on BOTH sides
+/// (`1.0E-7` and `1.0E30` alike, never `0.0000001` or a 31-digit integer).
+/// This test states both answers so neither function drifts into the other.
 #[tokio::test]
 async fn sparql_str_double_matches_serializer_canonical_form() {
     let fluree = FlureeBuilder::memory().build_memory();
@@ -4727,9 +4728,15 @@ async fn sparql_str_double_matches_serializer_canonical_form() {
     let cases = [
         ("ex:d1", "1E0", "1.0E0", "1"),
         ("ex:d2", "5.0", "5.0E0", "5"),
-        ("ex:d3", "1.0E6", "1.0E6", "1000000"),
+        // 1e6 sits just OUTSIDE the XPath decimal-notation range (the upper
+        // bound is exclusive), so cast and canonical agree here.
+        ("ex:d3", "1.0E6", "1.0E6", "1.0E6"),
         ("ex:d4", "0.001", "1.0E-3", "0.001"),
         ("ex:d5", "-12.5", "-1.25E1", "-12.5"),
+        // Outside the XPath range on either side the cast is the canonical
+        // scientific form too — the range rule is two-sided.
+        ("ex:d9", "1.0E-7", "1.0E-7", "1.0E-7"),
+        ("ex:d10", "1.0E30", "1.0E30", "1.0E30"),
         // Specials share one spelling across both renderings — and neither is
         // Rust's `Display` ("inf"), which is a lexical form of nothing.
         ("ex:d6", "NaN", "NaN", "NaN"),
@@ -4793,10 +4800,13 @@ async fn sparql_str_double_matches_serializer_canonical_form() {
     }
 }
 
-/// Issue #1695 (float sibling): a stored `xsd:float` materializes as an f32 in
-/// expression evaluation, and `STR()` on it must likewise agree with the
-/// serializer's rendering of the same term (canonical form of the stored
-/// value).
+/// Issue #1695 (float sibling): a stored `xsd:float` is carried as a
+/// full-precision f64 (ingest never narrows) and the serializer prints THAT
+/// value; expression evaluation narrows the same binding to an f32. `STR()`
+/// must side with the serializer — it reads the stored f64 off the binding
+/// (`stored_float_f64`) rather than spelling the f32 truncation. The
+/// f32-inexact rows are the ones that catch it: for `3.14159265358979` the
+/// truncated spelling would be `3.1415927E0`.
 #[tokio::test]
 async fn sparql_str_float_matches_serializer_canonical_form() {
     let fluree = FlureeBuilder::memory().build_memory();
@@ -4806,6 +4816,12 @@ async fn sparql_str_float_matches_serializer_canonical_form() {
         ("ex:f1", "1E0", "1.0E0"),
         ("ex:f2", "33.33", "3.333E1"),
         ("ex:f3", "-0.5", "-5.0E-1"),
+        // f32-INEXACT: the stored f64 keeps more precision than an f32 can
+        // hold, so serializer-vs-STR() agreement is only possible if STR()
+        // reads the stored f64. (These rows go red if STR() ever routes
+        // through the f32 materialization again.)
+        ("ex:f4", "3.14159265358979", "3.14159265358979E0"),
+        ("ex:f5", "1.23456789012345E-5", "1.23456789012345E-5"),
     ];
     let graph: Vec<serde_json::Value> = cases
         .iter()
--- a/fluree-db-api/tests/it_query_sparql.rs
+++ b/fluree-db-api/tests/it_query_sparql.rs
@@ -4703,6 +4703,196 @@ async fn sparql_double_canonical_lexical_form_across_formats() {
     );
 }
 
+/// Issue #1695: `STR()` on an `xsd:double` must return the same W3C canonical
+/// lexical form the serializer emits for the very same term — the companion
+/// of `sparql_double_canonical_lexical_form_across_formats`, which pins the
+/// serializer side but never exercises expression evaluation. Each row
+/// asserts `STR(?d) == serialized ?d` (self-anchoring: the two paths cannot
+/// drift apart again) plus the expected canonical spelling.
+///
+/// The `xsd:string()` cast is pinned alongside precisely because it must NOT
+/// follow: SPARQL §17.5 defers casting to XPath, whose double→string rule is
+/// plain decimal notation in `[1e-6, 1e6)`, and W3C `cast-string` requires
+/// `xsd:string("1E0"^^xsd:double)` to be `"1"`. The two functions answer
+/// different questions; this test states both answers so neither drifts into
+/// the other.
+#[tokio::test]
+async fn sparql_str_double_matches_serializer_canonical_form() {
+    let fluree = FlureeBuilder::memory().build_memory();
+    let ledger0 = genesis_ledger(&fluree, "canon:str-double");
+
+    // Representative doubles: the issue's repro (1E0), an integral value,
+    // a large exponent, a small exponent, a negative, and the specials.
+    // Columns: id, stored lexical, canonical (STR + serializer), cast form.
+    let cases = [
+        ("ex:d1", "1E0", "1.0E0", "1"),
+        ("ex:d2", "5.0", "5.0E0", "5"),
+        ("ex:d3", "1.0E6", "1.0E6", "1000000"),
+        ("ex:d4", "0.001", "1.0E-3", "0.001"),
+        ("ex:d5", "-12.5", "-1.25E1", "-12.5"),
+        // Specials share one spelling across both renderings — and neither is
+        // Rust's `Display` ("inf"), which is a lexical form of nothing.
+        ("ex:d6", "NaN", "NaN", "NaN"),
+        ("ex:d7", "INF", "INF", "INF"),
+        ("ex:d8", "-INF", "-INF", "-INF"),
+    ];
+    let graph: Vec<serde_json::Value> = cases
+        .iter()
+        .map(|(id, lex, _, _)| {
+            json!({"@id": id, "ex:score": {"@value": lex, "@type": "xsd:double"}})
+        })
+        .collect();
+    let insert = json!({
+        "@context": {
+            "ex": "http://example.org/ns/",
+            "xsd": "http://www.w3.org/2001/XMLSchema#"
+        },
+        "@graph": graph
+    });
+    let ledger = fluree.insert(ledger0, &insert).await.expect("seed").ledger;
+
+    let query = r"
+        PREFIX ex: <http://example.org/ns/>
+        PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+        SELECT ?s ?d (STR(?d) AS ?lex) (xsd:string(?d) AS ?cast)
+        WHERE { ?s ex:score ?d }
+    ";
+    let result = support::query_sparql(&fluree, &ledger, query)
+        .await
+        .expect("STR(double) query");
+    let sparql_json = result
+        .to_sparql_json(&ledger.snapshot)
+        .expect("to_sparql_json");
+    let bindings = sparql_json["results"]["bindings"]
+        .as_array()
+        .expect("bindings array");
+    assert_eq!(bindings.len(), cases.len(), "{sparql_json}");
+
+    for row in bindings {
+        let subject = row["s"]["value"].as_str().expect("subject IRI");
+        let serialized = row["d"]["value"].as_str().expect("serialized double");
+        let str_lex = row["lex"]["value"].as_str().expect("STR result");
+        let cast_lex = row["cast"]["value"].as_str().expect("xsd:string result");
+        // The core pin: expression evaluation agrees with the serializer on
+        // the SAME term in the SAME result set.
+        assert_eq!(
+            str_lex, serialized,
+            "STR(?d) diverges from the serializer for {subject}: {sparql_json}"
+        );
+        // And the shared form is the W3C canonical one.
+        let (_, _, canonical, cast) = cases
+            .iter()
+            .find(|(id, _, _, _)| subject.ends_with(&id[3..]))
+            .expect("known subject");
+        assert_eq!(serialized, *canonical, "{subject}: {sparql_json}");
+        // The cast keeps its own XPath-mandated rendering (W3C cast-string).
+        assert_eq!(
+            cast_lex, *cast,
+            "xsd:string(?d) must follow the XPath cast rule for {subject}: {sparql_json}"
+        );
+    }
+}
+
+/// Issue #1695 (float sibling): a stored `xsd:float` materializes as an f32 in
+/// expression evaluation, and `STR()` on it must likewise agree with the
+/// serializer's rendering of the same term (canonical form of the stored
+/// value).
+#[tokio::test]
+async fn sparql_str_float_matches_serializer_canonical_form() {
+    let fluree = FlureeBuilder::memory().build_memory();
+    let ledger0 = genesis_ledger(&fluree, "canon:str-float");
+
+    let cases = [
+        ("ex:f1", "1E0", "1.0E0"),
+        ("ex:f2", "33.33", "3.333E1"),
+        ("ex:f3", "-0.5", "-5.0E-1"),
+    ];
+    let graph: Vec<serde_json::Value> = cases
+        .iter()
+        .map(|(id, lex, _)| json!({"@id": id, "ex:reading": {"@value": lex, "@type": "xsd:float"}}))
+        .collect();
+    let insert = json!({
+        "@context": {
+            "ex": "http://example.org/ns/",
+            "xsd": "http://www.w3.org/2001/XMLSchema#"
+        },
+        "@graph": graph
+    });
+    let ledger = fluree.insert(ledger0, &insert).await.expect("seed").ledger;
+
+    let query = r"
+        PREFIX ex: <http://example.org/ns/>
+        SELECT ?s ?f (STR(?f) AS ?lex)
+        WHERE { ?s ex:reading ?f }
+    ";
+    let result = support::query_sparql(&fluree, &ledger, query)
+        .await
+        .expect("STR(float) query");
+    let sparql_json = result
+        .to_sparql_json(&ledger.snapshot)
+        .expect("to_sparql_json");
+    let bindings = sparql_json["results"]["bindings"]
+        .as_array()
+        .expect("bindings array");
+    assert_eq!(bindings.len(), cases.len(), "{sparql_json}");
+
+    for row in bindings {
+        let subject = row["s"]["value"].as_str().expect("subject IRI");
+        let serialized = row["f"]["value"].as_str().expect("serialized float");
+        let str_lex = row["lex"]["value"].as_str().expect("STR result");
+        assert_eq!(
+            str_lex, serialized,
+            "STR(?f) diverges from the serializer for {subject}: {sparql_json}"
+        );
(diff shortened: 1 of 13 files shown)


fluree-db-query/src/eval/value.rs
```diff
@@ -497,9 +497,16 @@ impl ComparableValue {
...
+            // canonical form the RDF-lexical serializers emit for the same
+            // term (#1695; the serializer side was #1445). `f64::to_string()`
+            // gave "1" for 1.0 and "inf" for INF, so `STR(?d)` disagreed with
+            // the value printed two fields over in the same result set.
+            ComparableValue::Double(d) => Some(ComparableValue::String(Arc::from(
+                fluree_graph_ir::canonical_xsd_double(d),
+            ))),
             ComparableValue::Float(f) => Some(ComparableValue::String(Arc::from(
-                super::cast::format_f32(f),
+                fluree_graph_ir::canonical_xsd_float(f),
```
bplatz, line 516: `STR()` on a stored `xsd:float` still diverges from the serializer. Floats are stored as `FlakeValue::Double(f64)` and ingest never narrows, so `"3.14159265358979"^^xsd:float` serializes as `3.14159265358979E0` while `STR()` returns `3.1415927E0` — `coerce_numeric_operand`'s `d as f32` truncates first. `GROUP_CONCAT` keeps the f64 and agrees with the serializer, so one term now has three renderings.

Simplest close: don't downcast on this path. `canonical_xsd_double` on the stored f64 matches the serializer exactly, and the widening artifact the new doc warns about can't arise here — the f64 came from parsing the lexical, not from widening an f32. The conflation underneath (serializers key on the `Double` variant, not the declared datatype) looks like follow-up issue material.

fluree-db-api/tests/it_query_sparql.rs
```diff
@@ -4703,6 +4703,196 @@ async fn sparql_double_canonical_lexical_form_across_formats() {
...
+/// serializer's rendering of the same term (canonical form of the stored
+/// value).
+#[tokio::test]
+async fn sparql_str_float_matches_serializer_canonical_form() {
+    let fluree = FlureeBuilder::memory().build_memory();
+    let ledger0 = genesis_ledger(&fluree, "canon:str-float");
+
+    let cases = [
+        ("ex:f1", "1E0", "1.0E0"),
+        ("ex:f2", "33.33", "3.333E1"),
```
bplatz, line 4817: All three cases are f32-exact, so `str_lex == serialized` can't fail here. `3.14159265358979` or `1.23456789012345E-5` breaks it (see the note on `eval/value.rs`). Worth one f32-inexact case whichever way the fix goes.

fluree-db-query/src/eval/value.rs
```diff
@@ -571,7 +581,11 @@ impl ComparableValue {
                 datatypes.xsd_double.clone(),
             )),
             // xsd:float renders its lexical form from the f32 (avoiding f64
-            // display artifacts) and carries the xsd:float datatype.
+            // display artifacts) and carries the xsd:float datatype. This
+            // CONSTRUCTS a float literal, so it keeps the cast-side rendering
+            // (`format_f32`) rather than STR()'s canonical form — `STR()` of
+            // the result still agrees with how that result serializes, which
+            // is the invariant #1695 is about.
             ComparableValue::Float(f) => Ok(Binding::lit(
```
bplatz, line 606: This builds a term's lexical form rather than a cast result, so `canonical_xsd_float` is arguably what belongs here: `BIND(?f + 0 AS ?y)` over a stored `1.0E0` float yields `"1"^^xsd:float`, so the same value spells two ways depending on provenance. `STR()` does agree with the serializer per-term, but on a non-canonical spelling — the class of thing #1445/#1695 exist to remove. Fine to keep deliberately; the comment should then say that.

fluree-db-query/src/eval/cast.rs
```diff
@@ -206,6 +211,25 @@ pub(crate) fn format_f32(f: f32) -> String {
     }
 }
 
+/// `f64` counterpart of [`format_f32`], for the `xsd:string()` cast of a
+/// double. Rust's `Display` supplies the decimal mantissa; the special values
+/// take their XSD spellings — `f64::to_string()` alone yields `"inf"`, which
+/// is not a valid lexical form under either the cast rules or the canonical
+/// ones (#1695).
+fn format_f64(d: f64) -> String {
```
bplatz, line 220: On the range gap you flagged in the description: it's two-sided. `xsd:string(1.0E-7)` returns `0.0000001` where XPath wants `1.0E-7`, same rule as the `>= 1e6` side. Worth filing — mention both bounds.

fluree-db-query/src/eval/cast.rs
```diff
@@ -320,6 +344,21 @@ fn cast_to_string(
         let s = canonical_decimal_string(d);
         return Some(ComparableValue::String(Arc::from(s)));
     }
+    // Doubles and floats likewise render by the XPath *cast* rule, not the
+    // canonical XSD lexical form `STR()` returns — see `format_f32` (#1695).
+    // Without these arms the cast would fall through to `into_string_value`
+    // and pick up STR()'s canonical form, failing W3C `cast-string`.
+    match &v {
+        ComparableValue::Double(d)
+        | ComparableValue::TypedLiteral {
```
bplatz, line 360: Nit: this arm ignores `dtc`, so a float arriving as `TypedLiteral { val: Double, .. }` takes the f64 branch while `Float` takes f32. Doesn't reproduce today (floats reach the cast as `Float`), but it's the same variant-vs-datatype conflation as the `STR()` issue.

fluree-graph-ir/src/xsd_double.rs
```diff
@@ -54,18 +54,21 @@ impl fmt::Write for StackBuf {
...
 /// zeros. Canonicalization is then purely syntactic: ensure the mantissa
 /// contains a `.` (insert `.0`) and uppercase the exponent marker.
-fn finite_canonical(d: f64) -> StackBuf {
-    debug_assert!(d.is_finite());
-
+///
+/// Callers must exclude NaN/±INF first — their `{:e}` forms carry no
+/// exponent, and their XSD spellings (`NaN`, `INF`, `-INF`) differ from
+/// Rust's.
+fn finite_canonical<F: fmt::LowerExp>(d: F) -> StackBuf {
```
bplatz: Nit: `debug_assert!(d.is_finite())` went away with the generic, so the contract is doc-only now. Both public wrappers guard, so nothing is broken — a sealed trait, or an assert at the two call sites, restores the machine check.

fluree-graph-ir/src/xsd_double.rs
```diff
@@ -103,6 +106,29 @@ pub fn canonical_xsd_double(d: f64) -> String {
...
+/// The canonical XSD lexical form of an `xsd:float` value, as a `String`.
+///
+/// Same grammar as [`canonical_xsd_double`] — mantissa with a decimal point,
+/// uppercase `E`, no `+`/leading zeros in the exponent, and the XSD special
+/// spellings `NaN`/`INF`/`-INF` — but the mantissa is the f32
+/// shortest-round-trip form, so a single-precision value never picks up
+/// double-precision widening artifacts (`33.33f32` stays `3.333E1`, not
+/// `3.333000183105469E1`).
+#[must_use]
+pub fn canonical_xsd_float(f: f32) -> String {
```
bplatz, line 184: Nit: the NaN/INF spelling is now written four times (here, `canonical_xsd_double`, `format_f32`, `format_f64`). One helper would do.

bplatz (approved):
Approving — the double side is solid. I reproduced the gates locally (graph-ir + `fluree-db-query --lib` green, W3C 36/0, clippy/fmt clean) and the non-vacuity claim: both `sparql_str_*` tests go red on revert. Keeping the cast separate is the right call and well evidenced.

Please look at the inline points before merging, though — the float sibling isn't actually closed. `STR()` still disagrees with the serializer for any stored `xsd:float` that isn't f32-exact, and the new float test can't catch it. Some of the rest are probably follow-up issues rather than changes here; fine either way as long as they get filed.

fluree-db-query/src/eval/value.rs
aaj3f: Right on the defect — and one refinement on the mechanism, because it ended up shaping the fix. For a stored float the truncation isn't in `coerce_numeric_operand` (that path only handles the string-backed cast/STRDT literals); it's at binding→`ComparableValue` conversion itself — `lit_to_comparable` (`eval.rs`) and the #1470 `EncodedLit` re-tag. Which means "don't downcast on this path" can't be done inside `into_string_value`: by the time the `Float` arm runs the f64 is gone, and widening it back is exactly the artifact case. And the general arm genuinely can't call `canonical_xsd_double` either — your "the f64 came from parsing the lexical" guarantee holds for a stored term, but a float computed inside the expression reaches the same arm as a widened f32, where `33.33f32 * 1` would spell itself `3.3330001831054688E1`.

So the shape that landed: `eval_str` reads the stored f64 straight off the binding for the bare-variable shape (`stored_float_f64` in `eval.rs`, covering both the `Lit` and `EncodedLit` lanes — same pattern as `eval_lang`'s existing binding fast path), and renders `canonical_xsd_double`; computed floats keep `canonical_xsd_float`. Verified live: stored `"3.14159265358979"^^xsd:float` — serializer `3.14159265358979E0`, `STR()` was `3.1415927E0`, now equal, on both lanes.

One wrinkle your GROUP_CONCAT sub-claim surfaced when I ran it: it agrees with the serializer on the plain lane (`GROUP BY ?s` → `agg_group_concat`'s `Double` arm → `3.14159265358979E0`), but `GROUP_CONCAT(?f) … GROUP BY ?f` gives a *third* spelling, `3.1415927` — the group-key lane round-trips the aggregate input through expression eval → `to_binding` → `format_f32`. That one, plus `STR(COALESCE(?f))` (the inner expression evaluates to an f32 before `STR()` can see the binding) and the serializer-precision question itself, are the umbrella you called: filed as #1776.

fluree-db-api/tests/it_query_sparql.rs
aaj3f: Confirmed exactly as you read it — with the fix reverted, the original three rows all stay green (they pass under either implementation) and only an inexact row discriminates. Added both of your candidates (`3.14159265358979`, `1.23456789012345E-5`); red without the fix (`left: "3.1415927E0"` / `right: "3.14159265358979E0"`), green with it. Also added the same inexact value on the `EncodedLit` lane (`sparql_encoded_xsd_float_str_matches_serializer`, reindexed ledger, next to the #1470 re-tag tests) since the memory-ledger tests never touch that lane — and proved it actually fires by disabling only the encoded arm of `stored_float_f64`: the encoded test goes red with the same truncation while the `Lit`-lane test stays green.

fluree-db-query/src/eval/value.rs
aaj3f: Reproduced as described: stored `"1.0E0"^^xsd:float` serializes `1.0E0`, `BIND(?f + 0 AS ?y)` yields `"1"^^xsd:float`. Two spellings by provenance, exactly.

I did try the switch before deciding, and it breaks something concrete: a minted float binding is a string-backed literal, and `xsd:string(?y)` reads that lexical verbatim (the `TypedLiteral{String}` fall-through in `cast_to_string`). Today that's `"1"` — which is what XPath requires for the cast, verified at head — so minting `"1.0E0"` instead makes the *rebound* cast wrong, and fixing that would mean teaching `cast_to_string` to parse float-typed strings back out, i.e. the split just moves. So: each minted term is internally consistent (serializer, `STR()`, and `xsd:string()` of it all return the one string), the split is between two terms rather than two renderings of one term — but you're right that the value-level split is the #1445/#1695 class. Kept deliberately, and the comment now says exactly that, with the rebound-cast reasoning inline. Cross-provenance canonicalization rides with the serializer-precision decision in #1776 — if we ever narrow the serializer to f32 by datatype, this arm collapses into it for free.

fluree-db-query/src/eval/cast.rs
aaj3f: Verified both bounds by execution at head: `xsd:string(1.0E-7)` → `0.0000001`, `xsd:string(1.0E30)` → the 31-digit integer. Two-sided, as you said.

Folded rather than filed, though — the "implementing XPath's algorithm" cost I punted on in the description had quietly collapsed once `canonical_xsd_double`/`canonical_xsd_float` existed: outside `[1e-6, 1e6)` XPath prescribes the canonical lexical representation, which is precisely those writers, so the whole fix is a range check in `format_f32`/`format_f64`. Unit pins on the bounds (1e-6 inclusive stays decimal, 1e6 exclusive goes scientific, zero stays `"0"`), integration cast columns pin `1.0E-7`/`1.0E6`/`1.0E30`, W3C still 36/0 — nothing in the corpus reaches either side. PR body updated to match.

fluree-db-query/src/eval/cast.rs
aaj3f: Confirmed — and it's unreachable by construction, not just unreproduced: I went through every `TypedLiteral` producer, and none wraps a `Double` (STRDT/STRLANG/casts wrap strings, `TryFrom<FlakeValue>` maps `Double` to the bare variant, `unwrap_typed_literal` extracts it back out, the temporal family is the only non-string wrap). Execution agrees — both float shapes take other arms (stored → `Float` → `3.1415927`, STRDT-minted → string-backed → verbatim). Left the arm as belt-and-braces and added the guarding comment naming the variant-vs-datatype conflation and pointing at #1776, so if a producer ever appears the arm's obligation to split on the datatype is written down where it'll be found.

fluree-graph-ir/src/xsd_double.rs
aaj3f: Took the sealed-trait shape — it turned out to pay for itself twice: `XsdFloat` (private `sealed` module, f32/f64 only) hands `finite_canonical` its `debug_assert!` back, and it's the same trait the NaN/INF dedupe from your other note hangs off, so it isn't speculative machinery. Call-site asserts would've been vacuous here anyway — each wrapper branches on the specials one line above the call.

fluree-graph-ir/src/xsd_double.rs
aaj3f: Done — `XsdFloat::nonfinite_xsd()`, a default method on the sealed trait returning `Option<&'static str>`, so the spelling exists exactly once and costs nothing. Wired through all four formatters here plus `format_f32`/`format_f64` in `eval/cast.rs` — your four sites turned out to be six once I counted the push/write variants, all on the one helper now.

PR #1752, opened 2026-08-31
split debug info out of dev-profile binaries

zonotope:
Cargo's dev profile defaults to packed DWARF, so every integration test binary carries its own copy of the debug info from the workspace rlibs it links.

split-debuginfo = "unpacked" keeps the full DWARF but stores it in .dwo files instead of duplicating it into each linked binary. No debugging capability is lost, and links get faster.

Applies to `cargo test` as well, since the test profile inherits dev.


bplatz (approved):
One thing worth addressing: `split-debuginfo = "unpacked"` is only supported on Linux and Apple targets — on `windows-msvc` stable rustc rejects it as a hard error, so this breaks `cargo build`/`cargo test` for anyone on Windows (release/dist builds are unaffected since they don't set it). Since Cargo has no per-platform profile syntax, consider moving this to `CARGO_PROFILE_DEV_SPLIT_DEBUGINFO=unpacked` in the CI workflow `env:` instead — one line covers both the workspace and testsuite-sparql jobs, macOS already defaults to unpacked, and Windows never sees it.

zonotope:
> One thing worth addressing: `split-debuginfo = "unpacked"` is only supported on Linux and Apple targets — on `windows-msvc` stable rustc rejects it as a hard error, so this breaks `cargo build`/`cargo test` for anyone on Windows (release/dist builds are unaffected since they don't set it). Since Cargo has no per-platform profile syntax, consider moving this to `CARGO_PROFILE_DEV_SPLIT_DEBUGINFO=unpacked` in the CI workflow `env:` instead — one line covers both the workspace and testsuite-sparql jobs, macOS already defaults to unpacked, and Windows never sees it.

Cargo queries the target's supported values and omits `-C split-debuginfo` entirely when the target doesn't support it.

Building `--target x86_64-pc-windows-msvc` on 1.97.0, the `rustc` invocation Cargo constructs is `-C embed-bitcode=no -C debuginfo=2 ... --target x86_64-pc-windows-msvc` (no `-C split-debuginfo`). Same on wasm32-unknown-unknown. The Linux invocation by contrast carries `-C debuginfo=2 -C split-debuginfo=unpacked`.

Since rust-toolchain.toml pins 1.97.0 and CI uses dtolnay/rust-toolchain@1.97.0, every developer gets that gating behavior.

I'd rather keep it in the manifest than move to `CARGO_PROFILE_DEV_SPLIT_DEBUGINFO` in CI env because the env var only helps CI and the saving that matters is local. This would give it back for every Linux and macOS developer. It also wouldn't add safety, since an env-set profile value goes through the same resolution path.

PR #1753, opened 2026-08-31
fix(import): publish bulk-import commit heads via ref fast-forward, not the queue contract

bplatz:
## Problem

Bulk import fails on every publish when the nameservice is the raft data plane:

```
storage: Storage error: publish_commit on <ledger>:main: per-branch queue is empty — nothing staged for this branch
```

`RefPublisher::publish_commit` means two different things depending on the implementation. The file/S3/Dynamo nameservices implement it as a plain strictly-newer head update — the semantics `fluree-db-api`'s bulk importer was written against. The raft nameservice (added with the consensus work) implements it as the second half of the transactor worker's two-phase protocol: *apply the staged front of the per-branch queue*. Bulk import is a one-shot admin flow that stages nothing, so under raft its very first head publish hits `peek_queue_front_id` on an empty queue and the whole import is rolled up as failed — after the chunks committed fine.

The combination simply had never been exercised: imports work everywhere the older nameservices run, and the raft posture is new. First hit in the standalone server with `FLUREE_AI_DATA_RAFT=true`, where the embedded importer is the only load path.

## Fix

Import stops calling the queue-contract verb it never satisfied and advances the ref directly: a new `publish_import_commit_head` helper wraps `RefPublisher::fast_forward_commit` and replaces the four `publish_commit` call sites (three periodic checkpoints, one final head).

- **Same semantics file mode always gave import**: advance if strictly newer. The one benign conflict is the final publish restating the last checkpoint's exact `(t, commit_id)` — import runs against a fresh ledger and publishes serially, so any other conflict (equal `t` under a different id, or a head past `t`) is a foreign writer and errors loudly; a head still behind `t` reports CAS-contention exhaustion. (Narrowed from equal-or-newer per review.)
- **Creation is the degenerate first fast-forward**: the importer `init()`s the ledger, `get_ref` reports the unborn head, and the CAS lands `t=1`. On the raft side this is `Command::CompareAndSetRef`; raft's `get_ref` reports a registered-but-unborn branch as `Some(RefValue { id: None, t: 0 })` (never `None`), so the creation CAS carries that as `expected` — pinned by the new `cas_ref_commit_initial_creation_with_expected_unborn_ref_value` state-machine test alongside the existing `expected = None` variant.
- **The raft queue contract is untouched** — `publish_commit` keeps protecting the transactor worker's stage→publish protocol; import just no longer pretends to be a worker.
- Works uniformly across every `NameServicePublisher` (file, S3, memory, raft), so there is no per-backend branching in the importer.

## Testing

- New unit test `publish_import_commit_head_tests::creates_advances_tolerates_exact_republish_rejects_divergence` covers the publish shapes the pipeline produces — creation on a fresh ledger, checkpoint fast-forward, exact `(t, commit_id)` republish as a no-op success — plus the two divergence shapes it never produces (equal `t` under a different id, stale `t`), which now fail loudly.
- Root-caused live on a standalone raft-mode server where every UI `.ttl` import failed with the queue error above (chunks committed, publish refused); the raft-side CAS behavior the fix relies on is pinned by the existing state-machine tests.
- `cargo check` / `cargo clippy` / the new test green on this branch.

### DIFF
--- a/fluree-db-api/src/import.rs
+++ b/fluree-db-api/src/import.rs
@@ -43,6 +43,45 @@ use std::sync::Arc;
 use std::time::Instant;
 use tracing::Instrument;
 
+/// Publish the import's commit head by fast-forwarding the ref directly.
+///
+/// Bulk import is a one-shot admin flow: nothing is staged on the
+/// per-branch transaction queue, so `publish_commit` — whose contract on
+/// queue-backed nameservices (the raft data plane) is "apply the staged
+/// queue front" — fails there with "per-branch queue is empty". A
+/// ref-level fast-forward carries the exact strictly-newer semantics the
+/// file/S3 nameservices give `publish_commit`, and every nameservice
+/// implements it via `RefPublisher`, so the bulk importer works under
+/// all data planes without touching the transactor's queue contract.
+///
+/// An equal-or-newer head is a no-op success, mirroring `publish_commit`:
+/// the final publish re-states the last checkpoint's `t`, and a re-run
+/// may find its own earlier head already in place.
+async fn publish_import_commit_head(
+    nameservice: &dyn crate::NameServicePublisher,
+    alias: &str,
+    t: i64,
+    commit_id: &ContentId,
+) -> std::result::Result<(), ImportError> {
+    use fluree_db_nameservice::{CasResult, RefValue};
+
+    let new_ref = RefValue {
+        id: Some(commit_id.clone()),
+        t,
+    };
+    match nameservice
+        .fast_forward_commit(alias, &new_ref, 5)
+        .await
+        .map_err(|e| ImportError::Storage(e.to_string()))?
+    {
+        CasResult::Updated => Ok(()),
+        CasResult::Conflict { actual } if actual.as_ref().is_some_and(|a| a.t >= t) => Ok(()),
+        CasResult::Conflict { actual } => Err(ImportError::Storage(format!(
+            "import commit-head publish for {alias} lost a race: head is {actual:?}, wanted t={t}"
+        ))),
+    }
+}
+
 const IMPORT_PIPELINE_WAIT_LOG_THRESHOLD_MS: u128 = 50;
 const LOCAL_RECHUNK_EVENT_CHANNEL_CAPACITY: usize = 2;
 
@@ -4180,10 +4219,13 @@ where
                 if env.config.publish_every > 0
                     && (next_expected + 1).is_multiple_of(env.config.publish_every)
                 {
-                    env.nameservice
-                        .publish_commit(env.alias, result.t, &result.commit_id)
-                        .await
-                        .map_err(|e| ImportError::Storage(e.to_string()))?;
+                    publish_import_commit_head(
+                        env.nameservice,
+                        env.alias,
+                        result.t,
+                        &result.commit_id,
+                    )
+                    .await?;
                     tracing::info!(
                         t = result.t,
                         chunk = next_expected + 1,
@@ -5050,10 +5092,7 @@ where
                 elapsed_secs: run_start.elapsed().as_secs_f64(),
             });
             if config.publish_every > 0 && (idx + 1).is_multiple_of(config.publish_every) {
-                nameservice
-                    .publish_commit(alias, result.t, &result.commit_id)
-                    .await
-                    .map_err(|e| ImportError::Storage(e.to_string()))?;
+                publish_import_commit_head(nameservice, alias, result.t, &result.commit_id).await?;
             }
         }
 
@@ -5312,10 +5351,8 @@ where
                     elapsed_secs: run_start.elapsed().as_secs_f64(),
                 });
                 if config.publish_every > 0 && (i + 1).is_multiple_of(config.publish_every) {
-                    nameservice
-                        .publish_commit(alias, result.t, &result.commit_id)
-                        .await
-                        .map_err(|e| ImportError::Storage(e.to_string()))?;
+                    publish_import_commit_head(nameservice, alias, result.t, &result.commit_id)
+                        .await?;
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


fluree-db-api/src/import.rs
```diff
@@ -43,6 +43,45 @@ use std::sync::Arc;
...
+    let new_ref = RefValue {
+        id: Some(commit_id.clone()),
+        t,
+    };
+    match nameservice
+        .fast_forward_commit(alias, &new_ref, 5)
+        .await
+        .map_err(|e| ImportError::Storage(e.to_string()))?
+    {
+        CasResult::Updated => Ok(()),
```
zonotope, line 80: This accepts any conflict where actual.t >= t. But import is gated on a fresh ledger (import.rs:3634), publishes are serial and monotonic, and the final publish uses `state.t`/`state.parent`, which are exactly the last chunk's (t, commit_id). So:
  - The only reachable benign case is equal t with the identical commit id
    (final publish restating the last checkpoint).
  - The `a.t > t` case is unreachable by construction. If it ever happens,
    it means another writer got ahead of the import mid-run, which the helper
    reports as success.
  - Equal t with a different id means two commits claim the same t. This is a 
    genuine divergence, also reported as success.

We should narrow the guard to `a.t == t && a.id.as_deref() == Some(commit_id)`.
It's free precision, and it turns a real anomaly into a loud failure instead of "import succeeded" over someone else's head.

fluree-db-api/src/import.rs
```diff
@@ -43,6 +43,45 @@ use std::sync::Arc;
...
+        CasResult::Conflict { actual } if actual.as_ref().is_some_and(|a| a.t >= t) => Ok(()),
+        CasResult::Conflict { actual } => Err(ImportError::Storage(format!(
+            "import commit-head publish for {alias} lost a race: head is {actual:?}, wanted t={t}"
+        ))),
```
zonotope: This arm catches two distinct outcomes from fast_forward_commit, because the retry loop returns the same Conflict shape for both:
  - a genuine divergence: `actual.t >= t`, the ref moved past us (but that's
    already absorbed by the guard on line 78, so it never reaches here)
  - retry-budget exhaustion: nameservice/src/lib.rs:1086-1088 falls out of the
    for loop and returns `Conflict { actual: current }` where `actual.t < t`, i.e.
    the fast-forward is still valid, we just lost the CAS 5 times in a row

Only the second can actually reach line 79, so in practice the message is always wrong about what happened. Distinguishing them on actual's t would let it say "contended, gave up after 5 attempts" instead.

zonotope (approved):
This looks good besides two minor suggestions

fluree-db-api/src/import.rs
bplatz: Narrowed in 91a34ca0f to `a.t == t && a.id.as_ref() == Some(commit_id)`. Verified the premise before changing it: `state.parent`/`state.t` are assigned together from the last chunk's `commit_cid`/`new_t` (fluree-db-transact/src/import.rs:1241-1242), so the exact-`(t, id)` republish really is the only benign shape. Equal-`t`-different-id and `a.t > t` now surface as diverged-head errors, and the unit test pins both rejections.

fluree-db-api/src/import.rs
bplatz: Split in 91a34ca0f: `actual.t < t` (and `actual: None`) now report retry-budget exhaustion under contention; anything else reports a diverged head with both sides' `(t, id)`. With the narrowed guard above, the divergence arm is reachable again, so the distinction carries real information rather than correcting an always-wrong message.

fluree-db-api/src/import.rs
```diff
@@ -43,6 +43,71 @@ use std::sync::Arc;
 use std::time::Instant;
 use tracing::Instrument;
 
+/// Publish the import's commit head by fast-forwarding the ref directly.
```
aaj3f: *Follow-up (sibling path; not this PR's to fix now that it's merged).* The same verb with the same precondition is still used by two other one-shot admin flows: the fast-forward branch of merge at `fluree-db-api/src/merge.rs:357` and the final head publish of rebase at `fluree-db-api/src/rebase.rs:603` both call `self.publisher()?.publish_commit(...)` without staging anything on the raft queue.

So under `FLUREE_AI_DATA_RAFT=true` a fast-forward `fluree merge` or a `fluree rebase` should hit `peek_queue_front_id`'s "per-branch queue is empty" exactly as import did — after their blobs are written. I've traced the mechanism, not run it under raft, so treat it as plausible rather than confirmed.

If it holds, lifting `publish_import_commit_head` into a shared `fast_forward_head` the three flows use is the natural fix. Commenting here because `merge.rs` / `rebase.rs` aren't in this diff.

fluree-db-api/src/import.rs
```diff
@@ -43,6 +43,71 @@ use std::sync::Arc;
...
+    let new_ref = RefValue {
+        id: Some(commit_id.clone()),
+        t,
+    };
+    match nameservice
+        .fast_forward_commit(alias, &new_ref, 5)
+        .await
+        .map_err(|e| ImportError::Storage(e.to_string()))?
+    {
+        CasResult::Updated => Ok(()),
```
aaj3f: *Praise.* Narrowing the benign arm to the exact `(t, commit_id)` republish is what turns "import succeeded over someone else's head" into a loud failure; and splitting contention (`actual` unborn or `actual.t < t`) from divergence in the error text is real information now that the divergence arm is reachable.

fluree-db-consensus/src/raft/state_machine.rs
```diff
@@ -4847,6 +4847,35 @@ mod tests {
         assert_eq!(entry.t, 1);
     }
 
+    #[test]
```
aaj3f: *Praise.* Pinning the `Some(RefValue { id: None, t: 0 })` creation shape — the one raft's `get_ref` actually reports for an unborn branch — is the test that protects the import's first publish; the existing `expected = None` test would not have.

